PR #38, opened 2019-06-06
Add Fedora dependency installation instructions

abrown:
This is the alternate dependency installation I had to do to get `iwasm` to compile on my Fedora machine. Also, I collapsed the Ubuntu steps into one line for conciseness.


abrown:
Another thought: how about a `.gitignore` file at the top level ignoring `build` so that first-time users of the instructions do not alter their Git state?

PR #89, opened 2019-08-02
Optimize samples build process

Weining2019:


Weining2019:
I will open another pull request with additional patches together.

PR #94, opened 2019-08-09
Add an IOT app-store demo

luhanzhl:
This app-store demo illustrates how to upload wasm applications, and then install both these locally uploaded apps as well as preloaded wasm apps to the target devices.


xwang98:
please ensure the font files integrated has no copyright issue.

PR #96, opened 2019-08-13
Add an IoT app-store demo

luhanzhl:


luhanzhl:
Already deleted unnecessary files by undoing last commit, and await your further comments.

Issue #44, opened 2019-07-21
Dockerfile for development


beriberikix:
I put together a simple `Dockerfile` so that I can easily develop on my macOS machine, found here: https://github.com/beriberikix/wamr-docker

Is there any interest in including it? I added the `test.c` from the `README` since it isn't in the main repo, but can easily remove it before merging.

xwang98:
It would be nice! Could you please advise how to include it?

beriberikix:
I created #97 with an a file that supports clang and updates to the README. Please leave comments in the PR if you need any corrections!

beriberikix:
Closing with #97

PR #97, opened 2019-08-13
Initial Dockerfile

beriberikix:
Dockerfile supports clang-8 and adds `iwasm` to `bin`. Addressing #44.


xwang98:
@beriberikix It is nice work! Thanks for contributing the patch.

PR #117, opened 2019-09-10
Add parameter module inst for native wrapper functions

wenyongh:
And add asm code of em64/arm/mips version to call native wrapper functions;
Fix some issues of calling wrapper functions;


wenyongh:
And add invokeNative function implemented with asm code of em64/arm/mips version to call native wrapper function;
Fix some issues of calling wrapper function;

PR #130, opened 2019-10-08
Add interpreter commands to get at interpreter specific data.

cmickeyb:
This is just a proposal for a more general API for getting/setting interpreter specific data in the thread data section. This assumes that the application will clean up any allocated memory.

Signed-off-by: Mic Bowman <mic.bowman@intel.com>


wenyongh:
Duplicated patch has been merged, please use the API wasm_runtime_set_custom_data() and wasm_runtime_get_custom_data() instead.

PR #131, opened 2019-10-11
Remove the binding between current thread and module instance

Weining2019:
Remove wasm_export_api.h that may confuse
Implement wasm_runtime_validate_app_str_addr()
Fix bugs of loader and pass more spec cases

Signed-off-by: Weining Lu <weining.x.lu@intel.com>


hangedfish:
@wenyongh After Three years, I found this missing position.
https://github.com/bytecodealliance/wasm-micro-runtime/blob/1e22d1a9e545f6efc8d793bdbe6d75165b62adf4/core/app-mgr/app-manager/watchdog.c#L108

Issue #123, opened 2019-09-17
WAMR relicensing coordination


joshtriplett:
WAMR is currently licensed under the Apache 2.0 license. To better collaborate with other projects like wasmtime, cranelift, lucet, and LLVM, Intel would like to match the license of those projects, which use the [Apache 2.0 license with LLVM exception](https://llvm.org/docs/DeveloperPolicy.html#new-llvm-project-license-framework). The LLVM exception keeps all the properties of the Apache 2.0 license, and improves license compatibility.

In order to make this change, we need your help; we're seeking permission from the non-Intel contributors to WAMR to make this change. The following people have made contributions to WAMR:

- [x] @zhujinzhou
- [x] @yiting16
- [x] @beriberikix
- [x] @jondong
- [x] @0xflotus
- [ ] @hakilebara
- [x] @pcmill
- [ ] @offercomes

If this seems reasonable to you, could you please respond to this issue with a comment? (Just saying "sure" is fine; we'll coordinate the list.)

Once we have approval from the folks listed above, I'll post a PR adding the LLVM exception text to the license.

Thank you again for your contributions to WAMR!

beriberikix:
Sure! Glad to see the collaboration 😁

0xflotus:
Sure :)

yiting16:
Sure.

jondong:
Sure! nice to see the collaboration

zhujinzhou:
sure!!!

pcmill:
Sure, I just made some changes to the README :smiley:

joshtriplett:
Fixed in #137.

PR #133, opened 2019-10-21
Error related to invokeNative_em64.s

oubotong:
There is a compiling error on OSX 10.15 which is supposed to relate to the Mach-O x86-64 ABI for OSX. 
Here is the error message:
```
invokeNative_em64.s:19:5: error: error: unknown directive
    .type invokeNative, @function
```
when using clang from Xcode:
```
Apple clang version 11.0.0 (clang-1100.0.33.8)
Target: x86_64-apple-darwin19.0.0
Thread model: posix
```
The new modification of this file solves the compiling issue. The iwasm tests well after the modification

### DIFF
--- a/core/iwasm/runtime/vmcore-wasm/invokeNative_em64.s
+++ b/core/iwasm/runtime/vmcore-wasm/invokeNative_em64.s
@@ -15,9 +15,9 @@
  */
     .text
     .align 2
-.globl invokeNative
-    .type    invokeNative, @function
-invokeNative:
+.globl _invokeNative
+    //.type    invokeNative, @function
+_invokeNative:
     /*  rdi - function ptr */
     /*  rsi - argv */
     /*  rdx - n_stacks */


wenyongh:
Hi @oubotong, it is nice of you to submit the patch pull request! Not sure whether the patch impacts other platform or not, like Linux. Is there any special MACRO defined in OSX platform? So we can use the macro to control the code.

Another issue is that WAMR will change the license to "Apache 2.0 license with LLVM exception", see #123. So currently it is not convenient to accept the code from outside. Shall we submit the patch by ourselves?

oubotong:
I don't know whether this issue is common among all the Unix like systems since I only tested it on my Mac. Currently it seems that the ABI on OSX with latest Xcode installed doesn't support `.type` and prefers an underline in front of the `.globl` function. Here is a quick guide of writing the 64bit MACRO on OSX: 
```
http://www.idryman.org/blog/2014/12/02/writing-64-bit-assembly-on-mac-os-x/
```
Maybe more tests need to be done on other systems before you create a common patch. 
Just for now, the patch I submitted solves the compiling issue on the latest OSX and hope you can take a look at it and make the right decision for this project.

Again, I appreciate it you guys create this amazing project which I also want to help it become better.

wenyongh:
Hi, I saw some projects like v8 and chromium use macro OS_MACOSX for Mac OSX system, could you help check whether the compiler has defined the macro? If not, could you define it by yourself? So we can add it into the invokeNative_em64.s. I am afraid that removing ".type    invokeNative, @function" in the file may cause compile failure in some platform.

oubotong:
Sure, I'll take a look at the compiler on MACOSX. And I will open a new issue once I find it.

wenyongh:
Yes, thanks a lot, now we have added the macro OS_MACOSX  in invokeNative_em64.s in the latest patch, please update to the latest version.

PR #169, opened 2020-02-19
Add SGX AOT support with SGX SDKv2.8 unpublic sgx_rsrv_mem_mngr.h

qdaoming-intel:


core/shared/platform/linux-sgx/bh_platform.c
```diff
@@ -69,23 +70,48 @@ int bh_vprintf_sgx(const char * format, va_list arg)
...
+        return NULL;
+    }
+    if (prot & MMAP_PROT_READ)
+        mprot |= SGX_PROT_READ;
+    if (prot & MMAP_PROT_WRITE)
+        mprot |= SGX_PROT_WRITE;
+    if (prot & MMAP_PROT_EXEC)
+        mprot |= SGX_PROT_EXEC;
+    st = sgx_tprotect_rsrv_mem(ret, alignedSize, mprot);
+    if (st != SGX_SUCCESS) bh_printf_sgx("bh_mmap(size=%d,prot=0x%x) failed to set protect.",size, prot);
```
wenyongh: Should check return value, if failed, free memory and return NULL.

qdaoming-intel: OK, it is reasonable comment.

PR #171, opened 2020-02-22
sdk build tools and app framework updates

xwang98:


samples/simple/src/iwasm_main.c
```diff
@@ -354,6 +354,25 @@ static host_interface interface = { .send = uart_send, .destroy = uart_destroy }
...
 #endif
 
+
+
+static attr_container_t * read_test_sensor(void * sensor)
+{
+    //luc: for test
+    attr_container_t *attr_obj = attr_container_create("read test sensor data");
+    if (attr_obj) {
+        attr_container_set_string(&attr_obj, "name", "read test sensor");
```
wenyongh: Should check return value, the original code doesn't check it, it may cause memory leak.

wenyongh: I have modified the code to add the return value check.

PR #176, opened 2020-02-23
Fixed C compiler detection error in CMakeList.txt

renjipanicker:
On *nix systems, (${CMAKE_C_COMPILER} resolves to /usr/bin/cc, which is usually a symlink to gcc or clang
The check in its present will not detect if it is a symlink to clang, and the flag gets added, which results in an error message from clang.
Checking if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang") is a better option.

### DIFF
--- a/product-mini/platforms/linux/CMakeLists.txt
+++ b/product-mini/platforms/linux/CMakeLists.txt
@@ -63,7 +63,7 @@ set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra -Wformat -Wformat-security")
 # set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wconversion -Wsign-conversion")
 
 if (WAMR_BUILD_TARGET MATCHES "X86_.*" OR WAMR_BUILD_TARGET STREQUAL "AMD_64")
-  if (NOT (${CMAKE_C_COMPILER} MATCHES ".*clang.*"))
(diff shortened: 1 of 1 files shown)


product-mini/platforms/linux/CMakeLists.txt
```diff
@@ -63,7 +63,7 @@ set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra -Wformat -Wformat-security")
 # set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wconversion -Wsign-conversion")
 
 if (WAMR_BUILD_TARGET MATCHES "X86_.*" OR WAMR_BUILD_TARGET STREQUAL "AMD_64")
-  if (NOT (${CMAKE_C_COMPILER} MATCHES ".*clang.*"))
+  if (NOT (CMAKE_C_COMPILER_ID STREQUAL "Clang"))
```
wenyongh: CMAKE_C_COMPILER_ID may be:
  AppleClang,  means Apple Clang (apple.com)
  Clang, means LLVM Clang (clang.llvm.org)
Seems it is better to use MACHES "*Clang" but not  STREQUAL.
Also the CMAKE_C_COMPILER_ID may be unuset if we set (CMAKE_C_COMPILER "clang") in this file.
So we had better check both of them:
  if (NOT (CMAKE_C_COMPILER MATCHES ".*clang.*" OR CMAKE_C_COMPILER_ID MATCHES ".*Clang"))

PR #180, opened 2020-02-25
Diasble AOT in SGX build by default, as it requires SGX SDKv2.8 or la…

qdaoming-intel:
…ter.


core/shared/platform/linux-sgx/bh_platform.c
```diff
@@ -7,7 +7,9 @@
 #include "bh_platform.h"
 
 #include <unistd.h>
+#if WASM_ENABLE_AOT
```
wenyongh: We usually use #if WASM_ENABLE_AOT != 0, had better keep the same code style.

PR #183, opened 2020-02-28
Fix bug of Dockerfile

daizoom:
Fixed the path.


simsinght:
I think the Dockerfile has fallen out of step. Building a container throws an error:
```
E: Unable to locate package lib32gcc-5-dev
```

But I think it might be out of date in other ways? Doesn't this project now use llvm-11. Would say how to fix but I can't even figure out how to compile 🙃

PR #194, opened 2020-03-10
let generic invokeNative know that it might have to return something

kfessel:
Compilers may use the return value register if function is not expected to return something (happend to me)
Adding returns into the invokeNative fixes that.

### DIFF
--- a/core/iwasm/common/arch/invokeNative_general.c
+++ b/core/iwasm/common/arch/invokeNative_general.c
@@ -6,81 +6,114 @@
 #include "../wasm_runtime_common.h"
 #include "../wasm_exec_env.h"
 
-void invokeNative(void (*native_code)(), uint32 argv[], uint32 argc)
+int64 invokeNative(void (*_native_code)(), uint32 argv[], uint32 argc)
 {
     bh_assert(argc >= sizeof(WASMExecEnv*)/sizeof(uint32));
 
+    int64 (*native_code)() = (int64 (*)()) _native_code;
     switch(argc) {
         case 0:
-            native_code();
+            return native_code();
             break;
         case 1:
-            native_code(argv[0]);
+            return native_code(argv[0]);
             break;
         case 2:
-            native_code(argv[0], argv[1]);
+            return native_code(argv[0], argv[1]);
             break;
         case 3:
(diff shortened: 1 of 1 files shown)


wenyongh:
Hi, the invokeNative() should be defined as return void here, it calls the native_code() function, which may has different return type, and in the assembly instructions of native_code(), the ret register(e.g. eax in x86-32 to return i32, rax in x86-64 to return i64) keeps the return value. And in runtime, it claims various return type propotype of invokeNative(), and calls them accordingly, in assembly code, the caller reads the ret register set by native_code() to get the return value.

If invokeNative() is defined as return int64, the ret register may be changed, e.g. call native_code() with return type f32/f64 may fail.

BTW: the invokeNative_general.c is only for experiment, it has potential issue for x86-64, arm, xtensa target, so it is not recommended to  use this file. Normally we should use the related assembly version. Not sure what is your target CPU?

kfessel:
My test-target is x86-32. GCC decided to use EAX in invokeNative and by that overwrote the return value of the called function, it did not use anything of the FPU so it left any floating point return value untouched. The issue is it touched EAX. Telling gcc that the function might return something which might have the size of an int64 from the function invokeNative calls tell it to leave the return register untouched even if there is nothing inside.

void means: i will never return anything so if i wirte some thing to eax my caller has to ignore it 

[invokeNative_general_after.s.txt](https://github.com/bytecodealliance/wasm-micro-runtime/files/4324294/invokeNative_general_after.s.txt)
[invokeNative_general_before.s.txt](https://github.com/bytecodealliance/wasm-micro-runtime/files/4324295/invokeNative_general_before.s.txt)

wenyongh:
It is really not recommended to modify the file invokeNative_general.c, could you please clone the source code, create you own product-mini project folder, and in your CMakeLists.txt, excluding file invokeNative_general.c and adding your own version? like:

    list (REMOVE_ITEM ${WAMR_RUNTIME_LIB_SOURCE}  ${WAMR_ROOT_DIR}/core/iwasm/common/arch/invokeNative_general.c)
    list (APPEND ${WAMR_RUNTIME_LIB_SOURCE}  your_file)

I haven't tested the script, hope it works for your.

Issue #238, opened 2020-04-25
Ill-formed wasm file makes wamrc crash


lei-april:
Hi,

First of all, thanks for your awesome work on this project :)

When experimenting with wamr, I mutated my wasm file into a invalid form, and it just crashes wamrc:
```
$ wamrc -o t.aot invalid.wasm
double free or corruption (out)
Aborted (core dumped)
```

This is the invalid wasm file [invalid.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/4532196/invalid.zip), and here's how I created it:
1. Write a text-form wasm module, and use some tool like `wasm2wat` to serialize it.
```
(module
  (func $foo
    (block
    )
  )
)
```
2. Find the opcode for `block` in the binary wasm file, and use some hex editor to change its immediate field to a invalid value. In this case, the opcode for `block` is 0x02 with immediate field 0x40, and I changed it to 0x41.
```
$ xxd  valid.wasm
00000000: 0061 736d 0100 0000 0104 0160 0000 0302  .asm.......`....
00000010: 0100 0a07 0105 0002 400b 0b              ........@..
$ xxd invalid.wasm
00000000: 0061 736d 0100 0000 0104 0160 0000 0302  .asm.......`....
00000010: 0100 0a07 0105 0002 410b 0b              ........A..
```

I investigated this issue a little further, and found that an invalid immediate field of `block`, which indicates the return type of the block structure, would cause function [`wasm_loader_push_frame_ref`](https://github.com/bytecodealliance/wasm-micro-runtime/blob/eab5e4085398a6c187bcd0a20b4fff9e26a4117d/core/iwasm/interpreter/wasm_loader.c#L2670) to access out-of-bound memory address.

A possible fix is to validate the return type of opcode `block`, `loop` and `if`, and stop the wasm loader if validation fails. I worked out a patch, and here's how it works:
```
$ wamrc -o t.aot invalid.wasm
WASM module load failed: invalid block type
```

I'll propose a PR later.

wenyongh:
@lei-april , thanks a lot for finding the issue! The patch has been merged now.

PR #239, opened 2020-04-25
bugfix: check type for opcode block, loop and if (#238)

lei-april:
Otherwise a block opcode with invalid type signature could crash the
wasm loader.


lei-april:
close issue #238

Issue #257, opened 2020-05-21
WAMR on Zephyr


liurenju:
Hi,

I'm currently trying to run wasm using wamr on zephyr (xtensa-qemu). However, when wasm_runtime_load() is called twice, the system crashes. Here's some sample code:

```
if (!(wasm_module = wasm_runtime_load(wasm_file_buf, wasm_file_size,
                                        error_buf, sizeof(error_buf)))){
  //some error checking
}
wasm_runtime_unload(wasm_module);
if (!(wasm_module2 = wasm_runtime_load(wasm_file_buf2, wasm_file_size2,
                                        error_buf2, sizeof(error_buf2)))){
 //some error checking
}
```

And the system shows me:
```[00:00:00.040,000] <err> os:  ** FATAL EXCEPTION
[00:00:00.040,000] <err> os:  ** CPU 0 EXCCAUSE 9 (load/store alignment)
[00:00:00.040,000] <err> os:  **  PC 0x6000deaf VADDR 0x6001d8fd
[00:00:00.040,000] <err> os:  **  PS 0x00060020
[00:00:00.040,000] <err> os:  **    (INTLEVEL:0 EXCM: 0 UM:1 RING:0 WOE:1 OWB:0 CALLINC:2)
[00:00:00.040,000] <err> os:  **  A0 0xa000e139  SP 0x60019360  A2 0x6001de78  A3 0x6001d978
[00:00:00.040,000] <err> os:  **  A4 0x60019400  A5 0x00000080  A6 0x00000000  A7 0x6001d8fd
[00:00:00.040,000] <err> os:  **  A8 0x6d736100  A9 0x60019340 A10 0x6001de78 A11 0x00000000
[00:00:00.040,000] <err> os:  ** A12 0x00000000 A13 0x6001b440 A14 0x6001dcc4 A15 0x00000000
[00:00:00.040,000] <err> os:  ** SAR 0x00000010
[00:00:00.040,000] <err> os: >>> ZEPHYR FATAL ERROR 0: CPU exception on CPU 0
[00:00:00.040,000] <err> os: Current thread: 0x6001da70 (unknown)
[00:00:00.050,000] <err> os: Halting system
```

Can anyone help me with this? thanks.

liurenju:
Found the bug, when declaring the wasm_file_buf array, it needs to align with the memory.

`unsigned char __aligned(4) wasm_test_file[] = {//binary code};`

I will submit a pull request for this.

PR #258, opened 2020-05-21
change the alignment, which might potentially casue problem.

liurenju:
Using __alignment(4) guarantees wasm_test_file[] will be aligned with the memory paging. Otherwise, it might cause the issues in 
https://github.com/bytecodealliance/wasm-micro-runtime/issues/257

Root cause:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/master/core/iwasm/interpreter/wasm_loader.c#L1909

This line will trigger the crash of the system because of the alignment problem.


xwang98:
@liurenju  thanks for the finding and the fix.

Issue #261, opened 2020-05-23
wamr-compiler compilation keeps failing in MacOS


dpinthinker:
After executing `build_llvm.sh`, execute the following commands:

```
mkdir build
cd build
cmake ..
build
```

then it fails.

```
➜  build git:(master) make
Scanning dependencies of target vmlib
[  2%] Building C object CMakeFiles/vmlib.dir/Users/zhujinpeng/workspace/webasm/wasm-micro-runtime/core/shared/platform/linux/platform_init.c.o
clang: error: unknown argument: '-mindirect-branch-register'
clang: warning: -Wl,-z,noexecstack,-z,relro,-z,now: 'linker' input unused [-Wunused-command-line-argument]
make[2]: *** [CMakeFiles/vmlib.dir/Users/zhujinpeng/workspace/webasm/wasm-micro-runtime/core/shared/platform/linux/platform_init.c.o] Error 1
make[1]: *** [CMakeFiles/vmlib.dir/all] Error 2
make: *** [all] Error 2
```

liurenju:
How did you configure the building environment? By the way, you should try to build the runtime on Linux since the one you used was supposed to be used in linux.

dpinthinker:
> How did you configure the building environment? By the way, you should try to build the runtime on Linux since the one you used was supposed to be used in linux.

Yes, I build it successfully on my Linux cloud machine. But I still want to build it on my MacOS laptop. Could you tell me how to build it on MacOS?

wenyongh:
@dpinthinker , could you please comment the -mindirect-branch-register flag in wamr-compiler/CMakeLists.txt, and build again?

    if (WAMR_BUILD_TARGET MATCHES "X86_.*" OR WAMR_BUILD_TARGET STREQUAL "AMD_64")
    set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -mindirect-branch-register")
    endif ()

to

    if (WAMR_BUILD_TARGET MATCHES "X86_.*" OR WAMR_BUILD_TARGET STREQUAL "AMD_64")
    #set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -mindirect-branch-register")
    endif ()

dpinthinker:
After comment the -mindirect-branch-register flag, this error is gone.
But another issue come out.

```
[...]

PR #265, opened 2020-05-28
add darwin compiling support for wamr-compiler

dpinthinker:
#261


wamr-compiler/CMakeLists.txt
```diff
@@ -76,7 +78,9 @@ add_definitions(${LLVM_DEFINITIONS})
 message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION}")
 message(STATUS "Using LLVMConfig.cmake in: ${LLVM_DIR}")
 
-set (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,--gc-sections")
+if (NOT WAMR_BUILD_PLATFORM MATCHES "darwin")
```
wenyongh, line 81: Actually it is not caused by darwin platform, it is caused by clang compiler, so we suggest modifying it as:

if (NOT (CMAKE_C_COMPILER MATCHES ".*clang.*" OR CMAKE_C_COMPILER_ID MATCHES ".*Clang"))
  set (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,--gc-sections")
endif()

Similar to product-mini/platforms/linux/CMakeLists.txt

PR #275, opened 2020-06-05
Missing SGX SDK Include fixed

J-Heinemann:
Hi, 
If I use `set(WAMR_BUILD_PLATFORM linux-sgx)` and `include(${WAMR_ROOT_DIR}/build-scripts/runtime_lib.cmake)` in order to build the wamr-lib for linux-sgx, I get an error message, that my compiler can't find `"sgx_thread.h"`. The following lines fixed that issue. 

Best regards,
Joshua


wenyongh:
Thanks a lot for finding the issue! We have similar process in product-mini/platforms/linux-sgx/CMakeLists.txt, but it seems that it is better to handle it here.

PR #279, opened 2020-06-10
wasm_export.h Function Description & SGX SDK Include path

J-Heinemann:
Hey, 
I added zero as possible return value into the description of function `wasm_runtime_lookup_function` and `wasm_runtime_create_exec_env`. I think its good to know in order to build error handling functions. 

Next, I changed the **shared_platform.cmake** because if you're using just a single CMakeFile to build the trusted and untrusted part and include this file, your untrusted part will also use the tlibc instead of the normal libc by default. 
I think its okay to remove them because if you want to link the wamr library to your enclave, the enclave already contains the nessessary symbols. Another option is that the developer has to split up the CMakeFile into two separate files for untrusted and trusted part. Then this would work. 

Best regards,
Joshua


wenyongh:
@J-Heinemann had better not directly remove the two lines in shared_platform.cmake, could we add a cmake variable to control it? e.g.

    include_directories (${SGX_SDK_DIR}/include)
    if (NOT BUILD_UNTRUST_PART EQUAL 1)
      include_directories (${SGX_SDK_DIR}/include/tlibc
                           ${SGX_SDK_DIR}/include/libcxx)
    endif ()
Then set BUILD_UNTRUST_PART to 1 when building the untrust part.

J-Heinemann:
@wenyongh Sounds good, I think this should work for the building process.

PR #293, opened 2020-06-28
Implement memory access bound check with hardware trap for 64-bit platforms

wenyongh:
Also implement native stack overflow check with hardware trap for 64-bit platforms
Refine classic interpreter and fast interpreter to improve performance
Update document


Kevin0626:
should we update the performance data?

PR #305, opened 2020-07-09
remove duplicated method resolve_main_func 

dpinthinker:


wenyongh:
Thanks a lot @dpinthinker ! I modified it with small changes.

PR #315, opened 2020-07-23
Enable post-MVP feature wasm-c-api

lum1n0us:


Rochet2:
Looks very exciting!

Does `wasm_instance_new` support instantiating multiple modules, one after the other, with the same store?
It seems with this API one can create their own dynamic linking schemes, as long as the modules are carefully designed.

lum1n0us:
YES. It is able to inject *imports* with `wasm_instance_new()`. And No, we'd better not use it when enabling *WAMR_BUILD_MULTI_MODULE*. It might leads to an unexpected situation.

wenyongh:
@Rochet2, yes, wasm_instance_new supports instantiating multiple modules with the same store, and also supports specifying imports of global and function (e.g. exports of other module, or new function with native function ptr) for it. And had better not enable WAMR_BUILD_MULTI_MODULE when using wasm-c-api currently.

The document can be found at: doc/wasm_c_api.md, and sample can be found at: samples/wasm-c-api.

PR #331, opened 2020-08-03
make include headers follow strict protoype rule 

kfessel:
I use an embedded OS that compiles using -Wstrict-prototypes. When i compile files that include I get  headers the following  Warning:

warning: function declaration isn’t a prototype [-Wstrict-prototypes]
  127 | wasm_runtime_init();

I think i would be reasonable to provide the include headers in a form does not produce a warning.

### DIFF
--- a/core/iwasm/common/wasm_runtime_common.h
+++ b/core/iwasm/common/wasm_runtime_common.h
@@ -81,7 +81,7 @@ set_error_buf_v(char *error_buf, uint32 error_buf_size, const char *format,
 
 /* See wasm_export.h for description */
 bool
-wasm_runtime_init();
+wasm_runtime_init(void);
 
 /* See wasm_export.h for description */
 bool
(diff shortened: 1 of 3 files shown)


wenyongh:
@kfessel could you please also modify the function declarations of wasm_runtime_init() and wasm_runtime_destroy() in file wasm_runtime_common.h? Or the function declarations are inconsistent.

kfessel:
what is the reason for it's double declareration? (there is a #include "../include/wasm_export.h" in wasm_runtime_common.h)

kfessel:
i would also like to fix some other issues since i have to switch of a great amount of warnings for the wamr/iwasm compilation

```
#less Wall TODO: get things fixed
CFLAGS :=  $(filter-out -pedantic, $(CFLAGS))
CFLAGS += -Wno-missing-include-dirs
CFLAGS += -Wno-format
CFLAGS += -Wno-strict-prototypes
CFLAGS += -Wno-old-style-definition
CFLAGS += -Wno-cast-function-type
```
is part of the makefile i use to call the cmake and make of wamr

wenyongh:
> I am happy to do so; done
> 
> What is the reason for it's double declareration? (there is a #include "../include/wasm_export.h" in wasm_runtime_common.h)

It is historical reason, the wasm_export.h isn't included by core/iwasm/common folder sometime in the past, so the function prototypes are declared in wasm_runtime_common.h for runtime internal use.

PR #332, opened 2020-08-03
Add wamrc AoT compiler building support for Windows(MSVC)

sophy228:
Just pass build, not test, since I found it is need /opt/wasi-sdk/bin/clang to build a wasm-app first ?

(Can someone share a prebuilt wasm-app to me)

Signed-off-by: Wu Zhongmin <vwzm@live.com>


xwang98:
Supporting Windows for wamrc is what we are really looking for. @sophy228

xujuntwt95329:
@sophy228 

Here is a sample hello world app:
[hello-world.tar.gz](https://github.com/bytecodealliance/wasm-micro-runtime/files/5021362/hello-world.tar.gz)

sophy228:
[helloworld.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/5028100/helloworld.zip)
@xujuntwt95329
Thanks very much for your sharing.
Now it can work with 
--format=llvmir-unopt and --format=llvmir-opt

for obj and AOT,  the program may crash .. may need further debug.

If you have time ,may help to review if the llvmir-opt and llvmir-unopt files in the attachment are expected.

xujuntwt95329:
@sophy228 The `llvm-ir` should be correct. 

I've tried to build `wamrc` based on your patch, but I failed to link the `pthread.lib`, seems the library is built as X86_32?

sophy228:
My patch build all libs as X86 not X86_64.  

I found the pthread.lib problem today,   pthread.lib is prebuilt by VS2019,  using previous VS version may cause link error.

You may need to built one by yourself  https://github.com/GerHobbelt/pthread-win32  

I am trying to fix this by adding pthread-win32 as a dependent project (just like llvm)

xujuntwt95329:
@sophy228  Thanks a lot !

I have successfully built `wamrc` as `x86_32`, now it works well with format `llvmir-opt`, `llvmir-unopt` and `object`. But it will throw an error `invaid llvm binary bin_type` if the output format is set as `aot`, because we haven't parsed the bin_type of windows object file `LLVMBinaryTypeCOFF`.
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/compilation/aot_emit_aot_file.c#L1496

> for obj and AOT, the program may crash .. may need further debug.

The program on my machine does not crash, I'm not sure what's the root cause of your problem. Actually I can't build the LLVM using the `build_llvm.py`, I use a prebuilt LLVM binary instead. You can download the prebuilt one from here [llvm_win32](https://github.com/WAVM/WAVM-LLVM/releases/download/10.x/Win32_Release.zip) [llvm_win64](https://github.com/WAVM/WAVM-LLVM/releases/download/10.x/Win64_Release.zip) and try again to check if it's caused by LLVM.

@wenyongh 
Our `AoT` file is designed to be platform independent, do we need to parse the Coff file? Maybe we can force the `LLVMTargeMachine` to `x86_64-unknown-linux-gnu` so we only need to process the `ELF` file.

wenyongh:
@XuJun2019 could you try adding "--target=i386(or x86_64) --abi=gnu" for wamrc?

sophy228:
@xujuntwt95329  that is a good news,  I think you may be right about LLVM, since I found the crash happened in LLVM codes. I will try your prebuilt LLVM lib.

By the way, what is your problem when using build_llvm.py ?

xujuntwt95329:
@wenyongh 
Cheers! After passing `--target=i386` and `--target-abi=gnu`, the wamrc can successfully generate the AoT file, and generated AoT file can be executed by `iwasm` on my Linux machine.

I've also tried to add `product-mini/platforms/windows`, the iwasm now can be built on windows, and it works well with `classic-interpreter` and `JIT`.

xujuntwt95329:
@sophy228 

> By the way, what is your problem when using build_llvm.py ?

Actually,  I'm not sure about what happened. The `build_llvm.py` can generate the `Visual Studio 2019` project successfully, but when I build the solution in VS, I get different errors for every build. Sometimes it reports `can't find LLVM-C.def`, sometimes `template argument mismatch`, and sometimes it says `cannot open xxx.lib`
I'm confused about this, and I give up building LLVM after 3 attempts because it cost about 2~3 hours for one build.

BTW, I also get another error when building wamrc:
``` C
static SymbolMap target_sym_map[] = {
    REG_COMMON_SYMBOLS
#if !defined(_WIN32)
    /* compiler-rt symbols that come from compiler(e.g. gcc) */
    REG_SYM(__divdi3),
    REG_SYM(__udivdi3),
    REG_SYM(__moddi3),
    REG_SYM(__umoddi3)
#endif
};
```
The MSVC reports `error: initializer element is not constant`, after removing the macro `REG_COMMON_SYMBOLS`, the error disappears.
Have you encountered this problem?

Thanks.

sophy228:
@xujuntwt95329  well,  LLVM prebuilt libs **Debug** version will cause the same crash, and **Release**  version not... It seems LLVM bugs?

And I meet the same error  REG_COMMON_SYMBOLS when build wasmrc as **Release** version, but no such error as **Debug**  version 

Maybe on **Release**, below functions are regard as inline, no symbol will be created ?

    REG_SYM(fminf),                       \
    REG_SYM(fmax),                        \
    REG_SYM(fmaxf),                       \
    REG_SYM(ceil),                        \
   REG_SYM(ceilf),                       \
    REG_SYM(floor),                       \
    REG_SYM(floorf),

xujuntwt95329:
> LLVM prebuilt libs Debug version will cause the same crash, and Release version not... It seems LLVM bugs?

I guess this may be caused by some LLVM assertions, in the release build the assertions are ignored.

> And I meet the same error REG_COMMON_SYMBOLS when build wasmrc as Release version, but no such error as Debug version
Maybe on Release, below functions are regard as inline, no symbol will be created ?

I've tried the **Debug** build without this error, too. Thanks a lot !  I'll try to find a workaround to solve this problem.

xujuntwt95329:
@sophy228 

I've added `iwasm` product build for windows and push the commit to your patch.
You can use this to run the wasm module on windows. `interpreter`, `JIT` and `AoT` mode all works well.

There seems something to be done before we merge this patch:
- [ ] Move `pthread_lib.lib` to somewhere out of this repo to reduce the patch size. I'll check if we can build it from source, or download it during building.
- [ ] Solve the problems with `invokeNative_i32.asm`, the call convention seems different between Windows and Linux, I‘ll take time to debug it.

This is an excellent patch, thanks for your contribution!

sophy228:
@xujuntwt95329 

For invokeNative_i32.asm issue ,  it was built as 16bit mode,   I put a new commit, should be fixed, please have a try.

xujuntwt95329:
Cool! 
It works now.

I've submitted a PR which contains the `iwasm` build for windows to your repo.

xujuntwt95329:
@sophy228 

Hi, 
we are going to merge this patch, however, we find that the `win_pthread` is released under `LGPL`, which is not suitable for our project. 

Actually the thread APIs are not necessary for building `wamrc` and `iwasm`. I've removed the pthread related files and created a PR to your repo, could you please help to merge it into your repo? Then we can merge this PR.

We'll implement the `win_thread.c` later based on windows APIs.

Thanks for your contribution!

sophy228:
@xujuntwt95329  OK， and thanks

sophy228:
@xujuntwt95329 Done

core/shared/utils/uncommon/bh_read_file.c
```diff
@@ -2,10 +2,10 @@
 
 #include <sys/stat.h>
 #include <fcntl.h>
-#if !defined(_WIN32) && !defined(_WIN32_)
-#include <unistd.h>
-#else
+#if defined(_WIN32) || defined(_WIN32_)
 include <io.h>
```
xujuntwt95329, line 6: #include 
^

wamr-compiler/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@
 
 cmake_minimum_required (VERSION 2.8)
 
-if (NOT MSVC)
+if (NOT WAMR_BUILD_PLATFORM STREQUAL "windows")
```
xujuntwt95329, line 6:
```
if (NOT WAMR_BUILD_PLATFORM STREQUAL "windows")
  ^
  seems this is not a space
```

PR #342, opened 2020-08-07
optimize: get_current_target for AOT thumb loader

kfessel:
debugging i found myself in the AOT loader part and saw there was a massive amount of string and format printing used for that simple function so i went to solve it.

(it had 2 format string and 1 string length calls to basically copy a constant)
i checked with godbolt but no compiler (and option) did get rid of these calls all of them made it into the assembler
sadly i cold net get the case conversion removed from the result.

i would do the same for aarch64 and arm if u like

### DIFF
--- a/core/iwasm/aot/arch/aot_reloc_thumb.c
+++ b/core/iwasm/aot/arch/aot_reloc_thumb.c
@@ -118,32 +118,32 @@ get_target_symbol_map(uint32 *sym_num)
 }
 
 
-#define BUILD_TARGET_V4 "thumbv4t"
+#define BUILD_TARGET_THUMB_V4T "thumbv4t"
 void
 get_current_target(char *target_buf, uint32 target_buf_size)
 {
     const char * s =  BUILD_TARGET;
     size_t s_size = sizeof(BUILD_TARGET);
+    char *d = target_buf;
 
-    //check if wenn need to add some subarchitecture specification
-    //this is some kind of thumb since we are in ...thumb.c
-    if(sizeof(BUILD_TARGET) == sizeof("thumb")){
-        s = BUILD_TARGET_V4;
-        s_size = sizeof(BUILD_TARGET_V4);
+    /* Set to "thumbv4t" by default if sub version isn't specified */
+    if (strcmp(s, "THUMB") == 0) {
+        s = BUILD_TARGET_THUMB_V4T;
+        s_size = sizeof(BUILD_TARGET_THUMB_V4T);
     }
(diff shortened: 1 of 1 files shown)


wenyongh:
Hi, thanks for uploading the patch to optimize the AOT loader, but it has some coding style issues, I tried to change it to:
```c
#define BUILD_TARGET_THUMB_V4T "thumbv4t"

void
get_current_target(char *target_buf, uint32 target_buf_size)
{
    const char *s = BUILD_TARGET;
    size_t s_size = sizeof(BUILD_TARGET);
    char *d;

    /* Set to "thumbv4t" by default if sub version isn't specified */
    if (strcmp(s, "thumb") == 0) {
        s = BUILD_TARGET_THUMB_V4T;
        s_size = sizeof(BUILD_TARGET_THUMB_V4T);
    }

    if (target_buf_size < s_size) {
        s_size = target_buf_size;
    }

    d = target_buf;
    while (--s_size) {
        if (*s >= 'A' && *s <= 'Z')
            *d++ = *s++ + 'a' - 'A';
        else
            *d++ = *s++;
    }
   /* Ensure the string is null byte ('\0') terminated */
    *d = '\0';
}
```
Could you please help test it? thanks.

kfessel:
Hello wenyongh,

thanks for reviewing 

This line:
`
if(strcmp(s, "thumb") == 0)
`
needs to have THUMB in upper case since BUILD_TARGET is in upper case (and strcmp is case sensitive).

The comparison of just the size does also works since we are in a thumb specific file (...thumb.c) BUILD_TARGET will either be THUMB or THUMB[sub] with some sub e.g. THUMBV4T or THUMBV6M other possible possible values which we do not want to replace will have an other number of letters and will just be converted to lower case.
`
if(sizeof(BUILD_TARGET) == sizeof("thumb"))
`
i tried to explain that with the comment
`
//check if wenn need to add some subarchitecture specification
//this is some kind of thumb since we are in ...thumb.c
`
which (my comment) seems to be suboptimal

Since all off the information that is used to build this function is static the commpieler is able to optimize it 
`
strcmp(s, "THUMB") == 0
`
To compare different variants there is a https://godbolt.org/z/rvcj3o

I updated my PR

wenyongh:
Yes, using `if(sizeof(BUILD_TARGET) == sizeof("thumb"))` seems to be good, but it makes the code not easy to understand, another issue is that developer might not use cmake, but use Makefile written by himself, which may pass other BUILD_TARGET string, so we had better use strcmp explicitly.

PR #379, opened 2020-09-09
Add Windows support for C-API and Runtime API libraries and examples.

qinxk-inter:


core/iwasm/include/wasm_export.h
```diff
@@ -713,7 +726,8 @@ wasm_runtime_get_native_addr_range(wasm_module_inst_t module_inst,
   *
   * @return true if success, false otherwise
   */
-bool wasm_runtime_register_natives(const char *module_name,
+WASM_RUNTIME_API_EXTERN bool
+wasm_runtime_register_natives(const char *module_name,
                                    NativeSymbol *native_symbols,
```
wenyongh, line 731: Indent: after modified, the place of first argument in last line is changed, could you please also change the indent of this line?

core/iwasm/include/wasm_export.h
```diff
@@ -727,7 +741,8 @@ bool wasm_runtime_register_natives(const char *module_name,
  * and write the return value back to args[0] with macro
  *   native_raw_return_type and native_raw_set_return
  */
-bool wasm_runtime_register_natives_raw(const char *module_name,
+WASM_RUNTIME_API_EXTERN bool
+wasm_runtime_register_natives_raw(const char *module_name,
                                        NativeSymbol *native_symbols,
```
wenyongh, line 746: Same as above.

core/shared/utils/uncommon/bh_getopt.c
```diff
@@ -0,0 +1,51 @@
+/* *****************************************************************
+*
+* Copyright 2016 Microsoft
+*
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
```
wenyongh, line 6: This file's copyright is Microsoft, and the license is Apache 2.0, but not Apache-2.0 WITH LLVM-exception, it seems that it is not very difficult to implement getopt function, could you please rewrite the API?

core/shared/utils/uncommon/bh_getopt.c
```diff
@@ -0,0 +1,51 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*
+******************************************************************/
+#ifdef WIN32
+#include "bh_getopt.h"
+#include <windows.h>
+char* optarg = NULL;
+int optind = 1;
```
wenyongh, line 23: The two variables arg global, could we move them into getopt as local variables?

qinxk-inter: No, we couldn't. At least optind cannot be changed to local variable, because some place like samples/basic/src/main.c use the optind out of the getopt function.
And also from the getopt(3) man page:

> SYNOPSIS
>
>      #include <unistd.h>
> 
>      extern char *optarg;
>      extern int optind;
>      extern int optopt;
>      extern int opterr;
>      extern int optreset;
> 
>      int
>      getopt(int argc, char * const argv[], const char *optstring);
I am afraid we need to keep them as global.

core/shared/utils/uncommon/bh_getopt.h
```diff
@@ -0,0 +1,37 @@
...
+extern char *optarg;
+extern int optind;
+
+int getopt(int argc, char *const argv[], const char *optstring);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif
```
xujuntwt95329, line 36: There are many `endif`, had better add a comment:
```suggestion
#endif /* end of GETOPT_H__ */
```

samples/wasm-c-api/CMakeLists.txt
```diff
@@ -39,24 +45,28 @@ if(NOT DEFINED WAMR_BUILD_FAST_INTERP)
...
+  endif ()
+endif()
 # build out vmlib
 set(WAMR_ROOT_DIR ${CMAKE_CURRENT_LIST_DIR}/../..)
 set(WAMRC ${WAMR_ROOT_DIR}/wamr-compiler/build/wamrc)
 include (${WAMR_ROOT_DIR}/build-scripts/runtime_lib.cmake)
 
 add_library(vmlib STATIC ${WAMR_RUNTIME_LIB_SOURCE})
+if (MSVC)
+target_compile_definitions(vmlib PRIVATE WASM_API_EXTERN=)
```
xujuntwt95329, line 68: Had better indent 2 spaces
```suggestion
  target_compile_definitions(vmlib PRIVATE WASM_API_EXTERN=)
```

samples/wasm-c-api/CMakeLists.txt
```diff
@@ -66,6 +76,9 @@ target_include_directories(c-api
   PRIVATE ${C_API_PATH}/include
 )
 target_link_libraries(c-api PRIVATE vmlib -lpthread -lm)
+if (MSVC)
+target_compile_definitions(c-api PRIVATE WASM_API_EXTERN=)
```
xujuntwt95329, line 80: Same as above

core/shared/utils/uncommon/bh_getopt.c
qinxk-inter: I just found this piece of code had an issue that optind will not increment if no ':' followed. We will reimplement it and fix the issue.

core/shared/utils/uncommon/bh_getopt.c
```diff
@@ -0,0 +1,55 @@
+#ifndef __GNUC__
```
wenyongh, line 1: Missing license header

qinxk-inter: Added, btw, the license header may need to be added for the bh_read_file.c also under the same directory.

PR #391, opened 2020-09-19
Introduce CI support

no1wudi:
Basic ci flow, only compile iwasm and samples/wasm-c-api for linux now.

I mainly work on the NuttX platform, so an ci flow is needed to test whether my modification has affected other platforms.

Maybe we need a more complete ci flow to cover more features, right?

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -1,0 +1,36 @@
+name: Linux
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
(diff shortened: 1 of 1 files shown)


.github/workflows/linux.yml
```diff
@@ -0,0 +1,36 @@
+name: Linux
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
```
wenyongh, line 4: Should be the main branch.

PR #397, opened 2020-09-23
Fix wasm loader malloc(0) issue which returns NULL is some platforms

no1wudi:
Detail can be found in comments.

Signed-off-by: Huang Qi <huangqi3@xiaomi.com>

### DIFF
--- a/core/iwasm/interpreter/wasm_loader.c
+++ b/core/iwasm/interpreter/wasm_loader.c
@@ -1691,8 +1691,14 @@ init_function_local_offsets(WASMFunction *func,
     uint32 i, local_offset = 0;
     uint64 total_size = sizeof(uint16) * ((uint64)param_count + local_count);
 
-    if (!(func->local_offsets =
-                loader_malloc(total_size, error_buf, error_buf_size))) {
+    /*
+     * Only allocate memory when total_size is not 0,
+     * or the return value of malloc(0) might be NULL on some platforms,
+     * which causes wasm loader return false.
+     */
+    if (total_size > 0
+        && !(func->local_offsets =
+               loader_malloc(total_size, error_buf, error_buf_size))) {
         return false;
     }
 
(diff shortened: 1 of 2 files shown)


core/iwasm/interpreter/wasm_loader.c
```diff
@@ -1692,8 +1692,16 @@ init_function_local_offsets(WASMFunction *func,
...
+            loader_malloc(total_size, error_buf, error_buf_size))) {
+        /*
+         * When total_size is 0, the return value of malloc(0)
+         * is NULL on some platform.
+         * Note: If size is 0, then malloc() returns either NULL,
+         * or a unique pointer value that can later be successfully passed to free().
+         */
+        if (total_size != 0) {
+            return false;
+        }
```
wenyongh, line 1704: How about changing to:
```C
        if (total_size > 0
            && !(func->local_offsets =
                 loader_malloc(total_size, error_buf, error_buf_size))) {
            return false;
        }
```
If total size is 0, just leave it NULL.

no1wudi: Done

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -1691,8 +1691,15 @@ init_function_local_offsets(WASMFunction *func,
...
     uint64 total_size = sizeof(uint16) * ((uint64)param_count + local_count);
 
-    if (!(func->local_offsets =
-                loader_malloc(total_size, error_buf, error_buf_size))) {
+    /*
+     * When total_size is 0, the return value of malloc(0)
+     * is NULL on some platform.
+     * Note: If size is 0, then malloc() returns either NULL,
+     * or a unique pointer value that can later be successfully passed to free().
+     */
```
wenyongh, line 1699: The comments make a little confused, could we change to be like:
Only allocate memory when total_size is not 0, or the return value of malloc(0) might be NULL on some platforms, which causes wasm loader return false.

PR #403, opened 2020-09-25
Add NULL check for memory inst in aot/wasm module malloc/free

no1wudi:


core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -1575,6 +1575,15 @@ wasm_module_malloc(WASMModuleInstance *module_inst, uint32 size,
     uint8 *addr = NULL;
     uint32 offset = 0;
 
+    /*
+     * This case is not common but exists, some compiler may generate
+     * code that use libc's malloc without initial memory or memory.grow
```
wenyongh, line 1580: should be that uses?

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -1575,6 +1575,15 @@ wasm_module_malloc(WASMModuleInstance *module_inst, uint32 size,
     uint8 *addr = NULL;
     uint32 offset = 0;
 
+    /*
+     * This case is not common but exists, some compiler may generate
+     * code that use libc's malloc without initial memory or memory.grow 
+     */
+    if (!memory) {
```
wenyongh, line 1582: Could you please also add check in wasm_module_free, and aot_module_malloc/free?

PR #425, opened 2020-10-15
add RIOT-OS support to WAMR

kfessel:
Add RIOT-os platform and simple example to WAMR


xwang98:
@kfessel It is great work! Thanks for contributing to the repo!

PR #466, opened 2020-12-09
product-mini/nuttx: Add support for repl mode

no1wudi:


product-mini/platforms/nuttx/main.c
```diff
@@ -296,6 +378,8 @@ main(int argc, char *argv[])
         goto fail3;
     }
 
+    if (is_repl_mode)
+        app_instance_repl(wasm_module_inst);
     if (func_name)
```
wenyongh, line 383: Should modify to else if (func_name) like Linux?

no1wudi: Fixed

PR #469, opened 2020-12-11
product-mini/nuttx: Use readline for repl mode

no1wudi:
The EOL of NuttX is configurable (CR/LF/CRLF), the implementation
of getline in NuttX need CR or LF as line delimiter(CRLF not supported).
Then we should use readline for better compatibility.

### DIFF
--- a/product-mini/platforms/nuttx/main.c
+++ b/product-mini/platforms/nuttx/main.c
@@ -10,6 +10,8 @@
 #include <stdio.h>
 #include <string.h>
 
+#include <system/readline.h>
+
 #include "bh_platform.h"
 #include "bh_read_file.h"
 #include "wasm_export.h"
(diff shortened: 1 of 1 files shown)


product-mini/platforms/nuttx/main.c
```diff
@@ -119,13 +121,15 @@ split_string(char *str, int *count)
 static void *
 app_instance_repl(wasm_module_inst_t module_inst)
 {
-    char *cmd = NULL;
-    size_t len = 0;
+    size_t len = 128;
+    char *cmd = malloc(len);
     ssize_t n;
 
-    while ((printf("webassembly> "), n = getline(&cmd, &len, stdin)) != -1) {
```
wenyongh, line 126: Should check whether cmd is NULL before while?

no1wudi: Thanks, I forget to force push null check to server
Done

no1wudi: Thanks, I forget to force push null check to server

Done

PR #492, opened 2021-01-05
Refine aot call_indirect opcode and interpreter br_table opcode

wenyongh:
Re-implement aot call_indirect opcode translation: when calling non-import function, translate it by LLVM call IR to call the function in AOTed code, so as to avoid calling runtime aot_call_indirect API which is much slower. For import function, keep calling aot_call_indirect API due to the possible pointer/string argument conversion.

And add prompt info while app heap is corrupted, change emit_leb to emit_uint32 inter fast-interp to refine footprint.


core/iwasm/aot/aot_runtime.c
```diff
@@ -1381,7 +1381,17 @@ aot_module_malloc(AOTModuleInstance *module_inst, uint32 size,
     }
 
     if (!addr) {
-        aot_set_exception(module_inst, "out of memory");
+        if (memory_inst->heap_handle.ptr
+            && mem_allocator_is_heap_corrupted(memory_inst->heap_handle.ptr)) {
+            LOG_ERROR("Error: app heap is corrupted, if the wasm file "
+                      "is compiled by wasi-sdk-12.0 or larger version, "
+                      "please add -Wl,--export=malloc -Wl,--export=free "
+                      " to export malloc and free functions.");
```
yamt, line 1389: why do you recommend to export malloc/free?

do you just mean it can be a workaround to avoid app heap corruption? (because these export disables app heap)

or, is there any incompatibility between wasi-sdk-12 and app heap, which can cause app heap corruptions?

PR #494, opened 2021-01-11
add porting codes of rt-thread.

alvkeke:
add porting codes and scons building scripts that are used by rt-thread, document was modified passingly.

### DIFF
--- a/SConscript
+++ b/SConscript
@@ -1,0 +1,27 @@
+# for module compiling
+import os
+
+from building import *
+
+objs = []
+cwd  = GetCurrentDir()
+list = os.listdir(cwd)
+
+if GetDepend(['PKG_USING_WAMR']):
+    wamr_entry_sconscript = os.path.join(cwd, "product-mini", "platforms", "rt-thread", 'SConscript')
+
+    if os.path.isfile(wamr_entry_sconscript):
+        objs = objs + SConscript(wamr_entry_sconscript)
+    else:
+        print("[WAMR] entry script wrong:", wamr_entry_sconscript)
+        Return('objs')
+
+    wamr_runlib_sconsript = os.path.join(cwd, "build-scripts", 'SConscript')
+
+    if os.path.isfile(wamr_runlib_sconsript):
+        objs = objs + SConscript(wamr_runlib_sconsript)
+    else:
+        print("[WAMR] runtime lib script wrong:", wamr_runlib_sconsript)
+
+Return('objs')
+
--- a/SConscript
+++ b/SConscript
@@ -1,3 +1,9 @@
+#
+# Copyright (c) 2021, RT-Thread Development Team
+#
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+
 # for module compiling
 import os
 
--- a/core/shared/platform/rt-thread/platform_internal.h
+++ b/core/shared/platform/rt-thread/platform_internal.h
@@ -1,0 +1,34 @@
+//
+// Created by alvis on 2020/12/14.
+//
+
+#ifndef RTTHREAD_PLATFORM_INTERNAL_H
+#define RTTHREAD_PLATFORM_INTERNAL_H
+
+#include <rtthread.h>
+#include <stdbool.h>
+#include <string.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <math.h>
+#include <stdint.h>
+#include <ctype.h>
+
+#ifndef BUILD_TARGET
+#define BUILD_TARGET "Unknown"
+#endif
+
+typedef rt_thread_t korp_tid;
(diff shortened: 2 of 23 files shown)


core/shared/platform/rt-thread/platform_internal.h
```diff
@@ -0,0 +1,34 @@
...
+#include <rtthread.h>
+#include <stdbool.h>
+#include <string.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <math.h>
+#include <stdint.h>
+#include <ctype.h>
+
+#ifndef BUILD_TARGET
```
wenyongh, line 17: Not sure whether we should define BUILD_TARGET here, normally we define BUILD_TARGET_XXX in build-scripts/config_common.cmake by parsing the WAMR_BUILD_TARGET=XXX passed by cmake variable (cmake -DWAMR_BUILD_TARGET=XXX).

And had better define the BH_PLATFORM_XXX macro in this file, so other files can use the macro, e.g. [wasm_native.c](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/common/wasm_native.c#L10):
#ifndef BUILD_PLATFORM_RT_THREAD
#define BUILD_PLATFORM_RT_THREAD
#endif

core/shared/platform/rt-thread/rtt_platform.c
```diff
@@ -0,0 +1,173 @@
...
+        rt_free(ptr->real);
+        rt_free(ptr);
+    }
+}
+
+
+void *os_malloc(unsigned size)
+{
+    void* buf = rt_malloc(size + 8);
+    if (((rt_ubase_t)buf)&7)
```
wenyongh, line 37: Not sure why here it doesn't check buf != NULL, but check (buf & 7), could buf's lowest 3 bits be zero, and higher bits be not zero? e.g. 0x11110, or 0x11118.

wenyongh (commented):
Thanks for submitting the patch!

core/shared/platform/rt-thread/rtt_platform.c
alvkeke: wamr need heap init struct be aligned to 8-bytes, otherwise iwasm will print error information: `[GC_ERROR]heap init struct buf not 8-byte aligned`. these codes are used to make sure the memories allocated are all align to 8-byte.

wenyongh: Got it, thanks a lot! And should we also check the input pointer ptr like os_free in os_realloc? The ptr might be not the original buffer returned by rt_malloc.

alvkeke: there is a list to save the map from fixed-address to origin-address named `mem_list`. if the address of memory allocated is not align to 8-byte, this address and fixed-address will be save into this list. while `os_free` be called, will also check this list to find the original address of the pointer pass in.

as for `os_realloc`, I have not done anything yet, I will resolve it later.

wenyongh: Hi, it seems that the performance might be not so good if saving the map to mem_list, e.g. in os_free, we need to traverse the list. And also for multi-thread, normally we should to add lock for the list. How about allocating memory with extra 16 bytes, saving the original address to the next 8-byte aligned address, and return the next next 8-byte aligned address? e.g.:
void *os_malloc(unsigned size)
{
    uint8 *buf = rt_malloc(size + 16);
    uint8 *buf_aligned;

    if (buf) {
        buf_aligned = (uint8 *)(((uintptr_t)buf + 7) & ~(uintptr_t)7);
        *(uint8 **)buf_aligned = buf;
        buf = buf_aligned + 8;
    }
    return buf;
}

void os_free(void *ptr)
{
    uint8 *buf_aligned;

    if (ptr) {
        buf_aligned = (uint8 *)ptr - 8;
        buf = *(uint8 **)buf_aligned;
        rt_free(buf);
    }
}

alvkeke: thinks for your reply and advice, It seems be a good idea, I will try this today.

SConscript
```diff
@@ -0,0 +1,27 @@
+# for module compiling
```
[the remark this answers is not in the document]
alvkeke: I have add license header to files I added, please check it.

core/shared/platform/rt-thread/platform_internal.h
alvkeke: sorry, this definiation was used in the past version and I forgot to delete it, I just remove it from `rt-thread/platform_internal.h`.

core/shared/platform/rt-thread/rtt_platform.c
alvkeke: I have rewrite these three method, please check.

PR #506, opened 2021-01-26
Fix memory alloc failed report when malloc(0) returns NULL.

no1wudi:
Since malloc(0) is an undefined behavior, NULL or a valid piece of memory may be returned.

When NULL returned, we should check the memory size requested, instead of just raise an error.


wenyongh (commented):
Hi, normally we should avoid allocating memory if the size is 0 and there are more places found, we will try to check and fix them.

Issue #322, opened 2020-07-28
Build error: Building samples/basic for AARCH64 target


xicocana:
I am trying to build the basic sample for aarch64 target.

On the CMakeLists.txt I have changed the WAMR_BUILD_TARGET from:
```
set (WAMR_BUILD_TARGET "X86_64")
```
to:
```
set (WAMR_BUILD_TARGET "AARCH64")
```

but i get the following build error (log.txt file in attachment) :
```
CMakeFiles/vmlib.dir/build.make:420: recipe for target 'CMakeFiles/vmlib.dir/home/user/wasm_runtimes/wasm-micro-runtime/core/iwasm/common/arch/invokeNative_aarch64.s.o' failed
make[2]: *** [CMakeFiles/vmlib.dir/home/user/wasm_runtimes/wasm-micro-runtime/core/iwasm/common/arch/invokeNative_aarch64.s.o] Error 1
CMakeFiles/Makefile2:97: recipe for target 'CMakeFiles/vmlib.dir/all' failed
make[1]: *** [CMakeFiles/vmlib.dir/all] Error 2
Makefile:103: recipe for target 'all' failed
make: *** [all] Error 2
BUILD_FAIL basic exit as 0\n
```

[log.txt](https://github.com/bytecodealliance/wasm-micro-runtime/files/4991650/log.txt)

Do I need to change something else?
Thanks

xujuntwt95329:
Seems your toolchain doesn't support `aarch64` instructions, have you changed the `CMAKE_C_COMPILER` to your cross-compiler ?
You can use `make VERBOSE=1` to check which compiler you are using.

xicocana:
thanks @xujuntwt95329 ! I indeed forgot about the toolchain. I'll give it a try and report back

xicocana:
I changed the CMAKE_C_COMPILER to my cross compiler toolchain ```aarch64-none-elf-gcc```.
But now I'm having the following problem:
```
aarch64-none-elf-gcc: error: unrecognized command line option '-mindirect-branch-register'

```

Can I remove this command line option so that my toolchain works? Or is it essential? If that's the case could you recommend a different toolchain?

xicocana:
@xujuntwt95329 I have changed the cross compiler toolchain to ```aarch64-none-linux-gnu-gcc``` and all is working fine now. thanks

PR #540, opened 2021-02-23
Fix AoT on Win32 & Win64

### LINKED ISSUE #531: AoT loading failure(s)
Hey there.

I've had great luck with the interpreter, but I'm wanting to give the AoT compilation a try. I'm on Windows, and trying to target x86 specifically (not x64). So, my .wasm runs great in the interpreter, and I compile it via `wamrc --target=i386`. When loading, however, `wasm_runtime_load` throws this error:

`AOT module load failed: resolve symbol ceilf failed`

In other tests, it also failed on a `__udivdi3` function, but this is the first time I've seen it reference a C library function. I'm using WASI, but without the linked library (just using the static routines, basically), so these should be linked into the .wasm (they run fine in the interpreter). Where should I be looking to resolve this?

Paril:
This fixes #531 and implements the missing functions that were patched out in PR #322. Allows AoT to work on Windows.

### DIFF
--- a/core/iwasm/aot/aot_reloc.h
+++ b/core/iwasm/aot/aot_reloc.h
@@ -37,19 +37,6 @@ typedef struct {
 #define REG_AOT_TRACE_SYM()
 #endif
 
-#if (defined(_WIN32) || defined(_WIN32_)) && defined(NDEBUG)
-#define REG_COMMON_SYMBOLS                \
-    REG_SYM(aot_set_exception_with_id),   \
-    REG_SYM(aot_invoke_native),           \
-    REG_SYM(aot_call_indirect),           \
-    REG_SYM(wasm_runtime_enlarge_memory), \
-    REG_SYM(wasm_runtime_set_exception),  \
-    {"memset", (void*)aot_memset},        \
-    {"memmove", (void*)aot_memmove},      \
-    REG_BULK_MEMORY_SYM()                 \
-    REG_ATOMIC_WAIT_SYM()                 \
-    REG_AOT_TRACE_SYM()
-#else /* else of (defined(_WIN32) || defined(_WIN32_)) && defined(NDEBUG) */
 #define REG_COMMON_SYMBOLS                \
     REG_SYM(aot_set_exception_with_id),   \
     REG_SYM(aot_invoke_native),           \
@@ -73,7 +60,6 @@ typedef struct {
     REG_BULK_MEMORY_SYM()                 \
     REG_ATOMIC_WAIT_SYM()                 \
     REG_AOT_TRACE_SYM()
-#endif /* end of (defined(_WIN32) || defined(_WIN32_)) && defined(NDEBUG) */
 
 #define CHECK_RELOC_OFFSET(data_size) do {                                  \
     if (!check_reloc_offset(target_section_size, reloc_offset, data_size,   \
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -2264,7 +2264,7 @@ aot_get_module_inst_mem_consumption(const AOTModuleInstance *module_inst,
         mem_conspn->memories_size +=
             mem_inst->num_bytes_per_page * mem_inst->cur_page_count;
         mem_conspn->app_heap_size =
-            mem_inst->heap_data_end.ptr - mem_inst->heap_data.ptr;
+            (uint8_t *) mem_inst->heap_data_end.ptr - (uint8_t *) mem_inst->heap_data.ptr;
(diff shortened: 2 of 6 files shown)


core/iwasm/aot/aot_reloc.h
```diff
@@ -37,19 +37,6 @@ typedef struct {
...
-    REG_SYM(aot_invoke_native),           \
-    REG_SYM(aot_call_indirect),           \
-    REG_SYM(wasm_runtime_enlarge_memory), \
-    REG_SYM(wasm_runtime_set_exception),  \
-    {"memset", (void*)aot_memset},        \
-    {"memmove", (void*)aot_memmove},      \
-    REG_BULK_MEMORY_SYM()                 \
-    REG_ATOMIC_WAIT_SYM()                 \
-    REG_AOT_TRACE_SYM()
-#else /* else of (defined(_WIN32) || defined(_WIN32_)) && defined(NDEBUG) */
```
xujuntwt95329, line 52: some symbols are not exists in Win64 Release build, maybe we need to reserve these code but change `WIN32` to `WIN64`

``` C
#if (defined(_WIN64) || defined(_WIN64_)) && defined(NDEBUG)
#define REG_COMMON_SYMBOLS                \
REG_SYM(aot_set_exception_with_id),       \
    REG_SYM(aot_invoke_native),           \
    REG_SYM(aot_call_indirect),           \
    REG_SYM(wasm_runtime_enlarge_memory), \
    REG_SYM(wasm_runtime_set_exception),  \
    {"memset", (void*)aot_memset},        \
    {"memmove", (void*)aot_memmove},      \
    REG_SYM(fmin),                        \
    REG_SYM(fminf),                       \
    REG_SYM(fmax),                        \
    REG_SYM(fmaxf),                       \
    REG_SYM(trunc),                       \
    REG_SYM(truncf),                      \
    REG_SYM(rint),                        \
    REG_SYM(rintf),                       \
    REG_BULK_MEMORY_SYM()                 \
    REG_ATOMIC_WAIT_SYM()                 \
    REG_AOT_TRACE_SYM()
#else /* else of (defined(_WIN64) || defined(_WIN64_)) && defined(NDEBUG) */
...
```

Paril: Oh interesting. I'll try the win64 build here. I'd rather get Win64 building properly than have them removed.

Paril: I pushed an update with x64 AoT relocs in place. I totally forgot about that, my bad!

xujuntwt95329 (commented):
Looks good to me.

core/iwasm/aot/aot_runtime.c
```diff
@@ -2264,7 +2264,7 @@ aot_get_module_inst_mem_consumption(const AOTModuleInstance *module_inst,
         mem_conspn->memories_size +=
             mem_inst->num_bytes_per_page * mem_inst->cur_page_count;
         mem_conspn->app_heap_size =
-            mem_inst->heap_data_end.ptr - mem_inst->heap_data.ptr;
+            (uint8_t *) mem_inst->heap_data_end.ptr - (uint8_t *) mem_inst->heap_data.ptr;
```
wenyongh, line 2267: Hi, @Paril, we defined uint8, int64, uint64 types in the platform header file, and usually use them inside runtime instead of uint8_t, int64_t, uint64_t and so on. Could you please modify the related changes so that the coding style is unified?

core/iwasm/aot/arch/aot_reloc_x86_32.c
```diff
@@ -8,20 +8,43 @@
...
+#if !defined(_WIN32) && !defined(_WIN32_)
 void __divdi3();
 void __udivdi3();
 void __moddi3();
 void __umoddi3();
+#else
+#pragma function (floor)
+#pragma function (ceil)
+
+int64_t __divdi3(int64_t a, int64_t b)
```
wenyongh, line 20: Same as above

core/iwasm/aot/arch/aot_reloc_x86_64.c
```diff
@@ -14,10 +14,37 @@
...
 void __udivdi3();
 void __moddi3();
 void __umoddi3();
+#else
+#pragma function (floor)
+#pragma function (ceil)
+#pragma function (floorf)
+#pragma function (ceilf)
+
+int64_t __divdi3(int64_t a, int64_t b)
```
wenyongh, line 28: Same as above

core/iwasm/aot/aot_runtime.c
Paril: Oops, will do.

PR #562, opened 2021-03-10
Build fixes for NuttX sim/macOS

yamt:


core/shared/platform/common/posix/posix_thread.c
```diff
@@ -166,17 +166,27 @@ int os_cond_wait(korp_cond *cond, korp_mutex *mutex)
...
+    long int tv_nsec_new;
+
+    /*
+     * Assumptions:
+     *  - time_t has no padding
+     *  - time_t is a signed integer
+     */
+
+    const time_t TIME_T_MAX =
+        (((time_t)1 << ((sizeof(time_t) * CHAR_BIT) - 2)) - 1) * 2 + 1;
```
xujuntwt95329, line 179: This code is hard to understand, is there any links or documentations to illustrate such kind of usage? I'm not sure if this is suitable for every platform.

If this is only to fix build issues on some specified platforms, I'd suggest you using platform defined macros.

yamt: @xujuntwt95329 
can you tell me which headers are appropriate for such macros? i'm not familiar with this codebase.
core/shared/platform/nuttx/platform_internal.h etc?

xujuntwt95329: The macro `BH_PLATFORM_XXX` should be available here.

But still, I'm worried about this statement `(((time_t)1 << ((sizeof(time_t) * CHAR_BIT) - 2)) - 1) * 2 + 1`, nobody will know what this exactly mean after a couple of days. Is this the only approach to fix your build?

yamt: it's just the easiest way it hopped up in my mind.

alternatively, i guess we can have for nuttx
```
#define BH_TIME_T_MAX 0xffffffff
```
and for others
```
#define BH_TIME_T_MAX LONG_MAX
```
how do you think?

xujuntwt95329: Looks good to me

yamt: done

yamt:
i marked it a draft for now as it seems to need some more fixes

yamt:
the last pushes are a simple conflict resolution + a few more commits for the recent changes.
i tested build + fib32.wasm with the latest nuttx sim/macos as of today.

yamt:
@xujuntwt95329 do you have any more concerns?

Issue #542, opened 2021-02-24
Debug memory leak about wasm app


JavanZhu:
I noticed the memory consumption of my wasm app is keep growing when running, and in the runtime, memory allocation call stack (AOT MODE) is following:
```
#01 wasm_runtime_realloc_internal at wasm_memory.c:153
#02 wasm_runtime_realloc at wasm_memory.c:182
#03 aot_enlarge_memory at aot_runtime.c:1506
#04 wasm_runtime_enlarge_memory at wasm_runtime_common.c:1614
```

memory allocation call stack (Intercepter MODE):
```
#00 wasm_runtime_realloc at wasm_memory.c:196
#01 wasm_enlarge_memory at wasm_runtime.c:1977
#02 wasm_interp_call_func_bytecode at wasm_interp_classic.c:1864
#03 wasm_interp_call_wasm at wasm_interp_classic.c:3362
#04 wasm_call_function at wasm_runtime.c:1621
#05 wasm_runtime_call_wasm at wasm_runtime_common.c:1042
```
It's look likes that memory leak is happened in wasm module heap, I would like to know if there are any tips on debugging memory leak like this, such as how to dump memory heap of wasm app.. thank you.

* Language: AssemblyScript

Paril:
There's a function that will dump the call stack out of the execution environment (if compiled with the DUMP_STACK feature enabled) - that's usually what I've done to find where allocations are happening.

wenyongh:
Hi, the runtime can detect the memory leaks by enabling the heap verification feature, please modify core/config.h, set `BH_ENABLE_GC_VERIFY` to 1. Note that there may be two heaps if the runtime allocator is set to allocate memory from a global heap buffer: the global heap and the app heap. Both of their memory leaks can be detected. And as app heap's space is inside the app's linear memory, it is not an issue in fact if there is a memory leak in app heap, since its space will be freed when the linear memory is destroyed.

JavanZhu:
Hi, @wenyongh 
I have set BH_ENABLE_GC_VERIFY to 1, and nothing output, sorry I'm not familar with it, but I will try it more.

I got following assemblyscript call stack of my module depends on @Paril 's reply (Thank you for @Paril's tip), 
```
#00 ~lib/rt/tlsf/growMemory 
#01 ~lib/rt/tlsf/allocateBlock 
#02 ~lib/rt/tlsf/__alloc 
#03 ~lib/rt/pure/__new 
[...]

PR #577, opened 2021-03-16
Adapt AssemblyScript's new/pin/unpin APIs(0.18).

JavanZhu:
Hi, In AssemblyScript 0.17, there exist a memory leak problem which I had opened a [issue](https://github.com/bytecodealliance/wasm-micro-runtime/issues/542) which caused by AssemblyScript lib, and this problem is solved in AssemblyScript 0.18. But I noticed that runtime is using the [deprecated api](https://www.assemblyscript.org/runtime.html), so I update it to adapt AssemblyScript 0.18 and have tested it in my wasm app.

ps: [AssemblyScript 0.18 GC API Doc](https://www.assemblyscript.org/garbage-collection.html#runtime-interface)

### DIFF
--- a/core/iwasm/aot/aot_loader.c
+++ b/core/iwasm/aot/aot_loader.c
@@ -1743,7 +1743,7 @@ load_from_sections(AOTModule *module, AOTSection *sections,
                     export_tmp = module->exports;
                     for (j = 0; j < module->export_count; j++, export_tmp++) {
                         if ((export_tmp->kind == EXPORT_KIND_FUNC)
-                            && (!strcmp(export_tmp->name, "__retain"))) {
+                            && (!strcmp(export_tmp->name, "__pin"))) {
                             func_index = export_tmp->index
                                             - module->import_func_count;
                             func_type_index =
@@ -1771,7 +1771,7 @@ load_from_sections(AOTModule *module, AOTSection *sections,
                 }
             }
             else if ((!strcmp(exports[i].name, "free"))
-                     || (!strcmp(exports[i].name, "__release"))) {
+                     || (!strcmp(exports[i].name, "__unpin"))) {
                 func_index = exports[i].index - module->import_func_count;
                 func_type_index = module->func_type_indexes[func_index];
                 func_type = module->func_types[func_type_index];
--- a/core/iwasm/aot/aot_loader.c
+++ b/core/iwasm/aot/aot_loader.c
@@ -1743,7 +1743,8 @@ load_from_sections(AOTModule *module, AOTSection *sections,
                     export_tmp = module->exports;
                     for (j = 0; j < module->export_count; j++, export_tmp++) {
                         if ((export_tmp->kind == EXPORT_KIND_FUNC)
-                            && (!strcmp(export_tmp->name, "__pin"))) {
+                            && (!strcmp(export_tmp->name, "__retain")
+                                || !strcmp(export_tmp->name, "__pin"))) {
                             func_index = export_tmp->index
                                             - module->import_func_count;
                             func_type_index =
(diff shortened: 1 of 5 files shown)


core/iwasm/aot/aot_loader.c
```diff
@@ -1743,7 +1743,7 @@ load_from_sections(AOTModule *module, AOTSection *sections,
                     export_tmp = module->exports;
                     for (j = 0; j < module->export_count; j++, export_tmp++) {
                         if ((export_tmp->kind == EXPORT_KIND_FUNC)
-                            && (!strcmp(export_tmp->name, "__retain"))) {
+                            && (!strcmp(export_tmp->name, "__pin"))) {
```
wenyongh, line 1746: There might be someone who is using the older asc compiler, we had better be compatible with the older version, could you please changing it to:
```C
&& (!strcmp(export_tmp->name, "__retain")
    || !strcmp(export_tmp->name, "__pin"))
```

core/iwasm/aot/aot_runtime.c
```diff
@@ -1501,7 +1501,7 @@ aot_module_malloc(AOTModuleInstance *module_inst, uint32 size,
             malloc_func_name = "__new";
             malloc_func_sig = "(ii)i";
             retain_func =
-                aot_lookup_function(module_inst, "__retain", "(i)i");
+                aot_lookup_function(module_inst, "__pin", "(i)i");
```
wenyongh, line 1504: lookup "__pin" firstly, if failed, lookup "__retain" again:
```C
retain_func = aot_lookup_function(.., "__pin", ..);
if (!retain_func)
    retain_func = aot_lookup_function(.., "__retain", ..);
...
```

core/iwasm/aot/aot_runtime.c
```diff
@@ -1566,7 +1566,7 @@ aot_module_free(AOTModuleInstance *module_inst, uint32 ptr)
             char *free_func_name;
 
             if (module->retain_func_index != (uint32)-1) {
-                free_func_name = "__release";
+                free_func_name = "__unpin";
             }
```
wenyongh, line 1570: Same as above

core/iwasm/aot/aot_loader.c
JavanZhu: Thank you for your thoughtful consideration, I'll change it and test it again.

core/iwasm/aot/aot_runtime.c
```diff
@@ -1566,13 +1568,15 @@ aot_module_free(AOTModuleInstance *module_inst, uint32 ptr)
...
             if (module->retain_func_index != (uint32)-1) {
-                free_func_name = "__unpin";
+                free_func_name = "__release";
             }
             else {
                 free_func_name = "free";
             }
             free_func =
                 aot_lookup_function(module_inst, free_func_name, "(i)i");
+            if (!free_func)
```
wenyongh, line 1578: Should add a condition:
```C
if (!free_func && module->retain_func_index != (uint32)-1)
```

JavanZhu: Hi @wenyongh, It would be more reasonable to add the condition and I will modify it. Could you please tell me when free_func_name is "free"? Does it exist an older version of AssemblyScript before 0.17?

wenyongh: Hi, the "free" function can be exported by wasi-sdk or emsdk for wasm application written by C language, it isn't exported by asc yet. Here retain_func_index is only set for asc compiler when both "__new" and "__retain" (or "__pin") functions are exported.

JavanZhu:
Hi @wenyongh, I have added the condition according to your suggestion, but pull request check is failed, I noticed that it's complained "wget: unable to resolve host address ‘github.com’" in Mac which look like a network problem probably, Could you please help me to check it? thanks.

PR #590, opened 2021-03-24
Android wasi support.

JavanZhu:
Android WASI support

Hi, I am trying to adapt WASI for Android platform. But there is some build error on Android because some system APIS  are
not well supported by some versions of Android system.

API name |  __ANDROID_API__
-- | --
preadv | 24
pwritev | 24
posix_fadvise | 21
posix_fallocate | 21
linkat | 21
symlinkat | 21
readlinkat | 21
telldir | 23
seekdir | 23
futimens | 19
getrandom | 28

So, here is some points to adapt it:
1. If ANDROID_PLATFORM is not set in CMakeLists.txt, the default value of __ANDROID_API__ is 16. In that situation, 
we define some empty functions to avoid build error and also print some errors to notice users.
2. And if ANDROID_PLATFORM is set to 28, all system api that libc WASI needed is supported, so there is no need to change current source code.


core/shared/platform/android/platform_init.c
```diff
@@ -5,6 +5,10 @@
 
 #include "platform_api_vmcore.h"
 #include "platform_api_extension.h"
+#include "bh_log.h"
+
+#define API_NOT_SUPPORT_ERROR(API, LEVEL) LOG_ERROR("%s() is only supported when compile option ANDROID_PLATFORM >= %s.", #API, #LEVEL);
```
wenyongh, line 10: In WAMR's design, the platform layer is the lowest layer, it should not depend any upper layer, so here we should not include bh_log.h and call LOG_ERROR. Could you please call __android_log_print instead, how about:
```C
#define API_NOT_SUPPORT_ERROR(API, VERSION) __android_log_print("%s() is only supported when __ANDROID_API__ >= %s.", #API, #VERSION)
```

core/shared/platform/android/platform_init.c
```diff
@@ -33,3 +37,69 @@ int os_vprintf(const char *fmt, va_list ap)
     return __android_log_vprint(ANDROID_LOG_INFO, "wasm_runtime::", fmt, ap);
 }
 
+#if __ANDROID_API__ < 19
+
+int futimens(int __dir_fd, const struct timespec __times[2]) {
```
wenyongh, line 42: Coding style, please put '{' to the next line:
```C
int futimens(int __dir_fd, const struct timespec __times[2])
{
  ...
}
```

core/shared/platform/android/platform_init.c
```diff
@@ -33,3 +37,69 @@ int os_vprintf(const char *fmt, va_list ap)
...
+    API_NOT_SUPPORT_ERROR(futimens, 19);
+    return -1;
+}
+
+#endif
+
+#if __ANDROID_API__ < 21
+
+int posix_fallocate(int __fd, off_t __offset, off_t __length) {
```
wenyongh, line 51: Same as above

core/shared/platform/android/platform_init.c
```diff
@@ -33,3 +37,69 @@ int os_vprintf(const char *fmt, va_list ap)
...
+    API_NOT_SUPPORT_ERROR(posix_fallocate, 21);
+    return -1;
+}
+
+int posix_fadvise(int fd, off_t offset, off_t len, int advice) {
+    API_NOT_SUPPORT_ERROR(posix_fadvise, 21);
+    return -1;
+}
+
+int linkat(int __old_dir_fd, const char *__old_path, int __new_dir_fd, const char *__new_path, int __flags) {
```
wenyongh, line 61: Coding style, the line is too long, please change it to
```C
int linkat(int __old_dir_fd, const char *__old_path,
           int __new_dir_fd, const char *__new_path, int __flags)
{
   ...
}
```

core/shared/platform/android/platform_internal.h
```diff
@@ -87,6 +87,44 @@ void os_sigreturn();
...
+
+#endif
+
+#if __ANDROID_API__ < 21
+
+int posix_fallocate(int __fd, off_t __offset, off_t __length);
+
+int posix_fadvise(int fd, off_t offset, off_t len, int advice);
+
+int linkat(int __old_dir_fd, const char *__old_path, int __new_dir_fd, const char *__new_path, int __flags);
```
wenyongh, line 104: Same as above, change to:
```C
int linkat(int __old_dir_fd, const char *__old_path,
           int __new_dir_fd, const char *__new_path, int __flags);
```

product-mini/platforms/android/CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ set (ANDROID_NDK $ENV{ANDROID_NDK_HOME})
 set (ANDROID_SDK $ENV{ANDROID_SDK_HOME})
 set (ANDROID_ABI "x86")
 set (ANDROID_LD lld)
+set (ANDROID_PLATFORM 24)
```
wenyongh, line 14: Seems that it isn't used, in fact the macro __ANDROID_API__  is used instead, could we remove this line?

JavanZhu: Some reasons make think we should "set (ANDROID_PLATFORM 24)" by default.
1.  refer to [Android NDK ANDROID PLATFORM Doc](https://developer.android.com/ndk/guides/cmake#android_platform),
seems that we should set "ANDROID_PLATFORM" to determine the value of macro __ANDROID_API__.
2. API level 24, equals Android system version 7.0, most of current android system version is larger that 7.0 .
image
In that situation, wamr libc WASI could supports most of APIs except getrandom().
(if we remove this line, instead we could tell developers that they must config "ANDROID_PLATFORM" if they want to support libc WASI  in the WAMR developer Document). 

If need, I can remove this line. How do you think? thanks.

wenyongh: Got it, but had better not hardcode it to 24, could you please modify it to
```bash
if (NOT DEFINED ANDROID_PLATFORM)
    set (ANDROID_PLATFORM 24)
endif ()
```
So developer can use `cmake -DANDROID_PLATFORM=xx` to set the version number if his version number is different.

JavanZhu: Ok, I'll change and test it later.

PR #616, opened 2021-04-19
Make it work for aarch64.

baohaojun:
Fixed a typo in aot_reloc_aarch64.c.

For aarch64 gcc, there's no -m32 switch.

For X86_64 gcc, there's now a build error when you use -m32 in ubuntu
20.04.

Anyway, let's fix the data type in the source code. All test cases are
passed.

Also, you may need to set APP_THREAD_STACK_SIZE_DEFAULT to 128K, because
with current value, there's a runtime error on our aarch64 board:

Invalid thread stack size 32768. Min stack size on Linux = 131072

Signed-off-by: Bao Haojun <baohaojun@lixiang.com>


test-tools/host-tool/src/transport.h
```diff
@@ -12,8 +12,8 @@ extern "C" {
 
 /* IMRT link message between host and WAMR */
 typedef struct {
-    unsigned short message_type;
-    unsigned long payload_size;
+    uint16_t message_type;
```
wenyongh, line 15: uint16_t requires <stdint.h>, normally we don't use uint16_t, uint32_t type except in core/iwasm/include/wasm_export.h, could you please remain it as unsigned short?

test-tools/host-tool/src/transport.h
```diff
@@ -12,8 +12,8 @@ extern "C" {
...
+    uint32_t payload_size;
```
wenyongh, line 16: Same as above, could you please change it to unsigned int?

samples/simple/profiles/arm64-interp/toolchain.cmake
```diff
@@ -0,0 +1,37 @@
...
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_GNU_SOURCE")
+
+
+if (NOT $ENV{ARM_A7_COMPILER_DIR} STREQUAL "")
+    SET (toolchain_sdk_dir $ENV{ARM_A7_COMPILER_DIR}/)
+endif ()
+
+if (NOT $ENV{ARM_A7_SDKTARGETSYSROOT} STREQUAL "")
+    SET(SDKTARGETSYSROOT $ENV{ARM_A7_SDKTARGETSYSROOT})
+    #SET(CMAKE_SYSROOT SDKTARGETSYSROOT)
```
wenyongh, line 19: Could you please remove this line? Seem it is useless.

samples/simple/profiles/arm64-aot/toolchain.cmake
```diff
@@ -0,0 +1,37 @@
...
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_GNU_SOURCE")
+
+
+if (NOT $ENV{ARM_A7_COMPILER_DIR} STREQUAL "")
+    SET (toolchain_sdk_dir $ENV{ARM_A7_COMPILER_DIR}/)
+endif ()
+
+if (NOT $ENV{ARM_A7_SDKTARGETSYSROOT} STREQUAL "")
+    SET(SDKTARGETSYSROOT $ENV{ARM_A7_SDKTARGETSYSROOT})
+    #SET(CMAKE_SYSROOT SDKTARGETSYSROOT)
```
wenyongh, line 19: Could you please remove this line? Seem it is useless.

PR #641, opened 2021-05-26
Allow calling pure C functions directly

pannous:
Added use_wrapper to structure NativeSymbol and corresponding methods to enable linking to pure C functions,  without prepending wasm_exec_env_t as first argument

example:
```
int square(int a) {
	return a * a;
}

static NativeSymbol native_symbols[] = {
				{       "square",  (void *) square,  "(i)i", NULL, /*use_wrapper*/ false}
};
```

### DIFF
--- a/core/iwasm/common/wasm_native.c
+++ b/core/iwasm/common/wasm_native.c
@@ -68,6 +68,15 @@ check_symbol_signature(const WASMType *type, const char *signature)
     if (*p++ != '(')
         return false;
 
+    uint32 param_count = (uint32) (p_end - p);
+    if(p_end[-1]!=')') param_count = param_count - 1;// no return
+    param_count = param_count - 1;// no wrapper 
+    bool params_missing = param_count != (uint32) (type->param_count );
+    if (params_missing){
+        printf("\nparams_missing in WASM function signature! Expected %d, got %d\n", param_count , type->param_count );
+        return false;
+    }
+
     if ((uint32)(p_end - p) < (uint32)(type->param_count + 1))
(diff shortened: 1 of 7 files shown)


wenyongh:
> Added use_wrapper to structure NativeSymbol and corresponding methods to enable linking to pure C functions, without prepending wasm_exec_env_t as first argument
> 
> example:
> 
> ```
> int square(int a) {
> 	return a * a;
> }
> 
> static NativeSymbol native_symbols[] = {
> 				{       "square",  (void *) square,  "(i)i", NULL, /*use_wrapper*/ false}
> };
> ```

Hi, adding the flag to control whether to pass the exec_env argument to native function or not may introduce more footprint and complexity for runtime, could you please simply add the wrapper function as normal, and call the pure C function in the wrapper function, e.g.:
```C
int square(int a) {
	return a * a;
}

int square_wrapper(wasm_exec_env_t exec_env, int a) {
	return square(a);
}
```
Thanks.

pannous:
Well this feature is only 8 lines of code, but if you don't want it for wider adoption that is OK.

I change the title in case others are looking for this functionality too.

pannous:
There are situations in which wasm-micro-runtime is used as a library and creating ad hoc wrappers is not an option.

tonibofarull:
If there are no additional checks, maybe it's easier to encapsulate your function with a macro.

wenyongh:
It is invalid to call pure C function directly since the wrapper function must have an extra argument `exec_env` which is passed from runtime when calling it.

Issue #219, opened 2020-03-31
format strings for fixed size integer should use fixed size format string macros


kfessel:
especially instead of buffer created format-strings 

    --- a/core/iwasm/common/wasm_runtime_common.c
    +++ b/core/iwasm/common/wasm_runtime_common.c
    @@ -1374,15 +1374,10 @@ wasm_application_execute_func(WASMModuleInstanceCommon *module_inst,
                break;
            case VALUE_TYPE_I64:
            {
    -            char buf[16];
                union { uint64 val; uint32 parts[2]; } u;
                u.parts[0] = argv1[0];
                u.parts[1] = argv1[1];
    -            if (sizeof(long) == 4)
    -                snprintf(buf, sizeof(buf), "%s", "0x%llx:i64");
    -            else
    -                snprintf(buf, sizeof(buf), "%s", "0x%lx:i64");
    -            os_printf(buf, u.val);
    +            os_printf("0x%"PRIx64":i64", u.val);
                break;
            }

wenyongh:
@kfessel thanks for the suggestion! Not sure whether all the platforms have this macro, seems it is better to check macro firstly to avoid compile error in some platform:

    #ifdef PRIx64
        os_printf("0x%"PRIx64":i64", u.val);
    #else
        use the original code
    #end

kfessel:
Since this is part of the C99 standard library and it is easy to implement i think most platforms support it.
lets do a quick check :
alios: gcc libc (+) or arm libc (+)
zephyr: minimal-libc(+) or newlibc (+)
since darwin is a full unix os using llvm i assume it has it in its libc
linux glibc musl  have it
androids boinc has it (copy from OpenBSD)
vwworks: windriver says they support C11 (i think this include C99)

even the avrlibc has it (since >12years)
microsoft was a bit late adding it 2015 since they focused on C++ 
according to https://www.cplusplus.com/reference/cinttypes/ these macros are also part of C++11

and if one needs to provide it there is the BSD licensed version from OpenBSD
https://android.googlesource.com/platform/bionic/+/refs/tags/android-wear-9.0.0_r14/libc/include/inttypes.h

but at this specific place the code currently in master is testing the compilers ability to deep dive into the code. The original should be:  ( at least give the compiler a chance to throw away the unneeded code)

        case VALUE_TYPE_I64:
        {
            char * buf;
            union { uint64 val; uint32 parts[2]; } u;
            u.parts[0] = argv1[0];
[...]

PR #654, opened 2021-06-18
Fix compile warnings with arm-none-eabi-gcc 9.3.1

masayuki2009:
## Summary

- This PR fixes compile warnings with arm-none-eabi-gcc

## Impact

- None

## Testing

- Tested with sabre-6quad (Cortex-A9/NuttX/QEMU)
- Tested with ubuntu20.04 x86_64

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -2300,7 +2300,7 @@ invoke_wasm_c_api_native(AOTModuleInstance *module_inst, void *func_ptr,
     }
     if (trap) {
         if (trap->message->data) {
-            snprintf(fmt, sizeof(fmt), "%%.%us", (uint32)trap->message->size);
+            snprintf(fmt, sizeof(fmt), "%%.%"PRIu32"s", (uint32)trap->message->size);
             snprintf(module_inst->cur_exception,
                      sizeof(module_inst->cur_exception),
                      fmt, trap->message->data);
--- a/core/iwasm/aot/arch/aot_reloc_thumb.c
+++ b/core/iwasm/aot/arch/aot_reloc_thumb.c
@@ -304,7 +304,7 @@ apply_relocation(AOTModule *module,
             if (error_buf != NULL)
                 snprintf(error_buf, error_buf_size,
                          "Load relocation section failed: "
-                         "invalid relocation type %d.",
+                         "invalid relocation type %"PRIu32".",
                          reloc_type);
             return false;
     }
--- a/core/iwasm/common/wasm_application.c
+++ b/core/iwasm/common/wasm_application.c
@@ -423,7 +423,7 @@ wasm_application_execute_func(WASMModuleInstanceCommon *module_inst,
         char *endptr = NULL;
         bh_assert(argv[i] != NULL);
         if (argv[i][0] == '\0') {
-            snprintf(buf, sizeof(buf), "invalid input argument %d", i);
+            snprintf(buf, sizeof(buf), "invalid input argument %"PRId32"", i);
             wasm_runtime_set_exception(module_inst, buf);
             goto fail;
         }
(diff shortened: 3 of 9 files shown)


core/iwasm/aot/aot_runtime.c
```diff
@@ -2300,7 +2300,7 @@ invoke_wasm_c_api_native(AOTModuleInstance *module_inst, void *func_ptr,
     }
     if (trap) {
         if (trap->message->data) {
-            snprintf(fmt, sizeof(fmt), "%%.%us", (uint32)trap->message->size);
+            snprintf(fmt, sizeof(fmt), "%%.%"PRIu32"s", (uint32)trap->message->size);
```
wenyongh, line 2303: %u, %d, %i and %x are common formats for printf/snprintf function, are they not supported by arm-none-eabi-gcc compiler? If we change code here, does it mean that we need to use PRIu32/PRId32/PRIx32 related strings in future code, which is really not convenient for most of the developers.

Another issue is that some platform doesn't support these strings, e.g. some embedded OSs, if we change the code, we had better add related macros in core/shared/platform/include/platform_common.h, e.g.:
```C
typedef int64_t int64;

#ifndef PRI64_PREFIX
# if UINTPTR_MAX == UINT64_MAX
#  define PRI64_PREFIX    "l"
#  define PRIPTR_PREFIX   "l"
# else
#  define PRI64_PREFIX    "ll"
#  define PRIPTR_PREFIX
# endif
#endif

#ifndef PRId32
# define PRId32     "d"
#endif
#ifndef PRIi32
# define PRIi32     "i"
#endif
#ifndef PRIu32
# define PRIu32     "u"
#endif
#ifndef PRIx32
# define PRIx32     "x"
#endif
#ifndef PRIX32
# define PRIX32     "X"
#endif
#ifndef PRIx64
# define PRIx64     PRI64_PREFIX "x"
#endif
```

masayuki2009: >%u, %d, %i and %x are common formats for printf/snprintf function, are they not supported by arm-none-eabi-gcc compiler?

Of course, they are supported by arm-none-eabi-gcc.
However, our environment (i.e. NuttX) typedefs _uint32_t as follows.

```
typedef unsigned long _uint32_t;
```

That's why the format warnings happen.

>Another issue is that some platform doesn't support these strings, e.g. some embedded OSs,

In my understanding, those strings should be supported by the recent compilers because they are defined in C99.
However, some RTOS (e.g. Zephyr) define those types in a header file in the RTOS.

core/iwasm/aot/arch/aot_reloc_thumb.c
```diff
@@ -304,7 +304,7 @@ apply_relocation(AOTModule *module,
             if (error_buf != NULL)
                 snprintf(error_buf, error_buf_size,
                          "Load relocation section failed: "
-                         "invalid relocation type %d.",
+                         "invalid relocation type %"PRIu32".",
```
wenyongh, line 307: Could we change to:
```C
snprintf(error_buf, error_buf_size,
         "Load relocation section failed: "
         "invalid relocation type %u.",
         (unsigned)reloc_type);
```

core/iwasm/aot/aot_runtime.c
masayuki2009: Yes, we can change.

core/iwasm/aot/arch/aot_reloc_thumb.c
masayuki2009: Yes, this can also be changed.

kfessel:
somehow i see #219 in this

masayuki2009:
> somehow i see #219 in this

@kfessel 
Thanks for the comments.
Let me close this PR to reconsider.

PR #675, opened 2021-07-28
Unify iwasm implement for POSIX compatible system

no1wudi:
Merge iwasm for NuttX、Linux、Darwin、VxWorks into one since they are POSIX compatible with little difference, such as global heap pool size.


wenyongh (commented):
Some main.c files are soft link files, right?
WAMR supports Windows platform and the repo may be cloned under Windows, using soft link file is not friendly for Windows, can we keep the original main.c file, and only add a line:
```C
#include "../posix/main.c"

no1wudi:
> Some main.c files are soft link files, right?
> WAMR supports Windows platform and the repo may be cloned under Windows, using soft link file is not friendly for Windows, can we keep the original main.c file, and only add a line:
> 
> ```c
> #include "../posix/main.c"
> ```

Done

PR #701, opened 2021-08-19
lib-pthread: Fix pthread_create

yamt:
The last argument of pthread_create is opaque for the library/host.
It should be returned to the app as it is without any interpretations.
There is no point to translate to native address.
Actually it harms because the validation seems broken.

With this change, pthread_create/join seems working reasonably
with libc-wasi on macOS.

### DIFF
--- a/core/iwasm/libraries/lib-pthread/lib_pthread_wrapper.c
+++ b/core/iwasm/libraries/lib-pthread/lib_pthread_wrapper.c
@@ -112,7 +112,7 @@ typedef struct {
     /* table elem index of the app's entry function */
     uint32 elem_index;
     /* arg of the app's entry function */
-    void *arg;
+    uint32 arg;
     wasm_module_inst_t module_inst;
 } ThreadRoutineArgs;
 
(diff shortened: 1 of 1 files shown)


xujuntwt95329 (commented):
LGTM

core/iwasm/libraries/lib-pthread/lib_pthread_wrapper.c
```diff
@@ -479,17 +479,8 @@ pthread_start_routine(void *arg)
...
-            all other instance in this cluster */
-        wasm_cluster_spread_exception(exec_env);
-        wasm_runtime_deinstantiate_internal(module_inst, true);
-        delete_thread_info_node(info_node);
-        return NULL;
-    }
-
     wasm_exec_env_set_thread_info(exec_env);
-    argv[0] = addr_native_to_app(routine_args->arg);
+    argv[0] = routine_args->arg;
```
wenyongh, line 483: The type of routine_args->arg is void *, converting to uint32 causes compilation warning, could you please change the type of field arg in ThreadRoutineArgs to uint32?

And also change the last argument in pthread_create_wrapper from `void *arg` to `uint32 arg`.

yamt: ok

yamt: done

PR #711, opened 2021-09-01
Make intrinsic flags readable

no1wudi:


core/iwasm/aot/aot_intrinsic.h
```diff
@@ -23,47 +23,47 @@ extern "C" {
  *   - The lower 48 bits are the intrinsic capability mask
  */
 
-#define AOT_INTRINSIC_FLAG(group, value)                                      \
-    ((((uint64)(group & 0xffffLL)) << 48) | (uint64)value)
+#define AOT_INTRINSIC_FLAG(group, number)                                     \
+    ((((uint64)(group & 0xffffLL)) << 48) | (uint64)(1 << number))
```
wenyongh, line 27: Should be ((uint64)1 << number), as (1 << number) is no more than UINT32_MAX, the highest 32 bits are 0 when converting it to uint64?

no1wudi: Done

PR #713, opened 2021-09-01
Move IntCond & FloatCond to aot.h

no1wudi:
Move them to aot.h from aot_compiler.h since they will be used in aot_intrinsics.c

### DIFF
--- a/core/iwasm/compilation/aot_compiler.h
+++ b/core/iwasm/compilation/aot_compiler.h
@@ -13,28 +13,8 @@
 extern "C" {
 #endif
 
-typedef enum IntCond {
-  INT_EQZ = 0,
-  INT_EQ,
-  INT_NE,
-  INT_LT_S,
-  INT_LT_U,
-  INT_GT_S,
-  INT_GT_U,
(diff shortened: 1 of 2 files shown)


core/iwasm/compilation/aot_compiler.h
```diff
@@ -13,29 +13,6 @@
...
-
-typedef enum FloatCond {
-  FLOAT_EQ = 0,
-  FLOAT_NE,
-  FLOAT_LT,
-  FLOAT_GT,
-  FLOAT_LE,
-  FLOAT_GE
-} FloatCond;
-
```
wenyongh, line 38: Is there any compile error or warning? Do we need to move them and why just move part of the enum types into aot.h?

no1wudi: No error or warning found with this patch since if the `aot_compiler.h` included the 'aot.h' had also included.

Move these flags because they will be used both in wamrc and runtime intrinsic (aot_intrinsics.c), some LLVM relative headers are not available on embedded platform.

PR #747, opened 2021-09-12
allocate correct space for results before native callback is run

### LINKED ISSUE #746: Possible bad allocation of `results` in `wasm_runtime_invoke_c_api_native()`
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/common/wasm_runtime_common.c#L4147-L4152

1) It looks like if the test for `func_type->result_count > 4` passes, we will malloc a new block of memory into `results`, this will cause us to loose the actual `results` data.

2) It looks like this code  should be run before https://github.com/bytecodealliance/wasm-micro-runtime/blob/e07b58062f101a85326c113c5bd783b220d31606/core/iwasm/common/wasm_runtime_common.c#L4120, so that the proper amount of storage for results is created upfront.

I would be happy to raise a quick PR if my understanding from reading the code is found to be correct.

saju:
Allocate space for results before invoking the native function in `wasm_runtime_invoke_c_api_native()`

fixes #746

### DIFF
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -4117,6 +4117,13 @@ wasm_runtime_invoke_c_api_native(WASMModuleInstanceCommon *module_inst,
         goto fail;
     }
 
+    if (func_type->result_count > 4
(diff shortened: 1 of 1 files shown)


wenyongh:
@saju Thanks for reporting the issue!

PR #766, opened 2021-09-28
Fix WASI misalignments

csegarragonz:
Some of the WASI definitions are out of sync with those of [wasi-libc](https://github.com/WebAssembly/wasi-libc/blob/main/libc-bottom-half/headers/public/wasi/api.h).

I think I have changed all differences, but the file structure in WAMR makes it difficult to diff both files.

### DIFF
--- a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/include/wasmtime_ssp.h
+++ b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/include/wasmtime_ssp.h
@@ -35,10 +35,9 @@ _Static_assert(_Alignof(int16_t) == 2, "non-wasi data layout");
 _Static_assert(_Alignof(uint16_t) == 2, "non-wasi data layout");
 _Static_assert(_Alignof(int32_t) == 4, "non-wasi data layout");
 _Static_assert(_Alignof(uint32_t) == 4, "non-wasi data layout");
-#if 0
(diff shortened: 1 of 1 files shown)


csegarragonz:
cc @wenyongh: android tests are failing and i am not sure why

csegarragonz:
@wenyongh yes, i had already seen that. i was wondering if you could point out what was the issue? i suspect maybe `sizeof(uint64_t)` in android is not 8, and thus why those checks fail. but i am not sure how to fix that neither.

kfessel:
``` 
typedef struct __wasi_filestat_t {
    __wasi_device_t st_dev;
    __wasi_inode_t st_ino;
    __wasi_filetype_t st_filetype;
    __wasi_linkcount_t st_nlink;
    __wasi_filesize_t st_size;
    __wasi_timestamp_t st_atim;
    __wasi_timestamp_t st_mtim;
    __wasi_timestamp_t st_ctim;
} __wasi_filestat_t __attribute__((aligned(8)));
```
it is about the alignment the struct description above means alling the struct by 8 bytes but doesn't tell the compiler what to do about its members  -> system alignment which maybe (aligned(4) (32bit) 

the gcc docu about this is a little sparse: https://gcc.gnu.org/onlinedocs/gcc-3.3/gcc/Type-Attributes.html (but somhow the information is in there)
kiel has a more detailed documentation: https://www.keil.com/support/man/docs/armclang_ref/armclang_ref_chr1385461015401.htm

PR #769, opened 2021-09-29
Implement source debugging for interpreter and AOT

wenyongh:
Implement source debugging feature for classic interpreter and AOT:
- use `cmake -DWAMR_BUILD_DEBUG_INTERP=1` to enable interpreter debugging
- use `cmake -DWAMR_BUILD_DEBUG_AOT=1` to enable AOT debugging

See doc/source_debugging.md for more details.


mbbill:
tbh this whole things is **very** unstable. I've found numerous programming errors on that branch, including some serious ones like misuse of mutex, using uninitialized memory and etc. The debugging feature barely works and I got tons of crashes in many different places.
It makes me worried more about the quality of the entire project. Reports from our internal tests shows the stability of wamr is worse than wasm3 and I hope you guys can put more effort on making sure things works before merging them into the trunk so quickly.
just saying. If you do care about the project quality, maybe I can help adding more tests, especially unit tests later on, but I'm not sure if that's on your list.

xwang98:
@mbbill Thank you for raising the issue for the source debugging patch. Firstly I recognize the stability of this patch is not at the quality level as we used to ask. There are a few reasons for the team deciding to merge it as an experimental feature.
1. This is the first large feature co-developed by the community since WAMR has been changed to an open governance model. The core team already realized the existing process can't work well for guaranteeing quality.  One action in progress is to enforce the code format for the patch submissions  (see https://github.com/bytecodealliance/wasm-micro-runtime/pull/770).  As reformatting the codebase will change every source file, merging such a big patch later will cause huge trouble in solving the conflicts. And more actions for enhancing the process and quality infrastructure are WIP.
2. We have tested the feature and it is basically working with a few known issues. The feature can be disabled by the compilation option. We have a plan to continue to polish this feature from the main branch.

Clearly quality is the top priority of WAMR project and the core team has spent significant efforts on quality assurance. As the project is becoming much larger scale in terms of contributor and code size since the initial release, ensuring high quality became challenging. This alerting message is just right in time and your offer to help the quality is very appreciated!

In our test of the debugging feature, the experience looks not as bad as yours. We wonder if there is any issue in the version or compilation under your environment. Could you please raise an issue for the problem you found and experienced? And possibly patches for fixing it would be awesome.

I am also very interested in your internal test report and I plan to drop you an email to see if we can get it privately. It will be excellent if you can raise issues for the problems found, and we are happy to fix them. Our internal test didn't show any quality advance by wasm3 while the WAMR code size is way bigger than wasm3. However anyway it will be really great to learn from other test reports.

core/iwasm/libraries/debug-engine/debug_engine.c
```diff
@@ -0,0 +1,1011 @@
...
+    }
+
+    if (!(instance = wasm_runtime_malloc(sizeof(WASMDebugInstance)))) {
+        LOG_ERROR("WASM Debug Engine error: failed to allocate memory");
+        return NULL;
+    }
+    memset(instance, 0, sizeof(WASMDebugInstance));
+    instance->cluster = cluster;
+    instance->control_thread =
+      wasm_debug_control_thread_create((WASMDebugObject *)instance);
```
mbbill, line 220: the idea of forced conversion between the `WASMDebugObject` and `WASMDebugInstance` is a huge red flag to me.  It's confusing, error-prone and in fact doesn't work. e.g. there's no guarantee that the compiler will generate exactly same layout for the two different structs. Having identical member in identical order doesn't automatically get that guarantee, although it *seems* work for most compilers. I just don't understand why ding this instead of simply aggregate them. btw, there's no document for the relationship between the two structs either. IMO it's a pretty bad design

mbbill: also, as mentioned above, the access to the debug object (or instance? whatever) needs to be protected from data race.

core/iwasm/libraries/debug-engine/packets.c
```diff
@@ -0,0 +1,173 @@
...
+    write_packet_bytes(gdbserver, buf, buf_num_bytes);
+    free(buf);
+}
+
+bool
+skip_to_packet_start(WASMGDBServer *gdbserver)
+{
+    ssize_t end = -1;
+
+    for (size_t i = 0; i < gdbserver->pkt.end; ++i)
```
mbbill, line 146: Another potential crash caused by using uninitialized `pkt.end`. Should've set it to zero before the first read.

core/iwasm/libraries/debug-engine/gdbserver.c
```diff
@@ -0,0 +1,177 @@
...
+
+static void
+process_packet(WASMGDBServer *server)
+{
+    uint8_t *inbuf = server->pkt.buf;
+    int inbuf_size = server->pkt.end;
+    uint8_t *packetend_ptr = (uint8_t *)memchr(inbuf, '#', inbuf_size);
+    int packetend = packetend_ptr - inbuf;
+    bh_assert('$' == inbuf[0]);
+    char request = inbuf[1];
```
mbbill, line 154: in some conditions `pkt.end` can be 1. which means the `inbuf[1]` and `inbuf[2]` will read invalid memory, which won't necessarily cause a crash but is still a serious issue.

mbbill: if there's a hard-coded `inbuf[2]` better put an assert but IMO it's the protocol handling code needs some more work.

mbbill (commented):
there are numerous other issues I hit couple times but couldn't remember.

core/iwasm/libraries/debug-engine/gdbserver.c
mbbill: btw, the name `pkt.end` is confusing. considering there's a `pkt.buf`, the `pkt.end` sounds like a pointer to the end. Should've used the name `size` instead.

xwang98:
@mbbill very appreciated!

hangedfish:
I get an error message `Failed to init stack guard pages` which execute `iwasm` with parameter `-g=127.0.0.1:1234`

```
➜  build ./iwasm ./main.wasm 
Hello world!
buf ptr: 0x114f0
buf: 1234
➜  build ./iwasm -g=127.0.0.1:1234 ./main.wasm
Failed to init stack guard pages
```

* CMake Flags: `-DWAMR_BUILD_DEBUG_INTERP=1`
* OS: MacOS 12.2.1
* CPU: Apple M1 Pro chip
* Compiler:
  Apple clang version 13.0.0 (clang-1300.0.29.30)
  Target: arm64-apple-darwin21.3.0
  Thread model: posix

core/iwasm/libraries/debug-engine/handler.c
```diff
@@ -0,0 +1,598 @@
...
+#include <limits.h>
+#include <string.h>
+#include <unistd.h>
+
+#include "debug_engine.h"
+#include "packets.h"
+#include "utils.h"
+#include "wasm_runtime.h"
+
+#define MAX_PACKET_SIZE (0x20000)
```
yamt, line 17: how did you calculate this size?
does this actually need this big buffer?
asking because i want to use this stuff on smaller machines.

PR #776, opened 2021-10-09
refine debug control thread and fix some memory issues

xujuntwt95329:
- fix data race between debug control thread and main thread
- fix possible memory leak in breakpoints list
- fix memory uninitialized issues
- remove unused data structures
- fix mini-loader issues
- fix config_common prompt


xwang98:
@mbbill I can't add your name into the reviewers. Your comments are appreciated. :-)

wenyongh:
LGTM

PR #807, opened 2021-10-28
support random debug port by assigning port = 0

xujuntwt95329:
If we don't want to specify a debug port, then we can use `-g=127.0.0.1:0` to get a random port

### DIFF
--- a/core/iwasm/libraries/debug-engine/gdbserver.c
+++ b/core/iwasm/libraries/debug-engine/gdbserver.c
@@ -104,7 +104,7 @@ wasm_launch_gdbserver(char *host, int port)
         goto fail;
     }
 
-    LOG_VERBOSE("Listening on %s:%d\n", host, ntohs(addr.sin_port));
+    LOG_WARNING("Listening on %s:%d\n", host, ntohs(addr.sin_port));
 
     ret = listen(listen_fd, 1);
     if (ret < 0) {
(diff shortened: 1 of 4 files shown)


core/iwasm/libraries/debug-engine/gdbserver.c
```diff
@@ -98,6 +98,14 @@ wasm_launch_gdbserver(char *host, int port)
         goto fail;
     }
 
+    socklen = sizeof(addr);
+    if (getsockname(listen_fd, (void *)&addr, &socklen) == -1) {
+        LOG_ERROR("%s", strerror(errno));
+        goto fail;
+    }
+
+    LOG_VERBOSE("Listening on %s:%d\n", host, ntohs(addr.sin_port));
```
wenyongh, line 107: Suggest to use LOG_WARNING() instead so user can always see the port number without adding "-v=5" to iwasm, or he won't be able to know the port number without "-v=5".

xujuntwt95329: Done

core/iwasm/libraries/debug-engine/gdbserver.c
```diff
@@ -98,6 +98,14 @@ wasm_launch_gdbserver(char *host, int port)
         goto fail;
     }
 
+    socklen = sizeof(addr);
+    if (getsockname(listen_fd, (void *)&addr, &socklen) == -1) {
+        LOG_ERROR("%s", strerror(errno));
+        goto fail;
+    }
+
+    LOG_WARNING("Listening on %s:%d\n", host, ntohs(addr.sin_port));
```
wenyongh, line 107: And shall we update the iwasm help and source debug document, to mention that when port is 0, a random port will be used?

xujuntwt95329: Done

PR #809, opened 2021-10-29
aot_reloc_x86_64: Fix pointer overflows

yamt:
Detected by UBSan

```
/root/src/wasm-micro-runtime/core/iwasm/aot/arch/aot_reloc_x86_64.c:232:43: runtime error: addition of unsigned offset to 0x000041209004 overflowed to 0x000041209000
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /root/src/wasm-micro-runtime/core/iwasm/aot/arch/aot_reloc_x86_64.c:232:43 in
```

### DIFF
--- a/core/iwasm/aot/arch/aot_reloc_x86_64.c
+++ b/core/iwasm/aot/arch/aot_reloc_x86_64.c
@@ -225,12 +225,13 @@ apply_relocation(AOTModule *module, uint8 *target_section_addr,
                       - get_plt_table_size()
                       + get_plt_item_size() * symbol_index;
                 target_addr = (intptr_t) /* L + A - P */
-                    (plt + reloc_addend - (target_section_addr + reloc_offset));
+                    ((uintptr_t)plt + reloc_addend
+                     - (uintptr_t)(target_section_addr + reloc_offset));
             }
             else {
                 target_addr = (intptr_t) /* L + A - P */
-                    ((uint8 *)symbol_addr + reloc_addend
-                     - (target_section_addr + reloc_offset));
+                    ((uintptr_t)symbol_addr + reloc_addend
+                     - (uintptr_t)(target_section_addr + reloc_offset));
             }
 
 #if defined(BH_PLATFORM_WINDOWS)


core/iwasm/aot/arch/aot_reloc_x86_64.c
```diff
@@ -225,12 +225,13 @@ apply_relocation(AOTModule *module, uint8 *target_section_addr,
...
-                    (plt + reloc_addend - (target_section_addr + reloc_offset));
+                    ((uintptr_t)plt + reloc_addend
+                     - (uintptr_t)(target_section_addr + reloc_offset));
             }
             else {
                 target_addr = (intptr_t) /* L + A - P */
-                    ((uint8 *)symbol_addr + reloc_addend
-                     - (target_section_addr + reloc_offset));
+                    ((uintptr_t)symbol_addr + reloc_addend
+                     - (uintptr_t)(target_section_addr + reloc_offset));
```
wenyongh, line 234: Shall we also fix other places, e.g. L168-L170?

wenyongh: And not known which expression causes integer overflow? target_section_addr + reloc_offset, plt + reloc_addend, or (plt + reloc_addend - (target_section_addr + reloc_offset))?  Here the type of plt and target_section_addr is `uint8 *`, the range is 0 ~ 2^64-1 in x86-64, a little confused why there is integer overflow.

yamt: * iirc, pointer overflows like these are undefined behavior in C. unsigned integer overflows are ok.
* i haven't checked other places. i just fixed a few places which actually got complaints from UBSAN during my tests.

wenyongh: OK, so the code inserted by the compiler detected pointer overflow and reported error, from the output log `runtime error: addition of unsigned offset to 0x000041209004 overflowed to 0x00004120900 ... aot_reloc_x86_64.c:232:43`，the overflow should occur  at `(uint8 *)symbol_addr + reloc_addend`. We will merge this PR firstly and check whether there are other places to fix. Thanks a lot!

PR #811, opened 2021-10-30
Build projects in wamr-test-suites with multiple cores

wenyongh:
Build several projects in wamr-test-suites with multiple cores to reduce
the compilation time.


tests/wamr-test-suites/test_wamr.sh
```diff
@@ -208,7 +208,7 @@ function unit_test()
 
         # keep going and do not care if it is success or not
         make -ki clean | true
-        cmake ${compile_flag} ${WORK_DIR}/../../unit && make
+        cmake ${compile_flag} ${WORK_DIR}/../../unit && make -j 4
```
xujuntwt95329, line 211: had better use `$(nproc)`
```suggestion
        cmake ${compile_flag} ${WORK_DIR}/../../unit && make -j $(nproc)
```

tests/wamr-test-suites/test_wamr.sh
```diff
@@ -506,7 +506,7 @@ function build_wamrc()
         && if [ -d build ]; then rm -r build/*; else mkdir build; fi \
         && cd build \
         && cmake .. \
-        && make
+        && make -j 4
```
xujuntwt95329, line 509: Same as above

tests/wamr-test-suites/test_wamr.sh
lum1n0us: nproc may not be installed. 4 might be a safe option

xujuntwt95329: Got it, thanks.

xujuntwt95329:
LGTM

tests/wamr-test-suites/test_wamr.sh
wenyongh: Yes, so let's use -j 4.
Another place to reduce the workloads compilation time is building iwasm in compilation_on_android_ubuntu_macos.yml, currently we get llvm libraries
before building iwasm for each mode, in fact we only need to get llvm libraries
for the mode jit and lazyjit. For fast/classic interpreter mode and aot mode,
we don't need to get the llvm libraires, is it easy to add condition check
in the scripts?
```sh
- name: Get LLVM libraries
        id: cache_llvm
        if: ${{ matrix.light == 'green' }}
        uses: actions/cache@v2
        with:
          path: |
            ./core/deps/llvm/build/bin
            ./core/deps/llvm/build/include
            ./core/deps/llvm/build/lib
            ./core/deps/llvm/build/libexec
            ./core/deps/llvm/build/share
          key: ${{ matrix.os }}-${{ env.LLVM_CACHE_SUFFIX }}
```
e.g. modify this line: if: ${{ matrix.light == 'green' }} to something like:
```sh
if: ${{ matrix.light == 'green'
        && (make_options_run_mode == $JIT_BUILD_OPTIONS
            || make_options_run_mode == $LAZY_JIT_BUILD_OPTIONS)
     }}
```

lum1n0us: Time consumption is defined by the longest job in parallel mode. In our workflow, the followings will not start until all five build_iwasm jobs are finished.

wenyongh: There should be total 243 build_iwasm jobs but not five in compilation_on_android_ubuntu_macos.yml:
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/1402357515
build_iwasm
It is the longest phase in all the workflows.
Currently each build_iwasm job will get LLVM libraries firstly which costs some time, but there
are a lot of interpreter/aot mode jobs which don't need llvm libraries, if we can skip this step for them,
we should be able to reduce the total time of  243 build_iwasm jobs, so as to reduce the total time of
all workflows.

PR #814, opened 2021-11-02
Reduce memory usage by string interning for AOT

no1wudi:
Rivisit: This patch may break the backward compatibility

It can save ~2392 byte for workload coremark, deeply embedded system may take benefit from it.

### DIFF
--- a/core/iwasm/aot/aot_loader.c
+++ b/core/iwasm/aot/aot_loader.c
@@ -215,29 +215,44 @@ static char *
 const_str_set_insert(const uint8 *str, int32 len, AOTModule *module,
                      char *error_buf, uint32 error_buf_size)
 {
-    HashMap *set = module->const_str_set;
-    char *c_str, *value;
+    char *c_str = (char *)str, *value;
 
-    if (!(c_str = loader_malloc((uint32)len + 1, error_buf, error_buf_size))) {
-        return NULL;
+    /* Insert string into set if load from sections */
+    if (module->is_load_from_sections) {
+        HashMap *set = module->const_str_set;
+
+        if (!(c_str =
+                  loader_malloc((uint32)len + 1, error_buf, error_buf_size))) {
+            return NULL;
+        }
+
+        bh_memcpy_s(c_str, (uint32)(len + 1), str, (uint32)len);
+        c_str[len] = '\0';
+
+        if ((value = bh_hash_map_find(set, c_str))) {
+            wasm_runtime_free(c_str);
+            return value;
+        }
+
+        if (!bh_hash_map_insert(set, c_str, c_str)) {
+            set_error_buf(error_buf, error_buf_size,
+                          "insert string to hash map failed");
+            wasm_runtime_free(c_str);
+            return NULL;
+        }
+    }
+    /* If load from file buffer and is relocation mode */
+    else if (module->native_symbol_count == 0) {
+        /* Convert string to '\0' in place */
+        value = c_str - sizeof(uint16);
+        c_str = memmove(value, c_str, len);
+        c_str[len] = '\0';   
+        printf("Len:%s,%d\n",c_str, len);
     }
 
-    bh_memcpy_s(c_str, (uint32)(len + 1), str, (uint32)len);
-    c_str[len] = '\0';
+    /* Direct return this string if in indirect mode */
 
-    if ((value = bh_hash_map_find(set, c_str))) {
-        wasm_runtime_free(c_str);
-        return value;
-    }
-
-    if (!bh_hash_map_insert(set, c_str, c_str)) {
-        set_error_buf(error_buf, error_buf_size,
-                      "insert string to hash map failed");
-        wasm_runtime_free(c_str);
-        return NULL;
-    }
-
-    return c_str;
+    return (char *)c_str;
 }
 
 static bool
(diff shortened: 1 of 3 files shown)


wenyongh:
@no1wudi there are two issues: one is app manager, the aot file stream is transferred from host, app manager allocates memory for each section when a whole section is received, and after loading, app manager will destroy some unused sections to reduce footprint. This patch will lead to illegal memory accessing, e.g. wasm_runtime_lookup_function() accesses the exported function name in export list. The other issue is that it breaks the backward compatibility, runtime might not be able to run aot file generated before.
To be discussed, have no good idea yet.

no1wudi:
@wenyongh  How about make it as a runtime option ? For advanced users, they can use it manually.
And another question, what's policy for breaking changes ?

wenyongh:
> @wenyongh How about make it as a runtime option ? For advanced users, they can use it manually. And another question, what's policy for breaking changes ?

Hi, it really reduces footprint, but we had better not change the AOT file format, I got a rough idea:
(1) Add a macro to control it, by default it is disabled
(2) When macro is enabled, no need to create string hash set, just use the 2-byte size to re-org
     the string: move the string to 2-byte backward and put '\0' at the end, or,
     size+string => string + '\0' + '\0'. And use that string directly.
(3) When macro is disabled, use the original process
(4) Enable the macro if required, for iwasm in some platforms, we can enable it

To be discussed with other TSC members to get more feedback.

no1wudi:
> Hi, it really reduces footprint, but we had better not change the AOT file format.
In this patch, aot file format has no actual change but emit string with an extra '\0',
is there any side effect? It should works well with old version.

The key point is what's the runtime does, for this change, I think no need to create a new macro to make build system more complex.

wenyongh:
> > Hi, it really reduces footprint, but we had better not change the AOT file format.
> 
> In this patch, aot file format has no actual change but emit string with an extra '\0', is there any side effect? It should works well with old version.
> 
> The key point is what's the runtime does, for this change, I think no need to create a new macro to make build system more complex.

The previous design of string format already aims to reduce the footprint, we prefer to not change it: (1) emitting '\0' at the end, and adding 2-byte size at the beginning, it looks a little strange and increases the AOT file size, (2) if the AOT file is not read-only and can be referred to after loading, we can just move string backward and append '\0' to make it as a normal string to use, no need to change the AOT file format, we had better not change AOT file format unless we had to do, (3) in fact the string in wasm file also isn't terminated with '\0', the format is size_of_leb + string, we can also adjust it to a normal string to use, so we can do the same change in both interpreter and aot to reduce the footprint.

For the runtime, adding option is also OK, but how to do? Adding an option in RuntimeInitArgs and setting it before calling wasm_runtime_full_init?

For the runtime behavior, if we can reuse the string in aot file, we can just use it after adjusting it, no need to add to string set.

no1wudi:
> For the runtime behavior, if we can reuse the string in aot file, we can just use it after adjusting it, no need to add to string set.

Yes

> (1) emitting '\0' at the end, and adding 2-byte size at the beginning, it looks a little strange and increases the AOT file size 

Do this to keep the backward capability , and it's not harmful but an extra '\0'.

> (2) if the AOT file is not read-only and can be referred to after loading, we can just move string backward and append '\0' to make it as a normal string to use, no need to change the AOT file format, we had better not change AOT file format unless we had to do

1. This change isn't harmful or break the capability
2. In XIP mode, aot file is read-only , we should consider

> For the runtime, adding option is also OK, but how to do? Adding an option in RuntimeInitArgs and setting it before calling wasm_runtime_full_init?

For aot, maybe we can add a param to `aot_load_from_aot_file` ?

wenyongh:
> > For the runtime behavior, if we can reuse the string in aot file, we can just use it after adjusting it, no need to add to string set.
> 
> Yes
> 
> > (1) emitting '\0' at the end, and adding 2-byte size at the beginning, it looks a little strange and increases the AOT file size
> 
> Do this to keep the backward capability , and it's not harmful but an extra '\0'.
> 
> > (2) if the AOT file is not read-only and can be referred to after loading, we can just move string backward and append '\0' to make it as a normal string to use, no need to change the AOT file format, we had better not change AOT file format unless we had to do
> 
> 1. This change isn't harmful or break the capability
> 2. In XIP mode, aot file is read-only , we should consider

Could we emit '\0' only when XIP mode is enabled by wamrc? For other modes, it really increases aot file size and it is unnecessary, we had better not do that.
And when option is disabled, keep the original behavior, create a new string and put it into the string set. When option is enabled, check whether the string is '\0' terminated, if yes, use it directly, else adjust the string and them use it.

> 
> > For the runtime, adding option is also OK, but how to do? Adding an option in RuntimeInitArgs and setting it before calling wasm_runtime_full_init?
> 
> For aot, maybe we can add a param to `aot_load_from_aot_file` ?

aot_load_from_aot_file is at a relatively upper layer, had better not change it. There are two APIs exported: 
```C
WASM_RUNTIME_API_EXTERN wasm_module_t
wasm_runtime_load(const uint8_t *buf, uint32_t size,
                  char *error_buf, uint32_t error_buf_size);
WASM_RUNTIME_API_EXTERN wasm_module_t
wasm_runtime_load_from_sections(wasm_section_list_t section_list, bool is_aot,
                                char *error_buf, uint32_t error_buf_size);
```
For wasm_runtime_load, as the input is a file buffer, currently we assume that the buffer can be referred after loading, e.g. wasm bytecode inside the file is used, and in aot code aot file is used when XIP is enabled. So for wasm_runtime_load, my personal understanding is that we can adjust the strings inside it in non XIP mode, and use its strings in XIP mode.
For wasm_runtime_load_from_sections, as the sections might be allocated individually and some sections might be destroyed after loading, we can not refer to the strings inside it after loading.

Both wasm_runtime_load and wasm_runtime_load_from_sections eventually call an internal static function load_from_sections() (in both aot and interpreter mode), so I think we can add an argument to function load_from_sections(), e.g. 'bool org_str_referable_to', then for wasm_runtime_load, set the argument to true, for wasm_runtime_load_from_sections, set the argument to false.  Note that similar changes can be applied to both aot and interpreter.

Add @xujuntwt95329, @lum1n0us, @qinxk-inter to discuss.

no1wudi:
@wenyongh Done

PR #819, opened 2021-11-07
[skip ci] speed up the spec test by parallelism

lum1n0us:
It will reduce `spec_test.yml` from ~14min to ~9min. It will reduce `ci_on_sgx` from ~18min to ~12min.


lum1n0us:
It will reduce ~4min if we can replace compiling WABT by using WABT binary release.

lum1n0us:
Update:
- *./test_wamr.sh* will keep the same experience likes before by default.
- *./test_wamr.sh -P* will enable parallelism mode
- in parallelism mode, *all.py* will be in a less-output mode. It only outputs the last words of failed cases and will not output logs for those passed

PR #820, opened 2021-11-08
Refine interp/aot string storage and emitting

wenyongh:
  same behavior as before


no1wudi:
LGTM

core/iwasm/aot/aot_loader.c
```diff
@@ -240,6 +247,64 @@ const_str_set_insert(const uint8 *str, int32 len, AOTModule *module,
...
+load_string(uint8 **p_buf, const uint8 *buf_end, AOTModule *module,
+            bool is_load_from_file_buf, char *error_buf, uint32 error_buf_size)
+{
+    uint8 *p = *p_buf;
+    const uint8 *p_end = buf_end;
+    char *str;
+    uint16 str_len;
+
+    CHECK_BUF(p, p_end, 1);
+    if (*p & 0x80) {
```
yamt, line 260: this looks problematic because:
* it checks a wrong byte if `p` is not uint16_t aligned. (missing `p = align_ptr(p, 2)`)
* it would break long strings (>= 0x80 in little endian)

yamt: @wenyongh
how do you think?

i can't think of any clean way to fix it except removing the optimization.

we probably can use alignment padding before and after the string to store extra data. but i'm not sure if it's worth doing.

wenyongh: @yamt Thanks for reporting the issue! Yes, seems that we had better remove the optimization of following code:
```C
        else if (is_load_from_file_buf) {
            /* As the file buffer can be referred to after loading,
               we use the 2 bytes of size to adjust the string:
               mark the flag with the highest bit of size[0],
               move string 1 byte backward and then append '\0' */
            *(p - 2) |= 0x80;
            bh_memmove_s(p - 1, (uint32)(str_len + 1), p, (uint32)str_len);
            p[str_len - 1] = '\0';
            str = (char *)(p - 1);
        }
```

wenyongh: @yamt There may be another solution: create another const string list in the module, after adjusting the string (move string 1 byte backward and append '\0', no need to mark the highest bit flag), add it into the const string list. And in the next time loading, check whether the `str = align_ptr(p, 2) + 1` is in the const string list, if yes, use it directly. The code may be like:
```C
    p = align_ptr(p, 2);
    CHECK_BUF(p, p_end, 1);
    if (find p+1 in str list) {
        return p + 1;
    else {
        /* The string hasn't been adjusted */
        read_uint16(p, p_end, str_len);
        CHECK_BUF(p, p_end, str_len);

        if (str_len == 0) {
            str = "";
        }
        else if (p[str_len - 1] == '\0') {
            /* The string is terminated with '\0', use it directly */
            str = (char *)p;
        }
        else if (is_load_from_file_buf) {
            bh_memmove_s(p - 1, (uint32)(str_len + 1), p, (uint32)str_len);
            p[str_len - 1] = '\0';
            str = (char *)(p - 1);
            add str to string list;
        }
        else {
            if (!(str = const_str_set_insert((uint8 *)p, str_len, module,
                                             error_buf, error_buf_size))) {
                goto fail;
            }
        }
        p += str_len;
    }
```
Is it a better way, what is your opinion?

yamt: > @yamt There may be another solution: create another const string list in the module, after adjusting the string (move string 1 byte backward and append '\0', no need to mark the highest bit flag), add it into the const string list. And in the next time loading, check whether the `str = align_ptr(p, 2) + 1` is in the const string list, if yes, use it directly. The code may be like:
> 
> ```c
>     p = align_ptr(p, 2);
>     CHECK_BUF(p, p_end, 1);
>     if (find p+1 in str list) {
>         return p + 1;
>     else {
>         /* The string hasn't been adjusted */
>         read_uint16(p, p_end, str_len);
>         CHECK_BUF(p, p_end, str_len);
> 
>         if (str_len == 0) {
>             str = "";
>         }
>         else if (p[str_len - 1] == '\0') {
>             /* The string is terminated with '\0', use it directly */
>             str = (char *)p;
>         }
>         else if (is_load_from_file_buf) {
>             bh_memmove_s(p - 1, (uint32)(str_len + 1), p, (uint32)str_len);
>             p[str_len - 1] = '\0';
>             str = (char *)(p - 1);
>             add str to string list;
>         }
>         else {
>             if (!(str = const_str_set_insert((uint8 *)p, str_len, module,
>                                              error_buf, error_buf_size))) {
>                 goto fail;
>             }
>         }
>         p += str_len;
>     }
> ```
> 
> Is it a better way, what is your opinion?

it sounds like a good trick.

yamt: https://github.com/bytecodealliance/wasm-micro-runtime/pull/1100

core/iwasm/aot/aot_loader.c
```diff
@@ -2062,13 +2138,7 @@ load_relocation_section(const uint8 *buf, const uint8 *buf_end,
...
 
         name_addr = symbol_buf + symbol_offsets[name_index];
-        str_len = *(uint16 *)name_addr;
-
-        if (!(group->section_name = const_str_set_insert(
-                  name_addr + sizeof(uint16), (int32)str_len, module, error_buf,
-                  error_buf_size))) {
-            goto fail;
-        }
+        read_string(name_addr, buf_end, group->section_name);
```
yamt, line 2141: doesn't the BH_PLATFORM_WINDOWS block above need similar changes to use read_string?

wenyongh: Should no need to change it, the block directly uses the raw string to compare with the string pointer and length, without caring whether it's terminated with '\0' or not, without using strlen.

PR #835, opened 2021-11-15
wamr: Add nuttx config for debug and  reduce flash size

DS-LK:
Applied on the nuttx platform, some memory configurations are added, and aot and interp can be completely closed or opened, thereby reducing the flash size。


no1wudi (approved):
LGTM

PR #838, opened 2021-11-16
Enable reference type and bulk memory in Wamrc

lum1n0us:
https://github.com/wasm-micro-runtime/wamr-internal-dev/issues/342

### DIFF
--- a/core/iwasm/interpreter/wasm_loader.c
+++ b/core/iwasm/interpreter/wasm_loader.c
@@ -6149,12 +6149,8 @@ get_table_seg_elem_type(const WASMModule *module, uint32 table_seg_idx,
                         uint32 error_buf_size)
 {
     if (table_seg_idx >= module->table_seg_count) {
-#if WASM_ENABLE_REF_TYPES != 0
         set_error_buf_v(error_buf, error_buf_size, "unknown elem segment %u",
                         table_seg_idx);
-#else
-        set_error_buf(error_buf, error_buf_size, "unknown table segment");
-#endif
         return false;
     }
 
--- a/core/iwasm/interpreter/wasm_loader.c
+++ b/core/iwasm/interpreter/wasm_loader.c
@@ -212,8 +212,7 @@ is_32bit_type(uint8 type)
 {
     if (type == VALUE_TYPE_I32 || type == VALUE_TYPE_F32
 #if WASM_ENABLE_REF_TYPES != 0
-        || (wasm_get_ref_types_flag()
-            && (type == VALUE_TYPE_FUNCREF || type == VALUE_TYPE_EXTERNREF))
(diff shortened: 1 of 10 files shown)


core/iwasm/interpreter/wasm_mini_loader.c
```diff
@@ -1234,13 +1217,11 @@ static bool
 check_table_index(const WASMModule *module, uint32 table_index, char *error_buf,
                   uint32 error_buf_size)
 {
-    if (
 #if WASM_ENABLE_REF_TYPES != 0
```
wenyongh, line 1220: Should be WASM_ENABLE_REF_TYPES == 0, like L1022

tests/wamr-test-suites/spec-test-script/runtest.py
```diff
@@ -939,12 +939,12 @@ def compile_wasm_to_aot(wasm_tempfile, aot_tempfile, runner, opts, r):
...
-        cmd.append("--enable-simd")
+    if not opts.simd:
+        cmd.append("--disable-simd")
 
-    if opts.ref_types:
-        cmd.append("--enable-ref-types")
-        cmd.append("--enable-bulk-memory")
+    if not opts.ref_types:
+        cmd.append("--disable-ref-types")
+        cmd.append("--disable-bulk-memory")
```
wenyongh, line 947: Should not disable bulk memory here, in other feature e.g. thread, it must be enabled.

core/iwasm/interpreter/wasm_mini_loader.c
wenyongh: When WASM_ENABLE_REF_TYPES != 0, the type_index can be larger than 0, it should not return false. Here we should change to WASM_ENABLE_REF_TYPES == 0?

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -6217,13 +6150,8 @@ get_table_seg_elem_type(const WASMModule *module, uint32 table_seg_idx,
...
 #if WASM_ENABLE_REF_TYPES != 0
-        if (!wasm_get_ref_types_flag()) {
-            set_error_buf(error_buf, error_buf_size, "unknown table segment");
-        }
-        else {
-            set_error_buf_v(error_buf, error_buf_size,
-                            "unknown elem segment %u", table_seg_idx);
-        }
+        set_error_buf_v(error_buf, error_buf_size, "unknown elem segment %u",
+                        table_seg_idx);
```
wenyongh, line 6154: How about merging L6153 to L6157 together, just replacing them with a line: 
```C
set_error_buf_v(error_buf, error_buf_size, "unknown elem segment %u",
                        table_seg_idx);
```

PR #841, opened 2021-11-20
Implement atomic and memset/memmove/memcpy intrinsic for riscv

no1wudi:
* Lower aotmic instruction to non-atomic form on some platform
* Implement memset/memmove/memcpy for XIP
* Disable rtti in cmake


core/iwasm/aot/aot_intrinsic.c
```diff
@@ -607,6 +640,20 @@ aot_intrinsic_fill_capability_flags(AOTCompContext *comp_ctx)
...
+                /* Target don't support hardware atomic instruction */
+                add_intrinsic_capability(comp_ctx,
+                                         AOT_INTRINSIC_FLAG_ATOMIC_LOAD);
+                add_intrinsic_capability(comp_ctx,
+                                         AOT_INTRINSIC_FLAG_ATOMIC_STORE);
+                add_intrinsic_capability(comp_ctx,
+                                         AOT_INTRINSIC_FLAG_CMPXCHG_4);
+                add_intrinsic_capability(comp_ctx,
+                                         AOT_INTRINSIC_FLAG_CMPXCHG_8);
+            }
```
wenyongh, line 655: Should dispose mac with LLVMDisposeMessage(), see llvm document:
```C
/** Returns the feature string used creating this target machine. See
  llvm::TargetMachine::getFeatureString. The result needs to be disposed with
  LLVMDisposeMessage. */
char *LLVMGetTargetMachineFeatureString(LLVMTargetMachineRef T);
```

core/iwasm/compilation/aot_emit_memory.c
```diff
@@ -951,13 +962,51 @@ aot_compile_op_memory_copy(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx)
...
+            aot_set_last_error("create LLVM function type failed.");
+            return false;
+        }
+
+        if (!(func_ptr_type = LLVMPointerType(func_type, 0))) {
+            aot_set_last_error("create LLVM function pointer type failed.");
+            return false;
+        }
+
+        int32 func_idx = aot_get_native_symbol_index(comp_ctx, "memmove");
```
wenyongh, line 984: Have better put variable definition after L967

core/iwasm/compilation/aot_emit_memory.c
```diff
@@ -981,11 +1030,49 @@ aot_compile_op_memory_fill(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx)
...
+            aot_set_last_error("create LLVM function type failed.");
+            return false;
+        }
+
+        if (!(func_ptr_type = LLVMPointerType(func_type, 0))) {
+            aot_set_last_error("create LLVM function pointer type failed.");
+            return false;
+        }
+
+        int32 func_idx = aot_get_native_symbol_index(comp_ctx, "memset");
```
wenyongh, line 1052: Same as above

core/iwasm/compilation/aot_emit_memory.c
```diff
@@ -1084,59 +1171,127 @@ aot_compile_op_atomic_cmpxchg(AOTCompContext *comp_ctx,
...
+            GET_AOT_FUNCTION(aot_intrinsic_cmpxchg_4, 5);
+        }
+
+        param_values[0] = maddr;
+        param_values[1] = expect;
+        param_values[2] = value;
+        param_values[3] =
+            I32_CONST((int)LLVMAtomicOrderingSequentiallyConsistent);
+        param_values[4] =
+            I32_CONST((int)LLVMAtomicOrderingSequentiallyConsistent);
```
wenyongh, line 1220: Should check return value of I32_CONST, theoretically it can be NULL

core/iwasm/aot/aot_intrinsic.c
```diff
@@ -474,6 +481,32 @@ aot_intrinsic_f64_cmp(AOTFloatCond cond, float64 lhs, float64 rhs)
...
 }
 
+int32
+aot_intrinsic_cmpxchg_4(int32 *ptr, int32 expected, int32 desired, int success,
+                        int failure)
+{
+    int32 ret = *ptr;
+#if !defined(BUILD_TARGET_X86_64) && !defined(BUILD_TARGET_X86_32) \
+    && !defined(BUILD_TARGET_AMD64)
+    __atomic_compare_exchange_4(ptr, &expected, desired, false, success,
```
wenyongh, line 491: Can we make sure that __atomic_compare_exchange_4 is supported by most compilers? Not sure whether it might cause link error.

no1wudi: How about only enable it on tested platform and raise a runtime log on others ?

wenyongh: Not known why we need to register aot_intrinsic_cmpxchg_4/8 global symbol map in aot_reloc.h? It seems that only riscv needs these APIs in multi-thread mode, shall we move them into aot_reloc_riscv.c?

no1wudi: Use LLVM builtin pass to convert these instructions

core/iwasm/compilation/aot_llvm_extra.cpp
```diff
@@ -106,6 +118,109 @@ WAMRCreateMCJITCompilerForModule(LLVMExecutionEngineRef *OutJIT,
...
+        }
+    }
+
+    return Changed;
+}
+
+void
+aot_add_expand_memory_op_pass(LLVMPassManagerRef pass)
+{
+    unwrap(pass)->add(new ExpandMemoryOpPass());
```
wenyongh, line 221: Should check return value of `new ExpandMemoryOpPass()`, and change return value of aot_add_expand_memory_op_pass from void to bool?

core/iwasm/compilation/aot_compiler.c
```diff
@@ -2625,6 +2658,27 @@ aot_compile_wasm(AOTCompContext *comp_ctx)
...
 
+    if (comp_ctx->optimize && comp_ctx->is_indirect_mode) {
+        LLVMPassManagerRef common_pass_mgr = NULL;
+
+        if (!(common_pass_mgr = LLVMCreatePassManager())) {
+            aot_set_last_error("create pass manager failed");
+            return false;
+        }
+
+        aot_add_expand_memory_op_pass(common_pass_mgr);
```
wenyongh, line 2669: Should check return value

core/iwasm/compilation/aot_llvm_extra.cpp
```diff
@@ -106,6 +118,109 @@ WAMRCreateMCJITCompilerForModule(LLVMExecutionEngineRef *OutJIT,
...
+        return "Expand memory operation intrinsics";
+    }
+
+    void getAnalysisUsage(AnalysisUsage &AU) const override
+    {
+        AU.addRequired<TargetTransformInfoWrapperPass>();
+    }
+};
+
+char ExpandMemoryOpPass::ID = 0;
```
wenyongh, line 144: Can we change the variable to static?

no1wudi: It's already static since it declared in class ExpandMemoryOpPass

wenyongh: Got it, thanks.

core/iwasm/compilation/aot_llvm_extra.cpp
no1wudi: If new operation failed, wamrc will crash and report an exception to notify user.
I think it's enough for debug purpose.

PR #844, opened 2021-11-22
Fix the bug that destroy_callback_list is not cleared when thread_man…

DS-LK:
The destroy_callback_list variable in thread_manager_destroy is not cleared. When wamr is started as a task, after exiting and restarting, it will loop endlessly in traverse_list

### DIFF
--- a/core/iwasm/libraries/thread-mgr/thread_manager.c
+++ b/core/iwasm/libraries/thread-mgr/thread_manager.c
@@ -240,6 +240,7 @@ void
 wasm_cluster_cancel_all_callbacks()
 {
     traverse_list(destroy_callback_list, free_node_visitor, NULL);
+    bh_list_init(destroy_callback_list);
 }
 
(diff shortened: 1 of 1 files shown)


core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -240,6 +240,7 @@ void
 wasm_cluster_cancel_all_callbacks()
 {
     traverse_list(destroy_callback_list, free_node_visitor, NULL);
+    memset(destroy_callback_list, 0, sizeof(*destroy_callback_list));
```
xujuntwt95329, line 243: Had better use `bh_list_init` API as we can't assume the list implementation from this file

```suggestion
    bh_list_init(destroy_callback_list);
```

DS-LK: OK, thank you.

PR #847, opened 2021-11-24
fix duplicate debug instance id in multiple native thread issue

xujuntwt95329:
The debug_instance_list is not protected by mutex, and the debug instance use
the length of this list as its id. If several threads create debug instance at
the same time, they may get duplicate ids. This patch add mutex to protect this
list

### DIFF
--- a/core/iwasm/libraries/debug-engine/debug_engine.c
+++ b/core/iwasm/libraries/debug-engine/debug_engine.c
@@ -21,10 +21,27 @@ typedef struct WASMDebugEngine {
     int platform_port;
     int process_base_port;
     bh_list debug_instance_list;
+    korp_mutex instance_list_lock;
     bool active;
 } WASMDebugEngine;
 
 static WASMDebugEngine *g_debug_engine;
+
+static uint32 current_instance_id = 1;
+
+static uint32
+allocate_instance_id()
+{
+    uint32 id;
+
+    bh_assert(g_debug_engine);
+
+    os_mutex_lock(&g_debug_engine->instance_list_lock);
+    id = current_instance_id++;
+    os_mutex_unlock(&g_debug_engine->instance_list_lock);
+
+    return id;
+}
 
 static bool
 should_stop(WASMDebugControlThread *control_thread)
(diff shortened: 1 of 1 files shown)


core/iwasm/libraries/debug-engine/debug_engine.c
```diff
@@ -21,11 +21,28 @@ typedef struct WASMDebugEngine {
...
+    korp_mutex instance_list_lock;
     bool active;
 } WASMDebugEngine;
 
 static WASMDebugEngine *g_debug_engine;
 
+static uint32 current_instance_id = 1;
+
+static uint32
+allocate_instance_id()
```
lum1n0us, line 33: how about [GCC atomic builtins](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html)?

xujuntwt95329: I‘ve also considered this, but I concern about the compiler compatibility. The GCC atomic builtins are optional features so there may be some compiler who doesn't implement it.

Also, in my opinion, these codes are not on the critical path, the mutex lock should not raise performance issues

@wenyongh how do you think about this?

wenyongh: Should not use the optional function which might lead to link error in some platform or compiler. Normally here we should use the API provided by platform layer.

core/iwasm/libraries/debug-engine/debug_engine.c
```diff
@@ -177,6 +197,14 @@ wasm_debug_engine_create()
     }
     memset(engine, 0, sizeof(WASMDebugEngine));
 
+    if (os_mutex_init(&engine->instance_list_lock) != 0) {
+        LOG_ERROR("WASM Debug Engine error: failed to init mutex");
```
wenyongh, line 201: Should free engine before return NULL

xujuntwt95329: Done

PR #851, opened 2021-11-25
Reorganize externref related public APIs.

lum1n0us:
Basically, users don't need to use `wasm_externref_obj2ref` before
calling `wasm_runtime_call_wasm_XXX`. User can pass object pointers
directlly as externref parameters.

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -2235,9 +2235,9 @@ aot_invoke_native(WASMExecEnv *exec_env, uint32 func_idx, uint32 argc,
     }
     else {
         signature = import_func->signature;
-        return wasm_runtime_invoke_native_raw(exec_env, func_ptr, func_type,
-                                              signature, attachment, argv, argc,
-                                              argv);
+        return wasm_runtime_invoke_native_raw(exec_env, func_ptr, false,
+                                              func_type, signature, attachment,
+                                              argv, argc, argv);
     }
 }
 
@@ -2301,9 +2301,9 @@ aot_call_indirect(WASMExecEnv *exec_env, uint32 tbl_idx, uint32 table_elem_idx,
         signature = import_func->signature;
         if (import_func->call_conv_raw) {
             attachment = import_func->attachment;
-            return wasm_runtime_invoke_native_raw(exec_env, func_ptr, func_type,
-                                                  signature, attachment, argv,
-                                                  argc, argv);
+            return wasm_runtime_invoke_native_raw(exec_env, func_ptr, true,
+                                                  func_type, signature,
+                                                  attachment, argv, argc, argv);
         }
     }
 
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -2301,7 +2301,7 @@ aot_call_indirect(WASMExecEnv *exec_env, uint32 tbl_idx, uint32 table_elem_idx,
         signature = import_func->signature;
         if (import_func->call_conv_raw) {
             attachment = import_func->attachment;
-            return wasm_runtime_invoke_native_raw(exec_env, func_ptr, true,
+            return wasm_runtime_invoke_native_raw(exec_env, func_ptr, false,
                                                   func_type, signature,
                                                   attachment, argv, argc, argv);
         }
@@ -2389,7 +2389,7 @@ aot_call_indirect(WASMExecEnv *exec_env, uint32 tbl_idx, uint32 table_elem_idx,
     }
     else {
         ret = invoke_native_internal(
-            exec_env, func_ptr, func_idx < aot_module->import_func_count,
+            exec_env, func_ptr, func_idx >= aot_module->import_func_count,
             func_type, signature, attachment, argv, argc, argv);
         if (clear_wasi_proc_exit_exception(module_inst))
             return true;
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -85,7 +85,7 @@ init_global_data(uint8 *global_data, uint8 type, WASMValue *initial_value)
     switch (type) {
         case VALUE_TYPE_I32:
         case VALUE_TYPE_F32:
-#if WASM_ENABLE_REF_TYPES
+#if WASM_ENABLE_REF_TYPES != 0
         case VALUE_TYPE_FUNCREF:
         case VALUE_TYPE_EXTERNREF:
 #endif
@@ -1294,10 +1294,9 @@ aot_signal_destroy()
 
 static bool
 invoke_native_with_hw_bound_check(WASMExecEnv *exec_env, void *func_ptr,
-                                  const WASMType *func_type,
+                                  bool is_aot_func, const WASMType *func_type,
                                   const char *signature, void *attachment,
-                                  bool aot_call_flag, uint32 *argv, uint32 argc,
-                                  uint32 *argv_ret)
+                                  uint32 *argv, uint32 argc, uint32 *argv_ret)
 {
     AOTModuleInstance *module_inst = (AOTModuleInstance *)exec_env->module_inst;
     WASMExecEnv **p_aot_exec_env = &aot_exec_env;
@@ -1353,14 +1352,14 @@ invoke_native_with_hw_bound_check(WASMExecEnv *exec_env, void *func_ptr,
             }
             else {
                 ret = wasm_runtime_invoke_native(
-                    exec_env, func_ptr, func_type, signature, attachment,
-                    aot_call_flag, argv, argc, argv_ret);
+                    exec_env, func_ptr, is_aot_func, func_type, signature,
+                    attachment, argv, argc, argv_ret);
             }
         }
         else {
-            ret = wasm_runtime_invoke_native(
-                exec_env, func_ptr, func_type, signature, attachment,
-                aot_call_flag, argv, argc, argv_ret);
+            ret = wasm_runtime_invoke_native(exec_env, func_ptr, is_aot_func,
+                                             func_type, signature, attachment,
+                                             argv, argc, argv_ret);
         }
 #ifdef BH_PLATFORM_WINDOWS
         if ((exce = aot_get_exception(module_inst))
@@ -1450,9 +1449,8 @@ aot_call_function(WASMExecEnv *exec_env, AOTFunctionInstance *function,
         }
 #endif
 
-        ret = invoke_native_internal(exec_env, function->u.func.func_ptr,
-                                     func_type, NULL, NULL, true, argv1, argc,
-                                     argv);
+        ret = invoke_native_internal(exec_env, function->u.func.func_ptr, true,
+                                     func_type, NULL, NULL, argv1, argc, argv);
 
         if (!ret || aot_get_exception(module_inst)) {
             if (argv1 != argv1_buf)
@@ -1515,9 +1513,8 @@ aot_call_function(WASMExecEnv *exec_env, AOTFunctionInstance *function,
         }
 #endif
 
-        ret = invoke_native_internal(exec_env, function->u.func.func_ptr,
-                                     func_type, NULL, NULL, true, argv, argc,
-                                     argv);
+        ret = invoke_native_internal(exec_env, function->u.func.func_ptr, true,
+                                     func_type, NULL, NULL, argv, argc, argv);
 
         if (clear_wasi_proc_exit_exception(module_inst))
             ret = true;
@@ -2232,9 +2229,9 @@ aot_invoke_native(WASMExecEnv *exec_env, uint32 func_idx, uint32 argc,
     }
     else if (!import_func->call_conv_raw) {
         signature = import_func->signature;
-        return wasm_runtime_invoke_native(exec_env, func_ptr, func_type,
-                                          signature, attachment, false, argv,
-                                          argc, argv);
+        return wasm_runtime_invoke_native(exec_env, func_ptr, false, func_type,
+                                          signature, attachment, argv, argc,
+                                          argv);
     }
     else {
         signature = import_func->signature;
@@ -2347,8 +2344,8 @@ aot_call_indirect(WASMExecEnv *exec_env, uint32 tbl_idx, uint32 table_elem_idx,
             cell_num += wasm_value_type_cell_num(ext_ret_types[i]);
         }
 
-        ret = invoke_native_internal(exec_env, func_ptr, func_type, signature,
-                                     attachment, true, argv1, argc, argv);
+        ret = invoke_native_internal(exec_env, func_ptr, true, func_type,
+                                     signature, attachment, argv1, argc, argv);
         if (!ret || aot_get_exception(module_inst)) {
             if (argv1 != argv1_buf)
                 wasm_runtime_free(argv1);
@@ -2391,8 +2388,9 @@ aot_call_indirect(WASMExecEnv *exec_env, uint32 tbl_idx, uint32 table_elem_idx,
         return true;
     }
     else {
-        ret = invoke_native_internal(exec_env, func_ptr, func_type, signature,
-                                     attachment, true, argv, argc, argv);
+        ret = invoke_native_internal(
+            exec_env, func_ptr, func_idx < aot_module->import_func_count,
+            func_type, signature, attachment, argv, argc, argv);
         if (clear_wasi_proc_exit_exception(module_inst))
             return true;
         return ret;
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -1296,7 +1296,8 @@ static bool
 invoke_native_with_hw_bound_check(WASMExecEnv *exec_env, void *func_ptr,
                                   const WASMType *func_type,
                                   const char *signature, void *attachment,
-                                  uint32 *argv, uint32 argc, uint32 *argv_ret)
+                                  bool aot_call_flag, uint32 *argv, uint32 argc,
+                                  uint32 *argv_ret)
 {
     AOTModuleInstance *module_inst = (AOTModuleInstance *)exec_env->module_inst;
     WASMExecEnv **p_aot_exec_env = &aot_exec_env;
@@ -1351,15 +1352,15 @@ invoke_native_with_hw_bound_check(WASMExecEnv *exec_env, void *func_ptr,
                 ret = aot_get_exception(module_inst) ? false : true;
             }
             else {
-                ret = wasm_runtime_invoke_native(exec_env, func_ptr, func_type,
-                                                 signature, attachment, argv,
-                                                 argc, argv_ret);
+                ret = wasm_runtime_invoke_native(
+                    exec_env, func_ptr, func_type, signature, attachment,
+                    aot_call_flag, argv, argc, argv_ret);
             }
         }
         else {
-            ret = wasm_runtime_invoke_native(exec_env, func_ptr, func_type,
-                                             signature, attachment, argv, argc,
-                                             argv_ret);
+            ret = wasm_runtime_invoke_native(
+                exec_env, func_ptr, func_type, signature, attachment,
+                aot_call_flag, argv, argc, argv_ret);
         }
 #ifdef BH_PLATFORM_WINDOWS
         if ((exce = aot_get_exception(module_inst))
@@ -1450,7 +1451,8 @@ aot_call_function(WASMExecEnv *exec_env, AOTFunctionInstance *function,
 #endif
 
         ret = invoke_native_internal(exec_env, function->u.func.func_ptr,
-                                     func_type, NULL, NULL, argv1, argc, argv);
+                                     func_type, NULL, NULL, true, argv1, argc,
+                                     argv);
 
         if (!ret || aot_get_exception(module_inst)) {
             if (argv1 != argv1_buf)
@@ -1514,7 +1516,8 @@ aot_call_function(WASMExecEnv *exec_env, AOTFunctionInstance *function,
 #endif
 
         ret = invoke_native_internal(exec_env, function->u.func.func_ptr,
-                                     func_type, NULL, NULL, argv, argc, argv);
+                                     func_type, NULL, NULL, true, argv, argc,
+                                     argv);
 
         if (clear_wasi_proc_exit_exception(module_inst))
             ret = true;
@@ -1557,15 +1560,7 @@ aot_create_exec_env_and_call_function(AOTModuleInstance *module_inst,
         }
     }
 
-#if WASM_ENABLE_REF_TYPES != 0
-    wasm_runtime_prepare_call_function(exec_env, func);
-#endif
-
     ret = aot_call_function(exec_env, func, argc, argv);
-
-#if WASM_ENABLE_REF_TYPES != 0
-    wasm_runtime_finalize_call_function(exec_env, func, ret, argv);
-#endif
 
     /* don't destroy the exec_env if it's searched from the cluster */
     if (!existing_exec_env)
@@ -2238,8 +2233,8 @@ aot_invoke_native(WASMExecEnv *exec_env, uint32 func_idx, uint32 argc,
     else if (!import_func->call_conv_raw) {
         signature = import_func->signature;
         return wasm_runtime_invoke_native(exec_env, func_ptr, func_type,
-                                          signature, attachment, argv, argc,
-                                          argv);
+                                          signature, attachment, false, argv,
+                                          argc, argv);
     }
     else {
         signature = import_func->signature;
@@ -2353,7 +2348,7 @@ aot_call_indirect(WASMExecEnv *exec_env, uint32 tbl_idx, uint32 table_elem_idx,
         }
 
         ret = invoke_native_internal(exec_env, func_ptr, func_type, signature,
-                                     attachment, argv1, argc, argv);
+                                     attachment, true, argv1, argc, argv);
         if (!ret || aot_get_exception(module_inst)) {
             if (argv1 != argv1_buf)
                 wasm_runtime_free(argv1);
@@ -2397,7 +2392,7 @@ aot_call_indirect(WASMExecEnv *exec_env, uint32 tbl_idx, uint32 table_elem_idx,
     }
     else {
         ret = invoke_native_internal(exec_env, func_ptr, func_type, signature,
-                                     attachment, argv, argc, argv);
+                                     attachment, true, argv, argc, argv);
         if (clear_wasi_proc_exit_exception(module_inst))
             return true;
         return ret;
--- a/core/iwasm/common/wasm_application.c
+++ b/core/iwasm/common/wasm_application.c
@@ -656,7 +656,7 @@ wasm_application_execute_func(WASMModuleInstanceCommon *module_inst,
 #if WASM_ENABLE_REF_TYPES != 0
             case VALUE_TYPE_FUNCREF:
             {
-                if (argv1[k] != NULL_REF && argv1[k] != (uint32)-1)
+                if (argv1[k] != NULL_REF)
                     os_printf("%u:ref.func", argv1[k]);
                 else
                     os_printf("func:ref.null");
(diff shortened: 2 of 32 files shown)


core/iwasm/common/wasm_application.c
```diff
@@ -354,13 +364,46 @@ union ieee754_double {
     } ieee;
 };
 
+#if WASM_ENABLE_REF_TYPES != 0
+static int32
+type_size_in_double_world(uint8 type)
```
xujuntwt95329, line 369: maybe `word` in this semantic

BTW, I wonder if there is an absolute definition of `WORD`? 
If I remembered correctly, the bit width of `WORD` may varies by platforms, it should represents `the most convenient block of data for the computer to deal with`

core/iwasm/common/wasm_application.c
```diff
@@ -404,8 +447,22 @@ wasm_application_execute_func(WASMModuleInstanceCommon *module_inst,
...
+            type_size_in_double_world(type->types[type->param_count + i]);
+    }
+    argc1 = param_size_in_double_world;
+    cell_num = (param_size_in_double_world >= result_size_in_double_world)
+                   ? param_size_in_double_world
+                   : result_size_in_double_world;
+#else
     argc1 = type->param_cell_num;
     cell_num = (argc1 > type->ret_cell_num) ? argc1 : type->ret_cell_num;
+#endif
```
wenyongh, line 465: Should not re-calculate the param cell num here, we can use the type->param_cell_num when wasm_value_type_cell_num() is corretly modified.

core/iwasm/common/wasm_application.c
```diff
@@ -529,23 +585,27 @@ wasm_application_execute_func(WASMModuleInstanceCommon *module_inst,
             }
             case VALUE_TYPE_EXTERNREF:
             {
-                if (strncmp(argv[i], "null", 4) == 0
-                    || strncmp(argv[i], "NULL", 4) == 0) {
+#ifdef BUILD_TARGET_X86_32
```
wenyongh, line 588: Don't use X86_32 macro, use `#if UINTPTR_MAX == UINT32_MAX` instead

core/iwasm/common/wasm_application.c
```diff
@@ -560,11 +620,14 @@ wasm_application_execute_func(WASMModuleInstanceCommon *module_inst,
             goto fail;
         }
     }
-    bh_assert(p == (int32)argc1);
```
wenyongh, line 563: Don't remove the code that isn't related to this patch?

core/iwasm/common/wasm_application.c
```diff
@@ -628,15 +691,23 @@ wasm_application_execute_func(WASMModuleInstanceCommon *module_inst,
...
             case VALUE_TYPE_EXTERNREF:
             {
-                if (argv1[k] != NULL_REF) {
-                    void *extern_obj = NULL;
-                    bool ret = wasm_externref_ref2obj(argv1[k], &extern_obj);
-                    bh_assert(ret);
-                    (void)ret;
-                    os_printf("%p:ref.extern", extern_obj);
+                if (argv1[k] != 0 && NULL_REF != argv1[k]) {
+#ifdef BUILD_TARGET_X86_32
```
wenyongh, line 695: `#if UINTPTR_MAX == UINT32_MAX`

core/iwasm/common/wasm_application.c
```diff
@@ -628,15 +691,23 @@ wasm_application_execute_func(WASMModuleInstanceCommon *module_inst,
             }
             case VALUE_TYPE_EXTERNREF:
             {
-                if (argv1[k] != NULL_REF) {
-                    void *extern_obj = NULL;
-                    bool ret = wasm_externref_ref2obj(argv1[k], &extern_obj);
-                    bh_assert(ret);
-                    (void)ret;
-                    os_printf("%p:ref.extern", extern_obj);
+                if (argv1[k] != 0 && NULL_REF != argv1[k]) {
```
wenyongh, line 694: argv1[k] is 32-bit, in 64-bit, we should check argv1[k], argv1[k+1]:
```C
#if UINTPTR_MAX == UINT32_MAX
                  if (argv1[k] != 0 && NULL_REF != argv1[k])
                       os_printf("%p:ref.extern", argv1[k]);
                  else
                       os_printf("extern:ref.null");
#else
                   union {
                        uintptr_t val;
                        uint32 parts[2];
                    } u;
                    u.parts[0] = argv1[k++];
                    u.parts[1] = argv1[k++];
                    if (u.val != 0 && NULL_REF != u.val)
                         os_printf("%p:ref.extern", (void *)u.val);
                    else
                         os_printf("extern:ref.null");
#endif
```

core/iwasm/common/wasm_native.c
```diff
@@ -78,7 +78,7 @@ check_symbol_signature(const WASMType *type, const char *signature)
              && type->types[i] <= VALUE_TYPE_I32
              && sig == sig_map[type->types[i] - VALUE_TYPE_F64])
 #if WASM_ENABLE_REF_TYPES != 0
-            || (sig == 'i' && type->types[i] == VALUE_TYPE_EXTERNREF)
+            || (sig == 'I' && type->types[i] == VALUE_TYPE_EXTERNREF)
```
wenyongh, line 81: Seems that both 'i' and 'I' are not good, how about changing to 'r' to represent a ref (also can represent other refs for GC feature), and update the document export_native_api.md?

wenyongh (commented):
partly reviewed

core/iwasm/common/wasm_application.c
lum1n0us: The size of an externref is different between inside of the call_WASM and outside of it. It is an uintptr_t outside and an uint32 inside. So existed cell_num variables are not correct here. Actually, this is the key point of the PR and made lots of things harder.

wenyongh: OK, how about rename the function to type_cell_num_out_of_runtime, type_cell_num_of_native, or type_cell_num_of_xxx? And had better call wasm_value_type_cell_num for other types:
```C
static uint32
type_cell_num_out_of_runtime(uint8 type)
{
    if (type == VALUE_TYPE_EXTERNREF)
        return sizeof(uintptr_t) / sizeof(uint32);
    else
        return wasm_value_type_cell_num(type);
}

File: core/iwasm/common/wasm_runtime_common.c [hunk 8]
```diff
@@ -1113,46 +1113,159 @@ wasm_runtime_lookup_function(WASMModuleInstanceCommon *const module_inst,
...
+    }
+    else {
+        for (param_i = 0; param_i < func_type->param_count && argv_i < argc
+                          && new_argv_i < func_type->param_cell_num;
+             param_i++) {
+            uint8 param_type = func_type->types[param_i];
+            if (VALUE_TYPE_EXTERNREF == param_type) {
+                void *externref_obj;
+                uint32 externref_index;
+#ifdef BUILD_TARGET_X86_32
```
wenyongh, line 1176: #if UINTPTR_MAX == UINT32_MAX

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1162,41 +1275,56 @@ wasm_runtime_call_wasm(WASMExecEnv *exec_env,
...
 #endif
 #if WASM_ENABLE_AOT != 0
     if (exec_env->module_inst->module_type == Wasm_Module_AoT)
-        ret = aot_call_function(exec_env, (AOTFunctionInstance *)function, argc,
-                                argv);
+        ret = aot_call_function(exec_env, (AOTFunctionInstance *)function,
+                                param_argc, new_argv);
 #endif
+    if (!ret) {
+        return false;
```
wenyongh, line 1308: new_argv might be allocated in wasm_runtime_prepare_call_function but isn't freed here, seems here we should:
```C
if (new_argv != argv)
    wasm_runtime_free(new_argv);
```

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1162,41 +1275,56 @@ wasm_runtime_call_wasm(WASMExecEnv *exec_env,
...
+    }
 
 #if WASM_ENABLE_REF_TYPES != 0
-    wasm_runtime_finalize_call_function(exec_env, function, ret, argv);
+    if (!wasm_runtime_finalize_call_function(exec_env, function, new_argv,
+                                             result_argc, argv)) {
+        LOG_ERROR("Failed to process result");
+        return false;
```
wenyongh, line 1315: Same as above, when wasm_runtime_finalize_call_function returns false, it doesn't free new_argv.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1233,16 +1361,38 @@ parse_args_to_uint32_array(WASMType *type, uint32 num_args, wasm_val_t *args,
...
             }
+#if WASM_ENABLE_REF_TYPES != 0
+            case WASM_FUNCREF:
+            {
+                out_argv[p++] = args[i].of.i32;
+                break;
+            }
+            case WASM_ANYREF:
+            {
+#ifdef BUILD_TARGET_X86_32
```
wenyongh, line 1372: #if UINTPTR_MAX == UINT32_MAX

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1288,22 +1438,77 @@ parse_uint32_array_to_results(WASMType *type, uint32 argc, uint32 *argv,
...
                 break;
         }
     }
-    bh_assert(argc == p);
-    return type->result_count;
 }
 
+#if WASM_ENABLE_REF_TYPES != 0
+static int32
+type_size_in_double_world(uint8 type)
```
wenyongh, line 1475: Had better change to be like:
```C
static uint32
type_cell_num_out_of_runtime(uint8 type)
{
    if (type == VALUE_TYPE_EXTERNREF)
        return sizeof(uintptr_t) / sizeof(uint32);
    else
        return wasm_value_type_cell_num(type);
}
```

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -2537,8 +2753,8 @@ static VoidFuncPtr invokeNative_Void = (VoidFuncPtr)(uintptr_t)invokeNative;
 bool
 wasm_runtime_invoke_native(WASMExecEnv *exec_env, void *func_ptr,
                            const WASMType *func_type, const char *signature,
-                           void *attachment, uint32 *argv, uint32 argc,
-                           uint32 *argv_ret)
+                           void *attachment, bool aot_call_flag, uint32 *argv,
+                           uint32 argc, uint32 *argv_ret)
```
wenyongh, line 2757: Seems not convert externref index into object pointer in this function (ARM_VFP, THUM_VFP and RISCV32)?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -2964,8 +3180,8 @@ word_copy(uint32 *dest, uint32 *src, unsigned num)
...
```
wenyongh, line 3184: Same as above.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3362,6 +3599,24 @@ wasm_runtime_invoke_native(WASMExecEnv *exec_env, void *func_ptr,
                 PUT_F64_TO_ADDR(
                     argv_ret, invokeNative_Float64(func_ptr, argv1, n_stacks));
                 break;
+#if WASM_ENABLE_REF_TYPES != 0
+            case VALUE_TYPE_EXTERNREF:
+            {
+                int32 ret = invokeNative_Int64(func_ptr, argv1, n_stacks);
```
wenyongh, line 3605: Seems should distinguish AOT func or native func, if AOT, the return value is externref index:
```C
int32 ret = invokeNative_Int32(...);
argv_ret[0] = ret;
```
if native func, the return value should be the object pointer:
```C
int64 ret = invokeNative_Int64(...);    => but not int32 ret
convert ret to externref_idx;
argv[0] = externref_idx;
```

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3616,6 +3871,11 @@ wasm_externref_obj2ref(WASMModuleInstanceCommon *module_inst, void *extern_obj,
     ExternRefMapNode *node;
     uint32 externref_idx;
 
+    if (NULL_REF == (uintptr_t)extern_obj) {
```
wenyongh, line 3874: Seems we should redefine the NULL_REF for 32-bit and 64-bit:
```C
#if UINTPTR_MAX == UINT32_MAX
#define NULL_REF (0xFFFFFFFF)
#else
#define NULL_REF (0xFFFFFFFFFFFFFFFFLL)
#endif
```

core/iwasm/common/wasm_runtime_common.c
lum1n0us: `wasm_runtime_finalize_call_function ` will not return false. It always frees *new_argv*. though we can bypass it here.

core/iwasm/common/wasm_runtime_common.c
lum1n0us: Yes. Plan to discuss the return value problem, but seems it has been already mentioned below.

core/iwasm/common/wasm_runtime_common.c
lum1n0us: `NULL_REF` is used inside. It should always be the same size with a reference index length. 32bit for NON-GC building.

core/iwasm/common/wasm_native.c
lum1n0us: If we are going to treat it (refs) as a pointer, `*` should be OK, and we can run an address check.
Using `r` also means we won't do an address check like a pointer of an object.  Is it OK ?

lum1n0us:
will update documents when the code is stable.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -2536,9 +2730,9 @@ static VoidFuncPtr invokeNative_Void = (VoidFuncPtr)(uintptr_t)invokeNative;
 
 bool
 wasm_runtime_invoke_native(WASMExecEnv *exec_env, void *func_ptr,
-                           const WASMType *func_type, const char *signature,
-                           void *attachment, uint32 *argv, uint32 argc,
-                           uint32 *argv_ret)
+                           bool is_aot_func, const WASMType *func_type,
+                           const char *signature, void *attachment,
+                           uint32 *argv, uint32 argc, uint32 *argv_ret)
```
wenyongh, line 2735: Should also handle EXTERNREF with wasm_externref_ref2obj in this function? see L2942

wenyongh (commented):
Partly reviewed, more detailed checks to be done for wasm_runtime_common.c

core/iwasm/common/wasm_runtime_common.c
lum1n0us: It seems there is no `funcref` and `externref` support when `defined(BUILD_TARGET_ARM_VFP) || defined(BUILD_TARGET_THUMB_VFP)`. Shall we enable them ?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1406,6 +1600,20 @@ wasm_runtime_call_wasm_v(WASMExecEnv *exec_env,
...
+#if WASM_ENABLE_REF_TYPES != 0
+            case VALUE_TYPE_FUNCREF:
+            {
+                args[i].kind = WASM_FUNCREF;
+                args[i].of.i32 = va_arg(vargs, uint32);
+                break;
+            }
+            case VALUE_TYPE_EXTERNREF:
+            {
+                args[i].kind = WASM_ANYREF;
```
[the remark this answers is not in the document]
lum1n0us: in wasm_c_api, prefer to use `anyref` to represent `externref`

core/iwasm/common/wasm_runtime_common.c
wenyongh: Yes, normally we should support it in all archs.

wenyongh (commented):
Several possible memory leaks in wasm_runtime_common.c aren't not fixed?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3121,12 +3375,31 @@ wasm_runtime_invoke_native(WASMExecEnv *exec_env, void *func_ptr,
...
-            case VALUE_TYPE_FUNCREF:
             case VALUE_TYPE_EXTERNREF:
-#endif
-                argv1[j++] = *argv++;
+                if (is_aot_func) {
+                    argv1[j++] = *argv++;
+                }
+                else {
+                    uint32 externref_idx = *argv++;
+                    if (is_aot_func) {
```
wenyongh, line 3387: No need to check is_aot_func again, L3382 has checked

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3121,12 +3375,31 @@ wasm_runtime_invoke_native(WASMExecEnv *exec_env, void *func_ptr,
...
+                    else {
+                        void *externref_obj;
+
+                        if (!wasm_externref_ref2obj(externref_idx,
+                                                    &externref_obj))
+                            goto fail;
+
+                        argv1[j++] = (uintptr_t)externref_obj;
+                    }
+                    break;
```
wenyongh, line 3399: No need to break, L3401 has a break

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3378,13 +3672,29 @@ wasm_runtime_invoke_native(WASMExecEnv *exec_env, void *func_ptr,
...
-                    stacks[n_stacks++] = *argv_src++;
-                break;
+            {
+                uint32 externref_idx = *argv_src++;
+                if (is_aot_func) {
+                    if (n_ints < MAX_REG_INTS)
+                        ints[n_ints++] = externref_idx;
+                    else
+                        stacks[n_stacks++] = externref_idx;
+                    break;
```
wenyongh, line 3683: Had better not break here, but break after L3696

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3378,13 +3672,29 @@ wasm_runtime_invoke_native(WASMExecEnv *exec_env, void *func_ptr,
...
+                    void *externref_obj;
+
+                    if (!wasm_externref_ref2obj(externref_idx, &externref_obj))
+                        goto fail;
+
+                    if (n_ints < MAX_REG_INTS)
+                        ints[n_ints++] = (uintptr_t)externref_obj;
+                    else
+                        stacks[n_stacks++] = (uintptr_t)externref_obj;
+                    break;
```
wenyongh, line 3695: Same as above

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3440,6 +3749,25 @@ wasm_runtime_invoke_native(WASMExecEnv *exec_env, void *func_ptr,
...
                 break;
+#if WASM_ENABLE_REF_TYPES != 0
+            case VALUE_TYPE_EXTERNREF:
+            {
+                if (is_aot_func) {
+                    argv_ret[0] = invokeNative_Int32(func_ptr, argv1, n_stacks);
+                }
+                else {
+                    uint32 externref_idx;
+                    int32 ret = invokeNative_Int64(func_ptr, argv1, n_stacks);
```
wenyongh, line 3760: Should use int64 ret, here it is 64-bit system, ref is 64-bit

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3694,6 +4022,11 @@ wasm_externref_obj2ref(WASMModuleInstanceCommon *module_inst, void *extern_obj,
     ExternRefMapNode *node;
     uint32 externref_idx;
 
+    if (NULL_REF == (uint32)(uintptr_t)extern_obj) {
```
wenyongh, line 4025: Should not use NULL_REF, and should not used uint32, in 64-bit, uintptr_t is truncated.
Had better use `if (UINTPTR_MAX == (uintptr_t)extern_obj)`

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3746,7 +4079,8 @@ wasm_externref_ref2obj(uint32 externref_idx, void **p_extern_obj)
     ExternRefMapNode *node;
 
     if (externref_idx == NULL_REF) {
-        return false;
+        *p_extern_obj = NULL;
```
wenyongh, line 4082: If the object is NULL, will it be converted into NULL_REF in obj2ref?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -4215,29 +4544,33 @@ wasm_runtime_invoke_c_api_native(WASMModuleInstanceCommon *module_inst,
...
-                                          * func_type->param_count))) {
-        wasm_runtime_set_exception(module_inst, "allocate memory failed");
-        return false;
+    if (func_type->param_count > 16) {
+        if (!(params = wasm_runtime_malloc(sizeof(wasm_val_t)
+                                           * func_type->param_count))) {
+            wasm_runtime_set_exception(module_inst, "allocate memory failed");
+            return false;
+        }
+        memset(params, 0, sizeof(wasm_val_t) * func_type->param_count);
```
wenyongh, line 4559: Had better use 'runtime_malloc(sizefo(wasm_val_t) * func_type->param_count, module_inst, NULL, 0)' instead, we have defined runtime_malloc function in this file, it will memset the memory to 0, and throw exception when malloc fails.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -4215,29 +4544,33 @@ wasm_runtime_invoke_c_api_native(WASMModuleInstanceCommon *module_inst,
...
         goto fail;
     }
 
-    if (func_type->result_count > 4
-        && !(results = wasm_runtime_malloc(sizeof(wasm_val_t)
-                                           * func_type->result_count))) {
-        wasm_runtime_set_exception(module_inst, "allocate memory failed");
-        goto fail;
+    if (func_type->result_count > 4) {
+        if (!(results = wasm_runtime_malloc(sizeof(wasm_val_t)
```
wenyongh, line 4568: Same as above

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1113,46 +1113,159 @@ wasm_runtime_lookup_function(WASMModuleInstanceCommon *const module_inst,
...
+                    uint32 parts[2];
+                } u;
+
+                u.parts[0] = argv[argv_i];
+                u.parts[1] = argv[argv_i + 1];
+                externref_obj = (void *)u.val;
+#endif
+                if (!wasm_externref_obj2ref(exec_env->module_inst,
+                                            externref_obj, &externref_index)) {
+                    return false;
```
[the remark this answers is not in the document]
wenyongh: Should free new_argv before return false, it is allocated but haven't set to *ret_argv.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1113,46 +1113,159 @@ wasm_runtime_lookup_function(WASMModuleInstanceCommon *const module_inst,
...
+        if (result_type == VALUE_TYPE_EXTERNREF) {
+            void *externref_obj;
+#if UINTPTR_MAX != UINT32_MAX
+            union {
+                uintptr_t val;
+                uint32 parts[2];
+            } u;
+#endif
+
+            if (!wasm_externref_ref2obj(argv[argv_i], &externref_obj)) {
```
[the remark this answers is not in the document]
wenyongh: Should free argv before return false, L1345 is to free it, but it isn't called here.

core/iwasm/common/wasm_runtime_common.c
lum1n0us: The code is trying to catch a variable with the initialized value with `ref.null`. So, it is a *NULL_REF* originally.

core/iwasm/common/wasm_runtime_common.c
lum1n0us: The code is trying to catch a variable initialized with `ref.null`. 

if the object is NULL, we shall give it a reference id instead of NULL_REF.

core/iwasm/aot/aot_runtime.c
```diff
@@ -2309,9 +2301,9 @@ aot_call_indirect(WASMExecEnv *exec_env, uint32 tbl_idx, uint32 table_elem_idx,
         signature = import_func->signature;
         if (import_func->call_conv_raw) {
             attachment = import_func->attachment;
-            return wasm_runtime_invoke_native_raw(exec_env, func_ptr, func_type,
-                                                  signature, attachment, argv,
-                                                  argc, argv);
+            return wasm_runtime_invoke_native_raw(exec_env, func_ptr, true,
```
wenyongh, line 2304: Should be false, here func_ptr is native func ptr.

core/iwasm/common/wasm_application.c
```diff
@@ -628,16 +665,26 @@ wasm_application_execute_func(WASMModuleInstanceCommon *module_inst,
...
-                if (argv1[k] != NULL_REF) {
-                    void *extern_obj = NULL;
-                    bool ret = wasm_externref_ref2obj(argv1[k], &extern_obj);
-                    bh_assert(ret);
-                    (void)ret;
-                    os_printf("%p:ref.extern", extern_obj);
+#if UINTPTR_MAX == UINT32_MAX
+                if (argv1[k] != 0 && NULL_REF != argv1[k]
+                    && argv1[k] != (uint32)-1) {
+                    os_printf("%p:ref.extern", argv1[k]);
```
wenyongh, line 671: (void *)(uintptr_t)argv[k] to advoid compile warning.

core/iwasm/common/wasm_application.c
```diff
@@ -628,16 +665,26 @@ wasm_application_execute_func(WASMModuleInstanceCommon *module_inst,
...
             case VALUE_TYPE_EXTERNREF:
             {
-                if (argv1[k] != NULL_REF) {
-                    void *extern_obj = NULL;
-                    bool ret = wasm_externref_ref2obj(argv1[k], &extern_obj);
-                    bh_assert(ret);
-                    (void)ret;
-                    os_printf("%p:ref.extern", extern_obj);
+#if UINTPTR_MAX == UINT32_MAX
+                if (argv1[k] != 0 && NULL_REF != argv1[k]
```
wenyongh, line 669: no need to chceck NULL_REF != argv1[k]

core/iwasm/common/wasm_application.c
```diff
@@ -628,16 +665,26 @@ wasm_application_execute_func(WASMModuleInstanceCommon *module_inst,
...
+                    && argv1[k] != (uint32)-1) {
+                    os_printf("%p:ref.extern", argv1[k]);
+                    k++;
```
wenyongh, line 672: Should move else condition in here, or for else, k isn't increased. Had better:
```
             if (...)
                 os_printf("%p:ref.extern", argv1[k]);
             else
                 os_printf("extern:ref.null");
             k++;
```

core/iwasm/common/wasm_application.c
```diff
@@ -628,16 +665,26 @@ wasm_application_execute_func(WASMModuleInstanceCommon *module_inst,
...
                 }
+#else
+                union {
+                    uintptr_t val;
+                    uint32 parts[2];
+                } u;
+                u.parts[0] = argv1[k];
+                u.parts[1] = argv1[k + 1];
+                k += 2;
+                if (u.val && u.val != NULL_REF && u.val != (uintptr_t)-1LL) {
```
wenyongh, line 682: No need to check u.val != NULL_REF?

core/iwasm/common/wasm_application.c
```diff
@@ -628,16 +665,26 @@ wasm_application_execute_func(WASMModuleInstanceCommon *module_inst,
...
+                    os_printf("%p:ref.extern", (void *)u.val);
+                }
```
wenyongh, line 684: Had better move else inside #endif:
```
if (...)
    os_printf("%p:ref.extern", (void *)u.val);
else
    os_printf("extern:ref.null");
```

core/iwasm/aot/aot_runtime.c
```diff
@@ -2396,8 +2388,9 @@ aot_call_indirect(WASMExecEnv *exec_env, uint32 tbl_idx, uint32 table_elem_idx,
         return true;
     }
     else {
-        ret = invoke_native_internal(exec_env, func_ptr, func_type, signature,
-                                     attachment, argv, argc, argv);
+        ret = invoke_native_internal(
+            exec_env, func_ptr, func_idx < aot_module->import_func_count,
```
wenyongh, line 2392: func_idx >= aot_module->import_func_count

core/iwasm/common/wasm_application.c
```diff
@@ -616,10 +653,10 @@ wasm_application_execute_func(WASMModuleInstanceCommon *module_inst,
                 os_printf("%.7g:f64", u.val);
                 break;
             }
-#if WASM_ENABLE_REF_TYPES
+#if WASM_ENABLE_REF_TYPES != 0
             case VALUE_TYPE_FUNCREF:
             {
-                if (argv1[k] != NULL_REF)
+                if (argv1[k] != NULL_REF && argv1[k] != (uint32)-1)
```
wenyongh, line 659: remove -1 check

PR #853, opened 2021-11-26
Fix crash issue if wasm file is malformed(armv7 32bit)

JavanZhu:
Environment
* cpu: armv7
* platform: android

aot file content like following will cause runtime to crash in 32bit system(the value of p is overflow):
```
00 61 6F 74 03 00 00 00 00 00 00 00 60 2B 05 E3
```
[malformed.aot.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/7607958/malformed.aot.zip)


wenyongh:
> Environment
> 
> * cpu: armv7
> * platform: android
> 
> aot file content like following will cause runtime to crash in 32bit system(the value of p is overflow):
> 
> ```
> 00 61 6F 74 03 00 00 00 00 00 00 00 60 2B 05 E3
> ```
> 
> [malformed.aot.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/7607958/malformed.aot.zip)

@JavanZhu Thanks for reporting the issue! Your fix might be not complete and some spec cases run failed. I updated have the patch according to my understanding, could you help review and test it?

PR #856, opened 2021-11-30
[source debug] refine some code in source debugging

xujuntwt95329:
- move the wait_cond from exec_env to debug_instance, so the debug thread can be waken up by any threads
- process more general query message from debugger
- refine debug instance create/destroy mechanism
- avoid creating debug instance during module instantiating
- avoid blocking execution thread during creating debug instance
- update related documents

### DIFF
--- a/core/iwasm/interpreter/wasm.h
+++ b/core/iwasm/interpreter/wasm.h
@@ -420,12 +420,15 @@ struct WASMModule {
 #endif
 
 #if WASM_ENABLE_DEBUG_INTERP != 0
-    /* count how many instances reference this module. When source
-        debugging feature enabled, the debugger may modify the code
-        section of the module, so we need to report a warning if user
-        create several instance based on the same module
-       sub_instances created by lib-pthread or spawn API will not
-        influence or check the ref count */
+    /**
+     * Count how many instances reference this module. When source
+     * debugging feature enabled, the debugger may modify the code
+     * section of the module, so we need to report a warning if user
+     * create several instance based on the same module
+     *
+     * Sub_instances created by lib-pthread or spawn API will not
+     * influence or check the ref count
+     */
     uint32 ref_count;
     korp_mutex ref_count_lock;
 #endif
--- a/core/iwasm/interpreter/wasm.h
+++ b/core/iwasm/interpreter/wasm.h
@@ -424,7 +424,7 @@ struct WASMModule {
      * Count how many instances reference this module. When source
      * debugging feature enabled, the debugger may modify the code
      * section of the module, so we need to report a warning if user
-     * create several instance based on the same module
+     * create several instances based on the same module
      *
      * Sub_instances created by lib-pthread or spawn API will not
      * influence or check the ref count
--- a/core/iwasm/interpreter/wasm.h
+++ b/core/iwasm/interpreter/wasm.h
@@ -419,6 +419,17 @@ struct WASMModule {
     uint64 buf_code_size;
 #endif
 
+#if WASM_ENABLE_DEBUG_INTERP != 0
+    /* count how many instances reference this module. When source
+        debugging feature enabled, the debugger may modify the code
+        section of the module, so we need to report a warning if user
+        create several instance based on the same module
+       sub_instances created by lib-pthread or spawn API will not
+        influence or check the ref count */
+    uint32 ref_count;
+    korp_mutex ref_count_lock;
+#endif
+
 #if WASM_ENABLE_CUSTOM_NAME_SECTION != 0
     const uint8 *name_section_buf;
     const uint8 *name_section_buf_end;
--- a/core/iwasm/interpreter/wasm_runtime.c
+++ b/core/iwasm/interpreter/wasm_runtime.c
@@ -1680,6 +1680,9 @@ wasm_create_exec_env_and_call_function(WASMModuleInstance *module_inst,
         }
 
 #if WASM_ENABLE_THREAD_MGR != 0
+#if WASM_ENABLE_DEBUG_INTERP != 0
+        wasm_runtime_start_debug_instance(exec_env);
+#endif
     }
 #endif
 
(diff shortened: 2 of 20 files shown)


xujuntwt95329:
@ApsarasX @qinxk-inter @mbbill  I made some refactor in source debugging for better embedder usage, also for further multi-thread wasm application support. Any comments are well appreciated!

doc/source_debugging.md
```diff
@@ -89,3 +89,51 @@ lldb-12 iwasm -- test.aot
...
+
+    You can use `-DWAMR_BUILD_DEBUG_INTERP=1` during cmake configuration
+
+    Or you can set it directly in `cmake` files:
+    ``` cmake
+    set (WAMR_BUILD_DEBUG_INTERP 1)
+    ```
+
+### Attentions
+- Debugging `multi-thread wasm module` is not supported, if your wasm module use pthread APIs (see [pthread_library.md](./pthread_library.md)), or the embedder use `wasm_runtime_spawn_thread` to create new wasm threads, then there may be **un-expected behaviour** during debugging.
```
mbbill, line 135: nit, *unexpected* is one word.

doc/source_debugging.md
```diff
@@ -89,3 +89,51 @@ lldb-12 iwasm -- test.aot
...
+
+    > Note: This attention is about "wasm thread" rather than native threads. Executing wasm functions in several different native threads will **not** affect the normal behaviour of debugging feature.
+
+- When using source debugging features, **don't** create multiple `wasm_instance` from the same `wasm_module`, because the debugger may change the bytecode (set/unset breakpoints) of the `wasm_module`. If you do need several instance from the same bytecode, you need to copy the bytecode to a new butter, then load a new `wasm_module`, and then instantiate the new wasm module to get the new instance.
```
mbbill, line 139: a sanity check would be better.

doc/source_debugging.md
xujuntwt95329: Done, thanks!

doc/source_debugging.md
xujuntwt95329: Thanks for your suggestion, a sanity check would greatly improve the user experience. I've added a check for this, and a warning will be produced once this happens.

core/iwasm/interpreter/wasm.h
```diff
@@ -419,6 +419,17 @@ struct WASMModule {
     uint64 buf_code_size;
 #endif
 
+#if WASM_ENABLE_DEBUG_INTERP != 0
+    /* count how many instances reference this module. When source
+        debugging feature enabled, the debugger may modify the code
+        section of the module, so we need to report a warning if user
+        create several instance based on the same module
+       sub_instances created by lib-pthread or spawn API will not
+        influence or check the ref count */
```
wenyongh, line 428: Had better make code indent the same?

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -1126,6 +1126,17 @@ wasm_instantiate(WASMModule *module, bool is_sub_inst, uint32 stack_size,
     if (!module)
         return NULL;
 
+#if WASM_ENABLE_DEBUG_INTERP != 0
+    if (!is_sub_inst) {
+        os_mutex_lock(&module->ref_count_lock);
+        if (module->ref_count != 0) {
+            LOG_WARNING("multiple instance referencing the same module may "
```
wenyongh, line 1133: Had better change to LOG_WARNING("warning: ...")

wenyongh: multiple instance**s** referencing?

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -1126,6 +1126,17 @@ wasm_instantiate(WASMModule *module, bool is_sub_inst, uint32 stack_size,
     if (!module)
         return NULL;
 
+#if WASM_ENABLE_DEBUG_INTERP != 0
+    if (!is_sub_inst) {
+        os_mutex_lock(&module->ref_count_lock);
```
wenyongh, line 1131: Seems no need to add lock, but adding is also OK.

core/iwasm/interpreter/wasm.h
xujuntwt95329: Done

core/iwasm/interpreter/wasm_runtime.c

core/iwasm/interpreter/wasm_runtime.c
xujuntwt95329: There may be potential data race in this implementation, the ref_count is not increased before finishing instantiation, if some other threads instantiate the at the same time, the ref_count will be zero as the previous instantiation is not finished.

I've changed this code, we increase the ref_count here, and decrease it if instantiation failed

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -1519,8 +1533,17 @@ wasm_instantiate(WASMModule *module, bool is_sub_inst, uint32 stack_size,
...
 #endif
     (void)global_data_end;
+
     return module_inst;
+
 fail:
+#if WASM_ENABLE_DEBUG_INTERP != 0
+    if (!is_sub_inst) {
+        os_mutex_lock(&module->ref_count_lock);
+        module->ref_count--;
```
wenyongh, line 1543: Seems no need to decrease it here, the following wasm_instantiate decreate it again

xujuntwt95329: 1: Yes, you're right. I move these line to the place where malloc memory for Instance failed.
2: AoTModule is not related to interpreter debugging, I'll consider that when refactoring AOT debugging, that is in lower priority currently.

core/iwasm/interpreter/wasm.h
```diff
@@ -419,6 +419,20 @@ struct WASMModule {
     uint64 buf_code_size;
 #endif
 
+#if WASM_ENABLE_DEBUG_INTERP != 0
+    /**
+     * Count how many instances reference this module. When source
+     * debugging feature enabled, the debugger may modify the code
+     * section of the module, so we need to report a warning if user
+     * create several instance based on the same module
```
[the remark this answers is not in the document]
xujuntwt95329: Done

PR #860, opened 2021-12-01
Don't throw exception while malloc failed

no1wudi:
* Exception will terminate the wasm app, it's not necessary since app should check the result of dynamic allocation, and they can do some cleanup or fallback operation on fail instead of 'crash' directly.
* In acquire_wait_info the key of hasn_map_find can be NULL thus there are
many sesenless error log

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -1774,7 +1774,7 @@ aot_module_malloc(AOTModuleInstance *module_inst, uint32 size,
             aot_set_exception(module_inst, "app heap corrupted");
         }
         else {
-            LOG_WARNING("alloc %d bytes memory failed", size);
+            LOG_WARNING("warning: allocate %u bytes memory failed", size);
         }
         return 0;
     }
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -1774,7 +1774,7 @@ aot_module_malloc(AOTModuleInstance *module_inst, uint32 size,
             aot_set_exception(module_inst, "app heap corrupted");
         }
         else {
-            aot_set_exception(module_inst, "out of memory");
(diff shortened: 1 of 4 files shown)


core/iwasm/aot/aot_runtime.c
```diff
@@ -1774,7 +1774,7 @@ aot_module_malloc(AOTModuleInstance *module_inst, uint32 size,
             aot_set_exception(module_inst, "app heap corrupted");
         }
         else {
-            aot_set_exception(module_inst, "out of memory");
+            LOG_ERROR("alloc %d bytes memory failed", size);
```
wenyongh, line 1777: Had better change to `LOG_WARNING("warning: allocate %d bytes memory failed", size)`? LOG_ERROR normally means some error happens.

core/shared/utils/bh_hashmap.c
```diff
@@ -127,8 +127,13 @@ bh_hash_map_find(HashMap *map, void *key)
     HashMapElem *elem;
     void *value;
 
-    if (!map || !key) {
-        LOG_ERROR("HashMap find elem failed: map or key is NULL.\n");
+    if (!map) {
```
wenyongh, line 130: My understanding is that we should modify acquire_wait_info in wasm_shared_memory.c but not here, e.g. change to:
```C
static AtomicWaitInfo *
acquire_wait_info(void *address, bool create)
{
    bh_list_status ret;

    if (address)    => add check here
        wait_info = (AtomicWaitInfo *)bh_hash_map_find(wait_map, address);

    if (!create)
        return wait_info;
```
@xujuntwt95329, what's you opinion? Qi found that "In acquire_wait_info the key of hasn_map_find can be NULL thus there are many senseless error logs".

xujuntwt95329: Yes, I agree with @wenyongh, the caller should avoid passing NULL pointer to `bh_hash_map_find`

PR #874, opened 2021-12-06
Add new E_TYPE_XIP to indicate XIP mode

no1wudi:
WAMRC will emit E_TYPE_XIP while enable indirect mode, this will break the capability.

### DIFF
--- a/core/iwasm/aot/aot_intrinsic.c
+++ b/core/iwasm/aot/aot_intrinsic.c
@@ -53,6 +53,8 @@ static const aot_intrinsic g_intrinsic_mapping[] = {
     { "f64_convert_i64_u", "aot_intrinsic_u64_to_f64", AOT_INTRINSIC_FLAG_U64_TO_F64 },
     { "f32_convert_i64_s", "aot_intrinsic_i64_to_f32", AOT_INTRINSIC_FLAG_I64_TO_F32 },
     { "f32_convert_i64_u", "aot_intrinsic_u64_to_f32", AOT_INTRINSIC_FLAG_U64_TO_F32 },
+    { "i32_trunc_f32_u", "aot_intrinsic_f32_to_u32", AOT_INTRINSIC_FLAG_F32_TO_U32 },
+    { "i32_trunc_f32_s", "aot_intrinsic_f32_to_i32", AOT_INTRINSIC_FLAG_F32_TO_I32 },
     { "i32_trunc_f64_u", "aot_intrinsic_f64_to_u32", AOT_INTRINSIC_FLAG_F64_TO_U32 },
     { "i32_trunc_f64_s", "aot_intrinsic_f64_to_i32", AOT_INTRINSIC_FLAG_F64_TO_I32 },
     { "f32_demote_f64", "aot_intrinsic_f64_to_f32", AOT_INTRINSIC_FLAG_F64_TO_F32 },
--- a/core/iwasm/aot/aot_intrinsic.c
+++ b/core/iwasm/aot/aot_intrinsic.c
@@ -5,8 +5,6 @@
 
 #include "aot_intrinsic.h"
 
-#include <float.h>
-
 typedef struct {
     const char *llvm_intrinsic;
     const char *native_intrinsic;
@@ -423,12 +421,14 @@ aot_intrinsic_f64_to_f32(float64 f)
     return (float32)f;
 }
 
+/* The epsilon value is from https://www.cplusplus.com/reference/cfloat/ */
+
 int32
 aot_intrinsic_f32_cmp(AOTFloatCond cond, float32 lhs, float32 rhs)
 {
     switch (cond) {
         case FLOAT_EQ:
-            return fabsf(lhs - rhs) <= __FLT_EPSILON__ ? 1 : 0;
+            return fabsf(lhs - rhs) <= 1e-5f ? 1 : 0;
 
         case FLOAT_LT:
             return lhs < rhs ? 1 : 0;
@@ -459,7 +459,7 @@ aot_intrinsic_f64_cmp(AOTFloatCond cond, float64 lhs, float64 rhs)
 {
     switch (cond) {
         case FLOAT_EQ:
-            return fabs(lhs - rhs) <= __DBL_EPSILON__ ? 1 : 0;
+            return fabs(lhs - rhs) <= 1e-9 ? 1 : 0;
 
         case FLOAT_LT:
             return lhs < rhs ? 1 : 0;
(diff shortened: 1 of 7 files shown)


core/iwasm/aot/aot_loader.c
```diff
@@ -2492,33 +2493,36 @@ destroy_sections(AOTSection *section_list, bool destroy_aot_text)
...
 static bool
-resolve_native_symbols(const uint8 *buf, uint32 size, uint32 *p_count,
-                       char *error_buf, uint32 error_buf_size)
+resolve_execute_mode(const uint8 *buf, uint32 size, bool *p_mode,
+                     char *error_buf, uint32 error_buf_size)
 {
     const uint8 *p = buf, *p_end = buf + size;
     uint32 section_type;
     uint32 section_size = 0;
+    uint16_t e_type = 0;
```
wenyongh, line 2502: uint16, normally we don't use _t inside runtime

core/iwasm/aot/aot_loader.c
```diff
@@ -2536,18 +2540,18 @@ create_sections(AOTModule *module, const uint8 *buf, uint32 size,
     AOTSection *section_list = NULL, *section_list_end = NULL, *section;
     const uint8 *p = buf, *p_end = buf + size;
     bool destroy_aot_text = false;
-    uint32 native_symbol_count = 0;
+    bool is_indirect_mode = 0;
```
wenyongh, line 2543: Had better change "= 0" to "= false"

core/iwasm/compilation/aot_emit_aot_file.c
```diff
@@ -2047,6 +2047,11 @@ aot_resolve_target_info(AOTCompContext *comp_ctx, AOTObjectData *obj_data)
         }
 
         elf_header = (struct elf32_ehdr *)elf_buf;
+
+        /* Emit eXecute In Place file type while in indirect mode */
+        if (comp_ctx->is_indirect_mode)
+            elf_header->e_type = 4;
```
wenyongh, line 2053: Should also set e_type before line 2028 and before line 2071.
And had better "define E_TYPE_XIP 4", e.g. in line 1911, before `#define IMAGE_FILE_MACHINE_AMD64 0x8664`

wenyongh (commented):
And should also modify function wasm_runtime_is_xip_file in wasm_runtime_common.c

core/iwasm/compilation/aot_emit_aot_file.c
no1wudi: Done

core/iwasm/compilation/aot_emit_aot_file.c
```diff
@@ -1909,6 +1909,8 @@ struct coff_hdr {
     U16 u16Characs;
 };
 
+#define E_TYPE_XIP 4
```
wenyongh, line 1912: Also define E_TYPE_REL 1

core/iwasm/compilation/aot_emit_aot_file.c
```diff
@@ -2025,7 +2027,13 @@ aot_resolve_target_info(AOTCompContext *comp_ctx, AOTObjectData *obj_data)
             return false;
         }
         coff_header = (struct coff_hdr *)elf_buf;
-        obj_data->target_info.e_type = 1;
+
+        /* Emit eXecute In Place file type while in indirect mode */
+        if (comp_ctx->is_indirect_mode)
+            obj_data->target_info.e_type = 4;
```
wenyongh, line 2033: Use E_TYPE_XIP and E_TYPE_REL but not 4 and 1.

core/iwasm/aot/aot_intrinsic.c
```diff
@@ -5,6 +5,8 @@
 
 #include "aot_intrinsic.h"
 
+#include <float.h>
```
wenyongh: Should not include float.h here, normally we just include bh_platform.h, and implement the API in platform layer if necessary.

core/iwasm/aot/aot_intrinsic.c
```diff
@@ -423,6 +427,9 @@ int32
 aot_intrinsic_f32_cmp(AOTFloatCond cond, float32 lhs, float32 rhs)
 {
     switch (cond) {
+        case FLOAT_EQ:
+            return fabsf(lhs - rhs) <= __FLT_EPSILON__ ? 1 : 0;
```
wenyongh: fabsf isn't supported in zephyr platform, in fact we implemented fabs for in with core/shared/platform/common/math/math.c which is ported from open sourced math lib.
Could we use fabs to implement the fabsf, like in interpreter: `use (float32)fabs(lhs - rhs)` instead

core/iwasm/aot/aot_intrinsic.c
```diff
@@ -419,10 +421,15 @@ aot_intrinsic_f64_to_f32(float64 f)
...
 }
 
+/* The epsilon value is from https://www.cplusplus.com/reference/cfloat/ */
+
 int32
 aot_intrinsic_f32_cmp(AOTFloatCond cond, float32 lhs, float32 rhs)
 {
     switch (cond) {
+        case FLOAT_EQ:
+            return fabsf(lhs - rhs) <= 1e-5f ? 1 : 0;
```
wenyongh, line 431: fabsf isn't supported in zephyr platform, whose math lib is ported from open source library, see core/shared/platform/common/math/math.c. Can we use `(float32)fabs(lhs - rhs)` instead?
Another issue is that do we need to check <= 1e-5f? In interpreter f32.cmp opcode is just interpreted with f1 == f2.

core/iwasm/aot/aot_intrinsic.c
no1wudi: Forget to remove this include

core/iwasm/aot/aot_intrinsic.c
no1wudi: It's better to check it.

wenyongh (commented):
Hi, after testing , we found several issues, could you please help modify:
1. aot_loader.c, about Line 1470:
Change
if ((module->code_size > 0) && (module->native_symbol_count == 0)) {
to 
if ((module->code_size > 0) && !module->is_indirect_mode) {

2. aot_loader.c, about Line 480:
remove
module->native_symbol_count = cnt;

3. aot_runtime.h
remove native_symbol_count field in struct AOTModule

To test spec cases for AOT with XIP mode, you can:
cd tests/wamr-test-suites
./test_wamr.sh -s spec -t aot -X

PR #891, opened 2021-12-13
remove is_aot_func from wasm_runtime_invoke_nativeXX

lum1n0us:


core/iwasm/common/wasm_runtime_common.c
```diff
@@ -2770,18 +2770,13 @@ wasm_runtime_invoke_native_raw(WASMExecEnv *exec_env, void *func_ptr,
             {
                 uint32 externref_idx = *argv_src++;
 
-                if (is_aot_func) {
-                    *(uint32 *)argv_dst = externref_idx;
-                }
-                else {
-                    void *externref_obj;
+                void *externref_obj;
```
wenyongh, line 2773: Seems we should replace `if (is_aot_func)` with `if (!signature)` but not remove it

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -2814,23 +2809,17 @@ wasm_runtime_invoke_native_raw(WASMExecEnv *exec_env, void *func_ptr,
...
             case VALUE_TYPE_EXTERNREF:
             {
-                if (is_aot_func) {
-                    /* return an externref idx */
-                    argv_ret[0] = *(uint32 *)argv1;
-                }
-                else {
-                    uint32 externref_idx;
-                    uint64 externref_obj;
+                uint32 externref_idx;
```
wenyongh, line 2812: Same as above

PR #894, opened 2021-12-15
xip: Lookup float constants from table to reduce relocations

wenyongh:
Lookup float/double constants from exec_env->native_symbol table
but not construct them with LLVMBuildConst if XIP mode is enabled,
these constants are introduced by f32/f64.const opcodes and some
float/double conversion opcodes, and make wamrc generate some
relocations in text section of AOT XIP file. This patch eliminates such
relocations when "--enable-indirect-mode" is added to wamrc.

### DIFF
--- a/core/iwasm/aot/aot_loader.c
+++ b/core/iwasm/aot/aot_loader.c
@@ -493,23 +493,26 @@ load_native_symbol_section(const uint8 *buf, const uint8 *buf_end,
 
         for (i = cnt - 1; i >= 0; i--) {
             read_string(p, p_end, symbol);
-            if (!strncmp(symbol, "i32_const#", strlen("i32_const#"))) {
+            if (!strncmp(symbol, "i32#", 4)) {
                 uint32 u32;
-                if (!str2uint32(symbol + strlen("i32_const#"), &u32)) {
+                if (!str2uint32(symbol + 4, &u32)) {
                     set_error_buf_v(error_buf, error_buf_size,
                                     "resolve symbol %s failed", symbol);
                     goto fail;
                 }
                 *(uint32 *)(&module->native_symbol_list[i]) = u32;
             }
-            else if (!strncmp(symbol, "i64_const#", strlen("i64_const#"))) {
+            else if (!strncmp(symbol, "i64#", 4)) {
                 uint64 u64;
-                if (!str2uint64(symbol + strlen("i64_const#"), &u64)) {
+                if (!str2uint64(symbol + 4, &u64)) {
                     set_error_buf_v(error_buf, error_buf_size,
                                     "resolve symbol %s failed", symbol);
                     goto fail;
                 }
                 *(uint64 *)(&module->native_symbol_list[i]) = u64;
+            }
+            else if (!strncmp(symbol, "__ignore", 8)) {
+                continue;
             }
             else {
                 module->native_symbol_list[i] =
--- a/core/iwasm/aot/aot_loader.c
+++ b/core/iwasm/aot/aot_loader.c
@@ -466,6 +466,12 @@ get_native_symbol_by_name(const char *name)
 }
 
 static bool
+str2uint32(const char *buf, uint32 *p_res);
+
+static bool
+str2uint64(const char *buf, uint64 *p_res);
+
+static bool
 load_native_symbol_section(const uint8 *buf, const uint8 *buf_end,
                            AOTModule *module, bool is_load_from_file_buf,
                            char *error_buf, uint32 error_buf_size)
@@ -487,11 +493,32 @@ load_native_symbol_section(const uint8 *buf, const uint8 *buf_end,
 
         for (i = cnt - 1; i >= 0; i--) {
             read_string(p, p_end, symbol);
-            module->native_symbol_list[i] = get_native_symbol_by_name(symbol);
-            if (module->native_symbol_list[i] == NULL) {
-                set_error_buf_v(error_buf, error_buf_size,
-                                "missing native symbol: %s", symbol);
-                goto fail;
+            if (!strncmp(symbol, "i32_const#", strlen("i32_const#"))) {
+                uint32 u32;
+                if (!str2uint32(symbol + strlen("i32_const#"), &u32)) {
+                    set_error_buf_v(error_buf, error_buf_size,
+                                    "resolve symbol %s failed", symbol);
+                    goto fail;
+                }
+                *(uint32 *)(&module->native_symbol_list[i]) = u32;
+            }
+            else if (!strncmp(symbol, "i64_const#", strlen("i64_const#"))) {
+                uint64 u64;
+                if (!str2uint64(symbol + strlen("i64_const#"), &u64)) {
+                    set_error_buf_v(error_buf, error_buf_size,
+                                    "resolve symbol %s failed", symbol);
+                    goto fail;
+                }
+                *(uint64 *)(&module->native_symbol_list[i]) = u64;
+            }
+            else {
+                module->native_symbol_list[i] =
+                    get_native_symbol_by_name(symbol);
+                if (module->native_symbol_list[i] == NULL) {
+                    set_error_buf_v(error_buf, error_buf_size,
+                                    "missing native symbol: %s", symbol);
+                    goto fail;
+                }
             }
         }
     }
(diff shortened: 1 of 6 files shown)


core/iwasm/aot/aot_loader.c
```diff
@@ -1711,7 +1741,6 @@ is_literal_relocation(const char *reloc_sec_name)
     return !strcmp(reloc_sec_name, ".rela.literal");
 }
 
-#if defined(BH_PLATFORM_WINDOWS)
 static bool
 str2uint32(const char *buf, uint32 *p_res)
```
lum1n0us, line 1745: what is the difference between `str2uint32` to the `strtol` from libc ?

core/iwasm/compilation/aot_llvm.c
```diff
@@ -2444,3 +2475,65 @@ aot_get_func_from_table(const AOTCompContext *comp_ctx, LLVMValueRef base,
...
+        case VALUE_TYPE_I32:
+            snprintf(buf, sizeof(buf), "i32#%08X", value->i32);
+            const_ptr_type = INT32_PTR_TYPE;
+            break;
+        case VALUE_TYPE_I64:
+            snprintf(buf, sizeof(buf), "i64#%016" PRIx64, value->i64);
+            const_ptr_type = INT64_PTR_TYPE;
+            break;
+        case VALUE_TYPE_F32:
+            snprintf(buf, sizeof(buf), "i32#%08X", value->i32);
```
xujuntwt95329, line 2498:
```suggestion
            snprintf(buf, sizeof(buf), "i32#%08X", value->f32);
```
Seems there should be f32?

core/iwasm/compilation/aot_llvm.c
```diff
@@ -2003,6 +2003,29 @@ aot_destroy_comp_context(AOTCompContext *comp_ctx)
...
+{
+    AOTNativeSymbol *sym = wasm_runtime_malloc(sizeof(AOTNativeSymbol));
+
+    if (!sym) {
+        aot_set_last_error("alloc native symbol failed.");
+        return false;
+    }
+
+    memset(sym, 0, sizeof(AOTNativeSymbol));
+    snprintf(sym->symbol, sizeof(sym->symbol), "%s", symbol);
```
xujuntwt95329, line 2017: Should we add an assert to ensure the symbol is no longer than the buffer?
``` C
bh_assert(strlen(symbol) <= sizeof(sym->symbol));
```

core/iwasm/aot/aot_loader.c
xujuntwt95329: @lum1n0us  Seems the result of str2uint32 is a `raw hex` rather than an integer

core/iwasm/compilation/aot_llvm.c
lum1n0us: f32#%08X?

core/iwasm/aot/aot_loader.c
```diff
@@ -487,11 +493,35 @@ load_native_symbol_section(const uint8 *buf, const uint8 *buf_end,
 
         for (i = cnt - 1; i >= 0; i--) {
             read_string(p, p_end, symbol);
-            module->native_symbol_list[i] = get_native_symbol_by_name(symbol);
-            if (module->native_symbol_list[i] == NULL) {
-                set_error_buf_v(error_buf, error_buf_size,
-                                "missing native symbol: %s", symbol);
-                goto fail;
+            if (!strncmp(symbol, "i32#", 4)) {
```
lum1n0us, line 496: may let *i32* be renamed *n32*. less confusion over *i* representing *integer* and *f* representing *float*.

core/iwasm/aot/aot_loader.c
wenyongh: Yes, it is raw int bits of f32 and f64.

core/iwasm/compilation/aot_llvm.c
wenyongh: OK

core/iwasm/compilation/aot_llvm.c
wenyongh: Should be i32, store the raw int bits to make it simple and avoid data loss (possible data loss when using f32, e.g. NaN values).

core/iwasm/aot/aot_loader.c
wenyongh: has renamed to "f32#"

yamt:
@wenyongh
do you think it makes sense to do similar for i32.const etc?
it can save some of relocations from wasi-libc printf implementation on x86-64.

wenyongh:
@yamt it depends, not sure how much footprint it can reduce and how much performance it will impact? I remember that many x86 instructions directly encode i32 immediate as its operand, the performance should be better than that of looking up i32 immediate from the constant table.

yamt:
> @yamt it depends, not sure how much footprint it can reduce and how much performance it will impact? I remember that many x86 instructions directly encode i32 immediate as its operand, the performance should be better than that of looking up i32 immediate from the constant table.

i'm not how much impact it is.
do you have a better idea to eliminate these relocations?

wenyongh:
Is it only required by XIP mode? For non-XIP, no need to eliminate these relocations, right? Seems no other good idea except do the similar for i32.const.
Add @no1wudi to discuss.

yamt:
> Is it only required by XIP mode? For non-XIP, no need to eliminate these relocations, right?

yes

> Seems no other good idea except do the similar for i32.const. Add @no1wudi to discuss.

ok

no1wudi:
@yamt Could you share more info (disassembly code, etc) about which instruction will lead to relocation in your application ?

yamt:
> @yamt Could you share more info (disassembly code, etc) about which instruction will lead to relocation in your application ?

just "hello world" using printf.
libc from wasi-sdk-14.0-macos.tar.gz.

for example, a constant 100000000 (00e1f505) is loaded from .rodata.cst4, it seems.
```
Contents of section .rodata.cst4:
 0000 01000000 00e1f505 0a000000           ............
```
```
    316d: 41 89 cb                      movl    %ecx, %r11d
    3170: 48 8b 7c 24 50                movq    80(%rsp), %rdi
    3175: 85 d2                         testl   %edx, %edx
    3177: 0f 88 01 01 00 00             js      0x327e <aot_func#29+0x18ae>
    317d: 31 ed                         xorl    %ebp, %ebp
    317f: 45 39 de                      cmpl    %r11d, %r14d
    3182: 48 bf 00 00 00 00 00 00 00 00 movabsq $0, %rdi
                0000000000003184:  R_X86_64_64  .rodata.cst4+0x4
    318c: 73 2e                         jae     0x31bc <aot_func#29+0x17ec>
    318e: 48 8b 84 24 90 00 00 00       movq    144(%rsp), %rax
    3196: 44 29 f0                      subl    %r14d, %eax
    3199: c1 f8 02                      sarl    $2, %eax
    319c: 8d 2c c0                      leal    (%rax,%rax,8), %ebp
    319f: 44 89 f0                      movl    %r14d, %eax
    31a2: 41 8b 44 05 00                movl    (%r13,%rax), %eax
    31a7: 83 f8 0a                      cmpl    $10, %eax
    31aa: 72 10                         jb      0x31bc <aot_func#29+0x17ec>
    31ac: b9 0a 00 00 00                movl    $10, %ecx
    31b1: ff c5                         incl    %ebp
    31b3: 01 c9                         addl    %ecx, %ecx
    31b5: 8d 0c 89                      leal    (%rcx,%rcx,4), %ecx
    31b8: 39 c8                         cmpl    %ecx, %eax
```

maybe this one in `printf_core`, but i'm not sure.
```
 001d87: 03 40                      |                             loop
 001d89: 20 11                      |                               local.get 17
 001d8b: 41 80 c2 d7 2f             |                               i32.const 100000000
 001d90: 6c                         |                               i32.mul
 001d91: 21 11                      |                               local.set 17
 001d93: 20 16                      |                               local.get 22
 001d95: 41 08                      |                               i32.const 8
 001d97: 6a                         |                               i32.add
 001d98: 22 16                      |                               local.tee 22
 001d9a: 0d 00                      |                               br_if 0
 001d9c: 0b                         |                             end
```

PR #913, opened 2021-12-23
Add a check about WASI ABI compatibility

lum1n0us:
Refer to:
https://github.com/WebAssembly/WASI/blob/main/design/application-abi.md

Plus,
- refactoring wasm_loader_find_export
- remove MULTI_MODULE related from mini_loader
- Update the sample of multi-modules
- Fix a "use-after-free" problem. Since we are reusing the memory instance of sub-module's,
just to pretect from freeing an imported memory instance is enough


core/iwasm/common/wasm_application.c
```diff
@@ -96,22 +92,25 @@ wasm_application_execute_main(WASMModuleInstanceCommon *module_inst, int32 argc,
...
+    /* In wasi mode, we should call function named "_start"
+       which initializes the wasi envrionment and then calls
+       the actual main function. Directly call main function
+       may cause exception thrown. */
+    if ((func = wasm_runtime_lookup_wasi_start_function(module_inst))) {
+        return wasm_runtime_create_exec_env_and_call_wasm(module_inst, func, 0,
+                                                          NULL);
+    }
+    else {
+        wasm_runtime_set_exception(
```
wenyongh, line 104: Should not set exception here: (1) Must the wasm app compiled as wasi mode export _start function? (2) Here the WASM_ENABLE_LIBC_WASI is enabled, but it doesn't require that the wasm app must be a wasi mode app, the app can be non-wasi mode also, for example `/opt/wasi-sdk/bin/clang -nostdlib --no-entry --export=__main_argc_argv ...`

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1183,7 +1183,7 @@ wasm_runtime_get_function_type(const WASMFunctionInstanceCommon *function,
 }
 
 WASMFunctionInstanceCommon *
-wasm_runtime_lookup_function(WASMModuleInstanceCommon *const module_inst,
+wasm_runtime_lookup_function(WASMModuleInstanceCommon *module_inst,
```
wenyongh, line 1186: Shall we not change the API's prototype? I had a test, it doesn't cause compile error or warning. Here using const is reasonable, and the API had been used for a long time.

core/iwasm/common/wasm_application.c
```diff
@@ -264,53 +265,33 @@ resolve_function(const WASMModuleInstanceCommon *module_inst, const char *name)
...
             LOG_DEBUG("can not find a sub module named %s", sub_module_name);
             goto LEAVE;
         }
     }
+    else {
+        target_inst = module_inst;
+    }
 #else
-    const char *function_name = name;
+    function_name = name;
```
wenyongh, line 279: Compile warning occurs: `assignment discards ‘const’ qualifier from pointer target type`

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -3440,9 +3457,62 @@ load(const uint8 *buf, uint32 size, WASMModule *module, char *error_buf,
...
+             * XXX reactor" */
+            LOG_VERBOSE("A command should export _start function by default");
+        }
+    }
+    else {
+        initialize =
+            wasm_loader_find_export(module, "", "_initialize", EXPORT_KIND_FUNC,
+                                    error_buf, error_buf_size);
+        if (!initialize) {
+            set_error_buf(
```
wenyongh, line 3490: The document mentions that `A reactor may export a function named _initialize.. If an _initialize export is present`, my understanding is that `_initialize` function can be exported and also be unexported?

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -1036,6 +1031,44 @@ sub_module_instantiate(WASMModule *module, WASMModuleInstance *module_inst,
...
+#if WASM_ENABLE_LIBC_WASI != 0
+        {
+            /*
+             * reactor instances may assume that _initialize will be called by
+             * the environment at most once, and that none of their other
+             * exports are accessed before that call.
+             */
+            WASMFunctionInstance *initialize =
+                wasm_lookup_function(sub_module_inst, "_initialize", NULL);
+            /* a strong restricttion that a reactor must export a _initialize
```
wenyongh, line 1044: Same as above, seems that it doesn't require that `_initialize` function must be exported.

wenyongh: And we should not use macro WASM_ENABLE_LIBC_WASI to check that the wasm app is in wasi mode or not, normally we should use `module->is_wasi_mode`

product-mini/platforms/linux-sgx/enclave-sample/Makefile
```diff
@@ -102,6 +102,7 @@ Enclave_Include_Paths := -IEnclave -I$(WAMR_ROOT)/core/iwasm/include \
 Enclave_C_Flags := $(SGX_COMMON_CFLAGS) -nostdinc -fvisibility=hidden -fpie -fstack-protector $(Enclave_Include_Paths)
 ifeq ($(SPEC_TEST), 1)
 	Enclave_C_Flags += -DWASM_ENABLE_SPEC_TEST=1
+	Enclave_C_Flags += -DSGX_DISABLE_WASI
```
wenyongh, line 105: Don't disable WASI by default, SGX supports WASI and enables WASI by default.

tests/wamr-test-suites/test_wamr.sh
```diff
@@ -527,6 +527,7 @@ function trigger()
     local EXTRA_COMPILE_FLAGS=""
     # default enabled features
     EXTRA_COMPILE_FLAGS+=" -DWAMR_BUILD_BULK_MEMORY=1"
+    EXTRA_COMPILE_FLAGS+=" -DWAMR_BUILD_LIBC_WASI=0"
```
wenyongh, line 530: Had better not disable wasi by default

core/iwasm/interpreter/wasm_loader.c
lum1n0us: A reactor always has an exported `_initialize`. Hasn't find any exception.

core/iwasm/interpreter/wasm_runtime.c
lum1n0us: Well, it is what I am worried about. `is_wasi_mode` doesn't always equal to `WASM_ENABLE_LIBC_WASI=1`. It represents a module depends on WASI APIs. If a reactor compiled by WASI_SDK and doesn't include any standard libraries, it will not have any imported entries but still a reactor.

``` c
void foo() {}
```

product-mini/platforms/linux-sgx/enclave-sample/Makefile
lum1n0us: Since SPEC test cases are all from WAST and can't be compiled with `wasi-sdk-clang` and they are not using any WASI apis. we'd better run spec test cases by avoiding WASI module rules.

core/iwasm/interpreter/wasm_loader.c
wenyongh: OK. And should we also check that the sub-module should not export _start func, and main-module should not export _initialize func?

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -3458,6 +3528,14 @@ wasm_loader_load(const uint8 *buf, uint32 size, char *error_buf,
         goto fail;
     }
 
+#if (WASM_ENABLE_MULTI_MODULE != 0) && (WASM_ENABLE_LIBC_WASI != 0)
+    /* do a check about WASI Application ABI */
+    if (!check_wasi_abi_compatibility(module, main_module, error_buf,
```
wenyongh, line 3533: Only check the compatibility when module is in wasi mode, or error might occur when it is non-wasi module (e.g. compile with `/opt/wasi-sdk/bin/clang -nostdlib`):
```C
if (module->is_wasi_mode
    && !check_wasi_abi_compatibility(module, main_module, error_buf,
```

core/iwasm/interpreter/wasm_runtime.c
wenyongh: Yes, it is still a reactor and exports _inititialize, but the function does nothing, whehter to call it or not seems meaningless. There ight be two choices: (1) Call _initialize it is found, (2) Call _initialize when it is found and the module is wasi-module.

And also we should check the func signature of `_start` and `_initialize`, they must be "()".

product-mini/platforms/linux-sgx/enclave-sample/Makefile
wenyongh: That's because that  we didn't check module->is_wasi_mode when calling `_start` and `_initialize`. My understanding is that we check the wasi module rules only when the module is wasi moduel, or many non wasi cases may run failed.

lum1n0us: The point here is we should run SPEC-TEST in **non-wasi-mode**.  the modification won't change the default setting of SGX.

core/iwasm/common/wasm_application.c
lum1n0us: If with WASI, a .wasm is either a command or a reactor. There should be not a third case. So. YES we run the WASM app if there is a `_start`.  And we quit with a message if there is no `_start`.

lum1n0us: Plus, we shall treat `-nostdlib` as a behavior of disabled WASI.

core/iwasm/interpreter/wasm_loader.c
lum1n0us: Not sure about that. Clang should make sure that `_start` and `_initialize` can not be exported at the same time. But it is OK if we want to be safe first. I am open to both.

lum1n0us: Plus, if needs a check about signatures of `_start` and `_initialize`, we should have it done here.

core/iwasm/interpreter/wasm_loader.c
lum1n0us: We shall treat `clang -nostdlib` equal to `WASM_ENABLE_LIBC_WASI=0`. I mean we don't accept the third kind of .wasm beyond command/reactor model when `WASM_ENABLE_LIBC_WASI=1

core/iwasm/interpreter/wasm_runtime.c
lum1n0us: I tend to make it simple without considering .wasm is import any `wasi_snapshot_preview1` function or not.

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -1036,6 +1031,39 @@ sub_module_instantiate(WASMModule *module, WASMModuleInstance *module_inst,
...
+            /*
+             * reactor instances may assume that _initialize will be called by
+             * the environment at most once, and that none of their other
+             * exports are accessed before that call.
+             *
+             * let the loader decide how to act if there is no _initialize
+             * in a reactor
+             */
+            WASMFunctionInstance *initialize =
+                wasm_lookup_function(sub_module_inst, "_initialize", NULL);
```
wenyongh, line 1046: Seems had better implement API wasm_runtime_lookup_wasi_initialize_function and check the signature

core/iwasm/common/wasm_application.c
```diff
@@ -100,18 +100,15 @@ wasm_application_execute_main(WASMModuleInstanceCommon *module_inst, int32 argc,
...
 #endif /* end of WASM_ENABLE_LIBC_WASI */
 
     if (!(func = wasm_runtime_lookup_function(module_inst, "main", NULL))
         && !(func = wasm_runtime_lookup_function(module_inst,
                                                  "__main_argc_argv", NULL))
         && !(func = wasm_runtime_lookup_function(module_inst, "_main", NULL))) {
-        wasm_runtime_set_exception(module_inst, "lookup main function failed");
+        wasm_runtime_set_exception(
+            module_inst,
+            "lookup the entry point symbol(like _start, _main, _main) failed");
```
wenyongh, line 111: How about "lookup the entry point symbol (e.g. main, _main, __main_argc_argv) failed", _start isn't to be find here, and "_main, _main" is duplicated

core/iwasm/common/wasm_runtime_common.c
wenyongh: Could we not change the function prototype?

core/iwasm/common/wasm_application.c
lum1n0us: plan to include the *else* case of L99 missing the `_start`? maybe we shall create two messages when 
- `WASM_EABLE_LIBC_WASI=1`. *lookup the entry point symbol (like _start, main, _main, __main_argc_argv) failed*
- `WAMS_ENABLE_LIBC_WASI=0`. *lookup the entry point symbol (like main, _main, __main_argc_argv) failed*

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -3440,9 +3457,65 @@ load(const uint8 *buf, uint32 size, WASMModule *module, char *error_buf,
...
+        if (!start) {
+            /* without _start(), it maybe a reactor which is run via "iwasm -f
+             * XXX reactor" */
+            LOG_VERBOSE("A command should export _start function by default");
+        }
+    }
+    else {
+        initialize =
+            wasm_loader_find_export(module, "", "_initialize", EXPORT_KIND_FUNC,
+                                    error_buf, error_buf_size);
```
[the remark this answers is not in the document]
lum1n0us: `wasm_runtime_lookup_wasi_start_function` requires a module instance and we don't have it yet.

Plan to have a type-check during loading.

core/iwasm/interpreter/wasm_runtime.c
lum1n0us: after the type check during loading, we can avoid all secondary check.

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -3440,9 +3457,129 @@ load(const uint8 *buf, uint32 size, WASMModule *module, char *error_buf,
...
+            set_error_buf(
+                error_buf, error_buf_size,
+                "The builtin _initiazlie() is with a wrong signature");
+            return false;
+        }
+    }
+
+    /* filter out non-wasi compatiable modules */
+    if (!module->is_wasi_module && !start && !initialize) {
+        return true;
```
wenyongh, line 3528: when return true, memory isn't checked if is_wasi_module is true

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -3440,9 +3457,129 @@ load(const uint8 *buf, uint32 size, WASMModule *module, char *error_buf,
...
+    }
+
+    /*
+     * it is ok a reactor acts as a main module,
+     * so skip the check about (with `_initialize`)
+     */
+
+    memory = wasm_loader_find_export(module, "", "memory", EXPORT_KIND_MEMORY,
+                                     error_buf, error_buf_size);
+    if (!memory) {
```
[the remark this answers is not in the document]
lum1n0us: non-wasi modules are supposed to be all filtered out in L3527

core/iwasm/interpreter/wasm_loader.c
lum1n0us: wasi compatible modules should have at least one of `_start` or `_initialize`.  observation 1 and 2.

core/iwasm/interpreter/wasm_loader.c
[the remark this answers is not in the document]
wenyongh: it only filters out non wasi module which doesn't export start and initialize
L3532 checks is_wasi_module again

core/iwasm/interpreter/wasm_loader.c
xujuntwt95329: `LIBC_WASI` and `LIBC_BUILTIN` can be enabled at the same time.

core/iwasm/interpreter/wasm_loader.c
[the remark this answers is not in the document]
lum1n0us: L3532 and L3545 are kinds of protection to make sure that from L3552, there are all cases about WASI modules.

PR #914, opened 2021-12-25
Implement GetCurrentThreadStackLimits() for Windows 7

wenyongh:
And fix compilation error of when build wamrc, fix compilation warnings


lum1n0us (approved):
go with risk  😆

Issue #884, opened 2021-12-11
Potential sensor framework issue regarding the timer


LinJian1982:
It looks like we shall reset the ms_to_next_check if it is **large** than "s->read_interval". 
for example, we have two sensors in the system. The read interval of them are 100ms and 1000ms. The current code causes the ms_to_next_check be set to 1000ms, so the callback of the shorter one only be called after 1000ms.

```
if (elpased_ms >= s->read_interval) {
            attr_container_t * data = s->read(s);
            if (data) {
                sensor_client_t * client = s->clients;
                while (client) {
                    client->client_callback(client, s->sensor_id, data);
                    client = client->next;
                }
                attr_container_destroy(data);
            }
            s->last_read = now;

            if (ms_to_next_check == -1 || (ms_to_next_check < s->read_interval))
                ms_to_next_check = s->read_interval;
        } else {
            int remaining = s->read_interval - elpased_ms;
            if (ms_to_next_check == -1 || (ms_to_next_check < remaining))
                ms_to_next_check = remaining;

        }
```

xwang98:
@LinJian1982 good catch! Just wondering if you can create a PR for the fix?

wenyongh:
@LinJian1982 Thanks for reporting the issue! The bug was fixed with #917, could you please try again?

PR #917, opened 2021-12-27
Fix sensor framework timer issue and update sensor sample

wenyongh:
Fix the sensor framework timer issue reported by #884 when setting
`ms_to_next_check`, and modify the type of time related args/vars to
uint32 and fix the compilation warnings.

And update the sensor sample by creating two sensors to confirm that
the fix works correctly.


core/app-framework/sensor/native/runtime_sensor.c
```diff
@@ -395,12 +395,12 @@ check_sensor_timers()
...
             s->last_read = now;
 
-            if (ms_to_next_check == -1 || (ms_to_next_check < s->read_interval))
+            if (s->read_interval < ms_to_next_check)
                 ms_to_next_check = s->read_interval;
         }
         else {
-            int remaining = s->read_interval - elpased_ms;
-            if (ms_to_next_check == -1 || (ms_to_next_check < remaining))
+            uint32 remaining = s->read_interval - elpased_ms;
```
lum1n0us, line 402: shall we be worried if `s->read_interval` is smaller than `elpased_ms` ?

wenyongh: This is the else branch, s->read_interval <= elpased_ms is handled in if branch, see Line 385:
`if (elpased_ms >= s->read_interval) {`

lum1n0us (approved):
LGTM

samples/simple/wasm-apps/sensor.c
```diff
@@ -6,46 +6,76 @@
...
-        sensor_config(sensor, 0, 0, 0);
+    if (NULL != sensor1) {
+        sensor_config(sensor1, 0, 0, 0);
     }
+
+    if (NULL != sensor2) {
+        sensor_config(sensor2, 0, 0, 0);
+    }
+
+    free(user_data);
```
xujuntwt95329, line 77: Had better check if `user_data` is NULL

wenyongh: Done

PR #920, opened 2021-12-29
Zephyr update

wallento:
I think the Zephyr instructions were not fully up-to-date anymore. Add info about espressif toolchain too.
Add ESP32C3 (RISC-V) and Particle Argon, more boards are possible, but the script doesn't scale well, can be improved in future.
Add Dockerfile to kickstart a Zephyr development environment as it can be rather confusing for new users..

### DIFF
--- a/doc/build_wamr.md
+++ b/doc/build_wamr.md
@@ -303,15 +303,24 @@ WAMR provides some features which can be easily configured by passing options to
 
 Zephyr
 -------------------------
-You need to download the Zephyr source code first and embed WAMR into it.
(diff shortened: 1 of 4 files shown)


wenyongh:
@wallento Thanks a lot for the update!

goodluckrhj:
@wallento If further update in the plan?
For reason that zephyr has updated its west commands since zephyr v3.20, the following command
`west espressif install`
has been deprecated[ West espressif install not working #51797 ](https://github.com/zephyrproject-rtos/zephyr/issues/51797). 
So, in the file build_wamr.md, "ESPRESS_TOOLCHAIN_PATH" which is  "~/espressif/tools/zephyr" will not be created and take no effect.

PR #944, opened 2022-01-08
Enable Windows XIP

wenyongh:
Enable Windows XIP
And add more strict checks for wamrc to report error when the input file
is same with output file, or the input file is AOT file but not wasm file.


xujuntwt95329 (commented):
LGTM

PR #960, opened 2022-01-14
fix pointer unchecked issue in thread-mgr

xujuntwt95329:


core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -354,19 +354,21 @@ wasm_cluster_spawn_exec_env(WASMExecEnv *exec_env)
         return NULL;
     }
 
+    if (module_inst) {
```
wenyongh, line 357: module_inst is used to get module in Line 347, wasm_exec_env_get_module doesn't check whether module_inst is NULL or not.
Can we change L347 to `wasm_module_t module;`, and add checks for module_inst and module at the beginning, L353:
```C
if (!module_inst || !(module = wasm_exec_env_get_module(module_inst))) {
    return NULL;
}
```
And here and L378, no need to check module_inst again

PR #962, opened 2022-01-15
Fix thread manager issues

wenyongh:
Fix the issue that possibly joining a thread after it was detached which
might causes joining hang, by adding wait_count to indicate whether
a thread needs to detach itself or not when it exits.
Add checks for the input exec_env for cluster's join/detach/cancel thread.


core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -625,16 +625,68 @@ wasm_cluster_set_debug_inst(WASMCluster *cluster, WASMDebugInstance *inst)
...
+        node = bh_list_first_elem(&cluster->exec_env_list);
+
+        while (node) {
+            if (node == exec_env) {
+                return true;
+            }
+            node = bh_list_elem_next(node);
+        }
+
+        cluster = bh_list_elem_next(cluster);
```
xujuntwt95329, line 646: should we search the whole cluster list?

Seems we should only search the current cluster, otherwise we may get some thread of other wasm instance?

wenyongh: But how to search the current cluster? Seems that we can only get it from exec_env->cluster, right? But if the exec_env has been removed from cluster and destroyed, accessing exec_env will cause crash.

How about, when a cluster is found that it contains the exec_env, then we check whether exec_env->cluster == cluster again? To ensure that the exec_env belongs to that cluster?

xujuntwt95329: > How about, when a cluster is found that it contains the exec_env, then we check whether exec_env->cluster == cluster again? To ensure that the exec_env belongs to that cluster?

Sounds reasonable, and I think an `assert` is enough because this should never happen, when it really happens, that means the whole system is in an unstable state

xujuntwt95329 (commented):
LGTM

PR #963, opened 2022-01-17
Correct stack base calculation on Mac and NuttX

no1wudi:


core/shared/platform/common/posix/posix_thread.c
```diff
@@ -327,8 +327,10 @@ os_thread_get_stack_boundary()
     if ((addr = (uint8 *)pthread_get_stackaddr_np(self))) {
         stack_size = pthread_get_stacksize_np(self);
         if (stack_size > max_stack_size)
-            addr -= max_stack_size;
-        else
+            stack_size = max_stack_size;
+
+        /* Check whether stack_addr is the base or end of the stack */
+        if (addr > (uint8 *)&stack_size)
```
wenyongh, line 333: Here `(addr > (uint8 *)&stack_size)` means that the addr is the base of the stack, right?
An issue is that when `addr <= (uint8 *)&stack_size)`, and line 330 is hit, then if APP_THREAD_STACK_SIZE_MAX is set to a small value, e.g. 1M, then the addr returned should be invalid, can we change L328 to L334 like:
```C
        stack_size = pthread_get_stacksize_np(self);

        /* Check whether stack_addr is the base or end of the stack,
            change it to the base if it is the end of stack */
        if (addr <= (uint8 *)&stack_size)
            addr = addr + stack_size;

        if (stack_size > max_stack_size)
            stack_size = max_stack_size;

         addr -= stack_size;
```

core/shared/platform/common/posix/posix_thread.c
```diff
@@ -326,10 +326,16 @@ os_thread_get_stack_boundary()
 #elif defined(__APPLE__) || defined(__NuttX__)
     if ((addr = (uint8 *)pthread_get_stackaddr_np(self))) {
         stack_size = pthread_get_stacksize_np(self);
+
+        /* Check whether stack_addr is the base or end of the stack,
+            change it to the base if it is the end of stack */
```
wenyongh, line 331: Code indent, had better make `change` be aligned with `Check` :)

PR #978, opened 2022-01-21
Use LLVM new pass manager for wamrc

wenyongh:
Use LLVM new pass manager for wamrc to replace the legacy pass manger,
so as gain better performance and reduce the compilation time.
Reference links:
- https://llvm.org/docs/NewPassManager.html
- https://blog.llvm.org/posts/2021-03-26-the-new-pass-manager

And add an option to use the old legacy pm mode when building wamrc:
cmake .. -DWAMR_BUILD_LLVM_LEGACY_PM=1

For JIT mode, keep unchanged as it only runs several function passes and using
new pass manager will increase the compilation time.

And refactor the codes of applying LLVM passes.


no1wudi:
> 
> And add an option to use the old legacy pm mode when building wamrc: cmake .. -DWAMR_BUILD_LLVM_LEGACY_PM=1
> 

Why keep this option ? It does'n make sense to wamrc's user if it's works fine.

wenyongh:
> > And add an option to use the old legacy pm mode when building wamrc: cmake .. -DWAMR_BUILD_LLVM_LEGACY_PM=1
> 
> Why keep this option ? It does'n make sense to wamrc's user if it's works fine. And it wouldn't introduce potential compatibility issues ?

Just to keep an option for user to tune the performance, for most cases, the performance of llvm new pass manager is better, but for some cases, it may be not as good as before. The original mode applies some function passes, vector related passes and lto passes, the performance is also relatively good, but the compilation time is really long.
Seems that the original APIs are also available, and there should be no compatibility issue currently, at least we can use them in some time. If the LLVM new version doesn't support it, we should be able to use the llvm version macro (e.g. LLVM_VERSION_MAJOR) to control the code.

PR #983, opened 2022-01-25
fix debug thread not create issue

xujuntwt95329:
And fix a double free issue when starting debug server failed

### DIFF
--- a/core/iwasm/common/wasm_application.c
+++ b/core/iwasm/common/wasm_application.c
@@ -105,6 +105,9 @@ wasm_application_execute_main(WASMModuleInstanceCommon *module_inst, int32 argc,
        the actual main function. Directly calling main function
        may cause exception thrown. */
     if ((func = wasm_runtime_lookup_wasi_start_function(module_inst))) {
+#if WASM_ENABLE_DEBUG_INTERP != 0
+        wasm_runtime_start_debug_instance(exec_env);
+#endif
         return wasm_runtime_call_wasm(exec_env, func, 0, NULL);
     }
 #endif /* end of WASM_ENABLE_LIBC_WASI */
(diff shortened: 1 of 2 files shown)


core/iwasm/common/wasm_application.c
```diff
@@ -105,6 +105,9 @@ wasm_application_execute_main(WASMModuleInstanceCommon *module_inst, int32 argc,
        the actual main function. Directly calling main function
        may cause exception thrown. */
     if ((func = wasm_runtime_lookup_wasi_start_function(module_inst))) {
+#if WASM_ENABLE_DEBUG_INTERP != 0
+        wasm_runtime_start_debug_instance(exec_env);
```
wenyongh, line 109: Should check the return value?

wenyongh: And after the following wasm_runtime_call_wasm, shall we destroy the debug instance?

core/iwasm/common/wasm_application.c
```diff
@@ -186,6 +189,10 @@ wasm_application_execute_main(WASMModuleInstanceCommon *module_inst, int32 argc,
             (uint32)wasm_runtime_addr_native_to_app(module_inst, argv_offsets);
     }
 
+#if WASM_ENABLE_DEBUG_INTERP != 0
+    wasm_runtime_start_debug_instance(exec_env);
```
wenyongh, line 193: Same as above

core/iwasm/common/wasm_application.c
xujuntwt95329: This API return the actual bind port number, or 0 if the debug instance not created. We don't check the return value here, because when user execute iwasm w/o `-g={ip}:{port}`, the API will return 0, but we still want the program to be executed

xujuntwt95329: > And after the following wasm_runtime_call_wasm, shall we destroy the debug instance?

We don't need to destroy the debug instance, it will be automatically destroyed when the corresponding cluster is destroyed.

PR #992, opened 2022-01-29
Fix native stack overflow check failed in interpreter

wenyongh:


xingkaiyu (commented):
I think you forget to change the comment after #endif in core/shared/platform/vxworks/platform_internal.h

wenyongh:
> I think you forget to change the comment after #endif in core/shared/platform/vxworks/platform_internal.h

Updated and tried not to modify the platform related files, could you please test again? Thanks.

Issue #993, opened 2022-02-01
MinGW support?


lucianoiam:
Question, not an issue.

I am working on desktop [audio plugins](https://github.com/lucianoiam/hiphop) that run AssemblyScript code for DSP. These plugins are compiled using MinGW on Windows.

Currently Wasmer is used as the engine but find it a little bloated because of Rust and also hit some obscure crashes on Windows. Already spent too much time on failed attempts to fix them. It works great on Linux and Mac though.

WAMR seems more suitable for use in a plugin environment because of its small footprint. Since the plugins rely on the Wasm C API, I have been able to quickly switch to WAMR and the result works great on Mac... however I hit a roadblock when I realized the library is not compiling on MinGW out of the box :/

Went some steps forward but this is starting to exponentially take more time, so it is worth asking:

Will be MinGW ever supported?  has anyone tried it?

lucianoiam:
This would make WAMR more FOSS friendly. It currently does not compile out of the box on MinGW but the missing gaps are minimal. Here is a workaround:

1. Make sure the correct CMake package is installed
```
pacman -R cmake
pacman -S mingw-w64-x86_64-cmake
```

2. Comment out this line in `core/shared/platform/windows/platform_internal.h`
```
#define OS_ENABLE_HW_BOUND_CHECK
```

3. Add `lWs2_32` to the DLL link options in the root CMakeLists.txt
```
target_link_libraries (iwasm_shared ${LLVM_AVAILABLE_LIBS} ${UV_A_LIBS} -lm -ldl -lpthread -lWs2_32)
```

4. Implement `timespec_get()` in `core/shared/platform/windows/win_time.c` which seems to be missing from MinGW. See code pasted below (requires Windows 8+)

5. Run CMake with some flags
```
mkdir build && cd build
cmake .. -G"Unix Makefiles" -DWAMR_BUILD_LIBC_WASI=0 -DWAMR_BUILD_INVOKE_NATIVE_GENERAL=1 -DWAMR_DISABLE_HW_BOUND_CHECK=1
```

6. Enjoy



win_time.c

```C
/*
 * Copyright (C) 2019 Intel Corporation.  All rights reserved.
 * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 */

#include "platform_api_vmcore.h"

// https://patches.dpdk.org/project/dpdk/patch/20210324005008.24705-2-dmitry.kozliuk@gmail.com/

[...]

PR #1000, opened 2022-02-05
Allow compilation on MinGW

lucianoiam:
This is the least invasive patch I could came up with. Note that WASI and some other smallish details are still not supported, yet we have a starting point.

Further discussion at https://github.com/bytecodealliance/wasm-micro-runtime/issues/993


wenyongh (commented):
Hi, thanks for supporting MinGW! Is it a new platform? If yes, how about adding MinGW to `Supported architectures and platforms` list on README.md?

lucianoiam:
Done.

PR #1006, opened 2022-02-14
Fix littlevgl link error issues

wenyongh:
The littlevgl library had changed its name and domain to [LVGL](https://lvgl.io).
See https://blog.lvgl.io/2020-06-01/announcement
We change some names and links accordingly.

Also remove the cloning for tlsf library as it isn't used now.


lum1n0us:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/ATTRIBUTIONS.md

PR #1015, opened 2022-02-21
Refine call native function from AOT code

wenyongh:
When calling native function from AOT code, current implementation is to return
back to runtime to call aot_invoke_native, which calls wasm_runtime_invoke_native
and the latter calls assembly code. We did it before as there may be pointer and
string arguments to check and convert if the native function's registered signature
has character '*' and '$'.
As the built-in native function's signatures can be gotten in compilation time, we
check the pointer/string arguments and convert them into native address in AOT
code, and then invoke the native function directly, so as to improve performance.

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -2464,6 +2464,7 @@ aot_check_app_addr_and_convert(AOTModuleInstance *module_inst, bool is_str,
     else {
         const char *str, *str_end;
 
+        /* The whole string must be in the linear memory */
         str = (const char *)native_addr;
         str_end = (const char *)memory_inst->memory_data_end.ptr;
         while (str < str_end && *str != '\0')
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -2436,8 +2436,8 @@ aot_call_indirect(WASMExecEnv *exec_env, uint32 tbl_idx, uint32 table_elem_idx,
  * and convert the app address into native address
  */
 bool
-aot_check_app_addr_and_convert(AOTModuleInstance *module_inst, bool is_str_arg,
-                               uint32 app_offset, uint32 buf_size,
+aot_check_app_addr_and_convert(AOTModuleInstance *module_inst, bool is_str,
+                               uint32 app_buf_addr, uint32 app_buf_size,
                                void **p_native_addr)
 {
     AOTMemoryInstance *memory_inst = aot_get_default_memory(module_inst);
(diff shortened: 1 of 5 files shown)


core/iwasm/aot/aot_runtime.c
```diff
@@ -2431,6 +2431,55 @@ aot_call_indirect(WASMExecEnv *exec_env, uint32 tbl_idx, uint32 table_elem_idx,
...
 
+/**
+ * Check whether the app address and the buf is inside the linear memory,
+ * and convert the app address into native address
+ */
+bool
+aot_check_app_addr_and_convert(AOTModuleInstance *module_inst, bool is_str_arg,
+                               uint32 app_offset, uint32 buf_size,
+                               void **p_native_addr)
+{
```
lum1n0us, line 2442: how about renaming parameters?
- `is_str_ag` -> `is_str_buf`
- `app_offset` -> `buf_offset` or `buf_mem_offset`
- `p_native_addr` -> `buf_native_addr`

core/iwasm/compilation/aot_emit_function.c
```diff
@@ -518,6 +518,111 @@ check_stack_boundary(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
...
+            return false;
+        }
+        if (!(func = aot_get_func_from_table(comp_ctx, func_ctx->native_symbol,
+                                             func_ptr_type, func_index))) {
+            return false;
+        }
+    }
+    else {
+        if (!(func = LLVMGetNamedFunction(func_ctx->module, func_name))
+            && !(func =
```
lum1n0us, line 580: `&&` -> `||` ?

core/iwasm/aot/aot_runtime.c
wenyongh: The function is somewhat like other APIs:
```
bool
wasm_runtime_validate_app_addr(WASMModuleInstanceCommon *module_inst,
                               uint32 app_offset, uint32 size);
bool
wasm_runtime_validate_native_addr(WASMModuleInstanceCommon *module_inst,
                                  void *native_ptr, uint32 size);
uint32
wasm_runtime_addr_native_to_app(WASMModuleInstanceCommon *module_inst,
                                void *native_ptr);
```
app_xxx means relative address of wasm app, native_xxx mean absolute address of native,
how about changing to:
``` 
bool
aot_check_app_addr_and_convert(AOTModuleInstance *module_inst, bool is_str_buf,
                               uint32 app_buf_addr, uint32 app_buf_size,
                               void **p_native_buf_addr);
```

core/iwasm/compilation/aot_emit_function.c
wenyongh: Should be `&&` like other places, when first time called, LLVMGetNamedFunction returns NULL and LLVMAddFunction adds the function, then when next time called, LLVMGetNamedFunction returns the function added by the first time.

core/iwasm/aot/aot_runtime.c
lum1n0us: got the point now. 

```
bool aot_check_app_addr_and_convert(AOTModuleInstance *module_inst, 
     bool is_str, uint32 app_addr, uint32 size_or_len, void **p_native_addr);
```

core/iwasm/compilation/aot_emit_function.c
lum1n0us: Oh, got it.

PR #1019, opened 2022-02-23
Fix C API mistake leading to double-free / use-after-free in `wasm_module_imports`

wildbook:
Two of the `if` branches when iterating imports in `wasm_module_imports` never set the `module_name` and `name` fields that are later passed as arguments to `wasm_importtype_new`.

This leads to reusing the same pointers for multiple imports, eventually causing double-free and/or use-after-free. It also leads to imports that hit either of those two branches to be returned with incorrect name fields, but compared to the UB caused that's a "minor" issue.

### DIFF
--- a/core/iwasm/common/wasm_c_api.c
+++ b/core/iwasm/common/wasm_c_api.c
@@ -1947,6 +1947,10 @@ wasm_module_imports(const wasm_module_t *module, own wasm_importtype_vec_t *out)
     for (i = 0; i != import_count; ++i) {
         char *module_name_rt = NULL, *field_name_rt = NULL;
 
+        memset(&module_name, 0, sizeof(module_name));
+        memset(&name, 0, sizeof(name));
+        extern_type = NULL;
+
         if (i < import_func_count) {
             wasm_functype_t *type = NULL;
             WASMType *type_rt = NULL;
--- a/core/iwasm/common/wasm_c_api.c
+++ b/core/iwasm/common/wasm_c_api.c
@@ -1974,16 +1974,6 @@ wasm_module_imports(const wasm_module_t *module, own wasm_importtype_vec_t *out)
                 continue;
             }
 
-            wasm_name_new_from_string(&module_name, module_name_rt);
-            if (strlen(module_name_rt) && !module_name.data) {
-                goto failed;
-            }
-
-            wasm_name_new_from_string(&name, field_name_rt);
-            if (strlen(field_name_rt) && !name.data) {
-                goto failed;
-            }
-
             if (!(type = wasm_functype_new_internal(type_rt))) {
                 goto failed;
             }
@@ -2019,16 +2009,6 @@ wasm_module_imports(const wasm_module_t *module, own wasm_importtype_vec_t *out)
 
             if (!module_name_rt || !field_name_rt) {
                 continue;
-            }
-
-            wasm_name_new_from_string(&module_name, module_name_rt);
-            if (strlen(module_name_rt) && !module_name.data) {
-                goto failed;
-            }
-
-            wasm_name_new_from_string(&name, field_name_rt);
-            if (strlen(field_name_rt) && !name.data) {
-                goto failed;
             }
 
             if (!(type = wasm_globaltype_new_internal(val_type_rt,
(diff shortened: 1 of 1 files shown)


core/iwasm/common/wasm_c_api.c
```diff
@@ -2021,6 +2021,16 @@ wasm_module_imports(const wasm_module_t *module, own wasm_importtype_vec_t *out)
                 continue;
             }
 
+            wasm_name_new_from_string(&module_name, module_name_rt);
```
lum1n0us, line 2024: maybe we should avoid copying `module_name` and `field_name` four times by doing it once?

wildbook: I was thinking the same but wasn't familiar enough with the codebase to say if the different order would matter for error handling / performance / something else, so I kept them within the branches figuring someone probably put them there for a reason.

Moving them out seems cleaner to me too if doing so seems good to you.

wenyongh: @wildbook Could you also memset module_name and field_name with 0 before wasm_name_new_from_string(&module_name/name)? If not, when goto failed in some places, the data of module_name and name may be initialized by last iteration and has been used by last time's wasm_importtype_new, and in the failed handling: wasm_byte_vec_delete(&module_name/name) may double free the vector data with wasm_importtype_vec_delete(out).

wildbook: @wenyongh Good eye.

It seems like `wasm_name_new_from_string` already does a [memset with zero](https://github.com/wildbook/wasm-micro-runtime/blob/44b4f2f94cb28f5ca3041ac309a15e38b06f32ff/core/iwasm/common/wasm_c_api.c#L121). That said, I think the current code would still allow for a double-free to happen if an import passes but the next one fails [early](https://github.com/wildbook/wasm-micro-runtime/blob/44b4f2f94cb28f5ca3041ac309a15e38b06f32ff/core/iwasm/common/wasm_c_api.c#L1978), right?

It looks to me like `wasm_byte_vec_delete` would delete the lingering data pointers from the last iteration, just for `wasm_importtype_vec_delete` to delete them again, since the struct is copied in `wasm_importtype_new`.

If I'm correct that behavior was caused by moving the copies out of the if statement, which might explain why they were duplicated in there in the first place.

Would zeroing `module_name`/`name` at the start of the loop where `module_name_rt` and `field_name_rt` are defined make more sense?

I'm familiar with C but I'm not used to writing it nowadays, so I might be wrong.

wildbook: And if so, should I also zero `extern_type` there to prevent the same situation with it?

wenyongh: @wildbook yes, thanks a lot, we should zero module_name/name at the start of the loop, and per my understanding, there are also several other issues:
1. Should also initialize extern_type to NULL at the start of loop, or wasm_externtype_delete(extern_type) at failure handling may delete the data of last iteration
2. No need to check `if (!extern_type) {  continue; }` as extern_type is converted from type and type is checked already, which must not be NULL
3. No need to `wasm_importtype_vec_delete(out)` as out is passed from outside, and should be initialized by outside,
4. In fact the failures are related to each iteration, we had better put the failure handing inside the loop, but not at the end of function.

I upload a patch to fix the them besides your update, could you please help check whether it is better?
[wasm_c_api.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/8123727/wasm_c_api.zip)

wildbook: I've committed the changes mentioned above. If you want them rolled back or done differently, or if I've misunderstood something, feel free to point it out and I'll go make changes.

wildbook: Oh, I missed the last comment, let me take care of that too then.

wildbook: @wenyongh Git doesn't let me apply the patch, but looking through the file I believe that looks good.

There's a change in behavior on failure that I personally like, but since this is implementing a pseudo-standard API that change might not be expected. I tried to quickly look for any mentions of how errors should be handled according to that standard but found nothing, so I'll assume that's fine until someone claims otherwise.

I'm happy with the changes in your patch, so if you're able to feel free to apply it and close the PR :)

lum1n0us: LGTM

wenyongh: > @wenyongh Git doesn't let me apply the patch, but looking through the file I believe that looks good.
> 
> There's a change in behavior on failure that I personally prefer, but since this is implementing a pseudo-standard API that change might not be expected. I tried to quickly look for any mentions of how errors should be handled according to that standard but found nothing, so I'll assume that's fine until someone claims otherwise.
> 
> I'm happy with the changes in your patch, so if you're able to, feel free to apply it and close this PR :)

Thanks, @wildbook, I created another PR to fix the issue and close this PR.

Issue #991, opened 2022-01-28
compiling with Wcast-align shows a huge number of warnings for THUMB


kfessel:
When compiling with Wcast-align and build traged beeing THUMB a huge number of warnings is shown.

RIOT uses Wcast-align by default i switched it of for the libwarm.a generation.

```
core/shared/mem-alloc/ems/ems_gc_internal.h:191:18: warning: cast increases required alignment of target type [-Wcast-align]
  191 |                ? (hmu_normal_node_t *)((uint8 *)node + node->next_offset)
```
<details>
```
-- The C compiler identification is GNU 9.2.1
-- The CXX compiler identification is GNU 9.2.1
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/arm-none-eabi-gcc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/arm-none-eabi-g++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- The ASM compiler identification is GNU
-- Found assembler: /usr/bin/arm-none-eabi-gcc
-- Build Configurations:
     Build as target THUMB
     CMAKE_BUILD_TYPE 
     WAMR Interpreter enabled
     WAMR AOT disabled
     WAMR JIT disabled
     Libc builtin enabled
     Libc WASI disabled
     Fast interpreter disabled
     Multiple modules disabled
     Reference types disabled
-- Configuring done
-- Generating done
-- Build files have been written to: .../examples/wasm/bin/nucleo-f767zi/pkg-build/wamr
make -C .../examples/wasm/bin/nucleo-f767zi/pkg-build/wamr 
Scanning dependencies of target wamr
[  3%] Building C object CMakeFiles/wamr.dir.../wamr/core/shared/platform/riot/riot_platform.c.obj
[  7%] Building C object CMakeFiles/wamr.dir.../wamr/core/shared/platform/riot/riot_thread.c.obj
[ 10%] Building C object CMakeFiles/wamr.dir.../wamr/core/shared/platform/riot/riot_time.c.obj
[ 14%] Building C object CMakeFiles/wamr.dir.../wamr/core/shared/mem-alloc/ems/ems_alloc.c.obj
[...]

PR #1020, opened 2022-02-23
Use -Werror for strict code checking

no1wudi:


build-scripts/runtime_lib.cmake
```diff
@@ -105,8 +105,7 @@ endif ()
 
 ####################### Common sources #######################
 if (NOT MSVC)
-    set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu99 -ffunction-sections -fdata-sections \
-                                         -Wall -Wno-unused-parameter -Wno-pedantic")
+    set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu99 -ffunction-sections -fdata-sections -Werror")
```
wenyongh, line 108: Will it be too strict for all platforms and targets? For example, #991 reports a lot of warnings for thumb when compiling with Wcast-align. Shall we only add '-Werror' for some common used platforms and targets, e.g. Linux/MacOS/Windows and x86-64?

no1wudi: Yes we can consider it, in fact I'm working with clang on my local machine to cleanup these warnings, may not works on all platforms.

I'll try to let this PR to pass CI first.

xujuntwt95329: I agree with @wenyongh , had better apply this strict check for some specific platforms. 

I remember that there are also a lot of warnings on Windows(MSVC).

no1wudi: It's not a problem since the best way is to use -Werror for all target, if we can't achive this, as a fallback we can apply it for partial target only.

PR #1025, opened 2022-03-01
Implement instantiation and classic interpreter for GC feature

wenyongh:


core/iwasm/common/wasm_application.c
```diff
@@ -657,23 +715,54 @@ wasm_application_execute_func(WASMModuleInstanceCommon *module_inst,
...
+#endif /* WASM_ENABLE_GC == 0 && WASM_ENABLE_REF_TYPES != 0 */
+            default:
             {
-                uint64 *v = (uint64 *)(argv1 + k);
-                os_printf("<0x%016" PRIx64 " 0x%016" PRIx64 ">:v128", *v,
-                          *(v + 1));
-                k += 4;
-                break;
+#if WASM_ENABLE_GC != 0
+                if (wasm_is_type_reftype(type->types[type->param_count + j])) {
```
no1wudi, line 722: Should we extract this code piece (and also in new other code) to a function for better readability ?

wenyongh: Yes, currently we just enable the patch to pass most spec cases, some codes may look a little ugly and need to be polished in the future.

PR #1030, opened 2022-03-08
Update CI files

wenyongh:
Update CI files:
- add another apt repository for spec test on x86-32 as some packages
  cannot be downloaded with the Github default repository
- trigger the CI file when it is modified


lum1n0us:
LGTM

PR #1016, opened 2022-02-22
Import Fast JIT framework

wenyongh:

PR #1032, opened 2022-03-08
Merge main into dev/fast_jit

wenyongh:


lum1n0us:
Shall we also merge #1016? Now it is a huge patch with ~1w line

wenyongh:
> Shall we also merge #1016? Now it is a huge patch with ~1w line

Yes, I will run CI again, if all workflows run OK, let's merge it.

PR #1036, opened 2022-03-09
Implement part of Berkeley Socket API for libc-wasi

wenyongh:
Refer to [ Networking API design](https://github.com/WebAssembly/WASI/issues/370)
and [feat(socket): berkeley socket API v2](https://github.com/WebAssembly/WASI/pull/459)
- Support the socket API of synchronous mode, including `socket/bind/listen/accept/send/recv/close/shutdown`,
    the asynchronous mode isn't supported yet.
- Support adding `--addr-pool=<pool1,pool2,..>` argument for command line to identify the valid ip address range.
- Add the sample and update the document


core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -2943,3 +3125,88 @@ fd_prestats_destroy(struct fd_prestats *pt)
...
+        return false;
+    }
+
+    if (!(next = wasm_runtime_malloc(sizeof(struct addr_pool)))) {
+        return false;
+    }
+
+    next->next = NULL;
+    next->mask = mask;
+    if (os_socket_inet_network(addr, &next->addr) != BHT_OK) {
```
yamt, line 3154: is there any reason you prefer to use legacy inet_network over, say, inet_addr/inet_aton?

PR #1042, opened 2022-03-10
Support `recvmsg()` and `sendmsg()`

lum1n0us:
Add an example, *send_recv.c*, to test with new socket APIs. Before testing, please uncomment lines in CMakeLists.txt

### DIFF
--- a/samples/socket-api/wasm-src/send_recv.c
+++ b/samples/socket-api/wasm-src/send_recv.c
@@ -17,39 +17,67 @@
 #include <wasi_socket_ext.h>
 #endif
 
+static pthread_mutex_t lock = { 0 };
+static pthread_cond_t cond = { 0 };
+static bool server_is_ready = false;
+
 void *
 run_as_server(void *arg)
 {
-    int sock = -1;
+    int sock = -1, on = 1;
     struct sockaddr_in addr = { 0 };
     int addrlen = 0;
     int new_sock = -1;
-    char *buf = "it is an example of socket";
-    struct iovec iov = { .iov_base = buf, .iov_len = strlen(buf) + 1 };
-    struct msghdr msg = { .msg_iov = &iov, .msg_iovlen = 1 };
+    char *buf[] = {
+        "The stars shine down", "It brings us light", "Light comes down",
+        "To make us paths",     "It watches us",      "And mourns for us",
+    };
+    struct iovec iov[] = {
+        { .iov_base = buf[0], .iov_len = strlen(buf[0]) + 1 },
+        { .iov_base = buf[1], .iov_len = strlen(buf[1]) + 1 },
+        { .iov_base = buf[2], .iov_len = strlen(buf[2]) + 1 },
+        { .iov_base = buf[3], .iov_len = strlen(buf[3]) + 1 },
+        { .iov_base = buf[4], .iov_len = strlen(buf[4]) + 1 },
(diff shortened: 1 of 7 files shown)


lum1n0us:
@wenyongh  what kind of copyright declaration should I add into [the example](samples/socket-api/wasm-src/send_recv.c) ?

lum1n0us:
@wenyongh here is the question,  the new API `socketpair()`
> On Linux, the only supported domains for this call are AF_UNIX (or synonymously, AF_LOCAL) and AF_TIPC (since Linux 4.12)

And WAMR socket supporting only covers `AF_INET`, which means either we expand the socket address family or we reject the requirement. 

BTW, the example *send_recv.c* will be failed in linux:
```
$ clang-13 -o send_recv ../wasm-src/send_recv.c
$ ./send_recv
socketpair err: Operation not supported
```

wenyongh:
> @wenyongh here is the question, the new API `socketpair()`
> 
> > On Linux, the only supported domains for this call are AF_UNIX (or synonymously, AF_LOCAL) and AF_TIPC (since Linux 4.12)
> 
> And WAMR socket supporting only covers `AF_INET`, which means either we expand the socket address family or we reject the requirement.
> 
> BTW, the example _send_recv.c_ will be failed in linux:
> 
> ```
> $ clang-13 -o send_recv ../wasm-src/send_recv.c
> $ ./send_recv
> socketpair err: Operation not supported
> ```

Please change AF_INET to AF_LOCAL.

lum1n0us:
In that case (`socketpair()` and `AF_UNIX`), we are looking at questions and they are all connected with rights control.
- `socketpair()` creates two ***unnamed*** sockets. It required a new way to announce "unnamed socket are allowed" since neither *--addr-pool* nor *--dir* covers it now
- because `socketpair()` brings `AF_UNIX` into WAMR socket APIs supporting range, are we going to expand `AF_UNIX` to `socket()` also?  It should be OK to use '--dir' to cover "pathname sockets". (because pathname includes files and dirs, it's better to rename `--dir` to `--path`). Just need to figure out two options to cover *unnamed sockets* (created by `socketpair()`) and *abstract sockets*.

Reference:
[AF_UNIX](https://man7.org/linux/man-pages/man7/unix.7.html)
[socketpair)(https://man7.org/linux/man-pages/man2/socketpair.2.html)

lum1n0us:
For the record, we plan to bypass the requirement of `socketpair` and only support `recvmsg()` and `sendmsg()`.

samples/socket-api/wasm-src/send_recv.c
```diff
@@ -0,0 +1,207 @@
...
+
+    if (pthread_cond_init(&cond, NULL)) {
+        perror("Initialize condition failed");
+        ret = EXIT_FAILURE;
+        goto DESTROY_MUTEX;
+    }
+
+    if (pthread_create(&cs[0], NULL, run_as_server, NULL)) {
+        perror("Create a server thread failed");
+        goto DESTROY_COND;
```
wenyongh, line 189: Should set `ret = EXIT_FAILURE;` before return?

samples/socket-api/wasm-src/send_recv.c
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    if (pthread_create(&cs[1], NULL, run_as_client, NULL)) {
+        perror("Create a client thread failed");
+        goto DESTROY_COND;
```
wenyongh, line 194: Same as above

PR #1044, opened 2022-03-13
Fix broken wamrc build on MinGW

lucianoiam:
- Exclude libc_wasi because it currently does not build on MinGW
- Exclude -ldl switch otherwise link fails


wamr-compiler/CMakeLists.txt
```diff
@@ -181,7 +181,9 @@ include (${SHARED_DIR}/utils/shared_utils.cmake)
 include (${SHARED_DIR}/utils/uncommon/shared_uncommon.cmake)
 include (${IWASM_DIR}/libraries/thread-mgr/thread_mgr.cmake)
 include (${IWASM_DIR}/libraries/libc-builtin/libc_builtin.cmake)
-include (${IWASM_DIR}/libraries/libc-wasi/libc_wasi.cmake)
+if (NOT MINGW)
+  include (${IWASM_DIR}/libraries/libc-wasi/libc_wasi.cmake)
```
wenyongh, line 185: Hi, what is the compilation error of libc wasi on MinGW, is it easy to fix? Just wondering why llvm can be built successfully, while libc-wasi gets failed?

wenyongh: Hi, the issue was introduced by adding libc-wasi to wamrc compilation,
and there is also compile error on Windows platform. In Windows we should
build libc-uvwasi but not libc-wasi.

I uploaded another PR #1046 to fix these issues, including wamrc compile
error in MinGW as mentioned in your patch. Could you please help check
whether it works on MinGW? It yes, we may merge it. Thanks.

lucianoiam: That makes sense. #1046 fix these issues and wamrc compiles nicely on MinGW after applying it. Looks good to merge, thank you!

lucianoiam:
Superseded by https://github.com/bytecodealliance/wasm-micro-runtime/pull/1046

PR #1049, opened 2022-03-17
Focus on IF block

lum1n0us:
Since `basic_block_else` is NULL, it meets a crash if there is a
IF block without a else branch. Like:

``` wat
(func (export "params-id") (param i32) (result i32)
  (i32.const 1)
  (if (param i32) (result i32) (local.get 0)
    (then)
  )
)
```

Consider the ELSE block will be created lazily, focus on
`basic_block_entry" here.

### DIFF
--- a/core/iwasm/fast-jit/fe/jit_emit_control.c
+++ b/core/iwasm/fast-jit/fe/jit_emit_control.c
@@ -128,8 +128,7 @@ push_jit_block_to_stack_and_pass_params(JitCompContext *cc, JitBlock *block,
     JitReg value;
     uint32 i, param_index, cell_num;
 
-    if (block->label_type == LABEL_TYPE_BLOCK
-        || (block->label_type == LABEL_TYPE_IF && !cond)) {
+    if (cc->cur_basic_block == basic_block) {
         /* Reuse the current basic block and no need to commit values,
            we just move param values from current block's value stack to
            the new block's value stack */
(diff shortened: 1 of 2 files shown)


core/iwasm/fast-jit/fe/jit_emit_control.c
```diff
@@ -584,16 +584,9 @@ jit_compile_op_block(JitCompContext *cc, uint8 **p_frame_ip,
...
-                        cc, block, block->basic_block_entry, value))
-                    goto fail;
-            }
-            else {
-                if (!push_jit_block_to_stack_and_pass_params(
-                        cc, block, block->basic_block_else, value))
-                    goto fail;
-            }
+            if (!push_jit_block_to_stack_and_pass_params(
+                    cc, block, block->basic_block_entry, value))
```
wenyongh, line 588: Should also modify L195, change `if (basic_block == block->basic_block_entry)` to:
`if (block->wasm_code_else)`, if has else branch, we should generate `BNE cond, basic_block_entry, basic_block_else`, if no, we should generate `BNE cond, basic_block_entry, basic_block_end`?

lum1n0us: That is the part I am not sure about it yet. Let me take more time to dig into it.

lum1n0us:
And there is another problem: after `load_block_params()`, shall we also push return values from the previous block ?

wenyongh:
> And there is another problem: after `load_block_params()`, shall we also push return values from the previous block ?

Yes, that's really an issue, I have tried to fix it in another PR:
https://github.com/bytecodealliance/wasm-micro-runtime/pull/1050/files#diff-16c33ead1da22f9948b953ef64d866ef0996bd1e82286d0cb4dee785c98d07aa

See load_block_results added, and the handle_op_end which calls load_block_results. Have tested some cases, but normally we should carefully check the codes and test more cases.

PR #1050, opened 2022-03-17
Implement part of codegen

wenyongh:


core/iwasm/fast-jit/fe/jit_emit_control.c
```diff
@@ -384,13 +449,18 @@ handle_op_end(JitCompContext *cc, uint8 **p_frame_ip)
 
         SET_BUILDER_POS(block->basic_block_end);
 
+        /* Pop block and load block results */
+        block = jit_block_stack_pop(&cc->block_stack);
+        if (!load_block_results(cc, block)) {
```
lum1n0us, line 454: L414 `if(!block->incoming_insn_for_end_bb)` also needs to transfer return values from the *if block* or *else block* to the outer block.

wenyongh: Seems no need to change the jit_frame, just move result values from popped block's value stack into previous block's value stack?

lum1n0us: Yep, only return values . No `basic_block`, No modification of `jit_frame`.

lum1n0us: But still, need to handle the special case of `LABE_TYPE_FUNCTION`

lum1n0us:
```suggestion
        if (block->label_type == LABEL_TYPE_FUNCTION) {
          handle_func_return(cc, block);
          SET_BB_END_BCIP(cc->cur_basic_block, *p_frame_ip - 1);
        }
        else {
          if (!load_block_result(cc, block) {
        }
```

core/iwasm/fast-jit/jit_ir.def
```diff
@@ -167,8 +167,9 @@ INSN(LOOKUPSWITCH, LookupSwitch, 1, 0)
 
 /* Call and return instructions */
 INSN(CALLNATIVE, VReg, 2, 1)
-INSN(CALLBC, Reg, 3, 0)
-INSN(RETURNBC, Reg, 1, 0)
+INSN(CALLBC, Reg, 3, 2)
+INSN(RETURNBC, Reg, 3, 0)
+INSN(RETURN, Reg, 1, 0)
```
lum1n0us, line 172: what is the difference between `CALLNATIVE` and `CALLBC`? and between `RETURNBC` and `RETURN` ?

wenyongh: CALLNATIVE is to call a native function, including runtime APIs (e.g., jit_set_exception_with_id) and native APIs (e.g., libc-builtin, libc-wasi, etc.), we must use `call` instruction to call that native function, and pass the arguments/results according the calling convention of host, e.g., ABI of Linux/Windows.

CALLBC means `call bytecode`, it is to call the wasm function, currently we only call the jited code of wasm function, in the future, we might return back to interpreter to call the wasm function, if it hasn't been compiled into jited code. Note that we use `jmp` instruction but not `call` instruction to jump to the jited code, and the arguments/results are passed by interpreter stack frame.

RETURNBC is to return to the caller from current jited code, if the caller is a wasm function, we jump to the next instruction of caller's `jmp` instruction (used by CALLBC) to let caller continue to run. If the caller is interpreter, we jump to the global code block `jit_globals->return_to_interp_from_jitted` to return to interpreter.

RETURN is used when exception was thrown, and we want to return to interpreter directly.

PR #1052, opened 2022-03-21
Partialy Implement GC allocator and collector.

linzang:
Implement the collector side of the GC.
Mainly changes in ems alloctor.

This patch does not include any change that required to be add
for root set enumeration.

basic tests in Sample has been tested to be work normally.

More runtime work need to change in future for fully support of wasm GC

Co-authored-by: Lin Zang <linzang@tencent.com>
Co-authored-by: Bin Liao <buddyliao@tencent.com>


core/iwasm/common/wasm_memory.c
```diff
@@ -147,23 +147,70 @@ wasm_runtime_free_internal(void *ptr)
...
+#if 0
     if (size == 0) {
         LOG_WARNING("warning: wasm_runtime_malloc with size zero\n");
         /* At lease alloc 1 byte to avoid malloc failed */
         size = 1;
     }
 
     return wasm_runtime_malloc_internal(size);
+#endif
+    return wasm_gc_malloc(size);
```
wenyongh, line 159: Should not change the original implementation, the allocation mode for runtime is initialized by wasm_runtime_init/full_init, it can be initialized as: allocate runtime memory from a global byte buffer, from os_malloc, from user defined malloc function.

core/iwasm/common/wasm_memory.c
```diff
@@ -147,23 +147,70 @@ wasm_runtime_free_internal(void *ptr)
...
 }
 
 void *
 wasm_runtime_realloc(void *ptr, unsigned int size)
 {
+#if 0
     return wasm_runtime_realloc_internal(ptr, size);
+#endif
```
wenyongh, line 167: Same as above, normally wasm_runtime_realloc should work and is required by loader and wasm_enlarge_memory.

core/iwasm/common/wasm_memory.c
```diff
@@ -147,23 +147,70 @@ wasm_runtime_free_internal(void *ptr)
...
+    return NULL;
 }
 
 void
 wasm_runtime_free(void *ptr)
+{
+#if 0
+  wasm_runtime_free_internal(ptr);
+#endif
```
wenyongh, line 176: Same as above

core/iwasm/common/wasm_memory.c
```diff
@@ -147,23 +147,70 @@ wasm_runtime_free_internal(void *ptr)
...
+
+#if WASM_ENABLE_GC != 0
+static inline void *
+wasm_gc_malloc_internal(unsigned int size)
+{
+    return mem_allocator_malloc_with_gc(pool_allocator, size);
+}
+ 
+void *
+wasm_gc_malloc(unsigned int size)
```
wenyongh, line 188: Could we change the function prototype to `void *wasm_runtime_gc_malloc(void *heap_handle, unsigned int size)`, my understanding is that we should not reuse the global heap's pool allocator, but we should initialize another global gc heap (as the global heap might be not created if developer initializes runtime with other allocation mode) or initialize gc heap for each module instance.

wenyongh: And like wasm_module_malloc, we can directly call `mem_allocator_malloc_with_gc(heap_handle, size);`

core/iwasm/common/wasm_memory.c
```diff
@@ -147,23 +147,70 @@ wasm_runtime_free_internal(void *ptr)
...
+    }
+    mem_allocator_free_with_gc(pool_allocator, ptr);
+}
+
+void
+wasm_gc_free(void *ptr)
 {
     wasm_runtime_free_internal(ptr);
 }
+#endif // WASM_GC_MANUALLY
```
wenyongh, line 215: Could you use `/* */` to comment the code, here it is C source and we normally use `/* */`

core/iwasm/include/wasm_export.h
```diff
@@ -225,6 +225,26 @@ wasm_runtime_realloc(void *ptr, unsigned int size);
 WASM_RUNTIME_API_EXTERN void
 wasm_runtime_free(void *ptr);
 
+#if WASM_ENABLE_GC != 0
```
wenyongh, line 228: No need to add WASM_ENABLE_GC in this header file, as developer may include this file and link iwasm with libiwasm.so, and doesn't add macro `WASM_ENABLE_GC` in the make file.

core/iwasm/include/wasm_export.h
```diff
@@ -225,6 +225,26 @@ wasm_runtime_realloc(void *ptr, unsigned int size);
...
+/**
+ * Allocate memory from object heap.
+ *
+ * @param size bytes need to allocate
+ *
+ * @return the pointer to memory allocated
+ */
+WASM_RUNTIME_API_EXTERN void *
+wasm_gc_malloc(unsigned int size);
```
wenyongh, line 237: Change to wasm_runtime_gc_malloc(void *heap_handle, unsigned size)?

core/iwasm/include/wasm_export.h
```diff
@@ -225,6 +225,26 @@ wasm_runtime_realloc(void *ptr, unsigned int size);
...
+
+#if WASM_GC_MANUALLY != 0
```
wenyongh, line 239: Same as above

core/shared/gc/ems_gc.h
```diff
@@ -1,15 +1,7 @@
 /*
- * Copyright (C) 2019 Intel Corporation.  All rights reserved.
+ * Copyright (C) 2022 Intel Corporation. All rights reserved.
```
wenyongh, line 2: Don't rename the folder, the folder structure was discussed and defined before, had better not change it as some users may manually write the Makefile (but not use cmake), and it is not convenient for them to update the Makefile.

core/shared/gc/ems_hmu.c
```diff
@@ -0,0 +1,86 @@
...
+	bh_assert(!(tot_size & 7));
+	bh_assert(hmu_get_ut(hmu) != HMU_VO || hmu_get_size(hmu) >= tot_size);
+
+	prefix = (gc_object_prefix_t *)(hmu + 1);
+	suffix = (gc_object_suffix_t *)((gc_uint8*)hmu + tot_size - OBJ_SUFFIX_SIZE);
+	prefix->file_name = file_name;
+	prefix->line_no = line_no;
+	prefix->size = tot_size;
+	for(i = 0;i < GC_OBJECT_PREFIX_PADDING_CNT;i++)
+	{
```
wenyongh, line 32: Coding style, `{` should be in the end of last line.
Please download clang-format 12.0.0 from https://github.com/llvm/llvm-project/releases/tag/llvmorg-12.0.0, cd <wamr_root_dir>, and run
`clang-format-12 -i --style=file <path_to_the_source_file>` to format the source codes

PR #1053, opened 2022-03-21
Fixed libc_wasi not working in spawned exec_env

hangedfish:
In `thread_manager.c:wasm_cluster_spawn_exec_env`, create a new module instance but not set wasi_ctx, so when the new exec_env call WebAssembly function that used WASI API, and the WASI API functions which in `libc_wasi_wrapper.c` call `get_wasi_ctx` will get null result and then return `wasi_errno`.

### DIFF
--- a/core/iwasm/libraries/thread-mgr/thread_manager.c
+++ b/core/iwasm/libraries/thread-mgr/thread_manager.c
@@ -393,6 +393,11 @@ wasm_cluster_spawn_exec_env(WASMExecEnv *exec_env)
     wasm_runtime_set_custom_data_internal(
         new_module_inst, wasm_runtime_get_custom_data(module_inst));
 
+#if WASM_ENABLE_LIBC_WASI != 0
+    WASIContext* wasi_ctx = wasm_runtime_get_wasi_ctx(module_inst);
+    wasm_runtime_set_wasi_ctx(new_module_inst, wasi_ctx);
+#endif
+
     new_exec_env = wasm_exec_env_create_internal(new_module_inst,
                                                  exec_env->wasm_stack_size);
     if (!new_exec_env)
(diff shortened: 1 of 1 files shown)


core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -393,6 +393,11 @@ wasm_cluster_spawn_exec_env(WASMExecEnv *exec_env)
     wasm_runtime_set_custom_data_internal(
         new_module_inst, wasm_runtime_get_custom_data(module_inst));
 
+#if WASM_ENABLE_LIBC_WASI != 0
+    WASIContext* wasi_ctx = wasm_runtime_get_wasi_ctx(module_inst);
```
wenyongh, line 397: Could you please define variable wasi_ctx at the beginning of the function, e.g. after line 365? Here it is C source code, we normally define variables at the beginning of function to avoid compile warnings reported by some compilers.

xujuntwt95329 (commented):
LGTM

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -362,6 +362,9 @@ wasm_cluster_spawn_exec_env(WASMExecEnv *exec_env)
     wasm_module_inst_t module_inst = get_module_inst(exec_env);
     wasm_module_t module;
     wasm_module_inst_t new_module_inst;
+#if WASM_ENABLE_LIBC_WASI != 0
+    WASIContext* wasi_ctx;
```
wenyongh, line 366: Coding guidelines check failed, should define as:
`WASIContext *wasi_ctx;` but not `WASIContext* wasi_ctx;`

PR #1062, opened 2022-03-25
Reset JitCompContext

lum1n0us:
Or else, all secondary IRs and machine code will append to the content
of first function

### DIFF
--- a/core/iwasm/fast-jit/jit_compiler.c
+++ b/core/iwasm/fast-jit/jit_compiler.c
@@ -152,42 +152,15 @@ jit_compiler_compile(WASMModule *module, uint32 func_idx)
 bool
 jit_compiler_compile_all(WASMModule *module)
 {
-    JitCompContext *cc;
-    char *last_error;
-    bool ret = true;
     uint32 i;
 
-    /* Initialize compilation context.  */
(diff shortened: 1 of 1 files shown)


core/iwasm/fast-jit/jit_compiler.c
```diff
@@ -161,12 +161,12 @@ jit_compiler_compile_all(WASMModule *module)
     if (!(cc = jit_calloc(sizeof(*cc))))
         return false;
 
-    if (!jit_cc_init(cc, 64)) {
-        jit_free(cc);
-        return false;
-    }
-
     for (i = 0; i < module->function_count; i++) {
+        if (!jit_cc_init(cc, 64)) {
```
lum1n0us, line 165: It is better if there is a function like `jit_cc_reset()` to reset `block_stack`, `jit_frame`, every `JitBasicBlock`.

Or else, a combination of `jit_cc_destroy()` and `jit_cc_init()` will do the work.

@wenyongh

lum1n0us: my bad, `jit_cc_destroy` fits. No necessary to create a `jit_cc_reset()`

PR #1064, opened 2022-03-30
Implement BR_TABLE

lum1n0us:


core/iwasm/fast-jit/fe/jit_emit_control.c
```diff
@@ -550,8 +550,12 @@ handle_op_else(JitCompContext *cc, uint8 **p_frame_ip)
         /* Clear value stack, restore param values and
            start to translate the else branch. */
         jit_value_stack_destroy(&block->value_stack);
+        /* Reload block parameters */
+        if (!load_block_params(cc, block)) {
+            return false;
+        }
```
wenyongh, line 556: Why moving these lines here? Here we load block params again for else branch as similar to if branch, moving here causes loading again for if branch and missing for else branch?

core/iwasm/fast-jit/fe/jit_emit_control.c
```diff
@@ -917,43 +832,37 @@ jit_compile_op_br(JitCompContext *cc, uint32 br_depth, uint8 **p_frame_ip)
...
     }
 
+    /* append IF to current basic block */
     POP_I32(cond);
 
     jit_frame = cc->jit_frame;
     cur_basic_block = cc->cur_basic_block;
     gen_commit_values(jit_frame, jit_frame->lp, jit_frame->sp);
-    clear_values(jit_frame);
+    SET_BB_END_BCIP(cur_basic_block, *p_frame_ip - 1);
```
wenyongh, line 865: Current block hasn't ended yet, it is br_if but not br, the following opcodes also belongs to current basic block.

core/iwasm/fast-jit/fe/jit_emit_control.c
```diff
@@ -966,65 +875,11 @@ jit_compile_op_br_if(JitCompContext *cc, uint32 br_depth, uint8 **p_frame_ip)
...
-        }
-        if (!jit_block_add_incoming_insn(block_dst, insn)) {
-            jit_set_last_error(cc, "add incoming insn failed");
-            goto fail;
-        }
-        SET_BB_END_BCIP(cc->cur_basic_block, *p_frame_ip - 1);
-    }
-
-    SET_BUILDER_POS(cur_basic_block);
+    if (!handle_op_br(cc, br_depth, p_frame_ip))
```
wenyongh, line 878: handle_op_br generates `JMP` instruction again, should not call it.
Had better restore the code of this function if there are no issues found.

For below wast file, the original code runs OK, but fails with this patch:

```wast
(func $1 (type $0)
    (result i32)
    i32.const 5678
    (block $l (param i32) (result i32)
      i32.const 1234
      i32.add
      i32.const 1
      br_if $l
      drop
      i32.const 666
    )
  )
```
The IRs generated by this patch:
```C
L2:
    ; PREDS(L0)
    ; BEGIN_BCIP=0x0027 END_BCIP=0x0033
    ADD             i7, 0x162e, 0x4d2
    STI32           i7, I0, 0x90
    CMP             i6, 0x1, 0x0
    BNE             i6, L3, VOID         => No ending IRs for this block
    ; SUCCS(L3 VOID)

L3:
    ; PREDS(L2)
    ; BEGIN_BCIP=0x0033 END_BCIP=0x0038
    JMP             L4
    LDI32           i8, I0, 0x90            => Unused IRs after JMP
    STI32           0x29a, I0, 0x90
    JMP             L4                           => Duplicated JMP
    ; SUCCS(L4)

L4:
```

core/iwasm/fast-jit/jit_frontend.c
```diff
@@ -683,7 +693,7 @@ jit_compile_func(JitCompContext *cc)
                 for (i = 0; i <= br_count; i++)
                     read_leb_uint32(frame_ip, frame_ip_end, br_depths[i]);
 
-                if (!jit_compile_op_br_table(cc, br_depths, br_count,
+                if (!jit_compile_op_br_table(cc, br_depths, br_count + 1,
```
wenyongh, line 696: The original code is ported from core/iwasm/compilation/aot_compiler.c:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/compilation/aot_compiler.c#L282-L283
Could we not change it and make it similar to aot_compiler.c? We can increase br_count in jit_emit_control.c

lum1n0us: Sure we can keep following AOT code. 

Just a little wired if the length parameter (br_count) is actually longer than the size of the array(br_depths).

core/iwasm/fast-jit/fe/jit_emit_control.c
```diff
@@ -1034,147 +889,48 @@ bool
...
-    }
-#endif
+    JitReg value;
+    JitInsn *insn;
+    uint32 i = 0;
+    JitOpndLookupSwitch *opnd = NULL;
 
-    POP_I32(value_cmp);
+    /* append LOOKUPSWITCH to current basic block */
+    gen_commit_values(cc->jit_frame, cc->jit_frame->lp, cc->jit_frame->sp);
```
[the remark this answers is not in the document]
lum1n0us: Sure.

core/iwasm/fast-jit/fe/jit_emit_control.c
```diff
@@ -917,43 +833,36 @@ jit_compile_op_br(JitCompContext *cc, uint32 br_depth, uint8 **p_frame_ip)
...
         return false;
     }
 
+    /* append IF to current basic block */
     POP_I32(cond);
 
     jit_frame = cc->jit_frame;
     cur_basic_block = cc->cur_basic_block;
     gen_commit_values(jit_frame, jit_frame->lp, jit_frame->sp);
-    clear_values(jit_frame);
```
wenyongh, line 956: Should not remove this line, current basic block's register may be used by other basic block, for example:
```wast
(module
  (func $test (export "test") (param $l0 i32) (result i32)
    i32.const 1
    (block $l (param i32) (result i32)
      i32.const 2
      i32.add
      local.get $l0
      local.get $l0
      br_if $l
      drop
      local.get $l0
      i32.add
    )
  )
)
```
The IRs in L3 generated are:
```bash
L3:
    ; PREDS()
    ; BEGIN_BCIP=0x002f END_BCIP=0x002f
    ADD             I21, I0, 0x94L
    STI32           i8, I21, 0x0                  => unknown/unloaded i8 used
    JMP             L4
    ; SUCCS(L4)
```
The running results:
```bash
iwasm -f test test.wasm 1
# return 3, the right result should be 1
```

core/iwasm/fast-jit/fe/jit_emit_control.c
```diff
@@ -1034,147 +889,51 @@ bool
...
-    ) {
-        if (!(jit_emit_exception(cc, func_ctx, EXCE_INTEGER_OVERFLOW, false,
-                                 NULL, NULL))) {
-            goto fail;
-        }
-        return jit_handle_next_reachable_block(cc, func_ctx, p_frame_ip);
+    POP_I32(value);
+
+    /* append LOOKUPSWITCH to current basic block */
+    gen_commit_values(cc->jit_frame, cc->jit_frame->lp, cc->jit_frame->sp);
```
wenyongh, line 903: Should also `clear_values()` after gen_commit_values.

PR #1071, opened 2022-04-01
fix constant space overflow issue

xujuntwt95329:


core/iwasm/interpreter/wasm_loader.c
```diff
@@ -4933,6 +4959,64 @@ wasm_loader_ctx_reinit(WASMLoaderContext *ctx)
     return true;
 }
 
+static void
+wasm_loader_emit_float64(WASMLoaderContext *ctx, float64 value)
```
wenyongh, line 4963: How about merging `wasm_loader_emit_float64/float32/uint64` into a single function 

```C
wasm_loader_emit_const(WASMLoaderContext *ctx, void *value, bool is_32_bit)
{
        bh_memcpy_s(ctx->p_code_compiled,
                    ctx->p_code_compiled_end - ctx->p_code_compiled, &value,
                    is_32_bit ? sizeof(uint32) : sizeof(uint64));
        ctx->p_code_compiled += is_32_bit ? sizeof(uint32) : sizeof(uint64);
        ...
}
```

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -5428,7 +5515,24 @@ wasm_loader_get_const_offset(WASMLoaderContext *ctx, uint8 type, void *value,
...
+            bytes_to_increase = 2;
+        }
+        else {
+            bytes_to_increase = 1;
+        }
+
+        /* The max cell num of const buffer is 32768 since the valid index range
+         * is -32768 ~ -1. Return an invalid index 0 to indicate the buffer is
+         * full */
+        if (ctx->const_cell_num + bytes_to_increase > 32768) {
```
wenyongh, line 5531: Had better use `ctx->const_cell_num >= INT32_MAX - bytes_to_increase` to check integer overflow, as recommended by SDL guideline

core/iwasm/interpreter/wasm_loader.c
xujuntwt95329: Done, thanks.

I remain emit_uint32 unchanged since it has been used in several places.

core/iwasm/interpreter/wasm_loader.c
xujuntwt95329: Done, thanks

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -4591,7 +4591,7 @@ wasm_loader_ctx_destroy(WASMLoaderContext *ctx)
 }
 
 static WASMLoaderContext *
-wasm_loader_ctx_init(WASMFunction *func)
+wasm_loader_ctx_init(WASMFunction *func, char *error_buf, uint32 error_buf_size)
 {
     WASMLoaderContext *loader_ctx =
         loader_malloc(sizeof(WASMLoaderContext), NULL, 0);
```
wenyongh, line 4597: Use `loader_malloc(sizeof(WASMLoaderContext), error_buf, error_buf_size)`, and no need to add line 4642. Same as following loader_malloc.

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -4623,6 +4623,14 @@ wasm_loader_ctx_init(WASMFunction *func)
               loader_malloc(loader_ctx->const_buf_size, NULL, 0)))
         goto fail;
 
+    if ((func->local_cell_num >= INT16_MAX)
+        || (func->param_cell_num >= INT16_MAX - func->local_cell_num)) {
```
wenyongh, line 4627: Integer overflow of `param_cell_num/ret_cell_num` is checked in `load_type_section`, L605.
Seems that we had better also check integer overflow of param_cell_num in `load_function_section`, L1956:
```C
if (wasm_get_cell_num(func->local_types, func->local_count)  >= INT16_MAX)
```
And here, had better convert INT16_MAX to uint32?
```C
if ((func->local_cell_num >= INT16_MAX)
        || (func->param_cell_num >= (uint32)INT16_MAX - func->local_cell_num)) {
```

xujuntwt95329: > Integer overflow of param_cell_num/ret_cell_num is checked in load_type_section, L605.

Yes, in L605 we check whether the cell num bigger than `UINT16_MAX (65535)`, but here we need to check `local_cell_num + param_cell_num` < `INT16_MAX(32767)`

> Seems that we had better also check integer overflow of param_cell_num in load_function_section, L1956:

I thinks you mean `local_cell_num` rather than `param_cell_num`?

xujuntwt95329: Seems we don't need to check in L1956, the `param_cell_num` and `local_cell_num` are both defined as `uint16`, so they won't be large than UINT16_MAX. 

Here we check whether `param_cell_num + local_cell_num < INT16_MAX`, this limit is from fast interpreter rather than wasm spec, because the upper bound of dynamic space is 32767

core/iwasm/interpreter/wasm_loader.c
xujuntwt95329: Done, thanks~

core/iwasm/interpreter/wasm_loader.c
wenyongh: > 



> > Integer overflow of param_cell_num/ret_cell_num is checked in load_type_section, L605.
> 
> Yes, in L605 we check whether the cell num bigger than `UINT16_MAX (65535)`, but here we need to check `local_cell_num + param_cell_num` < `INT16_MAX(32767)`

Yes they are different.

> 
> > Seems that we had better also check integer overflow of param_cell_num in load_function_section, L1956:
> 
> I thinks you mean `local_cell_num` rather than `param_cell_num`?

Yes, should be local_cell_num.

xujuntwt95329: Done.

I didn't add check in L1956 since `local_cell_num` is already uint16, it won't be large than UINT16_MAX

wenyongh: > Seems we don't need to check in L1956, the `param_cell_num` and `local_cell_num` are both defined as `uint16`, so they won't be large than UINT16_MAX.
> 

L1956 converts uint32 into uint16, note that return type of wasm_get_cell_num() is uint32, but local_cell_num type is uint16, had better check the integer overflow like L605

> Here we check whether `param_cell_num + local_cell_num < INT16_MAX`, this limit is from fast interpreter rather than wasm spec, because the upper bound of dynamic space is 32767

Yes, here we add extra check for `param_cell_num + local_cell_num < INT16_MAX`

xujuntwt95329: > L1956 converts uint32 into uint16, note that return type of wasm_get_cell_num() is uint32, but local_cell_num type is uint16, had better check the integer overflow like L605

Got it, thanks!
I'll check it.

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -4591,38 +4591,46 @@ wasm_loader_ctx_destroy(WASMLoaderContext *ctx)
...
-              loader_malloc(loader_ctx->const_buf_size, NULL, 0)))
+    if (!(loader_ctx->const_buf = loader_malloc(loader_ctx->const_buf_size,
+                                                error_buf, error_buf_size)))
         goto fail;
 
+    if (func->param_cell_num >= (uint32)INT16_MAX - func->local_cell_num) {
+        set_error_buf(error_buf, error_buf_size,
+                      "fast interpreter offset overflow");
+        wasm_loader_ctx_destroy(loader_ctx);
+        return NULL;
```
wenyongh, line 4631: Seems it can also be changed into `goto fail`, it depends on you.

core/iwasm/interpreter/wasm_loader.c
xujuntwt95329: Added check in L1956.

> And here, had better convert INT16_MAX to uint32?

INT16_MAX should be converted to `int32`, otherwise the final result will be treated as `uint32`, which may cause integer underflow

wenyongh: OK

PR #1072, opened 2022-04-02
preserve execution memory for debug instance

xujuntwt95329:


core/iwasm/libraries/debug-engine/debug_engine.c
```diff
@@ -1194,24 +1211,30 @@ wasm_debug_instance_mmap(WASMDebugInstance *instance, uint32 size,
...
-    /* TODO: malloc in wasi libc maybe not be thread safe, we hope LLDB will
-             always ask for memory when threads stopped */
-    offset = wasm_runtime_module_malloc((wasm_module_inst_t)module_inst, size,
-                                        &native_addr);
-    if (!offset)
+    if ((uint64)(instance->exec_mem_info.current_pos
+                 - instance->exec_mem_info.start_offset + size)
+        <= (uint64)instance->exec_mem_info.size) {
+        offset = instance->exec_mem_info.current_pos;
+        instance->exec_mem_info.current_pos += size;
```
wenyongh, line 1222: Directly adding the position instead of allocating memory, there is potential issue in multi-thread, should we initialize the preserved memory as an heap, and allocate memory from the heap?

core/iwasm/libraries/debug-engine/debug_engine.h
```diff
@@ -10,6 +10,8 @@
 #include "gdbserver.h"
 #include "thread_manager.h"
 
+#define DEBUG_EXECUTION_MEMORY_SIZE 0x85000
```
wenyongh, line 13: Could we define it in core/config.h, and use `#ifndef..#define..#endif`? Since in Embedded system, the memory may be limit, we had better let developer be able to adjust it.

xujuntwt95329: Yes, I moved it into `config.h`

BTW, this value is required by lldb, so user shouldn't change it, I leave a comment in config.h as well

core/iwasm/libraries/debug-engine/debug_engine.c
xujuntwt95329: The 'allocation' can only happen when the execution is stopped, so only the debug control thread of this debug instance can access this struct, that's why I add the position simply and without lock here.

Of course, we can use an allocator, but according to my observation, the lldb will never request to free this space, so I suggest to use the simplest solution, what's your opinion on this?

wenyongh: Yes, that should be OK.

PR #1075, opened 2022-04-07
Implement op_call for fast-jit

wenyongh:
Translate WASM_OP_CALL into JIT IR in the frontend, and translate
JIT_OP_CALLBC and JIT_OP_CALLNATIVE in the backend.
For calling wasm native API, simply call wasm_interp_call_func_native
to reduce the complexity.
And fix some issues, including wasm loader, frontend, register allocator,
and code gen.


core/iwasm/fast-jit/fe/jit_emit_function.c
```diff
@@ -4,11 +4,229 @@
...
+                outs_off -= 8;
+                GEN_INSN(STF64, value, cc->fp_reg, NEW_CONST(I32, outs_off));
+                break;
+            default:
+                bh_assert(0);
+                break;
+        }
+    }
+
+    /* Commit sp as the callee may use it to store the results */
```
lum1n0us, line 114: Does callee use the same `JitFrame` or a new one?

core/iwasm/fast-jit/jit_regalloc.c
```diff
@@ -247,12 +247,9 @@ rc_alloc_spill_slot(RegallocContext *rc, JitReg vreg)
 
 found:
     /* Now, i is the first slot for vreg.  */
-#if 0 /* TODO: check the spill */
-  if (rc->cc->interp_frame_size + (i + stride) * 4
-      > rc->cc->total_frame_size)
-    /* No frame space for the spill area.  */
-    return 0;
-#endif
+    if ((i + stride) * 4 > rc->cc->spill_cache_size)
```
lum1n0us, line 250: not sure about it. The original logic is to use `rc->spill_slots` as memory. If tend to use `spill_cache_size` in `WASMInterpFrame`, maybe we should also modify the other

core/iwasm/fast-jit/jit_regalloc.c
```diff
@@ -386,9 +383,8 @@ collect_distances(RegallocContext *rc, JitBasicBlock *basic_block)
 static JitReg
 offset_of_spill_slot(JitCompContext *cc, JitReg slot)
 {
-    /* TODO: check the spill */
-    return jit_cc_new_const_I32(
-        cc, /*cc->interp_frame_size + jit_cc_get_const_I32 (cc, slot) * 4*/ 0);
+    return jit_cc_new_const_I32(cc, cc->spill_cache_offset
```
lum1n0us, line 386: it may should longer than 16

core/iwasm/fast-jit/jit_regalloc.c
```diff
@@ -542,8 +538,7 @@ allocate_hreg(RegallocContext *rc, JitReg vreg, JitInsn *insn, int distance)
     if (hreg_num == 0)
     /* Unsupported hard register kind.  */
     {
-        /* TODO: how to set error */
-        /*jit_set_error (JIT_ERROR_UNSUPPORTED_HREG);*/
+        jit_set_last_error(rc->cc, "unsupported hard register num");
```
lum1n0us, line 541: it should mean there is no hardware register in this `kind`.

BTW, if no hardware register, we should spill all, right?

core/iwasm/fast-jit/fe/jit_emit_function.c
lum1n0us, line 114: To be clarified: the callee and the caller are using the one operand stack in `JitFrame` or not?

core/iwasm/fast-jit/jit_regalloc.c
wenyongh: Yes, we may change the error message.
If all hardware registers are used, we should spill part of them out to spill area, and allocate them for new virtual registers, and in some time, spill them in.

If no hardware registers are configured, e.g. no int64 registers in x86-32, register allocator cannot work if there are JIT IRs using i64 registers. Here is to check this situation.
And normally in this case, the codegen lower pass (lower_cg) should translate int64 related IRs into int32 related IRs, e.g. (convert one i64 IR `ADD I8, I9, I10` into two or more i32 IRs, e.g. `ADD ixx, ixx, ixx`)

core/iwasm/fast-jit/jit_regalloc.c
wenyongh: Same as above, the original code is:
```C
return jit_cc_new_const_I32(
        cc, cc->interp_frame_size + jit_cc_get_const_I32 (cc, slot) * 4);
```
I just replace `cc->interp_frame_size` to `cc->spill_cache_offset`

core/iwasm/fast-jit/jit_regalloc.c
wenyongh: Here the original logic is to check whether there is enough free space in the spill area:
```C
  if (rc->cc->interp_frame_size + (i + stride) * 4
      > rc->cc->total_frame_size)
    /* No frame space for the spill area.  */
    return 0;
```
The spill area resides in the end of the stack frame, from cc->interp_frame_size to cc->total_frame_size:
```C
/**
 * Calculate the size of interpreter area of frame of a method.
 *
 * @param all_cell_num number of all cells including local variables
 * and the working stack slots
 *
 * @return the size of interpreter area of the frame
 */
static inline unsigned
jeff_interp_interp_frame_size (unsigned all_cell_num)
{
  return align_uint (offsetof (JeffInterpFrame, lp) + all_cell_num * 5, 4);
}

/**
 * Calculate the total frame size of a method, including the Java
 * frame and the preserved space for JITed code.
 *
 * @param all_cell_num number of all cells including local variables
 * and the working stack slots
 *
 * @return the total frame size of the method
 */
static inline unsigned
jeff_interp_total_frame_size (unsigned all_cell_num)
{
  /* TODO: reduce the size after optimizing the regalloc of JIT.  */
  return offsetof (JeffInterpFrame, lp) + all_cell_num * 8 + 4 * 16;
}
```
In current design, we reserve a space inside stack frame but not in the end for the register allocator to spill. The start off set is cc->spill_cache_offset, and the size is cc->spill_cache_size. My understanding is that the logic is similar to the logic of original code.

lum1n0us: But seems like in `rc_alloc_spill_slot`, the search part, `for (i = 0; i < rc->spill_slot_num; i += stride)` is looking for enough space for different values, accroding to `get_reg_stride`. Every `JitReg` is a uint32, why using `stride` instead of a fixed size?

core/iwasm/fast-jit/jit_regalloc.c
lum1n0us: I mean `spill_cache` is a `uint32` array with 16 elements only, it might not be long enough.

lum1n0us:
LGTM 👍

PR #1087, opened 2022-04-12
IR for load/store integers

lum1n0us:


core/iwasm/fast-jit/fe/jit_emit_memory.c
```diff
@@ -5,18 +5,131 @@
...
+
+    GEN_INSN(CMP, cc->cmp_reg, offset1, memory_boundary);
+    if (!jit_emit_exception(cc, EXCE_OUT_OF_BOUNDS_MEMORY_ACCESS, JIT_OP_BGEU,
+                            cc->cmp_reg, 0)) {
+        goto fail;
+    }
+
+    /* seek */
+    memory_data = get_memory_data_reg(cc->jit_frame, mem_idx);
+    maddr = jit_cc_new_reg_I64(cc);
```
wenyongh, line 73: Should be jit_cc_new_reg_ptr(cc)

core/iwasm/fast-jit/fe/jit_emit_memory.c
```diff
@@ -5,18 +5,131 @@
...
+    GEN_INSN(ADD, maddr, memory_data, offset1);
```
wenyongh, line 74: Here it is only OK for 64-bit platform, in 32-bit, maddr and memory_data is 32-bit, and offset1 should be converted in 32-bit (I64TOI32) before using it.

In fact in AOT compilation, we don't extend memarg.offset + maddr to uint64, but check (1) whether memarg.offset + maddr is integer overflow, (2) whether the adding result is no larger than memory_boundary. Extending to 64bit, addin, checking and thrinking to 32bit again may introduce a lot instructions and degrade the performance.

core/iwasm/fast-jit/fe/jit_emit_memory.c
```diff
@@ -5,18 +5,131 @@
...
+
+    maddr =
+        check_and_seek(cc, addr, NEW_CONST(I64, offset), NEW_CONST(I64, bytes));
+    if (!maddr) {
+        goto fail;
+    }
+
+    value = jit_cc_new_reg_I32(cc);
+    if (sign) {
+        GEN_INSN(LDI32, value, maddr, NEW_CONST(I32, 0));
```
wenyongh, line 96: `bytes` argument is ignored, should use `LDI32`, `LDI16`, `LDU16`, `LDI8`, `LDU8`

core/iwasm/fast-jit/fe/jit_emit_memory.c
```diff
@@ -5,18 +5,131 @@
...
+
+    maddr =
+        check_and_seek(cc, addr, NEW_CONST(I64, offset), NEW_CONST(I64, bytes));
+    if (!maddr) {
+        goto fail;
+    }
+
+    value = jit_cc_new_reg_I64(cc);
+    if (sign) {
+        GEN_INSN(LDI64, value, maddr, NEW_CONST(I32, 0));
```
wenyongh, line 124: Same as above

core/iwasm/fast-jit/fe/jit_emit_memory.c
```diff
@@ -36,13 +149,43 @@ bool
...
+    POP_I32(value);
+    POP_I32(addr);
+
+    maddr =
+        check_and_seek(cc, addr, NEW_CONST(I64, offset), NEW_CONST(I64, bytes));
+    if (!maddr) {
+        goto fail;
+    }
+
+    GEN_INSN(STI32, value, maddr, NEW_CONST(I32, 0));
```
wenyongh, line 163: Same as above, should handle `bytes`

core/iwasm/fast-jit/fe/jit_emit_memory.c
```diff
@@ -5,18 +5,131 @@
...
+        }
+        default:
+        {
+            memory_boundary = get_memory_data_end_reg(cc->jit_frame, mem_idx);
+            break;
+        }
+    }
+
+    GEN_INSN(CMP, cc->cmp_reg, offset1, memory_boundary);
+    if (!jit_emit_exception(cc, EXCE_OUT_OF_BOUNDS_MEMORY_ACCESS, JIT_OP_BGEU,
```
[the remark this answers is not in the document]
wenyongh: Another issue is that when the linear memory size is 0, memory_boundary got is also 0, here the checking will run failed. In aot compilation, the solution is to check whether the memory's initial page count is 0, if yes, adding another check before L65: check whether memory_size (here we can use memory_boundary) is 0, if yes, throw `out of bounds`. If the initialize page count isn't 0, no need to add the extra check, so as to improve performance.

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
```diff
@@ -1242,9 +1244,12 @@ convert_imm_i32_to_r_u64(x86::Assembler &a, int32 reg_no, int32 data)
  * @return true if success, false otherwise
  */
 static bool
-convert_r_i32_to_r_u64(x86::Assembler &a, int32 reg_no_dst, int32 reg_no_src)
+convert_r_i32_to_r_i64(x86::Assembler &a, int32 reg_no_dst, int32 reg_no_src)
 {
-    return false;
+    if (reg_no_dst != reg_no_src) {
```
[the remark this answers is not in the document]
lum1n0us: redirection to `extend_r32_to_r64`

core/iwasm/fast-jit/fe/jit_emit_memory.c
lum1n0us: It seems there are two different comparison groups in AOT. Each for a different platform. Shall we do the same check or use the new solution w/o expanding to i64.

```
    LLVMValueRef offset_const = I32_CONST(offset);
    POP_I32(addr);

    if (is_target_64bit) {
        if (!(offset_const = LLVMBuildZExt(comp_ctx->builder, offset_const,
                                           I64_TYPE, "offset_i64"))
            || !(addr = LLVMBuildZExt(comp_ctx->builder, addr, I64_TYPE,
                                      "addr_i64"))) {
            aot_set_last_error("llvm build zero extend failed.");
            goto fail;
        }
    }

    /* offset1 = offset + addr; */
    BUILD_OP(Add, offset_const, addr, offset1, "offset1");

    if (!(mem_check_bound =
              get_memory_check_bound(comp_ctx, func_ctx, bytes))) {
        goto fail;
    }

    if (is_target_64bit) {
        /* if (offset1 > mem_check_bound) */
        BUILD_ICMP(LLVMIntUGT, offset1, mem_check_bound, cmp, "cmp");
    }
    else {
        /* if (offset1 < addr) */
        /* Check integer overflow */
        BUILD_ICMP(LLVMIntULT, offset1, addr, cmp1, "cmp1");
        /* if (offset1 > mem_check_bound) */
        BUILD_ICMP(LLVMIntUGT, offset1, mem_check_bound, cmp2, "cmp2");
        BUILD_OP(Or, cmp1, cmp2, cmp, "cmp");
    }
```

wenyongh: Yes, in AOT compilation, for 64-bit platform, it converts memarg.offset and addr into uint64, add them, and compares the result with mem_inst->mem_bound_check_nbytes:
```C
if ((uint64)memarg.offset + (uint64)addr) > mem_inst->mem_bound_check_nbytes)
  throw out_of_bounds exception;
```
For 32-bit platform:
it checks whether (memarg.offset + addr) is overflow, if not, checks (memarg.offset + addr) > mem_bound_check_nbytes):
```
if ((uint32)memarg.offset + (uint32)addr > (uint32)addr) //integer overflow
   throw out_of_bounds exception;
if ((uint32)memarg.offset + (uint32)addr > mem_bound_check_nbytes)
    throw out_of_bounds exception;
```
Note that mem_bound_check_nbytes is uint64 in 64-bit platform, and uint32 in 32-bit platform.

Seems that it is a little complex? How about we just use the solution for 64-bit platform, and refine it for 32-bit in the future? 64-bit solution is also good for 32-bit platform, but the performance is not the best for 32-bit platform.

Or you just want to unify to use the solution for 32-bit platform? Seems that it is also OK, the performance is also good, not sure whether the performance is better than 64-bit solution.

wenyongh: In 64-bit platform, with the solution `if ((uint64)memarg.offset + (uint64)addr) > mem_inst->mem_bound_check_nbytes)`, when checking and getting the final offset, there are 5 instructions:
```bash
extend memarg.offset to 64bit,
extend addr to 64bit
add: offset1 = (uint64)memarg.offset + (uint64)addr)
compare: if (offset1 > mem_bound_check_nbytes)
jxx to throw exception
directly use offset1 to get maddr, as it uint64: maddr = mem_data + offset1
```
with the solution `if ((uint32)memarg.offset + (uint32)addr > (uint32)addr) throw out_of_bounds exception; if ((uint32)memarg.offset + (uint32)addr > mem_bound_check_nbytes) throw out_of_bounds exception;`, when checking and getting the final offset, there are 5 instructions:
```bash
add: offset1 = (uint32)memarg.offset + (uint32)addr
compare, check integer overflow: if (offset1 > (uint32)addr)
jxx to throw exception if integer overflow
compare: if (offset1 > mem_bound_check_nbytes)
jxx to throw exception
extend offset1 to uint64, so maddr = memory_data + (uint64)offset1
```
Not sure whose performance is better, in my memory, the former seems to be better.

core/iwasm/fast-jit/fe/jit_emit_memory.c
```diff
@@ -5,18 +5,317 @@
...
+        JitReg memory_inst, cur_mem_page_count;
+
+        /* if (cur_mem_page_count == 0) goto EXCEPTION */
+        memory_inst = get_memory_inst_reg(cc->jit_frame, mem_idx);
+        cur_mem_page_count = jit_cc_new_reg_I32(cc);
+        GEN_INSN(LDI32, cur_mem_page_count, memory_inst,
+                 NEW_CONST(I32, offsetof(WASMMemoryInstance, cur_page_count)));
+        GEN_INSN(CMP, cc->cmp_reg, cur_mem_page_count, NEW_CONST(I32, 0));
+        if (!jit_emit_exception(cc, EXCE_OUT_OF_BOUNDS_MEMORY_ACCESS,
+                                JIT_OP_BEQ, cc->cmp_reg, 0)) {
```
wenyongh, line 134: The latest argument is `JitBasicBlock *cond_br_else_block`, had better use `NULL` instead of `0`

core/iwasm/fast-jit/fe/jit_emit_memory.c
```diff
@@ -5,18 +5,317 @@
...
+    value = jit_cc_new_reg_I32(cc);
+    switch (bytes) {
+        case 1:
+        {
+            if (sign) {
+                JitReg tmp1 = jit_cc_new_reg_I32(cc);
+                JitReg tmp2 = jit_cc_new_reg_I32(cc);
+                GEN_INSN(LDI8, tmp1, maddr, NEW_CONST(I32, 0));
+                GEN_INSN(SHL, tmp2, tmp1, NEW_CONST(I32, 24));
+                GEN_INSN(SHRS, value, tmp2, NEW_CONST(I32, 24));
```
wenyongh, line 188: Use LDU8 is OK: `GEN_INSN(LDI8, value , maddr, NEW_CONST(I32, 0));`
It is the codegen's responsibility to signed/unsigned extend the memory data to i32/i64.

core/iwasm/fast-jit/fe/jit_emit_memory.c
```diff
@@ -5,18 +5,317 @@
...
+            }
+            else {
+                JitReg tmp1 = jit_cc_new_reg_I32(cc);
+                GEN_INSN(LDU8, tmp1, maddr, NEW_CONST(I32, 0));
+                GEN_INSN(AND, value, tmp1, NEW_CONST(I32, 0x000000FF));
```
wenyongh, line 193: Use `LDU8` is already enough, it will unsigned extend to i32, see:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/dev/fast_jit/core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp#L543

core/iwasm/fast-jit/fe/jit_emit_memory.c
```diff
@@ -5,18 +5,317 @@
...
+            break;
+        }
+        case 2:
+        {
+            if (sign) {
+                JitReg tmp1 = jit_cc_new_reg_I32(cc);
+                JitReg tmp2 = jit_cc_new_reg_I32(cc);
+                GEN_INSN(LDI16, tmp1, maddr, NEW_CONST(I32, 0));
+                GEN_INSN(SHL, tmp2, tmp1, NEW_CONST(I32, 16));
+                GEN_INSN(SHRS, value, tmp2, NEW_CONST(I32, 16));
```
wenyongh, line 204: Same as above, `GEN_INSN(LDI16, value, maddr, NEW_CONST(I32, 0));`

core/iwasm/fast-jit/fe/jit_emit_memory.c
```diff
@@ -5,18 +5,317 @@
...
+            }
+            else {
+                JitReg tmp1 = jit_cc_new_reg_I32(cc);
+                GEN_INSN(LDU16, tmp1, maddr, NEW_CONST(I32, 0));
+                GEN_INSN(AND, value, tmp1, NEW_CONST(I32, 0x0000FFFF));
```
wenyongh, line 209: `GEN_INSN(LDU16, value, maddr, NEW_CONST(I32, 0));`

core/iwasm/fast-jit/fe/jit_emit_memory.c
```diff
@@ -5,18 +5,317 @@
...
+    }
+
+    value = jit_cc_new_reg_I64(cc);
+    switch (bytes) {
+        case 1:
+        {
+            if (sign) {
+                JitReg tmp1 = jit_cc_new_reg_I64(cc);
+                JitReg tmp2 = jit_cc_new_reg_I64(cc);
+                GEN_INSN(LDI8, tmp1, maddr, NEW_CONST(I32, 0));
```
wenyongh, line 256: Same as above, using LDI8/LDU8/LDI16/LDU16/LDI32/LDU32 should be OK, it is the codegen's responsibility to signed/unsigned extend the memory data to i32/i64.

core/iwasm/fast-jit/fe/jit_emit_memory.c
```diff
@@ -5,18 +5,317 @@
...
+                GEN_INSN(AND, value, tmp1, NEW_CONST(I32, (uint64)0xFFFF));
+            }
+            break;
+        }
+        case 4:
+        {
+            if (sign) {
+                JitReg tmp1 = jit_cc_new_reg_I64(cc);
+                JitReg tmp2 = jit_cc_new_reg_I64(cc);
+                GEN_INSN(LDI16, tmp1, maddr, NEW_CONST(I32, 0));
```
wenyongh, line 288: Should be LDI32 and LDU32

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
```diff
@@ -1109,7 +1139,10 @@ mov_r_to_r_f64(x86::Assembler &a, int32 reg_no_dst, int32 reg_no_src)
 static bool
 convert_imm_i32_to_r_i8(x86::Assembler &a, int32 reg_no, int32 data)
 {
-    return false;
+    /* lower quadratic */
```
wenyongh, line 1142: Not known what the comment `lower quadratic` means?

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
```diff
@@ -1227,13 +1260,23 @@ convert_r_i32_to_r_u16(x86::Assembler &a, int32 reg_no_dst, int32 reg_no_src)
...
-convert_imm_i32_to_r_u64(x86::Assembler &a, int32 reg_no, int32 data)
+convert_imm_i32_to_r_i64(x86::Assembler &a, int32 reg_no, int32 data)
 {
-    return false;
+    int64 long_data = 0;
+    if (data & 0x80000000) {
+        long_data = data | 0xFFFFFFFF00000000;
+    }
+    else {
+        long_data = data & 0x00000000FFFFFFFF;
```
wenyongh, line 1270: Directly using `int64 long_data = (int64)data;` is OK, no need to convert it manually.

If not, should be 0x0xFFFFFFFF00000000LL and 0x00000000FFFFFFFFLL, or it might be treated as int32.

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
```diff
@@ -2441,6 +2637,42 @@ static bool
...
+            break;
+        }
+        case SHRU:
+        {
+            data = data1_src >> data2_src;
+            if (data & (1 << 32)) {
+                /* signed */
+                uint32 mask = (1 << (32 - data2_src)) - 1;
+                /* patch sign bits */
+                data = data & mask;
```
wenyongh, line 2661: How about: `data = (uint32)data1_src >> (uint32)data2_src;`

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
```diff
@@ -2479,6 +2711,33 @@ static bool
...
+            a.shr(regs_i32[reg_no1_src], imm);
+            break;
+        }
+        default:
+        {
+            goto fail;
+        }
+    }
+
+    a.mov(regs_i32[reg_no_dst], regs_i32[reg_no1_src]);
```
wenyongh, line 2738: Using `mov_r_to_r(a, JIT_REG_KIND_I32, reg_no_dst, reg_no_src);`, so no need to emit mov instruction when reg_no_dst == reg_no1_src.

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
```diff
@@ -2497,6 +2756,31 @@ static bool
...
+            a.shr(regs_i32[reg_no1_src], regs_i32[reg_no2_src]);
+            break;
+        }
+        default:
+        {
+            goto fail;
+        }
+    }
+
+    a.mov(regs_i32[reg_no_dst], regs_i32[reg_no1_src]);
```
wenyongh, line 2781: Same as above, had better use `mov_r_to_r(a, JIT_REG_KIND_I32, reg_no_dst, reg_no1_src);`

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
```diff
@@ -2515,6 +2799,42 @@ static bool
...
+            break;
+        }
+        case SHRU:
+        {
+            data = data1_src >> data2_src;
+            if (data & (1L << 64)) {
+                /* signed */
+                uint64 mask = (1L << (64 - data2_src)) - 1;
+                /* patch sign bits */
+                data = data & mask;
```
wenyongh, line 2823: How about use `data = (uint64)data1_src >> (uint64)data2_src`

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
```diff
@@ -2553,6 +2873,33 @@ static bool
...
+            a.shr(regs_i64[reg_no1_src], imm);
+            break;
+        }
+        default:
+        {
+            goto fail;
+        }
+    }
+
+    a.mov(regs_i64[reg_no_dst], regs_i64[reg_no1_src]);
```
wenyongh, line 2900: Same as above, `mov_r_to_r(a, JIT_REG_KIND_I64, reg_no_dst, reg_no1_src);`

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
```diff
@@ -2571,6 +2918,31 @@ static bool
 shift_r_r_to_r_i64(x86::Assembler &a, SHIFT_OP op, int32 reg_no_dst,
                    int32 reg_no1_src, int32 reg_no2_src)
 {
+    switch (op) {
+        case SHL:
+        {
+            a.shl(regs_i64[reg_no1_src], regs_i32[reg_no2_src]);
```
wenyongh, line 2924: Same as above, no `shl/sar/shr r64, r64` instructions.

core/iwasm/fast-jit/fe/jit_emit_memory.c
```diff
@@ -5,18 +5,317 @@
...
+    GEN_INSN(U32TOI64, long_addr, addr);
+
+    /* offset1 = offset + long_addr */
+    offset1 = jit_cc_new_reg_I64(cc);
+    GEN_INSN(ADD, offset1, offset, long_addr);
+
+    /* if (offset1 > memory_boundary) goto EXCEPTION */
+    GEN_INSN(CMP, cc->cmp_reg, offset1, memory_boundary);
+    if (!jit_emit_exception(cc, EXCE_OUT_OF_BOUNDS_MEMORY_ACCESS, JIT_OP_BGTU,
+                            cc->cmp_reg, 0)) {
```
wenyongh, line 76: Had better use `cc->cmp_reg, NULL`, the last argument's type is "JitBasicBlock *`

core/iwasm/fast-jit/fe/jit_emit_memory.c
```diff
@@ -5,18 +5,317 @@
...
+    JitReg offset1;
+
+    /* offset1 = offset + addr */
+    offset1 = jit_cc_new_reg_I32(cc);
+    GEN_INSN(ADD, offset1, offset, addr);
+
+    /* if (offset1 < addr) goto EXCEPTION */
+    GEN_INSN(CMP, cc->cmp_reg, offset1, addr);
+    if (!jit_emit_exception(cc, EXCE_OUT_OF_BOUNDS_MEMORY_ACCESS, JIT_OP_BLTU,
+                            cc->cmp_reg, 0)) {
```
wenyongh, line 98: Same as above, `cc->cmp_reg, NULL`

core/iwasm/fast-jit/fe/jit_emit_memory.c
```diff
@@ -5,18 +5,317 @@
...
+        goto fail;
+    }
+
+    /* if (offset1 > memory_boundary) goto EXCEPTION */
+    GEN_INSN(CMP, cc->cmp_reg, offset1, memory_boundary);
+    if (!jit_emit_exception(cc, EXCE_OUT_OF_BOUNDS_MEMORY_ACCESS, JIT_OP_BGTU,
+                            cc->cmp_reg, 0)) {
```
wenyongh, line 105: Same as above

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
```diff
@@ -3398,7 +3772,7 @@ lower_bit(JitCompContext *cc, x86::Assembler &a, BIT_OP op, JitReg r0,
         if (jit_reg_is_const(r1))                                             \
             data1 = jit_cc_get_const_##kind(cc, r1);                          \
         if (jit_reg_is_const(r2))                                             \
-            data2 = jit_cc_get_const_##kind(cc, r2);                          \
+            data2 = jit_cc_get_const_I32(cc, r2);                             \
```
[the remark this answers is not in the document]
lum1n0us:
```
        CHECK_EQKIND(r0, r1);        
        CHECK_KIND(r2, JIT_REG_KIND_I32);   
```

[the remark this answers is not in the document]
wenyongh: The code was ported from Beihai JIT, which doesn't support 64-bit registers in the codegen, seems that we had better change `CHECK_KIND(r2, JIT_REG_KIND_I32);` to `bh_assert(jit_reg_kind(r2) == JIT_REG_KIND_I32 || jit_reg_kind(r2) == JIT_REG_KIND_I64)`, or `CHEKC_EQKIND(r1, r2)`, to make it consistent with wasm opcode? Or we might need to add more conversion in the frontend.

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
```diff
@@ -2497,6 +2756,31 @@ static bool
 shift_r_r_to_r_i32(x86::Assembler &a, SHIFT_OP op, int32 reg_no_dst,
                    int32 reg_no1_src, int32 reg_no2_src)
 {
+    switch (op) {
+        case SHL:
+        {
+            a.shl(regs_i32[reg_no1_src], regs_i32[reg_no2_src]);
```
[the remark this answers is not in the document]
lum1n0us: how about `shl/sar/shr r/m32, imm8`. 2**8 is enough for a i64 or a i32.

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
```diff
@@ -1307,6 +1350,129 @@ convert_r_i32_to_r_f64(x86::Assembler &a, int32 reg_no_dst, int32 reg_no_src)
...
+ * @param a the assembler to emit the code
+ * @param reg_no the no of dst int64 register
+ * @param data the src immediate uint32 data
+ *
+ * @return true if success, false otherwise
+ */
+static bool
+convert_imm_u32_to_r_i64(x86::Assembler &a, int32 reg_no, uint32 data)
+{
+    int64 long_data = data & 0x00000000FFFFFFFF;
```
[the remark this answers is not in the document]
lum1n0us: original `data` is `uint32`. how about changing it to `uint64 long_data = (uint64)data`

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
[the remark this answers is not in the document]
lum1n0us: Similar things at `LD_R_R_R` L3345 and L3359.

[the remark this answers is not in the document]
wenyongh: Yes, we limit type of r2 of LD_R_R_R and ST_R_R_R to i32, as it meets the requirement and is simple for codegen to emit machine code. For wasm load/store opcodes, the offset (r2) of LD/ST IRs generated can be set to i32 0, no need to set it to int64 type. It doesn't impact the performance in fact.

But for wasm 64 bit shift opcodes, the second operand is also 64-bit type, if we limit it to i32 type when generating IR, e.g., translating wasm opcode `i64.shl i64_0, i64_1, i64_2` into JIT IR `I64TOI32 i32_r2, i64_r2` and `SHL i64_r0, i64_r1, i32_r2`, there will be two IRs generated, and for codegen, there may be more instructions generated. It actually impacts the performance. Or is there any good method to improve performance when r2 is in32 type?

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
lum1n0us: `LL` is my mistake. 

not sure  `(int64)data` will let the compiler zero-extend or sign-extend the `data`, and make the code symmetrical, too

wenyongh: In my memory, it depends on the type of data: (int64)(int32)data signed extends i32 to i64, (int64)(uint32)data unsigned extends i32 to i64.
If not very sure, using your original code is also OK, but should add LL suffix.

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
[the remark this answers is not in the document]
wenyongh: It is OK.

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
lum1n0us: a little research shows in the expression `int64 long_data = (int64)data`.

-if `data` is signed, *clang* will use `movsxd` to extend it. *gcc* will use `cdqe`.
-if `data` is unsigned, both will `mov` `data` to a 32bit reg, like *eax*, and then use the whole 64bit reg, like `rax`.

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
lum1n0us: Sure we can. BTW, shall we make sure `int32 data2_src` is always `>=0` ?

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
wenyongh: OK, if the result isn't determinative, using the origin code (with LL) is better.

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
wenyongh: The definition of wasm opcode SHR_U:
ishr_u𝑁(𝑖1, 𝑖2)
• Let 𝑗2 be 𝑖2 modulo 𝑁.
• Return the result of shifting 𝑖1 right by 𝑗2 bits, extended with 0 bits.

In fact here we should mod data_src2 with 32 firstly, but seems that in x86, shift with an operand larger than 32 equals to shift with (operand % 32).

In classic interpreter, we return `data = (uint32)data1_src >> ((uint32)data2_src % 32)`, see:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_interp_classic.c#L2274
and
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_interp_classic.c#L418-L423

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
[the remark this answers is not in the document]
lum1n0us: So, we can use similar ways to handle, I mean let the compiler finish the rest, all immediates' type converting, Including integer to integer, and integer to float, and float to integer.

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
lum1n0us: Just want to make sure. 
Shall we merge all three  `mov_r_to_r` , `mov_r_to_r_i32`, `mov_r_to_r_i64` ?

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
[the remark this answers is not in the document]
wenyongh: Should be OK, as long as the final result is correct for all compilers.

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
wenyongh: No need to merge, how about using mov_r_ro_r_i32/mov_r_to_r_i64/mov_r_to_r_f32/mov_r_to_r_f64，as them are more concise/simple?
How about changing all mov_r_to_r(..JIT_REG_KIND_I32, ...) to mov_r_to_r_i32() in current code?

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
lum1n0us: Sure. will `mod right` in IR first.

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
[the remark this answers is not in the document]
lum1n0us: OK

PR #1090, opened 2022-04-14
protect wasm_valtype_new and wasm_tabletype_new from invalid input

lum1n0us:


core/iwasm/common/wasm_c_api.c
```diff
@@ -467,6 +471,10 @@ wasm_valtype_new(wasm_valkind_t kind)
 {
     wasm_valtype_t *val_type;
 
+    if (!(kind <= WASM_F64) && WASM_ANYREF != kind && WASM_FUNCREF != kind) {
```
wenyongh, line 474: `!(kind <= WASM_F64)` looks strange, had better change to `kind > WASM_F64`

core/iwasm/common/wasm_c_api.c
```diff
@@ -467,6 +471,10 @@ wasm_valtype_new(wasm_valkind_t kind)
 {
     wasm_valtype_t *val_type;
 
+    if (kind > WASM_F64 && WASM_ANYREF != kind && WASM_FUNCREF != kind) {
```
wenyongh, line 474: Shall we add `#if WASM_ENABLE_REF_TYPES != 0` macro control for `WASM_ANYREF`? Like the change below, L792?

PR #1091, opened 2022-04-15
Implement i32/i64 div and rem opcodes translation

wenyongh:


lum1n0us:
Still have a question, do we prevent those `JitReg`, like using `eax`, `edx`, from register allocation by satisfying the requirement of `jit_cc_is_hreg`? It will help us to keep a `JitReg`'s hardware register till the *codegen* phase.

But how to make sure those hardware registers, `edx`, `eax`, won't be used by other `JitReg` when special cases, like `div`, is happening?

PR #1094, opened 2022-04-18
Refine codes and fix several issues

wenyongh:
Add aot relocation for ".rodata.str" symbol to support more cases
Fix some coding style issues
Fix aot block/value stack destroy issue
Refine classic/fast interpreter codes
Clear compile warning of libc_builtin_wrapper.c in 32-bit platform


lum1n0us:
LGTM

PR #1097, opened 2022-04-18
IR for SHL/SHRU/SHRS

lum1n0us:


core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
```diff
@@ -2725,6 +2834,34 @@ static bool
 shift_r_imm_to_r_i64(x86::Assembler &a, SHIFT_OP op, int32 reg_no_dst,
                      int32 reg_no1_src, int64 data2_src)
 {
+    /* SHL/SHA/SHR r/m32, imm8 */
```
wenyongh, line 2837: Should be `r/m64`

core/iwasm/fast-jit/fe/jit_emit_numberic.c
```diff
@@ -472,16 +472,104 @@ jit_compile_op_i64_bitwise(JitCompContext *cc, IntBitwise bitwise_op)
...
+        mod_right = jit_cc_new_reg_I32(cc);
+        GEN_INSN(REM_U, mod_right, right, NEW_CONST(I32, 32));
+    }
+    else {
+        mod_right = jit_cc_new_reg_I64(cc);
+        GEN_INSN(REM_U, mod_right, right, NEW_CONST(I64, 64));
+    }
+
+    if (IS_CONST_ZERO(right)) {
+        res = left;
```
wenyongh, line 503: Code is duplicated with L487-489, do you want to check `IS_CONST_ZERO(mod_right)`?
But it is not const reg. How about change L487-489 to:
```
if (jit_reg_is_const_val(right)) {
    if ((is_i32 && jit_cc_get_const_I32(cc, right) % 32 == 0))
        || (!is_i32 && jit_cc_get_const_I64(cc, right) % 64 == 0)) {
        res = left;
        goto shortcut;
      }
  }
```

wenyongh, line 503: Code is duplicated with L487-489, do you want to check `IS_CONST_ZERO(mod_right)`?
But it is not const reg. How about change L487-489 to:
```
if (jit_reg_is_const_val(right)) {
    if ((is_i32 && jit_cc_get_const_I32(cc, right) % 32 == 0))
        || (!is_i32 && jit_cc_get_const_I64(cc, right) % 64 == 0)) {
        res = left;
        goto shortcut;
      }
  }
```

core/iwasm/fast-jit/fe/jit_emit_numberic.c
```diff
@@ -472,16 +472,104 @@ jit_compile_op_i64_bitwise(JitCompContext *cc, IntBitwise bitwise_op)
...
+    if (is_i32) {
+        res = jit_cc_new_reg_I32(cc);
+#if defined(BUILD_TARGET_X86_64) || defined(BUILD_TARGET_AMD_64)
+        GEN_INSN(MOV, ecx_hreg, mod_right);
+#endif
+    }
+    else {
+        res = jit_cc_new_reg_I64(cc);
+#if defined(BUILD_TARGET_X86_64) || defined(BUILD_TARGET_AMD_64)
+        GEN_INSN(MOV, rcx_hreg, mod_right);
```
wenyongh, line 517: Should add another INSN `MOV ecx_hreg, ecx_hreg` to actually make rcx_hreg usable here, as ecx is freely used as i32 registers.

wenyongh, line 517: Should add another INSN `MOV ecx_hreg, ecx_hreg` to actually make rcx_hreg usable here.

core/iwasm/fast-jit/fe/jit_emit_numberic.c
lum1n0us: Yes. I should check `mod_right` in the second `IS_CONST_ZERO`. But I am a little confused about that the real thing we should do is check the value of `JitReg`, no matter it is a const or not. I mean maybe we should create some `conditional branches` to handle that?

core/iwasm/fast-jit/fe/jit_emit_numberic.c
lum1n0us: Oh, I see. you are worried if some I32 *vr* might use `ecx` in the future.

core/iwasm/fast-jit/fe/jit_emit_numberic.c
wenyongh: My understanding is there is no need to check the value of `JitReg right`, no matter what the `right` value is, we didn't check it in interpreter/AOT. The spec defines:
```
ishl𝑁(𝑖1, 𝑖2)
• Let 𝑘 be 𝑖2 modulo 𝑁.
• Return the result of shifting 𝑖1 left by 𝑘 bits, modulo 2𝑁.
ishl𝑁(𝑖1, 𝑖2) = ibits− 𝑁1(𝑑𝑁 2 −𝑘 0𝑘) (if ibits𝑁(𝑖1) = 𝑑𝑘 1 𝑑𝑁 2 −𝑘 ∧ 𝑘 = 𝑖2 mod 𝑁)
```
It doesn't mention that we should add any check or to throw exception.

core/iwasm/fast-jit/fe/jit_emit_numberic.c
```diff
@@ -472,16 +472,110 @@ jit_compile_op_i64_bitwise(JitCompContext *cc, IntBitwise bitwise_op)
...
+            || (!is_i32 && (0 == jit_cc_get_const_I64(cc, right) % 64))) {
+            res = left;
+            goto shortcut;
+        }
+    }
+
+    /* right modulo N */
+    if (is_i32) {
+        mod_right = jit_cc_new_reg_I32(cc);
+        GEN_INSN(REM_U, mod_right, right, NEW_CONST(I32, 32));
```
wenyongh, line 498: When right is const, don't need to GEN_INSN, just `mod_right = NEW_CONST(I32, jit_jit_cc_get_const_I32(cc, right) % 32)`, shall we add the optimization?

And should fix the code guideline check failure.

core/iwasm/fast-jit/fe/jit_emit_numberic.c
```diff
@@ -472,16 +472,128 @@ jit_compile_op_i64_bitwise(JitCompContext *cc, IntBitwise bitwise_op)
...
+            }
+            else {
+                mod_right = NEW_CONST(I64, right_value);
+            }
+        }
+    }
+    else {
+        if (is_i32) {
+            mod_right = jit_cc_new_reg_I32(cc);
+            GEN_INSN(REM_U, mod_right, right, NEW_CONST(I32, 32));
```
wenyongh, line 515: An improvement: (right mod 32) equals to (right & 0x1F), do want want to change to `GEN_INSN(AND, mod_right, right, NEW_CONST(I32, 0x1F)`?

core/iwasm/fast-jit/fe/jit_emit_numberic.c
```diff
@@ -472,16 +472,128 @@ jit_compile_op_i64_bitwise(JitCompContext *cc, IntBitwise bitwise_op)
...
+        }
+        else {
+            mod_right = jit_cc_new_reg_I64(cc);
+            GEN_INSN(REM_U, mod_right, right, NEW_CONST(I64, 64));
```
wenyongh, line 519: Same as above, `GEN_INSN(AND, mod_right, right, NEW_CONST(I64, 0x3F)`?

core/iwasm/fast-jit/fe/jit_emit_numberic.c
lum1n0us: Sure, `AND` is better

PR #1100, opened 2022-04-19
aot_loader.c: Fix crashing issues in "Refine interp/aot string storage"

yamt:
The following commit [1] had a few issues.

* it looks a wrong byte for the mark
* it didn't work for long strings (>= 0x80 in case of little endian)

This commit fixes it by maintaining another hash instead of
the "0x80" marking.  (The idea from Wenyong Huang)

[1]
```
commit 403a7d3f4fe3eec3092f9fb7b8a71dc93e4589b1
Author: Wenyong Huang <wenyong.huang@intel.com>
Date:   Mon Nov 8 11:27:26 2021 +0800

    Refine interp/aot string storage and emitting (#820)

      same behavior as before

```

### DIFF
--- a/core/iwasm/aot/aot_loader.c
+++ b/core/iwasm/aot/aot_loader.c
@@ -253,6 +253,38 @@ const_str_set_insert(const uint8 *str, int32 len, AOTModule *module,
 }
 
 static char *
+const_inline_str_set_insert(char *c_str, AOTModule *module,
+                            char *error_buf, uint32 error_buf_size)
+{
+    HashMap *set = module->const_inline_str_set;
+    char *value;
+
+    /* Create const string set if it isn't created */
+    if (!set
+        && !(set = module->const_inline_str_set = bh_hash_map_create(
+                 32, false, (HashFunc)wasm_ptr_hash,
+                 (KeyEqualFunc)wasm_ptr_equal, NULL, NULL))) {
+        set_error_buf(error_buf, error_buf_size,
+                      "create const inline string set failed");
+        return NULL;
+    }
+
+    if ((value = bh_hash_map_find(set, c_str))) {
+        bh_assert(c_str == value);
+        return value;
+    }
+
+    if (!bh_hash_map_insert(set, c_str, c_str)) {
+        set_error_buf(error_buf, error_buf_size,
+                      "insert inline string to hash map failed");
+        return NULL;
+    }
+
+    return c_str;
+}
+
+
+static char *
 load_string(uint8 **p_buf, const uint8 *buf_end, AOTModule *module,
             bool is_load_from_file_buf, char *error_buf, uint32 error_buf_size)
 {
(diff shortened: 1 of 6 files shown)


yamt:
> (The idea from Wenyong Huang)

https://github.com/bytecodealliance/wasm-micro-runtime/pull/820#discussion_r851980844

core/iwasm/aot/aot_loader.c
```diff
@@ -261,14 +293,17 @@ load_string(uint8 **p_buf, const uint8 *buf_end, AOTModule *module,
     char *str;
     uint16 str_len;
 
-    CHECK_BUF(p, p_end, 1);
-    if (*p & 0x80) {
+    p = align_ptr(p, 2);
+    if (module->const_inline_str_set != NULL &&
+        (str = bh_hash_map_find(module->const_inline_str_set, p)) != NULL) {
```
wenyongh, line 298: @yamt Thanks for submitting the PR to fix the issue! It should be good, but after reviewing the source code and the AOT file format, I notice that we only repeatedly load the same string in relocation section, that means: in other sections, after the string is adjusted and returned, the loader won't load them again. In relocation section, when reading the string of section_name and symbol_name, the loader may duplicated read them, see:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/aot/aot_loader.c#L2226
and:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/aot/aot_loader.c#L2264
Here it gets the offset of the string and reads it, as the offsets may be same, the string may be loaded repeatedly.
So my understanding is that we can add the check only in `load_relocation_section`:
1. Here in `load_string`, we don't need to check whether the string is in const_inline_str_set or not, we just adjust the string and return it
2. In `load_relocation_section`, create the hash set, and before `read_string(name_addr, buf_end, group->section_name)` and `read_string(symbol_addr, buf_end, relocation->symbol_name)`, check whether `name_addr + 1`/`symbol_addr + 1` is in the hash set, if yes, directly use it, if not, load_string and add the return value in the hash set.
3. Destroy the hash set at the end of `load_relocation_section`. No need add extra hash set in aot module so as not to increase to footprint.

yamt: while i'm not very familiar with the aot file format, (is it documented somewhere?)
i'm not aware of any other places than load_relocation_section where a string can be repeatedly loaded.
so i guess you are right.

if it's ok to rely on that level of specifics of the file format, i guess it's simpler to make validate_symbol_table "adjust" all symbols in a consistent way. that way, the extra hash set is not necessary at all. how do you think?

wenyongh: Yes, that seems good, how about changing `validate_symbol_table` to `validate_and_adjust_symbol_table` or something else? And seems that we should add arguments `bool is_load_from_file_buf` to it, and inside it, handle the string according to is_load_from_file_buf: if it is true, adjust the string, else do nothing. And when read symbol string, if is_load_from_file_buf, use p+1, else use normal read_string with hash set?

yamt: after thinking a bit more, i feel it's simpler to maintain a list of loaded symbols like https://github.com/bytecodealliance/wasm-micro-runtime/pull/1102
how do you think?

wenyongh: Yes, it should be able to fix the issue, and is better and simpler.

wenyongh:
Close this PR as another PR had been merged to fix the issues.

PR #1101, opened 2022-04-19
IRs for bitwise

lum1n0us:


core/iwasm/fast-jit/fe/jit_emit_numberic.c
```diff
@@ -127,6 +127,11 @@ jit_compile_op_i64_popcnt(JitCompContext *cc)
     return false;
 }
 
+#define IS_CONST_ALL_ONE(val, is_i32)                    \
+    (jit_reg_is_const(val)                               \
+     && ((is_i32 && jit_cc_get_const_I32(cc, val) == -1) \
+         || (!is_i32 && jit_cc_get_const_I64(cc, val) == -1)))
```
wenyongh, line 133: Had better use `jit_cc_get_const_I64(cc, val) == -1LL`?

core/iwasm/fast-jit/fe/jit_emit_numberic.c
```diff
@@ -460,16 +465,142 @@ jit_compile_op_i64_arithmetic(JitCompContext *cc, IntArithmetic arith_op,
...
+        goto shortcut;
+    }
+
+    if (IS_CONST_ZERO(right)) {
+        res = left;
+        goto shortcut;
+    }
+
+    if (IS_CONST_ALL_ONE(left, is_i32) || IS_CONST_ALL_ONE(right, is_i32)) {
+        res = is_i32 ? NEW_CONST(I32, -1) : NEW_CONST(I64, -1);
```
wenyongh, line 514: Had better `NEW_CONST(I64, -1LL)`?

core/iwasm/fast-jit/fe/jit_emit_numberic.c
```diff
@@ -460,16 +465,142 @@ jit_compile_op_i64_arithmetic(JitCompContext *cc, IntArithmetic arith_op,
...
+    }
+
+    if (jit_reg_is_const(left) && jit_reg_is_const(right)) {
+        if ((is_i32
+             && (jit_cc_get_const_I32(cc, left)
+                 == jit_cc_get_const_I32(cc, right)))
+            || (!is_i32
+                && (jit_cc_get_const_I64(cc, left)
+                    == jit_cc_get_const_I64(cc, right)))) {
+            res = is_i32 ? NEW_CONST(I32, 0) : NEW_CONST(I64, 0);
```
wenyongh, line 544: If both left and right are const, just return xor or their const value: left_val ^ right_val

lum1n0us: seems all bi-operations of integers can use this improvement. shall we apply it to all ?

wenyongh: Yes, had better apply it to all

lum1n0us:
It is an ugly version but workable. Still looking for a graceful way.

core/iwasm/fast-jit/fe/jit_emit_numberic.c
```diff
@@ -173,9 +209,25 @@ compile_int_mul(JitCompContext *cc, JitReg left, JitReg right, bool is_i32)
...
+            int32 right_val = jit_cc_get_const_I32(cc, right);
+            res = NEW_CONST(I32, left_val * right_val);
+        }
+        else {
+            int64 left_val = jit_cc_get_const_I64(cc, left);
+            int64 right_val = jit_cc_get_const_I64(cc, right);
+            res = NEW_CONST(I64, left_val * right_val);
+        }
+        goto shortcut;
+    }
```
wenyongh, line 224: How about define these code piece to a macro, e.g. `CHECK_CONST(left, right, op)`?

core/iwasm/fast-jit/fe/jit_emit_numberic.c
```diff
@@ -367,6 +419,61 @@ compile_int_div(JitCompContext *cc, IntArithmetic arith_op, bool is_i32,
                                                 right, res);
             }
         }
+
+        if (jit_reg_is_const(left)) {
```
wenyongh, line 423: This code piece can not be executed as the code above has handled all possible branches. How about moving the code into `compile_int_div_no_check`, it doesn't need to check exception again as possible exception checks have been done before calling it.

core/iwasm/fast-jit/fe/jit_emit_numberic.c
wenyongh: Yes, how about adding a function to do that:
```C
bool check_const_and_operate(JitReg left, JitReg right, ArithOp op, JitReg *p_res)
{
    if (jit_reg_is_const(left) && jit_reg_is_const(right)) {
        if (is_i32) {
            int32 left_val = jit_cc_get_const_I32(cc, left);
            int32 right_val = jit_cc_get_const_I32(cc, right);
            int32 res_val;
            switch (op) {
                 case ADD: res_val = left_val + right_val;
                  break;
            res = NEW_CONST(I32, res_val);
        }
        else {
            int64 left_val = jit_cc_get_const_I64(cc, left);
            int64 right_val = jit_cc_get_const_I64(cc, right);
            ...
        }
        *p_res = res;
        return true;
    }
    return false;
}

Comment (lum1n0us) [reply core/iwasm/fast-jit/fe/jit_emit_numberic.c:224]: Sure. working on it.

File: core/iwasm/fast-jit/fe/jit_emit_numberic.c [hunk 6]
```diff
@@ -460,106 +618,392 @@ jit_compile_op_i64_arithmetic(JitCompContext *cc, IntArithmetic arith_op,
...
+    res = is_i32 ? jit_cc_new_reg_I32(cc) : jit_cc_new_reg_I64(cc);
+    GEN_INSN(OR, res, left, right);
+
+shortcut:
+    return res;
+}
+
+DEF_UNI_INT_CONST_OPS(xor)
+{
+    if (left == right) {
```
wenyongh, line 710: Had better compare with the actual const values but not JitReg ?

core/iwasm/fast-jit/fe/jit_emit_numberic.c
```diff
@@ -460,106 +618,392 @@ jit_compile_op_i64_arithmetic(JitCompContext *cc, IntArithmetic arith_op,
...
+        return is_i32 ? NEW_CONST(I32, 0) : NEW_CONST(I64, 0);
```
wenyongh, line 711: If left_val == 0, left xor right = right, do we want to add the optimization?
```C
if (IS_CONST_ZERO(left)) return right;
if (IS_CONST_ZERO(right)) return left;
```

PR #1102, opened 2022-04-19
aot_loader.c: Fix crashing issues in "Refine interp/aot string storage"

yamt:
The following commit [1] had a few issues.

* It looks a wrong byte for the mark
* It didn't work for long strings (>= 0x80 in case of little endian)

This commit fixes it by maintaining a list of loaded symbols to avoid
reading a string repeatedly.

[1]
```
commit 403a7d3f4fe3eec3092f9fb7b8a71dc93e4589b1
Author: Wenyong Huang <wenyong.huang@intel.com>
Date:   Mon Nov 8 11:27:26 2021 +0800

    Refine interp/aot string storage and emitting (#820)

      same behavior as before

```

### DIFF
--- a/core/iwasm/aot/aot_loader.c
+++ b/core/iwasm/aot/aot_loader.c
@@ -261,48 +261,34 @@ load_string(uint8 **p_buf, const uint8 *buf_end, AOTModule *module,
     char *str;
     uint16 str_len;
 
-    CHECK_BUF(p, p_end, 1);
-    if (*p & 0x80) {
-        /* The string has been adjusted */
-        str = (char *)++p;
-        /* Ensure the whole string is in range */
-        do {
-            CHECK_BUF(p, p_end, 1);
-        } while (*p++ != '\0');
+    read_uint16(p, p_end, str_len);
+    CHECK_BUF(p, p_end, str_len);
+
+    if (str_len == 0) {
+        str = "";
+    }
+    else if (p[str_len - 1] == '\0') {
+        /* The string is terminated with '\0', use it directly */
+        str = (char *)p;
+    }
+    else if (is_load_from_file_buf) {
+        /* As the file buffer can be referred to after loading,
+           we use the 2 bytes of size to adjust the string:
+           move string 2 byte backward and then append '\0' */
+        str = (char *)(p - 2);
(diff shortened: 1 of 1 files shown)


core/iwasm/aot/aot_loader.c
```diff
@@ -262,47 +262,35 @@ load_string(uint8 **p_buf, const uint8 *buf_end, AOTModule *module,
     uint16 str_len;
 
     CHECK_BUF(p, p_end, 1);
```
wenyongh, line 264: Can remove L264, as p is checked in read_uint16

core/iwasm/aot/aot_loader.c
```diff
@@ -2076,6 +2065,13 @@ load_relocation_section(const uint8 *buf, const uint8 *buf_end,
         goto fail;
     }
 
+    symbols = loader_malloc(sizeof(*symbols) * symbol_count, error_buf,
+                            error_buf_size);
+    if (symbols == NULL) {
+        goto fail;
+    }
+    memset(symbols, 0, sizeof(*symbols) * symbol_count);
```
wenyongh, line 2073: No need to memset again, loader_malloc already memset the buffer.

wenyongh:
Add @no1wudi to review

core/iwasm/aot/aot_loader.c
yamt: done

core/iwasm/aot/aot_loader.c
yamt: done

core/iwasm/aot/aot_loader.c
```diff
@@ -2076,6 +2063,12 @@ load_relocation_section(const uint8 *buf, const uint8 *buf_end,
         goto fail;
     }
 
+    symbols = loader_malloc(sizeof(*symbols) * symbol_count, error_buf,
```
wenyongh, line 2066: Had better change to `sizeof(*symbols) * (uint64)symbol_count` to avoid integer overflow: `sizeof(*symbols) * symbol_count` may lead to integer overflow firstly in 32-bit system before it is passed to loader_malloc, if symbol_count is large enough.

yamt: done

no1wudi:
LGTM

PR #1103, opened 2022-04-19
aot_runtime.c: Add static assertions for some offset assumptions

yamt:


core/iwasm/aot/aot_runtime.c
```diff
@@ -14,6 +14,21 @@
...
 
+/*
+ * Note: These offsets need to match the values hardcoded in
+ * aot_create_func_context.
+ */
+
+bh_static_assert(offsetof(WASMExecEnv, module_inst) == 2 * sizeof(uintptr_t));
+bh_static_assert(offsetof(WASMExecEnv, argv_buf) == 3 * sizeof(uintptr_t));
+bh_static_assert(offsetof(WASMExecEnv, native_stack_boundary)
+                 == 4 * sizeof(uintptr_t));
```
wenyongh, line 25: How about adding `suspend_flags` also? It may be accessed by aot code in multi-thread mode.
```C
bh_static_assert(offsetof(WASMExecEnv, suspend_flags)
                 == 5 * sizeof(uintptr_t));
```

yamt: i was not aware of it. thanks. done.

PR #1104, opened 2022-04-20
Add more wasm operand stack overflow checks for fast-interp

wenyongh:
And clear compile warnings on Windows


yamt:
is it normal to include a merge commit in a PR?
otherwise lgtm.

wenyongh:
> is it normal to include a merge commit in a PR? otherwise lgtm.

Yes, we will `squash and merge` the PR, the original commits won't be commited. But normally we had better submit a PR with few commits newer than the main branch.

xujuntwt95329 (commented):
LGTM

PR #1109, opened 2022-04-20
Fix build wamrc for xtensa failure

wenyongh:


yamt:
lgtm. why is this still WIP?

a question: do we need to keep old llvm support?
the xtensa llvm repo seems to have releases based on llvm 13 and even 14.

wenyongh:
> lgtm. why is this still WIP?
> 

Yes, it looks good, but not very sure, I need to test more.

> a question: do we need to keep old llvm support? the xtensa llvm repo seems to have releases based on llvm 13 and even 14.

Had better support llvm-11 and llvm-12, some developers may still use them. OK, seems that we should upgrade the script to use xtensa llvm 13.
For llvm-14, we may upgrade WAMR to support it in some time, besides with the toolchain emsdk and wasi-sdk, to use the latest version. Currently we are planning to create a new release, we may upgrade them after that.

core/iwasm/compilation/aot_llvm_extra.cpp
```diff
@@ -478,3 +479,4 @@ aot_apply_llvm_new_pass_manager(AOTCompContext *comp_ctx)
     }
 #endif
 }
+#endif
```
xujuntwt95329, line 482: Had better add a comment:
```suggestion
#endif /* end of LLVM_VERSION_MAJOR >= 12 */
```

xujuntwt95329 (commented):
LGTM

core/iwasm/compilation/aot_llvm_extra.cpp
wenyongh: Done

PR #1112, opened 2022-04-21
Refine opcode br_table for classic interpreter

wenyongh:
Refine opcode br_table for classic interpreter as there may be a lot of
leb128 decoding when the br count is big:
1. Use the bytecode itself to store the decoded leb br depths if each
    decoded depth can be stored with one byte
3. Create br_table cache to store the decode leb br depths if the decoded
    depth cannot be stored with one byte
After the optimization, the class interpreter can access the br depths array
with index, no need to decode the leb128 again.

And fix function record_fast_op() return value unchecked issue in source
debugging feature.


xujuntwt95329 (commented):
LGTM

PR #1114, opened 2022-04-21
IRs for ROTL and ROTR

lum1n0us:


core/iwasm/fast-jit/fe/jit_emit_numberic.c
```diff
@@ -863,129 +863,265 @@ modulo_rhs(JitCompContext *cc, JitReg rhs, bool is_i32)
     return res;
 }
 
+static void
+pre_compile_int_shift(JitCompContext *cc, bool is_i32, JitReg left,
+                      JitReg *left_as_val)
```
wenyongh, line 868: How about change the return type to JitReg, to align with other functions?

PR #1116, opened 2022-04-22
wasm_runtime.c: fix fault modification on data segment.

YoungWenMing:


core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -1333,12 +1333,14 @@ wasm_instantiate(WASMModule *module, bool is_sub_inst, uint32 stack_size,
                 goto fail;
             }
 
-            data_seg->base_offset.u.i32 =
+            base_offset =
                 globals[data_seg->base_offset.u.global_index].initial_value.i32;
         }
+        else {
+            /* check offset */
```
wenyongh, line 1340: The comment is for `if (base_offset > memory_size)`, could you please put it before line 1344?

YoungWenMing: you mean that move this `check offset` comment to line 1343?

PR #1117, opened 2022-04-23
support va_args in libc-builtin

xujuntwt95329:


wamr-sdk/app/libc-builtin-sysroot/include/stdarg.h
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright (C) 2019 Intel Corporation.  All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+
+#ifndef _WAMR_LIBC_STDARG_H
+#define _WAMR_LIBC_STDARG_H
+
+#ifndef __cplusplus
```
wenyongh, line 9: Should be below?
```C
#ifdef __cplusplus
extern "C" {
#endif
```
Or va_list/va_start/.. have been provided in C++ header files?

PR #1118, opened 2022-04-24
Fix calling fault imported function.

YoungWenMing:
Copy WASMFunctionImport to avoid calling unexpected wasm_func_t
after other instances were instantiated based on a same wasm module.

### DIFF
--- a/core/iwasm/interpreter/wasm_runtime.c
+++ b/core/iwasm/interpreter/wasm_runtime.c
@@ -599,6 +599,12 @@ functions_instantiate(const WASMModule *module, WASMModuleInstance *module_inst,
         }
 #endif /* WASM_ENABLE_MULTI_MODULE */
         function->u.func_import = &import->u.function;
+        if (!(function->u.func_import = runtime_malloc(
+                  sizeof(WASMFunctionImport), error_buf, error_buf_size))) {
+            return NULL;
+        }
+        bh_memcpy_s(function->u.func_import, sizeof(WASMFunctionImport),
+                    &import->u.function, sizeof(WASMFunctionImport));
         function->param_cell_num = import->u.function.func_type->param_cell_num;
         function->ret_cell_num = import->u.function.func_type->ret_cell_num;
         function->param_count =


core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -599,6 +599,12 @@ functions_instantiate(const WASMModule *module, WASMModuleInstance *module_inst,
         }
 #endif /* WASM_ENABLE_MULTI_MODULE */
         function->u.func_import = &import->u.function;
+        if (!(function->u.func_import = runtime_malloc(
```
xujuntwt95329, line 602: Hi @YoungWenMing 

By design, the `function->u.func_import` should not change no matter how many instances are created, if you really observe a change on this value, it seems that there are some memory issues, simply copy the data to another place doesn't really solve the problem, and may hide the root cause.

I'd suggest you to use `sanitizer` or `valgrind` to check if there are invalid memory access which modified the `function->u.func_import`

YoungWenMing: > Hi @YoungWenMing
> 
> By design, the `function->u.func_import` should not change no matter how many instances are created, if you really observe a change on this value, it seems that there are some memory issues, simply copy the data to another place doesn't really solve the problem, and may hide the root cause.
> 
> I'd suggest you to use `sanitizer` or `valgrind` to check if there are invalid memory access which modified the `function->u.func_import`

Thanks for replying~
Yes, the `function->u.func_import` does always point to a desired address, but function `wasm_c_api.c:interp_link_func` modifies the content on this address every time when we try to create an instance with one imported function.  However, each instance keep a pointer pointed to the same address kept by the module. It looks like that only the last created instance can access the imported function it wanted (if every instance import its own function). It seems that instances keep their own imported functions is more reasonable? 

We can easily reproduce this bug with this case:
[hello.txt](https://github.com/bytecodealliance/wasm-micro-runtime/files/8549342/hello.txt)
- replace `samples/wasm-c-api/src/hello.c` with this file as `hello.c`
- run `make hello`
- run `./hello`

xujuntwt95329: Hi, thanks for the details and resource, I can reproduce the issue on my side.

I agree with you that instance should keep their own imported functions, and I need to check if there are any other places may have the same problem, such as import globals and AOT mode. So I'll reply to you tomorrow.

YoungWenMing: OK～ ^_^

xujuntwt95329: Hi, @YoungWenMing In our current design, all the import lists are bound to module, including both interpreter and AoT module, so changing the import list to instance may influence a lot of code.

In this circumstance, we'd suggest to create a new module before instantiate it with a new import list.

Is there a specific use case where there must be multiple instance from the same module and with different import lists?

xujuntwt95329: Yes, the implementation in AoT mode works well, and your patch can solve the issue in interpreter mode, but I have a concern that copying the entire `WASMFunctionImport` may waste a lot of memory, since most of the fields in `WASMFunctionImport` are static informations, and only `function pointer` is required by each instance. 

We'd prefer this approach:
1. Add a `func_ptrs` array in `wasm instance` (just like AoT instance)
2. During instantiation, allocate memory for `func_ptrs` array
3. In this `functions_instantiate` function, just store the `function pointer` to the `func_ptrs` array
4. In `wasm_interp_call_func_native` function inside `wasm_interp_classic.c` and `wasm_interp_fast.c`, get the `function pointer` from `func_ptrs` array before invoking.
5. Of course we should also free the memory for `func_ptrs` array during deinstantiate.

This approach can reduce the memory consumption.

This is our preferred approach, and we look forward to hearing your comments.

If you think these modification may take too much effort, we can also do it together if you also agree with this solution.

Thanks

YoungWenMing: Thanks for this detailed and comprehensive solution which can save much more memories. I agree with this approach and I'd like to try to implement it within the next few days. ^_^

wenyongh: @YoungWenMing Thanks for the feedback, as there may be some places to change and it is a little complex, and also we hope to create a release soon, how about we fix the issue by ourselves?

YoungWenMing: OK

PR #1120, opened 2022-04-25
Enable register native with iwasm

wenyongh:
Enable register native with iwasm application, add sample and update document.

### DIFF
--- a/product-mini/platforms/posix/main.c
+++ b/product-mini/platforms/posix/main.c
@@ -40,7 +40,7 @@ print_help()
     printf("  --dir=<dir>            Grant wasi access to the given host directories\n");
     printf("                         to the program, for example:\n");
     printf("                           --dir=<dir1> --dir=<dir2>\n");
-    printf("  --addr-pool=           Grant wasi access to the given network addresses in\n");
+    printf("  --addr-pool=<addrs>    Grant wasi access to the given network addresses in\n");
     printf("                         CIRD notation to the program, seperated with ',',\n");
     printf("                         for example:\n");
     printf("                           --addr-pool=1.2.3.4/15,2.3.4.5/16\n");
@@ -49,7 +49,7 @@ print_help()
     printf("                         are shared object (.so) files, for example:\n");
     printf("                           --native-lib=test1.so --native-lib=test2.so\n");
 #if WASM_ENABLE_MULTI_MODULE != 0
-    printf("  --module-path=         Indicate a module search path. default is current\n"
+    printf("  --module-path=<path>   Indicate a module search path. default is current\n"
            "                         directory('./')\n");
 #endif
 #if WASM_ENABLE_LIB_PTHREAD != 0
--- a/product-mini/platforms/posix/main.c
+++ b/product-mini/platforms/posix/main.c
@@ -8,7 +8,9 @@
 #endif
 #include <stdlib.h>
 #include <string.h>
+#if defined(_POSIX_SOURCE) || defined(_POSIX_C_SOURCE)
 #include <dlfcn.h>
+#endif
 
 #include "bh_platform.h"
 #include "bh_read_file.h"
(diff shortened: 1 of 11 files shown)


samples/native-lib/README.md
```diff
@@ -0,0 +1,61 @@
...
+    *p_module_name = "env";
+    *p_native_symbols = native_symbols;
+    return sizeof(native_symbols) / sizeof(NativeSymbol);
+}
+```
+
+## Preparation
+
+Please install WASI SDK, download the [wasi-sdk release](https://github.com/CraneStation/wasi-sdk/releases) and extract the archive to default path `/opt/wasi-sdk`.
+And install wabt, download the [wabt release](https://github.com/WebAssembly/wabt/releases) and extract the archive to default path `/opt/wabt`
```
xujuntwt95329, line 34: Seems `wabt` is not required by this sample?

samples/native-lib/README.md
```diff
@@ -0,0 +1,61 @@
+# "native-lib" sample introduction
+
+This sample demonstrates how to write required interfaces in native library, build it into a shared library and
+register the shared library to iwasm.
```
xujuntwt95329, line 4: unnecessary line break

xujuntwt95329 (commented):
LGTM

samples/native-lib/test_add.c
```diff
@@ -0,0 +1,32 @@
...
+{
+    return x + y;
+}
+
+/* clang-format off */
+#define REG_NATIVE_FUNC(func_name, signature) \
+    { #func_name, func_name##_wrapper, signature, NULL }
+
+static NativeSymbol native_symbols[] = {
+    REG_NATIVE_FUNC(test_add, "(ii)i")
```
lum1n0us, line 22: there should be a function named `test_add`, right?

lum1n0us:
I got a rough idea about "register native". 

Some observations:

- the module name and the signature information of a register function depend on users' input
- users' compilers make sure wrappers are in the library

So, if I can replace the `REG_NATIVE_FUNC` with a .json file, and provide information including the module name, wrappers' names and signatures, wrappers' target,  and plus modifications on `load_and_register_native_libs`, we can let users register existed native function without recompiling their libraries.

Further, if users' can make sure all a WASM module's native dependencies are in existed libraries, and all import entries are using libraries name as modules name and using wasm function names as native function name, we can scan wasm import entries, and create such a .json automatically,

samples/native-lib/test_add.c
wenyongh: Yes, see file samples/native-lib/wasm-app/main.c: https://github.com/bytecodealliance/wasm-micro-runtime/pull/1120/files#diff-bc10ed7477a7549d93f8aca9e41f1fb832c68d4939f955bfe9f29c9769a521ff
it is compiled into wasm file, which calls to the imported function test_add/test_sqrt in its main function.

samples/native-lib/README.md
wenyongh: Done

samples/native-lib/README.md
wenyongh: Done

product-mini/platforms/posix/main.c
```diff
@@ -8,6 +8,9 @@
 #endif
 #include <stdlib.h>
 #include <string.h>
+#if defined(_POSIX_SOURCE) || defined(_POSIX_C_SOURCE)
```
yamt, line 11: what's the intention of this `defined(_POSIX_SOURCE) || defined(_POSIX_C_SOURCE)`?
is it meant to be an availability check of dlopen?

on nuttx, the availability of dlopen depends on a build-time config.
i feel it's more straightforward to control this with BH_HAS_DLFCN or something like that.
how do you think?

yamt: https://github.com/bytecodealliance/wasm-micro-runtime/pull/1134

PR #1121, opened 2022-04-25
Protect c_api from unlinked runtime objects

lum1n0us:


core/iwasm/common/wasm_c_api.c
```diff
@@ -2669,7 +2679,20 @@ wasm_func_call(const wasm_func_t *func, const wasm_val_vec_t *params,
...
     size_t param_count, result_count, alloc_count;
 
-    bh_assert(func && func->type && func->inst_comm_rt);
+    if (!func) {
+        return NULL;
+    }
+
+    if (!func->inst_comm_rt) {
+        wasm_name_t message = { 0 };
+        wasm_name_new_from_string(&message, "the function isn't linked");
```
wenyongh, line 2688: Do we need to check return value here?

core/iwasm/common/wasm_c_api.c
```diff
@@ -2669,7 +2679,20 @@ wasm_func_call(const wasm_func_t *func, const wasm_val_vec_t *params,
...
+        wasm_trap_t *trap;
```
wenyongh, line 2689: Put the variable definition at the beginning of `{}`

core/iwasm/common/wasm_c_api.c
```diff
@@ -4217,13 +4249,13 @@ wasm_instance_new_with_args(wasm_store_t *store, const wasm_module_t *module,
...
     /* link module and imports */
-    if (imports) {
+    if (imports && imports->num_elems) {
 #if WASM_ENABLE_INTERP != 0
         if ((*module)->module_type == Wasm_Module_Bytecode) {
             import_count = MODULE_INTERP(module)->import_count;
 
-            INIT_VEC(instance->imports, wasm_extern_vec_new_uninitialized,
-                     import_count);
+            // INIT_VEC(instance->imports, wasm_extern_vec_new_uninitialized,
```
wenyongh, line 4257: Why not remove the two lines? May be a little strange as instance->imports doesn't exist. If to keep them, please use `/* */` to comment them as here it is C source code.

core/iwasm/common/wasm_c_api.c
```diff
@@ -4246,8 +4278,8 @@ wasm_instance_new_with_args(wasm_store_t *store, const wasm_module_t *module,
                            + MODULE_AOT(module)->import_memory_count
                            + MODULE_AOT(module)->import_table_count;
 
-            INIT_VEC(instance->imports, wasm_extern_vec_new_uninitialized,
-                     import_count);
+            // INIT_VEC(instance->imports, wasm_extern_vec_new_uninitialized,
```
wenyongh, line 4281: Same as above

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1309,7 +1309,7 @@ wasm_runtime_finalize_call_function(WASMExecEnv *exec_env,
 
     bh_assert((argv && ret_argv) || (argc == 0));
 
-    if (argv == ret_argv || argc == 0) {
```
wenyongh, line 1312: Not sure why remove it, will it cause issue?

core/iwasm/common/wasm_c_api.c
lum1n0us: - if `wasm_name_new_from_string` failed, we have to return `NULL`
- a `NULL` message will lead `wasm_trap_new` returns `NULL` alone with a failure of creating a `wasm_trap_t`.

So I guess we can avoid checking here and let it go

core/iwasm/common/wasm_c_api.c
```diff
@@ -2669,7 +2679,22 @@ wasm_func_call(const wasm_func_t *func, const wasm_val_vec_t *params,
...
-    bh_assert(func && func->type && func->inst_comm_rt);
+    if (!func) {
+        return NULL;
+    }
+
+    if (!func->inst_comm_rt) {
+        wasm_name_t message = { 0 };
+        wasm_trap_t *trap;
+
+        wasm_name_new_from_string(&message, "the function isn't linked");
```
wenyongh, line 2690: message is not very friendly, had better be similiar with interpreter/aot, how about changing to `failed to call unlinked function`?

core/iwasm/common/wasm_c_api.c
wenyongh: If wasm_name_new_from_string failed, it doesn't return NULL, but the message vector is empty:
wasm_name_new_from_string(&message, ...)
And here `trap = wasm_trap_new(func->store, &message);`  will create an empty trap, but not NULL:
```C
wasm_trap_t *
wasm_trap_new(wasm_store_t *store, const wasm_message_t *message)
{
    wasm_trap_t *trap;
    if (!store || !message) {
        return NULL;
    }
    if (!(trap = malloc_internal(sizeof(wasm_trap_t)))) {
        return NULL;
    }
    INIT_VEC(trap->message, wasm_byte_vec_new, message->size, message->data); 
    ==> message->size is 0, an empty trap is created.

    return trap;
...
```
Why not check the result of wasm_name_new_from_string like others, e.g.:
```C
        wasm_name_new_from_string(&module_name, module_name_rt);
        if (strlen(module_name_rt) && !module_name.data) {
            goto failed;
        }
```

core/iwasm/common/wasm_runtime_common.c
lum1n0us: Yes, it will. If there is an `externref` in parameters, the code will malloc `new_argv` to transfer, and `new_argv` should be freed even if there is no result.

core/iwasm/common/wasm_c_api.c
lum1n0us: My bad explanation. I mean it doesn't matter if the `message` is empty or NULL.  The first priority is to raise a trap.  maybe I should change `wasm_trap_new`, like

`if (!store || !message)` => `if (!store)`

core/iwasm/common/wasm_c_api.c
lum1n0us: indeed.

core/iwasm/common/wasm_runtime_common.c
wenyongh: OK

core/iwasm/common/wasm_c_api.c
wenyongh: OK, let trap be thrown firstly.

core/iwasm/common/wasm_c_api.c
```diff
@@ -3965,30 +4007,22 @@ aot_link_func(const wasm_instance_t *inst, const AOTModule *module_aot,
...
     bh_assert(inst && module_aot && import);
 
     import_aot_func = module_aot->import_funcs + import_func_idx_rt;
     bh_assert(import_aot_func);
 
-    if (!(cloned = wasm_func_copy(import))) {
-        return false;
-    }
-
-    if (!bh_vector_append((Vector *)inst->imports, &cloned)) {
```
wenyongh, line 3979: Can also remove `imports` field in `wasm_instance_t` struct.

PR #1123, opened 2022-04-26
Impl memory grow

lum1n0us:


core/iwasm/fast-jit/fe/jit_emit_memory.c
```diff
@@ -408,9 +409,50 @@ jit_compile_op_memory_size(JitCompContext *cc)
...
+    grow_result = jit_codegen_get_hreg_by_name("eax");
+#else
+    grow_result = jit_cc_new_reg_I32(cc);
+#endif
+    module = get_module_reg(cc->jit_frame);
+    POP_I32(delta);
+    insn = GEN_INSN(CALLNATIVE, grow_result,
+                    NEW_CONST(PTR, (uintptr_t)wasm_enlarge_memory), 2);
+    if (insn) {
+        *(jit_insn_opndv(insn, 2)) = module;
```
wenyongh, line 439: Should be module_inst

wenyongh, line 439: Should be module_inst

core/iwasm/fast-jit/fe/jit_emit_memory.c
```diff
@@ -408,9 +409,50 @@ jit_compile_op_memory_size(JitCompContext *cc)
...
+        *(jit_insn_opndv(insn, 3)) = delta;
+    }
+
+#if defined(BUILD_TARGET_X86_64) || defined(BUILD_TARGET_AMD_64)
+    GEN_INSN(MOV, grow_result, grow_result);
```
wenyongh, line 444: Seems no need to add this line, as eax is used explicitly, not like edx in DIV/REM, which is used implicitly: used by x86 instruction and not used in INSN.
In aot_emit_function.c, eax is used, and no extra MOV is used, seems it works.

wenyongh, line 444: Seems no need to add this line, as eax is used explicitly, not like edx in DIV/REM, which is used implicitly: used by x86 instruction and not used in INSN.
In aot_emit_function.c, eax is used, and no extra MOV is used, seems it works.

core/iwasm/fast-jit/fe/jit_emit_memory.c
```diff
@@ -408,9 +409,50 @@ jit_compile_op_memory_size(JitCompContext *cc)
...
+#endif
+
+    /* check if enlarge memory success */
+    res = jit_cc_new_reg_I32(cc);
+    GEN_INSN(CMP, cc->cmp_reg, grow_result, NEW_CONST(I32, 1));
+    GEN_INSN(SELECTEQ, res, cc->cmp_reg, prev_page_count,
```
wenyongh, line 450: Had better check not zero, but not check whether it is 1, some systems' bool true might be not 1.
`CMP cmp_reg, grow_result, 0`, `SELECTNE`

core/iwasm/fast-jit/fe/jit_emit_memory.c
```diff
@@ -408,9 +409,50 @@ jit_compile_op_memory_size(JitCompContext *cc)
...
+        *(jit_insn_opndv(insn, 3)) = delta;
+    }
+
+#if defined(BUILD_TARGET_X86_64) || defined(BUILD_TARGET_AMD_64)
+    GEN_INSN(MOV, grow_result, grow_result);
+#endif
+
+    /* check if enlarge memory success */
+    res = jit_cc_new_reg_I32(cc);
+    GEN_INSN(CMP, cc->cmp_reg, grow_result, NEW_CONST(I32, 1));
```
wenyongh, line 449: Had better check not zero, but not check whether it is 1, some systems' bool true might be not 1.
`CMP cmp_reg, grow_result, 0`, `SELECTNE`

wenyongh, line 449: Had better check not zero, but not check whether it is 1, some systems' bool true might be not 1.
`CMP cmp_reg, grow_result, 0`, `SELECTNE`

PR #1124, opened 2022-04-27
Add missing wasm_export.h include in a few places

yamt:
This fixes a few pointer truncation crashes I've seen with WASI.

### DIFF
--- a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
+++ b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
@@ -21,6 +21,7 @@
 #include "refcount.h"
 #include "rights.h"
 #include "str.h"
+#include "wasm_export.h"
 
 #if 0 /* TODO: -std=gnu99 causes compile error, comment them first */
 // struct iovec must have the same layout as __wasi_iovec_t.
--- a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/str.c
+++ b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/str.c
@@ -14,6 +14,7 @@
 #include "ssp_config.h"
 #include "bh_platform.h"
 #include "str.h"
+#include "wasm_export.h"
 
 static char *
 bh_strndup(const char *s, size_t n)
--- a/core/shared/utils/bh_platform.h
+++ b/core/shared/utils/bh_platform.h
@@ -17,6 +17,9 @@
 #include "bh_queue.h"
 #include "bh_vector.h"
 #include "runtime_timer.h"
+#if !defined(WA_MALLOC) || !defined(WA_FREE)
+#include "wasm_export.h"
+#endif
 
 /**
  * WA_MALLOC/WA_FREE need to be redefined for both


core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -21,6 +21,7 @@
 #include "refcount.h"
 #include "rights.h"
 #include "str.h"
+#include "wasm_export.h"
```
wenyongh, line 24: Hi, the two files can be built without compilation warnings in Linux platform, you write your own Makefile to build them, and there are compilation warnings, right? It should be because that these files call `wasm_runtime_malloc`/`wasm_runtime_free`, and the prototypes are not declared.

In Linux, it uses cmake, and in core/iwasm/common/iwasm_common.cmake, it defines the macro `BH_MALLOC` and `BH_FREE` to `wasm_runtime_malloc` and `wasm_runtime_free`, see: https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/common/iwasm_common.cmake#L8-L9

And in platform_common.h, the prototypes of BH_MALLOC/BH_FREE are defined, see https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/shared/platform/include/platform_common.h#L40-L53

So could you please add the macros BH_MALLOC/BH_FREE in your own makefile like iwasm_common.cmake? In our original design, wasm_export.h is only for runtime embedder to use, we had better no use it inside runtime.

wenyongh, line 24: Hi, the two files can be built without compilation warnings in Linux platform, you write your own Makefile to build them, and there are compilation warnings, right? It should be because that these files call `wasm_runtime_malloc`/`wasm_runtime_free`, and the prototypes are not declared.

In Linux, it uses cmake, and in core/iwasm/common/iwasm_common.cmake, it defines the macro `BH_MALLOC` and `BH_FREE` to `wasm_runtime_malloc` and `wasm_runtime_free`, see: https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/common/iwasm_common.cmake#L8-L9

And in platform_common.h, the prototypes of BH_MALLOC/BH_FREE are defined, see https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/shared/platform/include/platform_common.h#L40-L53

So could you please add the macros BH_MALLOC/BH_FREE in your own makefile like iwasm_common.cmake? In our original design, wasm_export.h is only for runtime embedder to use, we had better no use it inside runtime.

yamt: ok. thank you for explanation. i will revisit.

yamt: > you write your own Makefile to build them, and there are compilation warnings, right? 

yes

> So could you please add the macros BH_MALLOC/BH_FREE in your own makefile like iwasm_common.cmake?

ok

> In our original design, wasm_export.h is only for runtime embedder to use, we had better no use it inside runtime.

shouldn't sandboxed-system-primitives use BH_MALLOC instead of wasm_runtime_malloc then?

yamt: https://github.com/bytecodealliance/wasm-micro-runtime/pull/1135

wenyongh: > > you write your own Makefile to build them, and there are compilation warnings, right?
> 
> yes
> 
> > So could you please add the macros BH_MALLOC/BH_FREE in your own makefile like iwasm_common.cmake?
> 
> ok
> 
> > In our original design, wasm_export.h is only for runtime embedder to use, we had better no use it inside runtime.
> 
> shouldn't sandboxed-system-primitives use BH_MALLOC instead of wasm_runtime_malloc then?

No, had better use wasm_runtime_malloc, if using BH_MALLOC and it isn't defined in makefile, BH_MALLOC equals to os_malloc, and when os_malloc isn't implemented (return NULL), here it may run failed.

yamt: > > shouldn't sandboxed-system-primitives use BH_MALLOC instead of wasm_runtime_malloc then?
> 
> No, had better use wasm_runtime_malloc, if using BH_MALLOC and it isn't defined in makefile, BH_MALLOC equals to os_malloc, and when os_malloc isn't implemented (return NULL), here it may run failed.

it's better than the current state.
currently it works only when BH_MALLOC=wasm_runtime_malloc.

wenyongh: No, here the libc-wasi is builtin lib for runtime and it must be built with runtime, the wasm_runtime_malloc must be implemented and we had better use wasm_runtime_malloc same as runtime uses, so as to manage all the memory in the same way. BH_MALLOC is only used in core/shared, when core/shared is built for runtime, we should define it to wasm_runtime_malloc, and when core shared isn't built for runtime, for example, in project test-tools/host-tool, we don't need to re-define it.

yamt: ok.
so, does it makes sense to somehow assert BH_MALLOC=wasm_runtime_malloc in runtime?

wenyongh: Yes, if to add assertion, had better add it in wasm_runtime_common.h or wasm_runtime_common.h.c.

yamt: https://github.com/bytecodealliance/wasm-micro-runtime/pull/1139

PR #1127, opened 2022-04-27
core/shared/platform/nuttx: mock several APIs for libc-wasi

yamt:

PR #1129, opened 2022-04-27
product-mini/platforms/nuttx/wamr.mk: Build libc-wasi

yamt:


yamt:
this is marked a draft as it depends on https://github.com/bytecodealliance/wasm-micro-runtime/pull/1127 (or real implementation in nuttx)

yamt:
> this is marked a draft as it depends on #1127 (or real implementation in nuttx)

now it's ready

PR #1130, opened 2022-04-27
store import function pointer in module instance

xujuntwt95329:
Thanks @YoungWenMing for reporting the issue that import functions will be overwrite by further instance in PR #1118 

After some discussions, we decide to use this approach since it avoid copying unnecessary static information into instance.


YoungWenMing:
My pleasure～

PR #1133, opened 2022-04-28
posix os_socket_inet_network: Use inet_addr instead of inet_network

yamt:
* inet_addr is in posix. inet_network is not.

* NuttX doesn't have inet_network.


lum1n0us:
LGTM. 😄

PR #1137, opened 2022-04-28
Fix execute_main not wait for other threads

wenyongh:
Fix wasm_application_execute_main/wasm_application_execute_func not waiting for
other threads to terminate in multi-thread mode, which causes that the exception
thrown by other threads may haven't been not spread the current main thread, and
cannot be detected by the caller, as reported in #1131.


xujuntwt95329 (commented):
LGTM

PR #1139, opened 2022-05-02
wasm_runtime_common.c: assert BH_MALLOC/BH_FREE

yamt:


core/iwasm/common/wasm_runtime_common.c
```diff
@@ -29,6 +29,17 @@
 #endif
 #include "../common/wasm_c_api_internal.h"
 
+/**
+ * For runtime build,
+ * BH_MALLOC/BH_FREE should be wasm_runtime_malloc/wasm_runtime_free.
+ */
+#if BH_MALLOC != wasm_runtime_malloc
```
wenyongh, line 36: @yamt Seems that such expression doesn't work, if I change it to `BH_MALLOC != os_malloc`, or keep it and remove `add_definitions(-DBH_MALLOC=wasm_runtime_malloc)` in iwasm_common.cmake, the compiler doesn't report error also. Is there other way to test the macros?

yamt: i pushed a working version. it's a bit more complicated than i want it to be though.

PR #1140, opened 2022-05-02
Fix aot compiler compilation errors with llvm-14

wenyongh:
Fix aot compiler compilation errors when the llvm version is 14.0,
and clear one compilation warning of thread_manager.c.


xujuntwt95329 (commented):
LGTM

PR #1147, opened 2022-05-06
wasm_export.h: Add a few comments about heap and threads

yamt:


core/iwasm/include/wasm_export.h
```diff
@@ -699,6 +705,10 @@ wasm_runtime_validate_app_addr(wasm_module_inst_t module_inst,
  * space or memory space. Moreover, it checks whether it is the offset of a
  * string that is end with '\0'.
  *
+ * Note: The validation result, especially the NUL termination check,
```
wenyongh, line 708: Should be NULL?

core/iwasm/include/wasm_export.h
```diff
@@ -729,6 +739,11 @@ wasm_runtime_validate_native_addr(wasm_module_inst_t module_inst,
 /**
  * Convert app address(relative address) to native address(absolute address)
  *
+ * Note that all native addresses to module instance memory can be
+ * invalidated on a memory growth. Thus, this API is basically unsafe
+ * to use for module instances with multiple threads unless you somehow
```
wenyongh, line 744: It should be safe when multi-thread mode is enabled in WAMR, in this mode the linear memory is shared and is allocated with its maximum size, and won't be re-allocated again, we just increase the memory's current page count when memory growing:

https://github.com/WebAssembly/threads/blob/main/proposals/threads/Overview.md#shared-linear-memory

https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_runtime.c#L2161

core/iwasm/include/wasm_export.h
yamt: i usually use NUL for '\0' and NULL for (void *)0.

core/iwasm/include/wasm_export.h
xujuntwt95329: WebAssembly's `thread` feature depends on the `shared_memory` feature, because the threads should have same "heap". If a wasm module's memory is not marked as `shared`, then it is not possible to create threads for it.

yamt: ok. thank you for explanation. i will update.

yamt: done

PR #1150, opened 2022-05-06
Fix some issues reported by Coverity

wenyongh:
module_wasm_app.c: add return value check for wasm_runtime_call_wasm
aot_runtime.c: add return value check for aot_get_default_memory
aot_runtime.c: add return value  check before calling wasm app malloc/free func
wasm_runtime_common.c: fix dead code warning in wasm_runtime_load_from_sections
aot_emit_memory.c: fix potential integer overflow issue
wasm_runtime.c: remove dead code in memory_instantiate, add assertion for globals
samples simple/gui/littlevgl: fix fields of struct sigaction initialization issue
host-tool: add return value check for sendto


xujuntwt95329 (commented):
LGTM

PR #1176, opened 2022-05-19
Implement SGX getrandom/getentropy without ocall

wenyongh:
Implement SGX getrandom with sgx_read_rand and getentropy with `rdseed` instruction
instead of ocall to improve the security.

### DIFF
--- a/core/shared/platform/linux-sgx/platform_internal.h
+++ b/core/shared/platform/linux-sgx/platform_internal.h
@@ -28,6 +28,7 @@
 #include "sgx_time.h"
 #include "sgx_socket.h"
 #include "sgx_signal.h"
+#include "sgx_trts.h"
 
 #ifdef __cplusplus
 extern "C" {
--- a/core/shared/platform/linux-sgx/sgx_file.c
+++ b/core/shared/platform/linux-sgx/sgx_file.c
@@ -891,29 +891,39 @@ sched_yield(void)
 ssize_t
 getrandom(void *buf, size_t buflen, unsigned int flags)
 {
-    ssize_t ret;
+    sgx_status_t ret;
 
-    if (ocall_getrandom(&ret, buf, buflen, flags) != SGX_SUCCESS) {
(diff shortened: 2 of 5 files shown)


zhenghaven:
The new `getrandom` implementation looks good to me. However, for `getentropy`, I'm not sure if using `sgx_read_rand` is secure enough for the entropy, since `sgx_read_rand` uses RDRAND instruction (reference: https://github.com/intel/linux-sgx/blob/master/external/rdrand/src/rdrand.c and https://github.com/intel/linux-sgx/blob/master/common/src/sgx_read_rand.cpp). Probably RDSEED instruction is a better option for entropy. As far as I know, Intel SGX SDK does not provide a high-level API to RDSEED instruction, but RDSEED instruction itself should be available inside of SGX. And I found Intel provides some sample code for using the RDSEED instruction: https://www.intel.com/content/www/us/en/developer/articles/guide/intel-digital-random-number-generator-drng-software-implementation-guide.html

wenyongh:
> The new `getrandom` implementation looks good to me. However, for `getentropy`, I'm not sure if using `sgx_read_rand` is secure enough for the entropy, since `sgx_read_rand` uses RDRAND instruction (reference: https://github.com/intel/linux-sgx/blob/master/external/rdrand/src/rdrand.c and https://github.com/intel/linux-sgx/blob/master/common/src/sgx_read_rand.cpp). Probably RDSEED instruction is a better option for entropy. As far as I know, Intel SGX SDK does not provide a high-level API to RDSEED instruction, but RDSEED instruction itself should be available inside of SGX. And I found Intel provides some sample code for using the RDSEED instruction: https://www.intel.com/content/www/us/en/developer/articles/guide/intel-digital-random-number-generator-drng-software-implementation-guide.html

Thanks, code of getentropy was updated by using `rdseed` instruction.

zhenghaven:
It compiles and works on my end. Thank you for addressing this issue!

PR #1179, opened 2022-05-20
aot_reloc_xtensa.c: define __packed if not available

yamt:


core/iwasm/aot/arch/aot_reloc_xtensa.c
```diff
@@ -130,6 +130,16 @@ static union {
 
 #define is_little_endian() (__ue.b == 1)
 
+#if !defined(__packed)
+/*
+ * Note: This version check is a bit relaxed.
+ * The __packed__ attribute has been there since gcc 2 era.
+ */
+#if __GNUC__ >= 3
+#define __packed __attribute__((__packed__))
```
wenyongh, line 139: Had better add else:
```C
#if __GNUC__ >= 3
#define __packed __attribute__((__packed__))
#else
#define __packed
#endif
```

yamt: given the way it's used in this file, i feel it's better to leave it undefined so that someone using an incompatible compiler get an error. hopefully he can easily add a proper definition for his compiler.

wenyongh: OK, thanks.

PR #1187, opened 2022-05-23
Extend lock range

lum1n0us:
Callees of `extend_vector` make sure the function is running under
thread-safe environment

### DIFF
--- a/core/shared/utils/bh_vector.c
+++ b/core/shared/utils/bh_vector.c
@@ -23,6 +23,10 @@ alloc_vector_data(size_t length, size_t size_elem)
     return data;
 }
 
+/**
+ * every caller of `extend_vector` must provide
+ * a thread-safe environment.
+ */
 static bool
 extend_vector(Vector *vector, size_t length)
 {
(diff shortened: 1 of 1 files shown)


core/shared/utils/bh_vector.c
```diff
@@ -143,24 +143,26 @@ bh_vector_insert(Vector *vector, uint32 index, const void *elem_buf)
 {
     size_t i;
     uint8 *p;
+    bool ret = false;
 
     if (!vector || !elem_buf) {
         LOG_ERROR("Insert vector elem failed: vector or elem buf is NULL.\n");
-        return false;
+        goto RETURN;
```
xujuntwt95329, line 150: Seems we can return false directly here, no need to add a `RETURN` label?

core/shared/utils/bh_vector.c
```diff
@@ -143,24 +143,26 @@ bh_vector_insert(Vector *vector, uint32 index, const void *elem_buf)
...
     }
 
     if (index >= vector->num_elems) {
         LOG_ERROR("Insert vector elem failed: invalid elem index.\n");
-        return false;
+        goto RETURN;
```
wenyongh, line 155: Can we rename the label to lowercase? Had better be consistent with current coding style. And label name of uppercase is not easy to read.

core/shared/utils/bh_vector.c
```diff
@@ -23,6 +23,10 @@ alloc_vector_data(size_t length, size_t size_elem)
     return data;
 }
 
+/**
+ * every callee of `extend_vector` must provide
```
xujuntwt95329, line 27: Should be caller?

core/shared/utils/bh_vector.c
lum1n0us: since `goto` is in the function, try using "one-exit-point"

PR #1189, opened 2022-05-24
Add comments on trailing uint8[1] members in the "Common" structures

yamt:


core/iwasm/common/wasm_runtime_common.h
```diff
@@ -303,7 +303,6 @@ typedef struct WASMModuleCommon {
        Wasm_Module_AoT, and this structure should be treated as
        AOTModule structure. */
     uint32 module_type;
-    uint8 module_data[1];
```
wenyongh, line 306: Adding `module_data[1]` here is to indicate that there is extra data in this structure, which makes the code easier to understand. My personal opinion is that we had better keep it, or add comments for it.
Is there any compilation error or execution error?

wenyongh, line 306: Adding `module_data[1]` here is to indicate that there is extra data in this structure, which makes the code easier to understand. My personal opinion is that we had better keep it, or add comments for it.
Is there any compilation error or execution error?

yamt: ok. i will add comments instead.

>  Is there any compilation error or execution error?

no.
the motivation of removal is that i saw someone trying to abuse this member to store extra data.

yamt: done.

wenyongh: Hi, my meaning is that we keep the field `module_data[1]`, and just add comments for it. I discussed with @XuJun2019, he also prefers not to remove it.

yamt: ok.
how about making them c99 flexible array member?
it would avoid casual readers from thinking there is a spare one byte to use.

xujuntwt95329: Hi, @yamt 

Flexible array member is introduced in C99, it is supported by most of the modern compilers, however, we used to encounter some compilation errors on some embedded device toolchains, the compiler complains they didn't recognize such kind of syntax (e.g. `module_data[]`, `module_data[0]`).

So IMO, using `module_data[1]` should be a more compatible approach. Had better add a comment to explain that the `module_data[1]` is actually a flexible array and can't be used to store user data.

yamt: ok. i made this just comments.

PR #1201, opened 2022-06-01
Fix fast jit issues

wenyongh:


core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
```diff
@@ -2902,9 +2964,9 @@ alu_imm_r_to_r_f32(x86::Assembler &a, ALU_OP op, int32 reg_no_dst,
     a.movaps(cache, regs_float[reg_no2_src]);
 
     /* imm -> gp -> xmm */
-    mov_imm_to_r_f32(a, reg_no2_src, data1_src);
+    mov_imm_to_r_f32(a, reg_no_dst, data1_src);
 
-    a.addss(regs_float[reg_no2_src], cache);
+    alu_r_m_float(a, op, reg_no_dst, cache, true);
     return true;
```
lum1n0us, line 2970: how about `return alu_r_m_float(...)`.

wenyongh: Done

core/iwasm/fast-jit/fe/jit_emit_function.c
```diff
@@ -341,23 +336,34 @@ jit_compile_op_call_indirect(JitCompContext *cc, uint32 type_idx,
...
-    *(jit_insn_opndv(insn, 4)) = element_indices;
-    *(jit_insn_opndv(insn, 5)) = NEW_CONST(I32, func_type->param_count);
-    *(jit_insn_opndv(insn, 6)) = argv;
+    *(jit_insn_opndv(insn, 4)) = elem_idx;
+    *(jit_insn_opndv(insn, 5)) = NEW_CONST(I32, type_idx);
+    *(jit_insn_opndv(insn, 6)) = NEW_CONST(I32, func_type->param_cell_num);
+    *(jit_insn_opndv(insn, 7)) = argv;
 
+#if defined(BUILD_TARGET_X86_64) || defined(BUILD_TARGET_AMD_64)
     jit_lock_reg_in_insn(cc, insn, native_ret);
```
lum1n0us, line 365: shall we lock `edx` and `r9' here?

lum1n0us: second thought, if `edx` and `r9` are "caller saved register", there is no need to lock them, right?

Or in other words, we should not lock any "caller saved register"

PR #1206, opened 2022-06-03
Refactor emitting callnative/callbc IRs

wenyongh:
And remove clear_values for opcode br_if as the jit registers
can be used in the same basic block.


core/iwasm/fast-jit/fe/jit_emit_compare.c
```diff
@@ -120,7 +121,7 @@ jit_compile_op_i64_compare(JitCompContext *cc, IntCond cond)
 static int32
 float_cmp_eq(float f1, float f2)
 {
-    if (isnan(f1) || isnan(f2))
+    if (isnanf(f1) || isnanf(f2))
```
lum1n0us, line 124: seems `isnanf` is not a standard c99 function in` math.h`

https://en.cppreference.com/w/c/numeric/math

core/iwasm/fast-jit/fe/jit_emit_conversion.c
```diff
@@ -50,31 +50,19 @@ jit_compile_op_i32_trunc_f32(JitCompContext *cc, bool sign, bool saturating)
...
-        /* Set nan_ret to x86::eax */
-        nan_ret = jit_codegen_get_hreg_by_name("eax");
-#else
+        /* If value is NaN, throw exception */
         nan_ret = jit_cc_new_reg_I32(cc);
-#endif
-        insn =
-            GEN_INSN(CALLNATIVE, nan_ret, NEW_CONST(PTR, (uintptr_t)isnanf), 1);
-        if (!insn) {
+        if (!jit_emit_callnative(cc, isnanf, nan_ret, &value, 1)) {
```
lum1n0us, line 55: I forgot to change here. So let me know if we are going to stick on the standard c99 function or not,  and I will scan all references of math functions, and make them align to the standard.

core/iwasm/fast-jit/fe/jit_emit_function.c
```diff
@@ -403,3 +378,122 @@ jit_compile_op_ref_func(JitCompContext *cc, uint32 func_idx)
...
 }
+
+#if defined(BUILD_TARGET_X86_64) || defined(BUILD_TARGET_AMD_64)
+bool
+jit_emit_callnative(JitCompContext *cc, void *native_func, JitReg res,
+                    JitReg *params, uint32 param_count)
+{
+    JitInsn *insn;
+    char *i64_arg_names[] = { "rdi", "rsi", "rdx", "rcx", "r8", "r9" };
+    char *f32_arg_names[] = { "xmm0", "xmm1", "xmm2", "xmm3", "xmm4", "xmm5" };
```
lum1n0us, line 389: if it is a *system V AMD64 ABI* , I thought we should use xmm0-7 to store floating-point arguments.

core/iwasm/fast-jit/fe/jit_emit_function.c
```diff
@@ -403,3 +378,122 @@ jit_compile_op_ref_func(JitCompContext *cc, uint32 func_idx)
...
+                return false;
+        }
+    }
+
+    if (res) {
+        GEN_INSN(MOV, res, res_hreg);
+    }
+
+    return true;
+}
```
lum1n0us, line 478: Actually, my opinion is to put all specific-hardware-register stuff in `lower_callnative()`.

- we can use least regalloc knowledge in frontend. It fits the abstraction of the frontend module.
- After regallocation promising the caller-save environment, we have no worried to use any hardware register as we want.

PR #1207, opened 2022-06-06
support emit specified custom sections to AoT file

xujuntwt95329:


core/iwasm/compilation/aot_llvm.c
```diff
@@ -2068,6 +2068,43 @@ aot_create_comp_context(AOTCompData *comp_data, aot_comp_option_t option)
...
+
+            content = wasm_loader_get_custom_section(
+                comp_data->wasm_module, option->custom_sections[i], &length);
+            if (!content) {
+                LOG_WARNING("Can't find custom section [%s], ignore it",
+                            option->custom_sections[i]);
+                continue;
+            }
+
+            if (!(section = wasm_runtime_malloc(sizeof(WASMSection)))) {
```
wenyongh, line 2087: Had better `(uint32)sizeof(WASMSection)` to avoid compile warning

core/iwasm/interpreter/wasm.h
```diff
@@ -10,6 +10,10 @@
 #include "bh_hashmap.h"
 #include "bh_assert.h"
 
+#if WASM_ENABLE_LOAD_CUSTOM_SECTION != 0
+#include "wasm_export.h"
```
wenyongh, line 14: wasm_export.h is included in wasm_runtime_common.h, how about not including wasm_export.h in this file?
And try using "struct wasm_section_t *custom_section_list" below?

core/iwasm/aot/aot_loader.c
```diff
@@ -674,6 +678,53 @@ load_custom_section(const uint8 *buf, const uint8 *buf_end, AOTModule *module,
...
+            WASMCustomSection *section;
+
+            if (p >= p_end) {
+                set_error_buf(error_buf, error_buf_size, "unexpected end");
+                goto fail;
+            }
+
+            read_string(p, p_end, section_name);
+
+            name_len = p - p_orig;
```
wenyongh, line 697: Should be `strlen(section_name)`? `p - p_orig` includes 2 bytes string length. section_name is already '\0' terminated.

core/iwasm/aot/aot_loader.c
```diff
@@ -3269,3 +3324,25 @@ aot_get_plt_table_size()
...
+
+#if WASM_ENABLE_LOAD_CUSTOM_SECTION != 0
+const uint8 *
+aot_get_custom_section(AOTModule *module, const char *name, uint32 *len)
+{
+    WASMCustomSection *section = module->custom_section_list;
+
+    while (section) {
+        if ((section->name_len == strlen(name))
+            && (memcmp(section->name_addr, name, section->name_len) == 0)) {
```
wenyongh, line 3336: Can just use `strcmp(section->section_name, name)`, but current code also works.

core/iwasm/aot/aot_runtime.h
```diff
@@ -733,6 +737,9 @@ aot_dump_call_stack(WASMExecEnv *exec_env);
 void
 aot_dump_perf_profiling(const AOTModuleInstance *module_inst);
 
+const uint8 *
+aot_get_custom_section(AOTModule *module, const char *name, uint32 *len);
```
wenyongh, line 741: Had better `const AOTModule *module`

core/iwasm/compilation/aot.c
```diff
@@ -4,6 +4,7 @@
  */
 
 #include "aot.h"
+#include "../interpreter/wasm_loader.h"
```
wenyongh, line 7: Had better restore the file if no need to change it?

core/iwasm/compilation/aot_emit_aot_file.c
```diff
@@ -939,6 +942,12 @@ get_aot_file_size(AOTCompContext *comp_ctx, AOTCompData *comp_data,
                      get_name_section_size(comp_data));
     }
 
+    if (comp_ctx->custom_sections_wp) {
+        /* custom sections */
+        size = align_uint(size, 4);
+        size += get_custom_sections_size(comp_ctx, comp_data);
```
wenyongh, line 948: How about not emitting content if `get_custom_sections_size` returns 0? e.g.
```C
uint32 size_custom_section = get_custom_sections_size(..);
if (size_custom_section > 0) {
    size = align_uint(size, 4);
    size += size_custom_section;
}
```

core/iwasm/aot/aot_loader.c
xujuntwt95329: Yes, things are much more easier when the string is NULL terminated in AoT, I've changed to strcmp

core/iwasm/aot/aot_loader.c
```diff
@@ -3269,3 +3307,24 @@ aot_get_plt_table_size()
...
 }
+
+#if WASM_ENABLE_LOAD_CUSTOM_SECTION != 0
+const uint8 *
+aot_get_custom_section(const AOTModule *module, const char *name, uint32 *len)
+{
+    WASMCustomSection *section = module->custom_section_list;
+
+    while (section) {
+        if (strcmp(section->name_addr, name)) {
```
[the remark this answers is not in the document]
xujuntwt95329: Yes, my mistake

Thanks~

PR #1208, opened 2022-06-06
Fix fast jit issues

wenyongh:
Move jit spill cache to the end of interp frame to reduce footprint
Fix codegen compare float issue: should not overwritten the source registers
Fix float to int conversion check integer overflow issue
Unify the float compare
Fix get_global issue

### DIFF
--- a/core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
+++ b/core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
@@ -82,6 +82,9 @@ x86::Gp regs_i64[] = {
     x86::r12, x86::r13, x86::r14, x86::r15,
 };
 
+#define REG_F32_FREE_IDX 15
+#define REG_F64_FREE_IDX 15
+
 x86::Xmm regs_float[] = {
     x86::xmm0,
     x86::xmm1,
@@ -349,6 +352,8 @@ cmp_r_and_jmp_label(JitCompContext *cc, x86::Assembler &a,
 
     bool fp_cmp = cc->last_cmp_on_fp;
 
+    bh_assert(!fp_cmp || (fp_cmp && (op == GES)));
+
     switch (op) {
         case EQ:
         {
@@ -362,60 +367,52 @@ cmp_r_and_jmp_label(JitCompContext *cc, x86::Assembler &a,
         }
         case GTS:
         {
-            if (fp_cmp) {
-                a.ja(imm);
-            }
-            else {
-                a.jg(imm);
-            }
+            a.jg(imm);
             break;
         }
         case LES:
         {
-            if (fp_cmp) {
-                a.jnb(imm);
-            }
-            else {
-                a.jng(imm);
-            }
+            a.jng(imm);
             break;
         }
         case GES:
         {
-            if (fp_cmp) {
-                a.jnb(imm);
-            }
-            else {
-
+            if (fp_cmp)
+                a.jae(imm);
+            else
                 a.jnl(imm);
-            }
             break;
         }
         case LTS:
         {
-            if (fp_cmp) {
-                a.ja(imm);
-            }
-            else {
-                a.jl(imm);
-            }
+            a.jl(imm);
             break;
         }
         case GTU:
+        {
             a.ja(imm);
             break;
+        }
         case LEU:
+        {
             a.jna(imm);
             break;
+        }
         case GEU:
+        {
             a.jnb(imm);
             break;
+        }
         case LTU:
+        {
             a.jb(imm);
             break;
+        }
         default:
+        {
             bh_assert(0);
             break;
+        }
     }
 
     if (r2) {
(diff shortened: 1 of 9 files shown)


core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
```diff
@@ -5174,10 +5152,13 @@ lower_select(JitCompContext *cc, x86::Assembler &a, COND_OP op, JitReg r0,
     CHECK_NCONST(r1);
     CHECK_KIND(r1, JIT_REG_KIND_I32);
 
-    if (r0 == r3 && r0 != r2) {
+    if (r0 == r3 && r0 != r2 && !cc->last_cmp_on_fp) {
```
lum1n0us, line 5155: just curious, if the frontend follows the SSA rule, this branch will never be used, right? because `r0` will not be equal with `r3`.

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
```diff
@@ -5090,18 +5068,13 @@ cmp_r_and_jmp_relative(JitCompContext *cc, x86::Assembler &a, int32 reg_no,
         }
         case LES:
         {
-            if (fp_cmp) {
-                a.jnb(target);
-            }
-            else {
-                a.jng(target);
-            }
+            a.jng(target);
```
lum1n0us, line 5071: YES. using `jnb` is a little wired here. 

But the full story is, `lhs<rhs` and `lhs<=rhs` are going to be translated to `rhs >= lhs` and `rhs > lhs`.

core/iwasm/fast-jit/fe/jit_emit_compare.c
```diff
@@ -199,7 +199,7 @@ jit_compile_op_compare_float_point(JitCompContext *cc, FloatCond cond,
             case FLOAT_LE:
             {
                 GEN_INSN(CMP, cc->cmp_reg, rhs, lhs);
-                GEN_INSN(SELECTLES, res, cc->cmp_reg, const_one, const_zero);
+                GEN_INSN(SELECTGES, res, cc->cmp_reg, const_one, const_zero);
```
lum1n0us, line 202: `lhs <= rhs` => `rhs` > `lhs` ?

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
wenyongh: Yes, definitely we can use other IRs or Asm Instructions to optimize it. For the following spec case:
```wasm
  (export "eq" (func $1))
  (export "ne" (func $2))
  (export "lt" (func $3))
  (export "le" (func $4))
  (export "gt" (func $5))
  (export "ge" (func $6))

  (func $1/$2/.../$6 (type $0)
    (param $0 f32)
    (param $1 f32)
    (result i32)
    local.get $0
    local.get $1
    f32.eq/f32.ne/f32.lt/f32.le/f32.gt/f32.ge
    )
```
The machine code generated by llvm is below (wamrc --format=object -o test.o test.wasm):
```asm
0000000000000000 <aot_func#0>:  => eq
   0:   c5 fa c2 c1 00          vcmpeqss %xmm1,%xmm0,%xmm0
   5:   c5 f9 7e c0             vmovd  %xmm0,%eax
   9:   83 e0 01                and    $0x1,%eax
   c:   c3                      retq
   d:   0f 1f 00                nopl   (%rax)

0000000000000010 <aot_func#1>:  =>ne
  10:   c5 fa c2 c1 04          vcmpneqss %xmm1,%xmm0,%xmm0
  15:   c5 f9 7e c0             vmovd  %xmm0,%eax
  19:   83 e0 01                and    $0x1,%eax
  1c:   c3                      retq
  1d:   0f 1f 00                nopl   (%rax)

0000000000000020 <aot_func#2>:   => lt
  20:   31 c0                   xor    %eax,%eax
  22:   c5 f8 2e c8             vucomiss %xmm0,%xmm1
  26:   0f 97 c0                seta   %al
  29:   c3                      retq
  2a:   66 0f 1f 44 00 00       nopw   0x0(%rax,%rax,1)

0000000000000030 <aot_func#3>:  => le
  30:   31 c0                   xor    %eax,%eax
  32:   c5 f8 2e c8             vucomiss %xmm0,%xmm1
  36:   0f 93 c0                setae  %al
  39:   c3                      retq
  3a:   66 0f 1f 44 00 00       nopw   0x0(%rax,%rax,1)

0000000000000040 <aot_func#4>:  => gt
  40:   31 c0                   xor    %eax,%eax
  42:   c5 f8 2e c1             vucomiss %xmm1,%xmm0
  46:   0f 97 c0                seta   %al
  49:   c3                      retq
  4a:   66 0f 1f 44 00 00       nopw   0x0(%rax,%rax,1)

0000000000000050 <aot_func#5>:  => ge
  50:   31 c0                   xor    %eax,%eax
  52:   c5 f8 2e c1             vucomiss %xmm1,%xmm0
  56:   0f 93 c0                setae  %al
  59:   c3                      retq
```
Note that it uses `vmovd %xmm0,%eax` or `seta` to set the result.

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
wenyongh: Yes, but it seems that the frontend doesn't follow the rule, the optimization was applied before, it happened when generating the IR for Jeff opcode.

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
wenyongh: Seems that we can also refine the f32.eq/ne without emitting callnative.

lum1n0us: Those are all AVX and AVX512 instructions, it would be better if we are allowed to use that. Currently, we limited ourselves in SSE/SSE2 instructions for the biggest compatibility.

| Instruction Sets | First Appears |
| --- | --- |
| SSE | 1999 |
| SSE2 | 2001 |
| SSE3 | 2004 |
| SSE4 |  2006 |
| AVX | 2008 |
| AVX2 | 2013 |
| AVX512 | 2015 |

lum1n0us: `cmp + setCC` seems can simple current `cmp + jmp` combination.

PR #1212, opened 2022-06-07
support print exception info in source debugger

xujuntwt95329:


xujuntwt95329:
This patch will enhance the user experience of source debugger, when exception occurs, it will report the exception, and this can be further displayed in tools like VSCode:

image

core/iwasm/libraries/debug-engine/handler.c
```diff
@@ -14,6 +14,9 @@
 static char tmpbuf[MAX_PACKET_SIZE];
 static korp_mutex tmpbuf_lock;
 
+extern WASMExecEnv *
+wasm_debug_instance_get_current_env(WASMDebugInstance *instance);
```
wenyongh, line 18: How about adding the declaration in debug_engine.h, and including it?

PR #1213, opened 2022-06-08
Fix fast jit int rem_s and const shl issues

wenyongh:
int rem_s -1 should return 0
int32 lhs << int32 rhs may cause sanitizer check failure
fix codegen I8TOI64, I16TOI64, I64TOI8, I64TOI16
implement codegen neg operations


lum1n0us:
LGTM

PR #1215, opened 2022-06-09
Refactory inn.trunc_fmm_sx and inn.trunc_fmm_sx_sat

lum1n0us:
passing conversions.wast

### DIFF
--- a/core/iwasm/fast-jit/fe/jit_emit_conversion.c
+++ b/core/iwasm/fast-jit/fe/jit_emit_conversion.c
@@ -9,16 +9,148 @@
 #include "../jit_codegen.h"
 #include "../jit_frontend.h"
 
-static double
-uint64_to_double(uint64 u64)
+#define F32_I32_S_MIN (-2147483904.0f)
+#define F32_I32_S_MAX (2147483648.0f)
+#define F32_I32_U_MIN (-1.0f)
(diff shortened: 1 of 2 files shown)


core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
```diff
@@ -352,7 +352,7 @@ cmp_r_and_jmp_label(JitCompContext *cc, x86::Assembler &a,
 
     bool fp_cmp = cc->last_cmp_on_fp;
 
-    bh_assert(!fp_cmp || (fp_cmp && (op == GES)));
+    bh_assert(!fp_cmp || (fp_cmp && (op == GES || op == GTS)));
```
wenyongh, line 355: Will this op can be GTS? I remember here only GES is introduced by jit_emit_conversion.c, GTS is not required. And if GTS for fp_cmp is used, we should change the handling of `case GTS` below: currently it doesn't check fp_cmp like `case GES`.

lum1n0us: theoretical speaking. we should support float-point comparison with both `GTS` and `GES`. 
Since we are planning to refractory this part with the new `CMPCC` IR.  I'd like to take this change back.

PR #1218, opened 2022-06-09
Enable opaque pointer for aot compiler to fit LLVM 14/15

wenyongh:
Enable opaque pointer for aot compiler to meet the requirements
of LLVM 15, modify LLVMBuildLoad/LLVMBuildCall/LLVMInBoundsGEP
to LLVMBuildLoad2/LLVMBuildCall2/LLVMInBoundsGEP2, and pass
the related type.
And update the build script to build LLVM 15 by default.


lum1n0us:
It seems some c-api are deprecated and should move to their new version
https://llvm.org/docs/OpaquePointers.html#frontends

PR #1220, opened 2022-06-09
To avoid modification on the register which is not destination

lum1n0us:


core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
```diff
@@ -3162,7 +3181,11 @@ alu_imm_r_to_r_f64(x86::Assembler &a, ALU_OP op, int32 reg_no_dst,
     /* imm -> gp -> xmm */
     mov_imm_to_r_f64(a, reg_no_dst, data1_src);
 
-    return alu_r_m_float(a, op, reg_no_dst, cache, false);
+    alu_r_m_float(a, op, reg_no_dst, cache, false);
+
+    /* restore xmm */
+    a.movupd(regs_float[reg_no2_src], cache);
```
wenyongh, line 3187: Same as above, should not restore xmm here?

wenyongh, line 3187: Same as above, should not restore xmm here

lum1n0us: because  we move `imm` to the `xmm` after moving `xmm` value to memory. So gona to restore the xmm when it is done

lum1n0us: are you suggesting something like:
```suggestion
    if (reg_no_dst != reg_no2_src)
        a.movupd(regs_float[reg_no2_src], cache);
```

wenyongh: My understanding is that we should not add this sentence at all, the code is:
```C
a.movupd(cache, regs_float[reg_no2_src]);
mov_imm_to_r_f64(a, reg_no_dst, data1_src);
alu_r_m_float(a, op, reg_no_dst, cache, false);
```
Two situations:
1. reg_no_dst != reg_no2_src
In this situation, reg_no2_src isn't changed, no need to store it.

2. reg_no_dst == reg_no2_src
In this situation, reg_no2_src is changed, but it is also the destination register, now it stores the result, we must not recover it, or the result will be overwritten.

lum1n0us: Oh, I see your point.

PR #1223, opened 2022-06-13
Fix fast jit issues

wenyongh:
Fix br if not clear_value issue
Fix i64.DIV/REM issue: i64 result rax/rdx may be overwritten by following i32 load eax/edx

### DIFF
--- a/core/iwasm/fast-jit/fe/jit_emit_numberic.c
+++ b/core/iwasm/fast-jit/fe/jit_emit_numberic.c
@@ -512,7 +512,7 @@ compile_int_div_no_check(JitCompContext *cc, IntArithmetic arith_op,
         case INT_DIV_S:
         case INT_DIV_U:
         {
-            JitInsn *insn = NULL;
+            JitInsn *insn = NULL, *insn1 = NULL;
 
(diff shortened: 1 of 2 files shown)


core/iwasm/fast-jit/fe/jit_emit_numberic.c
```diff
@@ -512,57 +512,75 @@ compile_int_div_no_check(JitCompContext *cc, IntArithmetic arith_op,
...
                     insn = GEN_INSN(REM_S, rdx_hreg, rax_hreg, right);
                 else
                     insn = GEN_INSN(REM_U, rdx_hreg, rax_hreg, right);
-
-                res = rdx_hreg;
             }
 
             jit_lock_reg_in_insn(cc, insn, eax_hreg);
             jit_lock_reg_in_insn(cc, insn, edx_hreg);
+
```
lum1n0us, line 571: seems `jit_lock_reg_in_insn` only locks the *eax_hreg* and *edx_hreg* in one instruction, and it is pretty clever to move the result to another register.

lum1n0us:
LGTM

PR #1224, opened 2022-06-13
Add a sanitizer to check if there is a definition of vreg before its

lum1n0us:
reference

### DIFF
--- a/core/iwasm/fast-jit/jit_regalloc.c
+++ b/core/iwasm/fast-jit/jit_regalloc.c
@@ -6,6 +6,10 @@
 #include "jit_utils.h"
 #include "jit_compiler.h"
 
+#if BH_DEBUG != 0
+#define VREG_DEF_SANITIZER
+#endif
+
 /**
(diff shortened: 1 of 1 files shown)


core/iwasm/fast-jit/jit_regalloc.c
```diff
@@ -350,6 +354,29 @@ is_alloc_candidate(JitCompContext *cc, JitReg reg)
...
+#ifdef VREG_DEF_SANITIZER
+static void
+check_vreg_definition(RegallocContext *rc, JitInsn *insn)
+{
+    unsigned j;
+    JitRegVec regvec = jit_insn_opnd_regs(insn);
+    unsigned first_use = jit_insn_opnd_first_use(insn);
+    JitReg *regp;
+
+    /* to check if a vr has a definition before it is refernced */
```
wenyongh, line 366: refernced -> referenced
How about change to "check if a vr has a definition before it is referenced"? No need to add `to` as it is what we are currently doing.

PR #1228, opened 2022-06-15
Fix fast jit issues and clear compile warnings

wenyongh:
Fix i8 bool result misused as i32 reg to compare with i32 0
Fix wasm_runtime_malloc 0 size memory warning
Fix codegen i64 ROTL translation issue
Refine rotate shift operations
Clear compilation warnings of codegen


lum1n0us:
LGTM

PR #1233, opened 2022-06-16
fix some kw issues

xujuntwt95329:


core/iwasm/aot/aot_loader.c
```diff
@@ -149,11 +149,11 @@ GET_U64_FROM_ADDR(uint32 *addr)
...
 
-#define read_byte_array(p, p_end, addr, len) \
-    do {                                     \
-        CHECK_BUF(p, p_end, len);            \
-        memcpy(addr, p, len);                \
-        p += len;                            \
+#define read_byte_array(p, p_end, addr, len)  \
+    do {                                      \
+        CHECK_BUF(p, p_end, len);             \
+        bh_memcpy_s(addr, p_end - p, p, len); \
```
wenyongh, line 155: Seems should be `bh_memcpy_s(addr, len, p, len)` according to the source code

PR #1234, opened 2022-06-17
the isatty definition on nuttx depends on CONFIG_SERIAL_TERMIOS

dongsheng28849455:
the isatty definition on nuttx depends on CONFIG_SERIAL_TERMIOS

### DIFF
--- a/core/shared/platform/nuttx/platform_internal.h
+++ b/core/shared/platform/nuttx/platform_internal.h
@@ -68,6 +68,13 @@ typedef pthread_t korp_thread;
 #define O_NOFOLLOW 0
 #endif
 
+#undef CONFIG_HAS_ISATTY
+#ifdef CONFIG_SERIAL_TERMIOS
+#define CONFIG_HAS_ISATTY 1
+#else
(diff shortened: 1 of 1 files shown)


wenyongh:
@no1wudi Could you please help review the PR?

core/shared/platform/nuttx/platform_internal.h
```diff
@@ -68,6 +68,13 @@ typedef pthread_t korp_thread;
 #define O_NOFOLLOW 0
 #endif
 
+#undef CONFIG_HAS_ISATTY
```
no1wudi, line 71: Is it enough to override globally ?

dongsheng28849455: CONFIG_HAS_ISATTY is defind in ssp_config.h to indicate that isatty is enabled, and only refered once in posix.c. ssp_config.h including is always ahead of bh_platform.h therefore all of them would be overridden

no1wudi (approved):

yamt:
https://github.com/bytecodealliance/wasm-micro-runtime/pull/1250

PR #1235, opened 2022-06-17
some symbols are missing when loading the xtensa aot file

dongsheng28849455:
some symbols are missing when loading the xtensa aot file

### DIFF
--- a/core/iwasm/aot/arch/aot_reloc_xtensa.c
+++ b/core/iwasm/aot/arch/aot_reloc_xtensa.c
@@ -22,6 +22,26 @@ void __modsi3();
 
 void __divdi3();
 
+void __udivdi3();
+void __unorddf2();
+void __adddf3();
+void __eqdf2();
+void __muldf3();
+void __gedf2();
+void __ledf2();
+void __fixunsdfsi();
+void __floatunsidf();
+void __subdf3();
+void __nedf2();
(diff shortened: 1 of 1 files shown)


core/iwasm/aot/arch/aot_reloc_xtensa.c
```diff
@@ -40,6 +60,28 @@ static SymbolMap target_sym_map[] = {
...
+    REG_SYM(__fixdfsi),
+    REG_SYM(__moddi3),
+    REG_SYM(__extendsfdf2),
+    REG_SYM(__truncdfsf2),
+    REG_SYM(__gtdf2),
+    REG_SYM(__umoddi3),
+    REG_SYM(__floatdidf),
+    REG_SYM(__divsf3),
+    REG_SYM(sqrt),
+    REG_SYM(sqrtf),
```
yamt, line 84: it seems awkward to export sqrt and sqrtf this way.
is there xtensa specific reasons to do this?

dongsheng28849455: no specific reasons, just find it in test. let me check.

yamt: i suspect your original module (before aot compilation) had these undefined symbols.

PR #1237, opened 2022-06-18
support integrate 3rd-party toolchains into wamrc

xujuntwt95329:


core/iwasm/compilation/aot_emit_aot_file.c
```diff
@@ -2692,8 +2692,38 @@ aot_obj_data_create(AOTCompContext *comp_ctx)
     memset(obj_data, 0, sizeof(AOTObjectData));
 
     bh_print_time("Begin to emit object file");
+    if (comp_ctx->external_llc_compiler || comp_ctx->external_asm_compiler) {
+#if !(defined(_WIN32) || defined(_WIN32_))
```
wenyongh, line 2696: Had better add `aot_set_last_error and goto fail` like L2728 when it is under windows

xujuntwt95329: Done, thanks

PR #1241, opened 2022-06-20
Make robust on choosing X86_32 rather than an assumption

cngzhnp:
Choosing right target was decided by checking `CMAKE_SIZEOF_VOID_P` variable. However, choosing `X86_32` platform is not doing specifically checking size of void pointer. It is kind a fallback target for others.

Somehow, `CMAKE_SIZEOF_VOID_P` was not set by CMake in the different environment. If not, it should throw an error to user to warn them something is wrong on the environment.

Expectation is, if `CMAKE_SIZEOF_VOID_P` was defined and set to **4**, it should be `X86_32`. Otherwise, it should generate an error.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,9 +25,11 @@ if (NOT DEFINED WAMR_BUILD_TARGET)
   elseif (CMAKE_SIZEOF_VOID_P EQUAL 8)
     # Build as X86_64 by default in 64-bit platform
     set (WAMR_BUILD_TARGET "X86_64")
-  else ()
+  elseif (CMAKE_SIZEOF_VOID_P EQUAL 4)
     # Build as X86_32 by default in 32-bit platform
     set (WAMR_BUILD_TARGET "X86_32")
+  else ()
+    message(SEND_ERROR "Unsupported build target platform!")
   endif ()
 endif ()
 
--- a/build-scripts/runtime_lib.cmake
+++ b/build-scripts/runtime_lib.cmake
@@ -41,9 +41,11 @@ if (NOT DEFINED WAMR_BUILD_TARGET)
     elseif (CMAKE_SIZEOF_VOID_P EQUAL 8)
         # Build as X86_64 by default in 64-bit platform
         set (WAMR_BUILD_TARGET "X86_64")
-    else ()
+    elseif (CMAKE_SIZEOF_VOID_P EQUAL 4)
         # Build as X86_32 by default in 32-bit platform
         set (WAMR_BUILD_TARGET "X86_32")
+    else ()
+        message(SEND_ERROR "Unsupported build target platform!")
     endif ()
 endif ()
 
--- a/product-mini/platforms/android/CMakeLists.txt
+++ b/product-mini/platforms/android/CMakeLists.txt
@@ -35,9 +35,11 @@ if (NOT DEFINED WAMR_BUILD_TARGET)
   if (CMAKE_SIZEOF_VOID_P EQUAL 8)
     # Build as X86_64 by default in 64-bit platform
     set (WAMR_BUILD_TARGET "X86_64")
-  else ()
+  elseif (CMAKE_SIZEOF_VOID_P EQUAL 4)
     # Build as X86_32 by default in 32-bit platform
     set (WAMR_BUILD_TARGET "X86_32")
+  else ()
+    message(SEND_ERROR "Unsupported build target platform!")
   endif ()
 endif ()
 
--- a/product-mini/platforms/darwin/CMakeLists.txt
+++ b/product-mini/platforms/darwin/CMakeLists.txt
@@ -22,9 +22,11 @@ if (NOT DEFINED WAMR_BUILD_TARGET)
   elseif (CMAKE_SIZEOF_VOID_P EQUAL 8)
     # Build as X86_64 by default in 64-bit platform
     set (WAMR_BUILD_TARGET "X86_64")
-  else ()
+  elseif (CMAKE_SIZEOF_VOID_P EQUAL 4)
     # Build as X86_32 by default in 32-bit platform
     set (WAMR_BUILD_TARGET "X86_32")
+  else ()
+    message(SEND_ERROR "Unsupported build target platform!")
   endif ()
 endif ()
 
--- a/product-mini/platforms/linux-sgx/CMakeLists.txt
+++ b/product-mini/platforms/linux-sgx/CMakeLists.txt
@@ -16,9 +16,11 @@ if (NOT DEFINED WAMR_BUILD_TARGET)
   if (CMAKE_SIZEOF_VOID_P EQUAL 8)
     # Build as X86_64 by default in 64-bit platform
     set (WAMR_BUILD_TARGET "X86_64")
-  else ()
+  elseif (CMAKE_SIZEOF_VOID_P EQUAL 4)
     # Build as X86_32 by default in 32-bit platform
     set (WAMR_BUILD_TARGET "X86_32")
+  else ()
+    message(SEND_ERROR "Unsupported build target platform!")
   endif ()
 endif ()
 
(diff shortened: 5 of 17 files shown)


build-scripts/config_common.cmake
```diff
@@ -52,23 +52,25 @@ if (CMAKE_BUILD_TYPE STREQUAL "Debug")
...
+  if (NOT WAMR_BUILD_PLATFORM STREQUAL "windows")
+    # Add -fPIC flag if build as 64-bit
+    if (CMAKE_SIZEOF_VOID_P EQUAL 8)
       set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
       set (CMAKE_SHARED_LIBRARY_LINK_C_FLAGS "${CMAKE_SHARED_LIBRARY_LINK_C_FLAGS} -fPIC")
     endif ()
-  else ()
+  endif ()
+else ()
+  if (CMAKE_SIZEOF_VOID_P EQUAL 4)
```
wenyongh, line 65: The change is invalid: when host is 64-bit (CAMKE_SIZEOF_VOID_P is 8) and target is set x86-32, it won't be handled, and flag `-m32` isn't passed to GCC, which causes compilation error, see CI: https://github.com/bytecodealliance/wasm-micro-runtime/runs/6964367552?check_suite_focus=true#step:6:138

How about not changing the handling of case `if (CMAKE_SIZEOF_VOID_P EQUAL 8)`, and adding the handling of case `(CMAKE_SIZEOF_VOID_P EQUAL 4)`:
```cmake
if (CMAKE_SIZEOF_VOID_P EQUAL 8)
   keep unchanged
elseif (CMAKE_SIZEOF_VOID_P EQUAL 4)
  if (WAMR_BUILD_TARGET STREQUAL "X86_32")
      add handing if needed
  endif()
else ()
   message(SEND_ERROR "Unsupported build target platform!")
endif ()
```

samples/littlevgl/vgl-native-ui-app/CMakeLists.txt
```diff
@@ -11,16 +11,14 @@ SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS}  -DPLATFORM_NATIVE_LINUX -DUSE_MONITOR -DUSE
...
-  if (${BUILD_AS_64BIT_SUPPORT} STREQUAL "YES")
-    # Add -fPIC flag if build as 64-bit
-    set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
-    set (CMAKE_SHARED_LIBRARY_LINK_C_FLAGS "${CMAKE_SHARED_LIBRARY_LINK_C_FLAGS} -fPIC")
-  else ()
-    add_definitions (-m32)
-    set (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -m32")
-    set (CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -m32")
-  endif ()
+if (${BUILD_AS_64BIT_SUPPORT} STREQUAL "YES")
```
wenyongh, line 14: Same as above

wamr-compiler/CMakeLists.txt
```diff
@@ -95,19 +97,21 @@ endif ()
 
 message ("-- Build as target ${WAMR_BUILD_TARGET}")
 
-if (CMAKE_SIZEOF_VOID_P EQUAL 8)
-    if (WAMR_BUILD_TARGET STREQUAL "X86_64" OR WAMR_BUILD_TARGET STREQUAL "AMD_64"
-        OR WAMR_BUILD_TARGET MATCHES "AARCH64.*" OR WAMR_BUILD_TARGET MATCHES "RISCV64.*")
-    if (NOT WAMR_BUILD_PLATFORM STREQUAL "windows")
-      # Add -fPIC flag if build as 64-bit
+if (WAMR_BUILD_TARGET STREQUAL "X86_64" OR WAMR_BUILD_TARGET STREQUAL "AMD_64"
```
wenyongh, line 100: Same as above

build-scripts/config_common.cmake
cngzhnp: Thanks for the hint, I updated the latest commit of mine after your suggestion. Could you please check & verify that?

build-scripts/config_common.cmake
```diff
@@ -60,17 +60,20 @@ if (CMAKE_SIZEOF_VOID_P EQUAL 8)
...
-    endif ()
+  endif()
+elseif (CMAKE_SIZEOF_VOID_P EQUAL 4)
+  include(CheckCCompilerFlag)
+  Check_C_Compiler_Flag(-m32 M32_OK)
+  if (M32_OK OR WAMR_BUILD_TARGET STREQUAL "X86_32")
+    add_definitions (-m32)
+    set (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -m32")
+    set (CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -m32")
   endif ()
```
wenyongh, line 71: Should not remove the `-m32` flags in the process of branch `if (CMAKE_SIZEOF_VOID_P EQUAL 8)`: size of void pointer is 8, meaning that the host is 64-bit, and when compiling 32-bit targets in 64-bit host, e.g. compile x86-32 in ubuntu 64-bit, we need to add the `-m32` flag, or compilation error will occur.

And if `(CMAKE_SIZEOF_VOID_P EQUAL 4)`, meaning that host is 32-bit, and when compiling 32-bit targets in it, we don't need to add `-m32` flag.

Seems that we don't need to change the script, could you please check whether it works for you if there is no modification in this file?

samples/littlevgl/vgl-native-ui-app/CMakeLists.txt
```diff
@@ -16,11 +16,13 @@ if (CMAKE_SIZEOF_VOID_P EQUAL 8)
     # Add -fPIC flag if build as 64-bit
     set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
     set (CMAKE_SHARED_LIBRARY_LINK_C_FLAGS "${CMAKE_SHARED_LIBRARY_LINK_C_FLAGS} -fPIC")
-  else ()
-    add_definitions (-m32)
-    set (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -m32")
-    set (CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -m32")
   endif ()
+elseif (CMAKE_SIZEOF_VOID_P EQUAL 4)
```
wenyongh, line 20: Same as above

wamr-compiler/CMakeLists.txt
```diff
@@ -96,18 +98,20 @@ endif ()
...
       # Add -fPIC flag if build as 64-bit
       set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
       set (CMAKE_SHARED_LIBRARY_LINK_C_FLAGS "${CMAKE_SHARED_LIBRARY_LINK_C_FLAGS} -fPIC")
     endif ()
-  else ()
-    add_definitions (-m32)
-    set (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -m32")
-    set (CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -m32")
   endif ()
+elseif (CMAKE_SIZEOF_VOID_P EQUAL 4)
```
wenyongh, line 109: Same as above

build-scripts/config_common.cmake
cngzhnp: OK, I checked twice and probably we do not need to change this script file that mentioned above. Now, I will try again that everything is working fine with it.

PR #1244, opened 2022-06-23
enable dump call stack to a buffer

xujuntwt95329:


core/iwasm/aot/aot_runtime.c
```diff
@@ -3059,14 +3049,96 @@ aot_dump_call_stack(WASMExecEnv *exec_env)
         frame.module_offset = 0;
         frame.func_index = cur_frame->func_index;
         frame.func_offset = 0;
+        frame.func_name_wp =
+            get_func_name_from_index(module_inst, cur_frame->func_index);
 
         if (!bh_vector_append(module_inst->frames.ptr, &frame)) {
             bh_vector_destroy(module_inst->frames.ptr);
```
wenyongh, line 3056: Should set module_inst->frame.ptr to NULL, or when next time aot_create_call_stack, L3039, the vector will be destroyed again with invalid pointer.

core/iwasm/aot/aot_runtime.c
```diff
@@ -3059,14 +3049,96 @@ aot_dump_call_stack(WASMExecEnv *exec_env)
...
+            if (string_len >= remain_len) {                            \
+                /* Buffer full */                                      \
+                return len;                                            \
+            }                                                          \
+            total_len += string_len;                                   \
+        }                                                              \
+        else {                                                         \
+            total_len += strlen(line_buf);                             \
+        }                                                              \
+    } while (0)
```
wenyongh, line 3084: The macro is called three times, how about changing it to a function to reduce the footprint? e.g.
```
static uint32
dump_call_stack(const char *line_buf, bool dump_or_print, char *buf, uint32 len)

buf is the place to output the line_buf, not the beginning of total buffer
len is the left size of the buf, not the total size of the buf
return value: the size of buffer printed
```

core/iwasm/aot/aot_runtime.c
```diff
@@ -3059,14 +3049,96 @@ aot_dump_call_stack(WASMExecEnv *exec_env)
...
+
+    snprintf(line_buf, sizeof(line_buf), "\n");
+    PRINT_OR_DUMP();
+
+    while (n < total_frames) {
+        WASMCApiFrame frame = { 0 };
+        uint32 line_length, i;
+
+        if (!bh_vector_get(module_inst->frames.ptr, n, &frame)) {
+            return 0;
```
wenyongh, line 3112: There are some lines printed already, is it reasonable to return 0?

core/iwasm/aot/aot_runtime.h
```diff
@@ -731,8 +731,24 @@ aot_alloc_frame(WASMExecEnv *exec_env, uint32 func_index);
 void
 aot_free_frame(WASMExecEnv *exec_env);
 
-void
-aot_dump_call_stack(WASMExecEnv *exec_env);
+bool
+aot_create_call_stack(struct WASMExecEnv *exec_env);
+
+/**
+ * @brief dump aot call stack or get the size
```
wenyongh, line 738: Dump?

core/iwasm/aot/aot_runtime.h
```diff
@@ -731,8 +731,24 @@ aot_alloc_frame(WASMExecEnv *exec_env, uint32 func_index);
...
+ *
+ * @param exec_env the execution environment
+ * @param print print to stdout
```
wenyongh, line 741: Had better `whether to print to stdout or not`

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -2560,20 +2560,94 @@ wasm_interp_dump_call_stack(struct WASMExecEnv *exec_env)
             }
         }
 
+        frame.func_name_wp = func_name;
+
+        bh_vector_append(module_inst->frames, &frame);
```
wenyongh, line 2565: Should check return value, if false, destroy module_inst->frames, and set it to NULL?

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -2560,20 +2560,94 @@ wasm_interp_dump_call_stack(struct WASMExecEnv *exec_env)
...
+            if (string_len >= remain_len) {                            \
+                /* Buffer full */                                      \
+                return len;                                            \
+            }                                                          \
+            total_len += string_len;                                   \
+        }                                                              \
+        else {                                                         \
+            total_len += strlen(line_buf);                             \
+        }                                                              \
+    } while (0)
```
wenyongh, line 2592: Same as above, how about make it as function?

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -2560,20 +2560,94 @@ wasm_interp_dump_call_stack(struct WASMExecEnv *exec_env)
...
+
+    snprintf(line_buf, sizeof(line_buf), "\n");
+    PRINT_OR_DUMP();
+
+    while (n < total_frames) {
+        WASMCApiFrame frame = { 0 };
+        uint32 line_length, i;
+
+        if (!bh_vector_get(module_inst->frames, n, &frame)) {
+            return 0;
```
wenyongh, line 2622: Same as above, is it reasonable to return 0?

core/iwasm/interpreter/wasm_runtime.h
```diff
@@ -432,8 +432,25 @@ wasm_get_table_inst(const WASMModuleInstance *module_inst, const uint32 tbl_idx)
 }
 
 #if WASM_ENABLE_DUMP_CALL_STACK != 0
-void
-wasm_interp_dump_call_stack(struct WASMExecEnv *exec_env);
+bool
+wasm_interp_create_call_stack(struct WASMExecEnv *exec_env);
+
+/**
+ * @brief dump wasm call stack or get the size
```
wenyongh, line 439: Dump?

core/iwasm/interpreter/wasm_runtime.h
```diff
@@ -432,8 +432,25 @@ wasm_get_table_inst(const WASMModuleInstance *module_inst, const uint32 tbl_idx)
...
+ *
+ * @param exec_env the execution environment
+ * @param print print to stdout
```
wenyongh, line 442: Whether to print to stdout or not?

core/iwasm/aot/aot_runtime.c
xujuntwt95329: `dump_call_stack` should have the same implementation in both `aot_runtime.c` and `wasm_runtime.c`

how about put it into wasm_runtime_common and rename to `wasm_runtime_dump_call_stack_impl`?

core/iwasm/aot/aot_runtime.c
xujuntwt95329: `bh_vector_destroy` doesn't really free the frames.ptr pointer, it just free its internal data and memset the frames.ptr with 0

core/iwasm/aot/aot_runtime.c
wenyongh: The function name is a little long, how about `wasm_runtime_dump_line_buf`?

core/iwasm/aot/aot_runtime.c
wenyongh: OK, thanks for the notice, had better not change.

core/iwasm/aot/aot_runtime.c
xujuntwt95329: Sure, this function will not be exported to developer, we can add some suffix to distinguish, e.g. `wasm_runtime_dump_line_buf_impl`

core/iwasm/aot/aot_runtime.c
xujuntwt95329: 0 means error here, even if some lines has been dumped to buffer, the content is not complete, so we return 0 to indicate that the content in buf may be not valid.

core/iwasm/interpreter/wasm_runtime.c
xujuntwt95329: Done, thanks~

core/iwasm/aot/aot_runtime.h
xujuntwt95329: Done

core/iwasm/interpreter/wasm_runtime.c

core/iwasm/aot/aot_runtime.c
wenyongh: OK, got it, thanks.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -4527,22 +4527,87 @@ wasm_externref_retain(uint32 externref_idx)
...
 #endif
 #if WASM_ENABLE_AOT != 0
     if (module_inst->module_type == Wasm_Module_AoT) {
-        aot_dump_call_stack(exec_env);
+        aot_dump_call_stack(exec_env, true, NULL, 0);
     }
 #endif
 }
+
+uint32_t
```
wenyongh, line 4571: Had better change to uint32

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -4527,22 +4527,87 @@ wasm_externref_retain(uint32 externref_idx)
...
+#if WASM_ENABLE_AOT != 0
+    if (module_inst->module_type == Wasm_Module_AoT) {
+        return aot_dump_call_stack(exec_env, false, NULL, 0);
+    }
+#endif
+
+    return 0;
+}
+
+uint32_t
```
wenyongh, line 4591: uint32

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -4527,22 +4527,87 @@ wasm_externref_retain(uint32 externref_idx)
...
+wasm_runtime_dump_call_stack_to_buf(wasm_exec_env_t exec_env, char *buf,
+                                    uint32_t len)
```
wenyongh, line 4593: uint32

xujuntwt95329: Done, thanks!

PR #1260, opened 2022-06-30
add a dev container to WAMR repo

lum1n0us:


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,95 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.195.0/containers/cpp/.devcontainer/base.Dockerfile
```
wenyongh, line 1: Should add license header, like ci/Dockerfile

PR #1261, opened 2022-06-30
debug_engine: fix a few typos

yamt:
no functional changes are intended.

cf.
https://sourceware.org/gdb/onlinedocs/gdb/General-Query-Packets.html#General-Query-Packets


xujuntwt95329 (commented):
LGTM

PR #1263, opened 2022-07-01
Refine fe translate op br_if and br_table

wenyongh:
Refine fast jit frontend translation of opcode br_if and br_table:
1. for br_if, no need to clear jit frame after handling new basic block,
    so as to re-use registers of current basic block
2. for br_table, no need to create new basic block to jump if there is
    no parameters/results to copy to new block, just jumping to current
    existing basic block

### DIFF
--- a/core/iwasm/fast-jit/fe/jit_emit_control.c
+++ b/core/iwasm/fast-jit/fe/jit_emit_control.c
@@ -449,9 +449,24 @@ handle_op_end(JitCompContext *cc, uint8 **p_frame_ip, bool is_block_polymorphic)
         incoming_insn = block->incoming_insns_for_end_bb;
         while (incoming_insn) {
             insn = incoming_insn->insn;
-            bh_assert(insn->opcode == JIT_OP_JMP || insn->opcode == JIT_OP_BNE);
-            *(jit_insn_opnd(insn, incoming_insn->opnd_idx)) =
-                jit_basic_block_label(block->basic_block_end);
+            bh_assert(insn->opcode == JIT_OP_JMP || insn->opcode == JIT_OP_BNE
+                      || insn->opcode == JIT_OP_LOOKUPSWITCH);
+            if (insn->opcode == JIT_OP_JMP || insn->opcode == JIT_OP_BNE) {
+                *(jit_insn_opnd(insn, incoming_insn->opnd_idx)) =
+                    jit_basic_block_label(block->basic_block_end);
+            }
+            else {
+                /* Patch LOOKUPSWITCH INSN */
+                JitOpndLookupSwitch *opnd = jit_insn_opndls(insn);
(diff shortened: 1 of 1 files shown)


core/iwasm/fast-jit/fe/jit_emit_control.c
```diff
@@ -832,15 +865,51 @@ jit_compile_op_br(JitCompContext *cc, uint32 br_depth, uint8 **p_frame_ip)
...
+static JitFrame *
+jit_frame_clone(const JitFrame *jit_frame)
+{
+    JitFrame *jit_frame_cloned;
+    uint32 max_locals = jit_frame->max_locals;
+    uint32 max_stacks = jit_frame->max_stacks;
+    uint32 total_size;
+
+    total_size = (uint32)(offsetof(JitFrame, lp)
+                          + sizeof(*jit_frame->lp) * (max_locals + max_stacks));
```
lum1n0us, line 877: I don't have any opinion about the code. But just want to clarify that in a frame, all we actually care about is data in both `locals` and `stack`, right?

core/iwasm/fast-jit/fe/jit_emit_control.c
```diff
@@ -832,15 +865,51 @@ jit_compile_op_br(JitCompContext *cc, uint32 br_depth, uint8 **p_frame_ip)
...
+jit_frame_copy(JitFrame *jit_frame_dst, const JitFrame *jit_frame_src)
+{
+    uint32 max_locals = jit_frame_src->max_locals;
+    uint32 max_stacks = jit_frame_src->max_stacks;
+    uint32 total_size;
+
+    total_size =
+        (uint32)(offsetof(JitFrame, lp)
+                 + sizeof(*jit_frame_src->lp) * (max_locals + max_stacks));
+    bh_memcpy_s(jit_frame_dst, total_size, jit_frame_src, total_size);
```
lum1n0us, line 899: shall we consider the possibility of the overlap between `jit_frame_dst` and `jit_frame_src`?

core/iwasm/fast-jit/fe/jit_emit_control.c
wenyongh: No, not only locals and stacks, but also other info, e.g. frame_sp, virtual fixed registers, e.g. memory/table related registers. Since when clear_values at L972, most of the frame registers are set 0, we need to backup them firstly and then restore them after translating the dst basic block.

core/iwasm/fast-jit/fe/jit_emit_control.c
wenyongh: There should be no overlap, the memory of jit_frame_dst is allocated in jit_frame_clone, it is another memory block different from jit_frame_src.

lum1n0us:
LGTM

PR #1264, opened 2022-07-01
Refine fast jit backend translation of IR lookupswitch

wenyongh:
If the branches of lookupswitch is relatively big, lookup the dst
label address to jump from the address table but not compare
the value of branch one by one.

### DIFF
--- a/core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
+++ b/core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
@@ -278,8 +278,10 @@ local_log2l(uint64 data)
 
 /* Jmp type */
 typedef enum JmpType {
-    JMP_DST_LABEL,     /* jmp to dst label */
-    JMP_END_OF_CALLBC, /* jmp to end of CALLBC */
+    JMP_DST_LABEL_REL,     /* jmp to dst label with relative addr */
+    JMP_DST_LABEL_ABS,     /* jmp to dst label with absolute addr */
+    JMP_END_OF_CALLBC,     /* jmp to end of CALLBC */
+    JMP_LOOKUPSWITCH_BASE, /* LookupSwitch table base addr */
 } JmpType;
 
 /**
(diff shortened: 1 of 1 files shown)


core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
```diff
@@ -5432,33 +5434,96 @@ lookupswitch_r(JitCompContext *cc, x86::Assembler &a, bh_list *jmp_info_list,
...
+                GOTO_FAIL;
+
+            node->type = JMP_DST_LABEL_REL;
+            node->label_src = label_src;
+            node->dst_info.label_dst = jit_reg_no(opnd->match_pairs[i].target);
+            node->offset = a.code()->sectionById(0)->buffer().size() + 2;
+            bh_list_insert(jmp_info_list, node);
+
+            imm.setValue(INT32_MAX);
+            a.je(imm);
```
lum1n0us, line 5460: in another PR, after `je` to `INT32_MAX`, there is an error checking. Do we need it here?

wenyongh: It is because that I don't want to write offset to the invalid place when asmjit error occurs. Normally we check the asm error in jit_codegen_gen_native, after generating machine code for each INSN, in other place, we can skip the check to reduce the footprint.

lum1n0us: but if the error was raised, https://github.com/bytecodealliance/wasm-micro-runtime/blob/7b72af3d618fe24027854d68309ab990cc0ad62b/core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp#L6513 will be triggered?

wenyongh: Yes, it will be triggered, in the beginning of jit_codegen_gen_native, we set the error handler of code buffer:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/7b72af3d618fe24027854d68309ab990cc0ad62b/core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp#L6049-L6051

And the JitErrorHandler is defined here:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/7b72af3d618fe24027854d68309ab990cc0ad62b/core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp#L221-L234

PR #1265, opened 2022-07-01
Szadys

sophia-szady:


wenyongh:
@sophia-szady Thanks for uploading the PR! Not sure what is the purpose of the samples?
And could you please remove the unused files, e.g. the unnecessary CMake files, and the binary files?

loganek:
@sophia-szady looks like this PR got stuck and there was no progress on that since last year. Do you think we should close it? @wenyongh any thoughts on that?

wenyongh:
@loganek Yes, there wasn't update for a long time, let's close this PR. @sophia-szady if there is new update, please re-open this PR or create another PR, thanks.

PR #1267, opened 2022-07-05
Refine fast jit frontend: merge cmp and if/br_if

wenyongh:
Merge below INSNs translated by wasm opcodes `cmp + if` or `cmp + br_if`:
    CMP, SELECTcc, CMP, Bcc
into
    CMP, Bcc
So as to reduce the instructions

### DIFF
--- a/core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
+++ b/core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
@@ -381,7 +381,7 @@ cmp_r_and_jmp_label(JitCompContext *cc, x86::Assembler &a,
 
     bool fp_cmp = cc->last_cmp_on_fp;
 
-    bh_assert(!fp_cmp || (fp_cmp && (op == GES)));
+    bh_assert(!fp_cmp || (fp_cmp && (op == GTS || op == GES)));
 
     switch (op) {
         case EQ:
@@ -396,7 +396,10 @@ cmp_r_and_jmp_label(JitCompContext *cc, x86::Assembler &a,
         }
         case GTS:
         {
-            a.jg(imm);
+            if (fp_cmp)
+                a.ja(imm);
+            else
+                a.jg(imm);
             break;
         }
         case LES:
@@ -5149,8 +5152,7 @@ cmp_r_and_jmp_relative(JitCompContext *cc, x86::Assembler &a, COND_OP op,
                        int32 offset)
 {
     Imm target(INT32_MAX);
-    char *stream = (char *)a.code()->sectionById(0)->buffer().data()
-                   + a.code()->sectionById(0)->buffer().size();
+    char *stream;
     bool fp_cmp = cc->last_cmp_on_fp;
 
     bh_assert(!fp_cmp || (fp_cmp && (op == GTS || op == GES)));
@@ -5223,8 +5225,14 @@ cmp_r_and_jmp_relative(JitCompContext *cc, x86::Assembler &a, COND_OP op,
         }
     }
 
-    /* The offset written by asmjit is always 0, we patch it again */
-    *(int32 *)(stream + 2) = offset;
+    JitErrorHandler *err_handler = (JitErrorHandler *)a.code()->errorHandler();
+
+    if (!err_handler->err) {
+        /* The offset written by asmjit is always 0, we patch it again */
+        stream = (char *)a.code()->sectionById(0)->buffer().data()
+                 + a.code()->sectionById(0)->buffer().size() - 6;
+        *(int32 *)(stream + 2) = offset;
+    }
     return true;
 }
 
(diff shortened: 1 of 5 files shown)


core/iwasm/fast-jit/fe/jit_emit_control.c
```diff
@@ -164,9 +164,40 @@ load_block_results(JitCompContext *cc, JitBlock *block)
...
+static void
+get_last_cmp_and_selectcc(JitCompContext *cc, JitReg cond, JitInsn **p_insn_cmp,
+                          JitInsn **p_insn_select)
+{
+    JitInsn *insn = jit_basic_block_last_insn(cc->cur_basic_block);
+
+    if (insn && insn->prev && insn->prev->opcode == JIT_OP_CMP
+        && insn->opcode >= JIT_OP_SELECTEQ && insn->opcode <= JIT_OP_SELECTLEU
+        && *jit_insn_opnd(insn, 0) == cond
+        && jit_reg_is_i32_const(cc, *jit_insn_opnd(insn, 2), 1)
```
lum1n0us, line 190: what if there are *i64* in `SELECTcc` ?

core/iwasm/fast-jit/fe/jit_emit_control.c
```diff
@@ -601,7 +655,7 @@ bool
 jit_compile_op_block(JitCompContext *cc, uint8 **p_frame_ip,
                      uint8 *frame_ip_end, uint32 label_type, uint32 param_count,
                      uint8 *param_types, uint32 result_count,
-                     uint8 *result_types)
+                     uint8 *result_types, bool merge_cmp_and_if)
```
lum1n0us, line 658: I thought the merge action is by default. Why need to disable it ?

core/iwasm/fast-jit/fe/jit_emit_control.c
```diff
@@ -227,11 +264,26 @@ push_jit_block_to_stack_and_pass_params(JitCompContext *cc, JitBlock *block,
         }
         else {
             /* IF block with condition br insn */
-            if (!GEN_INSN(CMP, cc->cmp_reg, cond, NEW_CONST(I32, 0))
-                || !(insn = GEN_INSN(BNE, cc->cmp_reg,
-                                     jit_basic_block_label(basic_block), 0))) {
-                jit_set_last_error(cc, "generate cond br failed");
-                goto fail;
+            if (insn_select && insn_cmp) {
```
lum1n0us, line 267: I thought `get_last_cmp_and_selectcc` will do the optimization. So, what does this `if(){}` do?

core/iwasm/fast-jit/fe/jit_emit_control.c
wenyongh: For the float/double comparison opcodes, we might translate them into calling native function instead, but not `CMP + SELECTcc`

core/iwasm/fast-jit/fe/jit_emit_control.c
wenyongh: get_last_cmp_and_selectcc just returns insn_select and insn_cmp found, here we replace insn_select with Bcc.

core/iwasm/fast-jit/fe/jit_emit_control.c
lum1n0us: Got it. Even i64 comparison would return i32 value

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
```diff
@@ -5223,8 +5225,14 @@ cmp_r_and_jmp_relative(JitCompContext *cc, x86::Assembler &a, COND_OP op,
...
-    /* The offset written by asmjit is always 0, we patch it again */
-    *(int32 *)(stream + 2) = offset;
+    JitErrorHandler *err_handler = (JitErrorHandler *)a.code()->errorHandler();
+
+    if (!err_handler->err) {
+        /* The offset written by asmjit is always 0, we patch it again */
+        stream = (char *)a.code()->sectionById(0)->buffer().data()
+                 + a.code()->sectionById(0)->buffer().size() - 6;
+        *(int32 *)(stream + 2) = offset;
+    }
```
lum1n0us, line 5235: add a `else` branch to handle error?

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
```diff
@@ -5302,24 +5310,30 @@ lower_select(JitCompContext *cc, x86::Assembler &a, COND_OP op, JitReg r0,
...
+#define JMP_TO_LABEL(label_dst, label_src)                                 \
+    do {                                                                   \
+        if (label_is_ahead(cc, label_dst, label_src)) {                    \
+            JitErrorHandler *err_handler =                                 \
+                (JitErrorHandler *)a.code()->errorHandler();               \
+            int32 _offset;                                                 \
+            char *stream;                                                  \
+            Imm imm(INT32_MAX);                                            \
+            a.jmp(imm);                                                    \
+            if (!err_handler->err) {                                       \
```
lum1n0us, line 5322: What kind of scenario would lead to the error? L5322 and L5332 seem to suggest both situations are acceptable.

core/iwasm/fast-jit/fe/jit_emit_control.c
lum1n0us: In that case, shall we tell that by checking `JIT_REG_KIND_XXX`?

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
wenyongh: The error is handled in jit_codegen_gen_native, after generation machine code for each INSN, seem no need to handle it here?
```C
            if (err_handler.err) {
                jit_set_last_error_v(cc,
                                     "failed to generate native code for JIT "
                                     "opcode 0x%02x, ErrorCode is %u",
                                     insn->opcode, err_handler.err);
                GOTO_FAIL;
            }
```

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
wenyongh: a.jmp(imm) may return error. The asmjit might calls re-alloc the memory of its code buffer, I tested cases with valgrind, and found that it sometimes reported that `*(int32 *)(stream + 2) = _offset` writes the old/obsolete memory before re-alloc, so I refactored the code. Another issue is that if a.jmp fails, we will write offset the invalid place.

lum1n0us:
LGTM

PR #1270, opened 2022-07-06
Fix verification of addresses in the address pool

loganek:
The existing validation didn't work as expected; e.g. for address
pool: 8.8.8.8/24 the application had access to the 127.0.0.1 address (which
should not).

### DIFF
--- a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
+++ b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
@@ -3270,7 +3270,19 @@ compare_address(const struct addr_pool *addr_pool_entry, const char *addr)
         return false;
     }
 
-    uint32 mask = addr_pool_entry->mask;
+    const uint32 max_mask_value = 32;
+    /* no support for invalid mask values */
+    if (addr_pool_entry->mask > max_mask_value) {
+        return false;
+    }
+
+    /* convert mask number into 32-bit mask value, i.e. mask /24 will be
+    converted to 4294967040 (binary: 11111111 11111111 11111111 00000000) */
+    uint32 mask = 0;
+    for (int i = 0; i < addr_pool_entry->mask; i++) {
+        mask |= 1 << (max_mask_value - 1 - i);
+    }
+
     uint32 first_address = address & mask;
     uint32 last_address = address | (~mask);
     return first_address <= target && target <= last_address;


core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -3270,7 +3270,11 @@ compare_address(const struct addr_pool *addr_pool_entry, const char *addr)
         return false;
     }
 
-    uint32 mask = addr_pool_entry->mask;
+    uint32 mask = 0;
+    for (int i = 0; i < addr_pool_entry->mask; i++) {
+        mask |= 1 << (31 - i);
```
lum1n0us, line 3275: nice catch. Would u mind adding protection about `i < 32`?

loganek: I updated the code. We have 2 options either truncating the number (`min(32, addr_pool_entry->mask)`) or rejecting invalid values. I've chosen thee latter as I thought is more natural in that case and I think it's aligned with the behavior when the ip address is invalid. Let me know if you prefer truncation, I don't mind changing that.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -3270,7 +3270,19 @@ compare_address(const struct addr_pool *addr_pool_entry, const char *addr)
...
+    /* no support for invalid mask values */
+    if (addr_pool_entry->mask > max_mask_value) {
+        return false;
+    }
+
+    /* convert mask number into 32-bit mask value, i.e. mask /24 will be
+    converted to 4294967040 (binary: 11111111 11111111 11111111 00000000) */
+    uint32 mask = 0;
+    for (int i = 0; i < addr_pool_entry->mask; i++) {
+        mask |= 1 << (max_mask_value - 1 - i);
```
lum1n0us, line 3283: It doesn't feel right. `mask` is used to identify the subnetwork. `first_address` is the first value in the subnetwork which should end with a 0, `last_address` is the last value in the range.

For example:
/0   --> 0xFFFF FFFF 
/1   --> 0xFFFF FFFE
/2   --> 0xFFFF FFFC
/24 --> 0xFF00 0000 
/31 --> 0x8000 0000
/32 --> 0x0, actually it is used for broadcast

lum1n0us: you are right. `mask` should be the amount number of `1` of the subnet address.

PR #1277, opened 2022-07-08
Add spec tests on OSX

lum1n0us:
Can't make a simple modification to resolve the issue of accidentally meeting "EOF" on MacOS. Has to

- Upgrade test scripts to python3
- Use a new `TTYRunner` on MacOS
- Abandon parallel running on MacOS

The left problem is pretty much. The time consumption on MacOS is super long. It will take *1.75 - 2 hours* 😮  per workflow.

### DIFF
--- a/tests/wamr-test-suites/spec-test-script/runtest.py
+++ b/tests/wamr-test-suites/spec-test-script/runtest.py
@@ -17,6 +17,8 @@ import struct
 import math
 import traceback
 
+from tty_runner import TTYRunner
+
 IS_PY_3 = sys.version_info[0] == 3
 
 test_aot = False
@@ -54,6 +56,10 @@ if platform.system().find("CYGWIN_NT") >= 0:
 else:
     sep = "\r\n"
 rundir = None
+
+def create_a_runner(cmd, no_pty=False):
+    # return TTYRunner(cmd) if sys.platform == "darwin" else Runner(cmd, no_pty)
+    return TTYRunner(cmd)
 
 class Runner():
     def __init__(self, args, no_pty=False):
@@ -148,7 +154,8 @@ class Runner():
                 self.stdout.close()
             self.stdin = None
             self.stdout = None
-            sys.exc_clear()
+            if not IS_PY_3:
+                sys.exc_clear()
 
 def assert_prompt(runner, prompts, timeout, is_need_execute_result):
     # Wait for the initial prompt
@@ -233,6 +240,9 @@ PY_SKIP_TESTS = (
         # endianness
         '.const 0x1.fff' )
 
+# TODO: replace it with shlex.shlex.
+# we should not check malformed in this script, and should leave it to
+# the wast2wasm program or the interpter
 def read_forms(string):
     forms = []
     form = ""
@@ -619,8 +629,8 @@ def value_comparison(out, expected):
     if not expected:
         return False
 
-    assert(':' in out), "out should be in a form likes numbers:type, but {}".format(out)
-    assert(':' in expected), "expected should be in a form likes numbers:type, but {}".format(expected)
+    assert(':' in out), "out should be in a form likes numbers:type, but {}".format(out.encode('utf-8'))
+    assert(':' in expected), "expected should be in a form likes numbers:type, but {}".format(expected.encode('utf-8'))
 
     if 'v128' in out:
         return vector_value_comparison(out, expected)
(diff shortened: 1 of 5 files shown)


tests/wamr-test-suites/spec-test-script/runtest.py
```diff
@@ -1122,11 +1135,24 @@ def test_assert_with_exception(form, wast_tempfile, wasm_tempfile, aot_tempfile,
...
+                    print("==>s: {}".format(s))
                     while s:
                         res = re.match("[^\"]*\"([^\"]*)\"(.*)", s, re.DOTALL)
-                        f.write(res.group(1).replace("\\", "\\x").decode("string_escape"))
+                        binary_module = res.group(1)
+
+                        print("==>1. SZ:{} binary_module: \"{}\"".format(len(binary_module), binary_module))
+                        if IS_PY_3:
+                            binary_module = binary_module.replace("\\", "\\x").encode("raw_unicode_escape").decode("unicode_escape")
+                            binary_module = binary_module.encode("latin-1")
```
yamt, line 1145: maybe it's simpler to use codecs.escape_decode
```
import codecs

b = binary_module.replace("\\", "\\x")
(binary_module, _) = codecs.escape_decode(b)
```

tests/wamr-test-suites/spec-test-script/runtest.py
```diff
@@ -148,7 +154,8 @@ def cleanup(self):
                 self.stdout.close()
             self.stdin = None
             self.stdout = None
-            sys.exc_clear()
+            if not IS_PY_3:
+                sys.exc_clear()
```
yamt, line 157: is it worth keeping py2 case?
we already use python3 for all.py.

yamt:
> Abandon parallel running on MacOS

why?

tests/wamr-test-suites/spec-test-script/runtest.py
lum1n0us: Would you mind sharing your local modification about py2->py3?

lum1n0us:
> > Abandon parallel running on MacOS
> 
> why?

`mp` doesn't show enough stability on our OSX machine. How about yours?

tests/wamr-test-suites/spec-test-script/runtest.py
lum1n0us: thanks. It works. 
But I can not make sure it is a standard function of codecs because I just can't locate `codecs.escape_decode` in [python3.7](https://docs.python.org/3.7/library/codecs.html), [python 3.8](https://docs.python.org/fr/3.8/library/codecs.html), [python 3.10](https://docs.python.org/3.10/library/codecs.html#module-codecs). But I did try `"escape_decode" in dir(codecs)` and got a `True`. 

Would you mind helping to make sure of it?

yamt: it seems that it's undocumented. i didn't know that.
maybe it's better to stick with documented methods.

tests/wamr-test-suites/spec-test-script/runtest.py
yamt: i was talking about eg. https://github.com/bytecodealliance/wasm-micro-runtime/blob/9d689b3a22d3ccd341393b5f76a071c89879371e/tests/wamr-test-suites/test_wamr.sh#L394
not local modification.

yamt:
> > > Abandon parallel running on MacOS
> > 
> > 
> > why?
> 
> `mp` doesn't show enough stability on our OSX machine. How about yours?

i vaguely remember that it used to work before. but i haven't tried recently.

tests/wamr-test-suites/spec-test-script/runtest.py
lum1n0us: For now, `runtest.py` is running under python2.7

yamt: > For now, `runtest.py` is running under python2.7

yes.
but this PR is making it compatible with py3, isn't it?
my suggestion is to drop py2 support as it would be simpler.
but if you want to keep it for now, it's fine.

yamt:
i'm using a local ad-hoc py3 mod to run runtest.py on macos. (as python2 is not available here)
as the changes in this PR looks more complete, i hope it lands sooner.

PR #1279, opened 2022-07-08
Fix an issue about calling imported functions in a wrong exec_env

lum1n0us:
Sub-modules and the main module should use their own exec_env.

In a reported case, the sub-module had a different aux_stack size with
the main module. And the main module called an imported function in the
main module exec_env. And when the sub-module function sets aux_stack
top, it hitted the aux_stack_bottom and lead the error.

### DIFF
--- a/core/iwasm/interpreter/wasm_interp_classic.c
+++ b/core/iwasm/interpreter/wasm_interp_classic.c
@@ -858,6 +858,9 @@ wasm_interp_call_func_import(WASMModuleInstance *module_inst,
     WASMFunctionImport *func_import = cur_func->u.func_import;
     uint8 *ip = prev_frame->ip;
     char buf[128];
+    WASMExecEnv *sub_module_exec_env = NULL;
+    uint32 aux_stack_origin_boundary = 0;
+    uint32 aux_stack_origin_bottom = 0;
 
     if (!sub_func_inst) {
         snprintf(buf, sizeof(buf),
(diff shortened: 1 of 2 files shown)


core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -871,17 +872,18 @@ wasm_interp_call_func_import(WASMModuleInstance *module_inst,
...
+        return;
+    }
 
-    /* call function of sub-module*/
-    wasm_interp_call_func_bytecode(sub_module_inst, exec_env, sub_func_inst,
-                                   prev_frame);
+    wasm_interp_call_func_bytecode(sub_module_inst, sub_module_exec_env,
+                                   sub_func_inst, prev_frame);
 
     /* restore ip and module_inst */
```
wenyongh, line 885: Had better change comment to `/* restore ip */`

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -936,17 +937,18 @@ wasm_interp_call_func_import(WASMModuleInstance *module_inst,
...
+        return;
+    }
 
-    /* call function of sub-module*/
-    wasm_interp_call_func_bytecode(sub_module_inst, exec_env, sub_func_inst,
-                                   prev_frame);
+    wasm_interp_call_func_bytecode(sub_module_inst, sub_module_exec_env,
+                                   sub_func_inst, prev_frame);
 
     /* restore ip and module_inst */
```
wenyongh, line 950: Same as above

lum1n0us:
Missing the point that "has to copy parameters and copy returns from the preivous `exec_env` to the current `exec_env` if using two `exec_env`". Now I understood the reason why we used one `exec_env` before.

Does it exist a simple solution for this multiple module case?

wenyongh:
> Missing the point that "has to copy parameters and copy returns from the preivous `exec_env` to the current `exec_env` if using two `exec_env`". Now I understood the reason why we used one `exec_env` before.
> 
> Does it exist a simple solution for this multiple module case?

Yes, the function arguments are stored in prev_frame's next frame, or its outs_area, we can copy them to a buffer, and then call wasm_interp_call_wasm with the buffer, which also returns the results. And after call, we can copy the results from buffer to prev_frame->sp from the buffer.

Here is a rough sample:
[multi_module.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/9079220/multi_module.zip)

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -1633,11 +1633,28 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                 global = globals + global_idx;
                 global_addr = get_global_addr(global_data, global);
                 aux_stack_top = *(uint32 *)(frame_sp - 1);
-                if (aux_stack_top <= exec_env->aux_stack_boundary.boundary) {
+                if (
+/* clang-format off */
+#if WASM_ENABLE_MULTI_MODULE != 0
+                    aux_stack_top <= module->module->aux_stack_bottom
+                                         - module->module->aux_stack_size
```
wenyongh, line 1640: Had better not use aux_stack_size directly, in multi-thread, the aux stack is divided into many sub areas according to the setting of max thread num.

lum1n0us: SO, which is better? 

Multiple modules in varients `exec_env` and stacks or multiple modules in one `exec_env`

wenyongh: My opinion is the first one is better, refer to the patch I uploaded, just like calling a normal wasm function from native/host.

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -871,17 +874,35 @@ wasm_interp_call_func_import(WASMModuleInstance *module_inst,
...
 
-    /* replace exec_env's module_inst with sub_module_inst so we can
-       call it */
+    /* Switch exec_env but keep using the same one by replacing necessary
+     * variables */
+    sub_module_exec_env = wasm_runtime_get_exec_env_singleton(
+        (WASMModuleInstanceCommon *)sub_module_inst);
+    if (!sub_module_exec_env) {
+        wasm_set_exception(module_inst, "create singleton exec_env failed");
+        return;
```
wenyongh, line 883: Can we put L873-L875 `prev_frame->ip = NULL;`  after L 885, or here restore the `prev_frame->ip` before return? So that it isn't changed when exception is thrown, if exception handling is enabled, `prev_frame->ip` might be used after handling the exception.

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -936,17 +939,35 @@ wasm_interp_call_func_import(WASMModuleInstance *module_inst,
...
 
-    /* replace exec_env's module_inst with sub_module_inst so we can
-       call it */
+    /* Switch exec_env but keep using the same one by replacing necessary
+     * variables */
+    sub_module_exec_env = wasm_runtime_get_exec_env_singleton(
+        (WASMModuleInstanceCommon *)sub_module_inst);
+    if (!sub_module_exec_env) {
+        wasm_set_exception(module_inst, "create singleton exec_env failed");
+        return;
```
wenyongh, line 948: Same as above

PR #1284, opened 2022-07-11
runtest.py: Fix traceback print

yamt:


tests/wamr-test-suites/spec-test-script/all.py
```diff
@@ -155,7 +155,7 @@ def test_case(
         CMD,
         bufsize=1,
         stdout=subprocess.PIPE,
-        stderr=subprocess.PIPE,
+        stderr=subprocess.DEVNULL,
```
lum1n0us, line 158: if so, we won't receive error information, and it may limit our choice.

yamt: do you mean it's better to keep the unused stderr plumbing?

tests/wamr-test-suites/spec-test-script/runtest.py
```diff
@@ -1274,7 +1274,7 @@ def test_assert_with_exception(form, wast_tempfile, wasm_tempfile, aot_tempfile,
             else:
                 raise Exception("unrecognized form '%s...'" % form[0:40])
     except Exception as e:
-        traceback.print_exc()
+        traceback.print_exc(file=sys.stdout)
```
loganek, line 1277: hi @yamt , could you clarify what is the benefit of this change? Whereas I understand what it does, it's not quite clear why do we want to change it?

wenyongh:
Close this PR since it hasn't been updated for a long time, and the changes seem unneeded.

PR #1305, opened 2022-07-19
Refine memory load/store opcode translation

wenyongh:
Merge two JIT IRs:
  ADD maddr, memory_data, offset1
  LOAD/STORE value, maddr, 0
into one JIT IR:
  LOAD/STORE value, memory_data, offset1


lum1n0us:
LGTM

PR #1306, opened 2022-07-19
Refine fast-jit load func pointers

wenyongh:
Copy fast jit function pointers array from wasm module into wasm
module instance, and load fast jit func pointer from module instance
instead wasm module, so as to reduce one load operation when callbc.


lum1n0us:
LGTM

PR #1308, opened 2022-07-19
Add import subtyping validation

lum1n0us:
according to [import_subtyping_validation](https://webassembly.github.io/spec/core/valid/types.html#import-subtyping),
wasm-c-api needs to check types when linking.

### DIFF
--- a/core/iwasm/common/wasm_c_api.c
+++ b/core/iwasm/common/wasm_c_api.c
@@ -669,6 +669,48 @@ wasm_functype_results(const wasm_functype_t *func_type)
     return func_type->results;
 }
 
+static bool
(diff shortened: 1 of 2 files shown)


core/iwasm/common/wasm_c_api.c
```diff
@@ -3812,6 +3854,11 @@ interp_link_func(const wasm_instance_t *inst, const WASMModule *module_interp,
     imported_func_interp = module_interp->import_functions + func_idx_rt;
     bh_assert(imported_func_interp);
 
+    /* type comparision */
```
wenyongh, line 3857: Should be `comparison`

Issue #537, opened 2021-02-21
Passing NULL from WASM to Native


Paril:
I ran into this issue in the project I'm working on, where at times a native function with a pointer signature like `(*)` may be given NULL. On the native side, however, this value is offset by the WASM heap address (it's the same value as `wasm_runtime_addr_app_to_native(module_inst, 0)`. Is this an oversight or something I'm unaware of? Is it ever possible for this value to come in as a non-NULL value? Should WAMR just treat 0 in pointer sigs as 0 instead of translating them?

wenyongh:
Hi, do you mean that the wasm app passes argument NULL pointer to native API, and the NULL pointer is converted to 0 by the WASM toolchain, so the native API gets 0 as offset to linear memory, and in WAMR, if the native API's argument signature is configured with `*` as pointer, function `wasm_runtime_addr_app_to_native` converts it into the native address of linear memory's begin address, but not NULL?

Yes, we might treat 0 as NULL to return in function wasm_runtime_addr_app_to_native, but it might not be a common case, for C it might be OK to convert 0 to NULL, but for other languages, e.g. Rust/AssemblyScript, it might be invalid. The WASM spec defines that the linear memory address range is 0 to linear_mem_size - 1, so 0 is also a valid address, we had better treat it as normal address when converting it to native address. If you want to treat it as NULL, could you please convert the native pointer into app address and then check whether the app address is 0 instead? e.g.:
```C
foo(wasm_exec_env_t exec_env, void *ptr)
{
    wasm_module_inst_t module_inst = get_module_inst(exec_env);
    uint32 offset = addr_native_to_app(ptr);
    if (offset == 0) // treat as NULL pointer
       report error;
    ...
}
```
This should be OK for wasi-sdk toolchain as the linear memory's offset 0-7 are reserved as in the wasm generated, the first offset of init data is 8.

Paril:
Gotcha, the last sentence was what I was interested in (whether 0 was indeed suitable for a "null" offset) since I wasn't sure if it was legal for legitimate memory to live at that address.

I don't recall this being noted in the docs, so it might be worth noting that if you use * that NULL won't expand to zero. I ran into the gotcha real quick in my case and worked around it with the fix you describe!

PR #1319, opened 2022-07-21
Implement wasi_addr_resolve function

### LINKED ISSUE #1336: Extend support for sockets in WASI
## Motivation
We'd like to be able to build TCP and UDP clients that fully run on WebAssembly runtime. We'd like to be able to build clients like HTTP or low-latency UDP-based protocols (e.g. RTP). The current implementation covers a lot of functionality we need, but there are gaps that  block some of our scenarios.

## High-level approach
We identified two approaches for addressing gaps in the implementation
1. Continue the development of lib-sockets in WAMR and expose POSIX-like (slightly simplified though) WASI syscalls for missing functionality.
2. Implement a current version of the [proposal](https://github.com/WebAssembly/wasi-sockets) (https://github.com/WebAssembly/wasi-sockets). The proposal is still in development phase, and there’s a risk it will significantly change - if that’s the case, we’ll end up with a significant amount of throw-away work.

Because of uncertainty of the proposal, we recommend following approach 1. We will however monitor the state of the proposal. We will also consider implementing the proposal as experiment in WAMR and provide feedback to the proposal champion; however, that is not our priority right now.

### Features/API changes
We'd like to extend WAMR to support the following features:

### IPv6 support
Current implementation supports IPv4 only. even though the data structures that we have in WAMR are already designed to handle IPv6, it's not been implemented in any of the function (e.g. `socket_open`, `socket_connect` or `socket_bind`).
There's no API change required for that, but the internal implementation needs to be able to handle IPv6

### ns-lookup
WAMR already has a placeholder called `addr_resolve`, but there's no implementation for that. As part of this we'll provide implementation for the placeholder.

**Security considerations**
WebAssembly code runs in sandboxed environment, and we need to make sure that any interaction with external service needs to be explicitly enabled by the user. We'll start simple by adding `--allow-ns-lookup`. However, we'll consider more fine-grained control later on as described in the [current socket proposal](https://github.com/WebAssembly/wasi-sockets/blob/main/GrantingAccess.md)

**API**
We expect the API to slightly change to accommodate for additional features and reuse existing data structures instead of using raw buffers.
```
typedef struct __wasi_addr_info_t {
    __wasi_addr_t addr;
    __wasi_sock_type_t type;
} __wasi_addr_info_t;

typedef struct __wasi_addr_info_hints_t {
    __wasi_sock_type_t type;
    __wasi_address_family_t family;
    // this is to workaround lack of optional parameters
    uint8_t hints_enabled;
} __wasi_addr_info_hints_t;

__wasi_errno_t
__wasi_sock_addr_resolve(const char *host, const char *service,
                                             __wasi_addr_info_hints_t *hints,
                                             __wasi_addr_info_t *addr_info,
                                             __wasi_size_t addr_info_size,
                                             __wasi_size_t *max_info_size)
```

### Socket options
In POSIX there's a pair of `getsockopt` and `setsockopt` functions for reading and setting the configuration. We considered providing a 1:1 mapping to the options, however, I don't think we want to expose all of the options through the API. Also 1:1 mapping makes the API very difficult to use, as depending on the option, the type of value might differ, so that makes the API more error-prone. There's at least few of them that I think is worth exposing (for example, `SO_RCVTIMEO`/`SO_SNDTIMEO`). We'll provide a set of set_/get_ methods to access the properties (please note there are already placeholders in WAMR for that, e.g. `set/get_recv_buf_size`).
  
### UDP gaps
Whereas it's currently possible to build a UDP client/server application using `connect` method, we think there's a value of providing send-to/recv-from functionality for ad-hoc communication with different endpoints. Please note the implementation of send-to/recv-from can be re-used in send/recv methods by passing NULL as a destination.

**API**
```
__wasi_errno_t __wasi_sock_send_to(
    __wasi_fd_t fd,
    const __wasi_ciovec_t *si_data,
    size_t si_data_len,
    __wasi_siflags_t si_flags,
    const __wasi_addr_t *destination,
    __wasi_size_t *retptr0
);

__wasi_errno_t __wasi_sock_recv_from(
    __wasi_fd_t fd,
    const __wasi_iovec_t *ri_data,
    size_t ri_data_len,
    __wasi_riflags_t ri_flags,
    __wasi_addr_t *source,
    __wasi_size_t *retptr0
);
```

### MVP milestone
We define Minimum Viable Product (MVP) as a solution that implements all the functionality mentioned above for POSIX platform (for other platforms, the features can be left unsupported, and return appropriate error code when called). We update documentation and perform necessary tests to make sure the existing functionality was not impacted, and new functionality works as expected. We also update a library (socket-lib) so the functionality can be easily accessed from C programs. We implement simple security rules (`--allow-ns-lookup`) which will be extended in the future.

We expect MVP milestone to be at the quality bar to merge it to main branch and expose to users.

### Post-MVP milestone
For Post-MVP milestone we add the same functionality for other supported platforms (Windows, POSIX-SGX). We implement more granular security rules.

loganek:
Implement wasi_addr_resolve function
    
I also slightly modified the interface to make it more accessible for the
end-user:
 * replaced port argument with the service - so the user can actually get the port for a given service if unknown
 * introduced __wasi_addr_info_t and used it as a buffer for addresses, instead of generic buffer
 * introduced __wasi_addr_info_hints_t so users can enable filtering on the syscall level (and therefore use smaller buffers for addresses)
 * added max_size parameter for the API as an output - in case the number of addresses is bigger than the buffer size, user can repeat the call with bigger buffer

This change is very minimalistic, and it doesn't include the following:
 1. implementation of getaddrinfo in the lib-socket
 2. sample application
 
The 1. and 2. will be added in the following change (I did want to keep the change as small as possible for easier review. 3. can be implemented if needed, but I'm not really sure if that is required at all

Fixes #1336

### DIFF
--- a/core/iwasm/libraries/lib-socket/inc/wasi_socket_ext.h
+++ b/core/iwasm/libraries/lib-socket/inc/wasi_socket_ext.h
@@ -59,6 +59,18 @@ typedef struct __wasi_addr_t {
 
 typedef enum { INET4 = 0, INET6 } __wasi_address_family_t;
 
+typedef struct __wasi_addr_info_t {
+    __wasi_addr_t addr;
+    __wasi_sock_type_t type;
+} __wasi_addr_info_t;
+
+typedef struct __wasi_addr_info_hints_t {
+    __wasi_sock_type_t type;
+    __wasi_address_family_t family;
+    // this is to workaround lack of optional parameters
+    uint8_t hints_enabled;
+} __wasi_addr_info_hints_t;
+
 #ifdef __wasi__
 /**
  * Reimplement below POSIX APIs with __wasi_sock_XXX functions.
@@ -149,30 +161,37 @@ __wasi_sock_addr_remote(__wasi_fd_t fd, uint8_t *buf, __wasi_size_t buf_len)
 }
 
 /**
- * Resolves a hostname and a port to one or more IP addresses. Port is optional
- * and you can pass 0 (zero) in most cases, it is used a hint for protocol.
+ * Resolve a hostname and a service to one or more IP addresses. Service is
+ * optional and you can pass empty string in most cases, it is used as a hint
+ * for protocol.
  *
  * Note: This is similar to `getaddrinfo` in POSIX
  *
  * When successful, the contents of the output buffer consist of a sequence of
- * IPv4 and/or IPv6 addresses. Each address entry consists of a addr_t object.
+ * IPv4 and/or IPv6 addresses. Each address entry consists of a wasi_addr_t
+ * object.
  *
- * This function fills the output buffer as much as possible, potentially
- * truncating the last address entry. It is advisable that the buffer is
+ * This function fills the output buffer as much as possible, truncating the
+ * entries that didn't fit into the buffer. A number of available addresses
+ * will be returned through the last parameter.
  */
 int32_t
-__imported_wasi_snapshot_preview1_addr_resolve(int32_t arg0, int32_t arg1,
-                                               int32_t arg2, int32_t arg3,
-                                               int32_t arg4)
+__imported_wasi_snapshot_preview1_sock_addr_resolve(int32_t arg0, int32_t arg1,
+                                                    int32_t arg2, int32_t arg3,
+                                                    int32_t arg4, int32_t arg5)
     __attribute__((__import_module__("wasi_snapshot_preview1"),
-                   __import_name__("addr_resolve")));
+                   __import_name__("sock_addr_resolve")));
 
 static inline __wasi_errno_t
-__wasi_addr_resolve(__wasi_fd_t fd, const char *host, __wasi_ip_port_t port,
-                    uint8_t *buf, __wasi_size_t size)
+__wasi_sock_addr_resolve(const char *host, const char *service,
+                         __wasi_addr_info_hints_t *hints,
+                         __wasi_addr_info_t *addr_info,
+                         __wasi_size_t addr_info_size,
+                         __wasi_size_t *max_info_size)
 {
-    return (__wasi_errno_t)__imported_wasi_snapshot_preview1_addr_resolve(
-        (int32_t)fd, (int32_t)host, (int32_t)port, (int32_t)buf, (int32_t)size);
+    return (__wasi_errno_t)__imported_wasi_snapshot_preview1_sock_addr_resolve(
+        (int32_t)host, (int32_t)service, (int32_t)hints, (int32_t)addr_info,
+        (int32_t)addr_info_size, (int32_t)max_info_size);
 }
 
 /**
--- a/core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
+++ b/core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
@@ -1041,10 +1041,23 @@ wasi_sock_addr_remote(wasm_exec_env_t exec_env, wasi_fd_t fd, uint8 *buf,
 }
 
 static wasi_errno_t
-wasi_sock_addr_resolve(wasm_exec_env_t exec_env, wasi_fd_t fd, const char *host,
-                       wasi_ip_port_t port, uint8 *buf, wasi_size_t size)
+wasi_sock_addr_resolve(wasm_exec_env_t exec_env, const char *host,
+                       const char *service, __wasi_addr_info_hints_t *hints,
+                       __wasi_addr_info_t *addr_info,
+                       __wasi_size_t addr_info_size,
+                       __wasi_size_t *max_info_size)
 {
-    return __WASI_ENOSYS;
+    wasm_module_inst_t module_inst = get_module_inst(exec_env);
+    wasi_ctx_t wasi_ctx = get_wasi_ctx(module_inst);
+    struct fd_table *curfds = NULL;
+
+    if (!wasi_ctx)
+        return __WASI_EACCES;
+
+    curfds = wasi_ctx_get_curfds(module_inst, wasi_ctx);
+
+    return wasi_ssp_sock_addr_resolve(curfds, host, service, hints, addr_info,
+                                      addr_info_size, max_info_size);
 }
 
 static wasi_errno_t
@@ -1390,7 +1403,7 @@ static NativeSymbol native_symbols_libc_wasi[] = {
     REG_NATIVE_FUNC(sock_accept, "(i*)i"),
     REG_NATIVE_FUNC(sock_addr_local, "(i*i)i"),
     REG_NATIVE_FUNC(sock_addr_remote, "(i*i)i"),
-    REG_NATIVE_FUNC(sock_addr_resolve, "(i*i*i)i"),
+    REG_NATIVE_FUNC(sock_addr_resolve, "($$**i*)i"),
     REG_NATIVE_FUNC(sock_bind, "(i*)i"),
     REG_NATIVE_FUNC(sock_close, "(i)i"),
     REG_NATIVE_FUNC(sock_connect, "(i*)i"),
--- a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/include/wasmtime_ssp.h
+++ b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/include/wasmtime_ssp.h
@@ -598,6 +598,18 @@ typedef struct __wasi_addr_t {
 
 typedef enum { INET4 = 0, INET6 } __wasi_address_family_t;
 
+typedef struct __wasi_addr_info_t {
+    __wasi_addr_t addr;
+    __wasi_sock_type_t type;
+} __wasi_addr_info_t;
+
+typedef struct __wasi_addr_info_hints_t {
+   __wasi_sock_type_t type;
+   __wasi_address_family_t family;
+   // this is to workaround lack of optional parameters
+   uint8_t hints_enabled;
+} __wasi_addr_info_hints_t;
+
 #if defined(WASMTIME_SSP_WASI_API)
 #define WASMTIME_SSP_SYSCALL_NAME(name) \
     asm("__wasi_" #name)
@@ -1021,6 +1033,16 @@ wasi_ssp_sock_bind(
     struct fd_table *curfds, struct addr_pool *addr_pool,
 #endif
     __wasi_fd_t fd, __wasi_addr_t *addr
+) __attribute__((__warn_unused_result__));
+
+__wasi_errno_t
+wasi_ssp_sock_addr_resolve(
+#if !defined(WASMTIME_SSP_STATIC_CURFDS)
+    struct fd_table *curfds,
+#endif
+    const char *host, const char* service,
+    __wasi_addr_info_hints_t *hints, __wasi_addr_info_t *addr_info,
+    __wasi_size_t addr_info_size, __wasi_size_t *max_info_size
 ) __attribute__((__warn_unused_result__));
 
 __wasi_errno_t
--- a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
+++ b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
@@ -199,6 +199,35 @@ convert_clockid(__wasi_clockid_t in, clockid_t *out)
     }
 }
 
+// Converts an IPv4 binary address object to WASI address.
+static void
+ipv4_addr_to_wasi_addr(uint32_t addr, __wasi_ip_port_t port, __wasi_addr_t *out)
+{
+    out->kind = IPv4;
+    out->addr.ip4.port = port;
+    out->addr.ip4.addr.n3 = (addr & 0xFF000000) >> 24;
+    out->addr.ip4.addr.n2 = (addr & 0x00FF0000) >> 16;
+    out->addr.ip4.addr.n1 = (addr & 0x0000FF00) >> 8;
+    out->addr.ip4.addr.n0 = (addr & 0x000000FF);
+}
+
+// Converts an IPv6 binary address object to WASI address object.
+static void
+ipv6_addr_to_wasi_addr(uint16_t addr[8], __wasi_ip_port_t port,
+                       __wasi_addr_t *out)
+{
+    out->kind = IPv6;
+    out->addr.ip6.port = port;
+    out->addr.ip6.addr.n0 = addr[0];
+    out->addr.ip6.addr.n1 = addr[1];
+    out->addr.ip6.addr.n2 = addr[2];
+    out->addr.ip6.addr.n3 = addr[3];
+    out->addr.ip6.addr.h0 = addr[4];
+    out->addr.ip6.addr.h1 = addr[5];
+    out->addr.ip6.addr.h2 = addr[6];
+    out->addr.ip6.addr.h3 = addr[7];
+}
+
 __wasi_errno_t
 wasmtime_ssp_clock_res_get(__wasi_clockid_t clock_id,
                            __wasi_timestamp_t *resolution)
@@ -2912,6 +2941,56 @@ wasi_ssp_sock_bind(
         return convert_errno(errno);
     }
 
+    return __WASI_ESUCCESS;
+}
+
+__wasi_errno_t
+wasi_ssp_sock_addr_resolve(
+#if !defined(WASMTIME_SSP_STATIC_CURFDS)
+    struct fd_table *curfds,
+#endif
+    const char *host, const char *service, __wasi_addr_info_hints_t *hints,
+    __wasi_addr_info_t *addr_info, __wasi_size_t addr_info_size,
+    __wasi_size_t *max_info_size)
+{
+    bh_addr_info_t *wamr_addr_info =
+        wasm_runtime_malloc(addr_info_size * sizeof(bh_addr_info_t));
+    uint8_t hints_is_ipv4 = hints->family == INET4;
+    uint8_t hints_is_tcp = hints->type == SOCKET_STREAM;
+    size_t _max_info_size;
+    size_t actual_info_size;
+
+    if (!wamr_addr_info) {
+        return __WASI_ENOMEM;
+    }
+
+    int ret = os_socket_addr_resolve(
+        host, service, hints->hints_enabled ? &hints_is_tcp : NULL,
+        hints->hints_enabled ? &hints_is_ipv4 : NULL, wamr_addr_info,
+        addr_info_size, &_max_info_size);
+
+    if (ret != BHT_OK) {
+        wasm_runtime_free(wamr_addr_info);
+        return convert_errno(errno);
+    }
+
+    *max_info_size = _max_info_size;
+    actual_info_size =
+        addr_info_size < *max_info_size ? addr_info_size : *max_info_size;
+
+    for (size_t i = 0; i < actual_info_size; i++) {
+        addr_info[i].type = wamr_addr_info[i].is_tcp ? SOCK_STREAM : SOCK_DGRAM;
+        if (wamr_addr_info[i].is_ipv4) {
+            ipv4_addr_to_wasi_addr(*(uint32_t *)wamr_addr_info[i].addr,
+                                   wamr_addr_info[i].port, &addr_info[i].addr);
+        }
+        else {
+            ipv6_addr_to_wasi_addr((uint16_t *)wamr_addr_info[i].addr,
+                                   wamr_addr_info[i].port, &addr_info[i].addr);
+        }
+    }
+
+    wasm_runtime_free(wamr_addr_info);
     return __WASI_ESUCCESS;
 }
 
--- a/core/shared/platform/common/posix/posix_socket.c
+++ b/core/shared/platform/common/posix/posix_socket.c
@@ -7,6 +7,7 @@
 #include "platform_api_extension.h"
 
 #include <arpa/inet.h>
+#include <netdb.h>
 
 static void
 textual_addr_to_sockaddr(const char *textual, int port, struct sockaddr_in *out)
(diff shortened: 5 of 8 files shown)


core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/include/wasmtime_ssp.h
```diff
@@ -598,6 +598,18 @@ typedef struct __wasi_addr_t {
...
+typedef struct __wasi_addr_info_t {
+    __wasi_addr_t addr;
+    __wasi_sock_type_t type;
+} __wasi_addr_info_t;
+
+typedef struct __wasi_addr_info_hints_t {
+   __wasi_sock_type_t type;
+   __wasi_address_family_t family;
+   // this is to workaround lack of optional parameters
+   uint8_t hints_enabled;
```
lum1n0us, line 610: May I ask why we need it or need it to be optional?

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -199,6 +201,55 @@ convert_clockid(__wasi_clockid_t in, clockid_t *out)
...
+    out->addr.ip6.addr.n2 = ntohs(addr16[2]);
+    out->addr.ip6.addr.n3 = ntohs(addr16[3]);
+    out->addr.ip6.addr.h0 = ntohs(addr16[4]);
+    out->addr.ip6.addr.h1 = ntohs(addr16[5]);
+    out->addr.ip6.addr.h2 = ntohs(addr16[6]);
+    out->addr.ip6.addr.h3 = ntohs(addr16[7]);
+}
+
+static int
+convert_getaddrinfo_error_to_wasi_error_code(int error)
```
lum1n0us, line 237: shall we make the name shorter?

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -2915,6 +2966,84 @@ wasi_ssp_sock_bind(
...
+                // Allow only UDP and TCP
+                || (res->ai_socktype != SOCK_DGRAM
+                    && res->ai_socktype != SOCK_STREAM)
+                || (res->ai_protocol != IPPROTO_TCP
+                    && res->ai_protocol != IPPROTO_UDP)) {
+                res = res->ai_next;
+                continue;
+            }
+
+            switch (res->ai_family) {
```
lum1n0us, line 3010: since we have already reduced values of `res->ai_family` to  either `AF_INET` or `AF_INET6`, maybe a smile `if...else..` is enough

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -2915,6 +2966,84 @@ wasi_ssp_sock_bind(
...
+    }
+
+    ret = getaddrinfo(host, strlen(service) == 0 ? NULL : service,
+                      hints->hints_enabled ? &posix_hints : NULL, &result);
+    if (ret != 0) {
+        return convert_getaddrinfo_error_to_wasi_error_code(ret);
+    }
+
+    res = result;
+    while (res) {
```
lum1n0us, line 2996: maybe something like:
```suggestion
    while (res && addr_info_size > pos)) {
```

lum1n0us (commented):
there is always a concern about `getaddrinfo()` in the community. WASI is built using capability-based security principles. It requires making a statement before using any system resource or accessing any outside resource.

So, there is a `--addr-pool` in *iwasm* command line option for creating sockets.

I am a little worried about if any resource will be involved during `getaddrinfo()` and should be reclaimed first

lum1n0us:
Plus, is there any modification in `wasi_socket_ext.c` ?

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/include/wasmtime_ssp.h
loganek: I made it optional in case the user want to request all possibilities (e.g. both IPv4 or IPv6). In POSIX the way it's achieved is by specifying the `AF_UNSPEC`, but I don't think we want to add that in the `__wasi_address_family_t` enum. I'm open for suggestions though, but I think giving people ability to disable a filter is important.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
loganek: We want to go through the full list to know its size (so we can return it through `max_info_size` parameter).

loganek:
> Plus, is there any modification in wasi_socket_ext.c ?

I mentioned it in the description:

> This change is very minimalistic, and it doesn't include the following:
> 
> 1. implementation of getaddrinfo in the lib-socket

I'd like to keep this pull request small but still self-contained. I'll definitely make a change in was_socket_ext.c but I'd rather keep this PR small (so it's easier for review) and update lib-socket in the following PR.

loganek:
> there is always a concern about `getaddrinfo()` in the community. WASI is built using capability-based security principles. It requires making a statement before using any system resource or accessing any outside resource.
> 
> So, there is a `--addr-pool` in _iwasm_ command line option for creating sockets.
> 
> I am a little worried about if any resource will be involved during `getaddrinfo()` and should be reclaimed first

I think this one is tricky as depending on the system, getaddrinfo will require different set of capabilities (e.g. on linux getaddrinfo references files from `/etc`, which is probably different for linux). One option I think we can consider is having an explicit permission to do ns-lookups (like `--allow-ns-lookup`) which will enable users to use the function. If you agree with that, I can do the work in a follow-up PR. But I'm also open for other suggestions.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/include/wasmtime_ssp.h
lum1n0us: how about disabling filters when `hints` is NULL?

loganek: Unfortunately, that's not that straightforward in WASM. When passing NULL in WASM to the function, it is translated to an address in linear memory, which no longer is NULL. We could follow the advice in [this issue](https://github.com/bytecodealliance/wasm-micro-runtime/issues/537) and use `addr_native_to_app` to see if the value passed is NULL; however, I'm still a bit concerned, as per the comment in the issue, it might not work for other languages than C. Is there already a function in WAMR that takes optional parameters so we can take it as an example? Or do you have any other suggestion for handling optional parameters?

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -1041,10 +1041,22 @@ wasi_sock_addr_remote(wasm_exec_env_t exec_env, wasi_fd_t fd, uint8 *buf,
 }
 
 static wasi_errno_t
-wasi_sock_addr_resolve(wasm_exec_env_t exec_env, wasi_fd_t fd, const char *host,
-                       wasi_ip_port_t port, uint8 *buf, wasi_size_t size)
+wasi_addr_resolve(wasm_exec_env_t exec_env, const char *host,
```
wenyongh, line 1044: Not sure whether change the function name, isn't it related with socket? If yes, had better not change the name, and rename wasi_ssp_addr_resolve to wasi_ssp_sock_addr_resolve, and os_addr_resolve to os_sock_addr_resolve.

core/iwasm/libraries/lib-socket/inc/wasi_socket_ext.h
```diff
@@ -59,6 +59,18 @@ typedef struct __wasi_addr_t {
...
+    __wasi_addr_t addr;
+    __wasi_sock_type_t type;
+} __wasi_addr_info_t;
+
+typedef struct __wasi_addr_info_hints_t {
+    __wasi_sock_type_t type;
+    __wasi_address_family_t family;
+    // this is to workaround lack of optional parameters
+    uint8_t hints_enabled;
+} __wasi_addr_info_hints_t;
```
wenyongh, line 72: Have we checked whether the PR works in 32-bit platform? Or the layout is same in both 32-bit and 64-bit platforms?

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -2915,6 +2944,52 @@ wasi_ssp_sock_bind(
...
+wasi_ssp_addr_resolve(
+#if !defined(WASMTIME_SSP_STATIC_CURFDS)
+    struct fd_table *curfds,
+#endif
+    const char *host, const char *service, __wasi_addr_info_hints_t *hints,
+    __wasi_addr_info_t *addr_info, __wasi_size_t addr_info_size,
+    __wasi_size_t *max_info_size)
+{
+    wamr_addr_info_t *wamr_addr_info =
+        (wamr_addr_info_t *)malloc(addr_info_size * sizeof(wamr_addr_info_t));
```
wenyongh, line 2957: Don't use malloc, use wasm_runtime_malloc/wasm_runtime_free instead, some platforms may not support malloc/free.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -2915,6 +2944,52 @@ wasi_ssp_sock_bind(
...
+    uint8_t hints_is_ipv4 = hints->family == INET4;
+    uint8_t hints_is_tcp = hints->type == SOCKET_STREAM;
+    size_t _max_info_size;
+    size_t actual_info_size;
+
+    int ret = os_addr_resolve(host, service,
+                              hints->hints_enabled ? &hints_is_tcp : NULL,
+                              hints->hints_enabled ? &hints_is_ipv4 : NULL,
+                              wamr_addr_info, addr_info_size, &_max_info_size);
```
wenyongh, line 2966: Should check the return value of wamr_addr_info before using it.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -2915,6 +2944,52 @@ wasi_ssp_sock_bind(
...
+
+    if (ret != BHT_OK) {
+        free(wamr_addr_info);
```
wenyongh, line 2969: Use wasm_runtime_free

core/shared/platform/include/platform_api_extension.h
```diff
@@ -339,6 +339,40 @@ os_socket_shutdown(bh_socket_t socket);
 int
 os_socket_inet_network(const char *cp, uint32 *out);
 
+typedef struct {
+    uint8_t addr[16];
+    uint16_t port;
+    uint8_t is_ipv4;
+    uint8_t is_tcp;
+} wamr_addr_info_t;
```
wenyongh, line 347: How about changing to bh_addr_info_t or bh_sock_addr_info_t? We define type bh_socket_t in platform layer, had better align to it.

core/shared/platform/common/posix/posix_socket.c
```diff
@@ -176,3 +177,100 @@ os_socket_inet_network(const char *cp, uint32 *out)
...
+                uint8_t *hint_is_ipv4, wamr_addr_info_t *addr_info,
+                size_t addr_info_size, size_t *max_info_size)
+{
+    struct addrinfo hints, *res, *result;
+    int hints_enabled = hint_is_tcp || hint_is_ipv4;
+    int ret;
+    size_t pos = 0;
+
+    if (hints_enabled) {
+        memset(&hints, 0, sizeof(hints));
```
wenyongh, line 221: Had better memset hints before L220 to avoid unexpected behavior.

core/shared/platform/include/platform_api_extension.h
```diff
@@ -339,6 +339,40 @@ os_socket_shutdown(bh_socket_t socket);
...
+
+/**
+ * Resolves a host a hostname and a service to one or more IP addresses
```
wenyongh, line 350: Resolve

wenyongh (commented):
@loganek thanks for implementing the new features, it seems that the code change is big (several PRs now and more PRs to be added), I am a little concerned that whether it is stable enough and whether it impacts current socket api feature, how about we creating a new branch (e.g. dev/socket_api) to develop these features? And after they are implemented and tested as stable, we can merge it into main branch.

core/shared/platform/include/platform_api_extension.h
loganek: sure, didn't know about the convention. What does 'BH' stand for?

core/shared/platform/common/posix/posix_socket.c
loganek: what unexpected behavior do you anticipate here? `hints` variable is only being used when `hints_enabled` is set to true:
```
getaddrinfo(host, service, hints_enabled ? &hints : NULL, &result);
```

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
loganek: The main reason I decided to rename this function was that the WASI function name that we export is `addr_resolve` (with no `socket_` prefix. Also, unlike other `socket_*` functions, this function does not operate on the socket that's being passed as argument. I see your point that it's part of the `socket` API so I'm not insisting on renaming it back to `socket_addr_resolve`, but in that case I think we should also change the name of the function in WASI to be consistent.

What do you think?

core/iwasm/libraries/lib-socket/inc/wasi_socket_ext.h
loganek: I'm testing my changes on 64-bit Ubuntu 20.04 and 32-bit Lubuntu 18.04 (through qemu). It worked well on both.

loganek:
> @loganek thanks for implementing the new features, it seems that the code change is big (several PRs now and more PRs to be added), I am a little concerned that whether it is stable enough and whether it impacts current socket api feature, how about we creating a new branch (e.g. dev/socket_api) to develop these features? And after they are implemented and tested as stable, we can merge it into main branch.

@wenyongh thanks for your support in reviewing the PRs. Regarding the interfaces, I think we agreed on the high-level strategy that we stick to close-to-POSIX interfaces until the [wasi-socket](https://github.com/WebAssembly/wasi-sockets/) proposal is finalized, so I don't expect us to introduce any major changes on the interface level. In terms of implementation - I agree, there are 3 approaches we can take for this change (and other major changes in the future):
1. We work in complete isolation on all the features and then make a huge PR with all the changes
2. We make small contributions to dev-socket branch and each contribution is reviewed, at the end of the work we test all of that and merge it to main branch
3. We make small contributions to main branch

I think 1. is unreasonable given the amount of changes we're planning for sockets; it'd be difficult to review and I think the feedback loop will be very slow. What you propose, as far as I understand, is 2., whereas what we're doing now is 3.
In this specific case, I personally don't see a significant difference between 2 and 3 mainly because of the following reasons:
* most of the changes (if not all) are around sockets library, and it's quite unlikely they interfere with other modules in WAMR
* each change has to be tested before we make a PR (either to main or to dev branch) anyway so we catch problems on the early stage.
* if there's any incremental work (like in this PR, we haven't updated examples and `wasi_socket_ext.*` files), we still can make sure it's not breaking anything by making sure the iterations are feature-completed

With all that above, I'm not against following the approach you propose. If that's what you prefer, can you please create a branch or give me permission to create branches in this repository? Also, if we follow that path, I think it'd be good to at least merge the PRs that we already have opened so we're not loosing the history. 

Please let me know what are your thoughts on that.
Thanks

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
wenyongh: Yes, had better change the related structure/function names to be `socket_` prefixed, it looks more clear, or it makes a little confused what wamr_addr_info_t means, bh_sock_addr_info_t or bh_sock_addr_t seems better.

core/shared/platform/common/posix/posix_socket.c
wenyongh: Yes, not sure whether it will impacts `getaddrinfo(host, service, hints_enabled ? &hints : NULL, &result);` in L230. It should be OK to memset here, we are just used to initialize the structure at the beginning, e.g. `struct addrinfo hints = { 0 } ` at the definition.

core/iwasm/libraries/lib-socket/inc/wasi_socket_ext.h
wenyongh: OK

core/shared/platform/include/platform_api_extension.h
wenyongh: bh is short for Beihai, which is a code name of internal project, some of WAMR core/share APIs were ported from it.

wenyongh:
> > @loganek thanks for implementing the new features, it seems that the code change is big (several PRs now and more PRs to be added), I am a little concerned that whether it is stable enough and whether it impacts current socket api feature, how about we creating a new branch (e.g. dev/socket_api) to develop these features? And after they are implemented and tested as stable, we can merge it into main branch.
> 
> @wenyongh thanks for your support in reviewing the PRs. Regarding the interfaces, I think we agreed on the high-level strategy that we stick to close-to-POSIX interfaces until the [wasi-socket](https://github.com/WebAssembly/wasi-sockets/) proposal is finalized, so I don't expect us to introduce any major changes on the interface level. In terms of implementation - I agree, there are 3 approaches we can take for this change (and other major changes in the future):
> 
> 1. We work in complete isolation on all the features and then make a huge PR with all the changes
> 2. We make small contributions to dev-socket branch and each contribution is reviewed, at the end of the work we test all of that and merge it to main branch
> 3. We make small contributions to main branch
> 
> I think 1. is unreasonable given the amount of changes we're planning for sockets; it'd be difficult to review and I think the feedback loop will be very slow. What you propose, as far as I understand, is 2., whereas what we're doing now is 3. In this specific case, I personally don't see a significant difference between 2 and 3 mainly because of the following reasons:
> 

Yes, approach 1 is not so good as the final PR to main branch is big, approach 3 is not preferred as it keeps making changes into main branch. Approach 2 should be better, shall we define some milestones, and submit PRs to the dev branch until one milestone is reached, and after that, we can (1) do some more tests, (2) update samples and documents, (3) apply static code analysis, e.g. Klocwork and Coverity, (4) fully review the code again, so as to ensure the code quality, and then merge it into main branch. What is your opinion?
And could you please file an issue in WAMR community to introduce the feature to develop? For example, including motivations, design goals, APIs to be added, platforms to be enabled, milestones, and so on? It will give developers a more clear perspective and what the progress is.

wenyongh:
BTW, what the dev branch is do you prefer? dev/socket_ipv6, dev/socket_api, or others?

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
lum1n0us: OK.

lum1n0us:
> > there is always a concern about `getaddrinfo()` in the community. WASI is built using capability-based security principles. It requires making a statement before using any system resource or accessing any outside resource.
> > So, there is a `--addr-pool` in _iwasm_ command line option for creating sockets.
> > I am a little worried about if any resource will be involved during `getaddrinfo()` and should be reclaimed first
> 
> I think this one is tricky as depending on the system, getaddrinfo will require different set of capabilities (e.g. on linux getaddrinfo references files from `/etc`, which is probably different for linux). One option I think we can consider is having an explicit permission to do ns-lookups (like `--allow-ns-lookup`) which will enable users to use the function. If you agree with that, I can do the work in a follow-up PR. But I'm also open for other suggestions.

love the idea of `--allow-ns-lookup`. Let's do it.

core/shared/platform/include/platform_api_extension.h
loganek: Thanks, we might want to do a rename at one point of time to avoid internal code names and make it more clear for people outside your organization (that will probably improve the onboarding experience), but let's keep it like that for now.

loganek:
> > > @loganek thanks for implementing the new features, it seems that the code change is big (several PRs now and more PRs to be added), I am a little concerned that whether it is stable enough and whether it impacts current socket api feature, how about we creating a new branch (e.g. dev/socket_api) to develop these features? And after they are implemented and tested as stable, we can merge it into main branch.
> > 
> > 
> > @wenyongh thanks for your support in reviewing the PRs. Regarding the interfaces, I think we agreed on the high-level strategy that we stick to close-to-POSIX interfaces until the [wasi-socket](https://github.com/WebAssembly/wasi-sockets/) proposal is finalized, so I don't expect us to introduce any major changes on the interface level. In terms of implementation - I agree, there are 3 approaches we can take for this change (and other major changes in the future):
> > 
> > 1. We work in complete isolation on all the features and then make a huge PR with all the changes
> > 2. We make small contributions to dev-socket branch and each contribution is reviewed, at the end of the work we test all of that and merge it to main branch
> > 3. We make small contributions to main branch
> > 
> > I think 1. is unreasonable given the amount of changes we're planning for sockets; it'd be difficult to review and I think the feedback loop will be very slow. What you propose, as far as I understand, is 2., whereas what we're doing now is 3. In this specific case, I personally don't see a significant difference between 2 and 3 mainly because of the following reasons:
> 
> Yes, approach 1 is not so good as the final PR to main branch is big, approach 3 is not preferred as it keeps making changes into main branch. Approach 2 should be better, shall we define some milestones, and submit PRs to the dev branch until one milestone is reached, and after that, we can (1) do some more tests, (2) update samples and documents, (3) apply static code analysis, e.g. Klocwork and Coverity, (4) fully review the code again, so as to ensure the code quality, and then merge it into main branch. What is your opinion? And could you please file an issue in WAMR community to introduce the feature to develop? For example, including motivations, design goals, APIs to be added, platforms to be enabled, milestones, and so on? It will give developers a more clear perspective and what the progress is.

Hi @wenyongh , Ok let's continue the development in the dev branch from now on. I'll open a ticket as well to define the plan as well so it's more clear for us (in fact, we internally at Amazon already have such a plan, somehow I didn't think of sharing it in github, but I think that's excellent idea).

loganek:
> BTW, what the dev branch is do you prefer? dev/socket_ipv6, dev/socket_api, or others?

@wenyongh could you please create a `dev/socket` branch? I'd like to keep the name generic as it is about 1) adding support for ipv6, 2) bugfixing some of the existing functionality, 3) extending the API. Please let me know when that's created, so we'll be raising PRs to that branch.

wenyongh:
> > BTW, what the dev branch is do you prefer? dev/socket_ipv6, dev/socket_api, or others?
> 
> @wenyongh could you please create a `dev/socket` branch? I'd like to keep the name generic as it is about 1) adding support for ipv6, 2) bugfixing some of the existing functionality, 3) extending the API. Please let me know when that's created, so we'll be raising PRs to that branch.

@loganek Thanks, the branch `dev/socket` was created.

loganek:
> > > there is always a concern about `getaddrinfo()` in the community. WASI is built using capability-based security principles. It requires making a statement before using any system resource or accessing any outside resource.
> > > So, there is a `--addr-pool` in _iwasm_ command line option for creating sockets.
> > > I am a little worried about if any resource will be involved during `getaddrinfo()` and should be reclaimed first
> > 
> > 
> > I think this one is tricky as depending on the system, getaddrinfo will require different set of capabilities (e.g. on linux getaddrinfo references files from `/etc`, which is probably different for linux). One option I think we can consider is having an explicit permission to do ns-lookups (like `--allow-ns-lookup`) which will enable users to use the function. If you agree with that, I can do the work in a follow-up PR. But I'm also open for other suggestions.
> 
> love the idea of `--allow-ns-lookup`. Let's do it.

Since we now do the development for the `dev/socket` branch, I think we can develop this particular feature in the next PR to not keep this one open for too long. We already have an issue for tracking the work (#1336) so it won't be forgotten for sure.

wenyongh:
@loganek Thanks a lot for the PRs and the issue #1328 you filed, I have merged this PR and changed the dst branch of other PRs to dev/socket. Thanks.

PR #1320, opened 2022-07-21
Implement sock_addr_local syscall

loganek:
I also slightly changed the interface - since we already have a `__wasi_addr_t`
structure which is an union, there's no need for passing length around - the
address buffer will always have the right length (i.e. max of all address
families).

### DIFF
--- a/core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
+++ b/core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
@@ -55,6 +55,38 @@ sockaddr_to_wasi_addr(const struct sockaddr *sock_addr, socklen_t addrlen,
 }
 
 static __wasi_errno_t
+wasi_addr_to_sockaddr(const __wasi_addr_t *wasi_addr,
+                      struct sockaddr *sock_addr, socklen_t *addrlen)
+{
+    switch (wasi_addr->kind) {
+        case IPv4:
+        {
+            struct sockaddr_in sock_addr_in = { 0 };
+            uint32_t s_addr;
+
+            s_addr = (wasi_addr.addr.ip4.addr.n0 << 24)
+                     | (wasi_addr.addr.ip4.addr.n1 << 16)
+                     | (wasi_addr.addr.ip4.addr.n2 << 8)
+                     | wasi_addr.addr.ip4.addr.n3;
+
+            sock_addr_in.sin_family = AF_INET;
+            sock_addr_in.sin_addr.s_addr = htonl(s_addr);
+            sock_addr_in.sin_port = htons(wasi_addr->addr.ip4.port);
+            memcpy(sock_addr, &sock_addr_in, sizeof(sock_addr_in));
+
+            *addrlen = sizeof(sock_addr_in);
+            break;
+        }
+        case IPv6:
+            // TODO: IPV6
+            return __WASI_ERRNO_AFNOSUPPORT;
+        default:
+            return __WASI_ERRNO_AFNOSUPPORT;
+    }
+    return __WASI_ERRNO_SUCCESS;
+}
+
+static __wasi_errno_t
 sock_addr_remote(__wasi_fd_t fd, struct sockaddr *sock_addr, socklen_t *addrlen)
 {
     __wasi_addr_t wasi_addr = { 0 };
@@ -67,30 +99,7 @@ sock_addr_remote(__wasi_fd_t fd, struct sockaddr *sock_addr, socklen_t *addrlen)
         return error;
     }
 
-    if (IPv4 == wasi_addr.kind) {
-        struct sockaddr_in sock_addr_in = { 0 };
-
-        s_addr = (wasi_addr.addr.ip4.addr.n0 << 24)
-                 | (wasi_addr.addr.ip4.addr.n1 << 16)
-                 | (wasi_addr.addr.ip4.addr.n2 << 8)
-                 | wasi_addr.addr.ip4.addr.n3;
-
-        sock_addr_in.sin_family = AF_INET;
-        sock_addr_in.sin_addr.s_addr = htonl(s_addr);
-        sock_addr_in.sin_port = htons(wasi_addr.addr.ip4.port);
-        memcpy(sock_addr, &sock_addr_in, sizeof(sock_addr_in));
-
-        *addrlen = sizeof(sock_addr_in);
-    }
-    else if (IPv6 == wasi_addr.kind) {
-        // TODO: IPV6
-        return __WASI_ERRNO_AFNOSUPPORT;
-    }
-    else {
-        return __WASI_ERRNO_AFNOSUPPORT;
-    }
-
-    return __WASI_ERRNO_SUCCESS;
+    return wasi_addr_to_sockaddr(&wasi_addr, sock_addr, addrlen);
 }
 
 int
@@ -259,4 +268,19 @@ socket(int domain, int type, int protocol)
     HANDLE_ERROR(error)
 
     return sockfd;
+}
+
+int
+getsockname(int sockfd, struct sockaddr *addr, socklen_t *addrlen)
+{
+    __wasi_addr_t wasi_addr = { 0 };
+    __wasi_errno_t error;
+
+    error = __wasi_sock_addr_local(sockfd, &wasi_addr);
+    HANDLE_ERROR(error)
+
+    error = wasi_addr_to_sockaddr(&wasi_addr, addr, addrlen);
+    HANDLE_ERROR(error)
+
+    return __WASI_ERRNO_SUCCESS;
 }
--- a/core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
+++ b/core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
@@ -1009,8 +1009,8 @@ wasi_sock_accept(wasm_exec_env_t exec_env, wasi_fd_t fd, wasi_fd_t *fd_new)
 }
 
 static wasi_errno_t
-wasi_sock_addr_local(wasm_exec_env_t exec_env, wasi_fd_t fd, uint8 *buf,
-                     wasi_size_t buf_len)
+wasi_sock_addr_local(wasm_exec_env_t exec_env, wasi_fd_t fd,
+                     __wasi_addr_t *addr)
 {
     wasm_module_inst_t module_inst = get_module_inst(exec_env);
     wasi_ctx_t wasi_ctx = get_wasi_ctx(module_inst);
@@ -1019,9 +1019,12 @@ wasi_sock_addr_local(wasm_exec_env_t exec_env, wasi_fd_t fd, uint8 *buf,
     if (!wasi_ctx)
         return __WASI_EACCES;
 
+    if (!validate_native_addr(addr, sizeof(__wasi_addr_t)))
+        return __WASI_EINVAL;
+
     curfds = wasi_ctx_get_curfds(module_inst, wasi_ctx);
 
-    return wasi_ssp_sock_addr_local(curfds, fd, buf, buf_len);
+    return wasi_ssp_sock_addr_local(curfds, fd, addr);
 }
 
 static wasi_errno_t
@@ -1401,7 +1404,7 @@ static NativeSymbol native_symbols_libc_wasi[] = {
     REG_NATIVE_FUNC(proc_raise, "(i)i"),
     REG_NATIVE_FUNC(random_get, "(*~)i"),
     REG_NATIVE_FUNC(sock_accept, "(i*)i"),
-    REG_NATIVE_FUNC(sock_addr_local, "(i*i)i"),
+    REG_NATIVE_FUNC(sock_addr_local, "(i*)i"),
     REG_NATIVE_FUNC(sock_addr_remote, "(i*i)i"),
     REG_NATIVE_FUNC(sock_addr_resolve, "($$**i*)i"),
     REG_NATIVE_FUNC(sock_bind, "(i*)i"),
--- a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
+++ b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
@@ -203,12 +203,14 @@ convert_clockid(__wasi_clockid_t in, clockid_t *out)
 static void
 ipv4_addr_to_wasi_addr(uint32_t addr, __wasi_ip_port_t port, __wasi_addr_t *out)
 {
+    addr = ntohl(addr);
+
     out->kind = IPv4;
     out->addr.ip4.port = port;
-    out->addr.ip4.addr.n3 = (addr & 0xFF000000) >> 24;
-    out->addr.ip4.addr.n2 = (addr & 0x00FF0000) >> 16;
-    out->addr.ip4.addr.n1 = (addr & 0x0000FF00) >> 8;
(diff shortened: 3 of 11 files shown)


core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -2852,16 +2881,60 @@ wasi_ssp_sock_addr_local(
...
+        return __WASI_EINVAL;
+    }
+    if (is_ipv4) {
+        ipv4_addr_to_wasi_addr(*(uint32_t *)inet_pton_dest, port, addr);
+    }
+    else {
+        ipv6_addr_to_wasi_addr((uint16 *)inet_pton_dest, port, addr);
+    }
+
+    // addr_len = sizeof(addr_storage);
```
lum1n0us, line 2915: might want to remove unused code

core/shared/platform/common/posix/posix_socket.c
```diff
@@ -176,3 +176,61 @@ os_socket_inet_network(const char *cp, uint32 *out)
...
+    }
+
+    *is_ipv4 = addr_storage.ss_family == AF_INET;
+
+    switch (addr_storage.ss_family) {
+        case AF_INET:
+            *port = ntohs(((struct sockaddr_in *)&addr_storage)->sin_port);
+            inet_ntop(AF_INET, &((struct sockaddr_in *)&addr_storage)->sin_addr,
+                      buf, buflen);
+            break;
```
lum1n0us, line 224:
```suggestion
            *is_ipv4 = true;
            break;
```

core/shared/platform/common/posix/posix_socket.c
```diff
@@ -176,3 +176,61 @@ os_socket_inet_network(const char *cp, uint32 *out)
...
+        case AF_INET6:
+            *port = ntohs(((struct sockaddr_in6 *)&addr_storage)->sin6_port);
+            inet_ntop(AF_INET6,
+                      &((struct sockaddr_in6 *)&addr_storage)->sin6_addr, buf,
+                      buflen);
+            break;
```
lum1n0us, line 230:
```suggestion
            *is_ipv4 = false;
            break;
```

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -2852,16 +2881,60 @@ wasi_ssp_sock_addr_local(
...
         fd_object_get(curfds, &fo, fd, __WASI_RIGHT_SOCK_ADDR_LOCAL, 0);
     if (error != __WASI_ESUCCESS)
         return error;
 
+    ret = os_socket_addr_local(fd_number(fo), buf, sizeof(buf) / sizeof(buf[0]),
+                               &port, &is_ipv4);
     fd_object_release(fo);
-    return __WASI_ENOSYS;
+    if (ret != BHT_OK) {
+        error = convert_errno(errno);
```
lum1n0us, line 2902: better don't go any further

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -2852,16 +2881,60 @@ wasi_ssp_sock_addr_local(
...
+    }
+
+    if (BHT_OK != os_socket_inet_pton(is_ipv4, buf, inet_pton_dest)) {
```
lum1n0us, line 2905: wondering why convert a network order binary to presentation in `os_socket_addr_local()` and convert it back here?

loganek: Yeah, you're right; I guess I was confused by the way it's done in `bind()` - I think we could pass binary representation there as well and that would simplify the code.

core/shared/platform/common/posix/posix_socket.c
```diff
@@ -176,3 +176,50 @@ os_socket_inet_network(const char *cp, uint32 *out)
...
+    switch (addr_storage.ss_family) {
+        case AF_INET:
+        {
+            struct sockaddr_in *addr_in = (struct sockaddr_in *)&addr_storage;
+
+            assert(buflen >= sizeof(addr_in->sin_addr));
+            *port = ntohs(addr_in->sin_port);
+            memcpy(buf, &addr_in->sin_addr, sizeof(addr_in->sin_addr));
+            *is_ipv4 = true;
+        } break;
```
lum1n0us, line 207:
```suggestion
          break;
        } 
```

core/shared/platform/common/posix/posix_socket.c
```diff
@@ -176,3 +176,50 @@ os_socket_inet_network(const char *cp, uint32 *out)
...
+            struct sockaddr_in6 *addr_in = (struct sockaddr_in6 *)&addr_storage;
+            assert(buflen >= sizeof(addr_in->sin6_addr));
+            *port = ntohs(addr_in->sin6_port);
+
+            for (int i = 0; i < 8; i++) {
+                ((uint16_t *)buf)[i] =
+                    ntohs(((uint16_t *)&addr_in->sin6_addr)[i]);
+            }
+            *is_ipv4 = false;
+        } break;
```
lum1n0us, line 219:
```suggestion
           break;
        }
```

lum1n0us (commented):
only some code-style problems

core/shared/platform/common/posix/posix_socket.c
loganek: Hmm, that seems to be automatically applied by clang-format-12. Also, I see it's the same in other files (e.g. [this one](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/samples/wasm-c-api/src/callback.c#L15))

core/shared/platform/common/posix/posix_socket.c
loganek: Hmm, that seems to be automatically applied by clang-format-12. Also, I see it's the same in other files (e.g. [this one](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/samples/wasm-c-api/src/callback.c#L15))

loganek:
> only some code-style problems

Thanks, already replied.

core/shared/platform/common/posix/posix_socket.c
lum1n0us: for some reason, we didn't apply our code style to `samples/wasm-c-api/src/`, so they are no good examples. please try to follow in `core/iwasm`

loganek: Please note if I make that change the build fails: https://github.com/bytecodealliance/wasm-micro-runtime/runs/7536836808?check_suite_focus=true and the error is:
```
--- core/shared/platform/common/posix/posix_socket.c failed on code style checking.
```

As I said, clang-format-12 enforces this kind of styling; we can change the configuration of clang formatter, but I think that's outside of the scope of this change, so I'd prefer not doing that here.

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
```diff
@@ -53,6 +53,34 @@ sockaddr_to_wasi_addr(const struct sockaddr *sock_addr, socklen_t addrlen,
...
+            sock_addr_in.sin_family = AF_INET;
+            sock_addr_in.sin_addr.s_addr = (wasi_addr->addr.ip4.addr.n3 << 24)
+                                           | (wasi_addr->addr.ip4.addr.n2 << 16)
+                                           | (wasi_addr->addr.ip4.addr.n1 << 8)
+                                           | wasi_addr->addr.ip4.addr.n0;
+            sock_addr_in.sin_port = htons(wasi_addr->addr.ip4.port);
+            memcpy(sock_addr, &sock_addr_in, sizeof(sock_addr_in));
+
+            *addrlen = sizeof(sock_addr_in);
+        } break;
```
lum1n0us, line 74:
```suggestion
            break;
        }
```

core/shared/platform/common/posix/posix_socket.c
lum1n0us:
```suggestion
            break;
        }
```

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
loganek: Hi @lum1n0us please see my previous comment above:

> Please note if I make that change the build fails: https://github.com/bytecodealliance/wasm-micro-runtime/runs/7536836808?check_suite_focus=true and the error is:
> 
> ```
> --- core/shared/platform/common/posix/posix_socket.c failed on code style checking.
> ```
> 
> As I said, clang-format-12 enforces this kind of styling; we can change the configuration of clang formatter, but I think that's outside of the scope of this change, so I'd prefer not doing that here.

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -1009,8 +1009,8 @@ wasi_sock_accept(wasm_exec_env_t exec_env, wasi_fd_t fd, wasi_fd_t *fd_new)
 }
 
 static wasi_errno_t
-wasi_sock_addr_local(wasm_exec_env_t exec_env, wasi_fd_t fd, uint8 *buf,
-                     wasi_size_t buf_len)
+wasi_sock_addr_local(wasm_exec_env_t exec_env, wasi_fd_t fd,
+                     __wasi_addr_t *addr)
```
wenyongh, line 1013: Should check whether addr is out of the range in linear memory, as the signature is register as "(i*)i", addr is pointer argument, runtime only checks its first byte isn't out of range before passing it into native API:
```
if (!validate_native_addr(addr, sizeof(__wasi_addr_t))
    return error;
```

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
lum1n0us: Sorry, just noticed that. We might have a glitch in the clang-format now.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -2852,16 +2881,34 @@ wasi_ssp_sock_addr_local(
 #if !defined(WASMTIME_SSP_STATIC_CURFDS)
     struct fd_table *curfds,
 #endif
-    __wasi_fd_t fd, uint8 *buf, __wasi_size_t buf_len)
+    __wasi_fd_t fd, __wasi_addr_t *addr)
 {
     struct fd_object *fo;
+    uint8_t buf[16];
+    __wasi_ip_port_t port;
+    uint8_t is_ipv4;
```
wenyongh, line 2889: Had better use `uint8`, we usually use type uint32/uint16/int8 instead of uint32_t/uint16_t/uint8_t inside runtime.

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
lum1n0us: Btw, may I have a look at the code causes the error?
"--- core/shared/platform/common/posix/posix_socket.c failed on code style checking." ?

loganek: This is the code you proposed as a change (putting `break` in a new line)

lum1n0us: I am hoping
```
        case IPv4:
        {
            struct sockaddr_in sock_addr_in = { 0 };

            sock_addr_in.sin_family = AF_INET;
            sock_addr_in.sin_addr.s_addr = (wasi_addr->addr.ip4.addr.n3 << 24)
                                           | (wasi_addr->addr.ip4.addr.n2 << 16)
                                           | (wasi_addr->addr.ip4.addr.n1 << 8)
                                           | wasi_addr->addr.ip4.addr.n0;
            sock_addr_in.sin_port = htons(wasi_addr->addr.ip4.port);
            memcpy(sock_addr, &sock_addr_in, sizeof(sock_addr_in));

            *addrlen = sizeof(sock_addr_in);
            break;
        }
```

loganek: Ah, sorry I misinterpreted your initial comment. I'll move break inside the {} then.

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
loganek: that's good to know, thanks for pointing that out.

wenyongh:
@loganek There are compilation errors on linux-sgx platform:
```
usr/bin/ld: libvmlib.a(posix.c.o): in function `wasi_ssp_sock_addr_local':
posix.c:(.text.wasi_ssp_sock_addr_local+0x157): undefined reference to `ntohl'
/usr/bin/ld: libvmlib.a(posix.c.o): in function `wasi_ssp_sock_addr_resolve':
posix.c:(.text.wasi_ssp_sock_addr_resolve+0x151): undefined reference to `ntohl'
```
These functions are not provided by sgx-sdk, but they were implemented in linux-sgx sgx_socket.c as static functions:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/shared/platform/linux-sgx/sgx_socket.c#L65-L130

Could you please add the API declarations in core/shared/platform/linux-sgx/platform_internal.h, and move the implementation into core/shared/platform/linux-sgx/platform_init.c and remove static flag?

loganek:
> @loganek There are compilation errors on linux-sgx platform:
> 
> ```
> usr/bin/ld: libvmlib.a(posix.c.o): in function `wasi_ssp_sock_addr_local':
> posix.c:(.text.wasi_ssp_sock_addr_local+0x157): undefined reference to `ntohl'
> /usr/bin/ld: libvmlib.a(posix.c.o): in function `wasi_ssp_sock_addr_resolve':
> posix.c:(.text.wasi_ssp_sock_addr_resolve+0x151): undefined reference to `ntohl'
> ```
> 
> These functions are not provided by sgx-sdk, but they were implemented in linux-sgx sgx_socket.c as static functions: https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/shared/platform/linux-sgx/sgx_socket.c#L65-L130
> 
> Could you please add the API declarations in core/shared/platform/linux-sgx/platform_internal.h, and move the implementation into core/shared/platform/linux-sgx/platform_init.c and remove static flag?

Hi @wenyongh , I've kept the implementation in sgx_socket.c but made it non-static. Is there any reason we'd prefer to move it to platform_init.c? What's the purpose of that file? I think similar functions follow the same pattern (i.e. the implementation exists in sgx_socket file).

Regarding the header file - I added the declaration to sgx_socket.h, as it looked like other functions are declared there, and the file is included through other headers.

wenyongh:
> Hi @wenyongh , I've kept the implementation in sgx_socket.c but made it non-static. Is there any reason we'd prefer to move it to platform_init.c? What's the purpose of that file? I think similar functions follow the same pattern (i.e. the implementation exists in sgx_socket file).
> 
> Regarding the header file - I added the declaration to sgx_socket.h, as it looked like other functions are declared there, and the file is already included through other headers.

@loganek It should be also OK, platform_init.c is to implement some common APIs for the platform, here it should be sgx_platform.c, for linux/windows/darwin, we named it platform_init.c. I will review the code again and merge it if there is no issue found.

PR #1327, opened 2022-07-25
Fix byte ordering in sockets API for systems where host byte order == network byte order

loganek:
- Documented data structures used for storing IP addresses.
- Fixed bug in bind and connect methods by updating a code in wasi_socket_ext.

### DIFF
--- a/core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
+++ b/core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
@@ -22,12 +22,13 @@
 static void
 ipv4_addr_to_wasi_addr(uint32_t addr_num, uint16_t port, __wasi_addr_t *out)
 {
+    addr_num = ntohl(addr_num);
     out->kind = IPv4;
     out->addr.ip4.port = ntohs(port);
-    out->addr.ip4.addr.n3 = (addr_num & 0xFF000000) >> 24;
-    out->addr.ip4.addr.n2 = (addr_num & 0x00FF0000) >> 16;
-    out->addr.ip4.addr.n1 = (addr_num & 0x0000FF00) >> 8;
-    out->addr.ip4.addr.n0 = (addr_num & 0x000000FF);
+    out->addr.ip4.addr.n0 = (addr_num & 0xFF000000) >> 24;
+    out->addr.ip4.addr.n1 = (addr_num & 0x00FF0000) >> 16;
+    out->addr.ip4.addr.n2 = (addr_num & 0x0000FF00) >> 8;
+    out->addr.ip4.addr.n3 = (addr_num & 0x000000FF);
 }
 
 static __wasi_errno_t
@@ -57,6 +58,7 @@ static __wasi_errno_t
 sock_addr_remote(__wasi_fd_t fd, struct sockaddr *sock_addr, socklen_t *addrlen)
 {
     __wasi_addr_t wasi_addr = { 0 };
+    uint32_t s_addr;
     __wasi_errno_t error;
 
     error =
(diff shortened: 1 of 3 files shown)


core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
```diff
@@ -68,11 +70,13 @@ sock_addr_remote(__wasi_fd_t fd, struct sockaddr *sock_addr, socklen_t *addrlen)
...
+                 | (wasi_addr.addr.ip4.addr.n1 << 16)
+                 | (wasi_addr.addr.ip4.addr.n2 << 8)
+                 | wasi_addr.addr.ip4.addr.n3;
+
         sock_addr_in.sin_family = AF_INET;
-        sock_addr_in.sin_addr.s_addr = (wasi_addr.addr.ip4.addr.n3 << 24)
-                                       | (wasi_addr.addr.ip4.addr.n2 << 16)
-                                       | (wasi_addr.addr.ip4.addr.n1 << 8)
-                                       | wasi_addr.addr.ip4.addr.n0;
+        sock_addr_in.sin_addr.s_addr = htonl(s_addr);
```
lum1n0us, line 79: `htons` ?

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
```diff
@@ -22,12 +22,13 @@
 static void
 ipv4_addr_to_wasi_addr(uint32_t addr_num, uint16_t port, __wasi_addr_t *out)
 {
+    addr_num = htonl(addr_num);
```
lum1n0us, line 25: in L42 where calls `ipv4_addr_to_wasi_addr`, we pass `sin_addr.s_addr` as `addr_num`, and I think it is in "network byte order".

loganek: good spot; `htonl` and `ntohl` are reverse operations so the code would still work well on both little and big endian, but I agree, better to not mix them. Updated the code

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
```diff
@@ -22,12 +22,13 @@
...
     out->addr.ip4.port = ntohs(port);
-    out->addr.ip4.addr.n3 = (addr_num & 0xFF000000) >> 24;
-    out->addr.ip4.addr.n2 = (addr_num & 0x00FF0000) >> 16;
-    out->addr.ip4.addr.n1 = (addr_num & 0x0000FF00) >> 8;
-    out->addr.ip4.addr.n0 = (addr_num & 0x000000FF);
+    out->addr.ip4.addr.n0 = (addr_num & 0xFF000000) >> 24;
+    out->addr.ip4.addr.n1 = (addr_num & 0x00FF0000) >> 16;
+    out->addr.ip4.addr.n2 = (addr_num & 0x0000FF00) >> 8;
+    out->addr.ip4.addr.n3 = (addr_num & 0x000000FF);
 }
```
lum1n0us, line 32: let's take a walk:
- `addr_num` as a parameter, is the big-endian
- L25, `addr_num` is host-byte-order
- L28 - L31, store the MSB to `n0` and the LSB to `n3`. So back to big-endian?

loganek: Yes, n0 in this case is the most significant byte (e.g. given 127.0.0.1, n0 is 127), which is the same behavior as we have before this PR. What this PR fixes though is the case when the host system is big endian. Even though big endian system is currently not that popular, I think we should write the code in a way it supports both architectures.

lum1n0us: - If the host system is little-endian, what we are doing is big-endian -> little-endian -> big-endian ?
- if the host system is big-endian, what we are doing is big-endian -> big-endian -> big-endian?

So, why not directly store the parameter `addr_num` into `_wasi_addr_t` in network-byte-order?

loganek: I'm no sure if we can serialise addrnum to wasi_addr_t directly - as this is violation of strict aliasing rules. So I think more accurate description of what we do here is:
Little-endian host: big endian -> little endian (through ntohl) -> wasi_addr_t
Big-endian host: big endian -> no-op (as ntohl in that case shoud do nothing) -> wasi_addr_t

So there's no redundant operation in any of the stages.

loganek: If we want to stick to using `&` operator for calculating each byte, w have to make sure the value is in the host byte order. The alternative here would be to convert addr_num to array of uint8 so we can safely rely on network byte order, but I find this approach a bit more elegant (that's very subjective though)

lum1n0us:
> w  have to make sure the value is in the host byte order.

May I ask the problem if `addr_num` is network byte order?

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
lum1n0us: LGTM

PR #1330, opened 2022-07-27
fix a potential integer overflow issue

lum1n0us:
- allow max_page_count to be 65536
- allow init_page_count to be 65536
- `num_bytes_per_page * XXX_page_count` will be great than UINT32_MAX

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -464,13 +464,13 @@ memory_instantiate(AOTModuleInstance *module_inst, AOTModule *module,
         }
         init_page_count += inc_page_count;
         max_page_count += inc_page_count;
-        if (init_page_count > 65536) {
+        if (init_page_count > DEFAULT_MAX_PAGES) {
             set_error_buf(error_buf, error_buf_size,
                           "memory size must be at most 65536 pages (4GiB)");
             return NULL;
         }
-        if (max_page_count > 65536)
-            max_page_count = 65536;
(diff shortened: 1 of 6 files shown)


core/iwasm/aot/aot_runtime.h
```diff
@@ -293,6 +293,7 @@ typedef struct AOTMemoryInstance {
     uint32 num_bytes_per_page;
     uint32 cur_page_count;
     uint32 max_page_count;
+    /*TODO:memory_data_size should be uint64*/
     uint32 memory_data_size;
```
lum1n0us, line 297: not sure if we should change `uint32 memory_data_size` -> `uint64 memory_data_size`?

lum1n0us:
There is a serious AOT compatible problem if updating `memory_data_size` of `AOTMemoryInstance` from `uint32` to `uint64`, and we are going to drop the intention and back to 
- limit `max_page_count` to 65535
- limit `init_page_count` to 65535
- So, `num_bytes_per_page * XXX_page_count` will be less than or equal to `UINT32_MAX`

All running modes, *interpreter*, *jit*, and *aot* shall have the same behaviors.

lum1n0us:
But

```wat
(module (memory 0 65536))
```

https://github.com/WebAssembly/spec/blob/e8c05f4b0d3a68bd3b4389b943fbb763a7d3a3ae/test/core/memory.wast#L8

lum1n0us:
https://github.com/bytecodealliance/wasm-micro-runtime/pull/1380

PR #1333, opened 2022-07-27
[bugfix] initialize "module->retain_function" for wasm_mini_loader

yaozhongxiao:
Before resolving the module function's export in wasm_mini_loader,
"module->retain_function" need to be initialized, otherwise,
the "__new" function export will lead to abort.

issue: https://github.com/bytecodealliance/wasm-micro-runtime/issues/1332


yaozhongxiao:
The failing cases seem unrelated to this commit, PTAL @wenyongh @XuJun2019 @xwang98

xujuntwt95329:
LGTM

PR #1339, opened 2022-07-28
Add our own legacy loop unswitch pass

penzn:
Since legacy binding for loop unswitch pass was removed and we can't get
it back implement its equivalent `aot_llvm_extra.cpp` and use in
`aot_compiler.c`.

Follow up to #1183


penzn:
I am on Windows today, didn't run the tests yet (assuming they require Linux).

PR #1355, opened 2022-08-05
wasm_c_api.c: improve unimplemented cases

yamt:
ASSERT_NOT_IMPLEMENTED is bh_assert, which might be no-op.
in that case, it's better to fall back to the "default" case,
which reports an error properly.

### DIFF
--- a/core/iwasm/common/wasm_c_api.c
+++ b/core/iwasm/common/wasm_c_api.c
@@ -3955,9 +3955,8 @@ interp_link(const wasm_instance_t *inst, const WASMModule *module_interp,
             }
             case IMPORT_KIND_MEMORY:
             case IMPORT_KIND_TABLE:
-                ASSERT_NOT_IMPLEMENTED();
-                break;
             default:
+                ASSERT_NOT_IMPLEMENTED();
                 LOG_WARNING("%s meets unsupported kind: %d", __FUNCTION__,
                             import_rt->kind);
                 goto failed;
@@ -4175,9 +4174,8 @@ aot_link(const wasm_instance_t *inst, const AOTModule *module_aot,
                 break;
             case WASM_EXTERN_MEMORY:
             case WASM_EXTERN_TABLE:
-                ASSERT_NOT_IMPLEMENTED();
-                break;
             default:
+                ASSERT_NOT_IMPLEMENTED();
                 goto failed;
         }
     }


core/iwasm/common/wasm_c_api.c
```diff
@@ -3956,7 +3956,6 @@ interp_link(const wasm_instance_t *inst, const WASMModule *module_interp,
             case IMPORT_KIND_MEMORY:
             case IMPORT_KIND_TABLE:
                 ASSERT_NOT_IMPLEMENTED();
```
wenyongh, line 3958: This leads to compilation warning in debug mode: `warning: this statement may fall through`
How about changing to:
```C
            case IMPORT_KIND_MEMORY:
            case IMPORT_KIND_TABLE:
                ASSERT_NOT_IMPLEMENTED();
                goto handle_default;
            handle_default:
            default:
```

core/iwasm/common/wasm_c_api.c
```diff
@@ -4176,7 +4175,6 @@ aot_link(const wasm_instance_t *inst, const AOTModule *module_aot,
             case WASM_EXTERN_MEMORY:
             case WASM_EXTERN_TABLE:
                 ASSERT_NOT_IMPLEMENTED();
```
wenyongh, line 4177: Same as above

shaischaudhry (approved):

core/iwasm/common/wasm_c_api.c
yamt: it's unfortunate that clang doesn't recognize the traditional `/* FALLTHROUGH */`.
how do you think about introducing something like `#defile BH_FALLTHROUGH __attribute__((fallthrough))`?

wenyongh: Had better not, it might be not friendly to some compilers.
We have the similar case in:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/compilation/aot_emit_numberic.c#L455-L458
The modification such as that should be OK.

Or how about handling default together with `case IMPORT_KIND_MEMORY/IMPORT_KIND_TABLE`:
```C
            case IMPORT_KIND_MEMORY:
            case IMPORT_KIND_TABLE:
            default:
                  ASSERT_NOT_IMPLEMENTED();
                  LOG_WARNING("%s meets unsupported kind: %d", __FUNCTION__,
                              import_rt->kind);
                  goto failed;
```

yamt: > Had better not, it might be not friendly to some compilers. We have the similar case in: https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/compilation/aot_emit_numberic.c#L455-L458 The modification such as that should be OK.

ok

> 
> Or how about handling default together with `case IMPORT_KIND_MEMORY/IMPORT_KIND_TABLE`:
> 
> ```c
>             case IMPORT_KIND_MEMORY:
>             case IMPORT_KIND_TABLE:
>             default:
>                   ASSERT_NOT_IMPLEMENTED();
>                   LOG_WARNING("%s meets unsupported kind: %d", __FUNCTION__,
>                               import_rt->kind);
>                   goto failed;
> ```

i think it makes the most sense this particular case. done.

PR #1359, opened 2022-08-05
SGX: printf returns the msg size stored the buffer

JamesMenetrey:
Dear WAMR developers,

When working with WASI calls inside Intel SGX, I discovered a glitch when writing into stdout/stderr using the code snippet present here:

https://github.com/bytecodealliance/wasm-micro-runtime/blob/0020b3ae6874d275a7de9fc95fe078355363aab5/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c#L1249-L1262

Indeed, when this code is executed for printing text into stdout/stderr, the implementation of `os_printf` for SGX always returns 0, since the return value of the OCALL is not considered. As a result, writing to stdout/stderr from the enclave creates an infinite loop, where the text is written in the console over and over.

While I could not find where the loop occurs (either in WAMR or in musl for Wasm), I propose to change the implementation of the `os_printf` for Intel SGX, so the number of bytes is returned by its implementation by using the return value of the formatting function (`vsnprintf`). Note that we could also use the return value of the `printf` called in the OCALL, but I prefer this approach since we can trust the return value of `vsnprintf`, which is not the case with the concrete `printf`.

Feel free to comment or suggest changes if you feel this approach is not the right way to fix the issue. I'll gladly apply them.

Cheers!

### DIFF
--- a/core/shared/platform/linux-sgx/sgx_platform.c
+++ b/core/shared/platform/linux-sgx/sgx_platform.c
@@ -60,16 +60,18 @@ os_set_print_function(os_print_function_t pf)
 int
 os_printf(const char *message, ...)
 {
+    int bytes_written = 0;
+
     if (print_function != NULL) {
         char msg[FIXED_BUFFER_SIZE] = { '\0' };
         va_list ap;
         va_start(ap, message);
-        vsnprintf(msg, FIXED_BUFFER_SIZE, message, ap);
+        bytes_written += vsnprintf(msg, FIXED_BUFFER_SIZE, message, ap);
         va_end(ap);
         print_function(msg);
     }
 
-    return 0;
+    return bytes_written;
 }
 
 int


core/shared/platform/linux-sgx/sgx_platform.c
```diff
@@ -60,16 +60,18 @@ os_set_print_function(os_print_function_t pf)
...
 os_printf(const char *message, ...)
 {
+    int bytes_written = 0;
+
     if (print_function != NULL) {
         char msg[FIXED_BUFFER_SIZE] = { '\0' };
         va_list ap;
         va_start(ap, message);
-        vsnprintf(msg, FIXED_BUFFER_SIZE, message, ap);
+        bytes_written += vsnprintf(msg, FIXED_BUFFER_SIZE, message, ap);
```
wenyongh, line 69: It seems work, but should not the best way, we had better get the exact size from print_function. Some suggestions:
1. In platform_internal.h, change return type of os_print_function_t to int

2. Here, change L71 to `bytes_written += print_function(msg)`
    And should also apply same changes for os_vprintf in this file.

3. Had better enlarge macro FIXED_BUFFER_SIZE in this file, it is `1 << 9`, or 512, not sure whether it is enough for most cases? How about 4096?

4. Modify file `product-mini/platforms/linux-sgx/enclave-sample/Enclave/Enclave.edl`
    and `product-mini/platforms/linux-sgx/enclave-sample/Enclave/Enclave_minimal.edl`
    Change `void ocall_print([in, string]const char* str);` to `int ocall_print([in, string]const char* str);`

5. Modify `product-mini/platforms/linux-sgx/enclave-sample/App/App.cpp`, the implementation of ocall_print:
```C
int
ocall_print(const char *str)
{
    return printf("%s", str);
}
```

wenyongh:
@JamesMenetrey I submitted PR #1387 to fix this issue, could you please help check whether it works for you? Thanks.

wenyongh:
Fixed with PR #1387.

JamesMenetrey:
Hello @wenyongh,

Many thanks for your update. I'll keep you informed in a couple of week (I'm currently in vacation).

Cheers!

JamesMenetrey:
Hey @wenyongh,

I am coming back to you regarding #1387. My code with your fix works perfectly! Thanks!

Cheers

PR #1360, opened 2022-08-05
Implement sock_addr_remote syscall

loganek:
I also slightly changed the interface - since we already have a `__wasi_addr_t`
structure which is an union, there's no need for passing length around - the
address buffer will always have the right length (i.e. max of all address
families).

### DIFF
--- a/core/iwasm/libraries/lib-socket/inc/wasi_socket_ext.h
+++ b/core/iwasm/libraries/lib-socket/inc/wasi_socket_ext.h
@@ -117,6 +117,9 @@ socket(int domain, int type, int protocol);
 
 int
 getsockname(int sockfd, struct sockaddr *addr, socklen_t *addrlen);
+
+int
+getpeername(int sockfd, struct sockaddr *addr, socklen_t *addrlen);
 #endif
 
 /**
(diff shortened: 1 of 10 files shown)


loganek:
All the checks pass except `Coding Guidelines`, which says `there are 2 commits in the PR` so looks like it's checking a commit that's not part of the PR (the PR only includes one commit). Any thoughts on that?

wenyongh:
> All the checks pass except `Coding Guidelines`, which says `there are 2 commits in the PR` so looks like it's checking a commit that's not part of the PR (the PR only includes one commit). Any thoughts on that?

@loganek Have you formatted the source code's coding style by following the prompt:
```
    1. Install clang-format-12.0.0
    Normally we can install it by `sudo apt-get install clang-format-12`,
    or download the `clang+llvm-12.0.0-xxx-tar.xz` package from
      https://github.com/llvm/llvm-project/releases/tag/llvmorg-12.0.0
    and install it

    2. Format the C/C++ source file
    ``` shell
    cd path/to/wamr/root
    clang-format-12 --style file -i path/to/file
    ```
```
Normally after installing clang-format-12, you can run:
```
cd <wamr_root>
clang-format-12 -i --style=file `find core -name "*.h"`
clang-format-12 -i --style=file `find core -name "*.c"`
clang-format-12 -i --style=file `find core -name "*.cpp"`
```
to format the source files under core folder.

loganek:
@wenyongh yeah I did exactly that but for some reason didn't push the change... anyways, that should work now.

PR #1368, opened 2022-08-09
Support custom stack guard size

no1wudi:
Add a new option `WAMR_BUILD_STACK_GUARD_SIZE` to use custom stack guard size,
for most RTOS port, we use stack base as check boundry which may not safety as POSIX based system (like linux).


core/shared/platform/common/posix/posix_thread.c
```diff
@@ -382,6 +386,10 @@ os_thread_get_stack_boundary()
     if ((addr = (uint8 *)pthread_get_stackaddr_np(self))) {
         stack_size = pthread_get_stacksize_np(self);
 
+#ifdef WASM_STACK_GUARD_SIZE
```
wenyongh, line 389: The macro is always defined, how about removing macro control here?

core/shared/platform/common/posix/posix_thread.c
```diff
@@ -368,7 +368,11 @@ os_thread_get_stack_boundary()
 #ifdef __linux__
     if (pthread_getattr_np(self, &attr) == 0) {
         pthread_attr_getstack(&attr, (void **)&addr, &stack_size);
+#if WASM_STACK_GUARD_SIZE == 0
         pthread_attr_getguardsize(&attr, &guard_size);
+#else
+        guard_size = WASM_STACK_GUARD_SIZE;
+#endif
```
wenyongh, line 375: if WASM_STACK_GUARD_SIZE is smaller than 1 page, L374 takes effect, which may be too small? How about changing to:
```
        pthread_attr_getguardsize(&attr, &guard_size);
        guard_size += WASM_STACK_GUARD_SIZE;
```

core/shared/platform/windows/win_thread.c
```diff
@@ -638,7 +638,12 @@ os_thread_get_stack_boundary()
 
     /* 4 pages are set unaccessible by system, we reserved
        one more page at least for safety */
+#if WASM_STACK_GUARD_SIZE == 0
     thread_stack_boundary = (uint8 *)(uintptr_t)low_limit + page_size * 5;
+#else
+    thread_stack_boundary =
+        (uint8 *)(uintptr_t)low_limit + page_size * 4 + WASM_STACK_GUARD_SIZE;
+#endif
```
wenyongh, line 646: How about changing to:
```
        if (WASM_STACK_GUARD_SIZE < page_size)
            thread_stack_boundary = (uint8 *)(uintptr_t)low_limit + page_size * 5;
        else
             thread_stack_boundary =
                (uint8 *)(uintptr_t)low_limit + page_size * 4 + WASM_STACK_GUARD_SIZE;
```

core/shared/platform/common/posix/posix_thread.c
no1wudi: Changed to `#if WASM_STACK_GUARD_SIZE == 0`

core/shared/platform/common/posix/posix_thread.c
no1wudi: OK, it's more suitable for powerful unix like systems.

core/shared/platform/windows/win_thread.c
no1wudi: OK

lum1n0us:
Just a little confused about the meaning of when `exec_env->native_stack_boundary` is `NULL`. Does it mean one of the below?
- be easy, there is no stack boundary at all. 
- or, be panic, there is something wrong. There should always be value.

no1wudi:
In current implementation,  `NULL` means there is no stack boundary at all.

wenyongh:
> @lum1n0us In current implementation, I think `NULL` means there is no stack boundary at all.

Some platforms doesn't provide the related API to get native stack boundary,
so we return NULL in API `os_thread_get_stack_boundary`, e.g. alios.

lum1n0us:
in that case, such a comparison looks a little wired. It will be passed but ...

```c
    //  if (&module_inst < NULL)
    if ((uint8 *)&module_inst < exec_env->native_stack_boundary) {
        aot_set_exception_with_id(module_inst, EXCE_NATIVE_STACK_OVERFLOW);
        return false;
    }
```

Would one of the below be better?
- `if (!exec_env->native_stack_boundary && &moudle_inst < exec_env->native_stack_bounbdary)` 
- Or use `UINT64_MAX`  instead of `NULL` if `native_stack_boundary` isn't provided.

build-scripts/config_common.cmake
```diff
@@ -262,3 +262,7 @@ if (WAMR_BUILD_LOAD_CUSTOM_SECTION EQUAL 1)
     add_definitions (-DWASM_ENABLE_LOAD_CUSTOM_SECTION=1)
     message ("     Load custom section enabled")
 endif ()
+if (WAMR_BUILD_STACK_GUARD_SIZE GREATER 0)
+    add_definitions (-DWASM_STACK_GUARD_SIZE=${WAMR_BUILD_STACK_GUARD_SIZE})
+    message("     Custom stack guard size: " ${WAMR_BUILD_STACK_GUARD_SIZE})
```
wenyongh, line 267: How about changing `message(` to `message (`?

wenyongh:
> in that case, such a comparison looks a little wired. It will be passed but ...
> 
> ```c
>     //  if (&module_inst < NULL)
>     if ((uint8 *)&module_inst < exec_env->native_stack_boundary) {
>         aot_set_exception_with_id(module_inst, EXCE_NATIVE_STACK_OVERFLOW);
>         return false;
>     }
> ```
> 
> Would one of the below be better?
> 
> * `if (!exec_env->native_stack_boundary && &moudle_inst < exec_env->native_stack_bounbdary)`
> * Or use `UINT64_MAX`  instead of `NULL` if `native_stack_boundary` isn't provided.

No need to do that, if exec_env->native_stack_boundary is NULL, this condition
`if ((uint8 *)&module_inst < exec_env->native_stack_boundary)` will be always false,
here the check is OK. The issue is in the platform API itself.

build-scripts/config_common.cmake
no1wudi: OK

PR #1375, opened 2022-08-11
Implement i64.div and i64.rem intrinsics

no1wudi:


core/iwasm/aot/aot_intrinsic.c
```diff
@@ -514,6 +542,15 @@ add_intrinsic_capability(AOTCompContext *comp_ctx, uint64 flag)
     }
 }
 
+static void
+add_i64_common_intrinsics(AOTCompContext *comp_ctx)
```
lum1n0us, line 546: who is going to call this function?

lum1n0us (commented):
Just one question

core/iwasm/aot/aot_intrinsic.c
no1wudi: I guess all 32-bit platform will need it, but only riscv32 verified now.

lum1n0us:
LGTM

PR #1377, opened 2022-08-11
let iwasm return non-zero value when running failed

xujuntwt95329:


product-mini/platforms/posix/main.c
```diff
@@ -294,6 +294,7 @@ static char global_heap_buf[10 * 1024 * 1024] = { 0 };
 int
 main(int argc, char *argv[])
 {
+    int32 ret = -1;
```
wenyongh, line 297: The compiler complains "warning: variable ‘ret’ set but not used" when `-DWAMR_BUILD_SPEC_TEST=1`. How about:
```
#if WASM_ENABLE_SPEC_TEST == 0
    int32 ret = -1;
#endif

File: product-mini/platforms/posix/main.c [hunk 2]
```diff
@@ -570,6 +570,8 @@ main(int argc, char *argv[])
     else
         app_instance_main(wasm_module_inst);
 
+    ret = 0;
```
wenyongh, line 573: Same as above, how about:
```
#if WASM_ENABLE_SPEC_EST == 0
    ret = 0;
#endif

Comment (wenyongh) [reply product-mini/platforms/posix/main.c:573]: Same for other files?

Comment (xujuntwt95329) [reply product-mini/platforms/posix/main.c:573]: How about using `(void)ret;` in L601?

Comment (wenyongh) [reply product-mini/platforms/posix/main.c:573]: OK, before return 0.

Comment (xujuntwt95329) [reply product-mini/platforms/posix/main.c:573]: Sure, updated.
Thanks!

**Comment by eloparco** [comment]
Hello, I was looking into adding some tests (for WASI) and I see that spec tests are always enabled https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/tests/wamr-test-suites/test_wamr.sh#L177.

What's the reason behind always returning 0 when spec tests are enabled i.e. `WASM_ENABLE_SPEC_TEST != 0`, as introduced by this PR?

In my case I'd like to get a non-zero value when there's a failure so I'll have to turn `WAMR_BUILD_SPEC_TEST` off.

**Comment by eloparco** [comment]
Maybe it's not necessary anymore? I ran the GitHub actions after replacing
```c
    (void)ret;
    return 0;
#else
    return ret;
#endif
```
with just `return ret` and the spec tests are passing. Am I missing anything?

wenyongh:
> Maybe it's not necessary anymore? I ran the GitHub actions after replacing
> 
> ```c
> #if WASM_ENABLE_SPEC_TEST != 0
>     (void)ret;
>     return 0;
> #else
>     return ret;
> #endif
> ```
> 
> with just `return ret` and the spec tests are passing. Am I missing anything?

Yes, seems that it's not necessary, it returned 0 before this PR, and in this PR we let it return 0 when spec test is enabled in case that the spec test script cannot handle the return value correctly when a non-zero is returned, e.g. loading a mal-formed wasm binary file. It seems that the spec test scripts work correctly and we can replace the code to `return ret`.

How about we fix it in the main branch, and then merge main branch into dev/wasi_threads?

PR #1378, opened 2022-08-11
Interp/Fast-JIT optimization

FromLiQg:
refine wasm_type_equal for call_indirect for intel internal #402
To Wenyong for code review, there is a bug in code and it cannot pass the case of spec (call_indirect.wast) while using  these parameters: 
`
Build iwasm with compile flags with  `
`-DWAMR_BUILD_TARGET=X86_64  `
`-DWAMR_BUILD_INTERP=0 `
`-DWAMR_BUILD_FAST_INTERP=0 `
`-DWAMR_BUILD_JIT=1 `
`-DWAMR_BUILD_AOT=1 `
`-DWAMR_BUILD_SPEC_TEST=1 `
`-DWAMR_BUILD_BULK_MEMORY=1 ` 
`-DWAMR_BUILD_MULTI_MODULE=0 `
`-DWAMR_BUILD_REF_TYPES=1 `
`-DWAMR_BUILD_SIMD=0  for spec test `
`Namespace(aot_flag=False, cases=[], clean_up_flag=True, gc_flag=False, multi_module_flag=False, multi_thread_flag=False, parl_flag=False, sgx_flag=False, simd_flag=False, target='X86_64', verbose_flag=True, xip_flag=False)`
`


core/iwasm/interpreter/wasm.h
```diff
@@ -626,11 +629,17 @@ wasm_get_smallest_type_idx(WASMType **types, uint32 type_count,
                            uint32 cur_type_idx)
 {
     uint32 i;
-
+#if WASM_ENABLE_TYPE_POINTER_COMPARE  != 0
+    for (i = 0; i < cur_type_idx; i++) {
+        if (types[cur_type_idx] != types[i])
```
lum1n0us, line 634: please inlight me here, L639 will return `i` when types are equal. But L634 will return `i` when types are not equal?

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -1419,10 +1419,18 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                     cur_func_type = cur_func->u.func_import->func_type;
                 else
                     cur_func_type = cur_func->u.func->func_type;
+#if WASM_ENABLE_TYPE_POINTER_COMPARE != 0
```
lum1n0us, line 1422: maybe?
```suggestion
   if (
#if WASM_ENABLE_TYPE_POINTER_COMPARE != 0
    cur_type != cur_func_type ||
#endif
    wasm_type_equal(cur_type, cur_func_type)) {
```

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -612,6 +628,22 @@ load_type_section(const uint8 *buf, const uint8 *buf_end, WASMModule *module,
...
             type->ret_cell_num = (uint16)ret_cell_num;
         }
+#if WASM_ENABLE_TYPE_POINTER_COMPARE != 0
+        /* Normalize the type list, check whehter two types in the type list
+           are equivalent, and if yes, free the later one and set it to the
+           former one */
+        for (i = 0; i < type_count - 1; i++) {
+            uint32 j;
+            for (j = i + 1; j < type_count; j++) {
+                if (module->types[j]->ref_count == 1
```
lum1n0us, line 638: Does L638 mean we don't normalize the type when `ref_count != 1` ?

lum1n0us (commented):
A little confused.

core/iwasm/interpreter/wasm_loader.c
wenyongh: Should initialize the type->ref_count, set it to 1 when initializing module->types[i].

And no need to check whether wasm_type_equal here, should add the check after module->types[i] is created, if there is same type in types[0] to types[i-1], increase the latter's ref_count and destroy current type.

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -415,6 +415,22 @@ const_str_list_insert(const uint8 *str, uint32 len, WASMModule *module,
     return node->str;
 }
 
+
```
wenyongh, line 418: Should also apply same changes to wasm_mini_loader.c, which can be compiled with 'cmake -DWAMR_BUILD_MINI_LOADER=1`.

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -611,6 +626,16 @@ load_type_section(const uint8 *buf, const uint8 *buf_end, WASMModule *module,
             }
             type->param_cell_num = (uint16)param_cell_num;
             type->ret_cell_num = (uint16)ret_cell_num;
+
+            /* If there is already a same type created, use it instead */
+            for (j = 0; j < i; j++) {
+                if (wasm_type_equal(type, module->types[j])) {
```
wenyongh, line 632: Had better check possible integer overflow before L633:
```C
if (module->types[j]->ref_count == UINT16_MAX) {
                  "wasm type's ref count too large");
    return false;
}
```
When module->types[j]->ref_count reaches UINT16_MAX, `module->types[j]->ref_count++` will be 0.

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -415,6 +415,20 @@ const_str_list_insert(const uint8 *str, uint32 len, WASMModule *module,
...
+destroy_wasm_type(WASMType *type)
+{
+    if (type->ref_count > 1) {
+        /* The type is referenced by other types
+           of current wasm module */
+        type->ref_count--;
+        return;
+    }
+
+    /* Free the type */
```
xujuntwt95329, line 428: This comment is un-necessary, the code is self explanation

core/iwasm/interpreter/wasm_mini_loader.c
```diff
@@ -248,6 +248,20 @@ const_str_list_insert(const uint8 *str, uint32 len, WASMModule *module,
...
+destroy_wasm_type(WASMType *type)
+{
+    if (type->ref_count > 1) {
+        /* The type is referenced by other types
+           of current wasm module */
+        type->ref_count--;
+        return;
+    }
+
+    /* Free the type */
```
xujuntwt95329, line 261:
```suggestion
```

xujuntwt95329 (commented):
LGTM

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -853,6 +853,9 @@ globals_instantiate(const WASMModule *module, WASMModuleInstance *module_inst,
     *p_global_data_size = global_data_offset;
     (void)module_inst;
     return globals;
+fail:
+    globals_deinstantiate(globals);
```
wenyongh, line 857: Use wasm_runtime_free

PR #1379, opened 2022-08-12
Remove some unused codes

wenyongh:
Remove some unused fields in module instance and the related codes so as
to reduce the footprint, which are introduced by emsdk non-standalone
mode (e.g. -DSIDE_MODULE=1) and are not required now.


lum1n0us:
LGTM

Issue #1371, opened 2022-08-10
Missing bounds check in code generated by AOT


lei-april:
The following wasm module:
```
(module
  (func (result i32)
    (i32.load (i32.const 0))
  )
  (memory 0)
)
```

when compiled by `wamrc` (version f3f8d684b319de3f7cb0895f4b7be88e796ec4a5) with option `--bounds-checks=1`, produces the following machine code:
```
0000000000000000 <aot_func#0>:
   0:	48 8b 47 10          	mov    0x10(%rdi),%rax
   4:	48 8b 80 58 01 00 00 	mov    0x158(%rax),%rax
   b:	8b 00                	mov    (%rax),%eax
   d:	c3                   	retq
```

Apparently, there lacks bounds check in the generated code. If `iwasm` is built with hardware bound check disabled, running this wasm module would probably crash.

penzn:
Wasm memory index starts with 0, this should be valid code. Try passing another `i32` and using it instead of `i32.const 0`, this should generate a bounds check in case passed value is too big.

lei-april:
> Wasm memory index starts with 0, this should be valid code. Try adding `i32` parameter to the function and using it instead of `i32.const 0`, this should generate a bounds check in case passed value is too big.

Yes, using a non-const address would make `wamrc` generate bounds check. But that doesn't mean bounds check is not needed for const addresses.

penzn:
Check for 0 addresses can probably be elided as always within bounds, right @wenyongh?

wenyongh:
> Check for 0 addresses can probably be elided as always within bounds, right @wenyongh?

Yes, if the page count is not 0. In fact, if the const address and its byte count to access are in the range of the initial page, we can ignore the check, as the linear memory cannot shrink -- it can only keep unchanged or grow by opcode memory.grow. For example, if the initial page count is 1 and the bytes of each page is 65536, then the initial linear memory size must be at least 65536, the const address for i32.load/store can be 0 to 65532, no need to add check for them.

@lei-april Thanks for reporting the issue! It should only occur when the initial page count is 0 and there is no memory.grow opcode, we will fix it soon.

penzn:
Sorry, I misread `memory 0` as memory # 0 that spec talks about sometimes.

wenyongh:
Close this issue as it was fixed.

Issue #1289, opened 2022-07-12
Fail to set the max memory with c-api ?


Zhangyx24:
## Steps to reproduce
I tried to execute the test.wat:
[test.txt](https://github.com/bytecodealliance/wasm-micro-runtime/files/9091706/test.txt) both in command line and c-api. 

#### (1)command line
`iwasm test.wat`

#### (2)c-api
The c file is 
[test-in-c.txt](https://github.com/bytecodealliance/wasm-micro-runtime/files/9091728/test-in-c.txt)
**Interpreter mode :** `cd build; cmake .. ; make; ./test`
**AOT mode :** `cd build; cmake -DWAMR_BUILD_INTERP=0 -DWAMR_BUILD_AOT=1 ..; make; ./test`

## Expected result
Allow the memory of 65536, and don't print error.

#### (2)c-api in **interpreter** mode **failed** and print:
![截屏2022-07-12 下午5 55 41](https://user-images.githubusercontent.com/28601300/178464082-8686fbc2-5a8a-4f3a-a414-2f65b24804e5.png)
**c-api in AOT mode succeed**.

Both macOS and ubuntu got the same result.

lum1n0us:
nice catching! will fix it soon.

Zhangyx24:
Thanks a lot !  Waiting for your fix.

lum1n0us:
After a long discussion, we are going to *keep different* results of interpreter and AOT when setting the memory minimum page count with 65536(4GB) in current phase.

The mainly reason on it is that interpreter and AOT are using different design to manage the linear memory boundary. And we are planing to unite both. And when it's done, both mode shall have same responses to 65536.

So, there will be no fix for the failure of meeting `(memory 65536)` in interpreter.

Zhangyx24:
Do you mean that after unite both, interpreter and AOT will response same to 65536?
Will they both allow 65536 or don't?

lum1n0us:
After uniting both, both interpreter and AOT will allow 65536 pages in the future. But not in the subsequent few releases.

lum1n0us:
https://github.com/bytecodealliance/wasm-micro-runtime/issues/1301

Zhangyx24:
Thanks for your answer!

PR #1380, opened 2022-08-12
Fix linear memory page count issues

wenyongh:
Fix issue reported in #1289 and #1371.
Enable to set the max page count to 65536.


core/iwasm/aot/aot_runtime.c
```diff
@@ -2062,94 +2070,103 @@ aot_enlarge_memory(AOTModuleInstance *module_inst, uint32 inc_page_count)
...
+        total_size_new = UINT32_MAX;
     }
 
 #if WASM_ENABLE_SHARED_MEMORY != 0
-    if (memory_inst->is_shared) {
-        /* For shared memory, we have reserved the maximum spaces during
-            instantiate, only change the cur_page_count here */
-        memory_inst->cur_page_count = total_page_count;
+    if (memory->is_shared) {
+        memory->num_bytes_per_page = UINT32_MAX;
```
lum1n0us, line 2083: If the code didn't go in the block between L2075 and L2078, do we still have to change `num_bytes_per_page`?

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -208,7 +208,7 @@ memory_instantiate(WASMModuleInstance *module_inst, uint32 num_bytes_per_page,
         num_bytes_per_page += heap_size;
         if (num_bytes_per_page < heap_size) {
             set_error_buf(error_buf, error_buf_size,
-                          "memory size must be at most 65536 pages (4GiB)");
+                          "failed to insert app heap into linear memory");
```
lum1n0us, line 211: shall we append something like "maybe use --heap_size=0 and try again"?

lum1n0us (commented):
a little comments

core/iwasm/aot/aot_runtime.c
wenyongh: No, only when the memory data size is out of range, or, it reaches 4GB (65536 pages * 65536 bytes_each_page), we need to resize it. As the type of aot_inst->memory_data_size is uint32, and its range is 0 to 4GB-1, and we had better keep the AOT ABI unchanged.

core/iwasm/interpreter/wasm_runtime.c
wenyongh: OK, let's add more prompt.

lum1n0us:
LGTM

PR #1382, opened 2022-08-12
reserve one pointer size for fast-interp compiled_code_size

xujuntwt95329:


core/iwasm/interpreter/wasm_loader.c
```diff
@@ -6630,7 +6644,7 @@ wasm_loader_prepare_bytecode(WASMModule *module, WASMFunction *func,
         }
         p = func->code;
         func->code_compiled = loader_ctx->p_code_compiled;
-        func->code_compiled_size = loader_ctx->code_compiled_size;
+        func->code_compiled_size = loader_ctx->code_compiled_peak_size;
```
wenyongh, line 6647: func->code_compiled_size is used by wasm_runtime wasm_get_func_code_end() function, which might impact the `frame_ip_end` in wasm_interp_fast.c and the interpreter's behavior? Shall we need to change it here?

xujuntwt95329: Yes, I modified it back, but the shortcoming is the memory profiler may not be so accurate

wenyongh: OK, no need to be so accurate for memory profiler, this almost doesn't occur except few few case.

PR #1386, opened 2022-08-15
nuttx: Provide a default WASM_STACK_GUARD_SIZE

yamt:
Unbreak builds with a bit older nuttx/apps for now.

Note: The default value here (0) was chosen to match the new default
in nuttx/apps, which is different from the old value used before
the introduction of WASM_STACK_GUARD_SIZE. (1024)


no1wudi:
`WASM_STACK_GUARD_SIZE` is a simple rename of `RESERVED_BYTES_TO_NATIVE_STACK_BOUNDARY` which is also not defined by NuttX, so I think it's OK to keep its origin behavior (use default 1024 for safety).

Is there any build break on older version of NuttX ?

yamt:
> `WASM_STACK_GUARD_SIZE` is a simple rename of `RESERVED_BYTES_TO_NATIVE_STACK_BOUNDARY` which is also not defined by NuttX, so I think it's OK to keep its origin behavior (use default 1024 for safety).

i prefer to be consistent with the value used by new version:
https://github.com/apache/incubator-nuttx-apps/blob/a3ccf59204c0716be18b1d2d43ad417ce5fd1c06/interpreters/wamr/Kconfig#L84-L89

> Is there any build break on older version of NuttX ?

yes. something like the following.
```
wamr/core/iwasm/common/wasm_exec_env.c:195:43: error: use of undeclared identifier 'CONFIG_INTERPRETERS_WAMR_STACK_GUARD_SIZE'
```

no1wudi (approved):

PR #1388, opened 2022-08-16
Expose wasm_runtime_get_exec_env_singleton to the API users

yamt:
Sometimes it can be useful to access the singleton.
Eg. use wasm_runtime_set_user_data on it.

### DIFF
--- a/core/iwasm/common/wasm_runtime_common.h
+++ b/core/iwasm/common/wasm_runtime_common.h
@@ -592,7 +592,8 @@ wasm_runtime_call_indirect(WASMExecEnv *exec_env, uint32 element_indices,
 bool
 wasm_runtime_create_exec_env_singleton(WASMModuleInstanceCommon *module_inst);
 
-WASMExecEnv *
+/* See wasm_export.h for description */
+WASM_RUNTIME_API_EXTERN WASMExecEnv *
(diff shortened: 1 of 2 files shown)


core/iwasm/common/wasm_runtime_common.h
```diff
@@ -592,9 +592,6 @@ wasm_runtime_call_indirect(WASMExecEnv *exec_env, uint32 element_indices,
 bool
 wasm_runtime_create_exec_env_singleton(WASMModuleInstanceCommon *module_inst);
 
-WASMExecEnv *
```
wenyongh, line 595: Had better not remove the declaration here, just add WASM_RUNTIME_API_EXTERN like others:
```C
WASM_RUNTIME_API_EXTERN WASMExecEnv *
wasm_runtime_get_exec_env_singleton(WASMModuleInstanceCommon *module_inst);
```

yamt: i moved this to wasm_export.h. do you mean to have it in both headers?

yamt: i think i got it now. although i feel it's a weird convention, i updated the patch to follow it. thank you.

wenyongh: Yes, prefer to have it in both files.

PR #1390, opened 2022-08-18
Enhance cmake makefiles

wenyongh:
Upgrade `cmake_minimum_required` from `(VERSION 2.8)` to `(VERSION 2.9)` to
yield the warning:
"Compatibility with CMake < 2.8.12 will be removed from a future version of CMake"

Add "-Wno-unused" for CMAKE_CXX_FLAGS to yield the compilation warnings
when build LLVM JIT.

Fix the link error when code coverage is enabled.


lum1n0us:
LGTM

PR #1401, opened 2022-08-22
Fix a bug that read a null pointer in jit_cc_destroy

FromLiQg:
Try to fix bug that it read a null pointer after initialize a compilation context for fast jit failed.
When memory_allocator_alloc return NULL, jit_cc_init will call jit_cc_destroy and we try to free the memory of cc.
But we ingored to do some check for cc->entry_label that may be 0 (This represents an initialization failure.), which causes the program to read a null pointer.
PS: This PR is for  [internal issues #403](https://github.com/wasm-micro-runtime/wamr-internal-dev/issues/403)


lum1n0us:
IIUC, `form_and_translate_func()` will make sure every function has an `entry` and an `exit`. So I am curious about when the new NULL check will work.

FromLiQg:
> IIUC, `form_and_translate_func()` will make sure every function has an `entry` and an `exit`. So I am curious about when the new NULL check will work.
We use "Instrument Test" to test abnormal cases.
In this case, we force the function `mem_allocator_malloc` to return `NULL` when **the specific call stack occur**.
And the function  `form_and_translate_func()` may not be called.
The detail case is in https://github.com/wasm-micro-runtime/wamr-internal-dev/issues/403.
I described the whole situation.

FromLiQg:
> IIUC, `form_and_translate_func()` will make sure every function has an `entry` and an `exit`. So I am curious about when the new NULL check will work.


We use "Instrument Test" to test abnormal cases.
In this case, we force the function mem_allocator_malloc to return NULL when the specific call stack occur.
And the function form_and_translate_func() may not be called.
The detail case is in https://github.com/wasm-micro-runtime/wamr-internal-dev/issues/403.
I described the whole situation.

FromLiQg:
> IIUC, `form_and_translate_func()` will make sure every function has an `entry` and an `exit`. So I am curious about when the new NULL check will work.

We use "Instrument Test" to test abnormal cases.
In this case, we force the function mem_allocator_malloc to return NULL when the specific call stack occur.
And the function form_and_translate_func() may not be called.
The detail case is in https://github.com/wasm-micro-runtime/wamr-internal-dev/issues/403.
I described the whole situation.
@lum1n0us

xujuntwt95329:
@lum1n0us This issue is found by instrumentation test (it is a destructive test to cover abnormal cases). 
This bug indicates that somewhere doesn't check and deal with the return value of `wasm_runtime_malloc` correctly.

FromLiQg:
I know what you mean.
The original design was:
```cpp

JitCompContext *
jit_cc_init(JitCompContext *cc, unsigned htab_size)
{
    JitBasicBlock *entry_block, *exit_block;
    unsigned i, num;

    memset(cc, 0, sizeof(*cc));
    cc->_reference_count = 1;
    jit_annl_enable_basic_block(cc);

    /* Create entry and exit blocks.  They must be the first two
       blocks respectively.  */
    if (!(entry_block = jit_cc_new_basic_block(cc, 0))
        || !(exit_block = jit_cc_new_basic_block(cc, 0)))
        goto fail;
/*ignored codes for convenience*/
    cc->entry_label = jit_basic_block_label(entry_block);
    cc->exit_label = jit_basic_block_label(exit_block);
/*ignored codes for convenience*/
fail:
    jit_cc_destroy(cc);
    return NULL;
}
``` 
And when exec `entry_block = jit_cc_new_basic_block(cc, 0)`，it goto fail.
So `cc->entry_label = 0`, and I think this represent the error case. So I use this case to detect NULL.
or do we need to do something in `jit_cc_new_basic_block`?
But I  have no idea dealing with the `else` case in this function to fix the bug.
```cpp

JitBasicBlock *
jit_cc_new_basic_block(JitCompContext *cc, int n)
{
    JitReg label = jit_cc_new_label(cc);
    JitBasicBlock *block = NULL;

    if (label && (block = jit_basic_block_new(label, n)))
        /* Void 0 register indicates error in creation.  */
        *(jit_annl_basic_block(cc, label)) = block;
    else
        jit_set_last_error(cc, "create basic block failed");

    return block;
}
```

PR #1402, opened 2022-08-22
Destroy fast-jit compiler after destroy loaded modules

wenyongh:
Destroy Fast-JIT compiler after destroying the modules loaded by
multi-module feature, since the Fast JIT's code cache allocator may
be used by these modules. If the Fast JIT's code cache allocator was
destroyed, then runtime will fail to destroy these modules.

And fix the issue of destroying import module's memory instance.


lum1n0us:
LGTM

PR #1403, opened 2022-08-22
debug: Remove unused platform_port

yamt:
I left a placeholder in RuntimeInitArgs just because I don't want to
bump the major verision for this change alone.


xujuntwt95329 (commented):
LGTM

PR #1407, opened 2022-08-23
wasm_application.c: Do not start debug instance automatically

yamt:
It's better to leave the control to embedders. (eg. product-mini)

background: I have a use case where
* The application runs multiple module instances.
* I want to debug only one of them.

### DIFF
--- a/product-mini/platforms/posix/main.c
+++ b/product-mini/platforms/posix/main.c
@@ -570,6 +570,23 @@ main(int argc, char *argv[])
         goto fail3;
     }
 
+#if WASM_ENABLE_DEBUG_INTERP != 0
+    if (ip_addr != NULL) {
+        wasm_exec_env_t exec_env =
+            wasm_runtime_get_exec_env_singleton(wasm_module_inst);
+        uint32_t debug_port;
+        if (exec_env == NULL) {
+            printf("%s\n", wasm_runtime_get_exception(wasm_module_inst));
+            goto fail4;
+        }
+        debug_port = wasm_runtime_start_debug_instance(exec_env);
+        if (debug_port == 0) {
+            printf("Failed to start debug instance\n");
+            goto fail4;
+        }
+    }
+#endif
+
     if (is_repl_mode)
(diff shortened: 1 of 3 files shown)


product-mini/platforms/posix/main.c
```diff
@@ -570,6 +570,22 @@ main(int argc, char *argv[])
...
 
+#if WASM_ENABLE_DEBUG_INTERP != 0
+    if (ip_addr != NULL) {
+        wasm_exec_env_t exec_env =
+            wasm_runtime_get_exec_env_singleton(wasm_module_inst);
+        if (exec_env == NULL) {
+            printf("%s\n", wasm_runtime_get_exception(wasm_module_inst));
+            goto fail4;
+        }
+        uint32_t debug_port = wasm_runtime_start_debug_instance(exec_env);
```
xujuntwt95329, line 581: Had better declare variables at the beginning of a scope, otherwise some compiler with `C90` may complain

product-mini/platforms/windows/main.c
```diff
@@ -437,6 +437,22 @@ main(int argc, char *argv[])
...
 
+#if WASM_ENABLE_DEBUG_INTERP != 0
+    if (ip_addr != NULL) {
+        wasm_exec_env_t exec_env =
+            wasm_runtime_get_exec_env_singleton(wasm_module_inst);
+        if (exec_env == NULL) {
+            printf("%s\n", wasm_runtime_get_exception(wasm_module_inst));
+            goto fail4;
+        }
+        uint32_t debug_port = wasm_runtime_start_debug_instance(exec_env);
```
xujuntwt95329, line 448: same as above

product-mini/platforms/posix/main.c
yamt: fixed. thank you.

product-mini/platforms/windows/main.c
yamt: fixed. thank you.

PR #1411, opened 2022-08-24
Add support for IPv6 in WAMR

loganek:
For now this implementation only covers posix platforms, as defined in MVP #1336

### DIFF
--- a/core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
+++ b/core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
@@ -59,7 +59,7 @@ sockaddr_to_wasi_addr(const struct sockaddr *sock_addr, socklen_t addrlen,
             ((struct sockaddr_in *)sock_addr)->sin_port, wasi_addr);
     }
     else if (AF_INET6 == sock_addr->sa_family) {
-        assert(sizeof(struct sockaddr_in6) <= addrlen);
+        assert(sizeof(struct sockaddr_in6) == addrlen);
         ipv6_addr_to_wasi_addr(
             (uint16_t *)((struct sockaddr_in6 *)sock_addr)->sin6_addr.s6_addr,
             ((struct sockaddr_in6 *)sock_addr)->sin6_port, wasi_addr);
--- a/core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
+++ b/core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
@@ -31,21 +31,38 @@ ipv4_addr_to_wasi_addr(uint32_t addr_num, uint16_t port, __wasi_addr_t *out)
     out->addr.ip4.addr.n3 = (addr_num & 0x000000FF);
 }
 
+static void
+ipv6_addr_to_wasi_addr(uint16_t *addr, uint16_t port, __wasi_addr_t *out)
+{
+    out->kind = IPv6;
+    out->addr.ip6.port = ntohs(port);
+    out->addr.ip6.addr.n0 = ntohs(addr[0]);
+    out->addr.ip6.addr.n1 = ntohs(addr[1]);
+    out->addr.ip6.addr.n2 = ntohs(addr[2]);
+    out->addr.ip6.addr.n3 = ntohs(addr[3]);
+    out->addr.ip6.addr.h0 = ntohs(addr[4]);
+    out->addr.ip6.addr.h1 = ntohs(addr[5]);
+    out->addr.ip6.addr.h2 = ntohs(addr[6]);
+    out->addr.ip6.addr.h3 = ntohs(addr[7]);
+}
+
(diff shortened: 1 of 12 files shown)


core/shared/platform/common/posix/posix_socket.c
```diff
@@ -9,28 +9,57 @@
...
+        v6->sin6_port = htons(port);
+        return true;
+    }
+
+    return false;
 }
 
 int
-os_socket_create(bh_socket_t *sock, int tcp_or_udp)
+os_socket_create(bh_socket_t *sock, int ip4_or_ip6, int tcp_or_udp)
```
wenyongh, line 38: The parameter definitions make a little confused, how about change to:
`bool is_ipv4, bool is_tcp`? And add comments in the API declaration.

core/shared/platform/common/posix/posix_socket.c
```diff
@@ -166,13 +201,28 @@ os_socket_shutdown(bh_socket_t socket)
 }
 
 int
-os_socket_inet_network(const char *cp, uint32 *out)
+os_socket_inet_network(uint8 ipv4_or_ipv6, const char *cp, uint8 *out)
```
wenyongh, line 204: Same as above, how about `bool is_ipv4`?
And for `uint8 *out`, there may be address unaligned issue, if the input is not 4-byte aligned, L218, `*ipv4_buffer = ntohl(*ipv4_buffer);` may cause core dump in some targets, e.g. arm?

How about define an union type, like:
```C
union {
    uint32 ipv4;
    uint16 ipv6[8];
};
```

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
```diff
@@ -31,21 +31,38 @@ ipv4_addr_to_wasi_addr(uint32_t addr_num, uint16_t port, __wasi_addr_t *out)
...
+    out->kind = IPv6;
+    out->addr.ip6.port = ntohs(port);
+    out->addr.ip6.addr.n0 = ntohs(addr[0]);
+    out->addr.ip6.addr.n1 = ntohs(addr[1]);
+    out->addr.ip6.addr.n2 = ntohs(addr[2]);
+    out->addr.ip6.addr.n3 = ntohs(addr[3]);
+    out->addr.ip6.addr.h0 = ntohs(addr[4]);
+    out->addr.ip6.addr.h1 = ntohs(addr[5]);
+    out->addr.ip6.addr.h2 = ntohs(addr[6]);
+    out->addr.ip6.addr.h3 = ntohs(addr[7]);
```
wenyongh, line 46: Found that there are some similar conversions, including PR #1413 [wasi_socket_ext.c](https://github.com/bytecodealliance/wasm-micro-runtime/pull/1413/files#diff-798a0826792097576b78eaa27018727f43ecae36c7296bdafada5bf4c614b54e), shall we add common APIs for the conversions?

core/shared/platform/common/posix/posix_socket.c
loganek: Good point, updated.

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
loganek: this is a copy-paste from that file. Once we merge one of the PRs, we can remove this part from the second. Just wanted to have both PRs working.

wenyongh:
LGTM

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
```diff
@@ -31,21 +31,38 @@ ipv4_addr_to_wasi_addr(uint32_t addr_num, uint16_t port, __wasi_addr_t *out)
...
+        assert(sizeof(struct sockaddr_in) <= addrlen);
 
         ipv4_addr_to_wasi_addr(
             ((struct sockaddr_in *)sock_addr)->sin_addr.s_addr,
             ((struct sockaddr_in *)sock_addr)->sin_port, wasi_addr);
     }
     else if (AF_INET6 == sock_addr->sa_family) {
-        // TODO: IPV6
-        ret = __WASI_ERRNO_AFNOSUPPORT;
+        assert(sizeof(struct sockaddr_in6) <= addrlen);
```
lum1n0us, line 62: IIUC, `sock_addr` should be either a value of `struct sockaddr_in` or `struct sockaddr_in6`. So, L55 using `<=` is right. But L62 should use `==`. There is no third option.

lum1n0us (approved):
Just one little suggestion.

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
loganek: done

lum1n0us:
LGTM

PR #1413, opened 2022-08-24
Implement getaddrinfo in wasi_socket_ext.c

loganek:
So getaddrinfo() can be used when compiling C programs


core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
```diff
@@ -78,8 +78,26 @@ wasi_addr_to_sockaddr(const __wasi_addr_t *wasi_addr,
...
+            uint16_t *addr_buf = (uint16_t *)sock_addr_in6.sin6_addr.s6_addr;
+
+            addr_buf[0] = htons(wasi_addr->addr.ip6.addr.n0);
+            addr_buf[1] = htons(wasi_addr->addr.ip6.addr.n1);
+            addr_buf[2] = htons(wasi_addr->addr.ip6.addr.n2);
+            addr_buf[3] = htons(wasi_addr->addr.ip6.addr.n3);
+            addr_buf[4] = htons(wasi_addr->addr.ip6.addr.h0);
+            addr_buf[5] = htons(wasi_addr->addr.ip6.addr.h1);
+            addr_buf[6] = htons(wasi_addr->addr.ip6.addr.h2);
+            addr_buf[7] = htons(wasi_addr->addr.ip6.addr.h3);
```
wenyongh, line 92: Had better add APIs to do the conversion of addr_buf and addr after PR #1411 is merged

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
```diff
@@ -282,3 +300,153 @@ getpeername(int sockfd, struct sockaddr *addr, socklen_t *addrlen)
...
+                                         &max_info_size);
+        if (error != __WASI_ERRNO_SUCCESS) {
+            free(addr_info);
+            HANDLE_ERROR(error);
+        }
+    } while (max_info_size > addr_info_size);
+
+    if (addr_info_size == 0) {
+        *res = NULL;
+        return __WASI_ERRNO_SUCCESS;
```
wenyongh, line 414: Should free(addr_info) before return? The addr_info will be allocated no matter the addr_info_size is 0 or not

samples/socket-api/wasm-src/addr_resolve.c
```diff
@@ -0,0 +1,66 @@
...
+
+    printf("Host: %s\n", host);
+    while (res) {
+        switch (res->ai_family) {
+            case AF_INET:
+                ptr = &((struct sockaddr_in *)res->ai_addr)->sin_addr;
+                break;
+            case AF_INET6:
+                ptr = &((struct sockaddr_in6 *)res->ai_addr)->sin6_addr;
+                break;
```
lum1n0us, line 44: `break` goes out of the `switch`, but still in `while`. So L46 will use `ptr` in an uninitialized value.

samples/socket-api/wasm-src/addr_resolve.c
```diff
@@ -0,0 +1,66 @@
...
+        }
+        inet_ntop(res->ai_family, ptr, addrstr, 100);
+        printf("IPv%d address: %s (%s)\n", res->ai_family == AF_INET6 ? 6 : 4,
+               addrstr, res->ai_socktype == SOCK_STREAM ? "TCP" : "UDP");
+        res = res->ai_next;
+    }
+
+    freeaddrinfo(result);
```
lum1n0us, line 52: `freeaddrinfo(...)` will `free(...)` something before `result`. 

`free( ((char *)result - offsetof(struct aibuf, ai)); )`

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
```diff
@@ -282,3 +300,153 @@ getpeername(int sockfd, struct sockaddr *addr, socklen_t *addrlen)
...
+
+    free(addr_info);
+
+    return __WASI_ERRNO_SUCCESS;
+}
+
+void
+freeaddrinfo(struct addrinfo *res)
+{
+    void *start_ptr = ((char *)res - offsetof(struct aibuf, ai));
```
lum1n0us, line 450: It is a little confused expression.  Since `res` actually points to `struct aibuf[]`, how about
```c
struct aibuf *aibuf_res = (struct aibuf*)res;
free(aibuf_res);
```

or just `free(res)`.

lum1n0us (commented):
Here are my opinions

samples/socket-api/wasm-src/addr_resolve.c
loganek: yes, it's expected to stay in `while`. The `break` here and in L41 is not expected to break the while loop. `ptr` is only uninitialized if the `ai_family` is set to a value different than `AF_INET` or `AF_INET6`. Having said that, we should have a `default` label here in case the `af_family` has an unexpected value.

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
loganek: Yeah, I got confused by the implementation in [musl](https://git.musl-libc.org/cgit/musl/tree/src/network/freeaddrinfo.c), but looking at the C standard 6.7.2.1-13, I see that actually it's safe to do `free(res)`:
> A pointer to a structure object, suitably converted, points to its initial member (or if that member is a bit-field, then to the unit in which it resides), and vice versa. There may be unnamed padding within a structure object, but not at its beginning.

samples/socket-api/wasm-src/addr_resolve.c
lum1n0us: If `res->ai_family` is neither `AF_INET` nor `AF_INET6`, `ptr` will be some random value, and `inet_ntop` will be failed, and `addrstr` will be some random value, and `printf` may access some random address.

loganek: yes, that's why I added `default` label to the switch.

lum1n0us:
LGTM

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
```diff
@@ -317,3 +317,153 @@ getpeername(int sockfd, struct sockaddr *addr, socklen_t *addrlen)
...
+        error = wasi_addr_info_to_addr_info(&addr_info[i], ai);
+        if (error != __WASI_ERRNO_SUCCESS) {
+            free(addr_info);
+            free(aibuf_res);
+            HANDLE_ERROR(error)
+        }
+        ai->ai_next = i == addr_info_size - 1 ? NULL : &aibuf_res[i + 1].ai;
+    }
+
+    free(addr_info);
```
wenyongh, line 460: Shall we free aibuf_res here? If not, when it is freed?

loganek: This is expected to be freed in `freeaddrinfo()` by the user.

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
```diff
@@ -317,3 +317,153 @@ getpeername(int sockfd, struct sockaddr *addr, socklen_t *addrlen)
...
+        *res = NULL;
+        return __WASI_ERRNO_SUCCESS;
+    }
+
+    aibuf_res = malloc(addr_info_size * sizeof(struct aibuf));
+    if (!aibuf_res) {
+        free(addr_info);
+        HANDLE_ERROR(__WASI_ERRNO_NOMEM)
+    }
+
```
wenyongh, line 440: Had better `memset(aibuf_res, 0, addr_info_size * sizeof(struct aibuf))` here, seems some fields are not set in the following loop?

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
wenyongh: OK, it can be freed correctly because `ai` is the first field of struct struct aibuf. Not very easy to understand, had better to add comment here?

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
loganek: changed `malloc` to `calloc`

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
loganek: added a comment in the code

PR #1417, opened 2022-08-25
Add symbol check in wasm_c_api linkage

linzang:
The orignal implementation does not consider the symbol (function name)
when linking an imported function. It can cause wrong linked function
when two function have same type. And it also introduce the unnecessary
limitation of the sequence that the external function declared in C must
be same with import sequence declared in wasm (a counter-example as
externs[] defined in callback_with_symbol.c when ).

This patch compare the symbol if it is specified in C. which could help
fix the aforementioned issues.

### DIFF
--- a/core/iwasm/common/wasm_c_api.c
+++ b/core/iwasm/common/wasm_c_api.c
@@ -2407,6 +2407,60 @@ wasm_func_new_basic(wasm_store_t *store, const wasm_functype_t *type,
 }
 
 static wasm_func_t *
+wasm_func_new_with_symbol_basic(wasm_store_t *store,
+                                const wasm_functype_t *type,
+                                wasm_func_callback_t func_callback,
+                                const char *module_name, const char *func_name)
+{
+    wasm_func_t *func = NULL;
+
+    if (!type) {
+        goto failed;
+    }
+
+    if (!(func = malloc_internal(sizeof(wasm_func_t)))) {
+        goto failed;
+    }
+
+    func->store = store;
+    func->kind = WASM_EXTERN_FUNC;
+    func->func_idx_rt = (uint16)-1;
+    func->with_env = false;
+    func->u.cb = func_callback;
+    func->module_name = NULL;
+    func->name = NULL;
+
+    /* take symbol */
+    if (module_name != NULL) {
+        func->module_name = malloc_internal(sizeof(wasm_name_t));
+        if (func->module_name == NULL) {
+            goto failed;
+        }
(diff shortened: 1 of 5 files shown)


yamt:
> The orignal implementation does not consider the symbol (function name) when linking an imported function. It can cause wrong linked function when two function have same type. And it also introduce the unnecessary limitation of the sequence that the external function declared in C must be same with import sequence declared in wasm (a counter-example as externs[] defined in callback_with_symbol.c when ).
> 
> This patch compare the symbol if it is specified in C. which could help fix the aforementioned issues.

with wasm-c-api, isn't it user's responsibility to prepare the correct sequence of externs using wasm_module_imports etc?
to me it seems quite wrong to perform name-based matching at this layer.

my suggestion is to write a utility function, which prepares the externs vec using wasm_module_imports etc.
something like https://docs.rs/wasmer-c-api/2.0.0/wasmer_c_api/wasm_c_api/wasi/fn.wasi_get_imports.html but for your host functions.

linzang:
Hi @yamt, 
    Thanks for your comments. 

>      with wasm-c-api, isn't it user's responsibility to prepare the correct sequence of externs using wasm_module_imports etc?

In my case, we are not the developer of the wasm program, instead we are sort of a platform service that provide several external APIs from C  for wasm developer to use. So we have no idea what the sequece of the "import func" for every specific wasm program that run with our APIs. 

And IMHO, when the number of extern APIs are huge, it is very error-prone by just compare the type of the func at link time,  e.g.  `int add(int, int)` and `int mul(int, int)` can be linked wrongly without any error if the sequece in wasm are reversed.

And Yes, your suggestion of utilizing wasi_get_imports is also a good way to solve my issue, we can even make it automatically, but it seems like an extra tool to workaround a problem that wasm linker can accept a potiential 
error linkage. 

Maybe I am wrong and wasm linker is designed to behave like this. This PR also targeting to help me understand more about the linker behavior of wasm. 

Thanks, 
Lin

lum1n0us:
External items comparison only consider *type* instead of name (https://webassembly.github.io/spec/core/valid/types.html#import-subtyping).

It is different from the *traditional linkage*. WAMR did implement some kind of  `dload()`  features. But it is a WAMR private feature.

According to the spec, it is allowed to not just link `mul` with `add` but also link host functions with `add`, as long as keeping exactly types.

The name comparison is considered as users' responsibility. Here is an example from [envoy](https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/master/src/wamr/wamr.cc#L216)

PR #1419, opened 2022-08-25
Added socket send and recv timeout options

cimacmillan:
Added socket send and recv timeout options with implementation for posix platform

This is part of a extending support for sockets in WASI. https://github.com/bytecodealliance/wasm-micro-runtime/issues/1336

I've also added a WIP sample that sets and reads back the send and receive timeouts using the native function binding, however I still need to add the posix-like shim.


samples/socket-api/wasm-src/socket_opts.c
```diff
@@ -0,0 +1,45 @@
...
+
+    printf("[Client] Create socket\n");
+    socket_fd = socket(AF_INET, SOCK_STREAM, 0);
+    if (socket_fd == -1) {
+        perror("Create socket failed");
+        return EXIT_FAILURE;
+    }
+
+// TODO: Replace this with posix sockopts when available
+#ifdef __wasi__
```
lum1n0us, line 31: We'd like to use the original API and translate them to Wasm. Users should not learn another set of APIs to develop Wasm application

Since timeout is an opt of a socket, `SO_SNDTIMEO`, using `setsockopt(...)` and `getsockopt(..)` is better.

wenyongh:
@loganek do you have any comments?

samples/socket-api/wasm-src/socket_opts.c
loganek: I think this is just an intermediate step and it will be replaced with `getsockopt()` and `secksockopt()` once we have it implemented in `wasi_socket_ext.c`, is that right, @cimacmillan ?

core/iwasm/libraries/lib-socket/inc/wasi_socket_ext.h
```diff
@@ -445,6 +445,62 @@ __wasi_sock_set_send_buf_size(__wasi_fd_t fd)
...
+                                                        int32_t arg1)
+    __attribute__((__import_module__("wasi_snapshot_preview1"),
+                   __import_name__("sock_set_recv_timeout")));
+
+static inline __wasi_errno_t
+__wasi_sock_set_recv_timeout(__wasi_fd_t fd, uint64_t *timeout_us)
+{
+    return (__wasi_errno_t)
+        __imported_wasi_snapshot_preview1_sock_set_recv_timeout(
+            (int32_t)fd, (int32_t)timeout_us);
```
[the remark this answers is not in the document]
cimacmillan: Here I'm converting the `timeout_us` pointer to `int32_t`, rather than passing the value as int64. It's because of this limitation:

> export_native_api.md - As function parameters are always passed in 32 bits numbers, you can also use 'i' for the pointer type argument, then you must do all the address conversion and boundary checking in your native function.

Tested that passing ULLONG_MAX (rounded down to nearest 1000) sets and receives properly. There should probably be a specific test case for this though

samples/socket-api/wasm-src/socket_opts.c
cimacmillan: That was my original plan, but I'll raise a revision for it as part of this PR if that's preferred

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -1171,6 +1203,22 @@ wasi_sock_set_recv_buf_size(wasm_exec_env_t exec_env, wasi_fd_t fd,
...
+    wasm_module_inst_t module_inst = get_module_inst(exec_env);
+    wasi_ctx_t wasi_ctx = get_wasi_ctx(module_inst);
+    struct fd_table *curfds = NULL;
+
+    if (!wasi_ctx)
+        return __WASI_EACCES;
+
+    curfds = wasi_ctx_get_curfds(module_inst, wasi_ctx);
+
+    return wasmtime_ssp_sock_set_recv_timeout(curfds, fd, *timeout_us);
```
loganek, line 1219: before passing the `timeout_us` further, we need to check do check the memory using `validate_native_addr()`. See usage of that macro in this file for examples.

core/iwasm/libraries/lib-socket/inc/wasi_socket_ext.h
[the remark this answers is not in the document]
loganek: Doesn't that refer to pointer types? I think it's possible to pass int64, that's why we have `I` (uppercase `i`) as a registration parameter. You can also see some examples in wasi libc, e.g. `fd_pwrite`'s `offset` parameter is of `wasi_filesize_t` type which is a type alias of `uint64_t`.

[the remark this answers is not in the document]
cimacmillan: Yeah you're right I can use `I`, misunderstood this part. Will update

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
```diff
@@ -282,3 +282,80 @@ getpeername(int sockfd, struct sockaddr *addr, socklen_t *addrlen)
...
+    switch (level) {
+        case SOL_SOCKET:
+            switch (optname) {
+                case SO_RCVTIMEO:
+                    assert(*optlen == sizeof(struct timeval));
+                    error = __wasi_sock_get_recv_timeout(sockfd, &timeout_us);
+                    HANDLE_ERROR(error);
+                    *(struct timeval *)optval = time_us_to_timeval(timeout_us);
+                    return error;
+                    break;
```
loganek, line 317: there's no need for `break` if you already have `return` one line above that.

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
```diff
@@ -282,3 +282,80 @@ getpeername(int sockfd, struct sockaddr *addr, socklen_t *addrlen)
...
+int
+getsockopt(int sockfd, int level, int optname, void *__restrict optval,
+           socklen_t *__restrict optlen)
+{
+    __wasi_errno_t error;
+    uint64_t timeout_us;
+
+    switch (level) {
+        case SOL_SOCKET:
+            switch (optname) {
```
loganek, line 310: `switch` inside another `switch` makes the code a bit less readable, consider refactoring the code to avoid deeply nested code.

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
cimacmillan: Forgot to remove, thanks for catching

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
cimacmillan: Agree, maybe separate functions depending on the level? What'd you recommend?

loganek: I think splitting it into functions is not a bad idea. Also, I don't expect we're going to add support to level different than `SOL_SOCKET`, so you might consider failing early at the beginning of the function:
```
if (level != SOL_SOCKET) {
  return -1;
}
```

cimacmillan: Thanks, will split because I was planning to include a couple `IPPROTO_TCP` ones

PR #1420, opened 2022-08-25
Implement ns lookup allowlist

loganek:
The ns-lookup accepts domain names as well as suffixes, e.g.:

```
--allow-resolve=* # allow all domain names
--allow-resolve=example.com # only allow example.com name resolution
--allow-resolve=example.com --allow-resolve=*.example.com # allow example.com and its subdomains' name resolution
```

### DIFF
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -2319,12 +2319,8 @@ wasm_runtime_enlarge_memory(WASMModuleInstanceCommon *module,
 
 #if WASM_ENABLE_LIBC_WASI != 0
 
-void
-wasm_runtime_set_wasi_args_ex(WASMModuleCommon *module, const char *dir_list[],
-                              uint32 dir_count, const char *map_dir_list[],
-                              uint32 map_dir_count, const char *env_list[],
-                              uint32 env_count, char *argv[], int argc,
-                              int stdinfd, int stdoutfd, int stderrfd)
+static WASIArguments *
+get_wasi_args_from_module(wasm_module_t module)
 {
     WASIArguments *wasi_args = NULL;
 
@@ -2336,6 +2332,18 @@ wasm_runtime_set_wasi_args_ex(WASMModuleCommon *module, const char *dir_list[],
     if (module->module_type == Wasm_Module_AoT)
         wasi_args = &((AOTModule *)module)->wasi_args;
 #endif
+
+    return wasi_args;
+}
+
+void
+wasm_runtime_set_wasi_args_ex(WASMModuleCommon *module, const char *dir_list[],
+                              uint32 dir_count, const char *map_dir_list[],
+                              uint32 map_dir_count, const char *env_list[],
+                              uint32 env_count, char *argv[], int argc,
+                              int stdinfd, int stdoutfd, int stderrfd)
+{
+    WASIArguments *wasi_args = get_wasi_args_from_module(module);
 
     if (wasi_args) {
         wasi_args->dir_list = dir_list;
@@ -2367,16 +2375,7 @@ void
 wasm_runtime_set_wasi_addr_pool(wasm_module_t module, const char *addr_pool[],
                                 uint32 addr_pool_size)
 {
-    WASIArguments *wasi_args = NULL;
-
-#if WASM_ENABLE_INTERP != 0 || WASM_ENABLE_JIT != 0
-    if (module->module_type == Wasm_Module_Bytecode)
-        wasi_args = &((WASMModule *)module)->wasi_args;
-#endif
-#if WASM_ENABLE_AOT != 0
-    if (module->module_type == Wasm_Module_AoT)
-        wasi_args = &((AOTModule *)module)->wasi_args;
-#endif
+    WASIArguments *wasi_args = get_wasi_args_from_module(module);
 
     if (wasi_args) {
         wasi_args->addr_pool = addr_pool;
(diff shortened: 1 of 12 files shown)


core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -161,6 +161,31 @@ convert_errno(int error)
     return errors[error];
 }
 
+static bool
```
lum1n0us, line 164: Is it OK if assume `list` is an array of domain names and domain patterns?, like `[["*", "example", "com"], ["www", "google", "com"]]`. 

If so, which function should split the input string "*.example.com" -> ["", "" ,""]

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -2319,14 +2319,31 @@ wasm_runtime_enlarge_memory(WASMModuleInstanceCommon *module,
...
+
 void
 wasm_runtime_set_wasi_args_ex(WASMModuleCommon *module, const char *dir_list[],
                               uint32 dir_count, const char *map_dir_list[],
                               uint32 map_dir_count, const char *env_list[],
                               uint32 env_count, char *argv[], int argc,
                               int stdinfd, int stdoutfd, int stderrfd)
 {
-    WASIArguments *wasi_args = NULL;
+    WASIArguments *wasi_args = get_wasi_args_from_module(module);
```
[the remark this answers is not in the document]
loganek: true, forgot to delete it when refactored the code.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -2367,30 +2384,75 @@ void
...
+
+        if (buf)
+            wasm_runtime_free(buf);
+        if (list)
+            wasm_runtime_free(list);
+        return false;
+    }
+
+    for (i = 0; i < array_size; i++) {
+        list[i] = buf + buf_offset;
```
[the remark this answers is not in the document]
loganek: @ L2423 we allocate array of pointers, whereas `buf` stores the actual values (which are then pointed to by elements in `list` array)

doc/socket_api.md
```diff
@@ -64,6 +64,12 @@ should be announced first. Every IP address should be in CIRD notation.
 $ iwasm --addr-pool=1.2.3.4/15,2.3.4.6/16 socket_example.wasm
 ```
 
+_iwasm_ also accepts list of domain names and domain name patterns for the address resolution via an option, `--allow-resolve`, to implement the capability control. Every domain that will be resolved using `sock_addr_resolve` needs to be added to the allowlist first.
+
+```bash
+$ iwasm --allow-resolve=*.example.com
```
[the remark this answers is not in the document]
loganek: done

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
loganek: `list` is an array of all the values domain names or domain patterns (i.e. list of strings), so for example running `iwasm` with the following parameters:
```
iwasm --allow-resolve=*.example.com --allow-resolve=domain.com
```
will result with having two elements in the array: `["*.example.com", "domain.com"]`. For now we allow for very simplistic pattern matching where wildcard can be used as a prefix, so it can only appear as a first character of the string.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -161,6 +161,31 @@ convert_errno(int error)
...
+ns_lookup_list_search(char **list, const char *host)
+{
+    size_t host_len = strlen(host), suffix_len;
+
+    while (*list) {
```
lum1n0us, line 169: There might be an issue, like "access out of boundary". 
`list++` might access some context behind `list` and not NULL (pass the check `*list`).

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -2367,30 +2375,75 @@ void
...
+                  char ***list_ptr, uint64 *out_buf_size)
+{
+    uint64 buf_size = 0, total_size;
+    uint32 buf_offset = 0, i;
+    char *buf = NULL, **list = NULL;
+
+    for (i = 0; i < array_size; i++)
+        buf_size += strlen(array[i]) + 1;
+
+    // We add +1 to generate null-terminated array of strings
```
wenyongh, line 2411: Use `/* .. */` to comment the code in core/iwasm as it is C source code, `//` is C++ code style.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
loganek: The list is guaranteed to end with `NULL`, see [this code](https://github.com/bytecodealliance/wasm-micro-runtime/commit/0f7dc0a38650da77c1ad333f685274894b2edc9f#diff-bfd0d83ff0f2d993326c15717688c3abb59d6f2384044769168b4316e0b6ea0bR2430)

lum1n0us:
LGTM

core/iwasm/interpreter/wasm.h
```diff
@@ -320,6 +320,8 @@ typedef struct WASIArguments {
     /* in CIDR noation */
     const char **addr_pool;
     uint32 addr_count;
+    const char **ns_lookup_pool;
+    uint32_t ns_lookup_count;
```
wenyongh, line 324: uint32

PR #1423, opened 2022-08-26
debug-engine: allocate "tmpbuf" dynamically

yamt:
Some configurations (eg. esp32/nuttx) have limited space for BSS.
0x20000 byte buffer is huge on embedded systems.
I guess in general malloc heap tends to have less limitations.
And probably more importantly, this change makes it allocated only
when the feature (debug engine) is actually used.

### DIFF
--- a/core/iwasm/libraries/debug-engine/handler.c
+++ b/core/iwasm/libraries/debug-engine/handler.c
@@ -11,18 +11,30 @@
 #include "wasm_runtime.h"
 
 #define MAX_PACKET_SIZE (0x20000)
-static char tmpbuf[MAX_PACKET_SIZE];
+static char *tmpbuf;
 static korp_mutex tmpbuf_lock;
 
 int
 wasm_debug_handler_init()
 {
-    return os_mutex_init(&tmpbuf_lock);
+    int ret;
+    tmpbuf = wasm_runtime_malloc(MAX_PACKET_SIZE);
+    if (tmpbuf == NULL) {
+        return BHT_ERROR;
+    }
+    ret = os_mutex_init(&tmpbuf_lock);
+    if (ret != BHT_OK) {
+        wasm_runtime_free(tmpbuf);
+        tmpbuf = NULL;
+    }
+    return ret;
 }
 
 void
 wasm_debug_handler_deinit()
 {
+    wasm_runtime_free(tmpbuf);
+    tmpbuf = NULL;
     os_mutex_destroy(&tmpbuf_lock);
 }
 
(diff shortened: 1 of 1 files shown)


core/iwasm/libraries/debug-engine/handler.c
```diff
@@ -11,18 +11,29 @@
...
 int
 wasm_debug_handler_init()
 {
-    return os_mutex_init(&tmpbuf_lock);
+    tmpbuf = wasm_runtime_malloc(MAX_PACKET_SIZE);
+    if (tmpbuf == NULL) {
+        return BHT_ERROR;
+    }
+    int ret = os_mutex_init(&tmpbuf_lock);
+    if (ret == BHT_ERROR) {
```
wenyongh, line 25: Had better use `ret != BHT_SUCCESS`

core/iwasm/libraries/debug-engine/handler.c
```diff
@@ -11,18 +11,29 @@
...
 
 int
 wasm_debug_handler_init()
 {
-    return os_mutex_init(&tmpbuf_lock);
+    tmpbuf = wasm_runtime_malloc(MAX_PACKET_SIZE);
+    if (tmpbuf == NULL) {
+        return BHT_ERROR;
+    }
+    int ret = os_mutex_init(&tmpbuf_lock);
```
wenyongh, line 24: Had better define ret at the beginning of the function

core/iwasm/libraries/debug-engine/handler.c
yamt: done (BHT_OK)

core/iwasm/libraries/debug-engine/handler.c
yamt: done

core/iwasm/libraries/debug-engine/handler.c
```diff
@@ -11,18 +11,30 @@
...
-static char tmpbuf[MAX_PACKET_SIZE];
+static char *tmpbuf;
 static korp_mutex tmpbuf_lock;
 
 int
 wasm_debug_handler_init()
 {
-    return os_mutex_init(&tmpbuf_lock);
+    int ret;
+    tmpbuf = wasm_runtime_malloc(MAX_PACKET_SIZE);
```
xujuntwt95329, line 21: On some constraint devices, there is a great possibility of malloc failure for such a large space, had better add `LOG_ERROR("Allocate debug package buffer failed");`, otherwise the developer may not know what happened

yamt: done.
btw, do you (or anyone) remember why this huge buffer is allocated here?
is it somehow required by the protocol?

PR #1429, opened 2022-08-29
Enable memory leak check

wenyongh:
Report the memory leak info when building iwasm with `cmake .. -DWAMR_BUILD_GC_VERIFY=1`.


core/shared/mem-alloc/ems/ems_kfc.c
```diff
@@ -132,10 +132,6 @@ gc_destroy_with_pool(gc_handle_t handle)
         && (hmu_t *)((char *)cur + hmu_get_size(cur)) != end) {
         os_printf("Memory leak detected:\n");
         gci_dump(heap);
-#if WASM_ENABLE_SPEC_TEST != 0
```
xujuntwt95329, line 135: Can we add a config `BH_EXIT_ON_GC_VERIFY_ERROR` and exit(-1) here when it is enabled? This will be more friendly to instrumentation test framework

wenyongh: Yes, will the return value -1 be gotten by the runtime embedder?

product-mini/platforms/posix/main.c
```diff
@@ -287,6 +287,8 @@ moudle_destroyer(uint8 *buffer, uint32 size)
 #if WASM_ENABLE_GLOBAL_HEAP_POOL != 0
 #ifdef __NuttX__
 static char global_heap_buf[WASM_GLOBAL_HEAP_SIZE * BH_KB] = { 0 };
+#elif WASM_ENABLE_SPEC_TEST != 0
```
no1wudi, line 290: How about merge these three definitions into one and define a platform dependent size (WASM_GLOBAL_HEAP_SIZE) for it ?

core/shared/mem-alloc/ems/ems_kfc.c
xujuntwt95329: > Yes, will the return value -1 be gotten by the runtime embedder?

the -1 will be the exit code of the process and will be gotten by another process (instrument daemon server)

wenyongh: Add the exit(-1) in wasm_runtime_memory_destroy(), wasm_memory.c, for global heap only. As wasm app heap also uses ems gc, if it has memory leak, it will exit firstly, the global heap memory leak check will be ignored.

product-mini/platforms/posix/main.c
wenyongh: Done, I defined the WASM_GLOBAL_HEAP_SIZE with unit bytes but not KB, is it OK for Nuttx?

no1wudi: It's better to change https://github.com/bytecodealliance/wasm-micro-runtime/blob/22c235b5ec5c473579984a51e2bd1207f4570201/product-mini/platforms/nuttx/wamr.mk#L264 together to get correct heap pool size.

wenyongh: How to change, like this?
```cmake
CFLAGS += -DWASM_GLOBAL_HEAP_SIZE=$(CONFIG_INTERPRETERS_WAMR_GLOBAL_HEAP_POOL_SIZE * 1024) 
```

no1wudi: Yes, it works.

PR #1433, opened 2022-08-30
Fix bugs: No error was detected and returned after the instruction or register generation failed.

FromLiQg:
In some function, like form_and_translate_func, when the micro GEN_INSN or other insn or reg generation function failed, we ignore to detect error and return false.


lum1n0us:
Not sure this is a real bug. If a process can not allocate such small memory resource, like 10 bytes or 20 bytes, I am afraid the whole system can not respond to anything and there is nothing we can do unless rebooting. 

IIUC, to catch all failed `GEN_INSN`, we need a lot of `if` branches. It will bring a negative influence on branch prediction but resolve a theoretical issue.

FromLiQg:
> Not sure this is a real bug. If a process can not allocate such small memory resource, like 10 bytes or 20 bytes, I am afraid the whole system can not respond to anything and there is nothing we can do unless rebooting.
> 
> IIUC, to catch all failed `GEN_INSN`, we need a lot of `if` branches. It will bring a negative influence on branch prediction but resolve a theoretical issue.

@lum1n0us 
I think what you said is right.
There are some many cases like this one, and It works normally in the release version (in the upper layer of the call stack, the error may be caught), so we need to discuss whether it is necessary to solve these bh_assert failed.

PR #1435, opened 2022-08-31
Fix multi-module and some other issues

wenyongh:
Fix multi-module issue:
don't call the sub module's function with "$sub_module_name$func_name"
Fix the aot_call_function free argv1 issue
Modify some API comments in wasm_export.h
Fix the wamrc help info

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -1483,7 +1483,8 @@ aot_call_function(WASMExecEnv *exec_env, AOTFunctionInstance *function,
 
 #if (WASM_ENABLE_DUMP_CALL_STACK != 0) || (WASM_ENABLE_PERF_PROFILING != 0)
         if (!aot_alloc_frame(exec_env, function->func_index)) {
-            wasm_runtime_free(argv1);
+            if (argv1 != argv1_buf)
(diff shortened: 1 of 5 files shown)


lum1n0us:
So, we are going to remove the ability of calling functions in a module's dependencies?

wenyongh:
> So, we are going to remove the ability of calling functions in a module's dependencies?

No, it is different concept, we only remove the ability to call a module's import function of the sub module through the name of "$sub_module_name$func_name", as it is really tricky and doesn't align with the spec: by the spec we can only invoke the exported function, but obviously "$sub_module_name$func_name" isn't the name of an export function. In fact we are calling an internal imported function of a module which isn't exported, that should not be allowed.

We support the ability to call an export function, which calls the imported function of a sub module, for example, for the wasm module below:
```wast
(module
  (import "mB" "B2" (func $B2 (result i32)))
  (func $C4 (type $3)
    (result i32)
    (local $0 i32)
    call $B2
  )
)
```
We can call "C4" to invoke the function "B2" of module "mB", but cannot call the function with "$mB$B2": B2 isn't exported.

yamt (approved):

lum1n0us (approved):

yamt:
i hope no one is relying on this functionality.

PR #1440, opened 2022-09-01
Fix potential warning of printf in posix/main.c

no1wudi:


product-mini/platforms/posix/main.c
```diff
@@ -464,7 +464,8 @@ main(int argc, char *argv[])
         else if (!strncmp(argv[0], "--version", 9)) {
             uint32 major, minor, patch;
             wasm_runtime_get_version(&major, &minor, &patch);
-            printf("iwasm %u.%u.%u\n", major, minor, patch);
+            printf("iwasm %" PRIu32 ".%" PRIu32 ".%" PRIu32 "\n", major, minor,
```
wenyongh, line 467: There are some other places, shall we modify them also?
```
wamr-compiler/main.c
product-mini/platforms/windows/main.c
product-mini/platforms/linux-sgx/enclave-sample/App/App.cpp
```

no1wudi: It seems some PRIxxx macro is not available in sgx enclave ?

wenyongh: Yes, seems not supported. Let's merge this PR firstly.

PR #1448, opened 2022-09-03
ci: Cover more configurations & platform on NuttX

no1wudi:
Enable build test for:
* cortex-m0
* cortex-m4
* cortex-m7
* rv32imac
* rv64imac
* rv64gc


.github/workflows/compilation_on_nuttx.yml
```diff
@@ -43,20 +43,36 @@ jobs:
...
           "CONFIG_INTERPRETERS_WAMR=y\\nCONFIG_INTERPRETERS_WAMR_AOT=y\\nCONFIG_INTERPRETERS_WAMR_LIBC_BUILTIN=y\\n",
           "CONFIG_INTERPRETERS_WAMR=y\\nCONFIG_INTERPRETERS_WAMR_AOT=y\\n",
+          "CONFIG_INTERPRETERS_WAMR=y\\nCONFIG_INTERPRETERS_WAMR_FAST=y\\n",
+          "CONFIG_INTERPRETERS_WAMR=y\\nCONFIG_INTERPRETERS_WAMR_CLASSIC=y\\n",
         ]
 
     steps:
       - name: Install Toolchains
-        run: sudo apt install -y gcc-arm-none-eabi kconfig-frontends-nox genromfs
+        run: sudo apt install -y gcc-arm-none-eabi gcc-riscv64-unknown-elf kconfig-frontends-nox genromfs
```
wenyongh, line 75: The toolchain `gcc-arm-none-eabi` and `gcc-riscv64-unknown-elf` are installed in every build, it may take some time but they are not required in each build: for x86 target, they are not needed, for arm target, gcc-riscv64 isn't needed, for riscv target, gcc-arm isn't needed. Shall we add conditions to control that? For example, split this into three steps:
step 1: install other toolchains for every target,
step 2: install gcc-arm-none-eabi, add a condition line:
```
if ${{ matrix.nuttx_board_config == "raspberrypi-pico:nsh" || matrix.nuttx_board_config == ... }}
``` 
step 3: install gcc-riscv64-unknown-elf, add a similar condition line
//////////////
Or install toolchain with a step, and set the install packages by condition sentences, not sure whether it works.

core/iwasm/aot/arch/aot_reloc_thumb.c
```diff
@@ -97,8 +97,24 @@ static SymbolMap target_sym_map[] = {
     /* clang-format off */
     REG_COMMON_SYMBOLS
     /* compiler-rt symbols that come from compiler(e.g. gcc) */
+#if __ARM_ARCH != 6
```
wenyongh, line 100: Do these sysbols really not needed by arm 6? Will it cause relocation errors in some boards?

no1wudi: It seems that armv6m's libgcc.a don't provide these symbols, when I try to build it, I got a error "undefined reference to [these symbol]".

.github/workflows/compilation_on_nuttx.yml
no1wudi: OK, I'll try to figure it out.

PR #1453, opened 2022-09-05
debug-engine: Handle wasm_debug_instance_get_current_object_name failure

yamt:
Use an empty string instead of stack garbage.

### DIFF
--- a/core/iwasm/libraries/debug-engine/handler.c
+++ b/core/iwasm/libraries/debug-engine/handler.c
@@ -108,8 +108,11 @@ process_xfer(WASMGDBServer *server, const char *name, char *args)
         os_mutex_lock(&tmpbuf_lock);
 #if WASM_ENABLE_LIBC_WASI != 0
         char objname[128];
-        wasm_debug_instance_get_current_object_name(
-            (WASMDebugInstance *)server->thread->debug_instance, objname, 128);
+        if (!wasm_debug_instance_get_current_object_name(
+                (WASMDebugInstance *)server->thread->debug_instance, objname,
+                128)) {
+            objname[0] = 0; /* use an empty string */
+        }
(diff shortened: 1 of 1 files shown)


core/iwasm/libraries/debug-engine/handler.c
```diff
@@ -108,8 +108,11 @@ process_xfer(WASMGDBServer *server, const char *name, char *args)
         os_mutex_lock(&tmpbuf_lock);
 #if WASM_ENABLE_LIBC_WASI != 0
         char objname[128];
-        wasm_debug_instance_get_current_object_name(
-            (WASMDebugInstance *)server->thread->debug_instance, objname, 128);
+        if (!wasm_debug_instance_get_current_object_name(
+                (WASMDebugInstance *)server->thread->debug_instance, objname,
+                128)) {
+            objname[0] = 0; /* use an empty string */
```
xujuntwt95329, line 114: How about zero initialize the buffer at L110： `char objname[128] = { 0 };`

And how about set to `nobody.wasm` if get object name failed?

yamt: > How about zero initialize the buffer at L110： char objname[128] = { 0 };

i don't like it because it assumes more about the implementation of wasm_debug_instance_get_current_object_name.

> And how about set to nobody.wasm if get object name failed?

why? it looks like an ordinary name. does lldb handle it specially?

xujuntwt95329: > i don't like it because it assumes more about the implementation of wasm_debug_instance_get_current_object_name.

Ok, agree with you.

> why? it looks like an ordinary name. does lldb handle it specially?

No, lldb doesn't handle it specially. Just because we use this string for LIBC_BUILTIN mode.
Empty string also OK to me.

xujuntwt95329:
LGTM

PR #1455, opened 2022-09-05
create a new branch to update the code

septa2112:
- Optimization jit_invoke_native
- There is few code updated.
- Create a dev/fast_jit branch to update the code

### DIFF
--- a/core/iwasm/fast-jit/fe/jit_emit_function.c
+++ b/core/iwasm/fast-jit/fe/jit_emit_function.c
@@ -182,6 +182,35 @@ fail:
     return false;
 }
 
+static JitReg
+create_first_res_reg(JitCompContext *cc, const WASMType *func_type)
+{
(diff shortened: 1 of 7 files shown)


core/iwasm/fast-jit/fe/jit_emit_function.c
```diff
@@ -247,6 +350,10 @@ jit_compile_op_call(JitCompContext *cc, uint32 func_idx, bool tail_call)
 
     /* Ignore tail call currently */
     (void)tail_call;
+
+    /*free argvs */
+    jit_free(argvs);
+
     return true;
 fail:
```
wenyongh, line 358: Should also free argvs if it is not NULL.

septa2112: OK,I will check my code.

PR #1467, opened 2022-09-08
Added http downloader and multicast socket options

cimacmillan:
Added a group of socket options that I found cURL & rust stdlib uses, as well as some UDP multicast options.

https://github.com/bytecodealliance/wasm-micro-runtime/issues/1336

### DIFF
--- a/core/iwasm/libraries/lib-socket/inc/wasi_socket_ext.h
+++ b/core/iwasm/libraries/lib-socket/inc/wasi_socket_ext.h
@@ -85,6 +85,20 @@ typedef struct __wasi_addr_info_hints_t {
     uint8_t hints_enabled;
 } __wasi_addr_info_hints_t;
 
+typedef struct __wasi_linger {
+    uint8_t l_onoff;
+    uint32_t l_linger_s;
+} __wasi_linger;
+
+typedef struct __wasi_in_addr_t {
+    uint32_t s_addr;
+} __wasi_in_addr_t;
+
+typedef struct __wasi_ip_mreq_t {
+    __wasi_in_addr_t imr_multiaddr;
+    __wasi_in_addr_t imr_interface;
+} __wasi_ip_mreq_t;
+
 #ifdef __wasi__
 /**
  * Reimplement below POSIX APIs with __wasi_sock_XXX functions.
@@ -93,9 +107,22 @@ typedef struct __wasi_addr_info_hints_t {
  * <sys/socket.h>
  * <sys/types.h>
  */
-
+#define SO_REUSEADDR 2
+#define SO_SNDBUF 7
+#define SO_RCVBUF 8
+#define SO_KEEPALIVE 9
+#define SO_LINGER 13
+#define SO_REUSEPORT 15
 #define SO_RCVTIMEO 20
 #define SO_SNDTIMEO 21
+
+#define TCP_KEEPIDLE 4
+#define TCP_KEEPINTVL 5
+#define TCP_FASTOPEN_CONNECT 30
+
+#define IP_MULTICAST_LOOP 34
+#define IP_ADD_MEMBERSHIP 35
+#define IP_DROP_MEMBERSHIP 36
 
 struct addrinfo {
     int ai_flags;             /* Input flags.  */
@@ -386,7 +413,7 @@ __imported_wasi_snapshot_preview1_sock_get_reuse_port(int32_t arg0,
                    __import_name__("sock_get_reuse_port")));
 
 static inline __wasi_errno_t
-__wasi_sock_get_reuse_port(__wasi_fd_t fd, int8_t *reuse)
+__wasi_sock_get_reuse_port(__wasi_fd_t fd, uint8_t *reuse)
 {
     return (__wasi_errno_t)
         __imported_wasi_snapshot_preview1_sock_get_reuse_port((int32_t)fd,
@@ -513,15 +540,17 @@ __wasi_sock_set_reuse_port(__wasi_fd_t fd, uint8_t reuse)
  * Note: This is similar to `setsockopt` in POSIX for SO_SNDBUF
  */
 int32_t
-__imported_wasi_snapshot_preview1_sock_set_send_buf_size(int32_t arg0)
+__imported_wasi_snapshot_preview1_sock_set_send_buf_size(int32_t arg0,
+                                                         int32_t arg1)
     __attribute__((__import_module__("wasi_snapshot_preview1"),
                    __import_name__("sock_set_send_buf_size")));
 
 static inline __wasi_errno_t
-__wasi_sock_set_send_buf_size(__wasi_fd_t fd)
+__wasi_sock_set_send_buf_size(__wasi_fd_t fd, __wasi_size_t buf_len)
 {
     return (__wasi_errno_t)
-        __imported_wasi_snapshot_preview1_sock_set_send_buf_size((int32_t)fd);
+        __imported_wasi_snapshot_preview1_sock_set_send_buf_size(
+            (int32_t)fd, (int32_t)buf_len);
 }
 
 int32_t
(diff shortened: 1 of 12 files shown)


core/shared/platform/common/posix/posix_socket.c
```diff
@@ -351,6 +353,306 @@ os_socket_convert_sockaddr(struct sockaddr *addr, uint8_t *buf, size_t buflen,
...
+    return BHT_OK;
+}
+
+int
+os_socket_get_send_buf_size(bh_socket_t socket, size_t *bufsiz)
+{
+    assert(bufsiz);
+
+    socklen_t bufsiz_len = sizeof(bufsiz);
+    if (getsockopt(socket, SOL_SOCKET, SO_SNDBUF, bufsiz, &bufsiz_len) != 0) {
```
loganek, line 399: I'd recommend using exact types defined for each property ([see here](https://pubs.opengroup.org/onlinepubs/009695399/functions/getsockopt.html)). For example, for `SO_SNDBUF` the value should be `int`.

core/shared/platform/common/posix/posix_socket.c
```diff
@@ -351,6 +353,306 @@ os_socket_convert_sockaddr(struct sockaddr *addr, uint8_t *buf, size_t buflen,
...
+}
+
+int
+os_socket_get_reuse_port(bh_socket_t socket, uint8 *enabled)
+{
+    return os_socket_getbooloption(socket, SOL_SOCKET, SO_REUSEPORT, enabled);
+}
+
+int
+os_socket_set_linger(bh_socket_t socket, bh_linger *bh_linger_opts)
```
loganek, line 517: I don't see this (and few other functions here) declared in `platform_api_extension.h` Is that intentional?

core/iwasm/libraries/lib-socket/inc/wasi_socket_ext.h
```diff
@@ -85,6 +85,20 @@ typedef struct __wasi_addr_info_hints_t {
     uint8_t hints_enabled;
 } __wasi_addr_info_hints_t;
 
+typedef struct __wasi_linger {
+    uint8_t l_onoff;
+    uint32_t l_linger_s;
+} __wasi_linger;
+
+typedef struct __wasi_in_addr_t {
```
loganek, line 93: could we use `__wasi_addr_ip4_t` instead, as this is a structure that we use for representing IPv4 addresses?

core/iwasm/libraries/lib-socket/inc/wasi_socket_ext.h
```diff
@@ -85,6 +85,20 @@ typedef struct __wasi_addr_info_hints_t {
     uint8_t hints_enabled;
 } __wasi_addr_info_hints_t;
 
+typedef struct __wasi_linger {
```
loganek, line 88: given this isn't going to be a commonly used structure, I wonder if we can just split it into two values rather than defining a new structure? The same about `__wasi_ip_mreq_t`

core/iwasm/libraries/lib-socket/inc/wasi_socket_ext.h
```diff
@@ -336,7 +363,7 @@ __imported_wasi_snapshot_preview1_sock_get_reuse_port(int32_t arg0,
                    __import_name__("sock_get_reuse_port")));
 
 static inline __wasi_errno_t
-__wasi_sock_get_reuse_port(__wasi_fd_t fd, int8_t *reuse)
+__wasi_sock_get_reuse_port(__wasi_fd_t fd, uint8_t *reuse)
```
lum1n0us, line 366: is there a particular reason about from `int8_t` to `uint8_t`

core/iwasm/libraries/lib-socket/inc/wasi_socket_ext.h
```diff
@@ -530,6 +559,198 @@ __wasi_sock_set_send_timeout(__wasi_fd_t fd, uint64_t timeout_us)
...
 }
 
+int32_t
+__imported_wasi_snapshot_preview1_sock_set_keep_alive(int32_t arg0,
+                                                      int32_t arg1)
+    __attribute__((__import_module__("wasi_snapshot_preview1"),
+                   __import_name__("sock_set_keep_alive")));
+
+static inline __wasi_errno_t
+__wasi_sock_set_keep_alive(__wasi_fd_t fd, uint8_t option)
```
lum1n0us, line 569: may I ask about possible values of `option`?

core/iwasm/libraries/lib-socket/inc/wasi_socket_ext.h
cimacmillan: Only reason is that for other functions I had used `uint8_t` for the enable/disable flag and I thought I would change this for consistency with those other set/get sock opts

core/iwasm/libraries/lib-socket/inc/wasi_socket_ext.h
cimacmillan: Just 0 and 1 like a bool

wenyongh:
@loganek @cimacmillan We have created the release WAMR-1.0.0, and merged the changes from branch main into branch dev/socket, as there is a conflict in file `samples/socket-api/CMakeLists.txt`, I did it manually with command line, please don't forget to sync up your local repo with the changes, thanks.

loganek:
Sure, thanks for letting us know

core/shared/platform/include/platform_api_extension.h
```diff
@@ -555,6 +622,376 @@ os_socket_set_recv_timeout(bh_socket_t socket, uint64 timeout_us);
...
+/**
+ * Enable re-use of local addresses
+ *
+ * @param socket the socket to set
+ * @param enable 1 to enable or 0 to disable
+ *
+ * @return 0 if success, -1 otherwise
+ */
+int
+os_socket_set_reuse_addr(bh_socket_t socket, bool enable);
```
loganek, line 634: [minor] wonder I wonder if more readable would be to have `is_enabled` for both `set` and `get`?

core/shared/platform/include/platform_api_extension.h
```diff
@@ -555,6 +622,376 @@ os_socket_set_recv_timeout(bh_socket_t socket, uint64 timeout_us);
...
+
+/**
+ * Set the linger options for the given socket
+ *
+ * @param socket the socket to set
+ * @param linger_opts linger options to set
+ * @return 0 if success, -1 otherwise
+ */
+int
+os_socket_set_linger(bh_socket_t socket, int l_onoff, int l_linger_s);
```
loganek, line 677: replace `l_onoff` with `is_enabled`? Also, `l_` prefix seems not needed given this function is all about linger anyway.

core/shared/platform/include/platform_api_extension.h
```diff
@@ -555,6 +622,376 @@ os_socket_set_recv_timeout(bh_socket_t socket, uint64 timeout_us);
...
+/**
+ * Drop membership of a group
+ *
+ * @param socket the socket to set timeout
+ * @param mreq the group multicast address and local interface address
+ *
+ * @return 0 if success, -1 otherwise
+ */
+int
+os_socket_set_ip_drop_membership(bh_socket_t socket, uint32_t imr_multiaddr,
```
loganek, line 854: we have `bh_inet_network_output_t` (which was renamed to `bh_ip_addr_buffer_t` in the `loganek/sendtorecvfrom` branch) for representing an IP address (both IPv6 and IPv4) - do you think it's worth using that so we can add support for both versions?

core/shared/platform/include/platform_api_extension.h
```diff
@@ -555,6 +622,376 @@ os_socket_set_recv_timeout(bh_socket_t socket, uint64 timeout_us);
...
+/**
+ * Join IPv6 multicast group
+ *
+ * @param socket the socket to set timeout
+ * @param mreq the group multicast address and local interface address
+ *
+ * @return 0 if success, -1 otherwise
+ */
+int
+os_socket_set_ipv6_join_group(bh_socket_t socket, uint16_t imr_multiaddr[8],
```
loganek, line 954: So, referring to my comment above - we could possibly merge this (and other methods that are for IPv4 and IPv6) into one method (rather than having 2) and pass `is_ipv6` flag - unless you think it's better to keep them separate?

core/shared/platform/common/posix/posix_socket.c
```diff
@@ -351,6 +353,473 @@ os_socket_convert_sockaddr(struct sockaddr *addr, uint8_t *buf, size_t buflen,
     return BHT_OK;
 }
 
+int
```
loganek, line 356: should that be a static function or are you going to export it from the unit and use somewhere else as well?

samples/socket-api/README.md
```diff
@@ -89,10 +89,44 @@ $ ./iwasm ./socket_opts.wasm
...
 The output is:
 ```bash
-[Client] Create socket
-recv_timeout is expected
-send_timeout is expected
-[Client] Close socket
+[Client] Create TCP socket
+[Client] Create UDP socket
+[Client] Create UDP IPv6 socket
+SO_RCVTIMEO tv_sec is expected
```
loganek, line 95: Not sure if we want to keep maintaining the list? Could we just explain what the output is but not pasting it here? It gets a bit long and might actually make the documentation less readable.

loganek (approved):
Thanks for the change, overall looks good to me. I left a couple of comments. The major question I have is whether we could simplify some of the get/set functions that are different in BSD sockets and have single function with address family as an argument instead of two separate functions for every getter setter?

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
```diff
@@ -506,6 +520,129 @@ get_sol_socket_option(int sockfd, int optname, void *__restrict optval,
...
+            HANDLE_ERROR(error);
+            ((struct linger *)optval)->l_onoff = l_onoff;
+            ((struct linger *)optval)->l_linger = l_linger_s;
+            return error;
+        case SO_BROADCAST:
+            assert(*optlen == sizeof(int));
+            error = __wasi_sock_get_broadcast(sockfd, (bool *)optval);
+            HANDLE_ERROR(error);
+            return error;
+    }
```
lum1n0us, line 560: Understand that all un-handling cases will go to L562 actually. But still feel adding a `default` branch is C like.

lum1n0us: same suggestion for below functions

cimacmillan: Addressed thanks

lum1n0us:
LGTM

loganek (approved):
Thanks, looks good. @wenyongh can we merge that?

PR #1480, opened 2022-09-13
Add check for code section size, fix interp float operations

wenyongh:


core/iwasm/interpreter/wasm_mini_loader.c
```diff
@@ -5545,7 +5549,7 @@ wasm_loader_prepare_bytecode(WASMModule *module, WASMFunction *func,
                     else if (is_64bit_type(*(loader_ctx->frame_ref - 1))) {
                         loader_ctx->frame_ref -= 2;
                         loader_ctx->stack_cell_num -= 2;
-#if (WASM_ENABLE_FAST_INTERP == 0) || (WASM_ENABLE_JIT != 0)
+#if WASM_ENABLE_FAST_INTERP == 0
```
xujuntwt95329, line 5552: Why we need to remove `WASM_ENABLE_JIT != 0` here?

xujuntwt95329 (approved):
LGTM

core/iwasm/interpreter/wasm_mini_loader.c
wenyongh: The classic interpreter is enabled when LLVM JIT is enabled, no need to add the macro control again.

PR #1487, opened 2022-09-15
Fix: thread-mgr: prevent an already detached thread from being detached again

shengyun-zhou:
If WASM app has called pthread_detach() to detach a thread, it will be detached again at thread exit. Attempting to detach an already detached thread results crash in musl-libc. This patch will fix it.

### DIFF
--- a/core/iwasm/libraries/thread-mgr/thread_manager.c
+++ b/core/iwasm/libraries/thread-mgr/thread_manager.c
@@ -697,6 +697,9 @@ wasm_cluster_join_thread(WASMExecEnv *exec_env, void **ret_val)
 int32
 wasm_cluster_detach_thread(WASMExecEnv *exec_env)
 {
+    if (exec_env->thread_is_detached)
+        return 0;
+
     int32 ret = 0;
 
     os_mutex_lock(&cluster_list_lock);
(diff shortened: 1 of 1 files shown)


core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -697,6 +697,9 @@ wasm_cluster_join_thread(WASMExecEnv *exec_env, void **ret_val)
 int32
 wasm_cluster_detach_thread(WASMExecEnv *exec_env)
 {
+    if (exec_env->thread_is_detached)
+        return 0;
```
wenyongh, line 701: Should move the code into the lock/unlock code piece, e.g. :
```C
    os_mutex_lock(&cluster_list_lock);
    if (exec_env->thread_is_detached) {
        os_mutex_unlock(&cluster_list_lock);
        return 0;
    }
    if (!clusters_have_exec_env(exec_env)) {
    ...
    os_mutex_unlock(&cluster_list_lock);
```
Or if another thread changes the state when current thread runs between L701 and L705, current thread will change again.

shengyun-zhou: OK, I have moved it into lock protection.

PR #1489, opened 2022-09-15
linux-sgx: Implement SGX IPFS as POSIX backend for files interaction

JamesMenetrey:
Dear WAMR developers,

Here is a PR that contributes to integrating an Intel SGX feature called [Intel Protection File System Library (IPFS)](https://www.intel.com/content/www/us/en/developer/articles/technical/overview-of-intel-protected-file-system-library-using-software-guard-extensions.html) in the runtime to create, operate and delete files inside the enclave, while guaranteeing the confidentiality and integrity of the data persisted.

This feature has been implemented so that when the runtime is compiled with a newly-introduced flag (`WAMR_SGX_IPFS`), the files interaction API of WASI will leverage IPFS, instead of the regular POSIX OCALLs. The implementation has been written with light changes to the POSIX sandbox for WASI, so all the security aspects WAMR relies on are conserved.

In addition to this integration, the following changes have been made:
 - The CI workflow has been adapted to test the compilation of the runtime with the flag `WAMR_SGX_IPFS` set to true,
 - Introduction of a new sample that demonstrates the interaction of the files (called `file`),
 - Documentation of this new feature ([preview here](https://github.com/X-Margin-Inc/wasm-micro-runtime/blob/ipfs_pr/doc/linux_sgx.md#sgx-intel-protected-file-system))

A proof of concept regarding WAMR and IPFS was already published by my university and me a year ago with many benchmarks to evaluate the overheads of IPFS file interaction. You can see our results in [this paper](https://arxiv.org/abs/2103.15860). With the help of the company Credora, we thought it would be a great asset to WAMR to push this research into the upstream WAMR.

Feel free to discuss the changes here, as I would be glad to perform any changes that are not compliant with the best practices of this project.

Cheers


build-scripts/config_common.cmake
```diff
@@ -287,3 +287,7 @@ if (WAMR_BUILD_STACK_GUARD_SIZE GREATER 0)
     add_definitions (-DWASM_STACK_GUARD_SIZE=${WAMR_BUILD_STACK_GUARD_SIZE})
     message ("     Custom stack guard size: " ${WAMR_BUILD_STACK_GUARD_SIZE})
 endif ()
+if (WAMR_SGX_IPFS EQUAL 1)
```
wenyongh, line 290: WAMR_BUILD_SGX_IPS

build-scripts/config_common.cmake
```diff
@@ -287,3 +287,7 @@ if (WAMR_BUILD_STACK_GUARD_SIZE GREATER 0)
...
+    add_definitions (-DWAMR_SGX_IPFS=1)
```
wenyongh, line 291: Had better use `WASM_ENABLE_SGX_IPFS` to align with other macros.
And add lines in `core/config.h`:
```C
#ifndef WASM_ENABLE_SGX_IPFS
#define WASM_ENABLE_SGX_IPFS 0
#endif
```

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -22,6 +22,10 @@
 #include "rights.h"
 #include "str.h"
 
+#if WAMR_SGX_IPFS
```
wenyongh, line 25: `#if WASM_ENABLE_SGX_IPFS != 0`

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -321,6 +325,9 @@ struct fd_object {
     struct refcount refcount;
     __wasi_filetype_t type;
     int number;
+#ifdef WAMR_SGX_IPFS
+    void *sgx_file;
+#endif
```
wenyongh, line 330: @JamesMenetrey It may be not a very good option to modify this file by adding macro controls, in our design, the related system APIs should be provided by the platform layer. As you have implemented the `ipfs_xxx` APIs in sgx_ipfs.c/h, shall we encapsulate a layer to re-implement the sgx platform APIs e.g. fopen/fclose/seek/.., for example, create a map and the operations between file descriptor (fd) and the file handle of sgx ipfs:
```
typedef struct {
    void *ipfs_file;
    uint32 ipfs_fd;
    ...
} IPFSFileMap;

IPFSFileMap *ipfs_file_list;            //Or an array with max size, or a hash map
os_corp_mutex ipfs_file_list_lock; //lock for multi-thread

int ipfs_init()  // To initialize the list, and call this in bh_platform_init()
{...
}
void ipfs_destroy() // To destroy the list, and call this in bh_platform_destroy()
{...
}

void * ipfs_fd2file(int ipfs_fd)
{...
}

int ipfs_file2fd(void *ipfs_file)
{...
}
```

And provide operations to allocate file descript for new file handle,
for API read in sgx_file.c, we may rewrite it like:
```C
ssize_t
read(int fd, void *buf, size_t size)
{
    ssize_t ret;
    int size_read_max = 2048, size_read, total_size_read = 0, count, i;
    char *p = buf;

#if WASM_ENABLE_SGX_IPFS != 0
   ipfs_file = ipfs_fd2file(fd);
   if (!ipfs_file) { report error }
   operate on the file handle
#endif

    if (buf == NULL) {
        TRACE_FUNC();
        return -1;
    }
    ...
}
```

core/shared/platform/linux-sgx/sgx_ipfs.c
```diff
@@ -0,0 +1,396 @@
+/*
+ * Copyright (C) 2022 Intel Corporation.  All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+
+#if WASM_ENABLE_SGX_IPFS
```
wenyongh, line 6: In most cases, we used style like `#if WASM_ENABLE_SGX_IPFS != 0` or `#if WASM_ENABLE_SGX_IPFS == 0` to control the source code

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
JamesMenetrey: Hey, thanks for the insights! I have refactored to implement all the IPFS hooks in the SGX dedicated platform file. I opted to use a hashmap to map the file descriptors to the SGX IPFS pointers, so the complexity of pointers resolution is O(1). I did not use a mutex manually in my code, since the hashmap provides this ability by setting the right argument in the constructor.

JamesMenetrey:
Hey @wenyongh, many thanks for this great feedback. I have improved the aspects you mentioned in your previous comments/reviews. Notably, I restored the initial state of the `posix.c` file and implemented hooked the SGX functions for file interactions to introduce IPFS. As such, this leads to better encapsulation.

Let me know if you see other points to fix/refactor.

Cheers!

.github/workflows/compilation_on_sgx.yml
```diff
@@ -140,6 +140,7 @@ jobs:
             # "-DWAMR_BUILD_SIMD=1",
             "-DWAMR_BUILD_TAIL_CALL=1",
             "-DWAMR_DISABLE_HW_BOUND_CHECK=1",
+            "-DWAMR_BUILD_SGX_IPFS=1",
```
wenyongh, line 143: How about also add compiling and running the sample in the related CI files:

https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/.github/workflows/compilation_on_android_ubuntu.yml#L409-L415

https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/.github/workflows/compilation_on_macos.yml#L352-L358

https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/.github/workflows/compilation_on_sgx.yml#L359-L365

wenyongh:
> Hey @wenyongh, many thanks for this great feedback. I have improved the aspects you mentioned in your previous comments/reviews. Notably, I restored the initial state of the `posix.c` file and implemented hooked the SGX functions for file interactions to introduce IPFS. As such, this leads to better encapsulation.
> 
> Let me know if you see other points to fix/refactor.
> 
> Cheers!

@JamesMenetrey Thanks a lot for implementing this feature, it is very helpful to SGX users! The PR seems good and there are only several comments from me, could you help check them?

core/shared/platform/linux-sgx/sgx_file.c
```diff
@@ -184,6 +188,18 @@ openat(int dirfd, const char *pathname, int flags, ...)
...
         errno = get_errno();
+
+#if WASM_ENABLE_SGX_IPFS != 0
+    // When WAMR uses Intel SGX IPFS to enabled, it opens a second
+    // file descriptor to interact with the secure file.
+    // The first fild descriptor opened earlier is used to interact
+    // with the metadata of the file (e.g., time, flags, etc.).
+    void *file_ptr = ipfs_fopen(fd, pathname, flags);
+    if (file_ptr == NULL) {
+        return -1;
```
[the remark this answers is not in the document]
JamesMenetrey: Indeed, I have added a call of `ocall_close` in case `ipfs_fopen` fails.

.github/workflows/compilation_on_sgx.yml
JamesMenetrey: I've added a lightweight `iwasm` runner to the sample to execute it in the CI you linked.

core/shared/platform/linux-sgx/sgx_ipfs.c
```diff
@@ -0,0 +1,453 @@
...
+fd2file(int fd)
+{
+    return bh_hash_map_find(ipfs_file_list, (void *)(intptr_t)fd);
+}
+
+int
+ipfs_init()
+{
+    ipfs_file_list = bh_hash_map_create(32, true, (HashFunc)fd_hash,
+                                        (KeyEqualFunc)fd_equal, NULL, NULL);
```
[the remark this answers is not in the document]
JamesMenetrey: Sure, that makes sense!

JamesMenetrey:
@wenyongh,

Thanks for this second pass of review. I have completed the different points you mentioned.

Don't hesitate to check it out and give me feedback on it!

Cheers

wenyongh:
LGTM

wenyongh (approved):

xujuntwt95329 (commented):
LGTM

JamesMenetrey:
Great, thanks! 🎉🚀

PR #1493, opened 2022-09-16
Dockerfile lint errors fix

TianlongLiang:
fix Dockerfile linter errors and most warnings


lum1n0us:
@wenyongh 
I am thinking that maybe we should remove *ci/Dockerfile* and always recommand *.devcontainer/Dockerfile*.

wenyongh:
> @wenyongh I am thinking that maybe we should remove _ci/Dockerfile_ and always recommand _.devcontainer/Dockerfile_.

Yes, if possible, more docker files make more confusing and result in more maintain burden.

lum1n0us:
@wenyongh  We still lefts some software versions relatied warnings.  Should we indiciate versions of all dependencies? Or keep using latest ones?

wenyongh:
> @wenyongh We still lefts some software versions relatied warnings. Should we indiciate versions of all dependencies? Or keep using latest ones?

Not sure, shall we use the latest ones by default, and add some comments to indicate which versions to use if the os is not the latest?

lum1n0us:
@TianlongLiang  
- please remove *ci/Dockerfile*, and recommaned *vscode development container* in documents.
- please make sure that all references to *ci/Dockerfile* are handled.

TianlongLiang:
OK, I will get to it.

lum1n0us:
> > @wenyongh We still lefts some software versions relatied warnings. Should we indiciate versions of all dependencies? Or keep using latest ones?
> 
> Not sure, shall we use the latest ones by default, and add some comments to indicate which versions to use if the os is not the latest?

IIUC,
- *apt-get install <package name>* will install the package with *the latest available* version. For example, there is a chance that *apt-get* will install difference version on Ubuntu 20.04 and Ubuntu 22.04. It depends on which version the package owner uploaded. Usually it dosn't matter since *apt-get* is so popular that any package problem will be fixed ASAP.
- *pip install* is the same story

TianlongLiang:
> @TianlongLiang
> 
> * please remove _ci/Dockerfile_, and recommaned _vscode development container_ in documents.
> * please make sure that all references to _ci/Dockerfile_ are handled.

so for the shell scripts in sample/workload/* and shell script in ci/, I should change the directory of Dockerfile to the one under .devcontainer, right?

lum1n0us:
> > @TianlongLiang
> > 
> > * please remove _ci/Dockerfile_, and recommaned _vscode development container_ in documents.
> > * please make sure that all references to _ci/Dockerfile_ are handled.
> 
> so for the shell scripts in sample/workload/* and shell script in ci/, I should change the directory of Dockerfile to the one under .devcontainer, right?

YES. that is the plan. let's see how it goes.

TianlongLiang:
Now I finished modifying the shell scripts `ci/build_wamr.sh` and `samples/workload/docker/build_workload.sh` to redirect them to use `.devcontainer/Dockerfile,` and they ran successfully. And I used this Dockerfile in the vscode, which was also fined. 

Also, I tested all the other Dockerfile and can successfully build them except `product-mini/platforms/zephyr/simple/Dockerfile.` It displays the following error messages when trying to build it even on the original unmodified Dockerfile:

```shell
Step 15/18 : RUN west espressif install
 ---> Running in 0e69b119aa3a
usage: west espressif [-h] [-b BAUD] [-p PORT] [-e ELF] {monitor}
west espressif: error: argument command: invalid choice: 'install' (choose from 'monitor')
The command '/bin/sh -c west espressif install' returned a non-zero code: 2
```

I tried to fix this issue today but so far no luck, I wonder if someone is more familiar with this `west` command or this `zephyr` platform that I could ask for help with. Meanwhile, I will continue to work on trying to fix this error

PR #1494, opened 2022-09-16
fix AOT debug for macOS

yamt:
note: macOS doesn't have elf.h.

### DIFF
--- a/core/iwasm/aot/debug/elf.h
+++ b/core/iwasm/aot/debug/elf.h
@@ -1,0 +1,358 @@
+/****************************************************************************
+ * include/elf.h
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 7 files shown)


core/iwasm/aot/debug/elf.h
```diff
@@ -0,0 +1,368 @@
...
+ * include/elf.h
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
```
wenyongh, line 11: @yamt, when introducing new files from other projects, we used to (1) add a file `LICENSE` under the related folder, (2) update the `<wamr-root>/ATTRIBUTIONS.md` to avoid the license issue. Could you do the similar operations? Thanks.

yamt: done.
i included their LICENSE/NOTICE files as they are. if it's preferred to strip unrelated parts, please let me know.

wenyongh: Yes, I found that LINCESE_NUTTX is really long, but we only need the first part, right? Can we only keep the content before L202?

yamt: ok. i removed unrelated parts.

PR #1497, opened 2022-09-16
Copy only received bytes from socket recv buffer into the app buffer

cimacmillan:
**What**

* Updated `copy_buffer_to_iovec_app` so that it copies as much of the buffer into the iovec as specified
* Throw invalid value when allocating an iovec of size 0

**Why**

* A bug found from TCP client example which allocates 1024 for the iovec size (where the buf size is also 1024) but received bytes is passed in as the `buf_size` argument to `copy_buffer_to_iovec_app`. This would return early after hitting this check `buf + data->buf_len > buf_begin + buf_size`. However, if the amount to copy is less than the iovec size, we should copy that much of the buf size. Eg TCP client sample receives 27(?) bytes at a time, and this copies 27 bytes into the iovec of size 1024
* The TCP client example attempts to recv bytes of size 0, this attempts to wasm malloc size 0, which outputs a warning. We should early return if recv bytes of size 0

### DIFF
--- a/core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
+++ b/core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
@@ -1865,11 +1865,20 @@ min(size_t a, size_t b)
 
 static wasi_errno_t
 copy_buffer_to_iovec_app(wasm_module_inst_t module_inst, uint8 *buf_begin,
-                         uint32 buf_size, iovec_app_t *data, uint32 data_len)
+                         uint32 buf_size, iovec_app_t *data, uint32 data_len,
+                         uint32 size_to_copy)
 {
     uint8 *buf = buf_begin;
     uint32 i;
-    uint32 length_to_copy;
+    uint32 size_to_copy_into_iovec;
+
+    if (buf_size < size_to_copy) {
+        return __WASI_EINVAL;
+    }
+
+    if (size_to_copy == 0) {
+        return __WASI_ESUCCESS;
+    }
 
     for (i = 0; i < data_len; data++, i++) {
         char *native_addr;
@@ -1879,17 +1888,24 @@ copy_buffer_to_iovec_app(wasm_module_inst_t module_inst, uint8 *buf_begin,
         }
 
         if (buf >= buf_begin + buf_size
-            || buf + data->buf_len < buf /* integer overflow */) {
+            || buf + data->buf_len < buf /* integer overflow */
+            || buf + data->buf_len > buf_begin + buf_size
+            || size_to_copy == 0) {
             break;
         }
 
-        // If our app buffer size is smaller than the amount to be copied,
-        // only copy the amount in the app buffer
-        length_to_copy = min(data->buf_len, buf_size);
+        /**
+         * If our app buffer size is smaller than the amount to be copied,
+         * only copy the amount in the app buffer. Otherwise, we fill the iovec
+         * buffer and reduce size to copy on the next iteration
+         */
+        size_to_copy_into_iovec = min(data->buf_len, size_to_copy);
 
         native_addr = (void *)addr_app_to_native(data->buf_offset);
-        bh_memcpy_s(native_addr, length_to_copy, buf, length_to_copy);
-        buf += length_to_copy;
+        bh_memcpy_s(native_addr, size_to_copy_into_iovec, buf,
+                    size_to_copy_into_iovec);
+        buf += size_to_copy_into_iovec;
+        size_to_copy -= size_to_copy_into_iovec;
     }
 
     return __WASI_ESUCCESS;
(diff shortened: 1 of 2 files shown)


core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -1867,14 +1879,17 @@ copy_buffer_to_iovec_app(wasm_module_inst_t module_inst, uint8 *buf_begin,
...
 
         if (buf >= buf_begin + buf_size
-            || buf + data->buf_len < buf /* integer overflow */
-            || buf + data->buf_len > buf_begin + buf_size) {
+            || buf + data->buf_len < buf /* integer overflow */) {
             break;
         }
 
+        // If our app buffer size is smaller than the amount to be copied,
+        // only copy the amount in the app buffer
```
wenyongh, line 1887: Could you use the `/* .. */` to add comments? We usually use it in the C source code.

lum1n0us:
LGTM

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
wenyongh: And wandering what the PR title means? Does it mean that we copy no more than the max size of app buffer from sender, but not the size that receiver hopes to receive?

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -1852,12 +1857,19 @@ allocate_iovec_app_buffer(wasm_module_inst_t module_inst,
...
 
+static inline size_t
+min(size_t a, size_t b)
+{
+    return a > b ? b : a;
+}
+
 static wasi_errno_t
 copy_buffer_to_iovec_app(wasm_module_inst_t module_inst, uint8 *buf_begin,
                          uint32 buf_size, iovec_app_t *data, uint32 data_len)
```
loganek, line 1868: I think for clarity we should update the signature of this function:
 * `buf_size` doesn't really represent a size of the buffer pointed by `buf_begin`; it represents a number of bytes to copy (at least based on the call in @L1924
 * we should probably pass both values to the function - size of the buffer (`buf_size`) and a number of bytes to copy (e.g. `size_to_copy`). If `buf_size < size_to_copy`, I think the function should return `EINVAL`, instead of silently swallowing the problem. 

Once we do that, later on in the function we know for sure that the buffer is enough to copy all the data, but I agree we should only copy the `size_to_copy` amount for efficiency.

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -1867,14 +1879,17 @@ copy_buffer_to_iovec_app(wasm_module_inst_t module_inst, uint8 *buf_begin,
         }
 
         if (buf >= buf_begin + buf_size
-            || buf + data->buf_len < buf /* integer overflow */
-            || buf + data->buf_len > buf_begin + buf_size) {
```
loganek, line 1871: I think this condition is fine, we just need to update the caller of this function so it passes a size of the buffer as a `buf_size` parameter.

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
cimacmillan: Will address thanks. I'll rename the PR. This will the received bytes into the app buffer, although it's a larger size. Eg with the TCP client example, the app buffer is 1024 bytes. However, the received bytes are 20 or so, so we will copy 20 bytes into the app buffer rather than the full 1024 bytes.

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
cimacmillan: Will address thanks for suggestions

wenyongh:
LGTM

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -1852,12 +1857,28 @@ allocate_iovec_app_buffer(wasm_module_inst_t module_inst,
...
 {
     uint8 *buf = buf_begin;
     uint32 i;
+    uint32 size_to_copy_into_iovec;
+
+    if (buf_size < size_to_copy) {
+        return __WASI_EINVAL;
+    }
+
+    if (size_to_copy == 0) {
```
loganek, line 1879: wonder if we need this check given that we already have this check in L1893

loganek (approved):
Looks good, just left one minor comment. Thanks.

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
cimacmillan: Yeah you're right. It saves one `validate_app_addr` call but then maybe it's preferred that if we pass in an invalid app addr, we always return `__WASI_EINVAL` even if the `size_to_copy` is 0

cimacmillan:
@wenyongh Above failing check is unrelated to the change. Is it OK to merge / or can the check be retried? I don't have permissions to trigger the workflow manually

PR #1502, opened 2022-09-18
Add a clone example

lum1n0us:
Use a clone example to show
- cold start and warm start
- how to clone from wasm_module_t and bypass loading phase

It turns out that there is a strong requirement of almost everything in wasm_instantiate() except funciton related if do a clone of wasm_instance_t.

Since we tend to isolate user data in threads, it needs to malloc and initialie
  - global. WASMGlobalIntance and global data
  - memory. WAAMMemoryInstance, memory_data and heap
  - table. WASMTableInstance, table_data
  - exports. all global, memory and table

### DIFF
--- a/samples/wasm-c-api/src/clone.c
+++ b/samples/wasm-c-api/src/clone.c
@@ -1,0 +1,518 @@
+/*
+ * Copyright (C) 2019 Intel Corporation.  All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+
+#include <pthread.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+#include "wasm_c_api.h"
+
+#define WORKER_NUMBER 10
+
+/******************************* VM *******************************/
+/* Use wasm_vm_t and vm_xxx to simulate a minimal Wasm VM in Envoy */
+
+typedef struct _vm {
+    wasm_engine_t *engine;
+    wasm_store_t *store;
+    wasm_module_t *module;
+    wasm_instance_t *instance;
+    wasm_func_t **function_list;
+    wasm_memory_t *memory;
+    wasm_table_t *table;
+    wasm_extern_vec_t *exports;
+} wasm_vm_t;
+
+typedef enum _clone_level {
+    not_cloneable = 0,
+    compiled_bytecode,
+    instantiated_module
+} clone_level;
+
+typedef struct _thread_arg_t {
+    char name[32];
+    bool *ready_go_flag;
+    pthread_mutex_t *ready_go_lock;
+    pthread_cond_t *ready_go_cond;
+    const wasm_vm_t *base_vm;
+} thread_arg_t;
+
+wasm_vm_t *
+vm_new()
+{
+    wasm_vm_t *vm = NULL;
+
+    vm = malloc(sizeof(struct _vm));
+    if (!vm)
+        goto fail;
+
+    memset(vm, 0, sizeof(wasm_vm_t));
+
+    vm->engine = wasm_engine_new();
+    if (!vm->engine)
+        goto fail;
+
+    vm->store = wasm_store_new(vm->engine);
+    if (!vm->store)
+        goto fail;
+
+    return vm;
+
+fail:
+    if (vm) {
+        if (vm->engine)
+            wasm_engine_delete(vm->engine);
+
+        free(vm);
+    }
+    return NULL;
+}
+
+bool
+vm_load(wasm_vm_t *vm, const wasm_byte_vec_t *binary)
+{
+    vm->module = wasm_module_new(vm->store, binary);
+    return vm->module != NULL;
+}
+
+bool
+vm_link(wasm_vm_t *vm, wasm_extern_vec_t *imports)
+{
+    vm->instance = wasm_instance_new(vm->store, vm->module, imports, NULL);
+    if (!vm->instance)
+        goto fail;
+
+    vm->exports = malloc(sizeof(wasm_extern_vec_t));
+    if (!vm->exports)
+        goto fail;
+
+    memset(vm->exports, 0, sizeof(wasm_extern_vec_t));
+    wasm_instance_exports(vm->instance, vm->exports);
+    /* an exported memory, and two exported functions */
+    assert(vm->exports->size == 3);
+
+    /* bind memory */
+    assert(wasm_extern_kind(vm->exports->data[0]) == WASM_EXTERN_MEMORY);
+    vm->memory = wasm_extern_as_memory(vm->exports->data[0]);
+
+    vm->function_list = malloc(2 * sizeof(wasm_func_t *));
+    if (!vm->function_list)
+        goto fail;
+
+    memset(vm->function_list, 0, sizeof(2 * sizeof(wasm_func_t *)));
+
+    /* bind wasm_set_byte(...) */
+    assert(wasm_extern_kind(vm->exports->data[1]) == WASM_EXTERN_FUNC);
+    vm->function_list[0] = wasm_extern_as_func(vm->exports->data[1]);
+
+    /* bind wasm_get_byte(...) */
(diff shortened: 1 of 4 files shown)


samples/wasm-c-api/src/clone.c
```diff
@@ -0,0 +1,465 @@
...
+    pthread_mutex_t *ready_go_lock;
+    pthread_cond_t *ready_go_cond;
+    const wasm_vm_t *base_vm;
+} thread_arg_t;
+
+static void *
+checked_malloc(size_t size)
+{
+    void *p = malloc(size);
+    assert(p);
```
wenyongh, line 48: Should not assert the result of malloc, it definitely can be NULL.
Had better change `checked_malloc` to other name (e.g. `my_malloc`?), and check its return value after calling it.

samples/wasm-c-api/src/clone.c
```diff
@@ -0,0 +1,465 @@
...
+    memset(p, 0, size);
```
wenyongh, line 49: Should be
```C
    if (p)
        memset(p, 0, size);
```

samples/wasm-c-api/src/clone.c
```diff
@@ -0,0 +1,465 @@
...
+    return p;
+}
+
+wasm_vm_t *
+vm_new()
+{
+    wasm_vm_t *vm = checked_malloc(sizeof(struct _vm));
```
wenyongh, line 56: Should check the return value

samples/wasm-c-api/src/clone.c
```diff
@@ -0,0 +1,465 @@
...
+
+    vm->engine = wasm_engine_new();
+    if (!vm->engine)
+        goto fail;
```
wenyongh, line 60: Should free vm before function return

samples/wasm-c-api/src/clone.c
```diff
@@ -0,0 +1,465 @@
...
+
+    vm->store = wasm_store_new(vm->engine);
+    if (!vm->store)
+        goto fail;
```
wenyongh, line 64: Should destroy `vm->engine` and free `vm` before function return

samples/wasm-c-api/src/clone.c
```diff
@@ -0,0 +1,465 @@
...
+}
+
+bool
+vm_link(wasm_vm_t *vm, wasm_extern_vec_t *imports)
+{
+    vm->instance = wasm_instance_new(vm->store, vm->module, imports, NULL);
+    if (!vm->instance)
+        goto fail;
+
+    vm->exports = checked_malloc(sizeof(wasm_extern_vec_t));
```
wenyongh, line 85: Should check the return value

samples/wasm-c-api/src/clone.c
```diff
@@ -0,0 +1,465 @@
...
+}
+
+wasm_vm_t *
+vm_clone_from_module(const wasm_vm_t *base)
+{
+    printf("Initializing...\n");
+    wasm_vm_t *secondary = vm_new();
+
+    printf("Reuse module and bypass vm_load()...");
+    secondary->module = base->module;
```
wenyongh, line 115: Should be
```C
if (secondary)
    secondary->module = base->module;
```

samples/wasm-c-api/src/clone.c
```diff
@@ -0,0 +1,465 @@
...
+    return NULL;
+}
+
+bool
+run_code_start(wasm_vm_t **out)
+{
+    bool ret = false;
+
+    printf("Initializing...\n");
+    wasm_vm_t *vm = vm_new();
```
wenyongh, line 281: Should check the return value

samples/wasm-c-api/src/clone.c
```diff
@@ -0,0 +1,465 @@
...
+    const char *file_name = "clone.aot";
+#else
+    const char *file_name = "clone.wasm";
+#endif
+    if (!load_wasm_file_content(file_name, &binary))
+        goto release_vm;
+
+    printf("Compiling module...\n");
+    ret = vm_load(vm, &binary);
+    wasm_byte_vec_delete(&binary);
```
wenyongh, line 295: Can we destroy the binary buffer of wasm/AOT file? It might be referred after loading?

samples/wasm-c-api/src/clone.c
```diff
@@ -0,0 +1,465 @@
...
+    if (!ret)
+        goto release_vm;
+
+    printf("Creating callback...\n");
+    wasm_functype_t *callback_type =
+        wasm_functype_new_1_0(wasm_valtype_new_i32());
+    wasm_func_t *callback = wasm_func_new(vm->store, callback_type, report_cb);
```
wenyongh, line 302: Should check the return value?

samples/wasm-c-api/src/clone.c
```diff
@@ -0,0 +1,465 @@
...
+    bool ret;
+    wasm_vm_t *secondary = vm_clone(first, compiled_bytecode);
+    if (!secondary)
+        return false;
+
+    printf("Creating callback...\n");
+    wasm_functype_t *callback_type =
+        wasm_functype_new_1_0(wasm_valtype_new_i32());
+    wasm_func_t *callback =
+        wasm_func_new(secondary->store, callback_type, report_cb);
```
wenyongh, line 333: Should check the return value?

samples/wasm-c-api/src/clone.c
```diff
@@ -0,0 +1,465 @@
...
+    thread_arg_t *thrd_arg = (thread_arg_t *)arg;
+    printf("Running warm start at %s...\n", thrd_arg->name);
+
+    pthread_setspecific(name_key, thrd_arg->name);
+
+    wasm_vm_t *vm;
+    if (!run_warm_start_level_1(thrd_arg->base_vm, &vm))
+        return NULL;
+
+    pthread_mutex_trylock(thrd_arg->ready_go_lock);
```
wenyongh, line 406: pthread_mutex_trylock isn't supported in lib-pthread library, had better use pthread_mutex_lock?

samples/wasm-c-api/src/clone.c
```diff
@@ -0,0 +1,465 @@
...
+    for (size_t i = 0; i < sizeof(tids) / sizeof(tids[0]); i++) {
+        thread_arg_t *thrd_arg = thrd_args + i;
+
+        snprintf(thrd_arg->name, 32, "Worker#%lu", i);
+        thrd_arg->ready_go_cond = &ready_go_cond;
+        thrd_arg->ready_go_lock = &ready_go_lock;
+        thrd_arg->ready_go_flag = &ready_go_flag;
+        thrd_arg->base_vm = base_vm;
+
+        pthread_create(&tids[i], NULL, thrd_func, thrd_arg);
```
wenyongh, line 449: Should check the return value?

samples/wasm-c-api/src/clone.c
lum1n0us: `wasm_module_new(...)` will clone `binary` and keep it in memory.

samples/wasm-c-api/src/clone.c
lum1n0us: https://linux.die.net/man/3/pthread_mutex_trylock is in *pthread.h*. 

Yes, using `pthread_mutex_lock` is OK. 

But may I ask that is the problem mentioned above in some specific  version or specific OSes?

samples/wasm-c-api/src/clone.c
wenyongh: OK, got it

samples/wasm-c-api/src/clone.c
wenyongh: In all OSs now, we haven't implemented `pthread_mutex_trylock` in `core/iwasm/libraries/lib-pthread`, wasm app should not be able to call `pthread_mutex_trylock`. Maybe we can enable it in the future.

lum1n0us: OIC. Actually, `thrd_func(...)` is a host function. It won't be converted into a  Wasm function.

PR #1508, opened 2022-09-20
AOT: fix crash in dumping call stack when the AOT file doesn't contain custom name section

shengyun-zhou:


core/iwasm/aot/aot_runtime.c
```diff
@@ -2917,6 +2917,8 @@ static const char *
 lookup_func_name(const char **func_names, uint32 *func_indexes,
                  uint32 func_index_count, uint32 func_index)
 {
+    if (!func_names || !func_indexes || func_index_count == 0)
+        return NULL;
```
wenyongh, line 2921: We used to define the variables at the beginning of a `{ .. }` as it is C source code, could you move the check to be ahead of L2925?

shengyun-zhou: OK, I have normalized it

PR #1509, opened 2022-09-20
Merge some APIs of interp and AOT modes

wenyongh:


core/iwasm/common/wasm_memory.c
```diff
@@ -185,3 +192,446 @@ wasm_runtime_get_mem_alloc_info(mem_alloc_info_t *mem_alloc_info)
...
+bool
+wasm_runtime_validate_app_addr(WASMModuleInstanceCommon *module_inst_comm,
+                               uint32 app_offset, uint32 size)
+{
+    WASMModuleInstance *module_inst = (WASMModuleInstance *)module_inst_comm;
+    WASMMemoryInstance *memory_inst;
+
+    bh_assert(module_inst_comm->module_type == Wasm_Module_Bytecode
+              || module_inst_comm->module_type == Wasm_Module_AoT);
+
```
lum1n0us, line 205: if `WASMModuleInstance` is compatible with both `Wasm_Module_Bytecode` and `Wasm_Module_AoT`, does it mean we `WASMModuleInstanceCommon` is not necessary?

wenyongh: Maybe, but it is used in wasm_export.h, we should carefully check whether we can remove it. My suggestion is to keep it currently, and try to remove it in the future.

Issue #1418, opened 2022-08-25
Project provided and bundled Dockerfiles generate flood of errors and warnings with hadolint


cosmo0920:
## Step To Reproduce

1. Install hadolint 2.10.0
2. execute hadolint for Dockerfiles
3. Boom. :boom:

### hadolint version

```console
$ hadolint --version
Haskell Dockerfile Linter 2.10.0
```

### projects' Dockerfiles

```console
$ git ls-files --exclude="*Dockerfile*" --cached --ignored
.devcontainer/Dockerfile
ci/Dockerfile
product-mini/platforms/zephyr/simple/Dockerfile
test-tools/IoT-APP-Store-Demo/wasm_django/Dockerfile
test-tools/IoT-APP-Store-Demo/wasm_django/server/Dockerfile
test-tools/wamr-ide/WASM-Debug-Server/Docker/Dockerfile
test-tools/wamr-ide/WASM-Toolchain/Docker/Dockerfile
```

## Result

```console
$ git ls-files --exclude="*Dockerfile*" --cached --ignored | xargs hadolint
.devcontainer/Dockerfile:12 DL3027 warning: Do not use apt as it is meant to be a end-user tool, use apt-get or apt-cache instead
.devcontainer/Dockerfile:22 DL4006 warning: Set the SHELL option -o pipefail before RUN with a pipe in it. If you are using /bin/sh in an alpine image or if your shell is symlinked to busybox then consider explicitly setting your SHELL to /bin/ash, or disable this check
.devcontainer/Dockerfile:22 DL3047 info: Avoid use of wget without progress bar. Use `wget --progress=dot:giga <url>`.Or consider using `-q` or `-nv` (shorthands for `--quiet` or `--no-verbose`).
.devcontainer/Dockerfile:22 DL3027 warning: Do not use apt as it is meant to be a end-user tool, use apt-get or apt-cache instead
.devcontainer/Dockerfile:31 DL3003 warning: Use WORKDIR to switch to a directory
.devcontainer/Dockerfile:33 DL3003 warning: Use WORKDIR to switch to a directory
.devcontainer/Dockerfile:42 DL3047 info: Avoid use of wget without progress bar. Use `wget --progress=dot:giga <url>`.Or consider using `-q` or `-nv` (shorthands for `--quiet` or `--no-verbose`).
[...]

PR #1511, opened 2022-09-22
Suppress hadolint warnings for pinning versions part

cosmo0920:
Related to https://github.com/bytecodealliance/wasm-micro-runtime/issues/1418.

I suppressed hadolint warnings for pinning version.
This is because these warnings for reproducible builds.
But, for development and CIs, ordinary cases developers have to use the latest packages.
So, it can be suppressed these linting errors.


TianlongLiang:
LGTM, thanks for bringing the docker linter issue to our attention and this awesome patch

cosmo0920:
Thanks !

PR #1513, opened 2022-09-22
Minor changes to support compilation with both clang and clang++

dzobbe:


loganek:
Could you provide a bit more details to the change? Looks like we're moving a lot of code from the `wasi_socket_ext.c` to a header file.

dzobbe:
Sorry my bad. I selected the wrong file. Will close this and open a new one with the rigth .c file

PR #1526, opened 2022-09-27
Allow passing custom error buffer to wasm_instance_new_with_args

loganek:
This is helpful if the user uses different logging technique and wants to get error message back to their system

This breaks public API and I'm not sure how do we proceed with this kind of changes (assuming you're happy to accept it). Could you please provide a recommendation or point me to a process related to releasing backward-incompatible changes? Optionally we can make the last two arguments default (e.g. set `error_buf` to `NULL` and if that's the case, the `wasm_instance_new_with_args` will use internal buffer) however I find this a bit polluting the API, so would rather avoid default arguments in that case.

### DIFF
--- a/core/iwasm/common/wasm_c_api.c
+++ b/core/iwasm/common/wasm_c_api.c
@@ -4284,17 +4284,20 @@ wasm_instance_t *
 wasm_instance_new(wasm_store_t *store, const wasm_module_t *module,
                   const wasm_extern_vec_t *imports, own wasm_trap_t **traps)
 {
+    char error_buf[128] = { 0 };
+
     return wasm_instance_new_with_args(store, module, imports, traps,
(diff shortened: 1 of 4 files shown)


lum1n0us:
In my opinion, I don't suggest to change signature of any released public API. I always tend to create a new one.

But in this case, I believe we should use `wasm_trap_t` and its `wasm_message_t` to pass error message. Caller could use `wasm_trap_message` to retrieve information, and `wasm_trap_trace` to get call stack if it is provided.

yamt:
i agree it's better to use a trap, at least for a subset of errors.

loganek:
makes sense, resolving as we already have https://github.com/bytecodealliance/wasm-micro-runtime/pull/1751

PR #1529, opened 2022-09-28
fix isssues detected by coverity

lum1n0us:


samples/socket-api/wasm-src/udp_client.c
```diff
@@ -73,6 +73,7 @@ main(int argc, char *argv[])
                    (struct sockaddr *)&server_address, &serverlen);
 
     if (ret > 0) {
+        buffer[ret] = '\0';
```
wenyongh, line 76: Will ret be equal to sizeof(buffer)? If yes, buffer[ret] will cause illegal memory access.
How about change to:
```
char msg[64];
snprintf(msg, sizeof(msg), "[Client] Buffer recieved: %%.%ds\n", ret);   => construct string "%.ns", n == ret
printf(msg, buffer);
```

samples/socket-api/wasm-src/udp_server.c
```diff
@@ -79,6 +79,7 @@ main(int argc, char *argv[])
             perror("Read failed");
             goto fail;
         }
+        buffer[ret] = '\0';
```
wenyongh, line 82: Same as above

samples/socket-api/wasm-src/udp_client.c
lum1n0us: according to code `core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c`,  it * returns the number of bytes read or -1 for errors.*.

loganek:
@lum1n0us @wenyongh is it possible to run coverity as part of the PR check?

wenyongh:
> @lum1n0us @wenyongh is it possible to run coverity as part of the PR check?

Maybe, we will check it in the future. To be frankly, it takes long time to compile all the possible feature combinations and not sure whether it is suitable to make it as a CI.

PR #1539, opened 2022-09-29
Socket: Explicit narrowing type cast and add missing static keywords.

JamesMenetrey:
Dear WAMR developers,

While compiling the file `wasi_socket_ext.c` with pedantic options (typically `-Wimplicit-int-conversion`, `-Wmissing-prototypes`), some warnings are raised.

This PR addresses those warnings by adding missing `static` statements before functions and explicitly casting a narrowing conversion.

Cheers

### DIFF
--- a/core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
+++ b/core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
@@ -143,7 +143,7 @@ accept(int sockfd, struct sockaddr *addr, socklen_t *addrlen)
     error = __wasi_sock_accept(sockfd, 0, &new_sockfd);
     HANDLE_ERROR(error)
 
-    error = getpeername(new_sockfd, addr, addrlen);
+    error = (__wasi_errno_t)getpeername(new_sockfd, addr, addrlen);
     HANDLE_ERROR(error)
 
     return new_sockfd;
@@ -543,7 +543,7 @@ freeaddrinfo(struct addrinfo *res)
     free(res);
 }
 
-struct timeval
+static struct timeval
 time_us_to_timeval(uint64_t time_us)
 {
     struct timeval tv;
@@ -552,13 +552,13 @@ time_us_to_timeval(uint64_t time_us)
     return tv;
 }
 
-uint64_t
+static uint64_t
 timeval_to_time_us(struct timeval tv)
 {
     return (tv.tv_sec * 1000000UL) + tv.tv_usec;
 }
 
-int
+static int
 get_sol_socket_option(int sockfd, int optname, void *__restrict optval,
                       socklen_t *__restrict optlen)
 {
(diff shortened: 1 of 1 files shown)


core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
```diff
@@ -143,7 +143,7 @@ accept(int sockfd, struct sockaddr *addr, socklen_t *addrlen)
     error = __wasi_sock_accept(sockfd, 0, &new_sockfd);
     HANDLE_ERROR(error)
 
-    error = getpeername(new_sockfd, addr, addrlen);
```
loganek, line 146: Actually, this should be handled in a different way. `getpeername` already sets `errno` and returns `-1` on error, so I passing it later to `HANDLE_ERROR` might override the correct `errno` code. I think the correct handling should be like:
```
if (getpeername(new_sockfd, addr, addrlen) == -1) {
  return -1;
}
```

loganek (changes requested):

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
wenyongh: Yes, seems it is better to change the code like you said:
```
RETURN VALUE
       On success, zero is returned.  On error, -1 is returned, and errno is set appropriately.
```

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
```diff
@@ -143,7 +143,7 @@ accept(int sockfd, struct sockaddr *addr, socklen_t *addrlen)
...
+    error = (__wasi_errno_t)getpeername(new_sockfd, addr, addrlen);
     HANDLE_ERROR(error)
```
loganek, line 147:
```suggestion
    if (getpeername(new_sockfd, addr, addrlen) == -1) { 
        return -1; 
    }
```

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
JamesMenetrey: Indeed, nicely catch. The change has been applied to this PR.

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
```diff
@@ -143,8 +143,9 @@ accept(int sockfd, struct sockaddr *addr, socklen_t *addrlen)
     error = __wasi_sock_accept(sockfd, 0, &new_sockfd);
     HANDLE_ERROR(error)
 
-    error = getpeername(new_sockfd, addr, addrlen);
-    HANDLE_ERROR(error)
+    if (getpeername(new_sockfd, addr, addrlen) == -1) { 
+        return -1;
```
wenyongh, line 147: @JamesMenetrey There are trailing spaces at the end of the two lines, which caused CI code guideline check error, could you please fix them? Thanks.

JamesMenetrey: Just trailed them :)

loganek (approved):

PR #1541, opened 2022-09-29
Add merge commit for dev/socket

loganek:
This change only introduces a merge commit for `dev/socket` branch and does not include any file changes. I noticed that the `dev/socket` work was squashed which means we're loosing a history of the work being done (in case we remove `dev/socket` branch). I think for the future workstreams we should aim to avoid squashing commits when merging larger branch and instead one of the below (in that order):
1. do a merge with fast forward (this keeps the tree very straight)
2. rebase and merge with fast forward (this also keeps the tree very straight, although might require fixing the same conflict multiple times for more complex change)
3. merge with merge commit (this one makes the tree a bit messy so I listed it as the least preferred option).

So we always have full history. Having said that, we should ask contributors to clean-up their individual PRs before they're merged (e.g. ask them to squash some of the commits) but I don't think maintainers should do that by themselves. Worth noting that squashing commits also overrides authors of the change in git log, and even if contributors are listed in the log message, it's not obvious to find their changes.


wenyongh:
@loganek Since development branch might have many temporarily/unused commits, we usually `Squash and merge` from it to main branch. For the dev/socket, if we want to keep the commits of dev/socket for main branch, we can merge it into main again with normal mode (non-squash), had better not create another PR from our forked repo.
I created another PR to do that: https://github.com/bytecodealliance/wasm-micro-runtime/pull/1542

loganek:
@wenyongh thanks a lot. I agree that development branches might be a bit polluted and contain a lot of 'ugly' commits. I think the way we could possibly approach it is to give developers who work on those branches write access (with force push option) to those particular branches so once the feature is completed, they can polish the history. What are your thoughts on that?

lum1n0us:
maybe use `git rebase -i` to rewriting commits before PR merging? 

https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

loganek:
Yes, but ideally this should be done by the developers working on the PR, not necessarily by the maintainers (as developers probably know best what changes should go as part of one commit, and what should not). It will also reduce the maintainers being a bottleneck, as sometimes rewriting a history takes quite a bit of effort. That's why I suggested giving write access to dev branches to developers so once the change is completed they can:
1. polish the history (using `git rebase -i`)
2. rebase to latest main (`git rebase origin/main`)

so then maintainers can merge with `fast-forward` option.
What do you think?

wenyongh:
> @wenyongh thanks a lot. I agree that development branches might be a bit polluted and contain a lot of 'ugly' commits. I think the way we could possibly approach it is to give developers who work on those branches write access (with force push option) to those particular branches so once the feature is completed, they can polish the history. What are your thoughts on that?

Unfortunately, we have no privilege to change the setting. I think we can ask the developer to submit relatively high quality commits, or let him roll up until a necessary change is reached. And we will polish the commit title and comments before merging it.

loganek:
ah, I see. I thought github allows to grant per-branch permissions. I guess one other option is for developers to then pull changes into their repository, do the cleanup and make a final PR from their fork to the WAMR's `main` branch.

wenyongh:
Close this PR as similar PR #1542 had been merged.

PR #1562, opened 2022-10-07
A workflow for releasing binaries

lum1n0us:
`release_process` is only triggered manually. It will create a git tag with the semantic versioning definition in *config_common.cmake*. And generate binaries for *iwasm* and *wamrc* on variant OSes.

### DIFF
--- a/.github/scripts/fetch_and_compare_version.py
+++ b/.github/scripts/fetch_and_compare_version.py
@@ -1,0 +1,123 @@
+#!/usr/bin/env python3
+#
+# Copyright (C) 2019 Intel Corporation.  All rights reserved.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+
+import re
+import shlex
+import subprocess
+import sys
+
+
+def fetch_version_from_code():
+    """
+    search the semantic version definition in build-scripts/config_common.cmake
+    """
+    major, minor, patch = "", "", ""
+    with open("build-scripts/config_common.cmake", encoding="utf-8") as f:
+        for line in f.readlines():
+            if "WAMR_VERSION" not in line:
+                continue
+
+            major_match = re.search(r"WAMR_VERSION_MAJOR (\d+)", line)
+            if major_match is not None:
+                major = major_match.groups()[0]
+                continue
+
+            minor_match = re.search(r"WAMR_VERSION_MINOR (\d+)", line)
+            if minor_match is not None:
+                minor = minor_match.groups()[0]
+                continue
+
+            patch_match = re.search(r"WAMR_VERSION_PATCH (\d+)", line)
+            if patch_match is not None:
+                patch = patch_match.groups()[0]
+
+    if len(major) == 0 or len(minor) == 0 or len(patch) == 0:
+        raise Exception(
+            "can't find the semantic version definition likes WAMR_VERSION_*"
+        )
+    return f"WAMR-{major}.{minor}.{patch}"
+
+
+def fetch_latest_git_tag():
+    list_tag_cmd = (
+        'git tag --list WAMR-*.*.* --sort=committerdate --format="%(refname:short)"'
+    )
+    p = subprocess.run(shlex.split(list_tag_cmd), capture_output=True, check=True)
+
+    all_tags = p.stdout.decode().strip()
+    latest_tag = all_tags.split("\n")[-1]
+    return latest_tag
+
+
+def match_version_pattern(v):
+    pattern = r"WAMR-\d+\.\d+\.\d+"
+    m = re.match(pattern, v)
+    return m is not None
+
+
+def split_version_string(v):
+    """
+    return the semantic version as an integer list
+    """
+    pattern = r"WAMR-(\d+)\.(\d+)\.(\d+)"
+    m = re.match(pattern, v)
+    return [int(x) for x in m.groups()]
+
+
+def compare_version_string(v1, v2):
+    """
+    return value:
+      - 1. if v1 > v2
+      - -1. if v1 < v2
+      - 0. if v1 == v2
+    """
+    if not match_version_pattern(v1):
+        raise Exception(f"{v1} doesn't match the version pattern")
+
+    if not match_version_pattern(v2):
+        raise Exception(f"{v2} doesn't match the version pattern")
+
+    v1_sem_ver = split_version_string(v1)
+    v2_sem_ver = split_version_string(v2)
+
+    return 0 if v1_sem_ver == v2_sem_ver else (1 if v1_sem_ver > v2_sem_ver else -1)
+
+
+def is_major_or_minor_changed(v1, v2):
+    """
+    return true if change either major of v2 or minor of v2
+    return false or else
+    """
+    if not match_version_pattern(v1):
+        raise Exception(f"{v1} doesn't match the version pattern")
+
+    if not match_version_pattern(v2):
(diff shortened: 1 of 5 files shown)


lum1n0us:
[It](https://github.com/lum1n0us/wasm-micro-runtime/releases) is a test run.

It is not the full process. Only includes *wamrc* and *iwasm*.

.github/scripts/fetch_compare_version.py
```diff
@@ -0,0 +1,133 @@
...
+#
+
+import re
+import shlex
+import subprocess
+import sys
+
+
+def fetch_version_from_code():
+    command = 'egrep "set *\(WAMR_VERSION" build-scripts/config_common.cmake'
```
loganek, line 14: I wonder if we can avoid using egrep and use native python libraries/functions to get the version? That would get us one step closer to have it working on Windows.

.github/scripts/fetch_compare_version.py
```diff
@@ -0,0 +1,133 @@
...
+            return 1
+        elif v1_minor < v2_minor:
+            return -1
+        else:
+            if v1_patch > v2_patch:
+                return 1
+            elif v1_patch < v2_patch:
+                return -1
+            else:
+                return 0
```
yamt, line 96: i guess it's simpler to compare tuples.
```
>>> (1,2,1) > (1,1,2)
True
>>> 
```
https://docs.python.org/3/reference/expressions.html#value-comparisons

lum1n0us: It is a really good one

.github/workflows/_build_iwasm_release.yml
```diff
@@ -0,0 +1,89 @@
...
+
+      - name: generate iwasm binary release
+        run: |
+          cmake -S . -B build \
+            -DWAMR_BUILD_AOT=1 -DWAMR_BUILD_INTERP=1 -DWAMR_BUILD_JIT=0 \
+            -DWAMR_BUILD_CUSTOM_NAME_SECTION=0 \
+            -DWAMR_BUILD_DEBUG_INTERP=0 \
+            -DWAMR_BUILD_DEBUG_AOT=0 \
+            -DWAMR_BUILD_DUMP_CALL_STACK=0 \
+            -DWAMR_BUILD_LIBC_BUILTIN=0 \
```
wenyongh, line 44: Should be 1 for WAMR_BUILD_LIBC_BUILTIN

.github/workflows/_build_iwasm_release.yml
```diff
@@ -0,0 +1,89 @@
...
+            -DWAMR_BUILD_LIBC_UVWASI=0 \
```
wenyongh, line 45: For Windows, WAMR_BUILD_LIBC_UVWASI should be 1

.github/workflows/_build_iwasm_release.yml
```diff
@@ -0,0 +1,89 @@
...
+            -DWAMR_BUILD_SIMD=1 \
+            -DWAMR_BUILD_SHARED_MEMORY=1 \
+            -DWAMR_BUILD_TAIL_CALL=1 \
+            -DWAMR_BUILD_THREAD_MGR=1
+          cmake --build build --config Release --parallel 4
+        working-directory: ${{ inputs.cwd }}
+
+      - name: compress the binary
+        run: |
+          tar czf iwasm-${{ inputs.ver_num }}-${{ inputs.runner }}.tar.gz iwasm
```
wenyongh, line 67: An issue is that Windows might not support tart command, host about we just create `.zip` file for Windows?

.github/workflows/_build_iwasm_release.yml
lum1n0us: we still have some problems on releases for windows. There will be anther PR for windows. I am focusing on Ubuntu and MacOS in this PR.

.github/workflows/_build_iwasm_release.yml
lum1n0us: we still have some problems on releases for windows. There will be anther PR for windows. I am focusing on Ubuntu and MacOS in this PR.

wenyongh: OK, got it

.github/workflows/_build_iwasm_release.yml
wenyongh: OK, got it

.github/workflows/_build_iwasm_release.yml
wenyongh: OK, got it

.github/scripts/fetch_compare_version.py
```diff
@@ -0,0 +1,123 @@
...
+    if not match_version_pattern(v2):
+        raise Exception(f"{v2} doesn't match the version pattern")
+
+    v1_sem_ver = split_version_string(v1)
+    v2_sem_ver = split_version_string(v2)
+
+    return 0 if v1_sem_ver == v2_sem_ver else (1 if v1_sem_ver > v2_sem_ver else -1)
+
+
+def change_minor(v1, v2):
```
wenyongh, line 89: According to the code, how about changing func name to `is_major_or_minor_changed`?
`change_minor` make confusing that we are changing the minor version?

.github/workflows/_build_iwasm_release.yml
```diff
@@ -0,0 +1,89 @@
+# Copyright (C) 2019 Intel Corporation.  All rights reserved.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+name: build iwasm release
```
wenyongh, line 3: Must the file name begin with `_`?

.github/workflows/_build_wamrc.yml
```diff
@@ -0,0 +1,90 @@
+# Copyright (C) 2019 Intel Corporation.  All rights reserved.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+name: build wamrc
```
wenyongh, line 3: Must the file name begin with `_`?

.github/workflows/_build_iwasm_release.yml
lum1n0us: just a habit. not necessary. Any suggestion?

.github/workflows/_build_wamrc.yml
lum1n0us: just a habit. not necessary. Any suggestion?

.github/workflows/_build_iwasm_release.yml
wenyongh: How about removing the prefix `_` for these files? And rename `fetch_compare_version.py` to `fetch_and_compare_version.py`?

.github/scripts/fetch_and_compare_version.py
```diff
@@ -0,0 +1,123 @@
...
+import sys
+
+
+def fetch_version_from_code():
+    """
+    search the semantic version definition in build-scripts/config_common.cmake
+    """
+    major, minor, patch = "", "", ""
+    with open("build-scripts/config_common.cmake", encoding="utf-8") as f:
+        for line in f.readlines():
```
loganek, line 19: given we iterate through the whole file (and not just need a single line), this could even be simplified to:
```
for line in f:
```
See [here](https://docs.python.org/3/tutorial/inputoutput.html#methods-of-file-objects) for details.

PR #1563, opened 2022-10-08
Release wamr sdk

lum1n0us:
It depends on PR https://github.com/bytecodealliance/wasm-micro-runtime/pull/1562


.github/workflows/_build_iwasm_release.yml
```diff
@@ -0,0 +1,89 @@
...
+
+      - name: generate iwasm binary release
+        run: |
+          cmake -S . -B build \
+            -DWAMR_BUILD_AOT=1 -DWAMR_BUILD_INTERP=1 -DWAMR_BUILD_JIT=0 \
+            -DWAMR_BUILD_CUSTOM_NAME_SECTION=0 \
+            -DWAMR_BUILD_DEBUG_INTERP=0 \
+            -DWAMR_BUILD_DEBUG_AOT=0 \
+            -DWAMR_BUILD_DUMP_CALL_STACK=0 \
+            -DWAMR_BUILD_LIBC_BUILTIN=0 \
```
wenyongh, line 44: DWAMR_BUILD_LIBC_BUILTIN should be set 1 by default

.github/workflows/_build_iwasm_release.yml
```diff
@@ -0,0 +1,89 @@
...
+            -DWAMR_BUILD_LIBC_UVWASI=0 \
```
wenyongh, line 45: For Windows, WAMR_BUILD_LIBC_UVWASI should be 1, since wasmtime wasi implemented cannot be built in Windows.

And WAMR_BUILD_LIBC_WASI isn't set, had better set it to 1 explicitly.

.github/workflows/compilation_on_android_ubuntu.yml
```diff
@@ -458,6 +458,14 @@ jobs:
           cmake --build . --config Release --parallel 4
           ./hello
 
+      - name: Build Sample [simple]
+        if: ${{ matrix.light == 'green' }}
+        run: |
+          ./build.sh -p host-interp
```
wenyongh, line 464: Should `cd samples/simple` firstly?

.github/workflows/compilation_on_android_ubuntu.yml
```diff
@@ -458,6 +458,14 @@ jobs:
...
+          python3 ./sample_test_run.py $(pwd)/out
+          exit $?
+        working-directory: ./samples/simple
```
wenyongh, line 467: Should be `./out/simple -s`? `-s` is to start app manager in server mode. But a question is that it never exits?

wamr-sdk/wamr_config_ubuntu_release.cmake
```diff
@@ -0,0 +1,39 @@
...
+set (WAMR_BUILD_AOT 1)
+set (WAMR_BUILD_INTERP 1)
+set (WAMR_BUILD_JIT 0)
+
+# Runtime SDK Features
+set (WAMR_BUILD_CUSTOM_NAME_SECTION 0)
+set (WAMR_BUILD_DEBUG_INTERP 0)
+set (WAMR_BUILD_DEBUG_AOT 0)
+set (WAMR_BUILD_DUMP_CALL_STACK 0)
+set (WAMR_BUILD_LIBC_BUILTIN 0)
```
wenyongh, line 14: Should enable libc buitin by default

.github/workflows/compilation_on_android_ubuntu.yml
lum1n0us: L467 `working-directory: ./samples/simple` will make sure of it

.github/workflows/compilation_on_android_ubuntu.yml
lum1n0us: we are using a python script to handle that:
- https://github.com/lum1n0us/wasm-micro-runtime/blob/release_wamr_sdk/samples/simple/sample_test_run.py#L20 starts the app manager in server mode.
- https://github.com/lum1n0us/wasm-micro-runtime/blob/release_wamr_sdk/samples/simple/sample_test_run.py#L95 always closes the app manager

wenyongh: Good, cool

samples/simple/sample_test_run.py
```diff
@@ -0,0 +1,101 @@
...
+        time.sleep(1)
+
+        print("--> Query all installed applications...")
+        query_installed_application(args.working_directory)
+        print("--> Install a Wasm application named request_handler...")
+        install_wasm_application(
+            "request_handler",
+            "./wasm-apps/request_handler.wasm",
+            args.working_directory,
+        )
```
wenyongh, line 85: only install one wasm app, how about install more:
timer.wasm
event_publisher
event_subscriber    (subscriber can then communicate with publisher)
Then uninstall them, and install
request_handler
request_sender

PR #1575, opened 2022-10-10
wrap wasi_socket_ext api with `extern "C"`

hangedfish:
Fixed wasi_socket_ext library failed to link with cxx project.

### DIFF
--- a/core/iwasm/libraries/lib-socket/inc/wasi_socket_ext.h
+++ b/core/iwasm/libraries/lib-socket/inc/wasi_socket_ext.h
@@ -12,6 +12,10 @@
 
 /*Be a part of <wasi/api.h>*/
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 typedef enum {
     SOCKET_DGRAM = 0,
     SOCKET_STREAM,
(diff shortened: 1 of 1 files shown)


loganek (approved):

core/iwasm/libraries/lib-socket/inc/wasi_socket_ext.h
```diff
@@ -140,6 +140,9 @@ struct addrinfo {
     struct addrinfo *ai_next; /* Pointer to next in list.  */
 };
 
+#ifdef __cplusplus
+extern "C" {
+#endif
```
wenyongh, line 145: A little strange that the macro only covers part of the code, prefer to wrap the whole file:
```C
#ifndef _WASI_SOCKET_EXT_H_
#define _WASI_SOCKET_EXT_H_

#include <stdbool.h>
#include <stddef.h>
#include <stdint.h>

#ifdef __cplusplus
extern "C" {
#endif
   => leave an empty line
...
   => leave an empty line
#ifdef __cplusplus
}
#endif
   => leave an empty line
#endif /* end of _WASI_SOCKET_EXT_H_ */
```

PR #1589, opened 2022-10-13
Enable backend threads pre-compilation for orcv2 JIT

wenyongh:


build-scripts/config_common.cmake
```diff
@@ -86,9 +86,11 @@ endif ()
 
 if (WAMR_BUILD_JIT EQUAL 1)
   if (WAMR_BUILD_AOT EQUAL 1)
+    add_definitions("-DWASM_ENABLE_JIT=1")
```
lum1n0us, line 89: Just a personal opinion, it is better to operate `WAMR_BUILD_XXX` variables only in this block. And define `WASM_ENABLE_XXX` from L140.

core/iwasm/aot/aot_loader.c
```diff
@@ -2891,45 +2979,113 @@ aot_load_from_comp_data(AOTCompData *comp_data, AOTCompContext *comp_ctx,
...
-            LLVMOrcLLLazyJITLookup(comp_ctx->orc_jit, &func_addr, func_name);
-#else
-        error = LLVMOrcLLJITLookup(comp_ctx->orc_jit, &func_addr, func_name);
-#endif
-        if (error != LLVMErrorSuccess) {
-            char *err_msg = LLVMGetErrorMessage(error);
-            set_error_buf_v(error_buf, error_buf_size,
-                            "failed to compile orc jit function: %s", err_msg);
-            LLVMDisposeErrorMessage(err_msg);
+    bh_print_time("Begin to lookup jit functions");
```
lum1n0us, line 2994: how about "Start to do compilation in background"

core/iwasm/aot/aot_loader.c
```diff
@@ -2891,45 +2979,113 @@ aot_load_from_comp_data(AOTCompData *comp_data, AOTCompContext *comp_ctx,
...
+            orcjit_stop_compiling = true;
+            for (j = 0; j < i; j++) {
+                os_thread_join(orcjit_threads[j], NULL);
+            }
             goto fail3;
         }
-        module->func_ptrs[i] = (void *)func_addr;
     }
 
+#if WASM_ENABLE_LAZY_JIT == 0
```
lum1n0us, line 3075: guess we don't need `WASM_ENABLE_LAZY_JIT == 0`

core/iwasm/compilation/aot_llvm.c
```diff
@@ -1372,70 +1395,46 @@ orc_jit_create(AOTCompContext *comp_ctx)
         goto fail;
     }
 
-    /* ownership transfer
-     * LLVMOrcJITTargetMachineBuilderRef -> LLVMOrcLLJITBuilderRef
-     */
-#if WASM_ENABLE_LAZY_JIT != 0
-    LLVMOrcLLLazyJITBuilderSetNumCompileThreads(
-        builder, WASM_ORC_JIT_COMPILE_THREAD_NUM);
+    if (num_compile_threads < 1)
```
lum1n0us, line 1398: how about 
```
#if WASM_ORC_JIT_COMPILE_THREAD_NUM < 1
#error "WASM_ORC_JIT_COMPILE_THREAD_NUM has to be greater than 0"
```

core/iwasm/compilation/aot_orc_extra.cpp
```diff
@@ -149,10 +182,13 @@ LLVMOrcCreateLLLazyJIT(LLVMOrcLLLazyJITRef *Result,
 
     if (!J) {
         Result = nullptr;
-        return wrap(J.takeError());
+        return NULL;
```
lum1n0us, line 185: `LLVMErrorSuccess` is `0`. `return NULL` will confuse the caller

core/iwasm/aot/aot_loader.c
wenyongh: I did some test, it reduces port of total lookup time, had better add it to startup as quick as we can.

core/iwasm/aot/aot_loader.c
wenyongh: If it is removed, the WASM_ENABLE_LAZY_JIT isn't used, do we want to disable the eager mode?

core/iwasm/compilation/aot_llvm.c
wenyongh: Good, done.

core/iwasm/compilation/aot_orc_extra.cpp
wenyongh: done

PR #1586, opened 2022-10-12
fix threads spec test issue

xujuntwt95329:

PR #1593, opened 2022-10-13
Merge main into dev/socket

wenyongh:


wenyongh:
> Hmm, I wonder why do we keep updating `dev/socket` branch if it was already merged?

The test cases of  threads spec proposal were updated, which causes CI run failed, @xujuntwt95329 fixed the issue with #1586, we had better merge the change into the dev branches, or their CIs will run failed.

loganek:
Yes, I understand, but why do we even keep that branch? Can we remove it completely given that it was merged or we prefer to keep all the branches forever?

wenyongh:
> Yes, I understand, but why do we even keep that branch? Can we remove it completely given that it was merged or we prefer to keep all the branches forever?

OK, I supposed that you will implement more socket related features and need to keep it for some time. Shall I remove the branch now?

PR #1596, opened 2022-10-14
posix_thread.c: Avoid sem_getvalue deprecation warning on macOS

yamt:


core/shared/platform/common/posix/posix_thread.c
```diff
@@ -228,7 +228,16 @@ os_sem_post(korp_sem *sem)
 int
 os_sem_getvalue(korp_sem *sem, int *sval)
 {
+#if defined(__APPLE__)
+    /*
+     * macOS doesn't have working sem_getvalue.
+     * Mock it up here to avoid complie-time deprecation warnings.
```
wenyongh, line 234: `complie-time` to `compile-time`

yamt: oops. thank you. fixed.

Issue #1478, opened 2022-09-13
Might need to update the minimum required version of CMake


lum1n0us:
Since WAMR is locking LLVM-13 and will upgrade to LLVM-14 in the future, it is better to keep same `cmake_minimum_required` on both side.

It is [*3.13.4*](https://llvm.org/docs/CMake.html#quick-start) now.

PR #1598, opened 2022-10-14
Use cmake POSITION_INDEPENDENT_CODE instead of hardcoding -pie -fPIE

yamt:
This fixes unused option warnings on -pie for macOS. (On macOS cmake produces "-fPIE -Xlinker -pie")

Bump required cmake version to 3.14 for CheckPIESupported.

References:
https://cmake.org/cmake/help/latest/prop_tgt/POSITION_INDEPENDENT_CODE.html https://cmake.org/cmake/help/latest/module/CheckPIESupported.html#module:CheckPIESupported

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -107,8 +107,6 @@ set (WAMR_ROOT_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 
 include (${WAMR_ROOT_DIR}/build-scripts/runtime_lib.cmake)
 
-set (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,--gc-sections -pie -fPIE")
-
 set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra -Wformat -Wformat-security -Wshadow")
 # set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wconversion -Wsign-conversion")
 
--- a/product-mini/platforms/linux/CMakeLists.txt
+++ b/product-mini/platforms/linux/CMakeLists.txt
@@ -1,7 +1,9 @@
 # Copyright (C) 2019 Intel Corporation.  All rights reserved.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-cmake_minimum_required (VERSION 2.9)
+cmake_minimum_required (VERSION 3.14)
+
+include(CheckPIESupported)
 
 project (iwasm)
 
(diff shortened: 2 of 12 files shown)


CMakeLists.txt
```diff
@@ -107,8 +107,6 @@ set (WAMR_ROOT_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 
 include (${WAMR_ROOT_DIR}/build-scripts/runtime_lib.cmake)
 
-set (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,--gc-sections -pie -fPIE")
```
yamt, line 110: i removed this because there is to executable target in this file.

product-mini/platforms/linux/CMakeLists.txt
```diff
@@ -1,7 +1,9 @@
 # Copyright (C) 2019 Intel Corporation.  All rights reserved.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-cmake_minimum_required (VERSION 2.9)
+cmake_minimum_required (VERSION 3.14)
```
wenyongh, line 4: A little afraid whether it is friendly to upgrade the cmake version, maybe we can resolve the `pie -fPIE` issue by checking whether the compiler is clang, for example for the CMakeLists.txt of wamrc:
```diff
--- a/wamr-compiler/CMakeLists.txt
+++ b/wamr-compiler/CMakeLists.txt
@@ -237,7 +237,7 @@ endif()
 if (NOT MSVC)
   add_definitions(-D_FORTIFY_SOURCE=2)
   set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -ftrapv")
-  if (NOT WIN32)
+  if (NOT CMAKE_C_COMPILER MATCHES ".*clang.*" AND NOT CMAKE_C_COMPILER_ID MATCHES ".*Clang")
     set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -pie -fPIE")
   endif()
 endif()
```

@loganek, @lum1n0us, @no1wudi What is your opinions?

lum1n0us: I am ok with the change of [cmake requirement](https://github.com/bytecodealliance/wasm-micro-runtime/issues/1478).

But how about other *CMakeLists.txt*?

loganek: I'm fine with 3.14 as well, it's already 3 years old and it comes with lots of good features (compared to 2.9) so I think it's worth taking that.

yamt: rebased

PR #1599, opened 2022-10-14
Maintain Release Notes

lum1n0us:
Github Actions will extract the content from *RELEASE_NOTES.md* and use it to create the Github Release

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,0 +1,118 @@
+## WAMR-X.Y.Z
+
+### Breaking Changes
+
+### New Features
+
+### Bug Fixes
+
+### Enhancements
+
+### Others
+
+---
+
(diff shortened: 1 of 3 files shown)


RELEASE_NOTES.md
```diff
@@ -0,0 +1,118 @@
+## WAMR-X.Y.Z
+
+### Breaking Changes
+
+### New Features
+
+### Bug Fixes
+
+### Performance Improvements
```
wenyongh, line 9: How about changing to `Enhancements`? It might include performance, footprint and some feature enhancements.

RELEASE_NOTES.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Other Changes
```
wenyongh, line 11: How about changing into `Others`?

PR #1602, opened 2022-10-16
add ios support for product-mini

hangedfish:
Signed-off-by: HangedFish <bravohangedman@outlook.com>

generate `xcodeproj`, that could build iwasm shared library for iOS or iOS Simulator


product-mini/platforms/ios/generate_xcodeproj.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
```
wenyongh, line 2: Should add the license header

product-mini/platforms/ios/generate_xcodeproj.sh
```diff
@@ -0,0 +1,7 @@
...
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```
wenyongh, line 3: Should also add `# Copyright (C) 2019 Intel Corporation. All rights reserved.` like sentence as the first line

PR #1608, opened 2022-10-17
Remove unnecessary app heap memory clean operations to reduce process RSS

shengyun-zhou:
With hardware boundary checking enabled, the app heap memory comes from `os_mmap()`. Clearing the whole heap in the memory allocator causes process RSS to reach maximum app heap size immediately and wastes lots of memory, so we should remove unnecessary memory clean operations in the memory allocator.


wenyongh:
LGTM

PR #1622, opened 2022-10-19
Build with GCC11

casaroli:
I am building this as part of NuttX with GCC11 and I needed these minor changes for GCC not to get angry.

### DIFF
--- a/core/iwasm/aot/arch/aot_reloc_xtensa.c
+++ b/core/iwasm/aot/arch/aot_reloc_xtensa.c
@@ -263,6 +263,12 @@ apply_relocation(AOTModule *module, uint8 *target_section_addr,
             imm16 = (int16)(relative_offset >> 2);
 
             /* write back the imm16 to the l32r instruction */
+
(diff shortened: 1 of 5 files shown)


core/iwasm/aot/arch/aot_reloc_xtensa.c
```diff
@@ -263,10 +263,17 @@ apply_relocation(AOTModule *module, uint8 *target_section_addr,
             imm16 = (int16)(relative_offset >> 2);
 
             /* write back the imm16 to the l32r instruction */
+#if __GNUC__ >= 9
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Waddress-of-packed-member"
```
wenyongh, line 268: @casaroli Could you add comments about the code? It is not easy to understand what the flag means here.

casaroli: added on fa2363ee05e3ae65a039f5bba2e12b599c84b2f3

PR #1627, opened 2022-10-21
Move indirect optimaztion to last of pipeline

no1wudi:
The general optimization may create some intrinsic function call like llvm.memset, so let's remove these function call at last.

### DIFF
--- a/core/iwasm/compilation/aot_compiler.c
+++ b/core/iwasm/compilation/aot_compiler.c
@@ -2715,7 +2715,7 @@ aot_compile_wasm(AOTCompContext *comp_ctx)
 
         /* Run specific passes for AOT indirect mode in last since general
            optimization may create some intrinsic function call like
-           llvm.memset, so let's remove these function call here.*/
+           llvm.memset, so let's remove these function calls here. */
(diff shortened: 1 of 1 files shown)


core/iwasm/compilation/aot_compiler.c
```diff
@@ -2721,6 +2712,17 @@ aot_compile_wasm(AOTCompContext *comp_ctx)
            possible core dump. */
         bh_print_time("Begin to run llvm optimization passes");
         aot_apply_llvm_new_pass_manager(comp_ctx, comp_ctx->module);
+
+        /* Run specific passes for AOT indirect mode in last since general
+           optimization may create some intrinsic function call like
```
wenyongh, line 2717: `function calls`

core/iwasm/compilation/aot_compiler.c
```diff
@@ -2721,6 +2712,17 @@ aot_compile_wasm(AOTCompContext *comp_ctx)
...
+           llvm.memset, so let's remove these function call here.*/
```
wenyongh, line 2718: call -> calls
`here.*/` -> `here. */`

PR #1628, opened 2022-10-21
Normalize how the global heap size is defined across iwasm apps.

JamesMenetrey:
Hello,

When creating some benchmarks for WAMR, I realized I needed a way to modify the global heap size provided in `core.h`. Nonetheless, this definition is not always picked up, depending on which *iwasm* application is used. For many of them, some sizes are hardcoded and cannot be easily changed except by editing the source code file directly.

Therefore, I propose a PR that unifies the global heap size in a single CMake configuration variable `WAMR_BUILD_GLOBAL_HEAP_SIZE`, that the developers may override when executing `cmake`. This variable is taken into account by all the *iwasm* implementations and adequately documented.

As a result, the developers who build *iwasm* can easily change the size of the global heap regardless of the *iwasm* implementation, without manually finding and patching the right location for that value.

As always, I'm open to any feedback regarding how it has been implemented.

Cheers!


JamesMenetrey:
Some updates for this PR: I added the flag `WAMR_BUILD_GLOBAL_HEAP_POOL` globally, to reflect the internal flag `WASM_ENABLE_GLOBAL_HEAP_POOL` that is used in some iwasm implementations.

Harmonizing this flag makes it convenient to compile every iwasm app with the global heap pool enabled, and with a given size. As such, we can now compile iwasm apps with these additional flags:

```bash
cmake .. -DWAMR_BUILD_GLOBAL_HEAP_POOL=1 -DWAMR_BUILD_GLOBAL_HEAP_SIZE=$((200 * 1024 * 1024))
make
```

tests/wamr-test-suites/test_wamr.sh
```diff
@@ -486,7 +486,7 @@ function build_iwasm_with_cfg()
         cd ${WAMR_DIR}/product-mini/platforms/linux-sgx \
         && if [ -d build ]; then rm -rf build/*; else mkdir build; fi \
         && cd build \
-        && cmake $* .. \
+        && cmake $* .. -DWAMR_BUILD_GLOBAL_HEAP_POOL=1 \
```
wenyongh, line 489: Seems no need to change here, I have a test, sgx spec cases can run successfully without global heap.

doc/build_wamr.md
```diff
@@ -114,6 +114,19 @@ Currently we only profile the memory consumption of module, module_instance and
 
 > The function name searching sequence is the same with dump call stack feature.
 
+#### **Enable the global heap**
+- **WAMR_BUILD_GLOBAL_HEAP_POOL**=1/0, default to disable if not set for all *iwasm* applications, except for the platforms Alios and Zephyr.
+
+> **WAMR_BUILD_GLOBAL_HEAP_POOL** is used in the *iwasm* applications provided in the directory `product-mini`. When writing your own host application using WAMR, you must set the initialization argument `mem_alloc_type` to allocate and use a global heap.
```
wenyongh, line 120: A little confusing about `you must set the initialization argument `mem_alloc_type` to allocate and use a global heap`, will it make developer think that he **must** set argument to `Alloc_With_Pool`?
How about changing to something like:
```md
When writing your own host application using WAMR, if you want to use a global heap and allocate memory from it, you must set the initialization argument `mem_alloc_type` to `Alloc_With_Pool`.
```

wenyongh:
@lum1n0us Could you help review this PR? And check whether we need to modify the makefile of nuttx?

product-mini/platforms/nuttx/wamr.mk
```diff
@@ -274,7 +274,11 @@ endif
 
 ifeq ($(CONFIG_INTERPRETERS_WAMR_GLOBAL_HEAP_POOL),y)
 CFLAGS += -DWASM_ENABLE_GLOBAL_HEAP_POOL=1
+ifdef WAMR_BUILD_GLOBAL_HEAP_SIZE
```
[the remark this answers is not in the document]
JamesMenetrey: If that change is not introduced, using `WAMR_BUILD_GLOBAL_HEAP_SIZE` won't set `WASM_GLOBAL_HEAP_SIZE` accordingly. The variable `CONFIG_INTERPRETERS_WAMR_GLOBAL_HEAP_POOL_SIZE` seems to be set outside of WAMR.

Do you think to another way to enforce the usage of `WAMR_BUILD_GLOBAL_HEAP_SIZE` for nuttx?

product-mini/platforms/linux-sgx/enclave-sample/Makefile
```diff
@@ -235,6 +240,7 @@ Enclave/Enclave_t.o: Enclave/Enclave_t.c
 	@echo "CC   <=  $<"
 
 Enclave/%.o: Enclave/%.cpp
+	@echo "CXX  $(Enclave_Cpp_Flags) <=  $<"
```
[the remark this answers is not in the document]
JamesMenetrey: My bad, I used this output because of an issue I had. It can be safely removed.

doc/build_wamr.md
JamesMenetrey: Much better!

doc/build_wamr.md
```diff
@@ -114,6 +114,19 @@ Currently we only profile the memory consumption of module, module_instance and
...
+> The global heap is defined in the documentation [Memory model and memory usage tunning](memory_tune.md).
+
+#### **Set the global heap size**
+- **WAMR_BUILD_GLOBAL_HEAP_SIZE**=n, default to 10 MB (10485760) if not set for all *iwasm* applications, except for the platforms Alios (256 kB), Riot (256 kB) and Zephyr (128 kB).
+
+> **WAMR_BUILD_GLOBAL_HEAP_SIZE** is used in the *iwasm* applications provided in the directory `product-mini`. When writing your own host application using WAMR, you must set the initialization argument `mem_alloc_option` to allocate and use a global heap.
```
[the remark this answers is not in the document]
JamesMenetrey: I have changed the text based on your previous suggestion, with some adaptations regarding the size of the heap pool.

JamesMenetrey:
Hello @wenyongh,

Thanks for your feedback! I have applied the changes related to not using the global heap pool for the SPEC test in SGX, but it seems the unit tests failed. Have you an idea to make it works? It seems you indicated to have made it works on your side.

I'm waiting on the feedback of @lum1n0us regarding the Makefile for the platform nuttx. :)

Thanks!

wenyongh:
> @lum1n0us Could you help review this PR? And check whether we need to modify the makefile of nuttx?

Sorry, should add @no1wudi to help review the PR.

wenyongh:
> Hello @wenyongh,
> 
> Thanks for your feedback! I have applied the changes related to not using the global heap pool for the SPEC test in SGX, but it seems the unit tests failed. Have you an idea to make it works? It seems you indicated to have made it works on your side.
> 
> I'm waiting on the feedback of @lum1n0us regarding the Makefile for the platform nuttx. :)
> 
> Thanks!

@JamesMenetrey, for the SGX spec case test issue, it is due to memory allocation failed, how about enlarging `HeapMaxSize` in Enclave.config.xml?
```diff
--- a/product-mini/platforms/linux-sgx/enclave-sample/Enclave/Enclave.config.xml
+++ b/product-mini/platforms/linux-sgx/enclave-sample/Enclave/Enclave.config.xml
@@ -3,7 +3,7 @@
   <ProdID>0</ProdID>
   <ISVSVN>0</ISVSVN>
   <StackMaxSize>0x100000</StackMaxSize>
-  <HeapMaxSize>0x2000000</HeapMaxSize>
+  <HeapMaxSize>0x8000000</HeapMaxSize>
```
Changing from 32MB to 128MB works for `./test_wamr.sh -s spec -t aot -S -x -p -P`

JamesMenetrey:
I have increased the heap max size as requested and the SPEC tests are passing now. :)

product-mini/platforms/nuttx/wamr.mk
no1wudi, line 277: It doesn't make sense since all variables in NuttX are from Kconfig and with a 'CONFIG_' prefix,
and in NuttX, we are already make the globl heap size configurable.

So we can revert the change in wamr.mk.

no1wudi (changes requested):

JamesMenetrey:
Hey @no1wudi, thanks for your review. I have removed the changes to nuttx.

PR #1633, opened 2022-10-24
Implement Fast JIT dump call stack and perf profiling

wenyongh:
Implement dump call stack and perf profiling features for Fast JIT,
and refine some code.

### DIFF
--- a/core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
+++ b/core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
@@ -5823,13 +5823,20 @@ lower_callnative(JitCompContext *cc, x86::Assembler &a, JitInsn *insn)
     a.call(regs_i64[REG_RAX_IDX]);
 
     if (ret_reg) {
-        bh_assert((jit_reg_kind(ret_reg) == JIT_REG_KIND_I32
-                   && jit_reg_no(ret_reg) == REG_EAX_IDX)
-                  || (jit_reg_kind(ret_reg) == JIT_REG_KIND_I64
-                      && jit_reg_no(ret_reg) == REG_RAX_IDX)
-                  || ((jit_reg_kind(ret_reg) == JIT_REG_KIND_F32
-                       || jit_reg_kind(ret_reg) == JIT_REG_KIND_F64)
-                      && jit_reg_no(ret_reg) == 0));
+        if (jit_reg_kind(ret_reg) == JIT_REG_KIND_I64) {
+            /* mov res, rax */
+            mov_r_to_r_i64(a, jit_reg_no(ret_reg), REG_RAX_IDX);
+        }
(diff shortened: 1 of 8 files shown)


core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
```diff
@@ -5823,13 +5823,23 @@ lower_callnative(JitCompContext *cc, x86::Assembler &a, JitInsn *insn)
...
-                  || (jit_reg_kind(ret_reg) == JIT_REG_KIND_I64
-                      && jit_reg_no(ret_reg) == REG_RAX_IDX)
-                  || ((jit_reg_kind(ret_reg) == JIT_REG_KIND_F32
-                       || jit_reg_kind(ret_reg) == JIT_REG_KIND_F64)
-                      && jit_reg_no(ret_reg) == 0));
+        uint32 ret_reg_no = jit_reg_no(ret_reg);
+        if (jit_reg_kind(ret_reg) == JIT_REG_KIND_I64) {
+            CHECK_I64_REG_NO(ret_reg_no);
+            /* mov res, rax */
+            mov_r_to_r_i64(a, ret_reg_no, REG_RAX_IDX);
```
lum1n0us, line 5830: based on the original `bh_assert`, `ret_reg_no` is equal to `REG_RAX_IDX`. so, why mov again ?

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
```diff
@@ -5823,13 +5823,23 @@ lower_callnative(JitCompContext *cc, x86::Assembler &a, JitInsn *insn)
...
+        }
+        else if (jit_reg_kind(ret_reg) == JIT_REG_KIND_F64) {
+            CHECK_F64_REG_NO(ret_reg_no);
+            /* mov res, xmm0_f64 */
+            mov_r_to_r_f64(a, ret_reg_no, 0);
```
lum1n0us, line 5835: same as above

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
wenyongh: It is an issue of emit_callnative when the return type is i64 or f64, in which rax/xmm0_f64 is used, and the IRs generated are like:
```
callnative rax, native_func, args
mov res, rax
```
But as eax is caller register, if it is used before/after callnative, register allocator will generate the spill out/in instructions for it, then the IRs may be like:
```
spill out eax
callnative rax, native_func, args
spill in eax
mov res, rax
```
Here the `spill in eax` will restore the value of eax, and overwrite the value of rax, and cause `mov res, rax` move unexpected value into res.

Here move the handling of i64/f64 ret value of callnative into codegen to resolve the issue.

PR #1635, opened 2022-10-24
Added macOS LLDB binary release

cimacmillan:
### Summary

* Added a release stage that builds LLDB for macOS. 
* Init the copied LLVM source as a git repo, as found an issue where the patch wasn't being applied
* Made some changes small changes to LLDB patch that fixed compatibility with latest xcode clang

Tested on x86_64 & ARM macbook pros

### DIFF
--- a/.github/workflows/build_wamr_lldb.yml
+++ b/.github/workflows/build_wamr_lldb.yml
@@ -51,10 +51,23 @@ jobs:
             ./core/deps/llvm-project/build/share
             ./core/deps/llvm-project/lldb/tools/
             ./core/deps/llvm-project/inst/
-          key: ${{ inputs.runner }}-lldb_build
+          key: ${{inputs.arch}}-${{ inputs.runner }}-lldb_build
 
-      - name: intall utils
-        if: steps.lldb_build_cache.outputs.cache-hit != 'true'
+      - name: setup xcode macos
+        if: steps.lldb_build_cache.outputs.cache-hit != 'true' && contains(inputs.runner, 'macos')
+        uses: maxim-lobanov/setup-xcode@v1
+        with:
+          xcode-version: latest-stable
+
+      # Remove xCode command line tools, to prevent duplicate symbol compilation failures
+      - name: install utils macos
+        if: steps.lldb_build_cache.outputs.cache-hit != 'true' && contains(inputs.runner, 'macos')
+        run: |
+          brew install swig cmake ninja libedit
+          sudo rm -rf /Library/Developer/CommandLineTools
+
+      - name: intsall utils ubuntu
+        if: steps.lldb_build_cache.outputs.cache-hit != 'true' && contains(inputs.runner, 'ubuntu')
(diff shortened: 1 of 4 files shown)


.github/workflows/build_wamr_lldb.yml
```diff
@@ -93,18 +109,53 @@ jobs:
...
+        if: steps.lldb_build_cache.outputs.cache-hit != 'true' && contains(inputs.runner, 'ubuntu')
+        run: |
+          cp build/lib/liblldb*.so inst/lib
+          cp build/lib/liblldb*.so.* inst/lib
+        working-directory: core/deps/llvm-project
+
+      - name: pack macos specific libraries
+        if: steps.lldb_build_cache.outputs.cache-hit != 'true' && contains(inputs.runner, 'macos')
+        run: |
+          cp build/lib/liblldb* inst/lib
```
loganek, line 156: what files are being copied here? Wonder if we can make this pattern more specific (e.g. `build/lib/liblldb*.dylib`) or do we need all the files?

.github/workflows/build_wamr_lldb.yml
```diff
@@ -53,29 +53,45 @@ jobs:
...
+          xcode-version: latest-stable
+
+      - name: install utils macos
+        if: steps.lldb_build_cache.outputs.cache-hit != 'true' && contains(inputs.runner, 'macos')
+        run: |
+          brew install swig cmake ninja libedit
+          /usr/bin/xcodebuild -version
+          ls /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/
+          clang -v
+          sudo rm -rf /Library/Developer/CommandLineTools
```
loganek, line 69: could you add a comment why do we do that?

.github/workflows/build_wamr_lldb.yml
```diff
@@ -93,18 +107,53 @@ jobs:
...
           mkdir -p inst/bin
           mkdir -p inst/lib
           cp build/bin/lldb* inst/bin
-          cp build/lib/liblldb*.so inst/lib
-          cp build/lib/liblldb*.so.* inst/lib
           cp lldb/tools/lldb-vscode/package.json inst
           cp -r lldb/tools/lldb-vscode/syntaxes/ inst
         working-directory: core/deps/llvm-project
 
+      - name: pack ubuntu specific libraries
```
lum1n0us, line 144: how about  `cp build/lib/liblldb*.* inst/lib`, and merge below two steps?

cimacmillan: I've made the macOS copy a little stricter since first revision as there are a bunch of static libs added to this folder, which don't need to be included https://pastebin.com/tpbjTXyt

lum1n0us: Got your point.

lum1n0us (approved):
LGTM

PR #1636, opened 2022-10-24
XIP adaptation for xtensa platform

dongsheng28849455:
1, Reading vram buffer with 4 bytes alignment
2, Eliminate the reference for below ones:
   i32_div_u, f32_min, f32_max, f32_ceil, f32_floor, f32_trunc, f32_rint


core/shared/utils/bh_common.c
```diff
@@ -5,6 +5,96 @@
...
+    else {
+        if (src + n > pa) {
+            read_size = src + n - pa;
+        }
+    }
+    if((pre_size + read_size + suf_size) != n) {
+        return -1;
+    }
+
+    // copy pre segment
```
wenyongh, line 71: Use `/* .. */` to comment the code

core/shared/utils/bh_common.c
```diff
@@ -5,6 +5,96 @@
...
+    unsigned int buff;
+    char* pbuff = (char*)&buff;
+    buff = (*(unsigned int*)p_pre);
+    bh_memcpy_s(dest, pre_size, pbuff + pre_offset,
+                pre_size);
+
+    // copy segment
```
wenyongh, line 78: Same as above

core/shared/utils/bh_common.h
```diff
@@ -19,6 +19,13 @@ extern "C" {
         bh_assert(_ret == 0);                                         \
     } while (0)
 
+#define bh_memcpy_aw(dest, dlen, src, slen)                           \
```
wenyongh, line 22: Not sure what `aw` means? Could you add comments?

core/iwasm/aot/aot_loader.c
```diff
@@ -123,6 +123,28 @@ GET_U64_FROM_ADDR(uint32 *addr)
...
 
+#if defined(BUILD_TARGET_XTENSA)
+
+static inline uint8
+GET_U8_FROM_ADDR(const uint8* p)
+{
+    bh_assert(p);
+    uint8 res = 0;
+
+    bh_memcpy_aw(&res, sizeof(uint8), p, sizeof(uint8));
```
wenyongh, line 134: Calling `bh_memcpy_raw` might be a little slow, how about
```C
const uint8 *p_aligned = align_ptr(p + 3, 4) - 4;
uint32 buf32 = *(const uint32 *)p_aligned;
const uint8 *buf = (const uint8 *)&buf32;
return buf[p - p_aligned];
```

core/iwasm/aot/aot_loader.c
```diff
@@ -123,6 +123,28 @@ GET_U64_FROM_ADDR(uint32 *addr)
     return u.val;
 }
 
+#if defined(BUILD_TARGET_XTENSA)
```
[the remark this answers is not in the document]
dongsheng28849455: using VRAM has two preconditions, one is xtensa target, another is xip mode. xip is a runtime variant read from file which need 4 bytes read
then we make VRAM enable/disable as compliling time not runtime? 
Or, we read xip mode using 4 bytes align anyway, and after that change the reading mode according to the XIP status?

core/iwasm/aot/aot_loader.c
```diff
@@ -123,6 +123,28 @@ GET_U64_FROM_ADDR(uint32 *addr)
...
+    return res;
+}
+
+static inline uint16
+GET_U16_FROM_ADDR(const uint8 *p)
+{
+    bh_assert(p);
+    uint16 res;
+
+    bh_memcpy_aw((uint8 *)&res, sizeof(uint16), p, sizeof(uint16));
```
wenyongh, line 144: How about
```C
const uint8 *p_aligned = align_ptr(p + 3, 4) - 4;
uint32 buf32 = *(const uint32 *)p_aligned;
const uint8 *buf = (const uint8 *)&buf32;
if (is_little_endian())
    return (buf[p - p_aligned + 1] << 1) | buf[p - p_aligned];
else
    return (buf[p - p_aligned] << 1) | buf[p - p_aligned + 1];
```

core/shared/utils/bh_common.h
dongsheng28849455: aw - Align Word

core/iwasm/aot/aot_loader.c
dongsheng28849455: can it work also for string or array bytes?

core/iwasm/aot/aot_loader.c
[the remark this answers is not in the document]
wenyongh: > using VRAM has two preconditions, one is xtensa target, another is xip mode. xip is a runtime variant read from file which need 4 bytes read then we make VRAM enable/disable as compliling time not runtime? 

Yes, had better make VRAM enable/disable as compiling time behavior, since for xtensa, not all situations require VRAM, it is only used in special situation. I believe there should be VRAM related macro defined, if not, you can define it in your system.

wenyongh:
@no1wudi, Could you help review the PR? It makes changes to AOT XIP behaviors.

core/iwasm/aot/aot_loader.c
wenyongh: Not sure, maybe you can optimize the algorithm of bh_memcpy_aw:
copy the leading 1-3 unaligned bytes if there are
copy 4-byte aligned bytes
copy trailing 1-3 unaligned bytes if there are

dongsheng28849455: OK, I'd like the bh_memcpy_aw can work for any length target.

no1wudi (changes requested):

core/iwasm/aot/aot_loader.c
dongsheng28849455: > 
For "const uint8 *buf" , the intent is "const uint16 *buf"?  and the operator for shift lengh should be 8, not 1?
And shall we need judge is_little_endian? copy the bytes as they are in src will follow what it is.

core/shared/utils/bh_common.c
```diff
@@ -5,6 +5,96 @@
...
+    }
+
+    if(s1max < n) {
+        return -1;
+    }
+
+    char *pa = align_ptr(src, sizeof(unsigned int));
+    char *pb = align_ptr((src + n), sizeof(unsigned int)) - sizeof(unsigned int);
+
+    char *p_pre = pa, *p_suf =pa, *p_read=pa;
```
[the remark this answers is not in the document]
dongsheng28849455: Why? why not put it near where it is used?

core/iwasm/aot/aot_loader.c
```diff
@@ -231,7 +293,11 @@ const_str_set_insert(const uint8 *str, int32 len, AOTModule *module,
         return NULL;
     }
 
-    bh_memcpy_s(c_str, (uint32)(len + 1), str, (uint32)len);
+    if(is_vram_align_word) {
+        bh_memcpy_aw(c_str, (uint32)(len + 1), str, (uint32)len);
+    } else {
+        bh_memcpy_s(c_str, (uint32)(len + 1), str, (uint32)len);
+    }
```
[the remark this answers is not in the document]
dongsheng28849455: since the is_vram_align_word is judged already by the macro read_string when it is defined:
`#define read_string(p, p_end, str)                                \
    do {                                                          \
        if (!(str = load_string((uint8 **)&p, p_end, module,      \
                                is_load_from_file_buf, **true**,      \
                                error_buf, error_buf_size)))      \
            goto fail;                                            \
    } while (0)`
and read_string has been wrappered by "defined(BUILD_TARGET_XTENSA) && defined(...)", shall we need re-wrap again?

core/iwasm/aot/aot_loader.c
wenyongh: > For "const uint8 *buf" , the intent is "const uint16 *buf"? and the operator for shift lengh should be 8, not 1?  And shall we need judge is_little_endian? copy the bytes as they are in src will follow what it is.

Hi, I just give an example, you can implement it according to your system. And yes, my mistake, the shift length should be 8. Maybe you can:
```C
const uint8 *p_aligned = align_ptr(p + 3, 4) - 4;
uint32 buf32 = *(const uint32 *)p_aligned;
const uint8 *buf = (const uint8 *)&buf32;

res = *(uint16 *)(buf + p - p_aligned);
or
if (is_little_endian())
    return (buf[p - p_aligned + 8] << 8) | buf[p - p_aligned];
else
    return (buf[p - p_aligned] << 8) | buf[p - p_aligned + 8];
```

core/iwasm/aot/aot_loader.c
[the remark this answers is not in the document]
wenyongh: Yes, had better wrap again here, if not, the bh_memcpy_aw will be linked, which introduces the footprint.

core/iwasm/aot/aot_loader.c
```diff
@@ -123,6 +123,42 @@ GET_U64_FROM_ADDR(uint32 *addr)
     return u.val;
 }
 
+#if defined(BUILD_TARGET_XTENSA) && (WASM_ENABLE_AOT_LOAD_FROM_VRAM != 0)
```
wenyongh, line 126: It seems that the feature is only related to VRAM, it doesn't matter with XTENSA, some other targets/platforms may also need the feature, my understanding is that here (and many other places) we can just use WASM_ENABLE_AOT_LOAD_FROM_VRAM to control the code:
```C
#if WASM_ENABLE_AOT_LOAD_FROM_VRAM != 0
...
#endif
```

core/iwasm/aot/aot_loader.c
```diff
@@ -140,10 +180,40 @@ GET_U64_FROM_ADDR(uint32 *addr)
...
+#define read_string(p, p_end, str)                                \
+    do {                                                          \
+        if (!(str = load_string((uint8 **)&p, p_end, module,      \
+                                is_load_from_file_buf, true,      \
+                                error_buf, error_buf_size)))      \
+            goto fail;                                            \
+    } while (0)
+
+#else /* else of defined(BUILD_TARGET_XTENSA)           \
+         && (WASM_ENABLE_AOT_LOAD_FROM_VRAM != 0) */
```
wenyongh, line 199: /* else of  WASM_ENABLE_AOT_LOAD_FROM_VRAM != 0 */

core/iwasm/aot/aot_loader.c
```diff
@@ -155,11 +225,19 @@ GET_U64_FROM_ADDR(uint32 *addr)
...
     do {                                                          \
         if (!(str = load_string((uint8 **)&p, p_end, module,      \
-                                is_load_from_file_buf, error_buf, \
-                                error_buf_size)))                 \
+                                is_load_from_file_buf,            \
+                                error_buf, error_buf_size)))      \
             goto fail;                                            \
     } while (0)
 
+#endif /* end of defined(BUILD_TARGET_XTENSA)           \
```
wenyongh, line 233: /* end of WASM_ENABLE_AOT_LOAD_FROM_VRAM != 0 */

core/iwasm/aot/aot_loader.c
```diff
@@ -211,6 +289,10 @@ loader_malloc(uint64 size, char *error_buf, uint32 error_buf_size)
 
 static char *
 const_str_set_insert(const uint8 *str, int32 len, AOTModule *module,
+#if defined(BUILD_TARGET_XTENSA) && (WASM_ENABLE_AOT_LOAD_FROM_VRAM != 0)
+                     bool is_vram_align_word,
+#endif /* end of defined(BUILD_TARGET_XTENSA) &&  \
+          (WASM_ENABLE_AOT_LOAD_FROM_VRAM != 0) */
```
wenyongh, line 295: No need to add comments if the `#else/#endif` is very close to `#if`, and the code is easy to read.

core/iwasm/aot/aot_loader.c
```diff
@@ -211,6 +289,10 @@ loader_malloc(uint64 size, char *error_buf, uint32 error_buf_size)
 
 static char *
 const_str_set_insert(const uint8 *str, int32 len, AOTModule *module,
+#if defined(BUILD_TARGET_XTENSA) && (WASM_ENABLE_AOT_LOAD_FROM_VRAM != 0)
```
wenyongh, line 292: `#if WASM_ENABLE_AOT_LOAD_FROM_VRAM != 0`

core/iwasm/aot/aot_loader.c
```diff
@@ -230,8 +312,16 @@ const_str_set_insert(const uint8 *str, int32 len, AOTModule *module,
     if (!(c_str = loader_malloc((uint32)len + 1, error_buf, error_buf_size))) {
         return NULL;
     }
+#if defined(BUILD_TARGET_XTENSA) && (WASM_ENABLE_AOT_LOAD_FROM_VRAM != 0)
```
wenyongh, line 315: `#if WASM_ENABLE_AOT_LOAD_FROM_VRAM != 0`

core/iwasm/aot/aot_loader.c
```diff
@@ -230,8 +312,16 @@ const_str_set_insert(const uint8 *str, int32 len, AOTModule *module,
...
+    if(is_vram_align_word) {
+        bh_memcpy_aw(c_str, (uint32)(len + 1), str, (uint32)len);
+    } else
+#endif /* end of defined(BUILD_TARGET_XTENSA) &&  \
+          (WASM_ENABLE_AOT_LOAD_FROM_VRAM !=0) */
```
wenyongh, line 320: Same as above

core/shared/utils/bh_common.c
```diff
@@ -5,6 +5,65 @@
...
+
+    char *pa = align_ptr(src, 4);
+    char *pb = align_ptr((src + n), 4);
+
+    if (pa > src) {
+        pa -= 4;
+    }
+
+    for (unsigned int *p = (unsigned int *)pa; p < (unsigned int *)pb; p++) {
+        unsigned int buff = *(p);
```
wenyongh, line 34: Don't define the variables in the for sentences, define them at the beginning of the function or `{ ... }`.

core/shared/utils/bh_common.c
```diff
@@ -5,6 +5,65 @@
...
+        const char *p_byte_read = ((char *)&buff);
+
+        if ((char *)p <= src) {
+            for (char *leading = src; leading < ((char *)p + 4); leading++) {
```
wenyongh, line 38: Same as abvoe

core/shared/utils/bh_common.c
```diff
@@ -5,6 +5,65 @@
...
+                if (leading >= src + n) {
+                    break;
+                }
+                p_byte_read = ((char *)&buff) + (leading - (char *)p);
+                *dest++ = *p_byte_read;
+            }
+        }
+        else if ((char *)p >= pb - 4) {
+            for (char *traling = (char *)p; traling < src + n; traling++) {
```
wenyongh, line 47: Same as above

core/shared/utils/bh_common.c
```diff
@@ -5,6 +5,65 @@
...
+                *dest++ = *p_byte_read++;
+            }
+        }
+        else {
+            if ((char *)p + 4 >= src + n) {
```
wenyongh, line 52: Same as above

wenyongh:
@dongsheng28849455 Could you please resolve the coding format issues with `clang-format-12`?

core/iwasm/aot/aot_loader.c
```diff
@@ -264,6 +354,15 @@ load_string(uint8 **p_buf, const uint8 *buf_end, AOTModule *module,
     if (str_len == 0) {
         str = "";
     }
+#if (WASM_ENABLE_WORD_ALIGN_READ != 0)
+    else if (is_vram_word_align) {
+        if (!(str = const_str_set_insert((uint8 *)p, str_len, module,
```
wenyongh, line 359: Does it mean that we will create a new string for the vram mode, no matter the original is terminated with '\0' or not? Is it due to we can not access a string in the vram like normal mode?

wenyongh (commented):
LGTM

PR #1640, opened 2022-10-26
Add locks to protect releasing multiple stores

lum1n0us:


core/iwasm/common/wasm_c_api.c
```diff
@@ -322,6 +331,10 @@ wasm_engine_new_internal(mem_alloc_type_t type, const MemAllocOption *opts)
     /* create wasm_store_vec_t */
     INIT_VEC(engine->stores, wasm_store_vec_new_uninitialized, 1);
 
+    if (os_mutex_init(&engine->lock) != BHT_OK)
```
wenyongh, line 334: When it is destroyed, cannot find `os_mutex_destroy`

core/iwasm/common/wasm_c_api.c
```diff
@@ -1930,6 +1944,9 @@ wasm_module_new(wasm_store_t *store, const wasm_byte_vec_t *binary)
         goto failed;
     }
 
+    os_mutex_init(&module_ex->lock);
```
wenyongh, line 1947: Same as above, cannot find os_mutex_destroy for it

core/iwasm/common/wasm_c_api.c
```diff
@@ -2,6 +2,7 @@
  * Copyright (C) 2019 Intel Corporation.  All rights reserved.
  * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
  */
+#include "platform_api_vmcore.h"
```
wenyongh, line 5: Don't include `platform_api_vmcore.h`, include "bh_platform.h" instead

core/iwasm/common/wasm_c_api.c
```diff
@@ -362,10 +375,19 @@ wasm_engine_new_with_args(mem_alloc_type_t type, const MemAllocOption *opts)
...
+    os_mutex_lock(&engine->lock);
+    engine->ref_count--;
+    if (engine->ref_count > 0) {
+        goto unlock;
     }
+
+    wasm_engine_delete_internal(engine);
+    singleton_engine = NULL;
+unlock:
+    os_mutex_unlock(&engine->lock);
```
wenyongh, line 390: My understanding is that engine->lock should be also destroyed when engine is destroyed. And there is only one engine `singleton_engine`, wasm_engine_new/delete should be called only once, why we need to add lock for them?

wenyongh: Seems that we create the engine->lock, and use it for the operations of stores in engine?

samples/wasm-c-api/src/utils/read_proc_status.c
```diff
@@ -0,0 +1,66 @@
...
+}
+
+int
+read_proc_status(pid_t pid, bool(filter)(const char *), char *buf,
+                 size_t buf_cap)
+{
+    int ret = -1;
+
+    char path[128] = { 0 };
+    snprintf(path, sizeof(path) / sizeof(path[0]), "/proc/%d/status", pid);
```
yamt, line 28: is it appropriate to use linux-specific feature here?

core/iwasm/common/wasm_c_api.c
```diff
@@ -261,7 +261,17 @@ static void
...
+        if (engine->stores) {
+            /* do lazy release.
+             * all stores have been deleted, just need to
+             * handle the vec */
+            bh_vector_destroy((Vector *)engine->stores);
+            wasm_runtime_free(engine->stores);
+            engine->stores = NULL;
+        }
+
+        os_mutex_destroy(&engine->lock);
```
wenyongh, line 273: May destroy an uninitialized lock if L333, L335 got failed.

core/iwasm/common/wasm_c_api.c
```diff
@@ -335,8 +348,16 @@ wasm_engine_new()
         singleton_engine =
             wasm_engine_new_internal(Alloc_With_System_Allocator, NULL);
     }
-    if (singleton_engine)
+    else {
+        os_mutex_lock(&singleton_engine->lock);
+
+        if (singleton_engine->ref_count == 0)
```
yamt, line 354: i don't understand what this check is for.

isn't it simpler to use a global lock to protect singleton_engine?

core/iwasm/common/wasm_c_api.c
```diff
@@ -414,34 +454,24 @@ wasm_store_new(wasm_engine_t *engine)
...
-                             NULL);
-            break;
-        }
-    }
+    /*
+     * every thread has on store. One store can't be shared in multile threads
+     * so a store can not be deleted twice
+     */
+    bh_assert(store->deleted == false);
+    store->deleted = true;
```
wenyongh, line 467: store is freed in L477, setting its field here is meaningless.

core/iwasm/common/wasm_c_api.c
```diff
@@ -414,34 +454,24 @@ wasm_store_new(wasm_engine_t *engine)
...
 
     DEINIT_VEC(store->modules, wasm_module_vec_delete);
     DEINIT_VEC(store->instances, wasm_instance_vec_delete);
     if (store->foreigns) {
         bh_vector_destroy(store->foreigns);
         wasm_runtime_free(store->foreigns);
+        store->foreigns = NULL;
```
wenyongh, line 474: store is freed in L477, setting its field here is meaningless.

core/iwasm/common/wasm_c_api.c
```diff
@@ -1855,6 +1885,8 @@ wasm_foreign_delete(wasm_foreign_t *foreign)
 struct wasm_module_ex_t {
     struct WASMModuleCommon *module_comm_rt;
     wasm_byte_vec_t *binary;
+    korp_mutex lock;
```
wenyongh, line 1888: Had better add a field `lock_inited`

core/iwasm/common/wasm_c_api.c
```diff
@@ -1930,6 +1962,9 @@ wasm_module_new(wasm_store_t *store, const wasm_byte_vec_t *binary)
         goto failed;
     }
 
+    os_mutex_init(&module_ex->lock);
```
wenyongh, line 1965: Should check the return value

core/iwasm/common/wasm_c_api.c
```diff
@@ -2440,6 +2488,38 @@ wasm_module_deserialize(wasm_store_t *store, const wasm_byte_vec_t *binary)
...
+    if (!store || !module)
+        return NULL;
+
+    module_ex = module_to_module_ext(module);
+
+    os_mutex_lock(&module_ex->lock);
+
+    if (module_ex->ref_count == 0) {
+        os_mutex_unlock(&module_ex->lock);
+        return new_module;
```
wenyongh, line 2506: Had better return NULL? It makes code easier to understand

core/iwasm/common/wasm_c_api.c
```diff
@@ -2440,6 +2488,38 @@ wasm_module_deserialize(wasm_store_t *store, const wasm_byte_vec_t *binary)
...
+    }
+
+    module_ex->ref_count++;
+
+    os_mutex_unlock(&module_ex->lock);
+
+    /* add it to a watching list in store */
+    if (!bh_vector_append((Vector *)store->modules, &module_ex)) {
+        wasm_module_delete_internal(module);
+        return new_module;
```
wenyongh, line 2516: Had better return NULL?
And L2509 is invalid for such case? How about move L2509-L2511 after current line:
```C
    /* add it to a watching list in store */
    if (!bh_vector_append((Vector *)store->modules, &module_ex)) {
        wasm_module_delete_internal(module);
        os_mutex_unlock(&module_ex->lock);
        return new_module;
    }

    module_ex->ref_count++;
    os_mutex_unlock(&module_ex->lock);
```

core/iwasm/common/wasm_c_api_internal.h
```diff
@@ -27,13 +27,15 @@ WASM_DECLARE_VEC(store, *)
 struct wasm_engine_t {
     /* support one store for now */
     wasm_store_vec_t *stores;
-    uint32_t ref_count;
+    korp_mutex lock;
```
wenyongh, line 30: Had better add a field `lock_inited`

core/iwasm/common/wasm_c_api.c
wenyongh: Yes, using a global lock would be a better way, but not all platforms have the global lock.

PR #1645, opened 2022-10-26
linux-sgx: use non-destructive modes for opening files using SGX IPFS.

JamesMenetrey:
Hello,

I'm benchmarking some applications using SGX IPFS, and I realized an issue when a file is opened with a _destructive mode_ (i.e., a mode that truncates the target file). Indeed, the following operations are performed when a file is opened with SGX IPFS:

 1. the WASI layer opens a file descriptor on the file (the standard fd),
 2. the SGX IPFS API opens a file descriptor on the same file (the IPFS fd).

While there is no issue with opening modes that do not truncate files, when a file is opened with the modes "w" or "w+", the two steps truncate the same file. As a result, some POSIX calls done on the standard file descriptor works on the truncated (no longer existing) file. This is typically the case with `fstat`, which returns some attributes of a file. When used on the file descriptor of the first step, the values returned are never updated (e.g., the file size).

The fix is to honour the original opening mode of the first step but uses an alternate (non-destructive) opening mode for the second step. This PR substitutes the destructive modes (w and w+) with non-destructive alternatives (r+). As a result, the truncate operation is still honoured by the first step, and the file descriptor of the first step remains valid afterwards.

I provided additional assertions in the file sample to validate this use case (typically, these assertions fail in the `main` branch).

Cheers

### DIFF
--- a/core/shared/platform/linux-sgx/sgx_ipfs.c
+++ b/core/shared/platform/linux-sgx/sgx_ipfs.c
@@ -260,22 +260,27 @@ ipfs_fopen(int fd, const char *filename, int flags)
     bool write_only = (flags & O_ACCMODE) == O_WRONLY;
     bool read_write = (flags & O_ACCMODE) == O_RDWR;
 
-    // The mapping of the mode are described in the table in the official
+    // The mapping of the mode is similar to the table in the official
     // specifications:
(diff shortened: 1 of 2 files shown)


JamesMenetrey:
It seems that the failing tests are not related to the codebase but rather to the CI.

PR #1648, opened 2022-10-27
Add CIs to release new version and publish binary files

wenyongh:
Add CIs to enable the release process of a new version of WAMR,
and build and publish the binary files when a version is released,
including iwasm, wamrc, lldb, vscode-extension and wamr-ide for
Ubuntu-20.04, Ubuntu-22.04 and MacOS.

And refine the CIs to test spec cases.


lum1n0us:
LGTM

PR #1657, opened 2022-10-29
Remove an improper assumption when creating wasm_trap

lum1n0us:
it is allowed to have mutltiple stores in an engine and instances in a store. Let a wasm_function_t pass its wasm_store_t is more efficient than searching in all.

### DIFF
--- a/core/iwasm/common/wasm_c_api.c
+++ b/core/iwasm/common/wasm_c_api.c
@@ -1637,11 +1637,11 @@ wasm_frame_func_offset(const wasm_frame_t *frame)
 }
 
 static wasm_trap_t *
-wasm_trap_new_internal(WASMModuleInstanceCommon *inst_comm_rt,
-                       const char *default_error_info)
+wasm_trap_new_internal(wasm_store_t *store,
+                       WASMModuleInstanceCommon *inst_comm_rt,
+                       const char *error_info)
 {
     wasm_trap_t *trap;
-    const char *error_info = NULL;
     wasm_instance_vec_t *instances;
     wasm_instance_t *frame_instance = NULL;
     uint32 i;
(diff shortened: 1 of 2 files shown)


core/iwasm/common/wasm_c_api.c
```diff
@@ -1651,39 +1651,23 @@ wasm_trap_new_internal(WASMModuleInstanceCommon *inst_comm_rt,
...
-    wasm_name_new_from_string_nt(trap->message, error_info);
-    if (strlen(error_info) && !trap->message->data) {
-        goto failed;
+        wasm_name_new_from_string_nt(trap->message, error_info);
+        if (!trap->message->data) {
+            goto failed;
+        }
     }
 
+    /* fill in frames */
```
wenyongh, line 1670: Can also consider merging L1671 to L683, currently the offset of field `frames` in WASMModuleInstance and AOTModuleInstance is the same.

wenyongh (commented):
LGTM

core/iwasm/common/wasm_c_api.c
```diff
@@ -1651,72 +1651,44 @@ wasm_trap_new_internal(WASMModuleInstanceCommon *inst_comm_rt,
...
-    }
 
-    for (i = 0; i < trap->frames->num_elems; i++) {
-        (((wasm_frame_t *)trap->frames->data) + i)->instance = frame_instance;
+        for (i = 0; i < trap->frames->num_elems; i++) {
+            (((wasm_frame_t *)trap->frames->data) + i)->instance =
+                frame_instance;
+        }
     }
+#endif /* WASM_ENABLE_DUMP_CALL_STACK != 0 */
```
wenyongh, line 1691: In the original code, this line is in L1699 (old code), and normally it should be move to L1673 (new code), now the macro wraps more codes, is it OK?

PR #1660, opened 2022-10-30
Add wasm_module_obtain() to clone another from one wasm_module_t

lum1n0us:
It depends on #1656

### DIFF
--- a/core/iwasm/common/wasm_c_api.c
+++ b/core/iwasm/common/wasm_c_api.c
@@ -17,10 +17,24 @@
 #endif /*WASM_ENABLE_JIT != 0 && WASM_ENABLE_LAZY_JIT == 0*/
 #endif /*WASM_ENABLE_AOT != 0*/
 
+/*
+ * Thread Model:
+ * - Only one wasm_engine_t in one process
+ * - One wasm_store_t is only accessed by one thread. wasm_store_t can't be
+ * shared in threads
+ * - wasm_module_t can be shared in threads
+ * - wasm_instance_t can not be shared in threads
+ */
+
 #define ASSERT_NOT_IMPLEMENTED() bh_assert(!"not implemented")
 #define UNREACHABLE() bh_assert(!"unreachable")
 
-typedef struct wasm_module_ex_t wasm_module_ex_t;
+typedef struct wasm_module_ex_t {
+    struct WASMModuleCommon *module_comm_rt;
+    wasm_byte_vec_t *binary;
+    korp_mutex lock;
+    uint32 ref_count;
+} wasm_module_ex_t;
 
 static void
 wasm_module_delete_internal(wasm_module_t *);
@@ -261,7 +275,24 @@ static void
 wasm_engine_delete_internal(wasm_engine_t *engine)
 {
     if (engine) {
-        DEINIT_VEC(engine->stores, wasm_store_vec_delete);
+        if (engine->stores) {
+            bh_vector_destroy((Vector *)engine->stores);
+            wasm_runtime_free(engine->stores);
+        }
+
+        /* clean all created wasm_module_t and their locks */
+        {
+            unsigned i;
+            for (i = 0; i < engine->modules.num_elems; i++) {
+                wasm_module_ex_t *module;
+                if (bh_vector_get(&engine->modules, i, &module)) {
+                    os_mutex_destroy(&module->lock);
+                    wasm_runtime_free(module);
+                }
+            }
+            bh_vector_destroy(&engine->modules);
+        }
+
         wasm_runtime_free(engine);
     }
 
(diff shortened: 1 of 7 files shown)


core/iwasm/common/wasm_c_api.c
```diff
@@ -2004,13 +2014,33 @@ wasm_module_delete_internal(wasm_module_t *module)
     }
 
     module_ex = module_to_module_ext(module);
+
+    /* deleting the module... */
+    if (module_ex->ref_count == 0) {
```
wenyongh, line 2019: Should move the check after locking the `module_ex->lock`? Suppose here the module_ex->ref_count read is 1 and we continue to run. And another thread decreases the count to 0 before this thread runs into lock L2024, then in L2028, after decreasing in L2027, the count is now -1 (or UINT32_MAX), the code will continue to run into L2033 to 2044, which is unexpected behavior?

core/iwasm/common/wasm_c_api.c
```diff
@@ -2472,6 +2502,34 @@ wasm_module_deserialize(wasm_store_t *store, const wasm_byte_vec_t *binary)
...
+{
+    wasm_module_ex_t *module_ex = NULL;
+
+    if (!store || !module)
+        return NULL;
+
+    module_ex = module_to_module_ext(module);
+
+    /* deleting the module... */
+    if (module_ex->ref_count == 0) {
```
wenyongh, line 2516: Same as above, should move the check after L2521, and if `fef_count == 0`, unlock the lock and return.

core/iwasm/common/wasm_c_api.c
lum1n0us: I thought there are cases when hitting L2024 : `os_mutex_lock(&module_ex->lock);`
- `module_ex` has been released. Dereference it is UB. It depends on callers' behavior to avoid it.
- `module_ex->lock` has been destroyed. It leads a failed return code of `os_muext_lock`.  Guess it is the "unexpected behavior` mentioned above?
- `module_ex->ref_count` is zero. In this case `ref_count--` will be `UINT32_MAX` and `if >0` will let `return` work
- `module_ex->ref_count` is great than zero.

yamt:
i guess it's better to introduce wasm_shared_module_t for compatibility. how do you think?

lum1n0us:
Slightly hesitate about the purpose of `wasm_shared_xxx_t`. `wasm_shared_xxx_t` should represent a core concept in Wasm spec.  Therefore didn't plan to support  `wasm_shared_module_t` until figure out.

yamt:
i'm talking about wasm_shared_module_t specifically, which seems to be used by other implementations.
not wasm_shared_xxx_t in general.

it even appears in an example in wasm-c-api repo:
https://github.com/WebAssembly/wasm-c-api/blob/b6dd1fb658a282c64b029867845bc50ae59e1497/example/threads.c#L25
and mentioned in the doc:
https://github.com/WebAssembly/wasm-c-api/blob/main/README.md#interfaces

core/iwasm/common/wasm_c_api.c
```diff
@@ -1939,34 +1950,62 @@ wasm_module_new(wasm_store_t *store, const wasm_byte_vec_t *binary)
...
+
+    module_ex->lock = malloc_internal(sizeof(korp_mutex));
+    if (module_ex->lock == NULL)
+        goto remove_last;
+
+    if (os_mutex_init(module_ex->lock) != BHT_OK)
+        goto free_lock;
 
+    if (!bh_vector_append(&singleton_engine->module_locks, &module_ex->lock))
+        goto free_lock;
```
wenyongh, line 1989: Should `goto destroy_lock` and add a destroy_lock label:
```C
destroy_lock:
    os_mutex_destroy(module_ex->lock);
```

core/iwasm/common/wasm_c_api.c
```diff
@@ -2472,6 +2523,34 @@ wasm_module_deserialize(wasm_store_t *store, const wasm_byte_vec_t *binary)
...
+        return NULL;
+
+    module_ex = module_to_module_ext(module);
+
+    os_mutex_lock(module_ex->lock);
+
+    /* deleting the module... */
+    if (module_ex->ref_count == 0) {
+        LOG_WARNING("re-enter a module under deleting.");
+        return NULL;
```
wenyongh, line 2541: Should unlock the lock before return.

core/iwasm/common/wasm_c_api.c
```diff
@@ -2472,6 +2523,34 @@ wasm_module_deserialize(wasm_store_t *store, const wasm_byte_vec_t *binary)
...
+
+    if (!store || !module)
+        return NULL;
+
+    module_ex = module_to_module_ext(module);
+
+    os_mutex_lock(module_ex->lock);
+
+    /* deleting the module... */
+    if (module_ex->ref_count == 0) {
```
wenyongh, line 2539: A question is that the module_ex might be freed at L2065, here accessing module_ex->lock/ref_count causes invalid memory dereferencing.

My understanding is that we can store the <module, lock> pair as element of the engine->module_locks, which is vector or hashmap, and create the vector or hashmap with use_lock = true. And when access module, we can search the vector/hashmap to get the lock, when deleting module, we can destroy the <module, lock> element in the vectory/hashmap.

samples/wasm-c-api/src/threads.c
```diff
@@ -0,0 +1,162 @@
+#include <inttypes.h>
```
wenyongh, line 1: Should add the license header

wenyongh (commented):
LGTM

PR #1671, opened 2022-11-02
Avoid initialize LLVM repeatedly

wenyongh:
Currently we initialize and destroy LLVM environment in aot_create_comp_context
and aot_destroy_comp_context, which are called in wasm_module_load/unload,
and the latter may be invoked multiple times, which leads to duplicated LLVM
initialization/destroy and may result in unexpected behaviors.

Move the LLVM init/destroy into runtime init/destroy to resolve the issue.

### DIFF
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -30,6 +30,9 @@
 #if WASM_ENABLE_FAST_JIT != 0
 #include "../fast-jit/jit_compiler.h"
 #endif
+#if WASM_ENABLE_JIT != 0 || WASM_ENABLE_WAMR_COMPILER != 0
+#include "../compilation/aot_llvm.h"
+#endif
 #include "../common/wasm_c_api_internal.h"
 #include "../../version.h"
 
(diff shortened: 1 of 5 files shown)


core/iwasm/common/wasm_runtime_common.c
```diff
@@ -30,6 +30,9 @@
 #if WASM_ENABLE_FAST_JIT != 0
 #include "../fast-jit/jit_compiler.h"
 #endif
+#if WASM_ENABLE_JIT != 0 || WASM_ENABLE_WAMR_COMPILER != 0
```
lum1n0us, line 33: just curious, not an opinion. Can we use `#if WASM_ENABLE_AOT != 0` here?

wenyongh: No, macro WASM_ENABLE_WAMR_COMPILER is defined by wamrc CMakeLists.txt, and is wildly used in wasm_loader.c and some other places.

Theoretically, LLVM JIT and wamrc don't definitely depend on AOT, using `WASM_ENABLE_AOT` is not a good choice. And enabling AOT doesn't mean that LLVM JIT and wamrc are enabled, using `WASM_ENABLE_AOT` will cause to compile error for iwasm when JIT isn't built in.

PR #1680, opened 2022-11-03
chore: Update MinGW docs

phated:
I'm not sure if these docs are outdated, but I'm able to build WAMR in a MinGW environment (for OCaml bindings). Was uvwasi support fixed for MinGW in the past?

While it builds fine, I'm investigating a `UVWASI_EINVAL` error, which shouldn't be related to these docs.

### DIFF
--- a/doc/build_wamr.md
+++ b/doc/build_wamr.md
@@ -385,18 +385,14 @@ arguments for cmake:
 
 ```Bash
 cmake .. -G"Unix Makefiles" \
-         -DWAMR_BUILD_LIBC_UVWASI=0 \
          -DWAMR_BUILD_INVOKE_NATIVE_GENERAL=1 \
          -DWAMR_DISABLE_HW_BOUND_CHECK=1
 ````
 
 Note that WASI will be disabled until further work is done towards full MinGW support.
 
-- uvwasi not building out of the box, though it reportedly supports MinGW.
 - Failing compilation of assembler files, the C version of `invokeNative()` will
 be used instead.
-- Compiler complaining about missing `UnwindInfoAddress` field in `RUNTIME_FUNCTION`
-struct (winnt.h).
 
 
 VxWorks


wenyongh:
@phated Thanks for the test and reminding! Maybe it happens to be fixed by some PR recently.

And @lucianoiam, could you help confirm it? Since the building of MinGW was enabled by you:
https://github.com/bytecodealliance/wasm-micro-runtime/pull/1000
Thanks a lot.

lucianoiam:
Unfortunately I forgot some details but think I never managed to test WASI on MinGW. What is the shortest path to reproduce the issue? Is it enough by cloning WAMR and running one of the demos?

phated:
I don't know how an average person would go about testing this. I'm writing bindings for OCaml and the "setup-ocaml" action configures the CI to use MinGW (because that is the only way OCaml will run on Windows).

phated:
Converting this to a Draft because I found that MinGW fails when trying to setup stdin/stdout/stderr - if removed, then it fails on `bh_assert(argc >= sizeof(WASMExecEnv *) / sizeof(uint32));` in `invokeNative_general.c`

wenyongh:
> Converting this to a Draft because I found that MinGW fails when trying to setup stdin/stdout/stderr - if removed, then it fails on `bh_assert(argc >= sizeof(WASMExecEnv *) / sizeof(uint32));` in `invokeNative_general.c`

@phated Had better not use `invokeNative_general.c`, it may only work for x86-32 and doesn't work for most targets, use the assembly versions instead.

phated:
@wenyongh Sorry to derail the thread but the assembly versions don't work on MinGW. Should I open an issue with the error?

wenyongh:
> @wenyongh Sorry to derail the thread but the assembly versions don't work on MinGW. Should I open an issue with the error?

Yes, it is up to you. But I just wonder why the MinGW cannot support the assembly code? It also uses cmake to build the project.

wenyongh:
Close this PR since PR #1753 had already been merged to fix the issues.

PR #1682, opened 2022-11-04
Add control for the native stack check with hardware trap

wenyongh:
Add new options to control the native stack hw bound check feature:
- Besides the original option `cmake -DWAMR_DISABLE_HW_BOUND_CHECK=1/0`,
  add a new option `cmake -DWAMR_DISABLE_STACK_HW_BOUND_CHECK=1/0`
- When the linear memory hw bound check is disabled, the stack hw bound check
   will be disabled automatically, no matter what the input option is
- When the linear memory hw bound check is enabled, the stack hw bound check
  is enabled/disabled according to the value of input option
- Besides the original option `--bounds-checks=1/0`, add a new option
  `--stack-bounds-checks=1/0` for wamrc


product-mini/platforms/nuttx/wamr.mk
```diff
@@ -268,8 +268,14 @@ endif
 
 ifeq ($(CONFIG_INTERPRETERS_WAMR_DISABLE_HW_BOUND_CHECK),y)
 CFLAGS += -DWASM_DISABLE_HW_BOUND_CHECK=1
+CFLAGS += -DWASM_DISABLE_STACK_HW_BOUND_CHECK=1
 else
 CFLAGS += -DWASM_DISABLE_HW_BOUND_CHECK=0
+  ifeq ($(CONFIG_INTERPRETERS_WAMR_DISABLE_STACK_HW_BOUND_CHECK),y)
```
no1wudi, line 274: No such option in NuttX now，we can disable it if CONFIG_INTERPRETERS_WAMR_DISABLE_HW_BOUND_CHECK disabled.
BTW, even CONFIG_INTERPRETERS_WAMR_DISABLE_HW_BOUND_CHECK exists but it doesn't works actually since most NuttX supported platform don't have MMU.

PR #1685, opened 2022-11-04
linux-sgx: allow to open files with arbitrary pathes in the sandbox using IPFS.

JamesMenetrey:
Hello,

A limitation of the current implementation of SGX IPFS in WAMR prevented to openof files which were not in the current directory. This restriction is lifted and can now open files in paths, similarly to the WASI openat call, which takes into account the sandbox of the file system.

Under the scenes, SGX IPFS retrieves the path of the symbolic link of the file descriptor using `/proc/self` and resolves the path of the opened file using `readlink`.

Cheers

### DIFF
--- a/core/shared/platform/linux-sgx/sgx_file.c
+++ b/core/shared/platform/linux-sgx/sgx_file.c
@@ -88,6 +88,8 @@ ocall_linkat(int *p_ret, int olddirfd, const char *oldpath, int newdirfd,
 int
 ocall_unlinkat(int *p_ret, int dirfd, const char *pathname, int flags);
 int
+ocall_readlink(ssize_t *p_ret, const char *pathname, char *buf, size_t bufsiz);
+int
 ocall_readlinkat(ssize_t *p_ret, int dirfd, const char *pathname, char *buf,
                  size_t bufsiz);
 int
@@ -190,17 +192,25 @@ openat(int dirfd, const char *pathname, int flags, ...)
         errno = get_errno();
 
 #if WASM_ENABLE_SGX_IPFS != 0
-    // When WAMR uses Intel SGX IPFS to enabled, it opens a second
-    // file descriptor to interact with the secure file.
-    // The first file descriptor opened earlier is used to interact
-    // with the metadata of the file (e.g., time, flags, etc.).
-    int ret;
-    void *file_ptr = ipfs_fopen(fd, pathname, flags);
-    if (file_ptr == NULL) {
-        if (ocall_close(&ret, fd) != SGX_SUCCESS) {
-            TRACE_OCALL_FAIL();
-        }
+    struct stat sb;
+    int ret = fstatat(dirfd, pathname, &sb, 0);
+    if (ret < 0)
         return -1;
+
+    // Ony files are managed by SGX IPFS
+    if (S_ISREG(sb.st_mode)) {
+        // When WAMR uses Intel SGX IPFS to enabled, it opens a second
+        // file descriptor to interact with the secure file.
+        // The first file descriptor opened earlier is used to interact
+        // with the metadata of the file (e.g., time, flags, etc.).
+        int ret;
+        void *file_ptr = ipfs_fopen(fd, flags);
+        if (file_ptr == NULL) {
+            if (ocall_close(&ret, fd) != SGX_SUCCESS) {
+                TRACE_OCALL_FAIL();
+            }
+            return -1;
+        }
     }
 #endif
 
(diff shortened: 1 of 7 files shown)


core/shared/platform/linux-sgx/sgx_file.c
```diff
@@ -190,17 +192,25 @@ openat(int dirfd, const char *pathname, int flags, ...)
...
-    // with the metadata of the file (e.g., time, flags, etc.).
-    int ret;
-    void *file_ptr = ipfs_fopen(fd, pathname, flags);
-    if (file_ptr == NULL) {
-        if (ocall_close(&ret, fd) != SGX_SUCCESS) {
-            TRACE_OCALL_FAIL();
-        }
+    struct stat sb;
+    int ret = fstatat(dirfd, pathname, &sb, 0);
+    if (ret < 0)
```
wenyongh, line 197: Should `ocall_close(.., fd)` before return -1

core/shared/platform/linux-sgx/sgx_file.c
```diff
@@ -190,17 +192,25 @@ openat(int dirfd, const char *pathname, int flags, ...)
...
         return -1;
+
+    // Ony files are managed by SGX IPFS
+    if (S_ISREG(sb.st_mode)) {
+        // When WAMR uses Intel SGX IPFS to enabled, it opens a second
+        // file descriptor to interact with the secure file.
+        // The first file descriptor opened earlier is used to interact
+        // with the metadata of the file (e.g., time, flags, etc.).
+        int ret;
```
wenyongh, line 206: ret is already defined at L196, had better not define another variable to shadow the previous one.

core/shared/platform/linux-sgx/sgx_ipfs.c
```diff
@@ -291,8 +291,31 @@ ipfs_fopen(int fd, const char *filename, int flags)
...
+    // Using the symbolic link in /proc/self allows to retrieve the same path as
+    // opened by the initial openat and respects the chroot of WAMR.
+    int size_of_path = snprintf(NULL, 0, "/proc/self/fd/%d", fd) + 1;
+    if (size_of_path < 0) {
+        errno = EINVAL;
+        return NULL;
+    }
+
+    char *symbolic_path = BH_MALLOC(size_of_path);
+    snprintf(symbolic_path, size_of_path, "/proc/self/fd/%d", fd);
```
wenyongh, line 305: How about just defining a buf with enough size? Seems `char symbolic_path[32] = { 0 };` should be enough. Using BH_MALLOC/BH_FREE is relatively slow and not necessary?

core/shared/platform/linux-sgx/sgx_ipfs.c
```diff
@@ -291,8 +291,31 @@ ipfs_fopen(int fd, const char *filename, int flags)
...
+
+    // Resolve the symbolic link to real absolute path, because IPFS can only
+    // open a file with a same file name it was initially created. Otherwise,
+    // IPFS throws SGX_ERROR_FILE_NAME_MISMATCH.
+    char real_path[PATH_MAX];
+    size_t ret = readlink(symbolic_path, real_path, PATH_MAX);
+    real_path[PATH_MAX - 1] = '\0';
```
wenyongh, line 312: Seems not a good choice, the actual length should be much smaller than `PATH_MAX - 1`, how about `char real_path[PATH_MAX] = { 0 };` and removing this line?

samples/file/wasm-app/main.c
```diff
@@ -12,7 +12,8 @@
 #include <sys/stat.h>
 #include <unistd.h>
 
-#define PATH_TEST_FILE "test.txt"
+#define PATH_TEST_FOLDER "/test"
+#define PATH_TEST_FILE (PATH_TEST_FOLDER "/test.txt")
```
wenyongh, line 16: Operating on folder and file under `/` may require root privilege, using current directory or temporarily directory may be a better way

core/shared/platform/linux-sgx/sgx_ipfs.c
JamesMenetrey: If BH_MALLOC/BH_FREE are slow, that totally makes sense to use a fixed-size array on the stack. I propose not to initialize it with zeroes, since `snprintf` takes care of adding the terminating null byte.

The new proposed code check for too-long symbolic paths and sets errno to `ENAMETOOLONG` if so.

core/shared/platform/linux-sgx/sgx_ipfs.c
JamesMenetrey: Indeed, I removed that line and initialized the array with zeroes. Furthermore, since we no longer set the last byte of the array after the syscall, I subtract the `bufsiz` of 1 when calling `readlink`, since the syscall does not append a terminating null byte.

JamesMenetrey:
Hey @wenyongh,

Thanks for your insights! I have reviewed, commented and applied the changes.
Let me know if you have other feedback or concerns.

Cheers

wenyongh:
@JamesMenetrey Thanks a lot! It is much better now, I will merge the PR after the CI checks finish.

PR #1687, opened 2022-11-07
Fix/nuttx spectest option and symbols

dongsheng28849455:
Enable spec test option on nuttx platform
Register more symbols in sym_map

### DIFF
--- a/core/iwasm/aot/aot_reloc.h
+++ b/core/iwasm/aot/aot_reloc.h
@@ -124,6 +124,8 @@ typedef struct {
     { "memset", (void*)aot_memset },      \
     { "memmove", (void*)aot_memmove },    \
     { "memcpy", (void*)aot_memmove },     \
+    { "sqrt", (void*)aot_sqrt },          \
+    { "sqrtf", (void*)aot_sqrtf },        \
     REG_SYM(fmin),                        \
     REG_SYM(fminf),                       \
     REG_SYM(fmax),                        \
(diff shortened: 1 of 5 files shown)


yamt (approved):
lgtm

core/iwasm/aot/aot_reloc.h
```diff
@@ -136,6 +136,8 @@ typedef struct {
     REG_SYM(truncf),                      \
     REG_SYM(rint),                        \
     REG_SYM(rintf),                       \
+    REG_SYM(sqrt),                        \
+    REG_SYM(sqrtf),                       \
```
wenyongh, line 140: sqrt/sqrtf might be a macro but not function, adding them here may cause compile error, see Windows CI report:
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/3406870901

Suggest to register them like aot_memset/memmove:
Declare and implement aot_sqrt and aot_sqrtf in aot_runtime.h and aot_runtime.c, just after the code of aot_memset and aot_memmove:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/aot/aot_runtime.h#L448-L452

And here:
```C
{ "sqrt", (void*)aot_sqrt },
{ "sqrtf", (void*)aot_sqrtf },
```

dongsheng28849455: Thanks. then how can we know which APIs may not function but macro? if in future we encounter similar problems again

PR #1699, opened 2022-11-11
Fix missing float cmp for XIP

no1wudi:


core/iwasm/compilation/aot_emit_numberic.c
```diff
@@ -234,15 +234,49 @@ compile_op_float_min_max(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
         nan = LLVMConstRealOfString(ret_type, "NaN");
     char *intrinsic = is_min ? (is_f32 ? "llvm.minnum.f32" : "llvm.minnum.f64")
                              : (is_f32 ? "llvm.maxnum.f32" : "llvm.maxnum.f64");
+    bool is_i32 = is_f32;
```
wenyongh, line 237: Seems that `is_i32` isn't used?

wenyongh (commented):
LGTM

core/iwasm/compilation/aot_emit_numberic.c
no1wudi: Emm, is_i32 used in macro `LLVM_BUILD_OP_OR_INTRINSIC`

wenyongh: Got it, thanks.

PR #1702, opened 2022-11-15
Fix sample ref-types/wasm-c-api build error with wab2wasm low version

wenyongh:
Fix build error in sample ref-types and wasm-c-api caused by the
version of wat2wasm: refine makefile to auto detect the wat2wasm
version and add the compile flag accordingly.


lum1n0us:
LGTM

PR #1720, opened 2022-11-18
remove unused  checker

lum1n0us:


core/iwasm/interpreter/wasm_loader.c
```diff
@@ -713,34 +712,8 @@ wasm_loader_find_export(const WASMModule *module, const char *module_name,
...
-
-    if (export->index >= export_index_boundary) {
-        LOG_DEBUG("%s in the module %s is out of index (%d >= %d )", field_name,
-                  module_name, export->index, export_index_boundary);
-        set_error_buf(error_buf, error_buf_size, "incompatible import type");
-        return NULL;
-    }
-
+    /* since there is a validation in load_export_section(), it is for sure
+     * export->index is valid*/
```
wenyongh, line 716: The export index is checked in wasm_loader.c load_export_section, but isn't checked in aot_loader.c load_export_section, shall we check it?
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/aot/aot_loader.c#L1765-L1772

PR #1721, opened 2022-11-18
fix issue with restore operand stack

xujuntwt95329:


core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -2027,6 +2027,8 @@ call_wasm_with_hw_bound_check(WASMModuleInstance *module_inst,
     WASMJmpBuf jmpbuf_node = { 0 }, *jmpbuf_node_pop;
     uint32 page_size = os_getpagesize();
     uint32 guard_page_count = STACK_OVERFLOW_CHECK_GUARD_PAGE_COUNT;
+    WASMRuntimeFrame *prev_frame = wasm_exec_env_get_cur_frame(exec_env);
+    void *current_top = exec_env->wasm_stack.s.top;
```
wenyongh, line 2031: How about defining as `uint8 *prev_top`?

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -2081,13 +2083,18 @@ call_wasm_with_hw_bound_check(WASMModuleInstance *module_inst,
         ret = false;
     }
 
+    /* Note: can't check wasm_get_exception(module_inst) here, there may be
+     * exception which is not caught by hardware (e.g. uninitialized elements),
+     * then the stack-frame is already freed inside wasm_interp_call_wasm */
     if (wasm_get_exception(module_inst)) {
```
wenyongh, line 2089: Should be `if (!ret)`

core/iwasm/interpreter/wasm_runtime.c

core/iwasm/interpreter/wasm_runtime.c
xujuntwt95329: forgot to update the code 😂
updated, thanks~

Issue #1727, opened 2022-11-19
fast-jit "jit compilation failed: unknown error"


yamt:
see https://github.com/yamt/toywasm/blob/424755e4210568ed112b155000c05d772b02e989/test/run-ffmpeg.sh#L26-L33

x86-64 macOS
WAMR commit 021130f4a3377726d9b3ab595caf76214c55fa80
iwasm built with `-DWAMR_BUILD_BULK_MEMORY=1 -DWAMR_BUILD_FAST_JIT=1`

wenyongh:
@yamt Thanks for reporting the issue! I uploaded PR #1731 to fix it, could you please try again?

BTW, currently the Fast JIT in main branch uses eager compilation mode, all wasm functions must be compiled before execution, so the startup time may be a little long. You can try the branch `dev/multi_tier_jit`, in which the Fast JIT enables the lazy compilation mode to speedup the launch process. Note that this patch should be applied manually for that branch.

yamt:
> @yamt Thanks for reporting the issue! I uploaded PR #1731 to fix it, could you please try again?

i tested it. it worked fine if i gave large enough --jit-codecache-size and --stack-size=20000.

> 
> BTW, currently the Fast JIT in main branch uses eager compilation mode, all wasm functions must be compiled before execution, so the startup time may be a little long. You can try the branch `dev/multi_tier_jit`, in which the Fast JIT enables the lazy compilation mode to speedup the launch process. Note that this patch should be applied manually for that branch.

ok.
i just happened to try fast-jit. i have no real use cases or performance concerns at this point.

PR #1731, opened 2022-11-21
Fix fast jit issue of translating opcode i32.rem_s/i64.rem_s

wenyongh:


yamt:
while i don't really understand this part of code, i tested it and it worked.
https://github.com/bytecodealliance/wasm-micro-runtime/issues/1727#issuecomment-1321499856

PR #1734, opened 2022-11-21
add bh_print_vm to dump memory consumption information

lum1n0us:


core/shared/utils/bh_log.c
```diff
@@ -79,3 +82,63 @@ bh_print_time(const char *prompt)
 
     last_time_ms = curr_time_ms;
 }
+
+#ifdef linux
```
loganek, line 86: This ifdef makes me think whether it actually should be in `shared` folder? If it's shared with all `posix`  or `linux` platforms, maybe we should have `posix` or `linux` shared folder instead?

lum1n0us: 😆  

OK. `#ifdef` represents that the implementation with */proc/self/status* only workable on Linux-like system. And still looking for solution on other platforms. Eventually, all platforms should have its own implementation to collect memory consumption information. In my mind, `bh_print_vm()` should be a common feature.

core/shared/utils/bh_log.c
```diff
@@ -2,6 +2,9 @@
  * Copyright (C) 2019 Intel Corporation.  All rights reserved.
  * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
  */
+#ifdef linux
+#include <stdio.h>
+#endif
```
wenyongh, line 7: Had better not include <stdio.h> in non-platform source files, `#include "bh_platform.h"` instead, which includes platform_internal.h, and the latter includes `<stdio.h>`.

core/shared/utils/bh_log.c
wenyongh: Is it better to use `#if defined(BH_PLATFORM_LINUX) || defined(BH_PLATFORM_DARWIN)`?

core/shared/utils/bh_log.h
```diff
@@ -73,6 +73,14 @@ bh_log(LogLevel log_level, const char *file, int line, const char *fmt, ...);
 void
 bh_print_time(const char *prompt);
 
+void
+bh_print_vm(const char *prompt);
+
+void
+bh_log_vm(const char *function, uint32 line);
```
wenyongh, line 80: How about renaming them to `bh_print_vm_mem` and `bh_log_log_vm_mem`, and renaming `LOG_MEM` to `LOG_VM_MEM`? They are more understandable.

lum1n0us: `vm` is for *Virtual memory* here. So, maybe `bh_print_mem` and `bh_log_mem` ?

wenyongh: vm makes a little confused, it may refer to `virtual machine`, how about `bh_print_virtual_mem` and `bh_log_virtual_mem` to make it clear?

core/shared/utils/bh_log.c
wenyongh: Agree to add the platform related code in platform related folder, for example, define API like `os_print_virt_mem` in `core/shared/platform/include/platform_api_extension.h`, and implement it in `core/shared/platform/common/posix/posix_malloc.c` for posix like platforms currently, and implement it for Windows in the future.

core/iwasm/common/wasm_c_api.c
```diff
@@ -258,6 +258,12 @@ WASM_DEFINE_VEC_OWN(module, wasm_module_delete_internal)
 WASM_DEFINE_VEC_OWN(store, wasm_store_delete)
 WASM_DEFINE_VEC_OWN(valtype, wasm_valtype_delete)
 
+#ifndef NDEBUG
+#define WASM_C_DUMP_PROC_MEM() LOG_PROC_MEM()
+#else
+#define WASM_C_DUMP_PROC_MEM()
```
wenyongh, line 264: Had better `#define WASM_C_DUMP_PROC_MEM() (void)0`, or when we call it with `WASM_C_DUMP_PROC_MEM();`, there will be an unused `;`, which might cause compilation warning.

core/shared/platform/common/posix/posix_malloc.c
```diff
@@ -22,3 +22,49 @@ os_free(void *ptr)
 {
     free(ptr);
 }
+
+int
+os_dumps_proc_mem_info(char *out, unsigned int size)
```
wenyongh, line 27: Had better rename to `os_dump_proc_mem_info`, using `dumps` makes confusing

core/shared/platform/common/posix/posix_malloc.c
```diff
@@ -22,3 +22,49 @@ os_free(void *ptr)
...
+    while (fgets(line, sizeof(line), f)) {
+#if WASM_ENABLE_MEMORY_PROFILING != 0
+        if (strncmp(line, "Vm", 2) == 0 || strncmp(line, "Rss", 3) == 0) {
+#else
+        if (strncmp(line, "VmRSS", 5) == 0
+            || strncmp(line, "RssAnon", 7) == 0) {
+#endif
+            size_t line_len = strlen(line);
+            if (line_len >= size - 1 - out_idx) {
+                return -1;
```
wenyongh, line 53: Should `fclose(f)` before return -1

core/shared/platform/include/platform_api_extension.h
```diff
@@ -977,6 +977,19 @@ os_socket_set_broadcast(bh_socket_t socket, bool is_enabled);
...
+ * It may have variant implementations in different platforms
+ *
+ * @param out the output buffer. It is for sure the return content
+ *            is a c-string which ends up with '\0'
+ * @param size the size of the output buffer
+ *
+ * @return 0 if success, -1 otherwise
+ */
+int
+os_dumps_proc_mem_info(char *out, unsigned int size);
```
wenyongh, line 991: os_dump_proc_mem_info

core/shared/platform/windows/win_malloc.c
```diff
@@ -22,3 +22,9 @@ os_free(void *ptr)
 {
     free(ptr);
 }
+
+int
+os_dumps_proc_mem_info(char *out, unsigned int size)
```
wenyongh, line 27: Should also implement API for other platforms, e.g. esp-idf, riot , rt-thread and zephyr, or when we build them with wasm_c_api.c in debug mode, there will be undefined reference linkage issue.

core/shared/platform/include/platform_api_extension.h
lum1n0us: like `fgets` means get as a string?

wenyongh: Other places we use bh_print_proc_mem, bh_log_proc_mem, wasm_runtime_dump_module_mem_consumption, wasm_runtime_dump_perf_profiling, using `dumps` makes inconsistent :)

loganek, line 991: [minor] could we use `size_t` -like type for the `size` variable?

loganek: Also, I think we could probably make the API more verbose - having `char *` type as a return value doesn't tell implementation and the user how the data should be structured (so on windows we might have different format, on linux there will be a different format).
I'd suggest using a well-defined structure that's being returned here so:
1. implementers know exactly what sort of data needs to be returned
2. users of that method know what to expect
3. we can have a consistent formatting defined in a common layer, rather than have different formatting for each system
4. If we ever want to emit metrics from WAMR or generate statistics, we can't rely on unstructured text.

This is just a suggestion - if returning an unstructured blob from this function is really what we want, that's fine, just wanted to raise the concern and make sure we're making a conscious decision.

core/shared/utils/bh_log.c
```diff
@@ -79,3 +79,29 @@ bh_print_time(const char *prompt)
...
+    char buf[1024] = { 0 };
+
+    if (log_verbose_level < BH_LOG_LEVEL_DEBUG)
+        return;
+
+    if (os_dumps_proc_mem_info(buf, sizeof(buf)) != 0)
+        return;
+
+    os_printf("%s\n", prompt);
+    os_printf("===== /proc/self/status =====\n");
```
loganek, line 95: This won't always be a buffer from `/proc/self/status` I guess? (e.g. not in windows) I suggested another approach for this, see my comment above.

core/shared/platform/include/platform_api_extension.h
lum1n0us: I am thinking that `os_dumps_proc_mem_info` should return some kind of *memory usage reports* of the current process. The developer should not be worried about how the memory data is structured or how different every platform is. It doesn't limit the output format and the output items.

wenyongh (commented):
LGTM

PR #1739, opened 2022-11-22
Fix issue for QEMU, support to run on QEMU-M3

letanphuc:


core/shared/platform/common/math/math.c
```diff
@@ -1554,11 +1554,11 @@ atan2(double y, double x)
...
-double
-sqrt(double x)
-{
-    return freebsd_sqrt(x);
-}
+// double
+// sqrt(double x) __attribute__((weak))
+// {
+//     return freebsd_sqrt(x);
+// }
```
wenyongh, line 1561: Could we use macro to control it? For example,
```C
#if !(defined(BH_PLATFORM_ZEPHYR) && KERNEL_VERSION_NUMBER >= xxx)
...
#endif
```
I remember that the early Zephyr version doesn't support sqrt/sqrtf yet.

core/shared/platform/zephyr/platform_internal.h
```diff
@@ -7,11 +7,10 @@
 #define _PLATFORM_INTERNAL_H
 
 #include <autoconf.h>
-#include <zephyr.h>
-#include <kernel.h>
+#include <zephyr/kernel.h>
 #include <version.h>
 #if KERNEL_VERSION_NUMBER >= 0x020200 /* version 2.2.0 */
-#include <sys/printk.h>
+// #include <sys/printk.h>
```
wenyongh, line 13: Could we use macro to control the code:
```C
#include <autoconf.h>
#include <version.h>

#if KERNERL_VERSION_NUMBER <= xxx
...
#elif KERNERL_VERSION_NUMBER <= xxx
...
#else
...
#endif
```

product-mini/app-samples/hello-world/main.c
```diff
@@ -6,24 +6,38 @@
 #include <stdio.h>
 #include <stdlib.h>
 
-int
-main(int argc, char **argv)
-{
-    char *buf;
+void fibo(int x) {
```
wenyongh, line 9: The sample is used by many places, shall we not change it?

product-mini/platforms/zephyr/simple/src/main.c
```diff
@@ -5,47 +5,22 @@
 
 #include <stdlib.h>
 #include <string.h>
-#include "bh_platform.h"
+#include <zephyr/kernel.h>
```
wenyongh, line 8: Many codes are removed, like riscv and esp32, must we remove them?

wenyongh:
@letanphuc I uploaded PR #1757 to fix the issues since several comments may be not easy for you to fix, could you help test it? Thanks.

letanphuc:
I tested, it works as expected.
Thanks for correcting me

wenyongh:
> I tested, it works as expected. Thanks for correcting me

OK, I merged that PR, if you have extra fixings, please rebase with the main branch and add changes in this PR.

PR #1746, opened 2022-11-24
Enlarge the default wasm operand stack size to 64KB

wenyongh:
Enlarge the default wasm operand stack size to 64KB since the original default
size 16KB is a little small, and the operand stack overflow exception is often
thrown when running wasm apps.


yamt (approved):
lgtm

Issue #1738, opened 2022-11-22
no way to obtain WASI exit code


yamt:
a straightforward implementation would introduce:

* wasm_runtime_set_wasi_exit_code
* wasm_runtime_get_wasi_exit_code

and use the former in libc-wasi/uwasi and use the latter in iwasm and other embedders.

if it sounds ok, i can submit an implementation.

wenyongh:
@yamt, do you mean to store the wasi exit code in native API `wasi_proc_exit(wasm_exec_env_t exec_env, wasi_exitcode_t rval)` to module instance, and get it from host embedder?

Seems that we only need to provide wasm_runtime_get_wasi_exit_code? Since the wasi exit code should not be set by host embedder.
And could you extend the wasi context to store the exit code? For example, for libc-wasi, add field `wasi_exitcode_t exit_code` in struct WASIContext. For libc-uvwasi, append extra `wasi_exitcode_t exit_code` to the end of uvwasi_t, and malloc uvwas_t with extra buffer when wasm_runtime_init_wasi:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/common/wasm_runtime_common.c#L2685

yamt:
> @yamt, do you mean to store the wasi exit code in native API `wasi_proc_exit(wasm_exec_env_t exec_env, wasi_exitcode_t rval)` to module instance, and get it from host embedder?

yes.

> Seems that we only need to provide wasm_runtime_get_wasi_exit_code? Since the wasi exit code should not be set by host embedder.

right.

> And could you extend the wasi context to store the exit code? For example, for libc-wasi, add field `wasi_exitcode_t exit_code` in struct WASIContext. For libc-uvwasi, append extra `wasi_exitcode_t exit_code` to the end of uvwasi_t, and malloc uvwas_t with extra buffer when wasm_runtime_init_wasi: https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/common/wasm_runtime_common.c#L2685

i was thinking to store the exit code in module_inst.
but it can be done that way too.

wenyongh:
[...]

PR #1748, opened 2022-11-24
Add wasm_runtime_get_wasi_exit_code

yamt:
https://github.com/bytecodealliance/wasm-micro-runtime/issues/1738

### DIFF
--- a/core/iwasm/common/wasm_runtime_common.h
+++ b/core/iwasm/common/wasm_runtime_common.h
@@ -369,9 +369,13 @@ typedef struct WASIContext {
     char **argv_list;
     char *env_buf;
     char **env_list;
+    uint32_t exit_code;
 } WASIContext;
 #else
-typedef uvwasi_t WASIContext;
+typedef struct WASIContext {
+    uvwasi_t uvwasi;
+    uint32_t exit_code;
+} WASIContext;
 #endif
 #endif
 
@@ -767,6 +771,10 @@ wasm_runtime_is_wasi_mode(WASMModuleInstanceCommon *module_inst);
 /* See wasm_export.h for description */
 WASM_RUNTIME_API_EXTERN WASMFunctionInstanceCommon *
 wasm_runtime_lookup_wasi_start_function(WASMModuleInstanceCommon *module_inst);
+
+/* See wasm_export.h for description */
+WASM_RUNTIME_API_EXTERN uint32_t
+wasm_runtime_get_wasi_exit_code(WASMModuleInstanceCommon *module_inst);
 
 bool
 wasm_runtime_init_wasi(WASMModuleInstanceCommon *module_inst,
--- a/core/iwasm/include/wasm_export.h
+++ b/core/iwasm/include/wasm_export.h
@@ -451,6 +451,18 @@ WASM_RUNTIME_API_EXTERN wasm_function_inst_t
 wasm_runtime_lookup_wasi_start_function(wasm_module_inst_t module_inst);
 
 /**
+ * Get WASI exit code.
+ *
+ * After a WASI command completed its execution, an embedder can
+ * call this function to get its exit code. (that is, the value given
+ * to proc_exit.)
+ *
+ * @param module_inst the module instance
+ */
+WASM_RUNTIME_API_EXTERN uint32_t
+wasm_runtime_get_wasi_exit_code(wasm_module_inst_t module_inst);
+
+/**
  * Lookup an exported function in the WASM module instance.
  *
  * @param module_inst the module instance
--- a/core/iwasm/libraries/libc-uvwasi/libc_uvwasi_wrapper.c
+++ b/core/iwasm/libraries/libc-uvwasi/libc_uvwasi_wrapper.c
@@ -11,9 +11,6 @@
 #define get_module_inst(exec_env) \
     wasm_runtime_get_module_inst(exec_env)
 
-#define get_wasi_ctx(module_inst) \
-    wasm_runtime_get_wasi_ctx(module_inst)
-
 #define validate_app_addr(offset, size) \
     wasm_runtime_validate_app_addr(module_inst, offset, size)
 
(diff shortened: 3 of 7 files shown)


yamt:
i plan to use this for https://github.com/WebAssembly/wasi-testsuite/pull/16

core/iwasm/common/wasm_runtime_common.h
```diff
@@ -369,9 +369,13 @@ typedef struct WASIContext {
     char **argv_list;
     char *env_buf;
     char **env_list;
+    uint32_t exit_code;
```
loganek, line 372: I don't see where we initialize `exit_code`; is there something like `memset(ctx, 0, sizeof(WASIContext))` in the startup code to set the value of `exit_code` to 0? Otherwise it'll contain garbage if the module doesn't call `proc_exit`.

core/iwasm/libraries/libc-uvwasi/libc_uvwasi_wrapper.c
```diff
@@ -72,9 +69,24 @@ typedef struct iovec_app {
     uint32 buf_len;
 } iovec_app_t;
 
+typedef struct WASIContext {
+    uvwasi_t uvwasi;
+    uint32_t exit_code;
+} WASIContext;
```
loganek, line 75: That's probably out of scope of this PR as it needs a bit larger refactoring, but I think we should strive to have one declaration per context, instead of duplicating it. I've opened an issue to track it: https://github.com/bytecodealliance/wasm-micro-runtime/issues/1749

core/iwasm/common/wasm_runtime_common.h
```diff
@@ -768,6 +772,10 @@ wasm_runtime_is_wasi_mode(WASMModuleInstanceCommon *module_inst);
 WASM_RUNTIME_API_EXTERN WASMFunctionInstanceCommon *
 wasm_runtime_lookup_wasi_start_function(WASMModuleInstanceCommon *module_inst);
 
+/* See wasm_export.h for description */
```
loganek, line 775: Could you please add documentation for this function in `wasm_export.h`?

product-mini/platforms/posix/main.c
```diff
@@ -643,14 +643,29 @@ main(int argc, char *argv[])
...
+    ret = 0;
+    if (is_repl_mode) {
         app_instance_repl(wasm_module_inst);
-    else if (func_name)
-        app_instance_func(wasm_module_inst, func_name);
-    else
-        app_instance_main(wasm_module_inst);
+    }
+    else if (func_name) {
+        if (app_instance_func(wasm_module_inst, func_name)) {
```
wenyongh, line 651: app_instance_func always return NULL, how about using `wasm_runtime_get_exception` to set the ret instead?

product-mini/platforms/posix/main.c
```diff
@@ -643,14 +643,29 @@ main(int argc, char *argv[])
...
+            /* got an exception */
+            ret = 1;
+        }
+    }
+    else {
+        if (app_instance_main(wasm_module_inst)) {
```
wenyongh, line 657: Same as above

product-mini/platforms/posix/main.c
```diff
@@ -643,14 +643,29 @@ main(int argc, char *argv[])
...
+            /* got an exception */
+            ret = 1;
+        }
+    }
 
-    ret = 0;
+#if WASM_ENABLE_LIBC_WASI != 0
+    if (ret == 0) {
+        /* propagate wasi exit code. */
+        ret = wasm_runtime_get_wasi_exit_code(wasm_module_inst);
```
wenyongh, line 666: Had better also apply similar changes to windows main.c?

core/iwasm/include/wasm_export.h
```diff
@@ -450,6 +450,9 @@ wasm_runtime_is_wasi_mode(wasm_module_inst_t module_inst);
 WASM_RUNTIME_API_EXTERN wasm_function_inst_t
 wasm_runtime_lookup_wasi_start_function(wasm_module_inst_t module_inst);
 
+WASM_RUNTIME_API_EXTERN uint32_t
```
wenyongh, line 453: Had better add comments?

core/iwasm/libraries/libc-uvwasi/libc_uvwasi_wrapper.c
yamt: i tend to agree. but it's how this code base is currently structured.

core/iwasm/common/wasm_runtime_common.h
yamt: done

product-mini/platforms/posix/main.c
yamt: oops, i forgot to add a few hunks. fixed.

product-mini/platforms/posix/main.c
yamt: done. i haven't tested it locally at all though.

core/iwasm/include/wasm_export.h
yamt: done

product-mini/platforms/posix/main.c
```diff
@@ -643,14 +643,29 @@ main(int argc, char *argv[])
...
+    }
+    else {
+        if (app_instance_main(wasm_module_inst)) {
+            /* got an exception */
+            ret = 1;
+        }
+    }
 
-    ret = 0;
+#if WASM_ENABLE_LIBC_WASI != 0
```
loganek, line 663: Just wanted to doublecheck - won't that break any user scenario right now? I guess many scripts assumed the return code is ignored by iwasm, so with the update it's possible this breaks some of the usecases. Just want to highlight we have to definitely include it in the release notes and probably highlight as backward incompatible change - any thoughts @wenyongh @yamt ?

wenyongh: Seems that it won't break any user scenarios for us - normally we only use iwasm to run spec cases and standalone cases. Agree to highlight the change in the release notes. And not sure whether it is very important to the users? iwasm just gives a sample, the host embedders can embed WAMR into their application and decide how to handle the exit code.

wenyongh (commented):
LGTM

PR #1751, opened 2022-11-24
Using trap for error message from wasm_instance_new

nchamzn:
* Similar to [this PR](https://github.com/bytecodealliance/wasm-micro-runtime/pull/1526) but uses a wasm_trap_t instead of adding error_buf to the API.
* Trap will need to be deleted by the caller but is not a breaking change as it is only created if trap is not NULL.
* Added error messages for all failure cases here, tried to make them accurate but welcome feedback for improvements.

### DIFF
--- a/core/iwasm/common/wasm_c_api.c
+++ b/core/iwasm/common/wasm_c_api.c
@@ -4427,25 +4427,25 @@ failed:
 
 wasm_instance_t *
 wasm_instance_new(wasm_store_t *store, const wasm_module_t *module,
-                  const wasm_extern_vec_t *imports, own wasm_trap_t **traps)
+                  const wasm_extern_vec_t *imports, own wasm_trap_t **trap)
 {
-    return wasm_instance_new_with_args(store, module, imports, traps,
+    return wasm_instance_new_with_args(store, module, imports, trap,
                                        KILOBYTE(32), KILOBYTE(32));
 }
 
 wasm_instance_t *
 wasm_instance_new_with_args(wasm_store_t *store, const wasm_module_t *module,
                             const wasm_extern_vec_t *imports,
-                            own wasm_trap_t **traps, const uint32 stack_size,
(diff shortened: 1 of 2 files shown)


loganek (approved):
LGTM

wenyongh (commented):
LGTM

core/iwasm/common/wasm_c_api.c
```diff
@@ -4501,18 +4507,22 @@ wasm_instance_new_with_args(wasm_store_t *store, const wasm_module_t *module,
...
         if (!import_count_verified) {
+            snprintf(sub_error_buf, sizeof(sub_error_buf),
+                     "Failed to verify import count");
             goto failed;
         }
     }
 
     instance->inst_comm_rt = wasm_runtime_instantiate(
-        *module, stack_size, heap_size, error_buf, sizeof(error_buf));
+        *module, stack_size, heap_size, sub_error_buf, sizeof(sub_error_buf));
```
lum1n0us, line 4517: L4519 `LOG_ERROR(error_buf)` should be removed

core/iwasm/common/wasm_c_api.c
```diff
@@ -4587,18 +4603,29 @@ wasm_instance_new_with_args(wasm_store_t *store, const wasm_module_t *module,
...
 
 failed:
-    LOG_DEBUG("%s failed", __FUNCTION__);
+    snprintf(error_buf, sizeof(error_buf), "%s failed: %s", __FUNCTION__,
+             sub_error_buf);
+    if (trap != NULL) {
+        wasm_message_t message = { 0 };
+        wasm_name_new_from_string(&message, error_buf);
+        *trap = wasm_trap_new(store, &message);
+    }
```
lum1n0us, line 4627:
```suggestion
      wasm_byte_vec_delete(&message);
    }
```

PR #1754, opened 2022-11-25
Compile WAMR on platforms that don't support IPV6

guangyuhu:
### Summary

Adding changes so that WAMR can compile on platforms that don't have IPV6 support.

### DIFF
--- a/core/shared/platform/common/posix/posix_socket.c
+++ b/core/shared/platform/common/posix/posix_socket.c
@@ -16,7 +16,9 @@ textual_addr_to_sockaddr(const char *textual, int port, struct sockaddr *out,
                          socklen_t *out_len)
 {
     struct sockaddr_in *v4;
+#ifdef IPPROTO_IPV6
     struct sockaddr_in6 *v6;
+#endif
 
     assert(textual);
 
@@ -28,6 +30,7 @@ textual_addr_to_sockaddr(const char *textual, int port, struct sockaddr *out,
         return true;
     }
 
+#ifdef IPPROTO_IPV6
     v6 = (struct sockaddr_in6 *)out;
     if (inet_pton(AF_INET6, textual, &v6->sin6_addr.s6_addr) == 1) {
         v6->sin6_family = AF_INET6;
@@ -35,6 +38,7 @@ textual_addr_to_sockaddr(const char *textual, int port, struct sockaddr *out,
         *out_len = sizeof(struct sockaddr_in6);
         return true;
     }
+#endif
 
     return false;
 }
@@ -55,6 +59,7 @@ sockaddr_to_bh_sockaddr(const struct sockaddr *sockaddr, socklen_t socklen,
             bh_sockaddr->is_ipv4 = true;
             return BHT_OK;
         }
+#ifdef IPPROTO_IPV6
         case AF_INET6:
         {
             struct sockaddr_in6 *addr = (struct sockaddr_in6 *)sockaddr;
@@ -75,6 +80,7 @@ sockaddr_to_bh_sockaddr(const struct sockaddr *sockaddr, socklen_t socklen,
             bh_sockaddr->is_ipv4 = false;
             return BHT_OK;
         }
+#endif
         default:
             errno = EAFNOSUPPORT;
             return BHT_ERROR;
@@ -92,6 +98,7 @@ bh_sockaddr_to_sockaddr(const bh_sockaddr_t *bh_sockaddr,
         addr->sin_addr.s_addr = htonl(bh_sockaddr->addr_bufer.ipv4);
         *socklen = sizeof(*addr);
     }
+#ifdef IPPROTO_IPV6
     else {
         struct sockaddr_in6 *addr = (struct sockaddr_in6 *)sockaddr;
         size_t i;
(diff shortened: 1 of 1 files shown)


core/shared/platform/common/posix/posix_socket.c
```diff
@@ -168,9 +176,15 @@ os_socket_bind(bh_socket_t socket, const char *host, int *port)
...
-    *port = ntohs(addr.ss_family == AF_INET
-                      ? ((struct sockaddr_in *)&addr)->sin_port
-                      : ((struct sockaddr_in6 *)&addr)->sin6_port);
+    if (addr.ss_family == AF_INET) {
+        *port = ntohs(((struct sockaddr_in *)&addr)->sin_port);
+    } else {
+#ifdef IPPROTO_IPV6
+        *port = ntohs(((struct sockaddr_in6 *)&addr)->sin6_port);
+#else
+    goto fail;
```
wenyongh, line 185: Coding style, should be 8 leading spaces

core/shared/platform/common/posix/posix_socket.c
```diff
@@ -881,15 +913,23 @@ os_socket_get_ip_multicast_ttl(bh_socket_t socket, uint8_t *ttl_s)
 int
 os_socket_set_ipv6_only(bh_socket_t socket, bool is_enabled)
 {
+#ifdef IPPROTO_IPV6
     return os_socket_setbooloption(socket, IPPROTO_IPV6, IPV6_V6ONLY,
                                    is_enabled);
+#else
+        return BHT_ERROR;
```
wenyongh, line 920: Should be 4 leading spaces

core/shared/platform/common/posix/posix_socket.c
```diff
@@ -881,15 +913,23 @@ os_socket_get_ip_multicast_ttl(bh_socket_t socket, uint8_t *ttl_s)
...
 }
 
 int
 os_socket_get_ipv6_only(bh_socket_t socket, bool *is_enabled)
 {
+#ifdef IPPROTO_IPV6
     return os_socket_getbooloption(socket, IPPROTO_IPV6, IPV6_V6ONLY,
                                    is_enabled);
+#else
+        return BHT_ERROR;
```
wenyongh, line 931: Same as above

wenyongh (commented):
Hi, there are some coding style issues, please help to resolve, or refer to:
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/3549594148/jobs/5962153784

```
    1. Install clang-format-12.0.0
    Normally we can install it by `sudo apt-get install clang-format-12`,
    or download the `clang+llvm-12.0.0-xxx-tar.xz` package from
      https://github.com/llvm/llvm-project/releases/tag/llvmorg-12.0.0
    and install it

    2. Format the C/C++ source file
    ``` shell
    cd path/to/wamr/root
    clang-format-12 --style file -i path/to/file
    ```
```

guangyuhu:
Thanks for your comments, ran clang-format over the changes so the style issues should be resolved now.

core/shared/platform/common/posix/posix_socket.c
```diff
@@ -168,9 +176,16 @@ os_socket_bind(bh_socket_t socket, const char *host, int *port)
...
     }
 
-    *port = ntohs(addr.ss_family == AF_INET
-                      ? ((struct sockaddr_in *)&addr)->sin_port
-                      : ((struct sockaddr_in6 *)&addr)->sin6_port);
+    if (addr.ss_family == AF_INET) {
+        *port = ntohs(((struct sockaddr_in *)&addr)->sin_port);
+    }
+    else {
+#ifdef IPPROTO_IPV6
```
loganek, line 183: I'm not sure if we ever actually get to the `else` branch, I'd expect the `bind()` to fail sooner if IPv6 is not supported.

I think we can probably fail way before making any syscall if IPv6 is not supported; my suggestion would be to move textual_addr_to_sockaddr at the very top, and move this if...else just after that. What do you think?

core/shared/platform/common/posix/posix_socket.c
```diff
@@ -318,6 +333,7 @@ os_socket_inet_network(bool is_ipv4, const char *cp, bh_ip_addr_buffer_t *out)
         /* Note: ntohl(INADDR_NONE) == INADDR_NONE */
         out->ipv4 = ntohl(out->ipv4);
     }
+#ifdef IPPROTO_IPV6
     else {
```
loganek, line 337: Do we handle it correctly? I think if `is_ipv4` is set to `false` and ipv6 is not supported, we should probably return an error instead (I think right now we return `BHT_OK`).

core/shared/platform/common/posix/posix_socket.c
guangyuhu: Makes sense, moved these bits to the start of this function.

core/shared/platform/common/posix/posix_socket.c
guangyuhu: You are right, changed this to return `BHT_ERROR` and set `EAFNOSUPPORT` as the `errno` where applicable.

loganek (approved):
LGTM

PR #1760, opened 2022-11-28
fix a typo in workflows

lum1n0us:


.github/workflows/compilation_on_android_ubuntu.yml
```diff
@@ -183,6 +183,12 @@ jobs:
             make_options_feature: "-DWAMR_BUILD_MINI_LOADER=1"
           - make_options_run_mode: $LLVM_EAGER_JIT_BUILD_OPTIONS
             make_options_feature: "-DWAMR_BUILD_MINI_LOADER=1"
+          # FAST-JIT mode doesn't support tail call
+          - make_options_run_mode: $FAST_JIT_BUILD_OPTIONS
+            make_options_feature: "-DWAMR_BUILD_TAIL_CALL=1"
```
wenyongh, line 188: The compilation error of fast jit when adding `-DWAMR_BUILD_TAIL_CALL` is due to we didn't remove the option control code, by default the tail call feature should be enabled when the WASM_ENABLE_TAIL_CALL macro is defined, and no need to use cc->enable_tail_call to control it again. We can fix it with:
```diff
--- a/core/iwasm/fast-jit/jit_frontend.c
+++ b/core/iwasm/fast-jit/jit_frontend.c
@@ -1286,10 +1286,6 @@ jit_compile_func(JitCompContext *cc)

 #if WASM_ENABLE_TAIL_CALL != 0
             case WASM_OP_RETURN_CALL:
-                if (!cc->enable_tail_call) {
-                    jit_set_last_error(cc, "unsupported opcode");
-                    return false;
-                }
                 read_leb_uint32(frame_ip, frame_ip_end, func_idx);
                 if (!jit_compile_op_call(cc, func_idx, true))
                     return false;
@@ -1301,11 +1297,6 @@ jit_compile_func(JitCompContext *cc)
             {
                 uint32 tbl_idx;

-                if (!cc->enable_tail_call) {
-                    jit_set_last_error(cc, "unsupported opcode");
-                    return false;
-                }
-
                 read_leb_uint32(frame_ip, frame_ip_end, type_idx);
```
How about applying the patch, and removing L186-188 here?

PR #1762, opened 2022-11-28
Add memory watchpoint support to debugger

cimacmillan:
### What

Added memory watchpoint support to the WAMR debug engine

### Why

This allows us to add watchpoints to variables for debugging. For instance:

```
breakpoint set variable var
```

Will pause WAMR execution when the address at `var` is written to. You can also set read (and r/w) watchpoints by passing this flag. This will pause execution when the address at var is read.

```
watchpoint set variable -w read var
```

### How

Added two linked lists for read/write watchpoints. When the debug message handler receives a watchpoint request, it adds/removes to one/both of these lists. In the interpreter, when an address is read or stored to, check whether the address is in these lists. If so, throw a sigtrap and suspend the process. 

This can support an arbitrary number of watchpoints. But there's some tradeoff that the more watchpoints, the more checks. This could be improved with a hashmap, although I think a small number of watchpoints (1 or 2) would be typical for most debugging sessions.

### DIFF
--- a/core/iwasm/interpreter/wasm_interp_classic.c
+++ b/core/iwasm/interpreter/wasm_interp_classic.c
@@ -71,20 +71,24 @@ typedef float64 CellType_F64;
     } while (0)
 
 #if WASM_ENABLE_DEBUG_INTERP != 0
-#define TRIGGER_WATCHPOINT_SIGTRAP()                          \
-    wasm_cluster_thread_send_signal(exec_env, WAMR_SIG_TRAP); \
-    CHECK_SUSPEND_FLAGS();
+#define TRIGGER_WATCHPOINT_SIGTRAP()                              \
+    do {                                                          \
+        wasm_cluster_thread_send_signal(exec_env, WAMR_SIG_TRAP); \
+        CHECK_SUSPEND_FLAGS();                                    \
+    } while (0)
 
-#define CHECK_WATCHPOINT(list, current_addr)                           \
-    WASMDebugWatchPoint *watchpoint = bh_list_first_elem(list);        \
-    while (watchpoint) {                                               \
-        WASMDebugWatchPoint *next = bh_list_elem_next(watchpoint);     \
-        if (watchpoint->addr <= current_addr                           \
-            && watchpoint->addr + watchpoint->length > current_addr) { \
-            TRIGGER_WATCHPOINT_SIGTRAP()                               \
-        }                                                              \
-        watchpoint = next;                                             \
-    }
+#define CHECK_WATCHPOINT(list, current_addr)                               \
+    do {                                                                   \
+        WASMDebugWatchPoint *watchpoint = bh_list_first_elem(list);        \
+        while (watchpoint) {                                               \
+            WASMDebugWatchPoint *next = bh_list_elem_next(watchpoint);     \
+            if (watchpoint->addr <= current_addr                           \
+                && watchpoint->addr + watchpoint->length > current_addr) { \
+                TRIGGER_WATCHPOINT_SIGTRAP();                              \
+            }                                                              \
+            watchpoint = next;                                             \
+        }                                                                  \
+    } while (0)
 
 #define CHECK_READ_WATCHPOINT(addr, offset) \
     CHECK_WATCHPOINT(watch_point_list_read, WASM_ADDR_OFFSET(addr + offset))
@@ -1813,8 +1817,8 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                 addr = POP_I32();
                 CHECK_MEMORY_OVERFLOW(4);
                 PUSH_I32(LOAD_I32(maddr));
-                (void)flags;
                 CHECK_READ_WATCHPOINT(addr, offset);
+                (void)flags;
                 HANDLE_OP_END();
             }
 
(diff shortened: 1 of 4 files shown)


core/iwasm/libraries/debug-engine/debug_engine.h
```diff
@@ -36,6 +36,12 @@ typedef struct WASMDebugBreakPoint {
     uint64 orignal_data;
 } WASMDebugBreakPoint;
 
+typedef struct WASMDebugWatchPoint {
+    struct WASMDebugWatchPoint *next;
```
xujuntwt95329, line 40:
```suggestion
    bh_list_link l;
```

Had better use `bh_list_link`

cimacmillan: Thanks for the feedback, will revise :)

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -70,6 +70,31 @@ typedef float64 CellType_F64;
             goto unaligned_atomic;                                   \
     } while (0)
 
+#if WASM_ENABLE_DEBUG_INTERP != 0
+#define TRIGGER_WATCHPOINT_SIGTRAP()                          \
```
wenyongh, line 74: Please use `do { .. } while (0)` to encapsulate the code, or when we call the macro `TRIGGER_WATCHPOINT_SIGTRAP();`, there will be two `;` when the macro is expanded, which might cause compile warning.

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -70,6 +70,31 @@ typedef float64 CellType_F64;
...
+    wasm_cluster_thread_send_signal(exec_env, WAMR_SIG_TRAP); \
+    CHECK_SUSPEND_FLAGS();
+
+#define CHECK_WATCHPOINT(list, current_addr)                           \
+    WASMDebugWatchPoint *watchpoint = bh_list_first_elem(list);        \
```
wenyongh, line 79: Same as above

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -1786,6 +1814,7 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                 CHECK_MEMORY_OVERFLOW(4);
                 PUSH_I32(LOAD_I32(maddr));
                 (void)flags;
+                CHECK_READ_WATCHPOINT(addr, offset);
```
wenyongh, line 1817: How about moving this line before line 1816? Normally we put `(void)xxx` at the end.

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -1800,6 +1829,7 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                 CHECK_MEMORY_OVERFLOW(8);
                 PUSH_I64(LOAD_I64(maddr));
                 (void)flags;
+                CHECK_READ_WATCHPOINT(addr, offset);
```
wenyongh, line 1832: Same as above

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -1988,6 +2030,7 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                 }
 
                 (void)flags;
+                CHECK_WRITE_WATCHPOINT(addr, offset);
```
wenyongh, line 2033: Same as above

core/iwasm/libraries/debug-engine/handler.c
```diff
@@ -643,46 +651,132 @@ handle_get_write_memory(WASMGDBServer *server, char *payload)
...
+        LOG_ERROR("Unsupported number of remove break arguments %d", arg_c);
+        write_packet(server, "");
+        return;
+    }
+
+    switch (type) {
+        case eBreakpointSoftware:
+            handle_breakpoint_software_remove(server, addr, length);
+            break;
+            break;
```
wenyongh, line 762: Duplicated break

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -70,6 +70,35 @@ typedef float64 CellType_F64;
...
+        }                                                                  \
+    } while (0)
+
+#define CHECK_READ_WATCHPOINT(addr, offset) \
+    CHECK_WATCHPOINT(watch_point_list_read, WASM_ADDR_OFFSET(addr + offset))
+#define CHECK_WRITE_WATCHPOINT(addr, offset) \
+    CHECK_WATCHPOINT(watch_point_list_write, WASM_ADDR_OFFSET(addr + offset))
+#else
+#define CHECK_READ_WATCHPOINT(addr, offset)
+#define CHECK_WRITE_WATCHPOINT(addr, offset)
```
wenyongh, line 99: Had better define them as `(void)0`, or when it is called with `CHECK_READ_WATCHPOINT(addr, offset);`, there will be a single `;` after extended, and may cause compile warning.

core/iwasm/interpreter/wasm_interp_classic.c
wenyongh: Yes, it looks better :)

wenyongh (commented):
@cimacmillan Only a comment left from me, could you help modify the code again? Thanks.

core/iwasm/interpreter/wasm_interp_classic.c
cimacmillan: Good catch thanks

PR #1765, opened 2022-11-28
Adding option to pass user data to allocator functions

nchamzn:
Adds an option to pass user data to the allocator functions. It's common to do this so that if I am embedding Wamr in to my application then I can pass a struct as user data and access that struct from the allocator which gives me the ability to do things such as track allocation statistics within my allocator.

### Notes

* I added this as a new option so I didn't break the existing API for users.
* I moved the definitions of ``MemAllocOption`` and ``mem_alloc_type_t`` to a new header, I don't know if there was a reason they were duplicated.

### Example usage:
```
Engine::Engine() : _total_allocations(0)
{
    RuntimeInitArgs args;
    memset(&args, 0, sizeof(RuntimeInitArgs));

    args.mem_alloc_type = Alloc_With_Allocator_With_Data;
    args.mem_alloc_option.allocator_with_data.data = reinterpret_cast<void*>(this);
    args.mem_alloc_option.allocator_with_data.malloc_func = reinterpret_cast<void*>(engine_malloc);
    args.mem_alloc_option.allocator_with_data.realloc_func = reinterpret_cast<void*>(engine_realloc);
    args.mem_alloc_option.allocator_with_data.free_func = reinterpret_cast<void*>(engine_free);

    // ...
    auto engine = wasm_engine_new_with_args(args.mem_alloc_type, &args.mem_alloc_option);

    // ...
}

void* Engine::malloc(void* data, size_t size)
{
    auto that = reinterpret_cast<Engine*>(data);
    that->_total_allocations++;
    // Now allocate some memory
}

void Engine::free(void* data, void* ptr)
{
    auto that = reinterpret_cast<Engine*>(data);
    that->_total_allocations--;
    // Now free some memory
}
```


core/iwasm/include/mem_alloc_export.h
```diff
@@ -0,0 +1,40 @@
...
+/* Memory allocator type */
+typedef enum {
+    /* pool mode, allocate memory from user defined heap buffer */
+    Alloc_With_Pool = 0,
+    /* user allocator mode, allocate memory from user defined
+       malloc function */
+    Alloc_With_Allocator,
+    /* user allocator mode, allocate memory from user defined
+       malloc function with user data */
+    Alloc_With_Allocator_With_Data,
```
loganek, line 15: Should that be moved to the bottom as a last option? Otherwise this change breaks ABI compatibility.

nchamzn: Good point, updated

core/iwasm/include/mem_alloc_export.h
```diff
@@ -0,0 +1,40 @@
...
+        void *malloc_func;
+        void *realloc_func;
+        void *free_func;
+    } allocator;
+    struct {
+        void *malloc_func;
+        void *realloc_func;
+        void *free_func;
+        void *data;
+    } allocator_with_data;
```
loganek, line 37: I'm curious if you also explored alternative approach where we have a single struct (`allocator`) with `void* data` inside `ifdef`? It **might** make the code less readable, but that **possibly** would simplify the implementation? That's just an idea though, not sure if you already thought about it.

wenyongh:
@nchamzn, seems that in you host application, you pass the malloc_func/free_func callback (mode Alloc_With_Allocator) to runtime init, and want to  track the allocation statistics, right? Not sure why not wrap the malloc_func/free_func, do the statistics in the wrapper APIs, and pass the wrapper APIs to runtime? For example:
```C
static data;
void *my_malloc(uint32 size) {
   data->alloc_count++;
   return engine_malloc(size);
}

void my_free(void *ptr) {
    data->free_count++;
    engine_free(ptr);
}

    RuntimeInitArgs args;
    memset(&args, 0, sizeof(RuntimeInitArgs));

    args.mem_alloc_type = Alloc_With_Allocator;
    args.mem_alloc_option.allocator.malloc_func = reinterpret_cast<void*>(my_malloc);
    args.mem_alloc_option.allocator.realloc_func = reinterpret_cast<void*>(my_realloc);
    args.mem_alloc_option.allocator.free_func = reinterpret_cast<void*>(my_free);

    // ...
    auto engine = wasm_engine_new_with_args(args.mem_alloc_type, &args.mem_alloc_option);
```

core/iwasm/include/mem_alloc_export.h
loganek: Yes, I meant the build option, something like:
```c
typedef union MemAllocOption {
    struct {
        void *heap_buf;
        uint32_t heap_size;
    } pool;
    struct {
        void *malloc_func;
        void *realloc_func;
        void *free_func;
#ifdef MEM_ALLOC_WITH_USER_DATA
        void *user_data;
#endif
    } allocator;
} MemAllocOption;
```

loganek:
@wenyongh I agree better to not modify the API, and actually in its current design, where the engine itself and allocation functions are singletons, there's no much benefit of having user data, because it can be a global variable instead (like you suggested in your comment).

Having said that, I wonder if we should start working towards having a concept of contexts instead of singletons/global objects so it's possible to embed multiple VMs in a single process? For example:
- CPython follows similar approach and only allows having a single interpreter instance per process (although there's a lot of differences between that and WAMR, and CPython)
- LUA has `lua_State` which holds all the information about the execution environment (including allocation methods) so we can have multiple scripts running independently in a single process

I presume moving away from global objects would require much larger refactoring and probably needs more through analysis; also, I'm aware I don't know have all the data behind decisions that were made in WAMR a long time ago, and there might be good reasons to keep it as it is. I've added this topic to the next TSC meeting agenda to discuss this topic with the others and gather their feedback.

nchamzn:
The suggestion is how we have implemented this at the moment but I see it as unnecessary - not passing user data forces the embedding application to use a singleton/static even though it shouldn't be required.

I already have a singleton `Engine` class and the allocator should have the same lifetime, as it does in Wamr (See: `wasm_runtime_memory_init`/`wasm_runtime_memory_destroy`), but to ensure my own allocator has the same lifetime I need to add functions to mutate this singleton from my Engine constuctor/destructor whereas in an ideal world it would just be a member of the `Engine` class:

```
class Engine
{
  Engine& get_instance();

  AllocStats get_alloc_stats();
private:
  Allocator _allocator;
};
```

For what it's worth, this pattern is used by all other Vms I've embedded before:
* [quickjs](https://github.com/bellard/quickjs/blob/master/quickjs.h#L319)
* [duktape](https://github.com/svaarala/duktape/blob/4a07a8be2e820cd5c33c0d4fbe8fdc155eb69e81/src-input/duk_api_heap.c#L20)
* [lua](https://github.com/lua/lua/blob/be908a7d4d8130264ad67c5789169769f824c5d1/lua.h#L153)

wenyongh:
> @wenyongh I agree better to not modify the API, and actually in its current design, where the engine itself and allocation functions are singletons, there's no much benefit of having user data, because it can be a global variable instead (like you suggested in your comment).
> 
> Having said that, I wonder if we should start working towards having a concept of contexts instead of singletons/global objects so it's possible to embed multiple VMs in a single process? For example:
> 
> * CPython follows similar approach and only allows having a single interpreter instance per process (although there's a lot of differences between that and WAMR, and CPython)
> * LUA has `lua_State` which holds all the information about the execution environment (including allocation methods) so we can have multiple scripts running independently in a single process
> 
> I presume moving away from global objects would require much larger refactoring and probably needs more through analysis; also, I'm aware I don't know have all the data behind decisions that were made in WAMR a long time ago, and there might be good reasons to keep it as it is. I've added this topic to the next TSC meeting agenda to discuss this topic with the others and gather their feedback.

@loganek Per my understanding, you mainly idea is to create runtime/vm context multiple times to eliminate the singletons/global objects in wasm_memory.c and some other places, and for each wasm_runtime_init/full_init, we return a runtime context instead, right? By this way we can custom the individual behavior for each runtime context. My concern is that this would greatly change the runtime APIs and cause incompatibility with old versions, for example, for wasm_runtime_load/wasm_runtime_instantiate and some other APIs, we should add runtime context as a new argument. And not sure whether it is necessary? Since we have already supported created multiple wasm module/module_inst in a thread.

Or do we just want to support using different memory allocator for a module/module_inst? That seems easier but should also require lots of changes.

loganek:
> @loganek Per my understanding, your mainly idea is to create runtime/vm context multiple times to eliminate the singletons/global objects in wasm_memory.c and some other places, and for each wasm_runtime_init/full_init, we return a runtime context instead, right? By this way we can custom the individual behavior for each runtime context. My concern is that this would greatly change the runtime APIs and cause incompatibility with old versions, for example, for wasm_runtime_load/wasm_runtime_instantiate and some other APIs, we should add runtime context as a new argument. And not sure whether it is necessary? Since we have already supported created multiple wasm module/module_inst in a thread.

Yes, I thought of having a context per runtime. And yes, I agree that will be a lot of changes, however, none of that should at least affect `wasm_c_api.h` changes, although there need to be changes in `wasm_export.h`. I think we don't have to modify the API, but instead, we can extend it by adding a new function (e.g. with `_ex` or `_with_ctx` suffix) for every function that assumes the singleton. 

> Or do we just want to support using different memory allocator for a module/module_inst? That seems easier but should also require lots of changes.

No, I didn't think of having per-module allocator, although if there's a need for that, and it doesn't cause any risk, we might consider that in the design as well.

As I said, I think this is a complex change and we should discuss it before jumping into it.

core/iwasm/include/mem_alloc_export.h
```diff
@@ -0,0 +1,34 @@
...
+        void *heap_buf;
+        uint32_t heap_size;
+    } pool;
+    struct {
+        void *malloc_func;
+        void *realloc_func;
+        void *free_func;
+#ifdef MEM_ALLOC_WITH_USER_DATA
+        void *user_data;
+#endif
```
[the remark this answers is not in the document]
nchamzn: Yeah, I mentioned this [here](https://github.com/bytecodealliance/wasm-micro-runtime/pull/1765#discussion_r1034548985).  If you are happy with this being always defined then that is much easier.  Thanks for the link, that helps

core/iwasm/include/mem_alloc_export.h
```diff
@@ -0,0 +1,34 @@
...
+    } allocator;
+} MemAllocOption;
```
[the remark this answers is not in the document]
nchamzn: I don't see why you would want this duplicated?  Happy to revert but seems an easy change/fix for that issue.

core/iwasm/include/mem_alloc_export.h
loganek: Good point, didn't realize that.

core/iwasm/common/wasm_c_api.c
```diff
@@ -340,6 +340,10 @@ wasm_engine_new_internal(mem_alloc_type_t type, const MemAllocOption *opts)
             opts->allocator.free_func;
         init_args.mem_alloc_option.allocator.realloc_func =
             opts->allocator.realloc_func;
+#ifdef MEM_ALLOC_WITH_USER_DATA
```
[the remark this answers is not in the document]
loganek: Just out of curiosity, is there any reason (other than historical) for that pattern? I saw it's more common to use `ifndef/ifdef` pattern and the config file have all possible options listed, but commented if not used.

core/iwasm/include/mem_alloc_export.h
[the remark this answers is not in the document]
wenyongh: Had better not to introduce a new file for that, for the wasm-c-api spec proposal, developer can just copy the header file wasm_c_api.h and include it. And for using runtime export APIs, developer can just copy wasm_export.h/lib_export.h and use them. Had better not to make big changes now.

nchamzn:
Had to make some more changes because when the build definition is set, the function type is different which leads to some issues when the build definition is set and the user decides to use the system allocator.

core/iwasm/include/mem_alloc_export.h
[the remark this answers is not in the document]
nchamzn: OK, reverted that and modified in 2 places.

core/iwasm/common/wasm_memory.c
```diff
@@ -11,16 +11,25 @@
 typedef enum Memory_Mode {
     MEMORY_MODE_UNKNOWN = 0,
     MEMORY_MODE_POOL,
+    MEMORY_MODE_SYSTEM_ALLOCATOR,
```
wenyongh, line 14: How about moving `MEMORY_MODE_SYSTEM_ALLOCATOR` after `MEMORY_MODE_ALLOCATOR`, same as order in wasm_export.h?

core/iwasm/common/wasm_memory.c
```diff
@@ -63,12 +91,22 @@ wasm_runtime_memory_init(mem_alloc_type_t mem_alloc_type,
         return wasm_memory_init_with_pool(alloc_option->pool.heap_buf,
                                           alloc_option->pool.heap_size);
     else if (mem_alloc_type == Alloc_With_Allocator)
+#if WASM_MEM_ALLOC_WITH_USER_DATA != 0
+        return wasm_memory_init_with_allocator_with_user_data(
+            alloc_option->allocator.user_data,
+            alloc_option->allocator.malloc_func,
+            alloc_option->allocator.realloc_func,
+            alloc_option->allocator.free_func);
+#else
```
wenyongh, line 100: Code is not easy to read, had better add `{ .. }` to wrap them:
```C
    else if (mem_alloc_type == Alloc_With_Allocator) {
#if WASM_MEM_ALLOC_WITH_USER_DATA != 0
        return wasm_memory_init_with_allocator_with_user_data(
            alloc_option->allocator.user_data,
            alloc_option->allocator.malloc_func,
            alloc_option->allocator.realloc_func,
            alloc_option->allocator.free_func);
#else
        return wasm_memory_init_with_allocator(
            alloc_option->allocator.malloc_func,
            alloc_option->allocator.realloc_func,
            alloc_option->allocator.free_func);
#endif
    }
```

core/iwasm/common/wasm_memory.c
nchamzn: Will also reorder the if/else chains to match

core/iwasm/common/wasm_c_api.c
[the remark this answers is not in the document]
wenyongh: Yes, the `ifndef/ifdef` pattern is also very common, but it may be not very convenience to use the config file: if it isn't created and all the configurations are controlled by makefile, it may be not easy to learn all the configs in the makefile. And if the config header file is used, (1) if it is auto generated e.g. autoconf.h, it may be under the output build folder and not easy to find it, (2) if it is one file of the project and all the configs are listed, then as you mentioned, if we want to enable/disable one config, we may need to uncomment/comment the macro manually or by makefile, just not so convenience as `#if xxx != 0`/`#if xxx == 0` pattern with which normally we don't need to modify the config file.

core/iwasm/common/wasm_memory.c
wenyongh: Yes, it really makes confusing, in my personal opinion, it is recommended to always use the braces, at least for the code snippet with several lines.

core/iwasm/common/wasm_memory.c
```diff
@@ -54,23 +81,37 @@ wasm_memory_init_with_allocator(void *_malloc_func, void *_realloc_func,
...
 {
-    if (mem_alloc_type == Alloc_With_Pool)
+    if (mem_alloc_type == Alloc_With_Pool) {
         return wasm_memory_init_with_pool(alloc_option->pool.heap_buf,
                                           alloc_option->pool.heap_size);
-    else if (mem_alloc_type == Alloc_With_Allocator)
+    }
+    else if (mem_alloc_type == Alloc_With_Allocator) {
+#if WASM_MEM_ALLOC_WITH_USER_DATA != 0
+        return wasm_memory_init_with_allocator_with_user_data(
```
wenyongh, line 96: Should be wasm_memory_init_with_allocator

nchamzn: Sorry that was a miss.  Pushed a new version and tested building and running with all combinations of these flags again.

PR #1766, opened 2022-11-29
Initialize/destroy thread_signal_inited multiple times in one thread

lum1n0us:
wasm_c_api brings a senario that both os_thread_signal_init and os_thread_signal_destroy could be called multiple times in one thread. it makes the boolean value false after first calling of os_thread_signal_destroy.


core/iwasm/common/wasm_c_api.c
```diff
@@ -281,24 +288,23 @@ static void
...
             }
-            bh_vector_destroy(&engine->modules);
         }
 
+        bh_vector_destroy(&engine->modules);
+
+#ifndef os_thread_local_attribute
+        bh_vector_destroy(&engine->stores_by_tid);
+#endif
+
```
wenyongh, line 307: The original code will destroy and free the engine->stores:
```C
        if (engine->stores) {
            bh_vector_destroy((Vector *)engine->stores);
            wasm_runtime_free(engine->stores);
        }
```
Wondering why the code is removed? If it isn't used, we should remove all the code related to engine->stores.

lum1n0us: `engine->stores` is not necessary any more. So I intend to remove all

core/iwasm/common/wasm_c_api.c
```diff
@@ -460,9 +532,17 @@ wasm_store_new(wasm_engine_t *engine)
...
+    if (thread_local_stores_num++ == 0)
+#else
+    if (!retrive_thread_local_store_num(&engine->stores_by_tid,
+                                        os_self_thread()))
+#endif
+    {
+        if (!wasm_runtime_init_thread_env()) {
+            LOG_ERROR("init thread environment failed");
+            return NULL;
+        }
```
wenyongh, line 545: Not a better way to wasm_runtime_init_thread_env and increase thread_local_stores_num here:
- when malloc store fails or init store fails below (L550 and L562), we don't decrease thread_local_stores_num and wasm_runtime_destroy_thread_env
- when INIT_VEC(store->modules) and INIT_VEC(store->instances) fail (see INIT_VEC macro), we don't decrease thread_local_stores_num and wasm_runtime_destroy_thread_env

Suggest removing code here, remove L549, init wasm_runtime_init_thread_env and increase thread_local_stores_num only when initialize store successfully, at L565.

core/iwasm/common/wasm_c_api.c
```diff
@@ -449,6 +463,64 @@ wasm_engine_delete(wasm_engine_t *engine)
...
+{
+    unsigned i = 0;
+    thread_local_stores ts = { 0 };
+
+    for (; i < stores_by_tid->num_elems; i++) {
+        bh_assert(bh_vector_get(stores_by_tid, i, &ts));
+
+        if (ts.tid == tid) {
+            ts.stores_num--;
+            bh_assert(bh_vector_set(stores_by_tid, i, &ts));
```
wenyongh, line 515: Should we remove ts from vector when stores_num is 0? Will there the case that stores_num is 0, and then we decrease it again?

core/iwasm/common/wasm_c_api.c
```diff
@@ -511,7 +591,18 @@ wasm_store_delete(wasm_store_t *store)
     }
 
     wasm_runtime_free(store);
-    wasm_runtime_destroy_thread_env();
+
+#ifdef os_thread_local_attribute
+    if (--thread_local_stores_num == 0)
```
wenyongh, line 596: Should we make sure that thread_local_stores_num is larger than 0? Will there be case that thread_local_stores_num is 0 and then we decrease it again?
How about changing to
`if (thread_local_stores_num > 0 && --thread_local_stores_num == 0)`?

lum1n0us: it is an unsigned integer.  `static os_thread_local_attribute unsigned thread_local_stores_num = 0;`

core/iwasm/common/wasm_c_api.c
lum1n0us: `engine->stores_by_tid` is a vector of `struct thread_local_stores` (instead of `therad_local_store*`), so I think it is ok to skip *remove*. Besides, I assume most of cases are about thread pool. If the counter is right, only meets *0* when quitting.
```
#ifndef os_thread_local_attribute
    if (!bh_vector_init(&engine->stores_by_tid, DEFAULT_VECTOR_INIT_SIZE,
                        sizeof(thread_local_stores), true))
        goto failed;
#endif
```

core/iwasm/common/wasm_c_api.c
wenyongh: Yes, I mean if it is decrease to 0, and then wasm_runtime_destroy_thread_env is called, and after that, will it be decreased again? If yes, then it will be decreased to UINT32_MAX, and if wasm_store_new is called again, the wasm_runtime_init_thread_env won't be called since it isn't 0 and error may occur when calling wasm func.

wenyongh: Or had better add bh_assert(thread_local_stores_num > 0) before decrease it?

core/iwasm/common/wasm_c_api.c
wenyongh: OK, how about adding `bh_assert(ts.stores_num > 0);` before `ts.stores_num--;`?

core/iwasm/common/wasm_c_api.c
```diff
@@ -449,25 +464,186 @@ wasm_engine_delete(wasm_engine_t *engine)
...
+#ifndef os_thread_local_attribute
+static bool
+search_thread_local_store_num(Vector *stores_by_tid, korp_tid tid,
+                              thread_local_stores *out_ts, unsigned *out_i)
+{
+    unsigned i;
+
+    for (i = 0; i < stores_by_tid->num_elems; i++) {
+        bool ret = bh_vector_get(stores_by_tid, i, out_ts);
+        bh_assert(ret);
```
wenyongh, line 476: Had better add `(void)ret;`, or compiler may report warning.

PR #1773, opened 2022-12-01
[DO NOT MERGE] Add lldb binary for linux and macos for wamr-ide

NingW101:
- please unzip the `.zip` file before running `wamr-ide`


loganek:
Is it necessary to store those files in git? I'd rather avoid having them in git history because of its size; don't we already publish them as release on Github? @cimacmillan

PR #1781, opened 2022-12-05
Enable bulk memory by default

wenyongh:
Enable bulk memory by default since it is a finished wasm spec proposal
and is enabled by default in latest wasi-sdk. Developer may encounter
"invalid section id" or "unsupported opcode" error if it isn't enabled.


xujuntwt95329 (commented):
LGTM

PR #1638, opened 2022-10-25
wasi-threads: basic simplistic implementation of wasi-threads proposal

loganek:
wasi-threads proposal: https://github.com/WebAssembly/wasi-threads

This is very very basic implementation, probably doesn't cover a lot of corner cases. I open the PR here to get early feedback on the direction, missing bits and suggestions.

Also, I think it'd be good to have a separate branch for that (dev/wasi-threads?) so if we can have it setup that'd be great.

PR #1783, opened 2022-12-05
Create a placeholder for WASI threads implementation

loganek:
This a simpler version of the PR: https://github.com/bytecodealliance/wasm-micro-runtime/pull/1638
To simplify the review, I'll split it into a few smaller changes, starting from this one, which just creates a function stub and a simplistic example.


yamt:
> This a simpler version of the PR: #1638 To simplify the review, I'll split it into a few smaller changes, starting from this one, which just creates a function stub and a simplistic example.

i feel the original version is simple enough.

loganek:
@yamt it is indeed, but it's also a bit incorrect, and I'm not happy about the quality of that at all - I only made a PR to save my work and give others opportunity to provide higher-level feedback; I actually never intended to merge it as-it-is. As I came back to that work, I thought I'd use it as opportunity to split it into two separate PRs, where the first one is just to provide a template, and the second will be about the actual implementation (with some TODOs, which we address later on - that's why the development will happen on the `dev/wasi_threads` branch until we're fully ready). Having said that, many thanks for all your comments in the previous PR, I'll take them into account.

samples/wasi-threads/wasm-apps/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
+
```
wenyongh, line 1: Should add the license header

samples/wasi-threads/wasm-apps/no_pthread.c
```diff
@@ -0,0 +1,51 @@
+#ifndef __wasi__
```
wenyongh, line 1: Should add the license header.

wenyongh (approved):
LGTM

PR #1787, opened 2022-12-07
Fix wasm-c-api import func link issue in wasm_instance_new

wenyongh:
When a wasm module is duplicated instantiated with wasm_instance_new,
the function import info of the previous instantiation may be overwritten by
the later instantiation, which may cause unexpected behavior.

Store the function import info into the module instance to fix the issue.


lum1n0us:
LGTM. I've uploaded an updated PR in envoy and see how it goes.

PR #1793, opened 2022-12-08
Enable XIP in daily test

no1wudi:


no1wudi:
This patch depends on #1792

Issue #1790, opened 2022-12-07
WASI threads support


loganek:
I create this issue to track implementation of the [`WASI threads`](https://github.com/WebAssembly/wasi-threads) proposal, define technical approach and give opportunity for discussion.

## High-level approach
The high-level approach for implementing `WASI threads` is to create a new module instance for every new thread and only share the memory between created instances.

## API
As of today there's a single API that has to be added to the host:
```c
int32_t thread_spawn(void * start_arg)
```
where `start_arg` is an opaque pointer and its value should not be inspected by the host environment; instead, it should be passed back to the WASM code. The function return unique non-negative thread identifier integer.
`thread_spawn` instantiates new thread, and must call the following exported function in a new thread once the thread is ready:
```c
void wasi_thread_start(int thread_id, void *start_arg);
```
where the `start_arg` is exactly the same value as the one being passed to the `thread_spawn`, and `thread_id` is the same thread ID as the one returned from `thread_spawn`.

## Project structure
We considered two approaches:
 * implement proposal in `libc-wasi`, given it is already part of wasi sdk
 * create a new library (in `iwasm/core/libraries`) that includes the implementation

The main benefit of the former approach is consistency with other APIs; however, we decided to go for the latter approach due to the following reasons:
* `WASI threads` is still in progress and some of the concepts haven't been fully clarified yet; having it as a separate library help emphasizing the instability of the module to potential users
* conditional compilation of the module won't require too many `ifdef`'s here and there as the whole implementation will live in a separate place; that will enable us to have cleaner and more maintainable code

## Compatibility
WAMR already allows for spawning new threads, either using [pthread library](https://github.com/bytecodealliance/wasm-micro-runtime/blob/c3d66f916ef8093e5c8cacf3329ed968f807cf58/doc/pthread_library.md) or [wasm_runtime_spawn_thread](https://github.com/bytecodealliance/wasm-micro-runtime/blob/d974452a6d17a6a802149d32509c2aa9e151bdea/core/iwasm/common/wasm_runtime_common.c#L4269).

Our goal as part of wasi threads support is to provide a replacement for pthread library (through wasi-libc) therefore we won't guarantee compatibility with pthread library (i.e. both wasi threads and pthread library can not be used at the same time). If we see that using both implementations at the same time might actually be harmful, we might even restrict enabling both features at the same time at build time.
[...]

PR #1799, opened 2022-12-08
Enable aux stack allocations on application heap

loganek:
This is necessary for WASI threads as the aux stack should be managed by the application. See https://github.com/bytecodealliance/wasm-micro-runtime/issues/1790 for details.


core/iwasm/libraries/lib-pthread/lib_pthread_wrapper.c
```diff
@@ -527,9 +527,6 @@ pthread_start_routine(void *arg)
     /* destroy pthread key values */
     call_key_destructor(exec_env);
 
-    /* routine exit, destroy instance */
-    wasm_runtime_deinstantiate_internal(module_inst, true);
-
```
yamt, line 532: how is this related to the rest of the PR?

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -77,8 +77,18 @@ traverse_list(bh_list *l, list_visitor visitor, void *user_data)
...
 static bool
-allocate_aux_stack(WASMCluster *cluster, uint32 *start, uint32 *size)
+allocate_aux_stack(WASMExecEnv *exec_env, uint32 *start, uint32 *size)
 {
+    WASMCluster *cluster = wasm_exec_env_get_cluster(exec_env);
+#if WASM_ENABLE_HEAP_AUX_STACK_ALLOCATION == 1
+    WASMModuleInstanceCommon *module_inst =
+        wasm_exec_env_get_module_inst(exec_env);
+
+    *start = wasm_runtime_module_malloc(module_inst, cluster->stack_size, NULL);
```
yamt, line 87: i'm not sure what's the intention of this.
for wasi-threads, we will disable this allocation, right?

loganek: Yes, it will be disabled for wasi threads, but even with that, we need to keep `wasm_cluster_spawn_exec_env()` working; aux stack allocation is needed for that.

core/iwasm/libraries/lib-pthread/lib_pthread_wrapper.c
loganek: If `WASM_ENABLE_HEAP_AUX_STACK_ALLOCATION` is enabled, module instance can't be deinstantiate untill the memory allocated for the aux stack is released.

yamt: * with this change, where will it be deinstantiated instead?
* how about the other wasm_runtime_deinstantiate_internal call in this function? (L507)

core/iwasm/libraries/thread-mgr/thread_manager.c
yamt: ok

core/iwasm/libraries/lib-pthread/lib_pthread_wrapper.c
loganek: > with this change, where will it be deinstantiated instead?
It's deinstantiated in the `thread_manager_start_routine()`

> how about the other wasm_runtime_deinstantiate_internal call in this function? (L507)
Yes, this one, and one in the `lib_wasi_threads_wrapper.c` have to be removed. Already updated the code, thanks.

yamt: do you mean already updated locally but not pushed yet?

loganek: Yes sorry, didn't push my change. Should be updated now.

core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads.cmake
```diff
@@ -3,7 +3,7 @@
 
 set (LIB_WASI_THREADS_DIR ${CMAKE_CURRENT_LIST_DIR})
 
-add_definitions (-DWASM_ENABLE_LIB_WASI_THREADS=1)
+add_definitions (-DWASM_ENABLE_LIB_WASI_THREADS=1 -DWASM_ENABLE_HEAP_AUX_STACK_ALLOCATION)
```
wenyongh, line 6: Should be `-DWASM_ENABLE_HEAP_AUX_STACK_ALLOCATION=1`

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -77,8 +77,18 @@ traverse_list(bh_list *l, list_visitor visitor, void *user_data)
 }
 
 static bool
-allocate_aux_stack(WASMCluster *cluster, uint32 *start, uint32 *size)
+allocate_aux_stack(WASMExecEnv *exec_env, uint32 *start, uint32 *size)
 {
+    WASMCluster *cluster = wasm_exec_env_get_cluster(exec_env);
+#if WASM_ENABLE_HEAP_AUX_STACK_ALLOCATION == 1
```
wenyongh, line 83: Had better `#if WASM_ENABLE_HEAP_AUX_STACK_ALLOCATION != 0`

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -100,11 +110,21 @@ allocate_aux_stack(WASMCluster *cluster, uint32 *start, uint32 *size)
...
     os_mutex_unlock(&cluster->lock);
     return false;
+#endif
 }
 
 static bool
-free_aux_stack(WASMCluster *cluster, uint32 start)
+free_aux_stack(WASMExecEnv *exec_env, uint32 start)
 {
+#if WASM_ENABLE_HEAP_AUX_STACK_ALLOCATION == 1
```
wenyongh, line 119: Same as above

PR #1803, opened 2022-12-12
64-bit const value accompany with __ignore in lookup table for 32-bit target.

dongsheng28849455:
1, i64.const should be handled.
2, xtensa platform is 32-bit target, we need register for it also. however, to avoid hard coded for each 32-bit target, just using comp_ctx->pointer_size == sizeof(uint32) as the only condition.


no1wudi (approved):
LGTM

PR #1806, opened 2022-12-12
Fix watchpoint segfault when using debug interpreter without server

cimacmillan:
### Summary

In a [previous PR](https://github.com/bytecodealliance/wasm-micro-runtime/pull/1762) I had introduced memory watchpoint support. This adds two linked lists containing watchpoint locations that are checked on variable reads/writes. However, there is an edge-case where one could be using the `WASM_ENABLE_DEBUG_INTERP`, but not within a debugging instance. 

In this, the interpreter attempts to read the debugging instance lists which are not initialised. This results in a segmentation fault in `bh_list_first_elem`

This change checks whether the interpreter is running within a debugging instance. If not, it assigns the list pointers to null and `bh_list_first_elem` handles this gracefully

### DIFF
--- a/core/iwasm/interpreter/wasm_interp_classic.c
+++ b/core/iwasm/interpreter/wasm_interp_classic.c
@@ -1157,8 +1157,10 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
 #if WASM_ENABLE_DEBUG_INTERP != 0
     uint8 *frame_ip_orig = NULL;
     WASMDebugInstance *debug_instance = wasm_exec_env_get_instance(exec_env);
-    bh_list *watch_point_list_read = &debug_instance->watch_point_list_read;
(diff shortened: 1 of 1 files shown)


cimacmillan:
@wenyongh @lum1n0us Can we either include this in the WAMR 1.1.2 release, or omit the previous PR from the release until this is merged? https://github.com/bytecodealliance/wasm-micro-runtime/pull/1762

wenyongh:
> @wenyongh @lum1n0us Can we either include this in the WAMR 1.1.2 release, or omit the previous PR from the release until this is merged? #1762

Sure, let's merge this PR before release 1.1.2.

wenyongh:
@cimacmillan, @loganek The WAMR-1.1.2 was created:
 https://github.com/bytecodealliance/wasm-micro-runtime/releases/tag/WAMR-1.1.2

PR #1809, opened 2022-12-13
Improve WASI thread id allocator to reuse identifiers

eloparco:
At the moment, when created, a thread receives an incremental number as an identifier.

This PR allows reusing thread ids once they are released. That is done by using a stack data structure to keep track of the used ids.
When a thread is created, it takes an available identifier from the stack. When the thread exits, it returns the id to the stack of available identifiers.

### DIFF
--- a/core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
+++ b/core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
@@ -18,6 +18,14 @@ static const char *THREAD_START_FUNCTION = "wasi_thread_start";
 
 static korp_mutex thread_id_lock;
 
+// Stack data structure to track available thread identifiers
+#define AVAIL_TIDS_INIT_SIZE CLUSTER_MAX_THREAD_NUM
+typedef struct {
+    int32 *ids;
+    uint32 pos, size;
+} AvailableThreadIds;
+static AvailableThreadIds avail_tids;
+
 typedef struct {
     /* app's entry function */
     wasm_function_inst_t start_func;
@@ -30,14 +38,54 @@ typedef struct {
 static int32
 allocate_thread_id()
 {
-    static int32 thread_id = 0;
-
-    int32 id;
+    int32 id = -1;
 
     os_mutex_lock(&thread_id_lock);
-    id = thread_id++;
+    if (avail_tids.pos == 0) { // Resize stack and push new thread ids
+        uint32 old_size = avail_tids.size;
+        uint32 new_size = avail_tids.size * 2;
+        if (new_size / 2 != avail_tids.size) {
+            LOG_ERROR("Overflow detected during new size calculation");
+            goto return_id;
+        }
+
+        size_t realloc_size = new_size * sizeof(int32);
+        if (realloc_size / sizeof(int32) != new_size) {
+            LOG_ERROR("Overflow detected during realloc");
+            goto return_id;
+        }
+        int32 *tmp =
+            (int32 *)wasm_runtime_realloc(avail_tids.ids, realloc_size);
+        if (tmp == NULL) {
+            LOG_ERROR("Thread ID allocator realloc failed");
+            goto return_id;
+        }
+
+        avail_tids.size = new_size;
+        avail_tids.pos = old_size;
+        avail_tids.ids = tmp;
+        for (uint32 i = 0; i < old_size; i++)
+            avail_tids.ids[i] = new_size - i;
+    }
+
+    // Pop available thread identifier from `avail_tids` stack
+    id = avail_tids.ids[--avail_tids.pos];
+
+return_id:
     os_mutex_unlock(&thread_id_lock);
     return id;
+}
+
+void
+deallocate_thread_id(int32 thread_id)
+{
+    os_mutex_lock(&thread_id_lock);
+
+    // Release thread identifier by pushing it into `avail_tids` stack
+    bh_assert(avail_tids.pos < avail_tids.size);
+    avail_tids.ids[avail_tids.pos++] = thread_id;
+
+    os_mutex_unlock(&thread_id_lock);
 }
 
 static void *
@@ -57,6 +105,8 @@ thread_start(void *arg)
             wasm_cluster_spread_exception(exec_env);
     }
 
+    // Routine exit
+    deallocate_thread_id(thread_arg->thread_id);
     wasm_runtime_free(thread_arg);
     exec_env->thread_arg = NULL;
 
@@ -101,15 +151,19 @@ thread_spawn_wrapper(wasm_exec_env_t exec_env, uint32 start_arg)
     if (!start_func) {
         LOG_ERROR("Failed to find thread start function %s",
                   THREAD_START_FUNCTION);
-        goto thread_spawn_fail;
+        goto thread_preparation_fail;
     }
 
     if (!(thread_start_arg = wasm_runtime_malloc(sizeof(ThreadStartArg)))) {
         LOG_ERROR("Runtime args allocation failed");
-        goto thread_spawn_fail;
+        goto thread_preparation_fail;
     }
 
     thread_start_arg->thread_id = thread_id = allocate_thread_id();
+    if (thread_id < 0) {
+        LOG_ERROR("Failed to get thread identifier");
+        goto thread_preparation_fail;
+    }
     thread_start_arg->arg = start_arg;
     thread_start_arg->start_func = start_func;
 
@@ -117,7 +171,6 @@ thread_spawn_wrapper(wasm_exec_env_t exec_env, uint32 start_arg)
     ret = wasm_cluster_create_thread(exec_env, new_module_inst, thread_start,
                                      thread_start_arg);
     if (ret != 0) {
-        os_mutex_unlock(&exec_env->wait_lock);
         LOG_ERROR("Failed to spawn a new thread");
         goto thread_spawn_fail;
     }
@@ -126,9 +179,12 @@ thread_spawn_wrapper(wasm_exec_env_t exec_env, uint32 start_arg)
     return thread_id;
 
 thread_spawn_fail:
+    os_mutex_unlock(&exec_env->wait_lock);
+    deallocate_thread_id(thread_id);
+
+thread_preparation_fail:
     if (new_module_inst)
         wasm_runtime_deinstantiate_internal(new_module_inst, true);
-
     if (thread_start_arg)
         wasm_runtime_free(thread_start_arg);
 
@@ -156,11 +212,25 @@ lib_wasi_threads_init(void)
     if (0 != os_mutex_init(&thread_id_lock))
         return false;
 
+    // Initialize stack to store thread identifiers
+    avail_tids.size = AVAIL_TIDS_INIT_SIZE;
+    avail_tids.pos = avail_tids.size;
+    avail_tids.ids =
+        (int32 *)wasm_runtime_malloc(avail_tids.size * sizeof(int32));
+    if (avail_tids.ids == NULL) {
+        os_mutex_destroy(&thread_id_lock);
+        return false;
+    }
+    for (uint32 i = 0; i < avail_tids.size; i++)
+        avail_tids.ids[i] = avail_tids.size - i;
+
     return true;
 }
 
 void
 lib_wasi_threads_destroy(void)
 {
+    wasm_runtime_free(avail_tids.ids);
+    avail_tids.ids = NULL;
     os_mutex_destroy(&thread_id_lock);
 }


samples/wasi-threads/wasm-apps/CMakeLists.txt
```diff
@@ -1,6 +1,12 @@
 # Copyright (C) 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+if (APPLE)
+    set (HAVE_FLAG_SEARCH_PATHS_FIRST 0)
+    set (CMAKE_C_LINK_FLAGS "")
+    set (CMAKE_CXX_LINK_FLAGS "")
+endif ()
```
yamt, line 8: why?

core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
```diff
@@ -31,16 +39,43 @@ typedef struct {
...
-    static int32 thread_id = 0;
-
     int32 id;
 
     os_mutex_lock(&thread_id_lock);
-    id = thread_id++;
+    if (avail_tids.pos < 0) {
+        // Resize stack and push new thread ids
+        uint32 old_size = avail_tids.size;
+        avail_tids.size *= 2;
```
yamt, line 48: check overflow?

core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
```diff
@@ -31,16 +39,43 @@ typedef struct {
...
+        avail_tids.pos = old_size - 1;
+        avail_tids.ids = (int32 *)wasm_runtime_realloc(
+            avail_tids.ids, avail_tids.size * sizeof(int32));
```
yamt, line 51: check overflow?

core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
```diff
@@ -160,11 +201,25 @@ lib_wasi_threads_init(void)
...
+    avail_tids.size = AVAIL_TIDS_INIT_SIZE;
+    avail_tids.pos = avail_tids.size - 1;
+    avail_tids.ids =
+        (int32 *)wasm_runtime_malloc(avail_tids.size * sizeof(int32));
+    if (avail_tids.ids == NULL) {
+        os_mutex_destroy(&thread_id_lock);
+        return false;
+    }
+    for (int32 i = 0; i < avail_tids.size; i++)
+        avail_tids.ids[i] = avail_tids.size - 1 - i;
```
yamt, line 214: please do not use TID 0, which has a special meaning in wasi-libc.

samples/wasi-threads/wasm-apps/CMakeLists.txt
eloparco: it's the same as in the others `samples/<sample-name>/wasm-apps/CMakeLists.txt` files; in particular, without the `set (CMAKE_C_LINK_FLAGS "")` I get linking errors when compiling on a macbook:
```
wasm-ld: error: unknown argument: -search_paths_first
wasm-ld: error: unknown argument: -headerpad_max_install_names
```
I put the entire `if(APPLE)/endif()` block (and not only the `CMAKE_C_LINK_FLAGS` part) for consistency with the other samples

yamt: ok. thank you for explanation.

i think it's cleaner to use ExternalProject as some of other samples do. but i guess it's ok for this PR.

core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
eloparco: right, I'll update it to
```c
if (avail_tids.size * 2 < avail_tids.size) { // Overflow check
    os_mutex_unlock(&thread_id_lock);
    return -1;
}
avail_tids.size *= 2;
```

core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
eloparco: similar here
```
size_t realloc_size = avail_tids.size * sizeof(int32);
if (realloc_size / sizeof(int32) != avail_tids.size) { ... }
```

core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
loganek: Also, worth adding error message in spawn_thread so it's clear why creating a new thread couldn't be completed.

samples/wasi-threads/wasm-apps/CMakeLists.txt
loganek: I'm not a big fan of external project because it doesn't automatically recompile sources from root on change. Maybe we can come up with a generic solution for all the samples, but that feels like outside of the scope for this pr

yamt: i'm not a fan of ExternalProject either. but it's the cleanest way i'm aware of.

either ways, i agree it's nice to be consistent and it's out of the scope of this PR.

core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
```diff
@@ -122,6 +162,7 @@ thread_spawn_wrapper(wasm_exec_env_t exec_env, uint32 start_arg)
                                      thread_start_arg);
     if (ret != 0) {
         os_mutex_unlock(&exec_env->wait_lock);
+        deallocate_thread_id(thread_id);
```
loganek, line 165: Maybe move it under the thread_spawn_fail label in case we'll have more fail branches after this one in the future? And even if not, I think it makes sense to have error handling in one place.

core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
yamt: if you want to use them, i guess you need to introduce macros with plain C fallback.

core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
```diff
@@ -31,16 +39,43 @@ typedef struct {
...
     int32 id;
 
     os_mutex_lock(&thread_id_lock);
-    id = thread_id++;
+    if (avail_tids.pos < 0) {
+        // Resize stack and push new thread ids
+        uint32 old_size = avail_tids.size;
+        avail_tids.size *= 2;
+        avail_tids.pos = old_size - 1;
+        avail_tids.ids = (int32 *)wasm_runtime_realloc(
```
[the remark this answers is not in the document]
eloparco: good advice!

core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
```diff
@@ -31,14 +40,57 @@ typedef struct {
...
+        avail_tids.ids = tmp;
+
+        for (uint32 i = 0; i < old_size; i++)
+            avail_tids.ids[i] = avail_tids.size - i;
+
+        avail_tids.pos = old_size - 1;
+    }
+
+    // Pop available thread identifier from `avail_tids` stack
+    id = avail_tids.ids[avail_tids.pos--];
```
loganek, line 75: [minor] if you initialize I'd with -1 you can skip duplication, just rename the resize_fail label to e.g. final.

loganek (approved):
Lgtm

eloparco:
I'm doing some additional refactoring to have both `pos` and `size` as `unsigned` so that comparisons are safe. I just need to get `pos` to point to the first empty position in the stack rather than at the top element.

core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
```diff
@@ -114,21 +164,28 @@ thread_spawn_wrapper(wasm_exec_env_t exec_env, uint32 start_arg)
...
-        os_mutex_unlock(&exec_env->wait_lock);
         LOG_ERROR("Failed to spawn a new thread");
-        goto thread_spawn_fail;
+        goto tid_allocation_fail;
     }
     os_mutex_unlock(&exec_env->wait_lock);
 
     return thread_id;
 
+tid_allocation_fail:
```
yamt, line 185: i feel this label is a bit confusing because it's actually used when tid allocation succeeded.

core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
```diff
@@ -31,16 +39,57 @@ typedef struct {
...
+        avail_tids.size *= 2;
+
+        size_t realloc_size = avail_tids.size * sizeof(int32);
+        if (realloc_size / sizeof(int32) != avail_tids.size) {
+            LOG_ERROR("Overflow detected during realloc");
+            goto return_id;
+        }
+        int32 *tmp =
+            (int32 *)wasm_runtime_realloc(avail_tids.ids, realloc_size);
+        if (avail_tids.ids == NULL) {
```
yamt, line 61: tmp == NULL

core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
```diff
@@ -31,16 +39,57 @@ typedef struct {
...
     os_mutex_lock(&thread_id_lock);
-    id = thread_id++;
+    if (avail_tids.pos == 0) { // Resize stack and push new thread ids
+        uint32 old_size = avail_tids.size;
+        uint32 new_size = avail_tids.size * 2;
+        if (new_size / 2 != avail_tids.size) {
+            LOG_ERROR("Overflow detected during new size calculation");
+            goto return_id;
+        }
+        avail_tids.size *= 2;
```
yamt, line 52: move this to below because it can still fail.

core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
```diff
@@ -31,16 +39,57 @@ typedef struct {
...
+            goto return_id;
+        }
+        int32 *tmp =
+            (int32 *)wasm_runtime_realloc(avail_tids.ids, realloc_size);
+        if (avail_tids.ids == NULL) {
+            LOG_ERROR("Thread ID allocator realloc failed");
+            goto return_id;
+        }
+        avail_tids.ids = tmp;
+
```
yamt, line 66: avail_tids.size = new_size;

core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
```diff
@@ -31,16 +39,57 @@ typedef struct {
...
-
-    int32 id;
+    int32 id = -1;
 
     os_mutex_lock(&thread_id_lock);
-    id = thread_id++;
+    if (avail_tids.pos == 0) { // Resize stack and push new thread ids
+        uint32 old_size = avail_tids.size;
+        uint32 new_size = avail_tids.size * 2;
+        if (new_size / 2 != avail_tids.size) {
```
loganek, line 48: [minor] you might just compare `old_size` with the `new_size` to avoid additional division.

core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
```diff
@@ -31,16 +39,57 @@ typedef struct {
...
     return id;
 }
 
+void
+deallocate_thread_id(int32 thread_id)
+{
+    os_mutex_lock(&thread_id_lock);
+
+    // Release thread identifier by pushing it into `avail_tids` stack
+    assert(avail_tids.pos < avail_tids.size);
```
loganek, line 87: use `bh_assert` instead.

core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
```diff
@@ -114,21 +164,28 @@ thread_spawn_wrapper(wasm_exec_env_t exec_env, uint32 start_arg)
     }
 
     thread_start_arg->thread_id = thread_id = allocate_thread_id();
+    if (thread_id < 0) {
+        LOG_ERROR("Failed to get thread identifier");
+        goto thread_spawn_fail;
```
loganek, line 169: the label is a bit confusing; we didn't get to spawning thread yet, so I'd expect we should use a `tid_allocation_fail` label instead (and probably change the implementation of that below).

loganek (changes requested):

core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
```diff
@@ -130,9 +182,12 @@ thread_spawn_wrapper(wasm_exec_env_t exec_env, uint32 start_arg)
     return thread_id;
 
 thread_spawn_fail:
+    os_mutex_unlock(&exec_env->wait_lock);
```
wenyongh, line 185: Not sure why moving `os_mutex_unlock` to here, putting it in the original place (L177) seems to be better?

eloparco: To separate the error handling logic. It's even more useful if in the future we put additional operations between L179 and L180 and we want to release the lock in case of errors.

core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
```diff
@@ -31,16 +39,56 @@ typedef struct {
...
+            goto return_id;
+        }
+
+        size_t realloc_size = new_size * sizeof(int32);
+        if (realloc_size / sizeof(int32) != new_size) {
+            LOG_ERROR("Overflow detected during realloc");
+            goto return_id;
+        }
+        int32 *tmp =
+            (int32 *)wasm_runtime_realloc(avail_tids.ids, realloc_size);
```
yamt, line 59: you don't need a cast

yamt (approved):
looks ok to me

core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
eloparco: Should I get rid of it? I put it (and same for the malloc) for consistency with the rest of the codebase

core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
eloparco: I used the division to make it similar to L54. I'm not sure a simple comparison would work for L54 (maybe when overflowing/wrapping the result can be bigger than the initial value?)

loganek: yes, one way to detect integer overflow here is to check if `new_size < avail_tids.size`. It's not a major callout though.

loganek (approved):

wenyongh (commented):
LGTM

@eloparco Can you resolve the conflicts?

PR #1812, opened 2022-12-14
Exception propagation with WASI threads

eloparco:
As part of the [WASI thread support roadmap](https://github.com/bytecodealliance/wasm-micro-runtime/issues/1790), we want to make sure that exceptions generated in one thread are gracefully propagated to the other threads.
Here, an example of exception propagation is provided. It will later converted into a test when we have a test suite in place.

### DIFF
--- a/samples/wasi-threads/README.md
+++ b/samples/wasi-threads/README.md
@@ -1,1 +1,24 @@
 # "WASI threads" sample introduction
+
+Currently, since the `wasi-sdk` does not have thread support in the latest release, make sure to have [wasi-libc](https://github.com/WebAssembly/wasi-libc) installed. Build it with threads enabled, e.g.
+
+```shell
+make \
+    AR=/opt/wasi-sdk/bin/llvm-ar \
+    NM=/opt/wasi-sdk/bin/llvm-nm \
+    CC=/opt/wasi-sdk/bin/clang \
+    THREAD_MODEL=posix
+```
+
+Build and run the samples
+
+```shell
+$ mkdir build
+$ cd build
+$ cmake -DWASI_SYSROOT=/path/to/wasi-libc/sysroot ..
+$ make
+...
+$ ./iwasm wasm-apps/no_pthread.wasm
+...
+$ ./iwasm wasm-apps/exception_propagation.wasm
+```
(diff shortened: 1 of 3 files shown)


samples/wasi-threads/README.md
```diff
@@ -1 +1,24 @@
...
\ No newline at end of file
+# "WASI threads" sample introduction
+
+Currently, since the `wasi-sdk` does not have thread support in the latest release, make sure to have [wasi-libc](https://github.com/WebAssembly/wasi-libc) installed. Build it with threads enabled, e.g.
+
+```shell
+make \
+    AR=wasm-micro-runtime/core/deps/llvm/build-lldb/bin/llvm-ar \
+    NM=wasm-micro-runtime/core/deps/llvm/build-lldb/bin/llvm-nm \
+    CC=/opt/wasi-sdk/bin/clang \
```
loganek, line 9: Since we use clang from wasi-sdk, can we also for consistency suggest using the other tools from wasi-sdk?

samples/wasi-threads/wasm-apps/exception_propagation.c
```diff
@@ -0,0 +1,86 @@
...
+
+#define NUM_THREADS 3
+static sem_t sem;
+
+void
+run_long_task()
+{
+    // Busy waiting to be interruptible by exception
+    while (true)
+        sleep(1);
```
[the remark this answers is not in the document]
eloparco: Actually, that makes much more sense. In that way, the assertions `assert(false && "Unreachable");` are useful.
Tested by commenting `wasm_cluster_spread_exception`: in that case, there is no exception propagation and the assertion fails.

samples/wasi-threads/wasm-apps/exception_propagation.c
```diff
@@ -0,0 +1,87 @@
...
+
+__attribute__((export_name("wasi_thread_start"))) void
+wasi_thread_start(int thread_id, int *start_arg)
+{
+    bool has_to_throw_exception = (bool)start_arg;
+
+    if (has_to_throw_exception) {
+        // Wait for all other threads (including main thread) to be ready
+        printf("Waiting before throwing exception\n");
+        for (int i = 0; i < NUM_THREADS; i++)
```
loganek, line 37: Should that be `NUM_THREADS - 1`?

loganek (approved):

PR #1815, opened 2022-12-15
vscode-extension: Run the docker image with the same version as WAMR

TianlongLiang:
Modify the VS Code extension, making it run the docker image with the same version of WAMR release, other than fixed version 1.0

### DIFF
--- a/test-tools/wamr-ide/VSCode-Extension/src/extension.ts
+++ b/test-tools/wamr-ide/VSCode-Extension/src/extension.ts
@@ -45,8 +45,9 @@ export async function activate(context: vscode.ExtensionContext) {
         includePathArr = new Array(),
         /* exclude files array used for written into config file */
         excludeFileArr = new Array(),
-        scriptMap = new Map(),
-        wamrVersion = '';
+        scriptMap = new Map();
+
+    const wamrVersion = getWAMRExtensionVersion(context);
 
     /**
      * Get OS platform information for differ windows and linux execution script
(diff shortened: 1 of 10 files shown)


test-tools/wamr-ide/VSCode-Extension/src/extension.ts
```diff
@@ -83,7 +88,9 @@ export async function activate(context: vscode.ExtensionContext) {
     typeMap.set('Debug', 'Debug');
     typeMap.set('Destroy', 'Destroy');
 
-    wasmTaskProvider = new WasmTaskProvider(typeMap, scriptMap);
+    wamrVersion = getWAMRExtensionVersion(context);
```
loganek, line 91: could that be a constant declared here instead of making it `var` in L49?

test-tools/wamr-ide/VSCode-Extension/src/taskProvider.ts
```diff
@@ -40,7 +45,11 @@ export class WasmTaskProvider implements vscode.TaskProvider {
                             : (this._script.get('buildScript') as string),
                     options: {
                         executable: this._script.get('buildScript'),
-                        shellArgs: [targetName, os.platform()],
+                        shellArgs: [
+                            targetName,
+                            this._wamrVersion,
+                            os.platform(),
```
NingW101, line 51: maybe the `os.platform()` can be removed since it is not used by `build scripts (build.bat & build.sh)`

wenyongh: Should be also OK to reserve the argument, maybe we need to use it in the future.

test-tools/wamr-ide/VSCode-Extension/src/extension.ts
TianlongLiang: I guess it's perfectly fine to declare a const here, but I think maybe it's better just to follow the declare convention of other variables:)

PR #1819, opened 2022-12-16
Add CI for WASI threads

eloparco:


eloparco:
Not sure why the `compilation on macos-latest` job is failing, it works properly when I try it directly from my forked repo https://github.com/eloparco/wasm-micro-runtime/actions/runs/3713650979/jobs/6296822730.

.github/workflows/compilation_on_macos.yml
```diff
@@ -210,7 +210,7 @@ jobs:
             #$AOT_BUILD_OPTIONS,
           ]
         os: [macos-latest]
-        wasi_sdk_release: ["https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-12/wasi-sdk-12.0-macos.tar.gz"]
+        wasi_sdk_release: ["https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-17/wasi-sdk-17.0-macos.tar.gz"]
```
loganek, line 213: Wonder if we want to continue building for 12 to ensure backward compatibility? Or do we want to drop 12 and ask users to migrate to >= 17 now?

eloparco: I tried `wasi-sdk` version 12 and `wasi-libc` branch `wasi-sdk-17` [here](https://github.com/eloparco/wasm-micro-runtime/actions/runs/3730554354/jobs/6327794669) but the build failed.
It was not immediate to find a compatible combination of `wasi-sdk` and `wasi-libc` for the different targets (on ubuntu I'm using `wasi-sdk` version 16 in this PR since version 17 was not working).

eloparco:
It looks like all jobs are succeeding now, using `wasi-sdk` version `16` and `wasi-libc` with branch `wasi-sdk-17`  on all targets.

eloparco:
Actually, looking at the output (and at some tests that I was doing locally) I see a regression:
```
New thread ID: 1, starting parameter: 52
Updated value: 8
Thread completed, new value: 0, thread id: 0
```
The example is compiled in release mode so the assert `assert(thread_id == data.thread_id);` doesn't fail, but the value of the thread id is wrong.

wenyongh (commented):
LGTM

@eloparco Shall I merge the PR?

samples/wasi-threads/wasm-apps/CMakeLists.txt
```diff
@@ -29,6 +29,8 @@ function (compile_sample SOURCE_FILE)
     LINKER:--export=__data_end
     LINKER:--shared-memory,--max-memory=1966080
     LINKER:--export=wasi_thread_start
+    LINKER:--export=malloc
+    LINKER:--export=free
```
yamt, line 33: why?

eloparco: Because I was getting this error without the additional exports:
```
[09:52:39:510 - 101150580]: Memory instantiate success.
[09:52:39:614 - 101150580]: Error: app heap is corrupted, if the wasm file is compiled by wasi-sdk-12.0 or higher version, please add -Wl,--export=malloc -Wl,--export=free to export malloc and free functions. If it is compiled by asc, please add --exportRuntime to export the runtime helpers.
[09:52:39:618 - 101150580]: thread manager error: failed to allocate aux stack space for new thread
[09:52:39:620 - 101150580]: Failed to spawn a new thread
[09:52:39:623 - 101150580]: /Users/eloparco/dev/forks/wasm-micro-runtime/core/iwasm/interpreter/wasm_interp_fast.c, line 3959, meet an exception Exception: app heap corrupted
Exception: app heap corrupted
```

eloparco:
@wenyongh let's wait for @loganek's feedback before merging

samples/wasi-threads/wasm-apps/CMakeLists.txt
yamt: the aux stack allocation should be disabled for wasi-threads because it's libc's responsibility to allocate stack for new threads.

no_pthread.c seems broken. its wasi_thread_start should either update the stack pointer global or avoid using C stack.

eloparco: > the aux stack allocation should be disabled for wasi-threads because it's libc's responsibility to allocate stack for new threads.

https://github.com/bytecodealliance/wasm-micro-runtime/blob/dev/wasi_threads/core/config.h#L172

> no_pthread.c seems broken. its wasi_thread_start should either update the stack pointer global or avoid using C stack.

can you elaborate on that and give me some hints on what would need to be changed? my knowledge on the topic is quite limited

tonibofarull: According to [this](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/doc/memory_tune.md),

> app heap and libc heap: the heap to allocate memory for wasm app, note that app heap is created only when the malloc/free functions (or __new/__release functions for AssemblyScript) are not exported and runtime can not detect the libc heap.

eloparco: I've read that one already, but if I don't export malloc and free I'm getting the corruption error

yamt: > > the aux stack allocation should be disabled for wasi-threads because it's libc's responsibility to allocate stack for new threads.
> 
> https://github.com/bytecodealliance/wasm-micro-runtime/blob/dev/wasi_threads/core/config.h#L172

https://github.com/bytecodealliance/wasm-micro-runtime/pull/1799#discussion_r1045454952

> 
> > no_pthread.c seems broken. its wasi_thread_start should either update the stack pointer global or avoid using C stack.
> 
> How do I use the app heap? if I don't export the malloc and free functions I get `Error: app heap is corrupted`

wamr doesn't need to (and shouldn't) use the heap (app heap or libc heap) at least for this purpose (stack allocation for wasi-threads) because the stack is managed inside the module itself. (usually by libc)

.github/workflows/compilation_on_macos.yml
lum1n0us: I am a little confused here. It seems the CI depends on the combination of `wasi-sdk-X` + a specific release of `wasi-libc` now. Is there a plan that a future wasi-sdk release include everything and make the local combination not necessary?

lum1n0us:
BTW, is there an issue to track the compilation error in https://github.com/eloparco/wasm-micro-runtime/actions/runs/3730554354/jobs/6327794669?

wenyongh:
@eloparco Wondering why you use wasi-sdk-16.0 for the CI and use wasi-sdk-17.0 for wasi-thread sample? Why not use wasi-sdk-17.0 for all?
We are upgrading the toolkits (llvm, wasi-sdk, emsdk, binaryen and so on), and are considering upgrade wasi-sdk to 17.0 directly.

eloparco:
> @eloparco Wondering why you use wasi-sdk-16.0 for the CI and use wasi-sdk-17.0 for wasi-thread sample? Why not use wasi-sdk-17.0 for all?

@wenyongh That's what I tried first, but it was failing on ubuntu/android: https://github.com/eloparco/wasm-micro-runtime/actions/runs/3703249930/jobs/6274621343

> BTW, is there an issue to track the compilation error in https://github.com/eloparco/wasm-micro-runtime/actions/runs/3730554354/jobs/6327794669?

@lum1n0us Not yet, should I create one in this repo?

samples/wasi-threads/wasm-apps/CMakeLists.txt
loganek: Threads implementation in WAMR is not fully completed yet. One of the things is to *always* (regardless of compilation flags) use on-heap aux stack allocation. I have change for that already but didn't test it yet; I'll make the change public beginning of next month when I'm back from holidays, until then I'd suggest waiting with merging this PR.

loganek:
> > @eloparco Wondering why you use wasi-sdk-16.0 for the CI and use wasi-sdk-17.0 for wasi-thread sample? Why not use wasi-sdk-17.0 for all?
> 
> @wenyongh That's what I tried first, but it was failing on ubuntu/android: https://github.com/eloparco/wasm-micro-runtime/actions/runs/3703249930/jobs/6274621343

@eloparco are you going to fix the issue? Or should we at least have an issue to upgrade to v17?

eloparco:
@loganek @lum1n0us I'll create an issue to track the upgrade to v17, putting links to the failing jobs

.github/workflows/compilation_on_android_ubuntu.yml
```diff
@@ -323,6 +323,17 @@ jobs:
           sudo tar -xzf wabt-1.0.24-*.tar.gz
           sudo mv wabt-1.0.24 wabt
 
+      - name: build wasi-libc (needed for wasi-threads)
+        run: |
+          git clone --branch wasi-sdk-17 https://github.com/WebAssembly/wasi-libc
```
wenyongh, line 328: Branch wasi-sdk-17 isn't found in the wasi-libc repo, seems that main branch is OK.

.github/workflows/compilation_on_macos.yml
```diff
@@ -266,6 +266,17 @@ jobs:
           sudo tar -xzf wabt-1.0.24-*.tar.gz
           sudo mv wabt-1.0.24 wabt
 
+      - name: build wasi-libc (needed for wasi-threads)
+        run: |
+          git clone --branch wasi-sdk-17 https://github.com/WebAssembly/wasi-libc
```
wenyongh, line 271: Same as above

.github/workflows/compilation_on_sgx.yml
```diff
@@ -268,6 +268,17 @@ jobs:
           sudo tar -xzf wabt-1.0.24-*.tar.gz
           sudo mv wabt-1.0.24 wabt
 
+      - name: build wasi-libc (needed for wasi-threads)
+        run: |
+          git clone --branch wasi-sdk-17 https://github.com/WebAssembly/wasi-libc
```
wenyongh, line 273: Same as above

wenyongh:
> > @eloparco Wondering why you use wasi-sdk-16.0 for the CI and use wasi-sdk-17.0 for wasi-thread sample? Why not use wasi-sdk-17.0 for all?
> 
> @wenyongh That's what I tried first, but it was failing on ubuntu/android: https://github.com/eloparco/wasm-micro-runtime/actions/runs/3703249930/jobs/6274621343
> 

Got it, I tested building wasi-libc and other wasm app with wasi-sdk-17 on Ubuntu-20.04, and got the same errors:
```bash
/opt/wasi-sdk/bin/clang: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /opt/wasi-sdk/bin/clang)
/opt/wasi-sdk/bin/clang: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /opt/wasi-sdk/bin/clang)
/opt/wasi-sdk/bin/clang: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /opt/wasi-sdk/bin/clang)
```
It seems that wasi-sdk-17 requires that libc-2.32 or newer version is installed, while by default in Ubuntu-20.04 libc-2.31 is installed. It may work in Ubuntu-22.04. But it seems that we had better just upgrade scripts to support wasi-sdk-16 now.

.github/workflows/compilation_on_android_ubuntu.yml
eloparco: It's a tag https://github.com/WebAssembly/wasi-libc/tree/wasi-sdk-17. 

I'd prefer to pin the version/tag rather than using main. Unless we want to use the current latest commit from main and we checkout that specific one.

eloparco: I remember that the combination `wasi-sdk-16` + `main` from `wasi-libc` was failing on Ubuntu.

wenyongh: OK. So shall we merge the PR now? @loganek, are there any comments from you? Thanks.

eloparco: I think so, to recap:
- we're using `wasi-sdk-16` because `wasi-sdk-17` would require a newer version of libc wrt the one on Ubuntu-20.04 used in the pipeline (the alternative would be to update the Ubuntu image)
- `--export=malloc` and `--export=free` are used here otherwise the pipeline would fail when executing the `wasi-threads` samples (the alternative would be to comment the assert in `samples/wasi-threads/wasm-apps/no_pthread.c`)

PR #1823, opened 2022-12-20
consistent naming of generated binaries in release CI and update readme

TianlongLiang:
consistent naming of generated binaries in release CI and update readme

### DIFF
--- a/test-tools/wamr-ide/README.md
+++ b/test-tools/wamr-ide/README.md
@@ -20,11 +20,13 @@ them eg. by copying them from Windows.
 Similarly, do not drop execution permission when copying `lldb` binaries
 under `resource/debug/bin`.
 
-#### 1. Install `VSCode` on host.
+### 1. Preparation
+
+#### 1.1. Install `VSCode` on host
 
 - make sure the version of [vscode](https://code.visualstudio.com/Download) you installed is at least _1.59.0_
 
-#### 2. Install `Docker` on host.
+#### 1.2. Install `Docker` on host
 
     1. [Windows: Docker Desktop](https://docs.docker.com/desktop/windows/install/)
     2. [Ubuntu: Docker Engine](https://docs.docker.com/engine/install/ubuntu)
@@ -37,11 +39,13 @@ under `resource/debug/bin`.
        - Ubuntu Bionic 18.04(LTS)
        ```
 
-#### 3. Load docker images from the release tar file or build docker images on the host
+### 2. WAMR VS Code extension: download from the GitHub release or build locally
 
-##### 3.1 Load docker images from the release tar file
+#### 2.1 Download WAMR VS Code extension from the GitHub release(Recommended approach)
 
-From now on, for each release, we have the same version tagged docker image saved as a tar file, which you can find and download in the release.
+##### 2.1.1 Load docker images from the GitHub release tar file
+
+From now on, for each GitHub release, we have the same version tagged docker image saved as a tar file, which you can find and download in the GitHub release.
 
 You could download the tar archive files for docker images from the release, and then load them using the following commands:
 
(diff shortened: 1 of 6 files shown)


test-tools/wamr-ide/README.md
```diff
@@ -68,7 +72,15 @@ docker load --input ./wasm-toolchain.tar
 docker load --input ./wasm-debug-server.tar
 ```
 
-##### 3.2 Build docker images on host
+##### 2.1.2 Download the VS Code extension installation file from the GitHub release
+
+From now on, for each GitHub release, we have the same version tagged zip/tar.gz file(e.g. decompress `wamr_ide-1.1.2.tar.gz` `wamr_ide-1.1.2.zip`), which contains .vsix VS Code extension installation file. You can find and easily download in the GitHub release.
```
wenyongh, line 77: Do we need to change `wamr_ide-1.1.2.tar.gz` `wamr_ide-1.1.2.zip` to `wamr-ide-1.1.2.tar.gz` `wamr-ide-1.1.2.zip`?
Or change to `wamr-ide-1.2.0.tar.gz` `wamr-ide-1.2.0.zip` when we are to release 1.2.0?

test-tools/wamr-ide/README.md
```diff
@@ -68,7 +72,15 @@ docker load --input ./wasm-toolchain.tar
...
+
+#### 2.2 Build WAMR VS Code extension locally(Alternative approach)
+
+You could also build the VS Code extension locally, the following instruction provides a thorough tutorial. It's worth noting that in the local build tutorial we use tag version 1.0 other than the semantic version of WAMR.
```
wenyongh, line 81: Do we use tag version 1.0 currently? I remember we have changed to 1.1.2?

test-tools/wamr-ide/README.md
```diff
@@ -114,15 +126,15 @@ $ docker build --no-cache --build-arg http_proxy=http://proxy.example.com:1234
...
 
-#### If you encounter the problem `failed to solve with frontend dockerfile.v0: failed to create LLB definition`, please config your docker desktop
+##### 2.2.4 If you encounter the problem `failed to solve with frontend dockerfile.v0: failed to create LLB definition`, please config your docker desktop
 
 ![docker-engine-config](./Media/docker_engine_config.png)
 
-#### Points To Remember
+##### 2.2.5 Points To Remember
 
 - Make sure that the `wasm-toolchain:1.0` and `wasm-debug-server:1.0` docker images are both successfully built before using `WAMR IDE`, otherwise `Build`, `Run` and `Debug` will not work.
```
wenyongh, line 135: Do we use `wasm-toolchain:1.0` and `wasm-debug-server:1.0` now?

test-tools/wamr-ide/README.md
```diff
@@ -114,15 +126,15 @@ $ docker build --no-cache --build-arg http_proxy=http://proxy.example.com:1234
...
 
-#### 4. Generate wamride extension package file
+##### 2.2.6 Generate wamride extension package file
 
 `wamride-1.0.0.vsix` can be packaged by [`npm vsce`](https://code.visualstudio.com/api/working-with-extensions/publishing-extension).
```
wenyongh, line 139: Same as above, seems to be `wamr-ide-1.1.2.vsix`?

test-tools/wamr-ide/README.md
TianlongLiang: If the user decides to build locally, the default version is 1.0 inside package.json(unlike inside the CI, we change it dynamically based on the release version), and the local build docker image is also tagged with version 1.0 so that they are consistent.

test-tools/wamr-ide/README.md
TianlongLiang: I will rephrase and be more clear. `wamr-ide-1.1.2.tar.gz` means to be an example. What I am just trying to say is that in the future(for example, there are multiple releases like 1.1.2, 1.1.3, and 1.2.0), users can choose to download which version they prefer

wenyongh (commented):
LGTM

PR #1824, opened 2022-12-20
libc-wasi: Fix spurious poll timeout

yamt:
This reverts the "Fix libc-wasi poll_oneoff hang issue" change. https://github.com/bytecodealliance/wasm-micro-runtime/pull/1300

My app usng poll(,,-1) timed out and surprised me today.

### DIFF
--- a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
+++ b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
@@ -2696,7 +2696,7 @@ wasmtime_ssp_poll_oneoff(
         timeout = ts > INT_MAX ? -1 : (int)ts;
     }
     else {
-        timeout = 1000;
+        timeout = -1;
     }
     int ret = poll(pfds, nsubscriptions, timeout);
 


yamt:
@wenyongh can you explain "libc-wasi poll_oneoff hang issue" which https://github.com/bytecodealliance/wasm-micro-runtime/pull/1300 fixed?

wenyongh:
> @wenyongh can you explain "libc-wasi poll_oneoff hang issue" which #1300 fixed?

@yamt Using `-1` makes the following wasm app hang in some environments:
```
iwasm --dir=. qjs.wasm
QuickJS - Type "\h" for help
qjs > \h          ==> hang when we input "\h" or "console.log('hello')"
```
[qjs.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/10273757/qjs.zip)

yamt:
> > @wenyongh can you explain "libc-wasi poll_oneoff hang issue" which #1300 fixed?
> 
> @yamt Using `-1` makes the following wasm app hang in some environments:
> 
> ```
> iwasm --dir=. qjs.wasm
> QuickJS - Type "\h" for help
> qjs > \h          ==> hang when we input "\h" or "console.log('hello')"
> ```
> 
> [qjs.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/10273757/qjs.zip)

* do you have a source of the wasm file?
* are you sure if it was not a problem in the app?
* which environment was it? i couldn't reproduce it on my env. (macOS x86-64, wamr version of this PR, fast interpreter)

yamt:
a found a wasmtime gh issue with the identical qjs.wasm binary and a similar symptom.
i guess it's a problem in the wasm module itself.

wenyongh:
> a found a [wasmtime gh issue](https://github.com/bytecodealliance/wasmtime/issues/4461) with the identical qjs.wasm binary and a similar symptom. i guess it's a problem in the wasm module itself.

Yes, it is the same file that I tested, it works in Ubuntu 20.04 physical machine but fails in VirtualBox Ubuntu 18.04. And the issue was reported by my colleague long time ago.
The wasm file is really an old build since it uses `wasi_unstable`. Seems that it is the problem of case itself. If you are very sure that using `-1` is correct, I think we can revert the change.

yamt:
> > a found a [wasmtime gh issue](https://github.com/bytecodealliance/wasmtime/issues/4461) with the identical qjs.wasm binary and a similar symptom. i guess it's a problem in the wasm module itself.
> 
> Yes, it is the same file that I tested, it works in Ubuntu 20.04 physical machine but fails in VirtualBox Ubuntu 18.04. And the issue was reported by my colleague long time ago. The wasm file is really an old build since it uses `wasi_unstable`. 

intersting. maybe some kind of race condition in somewhere.

> Seems that it is the problem of case itself. If you are very sure that using `-1` is correct, I think we can revert the change.

i'd suggest to revert it because it's clear it broke legitimate applications. eg. `poll(,, -1)`

on the other hand, what issue it has worked around is not clear. it might or might not even exist nowadays. (as wasmtime person speculated on their gh issue.)

PR #1825, opened 2022-12-21
Port WAMR to the FreeBSD platform and update document

TianlongLiang:
Port WAMR to the FreeBSD platform and update document.


wenyongh (commented):
LGTM

PR #1826, opened 2022-12-21
Add example with multiple threads creation

eloparco:
The example shows the creation of multiple threads. Thread spawn failure and thread identifier allocation are tested.
Now written as an example, it uses asserts so that it can be easily converted into a test once a test suite is created #1790.

### DIFF
--- a/samples/wasi-threads/wasm-apps/multiple_threads.c
+++ b/samples/wasi-threads/wasm-apps/multiple_threads.c
@@ -1,0 +1,84 @@
+/*
+ * Copyright (C) 2022 Amazon.com Inc. or its affiliates. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+#ifndef __wasi__
+#error This example only compiles to WASM/WASI target
+#endif
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <assert.h>
+#include <wasi/api.h>
+#include <stdbool.h>
+
+#define MAX_NUM_THREADS 4 // Default maximum number of threads for iwasm
+static const int64_t SECOND = 1000 * 1000 * 1000;
+
+typedef struct {
+    int th_ready;
+    int th_done;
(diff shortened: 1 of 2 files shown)


yamt:
FYI, there are a few test efforts:
wasi level tests: https://github.com/WebAssembly/wasi-threads/pull/19
libc level tests: https://github.com/WebAssembly/wasi-libc/pull/369

samples/wasi-threads/wasm-apps/multiple_threads.c
```diff
@@ -0,0 +1,84 @@
...
+        assert(false && "Wait operation in main thread failed");
+
+    printf("Try to create additional thread after previous was released\n");
+    shared_t data_succeed = { 0 };
+    thread_id = __wasi_thread_spawn(&data_succeed);
+    assert(thread_id >= 0 && "Thread creation must succeed");
+
+    printf("Created thread with id = %d\n", thread_id);
+    assert(thread_id == thread_ids[0]
+           && "New thread must reuse previously-released identifier");
```
yamt, line 73: * it seems racy as the first thread might be still running
* i'm not sure if it's a good idea to test this kind of implementation details (how ids are reused) in these samples.

eloparco: Not sure about the race condition, but I definitely agree that the check on the id is an implementation detail and should not be tested.

eloparco: What's unsafe is maybe that, after `__builtin_wasm_memory_atomic_wait32()` in the main thread, I can't assume that the thread created has finished and it's possible to create a new thread.

loganek:
I wonder if we actually need this example as it doesn't present any new feature that's already not in `no_pthread.c` already?

eloparco:
It's to test the failure and re-attempt on thread creation once the thread number limit is reached. Probably it doesn't add anything, I'll close the PR

Issue #1761, opened 2022-11-28
Integrate WASI test suite to a WAMR build


loganek:
Bytecode Alliance started a new project for unifying WASI tests; we already have a [repository](https://github.com/WebAssembly/wasi-testsuite) with quite a few test cases (we'll be expanding it over time, contributions are also welcome) and test harness to execute tests.
@yamt already added an adapter for WAMR here: https://github.com/WebAssembly/wasi-testsuite/pull/16 so the runtime can be already validated. The next step is to integrate the test suite with WAMR's CI; this can be done by following the steps in [documentation](https://github.com/WebAssembly/wasi-testsuite#getting-started). I already checked locally, at the moment WAMR passes all of the tests defined in the repository on Linux platform.

WASI test harness also allows executing tests for APIs that are not part of the snapshot, so in the future we can implement and run tests for e.g. pthread library or sockets.

wenyongh:
Done with #1832.

PR #1832, opened 2022-12-23
Integrate WASI test suite

eloparco:
Add [wasi-testsuite](https://github.com/WebAssembly/wasi-testsuite) to set of tests executed from `test_wamr.sh`. 
Additional description here: #1761

### DIFF
--- a/.github/workflows/compilation_on_android_ubuntu.yml
+++ b/.github/workflows/compilation_on_android_ubuntu.yml
@@ -61,6 +61,7 @@ env:
   SIMD_TEST_OPTIONS: "-s spec -b -S -P"
   THREADS_TEST_OPTIONS: "-s spec -b -p -P"
   X86_32_TARGET_TEST_OPTIONS: "-m x86_32 -P"
+  WASI_TEST_OPTIONS: "-s wasi_certification"
 
 jobs:
   build_llvm_libraries:
@@ -376,7 +377,7 @@ jobs:
           exit $?
         working-directory: ./samples/simple
 
-  spec_test:
+  test:
     needs: [build_iwasm, build_llvm_libraries, build_wamrc]
     runs-on: ubuntu-20.04
     strategy:
@@ -389,6 +390,7 @@ jobs:
             $MULTI_MODULES_TEST_OPTIONS,
             $SIMD_TEST_OPTIONS,
             $THREADS_TEST_OPTIONS,
+            $WASI_TEST_OPTIONS,
           ]
         exclude:
           # uncompatiable modes and features
@@ -400,6 +402,9 @@ jobs:
           # aot and jit don't support multi module
           - running_mode: "aot"
             test_option: $MULTI_MODULES_TEST_OPTIONS
+          # aot is WAMR-specific while wasi-testsuite is generic
+          - running_mode: "aot"
+            test_option: $WASI_TEST_OPTIONS
           - running_mode: "jit"
             test_option: $MULTI_MODULES_TEST_OPTIONS
           # fast-jit is only tested on default mode, exclude other three
@@ -441,7 +446,7 @@ jobs:
         if: env.USE_LLVM == 'true' && steps.cache_llvm.outputs.cache-hit != 'true'
         run: echo "::error::can not get prebuilt llvm libraries" && exit 1
 
-      - name: run spec tests default and extra
+      - name: run tests
         run: ./test_wamr.sh ${{ matrix.test_option }} -t ${{ matrix.running_mode }}
         working-directory: ./tests/wamr-test-suites
 
@@ -456,7 +461,7 @@ jobs:
           sudo apt-get update &&
           sudo apt install -y g++-multilib lib32gcc-9-dev
 
-      - name: run spec tests x86_32
+      - name: run tests x86_32
         if: env.TEST_ON_X86_32 == 'true'
         run: ./test_wamr.sh ${{ env.X86_32_TARGET_TEST_OPTIONS }} ${{ matrix.test_option }} -t ${{ matrix.running_mode }}
         working-directory: ./tests/wamr-test-suites
--- a/tests/wamr-test-suites/test_wamr.sh
+++ b/tests/wamr-test-suites/test_wamr.sh
@@ -46,6 +46,7 @@ PLATFORM=$(uname -s | tr A-Z a-z)
 PARALLELISM=0
 ENABLE_QEMU=0
 QEMU_FIRMWARE=""
+WASI_TESTSUITE_COMMIT="1d913f28b3f0d92086d6f50405cf85768e648b54"
 
 while getopts ":s:cabt:m:MCpSXxPQF:" opt
 do
@@ -447,7 +448,10 @@ function spec_test()
     cd ${WORK_DIR}
     echo "python3 ./all.py ${ARGS_FOR_SPEC_TEST} | tee -a ${REPORT_DIR}/spec_test_report.txt"
     python3 ./all.py ${ARGS_FOR_SPEC_TEST} | tee -a ${REPORT_DIR}/spec_test_report.txt
-    [[ ${PIPESTATUS[0]} -ne 0 ]] && exit 1
+    if [[ ${PIPESTATUS[0]} -ne 0 ]];then
+        echo -e "\nspec tests FAILED" | tee -a ${REPORT_DIR}/spec_test_report.txt
+        exit 1
+    fi
     cd -
 
     echo -e "\nFinish spec tests" | tee -a ${REPORT_DIR}/spec_test_report.txt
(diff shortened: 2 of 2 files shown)


eloparco:
What's missing in the PR (I'll resume the work after I return from vacation, but I pushed the change to get some feedback):
- is the `return 0` [here](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/product-mini/platforms/posix/main.c#L701) actually needed? I commented it out and ran the spec test in GitHub actions and they succeed without that part; I asked already in #1377
- [one test](https://github.com/WebAssembly/wasi-testsuite/blob/main/tests/c/testsuite/stat-dev-ino.c) fails right now, I need to look into it

product-mini/platforms/posix/main.c
```diff
@@ -696,10 +696,10 @@ main(int argc, char *argv[])
...
-    (void)ret;
-    return 0;
-#else
+    // #if WASM_ENABLE_SPEC_TEST != 0
+    //     (void)ret;
+    //     return 0;
+    // #else
     return ret;
-#endif
+    // #endif
```
yamt, line 704: how is this related to the rest of the PR?

tests/wamr-test-suites/test_wamr.sh
```diff
@@ -453,7 +453,7 @@ function spec_test()
     echo -e "\nFinish spec tests" | tee -a ${REPORT_DIR}/spec_test_report.txt
 }
 
-function wasi_test()
+function wasi_test_OUTDATED()
```
yamt, line 456: at this point it has very different coverage to simply replace.

product-mini/platforms/posix/main.c
eloparco: In [wasi-testsuite](https://github.com/WebAssembly/wasi-testsuite) there are some tests that check the return code and, because of the hardcoded `return 0`, would fail.
It has been fixed in the `main` branch now #1377 #1839.

To merge this PR, we need https://github.com/WebAssembly/wasi-testsuite/pull/40 to be merged and `main` to be merged into [dev/wasi_threads](https://github.com/bytecodealliance/wasm-micro-runtime/tree/dev/wasi_threads).

loganek: @wenyongh would you be able to help with the second point and merge `main` to `dev/wasi_threads` and unblock this PR?

tests/wamr-test-suites/test_wamr.sh
loganek: Agree, I'd rather keep this name unchanged, and maybe find another name for the new test suite (`wasi_certification_tests()`?). Once we have the same coverage, we can delete this test suite completely.

tests/wamr-test-suites/test_wamr.sh
```diff
@@ -469,6 +469,37 @@ function wasi_test()
...
+    else
+        echo "wasi exists, pulling updates from github"
+        pushd wasi-testsuite
+        git fetch
+        git reset --hard origin/prod/testsuite-base
+        popd
+    fi
+
+    cd wasi-testsuite
+    python3 -m pip install -r test-runner/requirements.txt
```
loganek, line 490: I wonder if we can do it in virtual environment [1] so we don't pollute host's environment with packages that they might not want to have? 

[1] https://docs.python.org/3/tutorial/venv.html

tests/wamr-test-suites/test_wamr.sh
```diff
@@ -469,6 +469,37 @@ function wasi_test()
...
+    IWASM_PATH=$(dirname ${IWASM_CMD})
+    PATH=${PATH}:${IWASM_PATH} python3 test-runner/wasi_test_runner.py \
+                -r adapters/wasm-micro-runtime.sh \
+                -t \
+                    tests/c/testsuite/ \
+                    tests/assemblyscript/testsuite/ \
+                | tee -a ${REPORT_DIR}/wasi_test_report.txt
+    [[ ${PIPESTATUS[0]} -ne 0 ]] && exit 1
```
loganek, line 498: does it mean that in case of failure the L500 won't be executed? Is that desired? If we don't have the last line, we probably wouldn't know from this log file whether the test execution completed or just broke half way through.

eloparco: I was replicating what's done already for spec tests https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/tests/wamr-test-suites/test_wamr.sh#L450. I'll update it in both cases to add something to the log file before exiting.

product-mini/platforms/posix/main.c
wenyongh: @eloparco @loganek I have merged `main` into `dev/wasi_threads`, since there are code conflicts, I applied the merging manually, please check whether it works for you:
 https://github.com/bytecodealliance/wasm-micro-runtime/commit/3403f0a7f4d838f260b55b327da8be00a8644229

loganek: @wenyongh thanks a lot, do you remember which files required manual conflict resolution?

tests/wamr-test-suites/test_wamr.sh
eloparco: and for the new `wasi_certification_tests`, should I just skip the coverage generation? I don't think we can use the same (`lcov`) as in the other tests?

product-mini/platforms/posix/main.c
wenyongh: @loganek Not the files related to wasi-threads, but some files modified in main branch. I used mode `rebase and merge` instead of mode `merge` to merge `main` into `dev/wasi_threads` last time, it didn't produce "a new generated commit" and seems that `dev/wasi_threads` didn't remember the sync-up status from `main`, so when the same line of the same file was modified in main branch by new commit, conflict occurs. Similar conflicts occurred when I tried merging main into `dev/fast_jit_simd` and `gitbook`, I think I should use `merge` mode to merge main into other branches next time.

loganek: ok, thanks. Looks good to me. I think once we're ready with the dev branch, I'll rebase it completely on top of main and open a final PR to rebase-merge it, to keep the git tree clean.

tests/wamr-test-suites/test_wamr.sh
loganek: I'd keep the integration simple for now, and we can add more features in the future if needed

tests/wamr-test-suites/test_wamr.sh
```diff
@@ -469,6 +472,43 @@ function wasi_test()
...
+    python3 -m pip install -r test-runner/requirements.txt
+    IWASM_PATH=$(dirname ${IWASM_CMD})
+    PATH=${PATH}:${IWASM_PATH} python3 test-runner/wasi_test_runner.py \
+                -r adapters/wasm-micro-runtime.sh \
+                -t \
+                    tests/c/testsuite/ \
+                    tests/assemblyscript/testsuite/ \
+                | tee -a ${REPORT_DIR}/wasi_test_report.txt
+    exit_code=${PIPESTATUS[0]}
+    deactivate
```
loganek, line 503: what's that for?

eloparco: to stop using the `wasi-env` virtual environment created before; if another test suite has dependencies it will create another virtual environment with its own dependencies

loganek: Ah right, thanks.

.github/workflows/compilation_on_android_ubuntu.yml
```diff
@@ -475,7 +477,7 @@ jobs:
           sudo apt-get update &&
           sudo apt install -y g++-multilib lib32gcc-9-dev
 
-      - name: run spec tests x86_32
-        if: env.TEST_ON_X86_32 == 'true'
+      - name: run tests x86_32
+        if: env.TEST_ON_X86_32 == 'true' && matrix.test_option != '$WASI_TEST_OPTIONS'
```
loganek, line 481: What's the reason for not running wasi tests on x86-32?

loganek:
I don't think the PR has anything to do with wasi threads, could you change the destination branch to `main`?

tests/wamr-test-suites/test_wamr.sh
```diff
@@ -469,6 +472,43 @@ function wasi_test()
...
+    cd ${WORK_DIR}
+    if [ ! -d "wasi-testsuite" ]; then
+        echo "wasi not exist, clone it from github"
+        git clone -b prod/testsuite-base \
+            --single-branch https://github.com/WebAssembly/wasi-testsuite.git
+    else
+        echo "wasi exists, pulling updates from github"
+        pushd wasi-testsuite
+        git fetch
+        git reset --hard origin/prod/testsuite-base
```
loganek, line 488: I'd suggest at least for now we point to a specific commit. Using always the latest testsuite is a bit risky and might block the pipeline, especially that there's a lot going on in the wasi-testsuite repo now. We might want to choose to use always the latest version in the future, but I'd point to specific commit for now.

.github/workflows/compilation_on_android_ubuntu.yml
loganek: I think it'd be worth running those tests on x86-32 as well (to make sure WASI implementation works on those platforms as well).

wenyongh: Do we want to run wasi tests on x86-32? For the output of the CI, I found that x86-32 wasi tests are ignored:
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/3834321537/jobs/6526922968
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/3834321537/jobs/6526923240

Seems that we need to change the below lines in the CI file
```yml
- name: set env variable(if x86_32 test needed)
        if: >
          (matrix.test_option == '$DEFAULT_TEST_OPTIONS' || matrix.test_option == '$THREADS_TEST_OPTIONS')
          && matrix.running_mode != 'fast-jit' && matrix.running_mode != 'jit'
        run: echo "TEST_ON_X86_32=true" >> $GITHUB_ENV
```
into:
```yml
- name: set env variable(if x86_32 test needed)
        if: >
          (matrix.test_option == '$DEFAULT_TEST_OPTIONS' || matrix.test_option == '$THREADS_TEST_OPTIONS'
           || matrix.test_option == '$WASI_TEST_OPTIONS')
          && matrix.running_mode != 'fast-jit' && matrix.running_mode != 'jit'
        run: echo "TEST_ON_X86_32=true" >> $GITHUB_ENV
```

lum1n0us:
LGTM

tests/wamr-test-suites/test_wamr.sh
loganek: It's not an official concept (at least not yet), but we strive to provide an extensive coverage so the testsuite can act as certification tests.

loganek (approved):
LGTM

PR #1837, opened 2022-12-26
Upgrade toolkits: llvm, wasi-sdk, emsdk, wabt and so on

wenyongh:


lum1n0us:
I thought we didn't need `test-tools/build-wasi-sdk` long ago.

wenyongh:
> I thought we didn't need `test-tools/build-wasi-sdk` long ago.

Yes, removing `build-wasi-sdk` would be better if we can. Do you mean to use the recent versions of wasi-sdk which has supported SIMD? There are issues when using wasi-sdk-16.0 to build the bwa workload:
- samples/workload/bwa/bwa/kthread.c:1:10: fatal error: 'pthread.h' file not found
- fail to use some SIMD header files: mmintrin.h, xmmintrin, immintrin.h etc:
```cmake
/opt/wasi-sdk-12.0/lib/clang/11.0.0/include/mmintrin.h:50:12: error: invalid conversion between vector type '__m64' (vector of 1 'long long' value) and integer type 'int' of different size
    return (__m64)__builtin_ia32_vec_init_v2si(__i, 0);
           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/opt/wasi-sdk-12.0/lib/clang/11.0.0/include/mmintrin.h:129:12: error: invalid conversion between vector type '__m64' (vector of 1 'long long' value) and integer type 'int' of different size
    return (__m64)__builtin_ia32_packsswb((__v4hi)__m1, (__v4hi)__m2);
```

lum1n0us:
- there is no *pthread* suppoting in `wasi-sdk` for sure. If it is required, we shall add our [pthread.h](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/wamr-sdk/app/libc-builtin-sysroot/include/pthread.h) into include paths.

- SIMD supporting comes from clang headers. https://github.com/llvm/llvm-project/tree/main/clang/lib/Headers. There must be some reasons for above failures. Let me take a look firstly

doc/build_wasm_app.md
```diff
@@ -3,12 +3,12 @@
...
+For C and C++, WASI-SDK version 16.0+ is the major tool supported by WAMR to build WASM applications. Also, we can use [Emscripten SDK (EMSDK)](https://github.com/emscripten-core/emsdk), but it is not recommended. And there are some other compilers such as the standard clang compiler, which might also work [here](./other_wasm_compilers.md).
 
 To install WASI SDK, please download the [wasi-sdk release](https://github.com/CraneStation/wasi-sdk/releases) and extract the archive to default path `/opt/wasi-sdk`.
 
 The official *wasi-sdk release* doesn't fully support *latest 128-bit SIMD spec* yet. WAMR provides a script in [build-wasi-sdk](../test-tools/build-wasi-sdk/) to generate
-another wasi-sdk with *llvm-13* from source code and installs it at *../test-tools/wasi-sdk*. If you plan to build WASM applications with *latest 128-bit SIMD*, please use it instead of the official release.
+another wasi-sdk with *llvm-15* from source code and installs it at *../test-tools/wasi-sdk*. If you plan to build WASM applications with *latest 128-bit SIMD*, please use it instead of the official release.
```
lum1n0us, line 11: It may not be right to say "not fully support" now.

Issue #1728, opened 2022-11-19
Unexpected value on the stack (classic interpreter)


erxiaozhou:
# Build commands
I compile the code with commit id f59ffa0d63f832982da1caa524aaac146f3f1422.
Platform: Ubuntu 20.04
CPU: amd64 
compile:
```shell
export CC=/usr/bin/clang-14
export CXX=/usr/bin/clang++-14
cd product-mini/platforms/linux/;rm -rf build
cmake  -DWAMR_BUILD_INTERP=1 -DWAMR_BUILD_AOT=0 -DCMAKE_BUILD_TYPE=Debug -DWAMR_BUILD_FAST_INTERP=0 -Bbuild
cd build
make -j40
```
execute:
```shell
iwasm --heap-size=0 -f to_test <tc_name>
```
[iwasm_nan1.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/10047806/iwasm_nan1.zip)

# Actual behavior:
The value left on the stack is ['0x39', '0x0', '0x0', '0x0', '0x0', '0x0', '0xf0', '0x7f']

# Expected behavior:
The value left on the stack is ['0x39', '0x0', '0x0', '0x0', '0x0', '0x0', '0xf8', '0x7f'] (The most significant bit of payload is 1)

wenyongh:
Hi, thanks for reporting the issue! I submitted PR #1733 to fix it, could you please help review the patch and try again? Thanks.

PR #1846, opened 2022-12-29
Using float version library routine for XIP

no1wudi:


core/iwasm/aot/aot_intrinsic.c
```diff
@@ -223,7 +223,7 @@ aot_intrinsic_fmin_f32(float32 a, float32 b)
     else if (isnan(b))
         return b;
     else
-        return (float32)fmin(a, b);
+        return fminf(a, b);
```
wenyongh, line 226: Not very sure whether the original implementations of aot_intrinsic_fmin_f32/f64 and aot_intrinsic_fmax_f32/f64 are correct, here codes of aot_intrinsic_fmin_f32 and aot_intrinsic_fmin_f64 are inconsistent. Developer ever reported error for the implementation of interpreter:
https://github.com/bytecodealliance/wasm-micro-runtime/issues/1728
And now we have changed them to:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_interp_classic.c#L138-L180

Suggest to use the same implementations of interpreter for aot_intrinsic_fmin_f32/f64 and aot_intrinsic_fmax_f32/f64.

no1wudi: Yes, current implementations can't pass the spec test, I had fixed it but not submitted yet.
I prefer to split these fixes to small patch to simplify patch management (a patch only do one thing).

wenyongh: OK, so let's merge this PR first.

Issue #1805, opened 2022-12-12
Important synchronization issue


qianyimao:
it will miss notification in "os_cond_signal" while "os_cond_wait_internal" updating "korp_cond.wait_list_lock" or there isn't a thread is waiting. windows platform.

wenyongh:
Hi, yes, but it really requires preparation for os_cond_wait to start waiting the notification. Is there any idea from you? Thanks.

qianyimao:
append a "count" field in "korp_cond" which indicates the missed signal count.
increase "count" when "thread_wait_list" is null in os_cond_signal.
check "count" before waiting in os_cond_wait_internal.
maybe it work.

qianyimao:
or limit the ”count” in 0,1.

qianyimao:
1.platform_internal.h
typedef struct korp_cond {
    korp_mutex wait_list_lock;
    os_thread_wait_list thread_wait_list;
    int signaled;
}
 
2.win_thread.c!os_cond_init
cond->thread_wait_list = NULL;
cond->signaled = FALSE;
 
3.win_thread.c!os_cond_wait_internal
os_mutex_lock(&cond->wait_list_lock);
if (cond->signaled) {
    node = NULL;
    cond->signaled = FALSE;
}
else {
    ...
}
os_mutex_unlock(&cond->wait_list_lock);
if (!node) { return BHT_OK; }
 
4.win_thread.c!os_cond_signal
if (cond->thread_wait_list)
    os_sem_signal(&cond->thread_wait_list->sem);
else
    cond->signaled = TRUE;

qianyimao:
maybe the patch work.

wenyongh:
Hi, there may be issue for this patch: the signal thread sets `cond->signaled` to `true` and expects the waiting thread to directly return when it detects that its `cond->signaled` is set to true, but suppose that the waiting thread doesn't call os_cond_wait immediately, instead, it does something for some time (e.g. a loop or sleep(5)), and then calls os_cond_wait the cond, is it correct to return directly in this situation?

My understanding of your concern is that appending the wait node to the end of wait node list may take some time:
```C
        /* Add to end of wait list */
        os_thread_wait_node *p = cond->thread_wait_list;
        while (p->next)
            p = p->next;
        p->next = node;
```
Maybe we can refine the code to append the node quickly, is it OK for you?

qianyimao:
[...]

PR #1853, opened 2023-01-03
Refine Windows thread waiting list operations

wenyongh:
Add thread_wait_list_end node for thread data and cond for Windows platform
to speedup the thread join and cond wait operations: no need to traverse the
wait list to get the end node to append the wait node.

### DIFF
--- a/core/shared/platform/windows/platform_internal.h
+++ b/core/shared/platform/windows/platform_internal.h
@@ -72,6 +72,7 @@ typedef struct os_thread_wait_node *os_thread_wait_list;
 typedef struct korp_cond {
     korp_mutex wait_list_lock;
     os_thread_wait_list thread_wait_list;
+    struct os_thread_wait_node *thread_wait_list_end;
 } korp_cond;
 
 #define bh_socket_t SOCKET
--- a/core/shared/platform/windows/win_thread.c
+++ b/core/shared/platform/windows/win_thread.c
@@ -37,6 +37,8 @@ typedef struct os_thread_data {
     korp_mutex wait_lock;
     /* Waiting list of other threads who are joining this thread */
     os_thread_wait_list thread_wait_list;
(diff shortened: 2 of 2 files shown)


loganek:
Could you clarify what's the problem and how the change solves it? I understand now that appending new node can be faster than before, but why was the slowness a problem at all (and was it really that slow)? I saw this issue https://github.com/bytecodealliance/wasm-micro-runtime/issues/1805 but still not clear to me what are we trying to achieve here. Thanks

wenyongh:
> Could you clarify what's the problem and how the change solves it? I understand now that appending new node can be faster than before, but why was the slowness a problem at all (and was it really that slow)? I saw this issue #1805 but still not clear to me what are we trying to achieve here. Thanks

Here it is just to improve the performance, a defect of the original implementation is that it may be a little slow when traversing the list to get the end of the list (when the list is long):
```C
         /* Add to end of wait list */
        os_thread_wait_node *p = cond->thread_wait_list;
        while (p->next)
            p = p->next;
        p->next = node;
```
Using `thread_wait_list_end` to mark the end node of the list is to skip the searching since we can append the node to the end directly.

loganek (approved):
Ok, that's what I thought although I didn't expect that particular list can get so long it's even noticeable. 

LGTM.

yamt:
i can understand the motivation to make the cond list a fifo.
however, for os_thread_data, it seems simpler to make it add a node to the head of the list.

wenyongh:
> i can understand the motivation to make the cond list a fifo. however, for os_thread_data, it seems simpler to make it add a node to the head of the list.

Yes, refer to the pthread_join manual, "If multiple threads simultaneously try to join with the same thread, the results are undefined", so maybe we can do like you said. Here a consideration is that the first thread to join had better be notified first, or fifo.

PR #1859, opened 2023-01-04
Quit if meeting unimport globals

lum1n0us:


core/iwasm/aot/aot_runtime.c
```diff
@@ -976,6 +976,25 @@ execute_memory_init_function(AOTModuleInstance *module_inst)
...
+{
+    uint32 i;
+
+    /* init_func_ptrs() will go through import functions */
+
+    for (i = 0; i < module->import_global_count; i++) {
+        AOTImportGlobal *global = module->import_globals + i;
+        if (!global->is_linked) {
+            LOG_WARNING("warning: failed to link import global (%s, %s)",
+                        global->module_name, global->global_name);
```
wenyongh, line 990: Should be an error here, use `set_error_buf`

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -1250,8 +1251,9 @@ check_linked_symbol(WASMModuleInstance *module_inst, char *error_buf,
             return false;
 #else
 #if WASM_ENABLE_WAMR_COMPILER == 0
-            LOG_DEBUG("warning: failed to link import global (%s, %s)",
-                      global->module_name, global->field_name);
+            LOG_WARNING("warning: failed to link import global (%s, %s)",
+                        global->module_name, global->field_name);
```
wenyongh, line 1255: Same as above, should be an error

Issue #1831, opened 2022-12-22
Build should check for global libuv and uvwasi before downloading


phated:
I've run into some problems packaging WAMR for OCaml because libuv and uvwasi are **always** downloaded when setting `WAMR_BUILD_LIBC_UVWASI=1`. I believe that the CMake should be altered to locate libuv and uvwasi in the environment before downloading them. This is actually a feature of `FetchContent_MakeAvailable` after CMake 3.24 but I think WAMR allows an older version (and doesn't use `FetchContent_MakeAvailable`).

lum1n0us:
IIUC, 
- `cmake_minimum_required` just limited and check the minimum version of the cmake. It actually use the cmake on the host machine in the system path as long as it passed the requirement.
- `FetchContent_MakeAvailable`  equivalent logic can be obtained by calling  `FetchContent_GetProperties` and `FetchContent_Populate`.  But yes, 
>  it is simpler and provides additional features over the pattern above
- `FetchContent_MakeAvailable` is a new feature in CMake 3.14.

At first, I was also confused by the message *Fetching uvwasi ..* and *Featching libuv ..* and time pausing when cmake configuring. But after took a close look of the timestamp of a download file, `build/_deps/libuv-src/XXX` and `build/_deps/uvwasi-src/XXX`, I realized cmake actually reused downloaded files unless `build/_dpes/***` are removed.

phated:
You are correct that it will only be downloaded once, as long as sources aren't removed; however, some tools build in sandbox and clear all sources.

This issue is about using find_library before downloading anything, so libuv and uvwasi can be provided externally and are never downloaded.

lum1n0us:
True. it is better if we can store it externally. Let's see what we can do. thanks for asking.

phated:
This should be completed with the merge of #1861

PR #1861, opened 2023-01-04
feat: Use system libuv if available

phated:
This is my attempt at searching for the system libuv ~~before~~ instead of downloading it, if libuv exists on the system already. As I mentioned in https://github.com/bytecodealliance/wasm-micro-runtime/issues/1831, this is needed because some tools build in a sandbox and clear the extra sources.

I didn't implement this for uvwasi because I wanted to get feedback first.


core/iwasm/libraries/libc-uvwasi/libc_uvwasi.cmake
```diff
@@ -8,21 +8,33 @@ set (LIBUV_VERSION v1.44.2)
 add_definitions (-DWASM_ENABLE_LIBC_WASI=1 -DWASM_ENABLE_UVWASI=1)
 
 include(FetchContent)
+include(FindPackageHandleStandardArgs)
+
+# Based on https://github.com/xenoscopic/libuv-cmake/blob/fc164590c6c723fd474e35b0b52022ebaecef745/Findlibuv.cmake
+find_path (libuv_INCLUDE_DIR NAMES uv.h)
+find_library (libuv_LIBRARY NAMES uv libuv)
+find_package_handle_standard_args(libuv libuv_LIBRARY libuv_INCLUDE_DIR)
```
lum1n0us, line 16: IIUC, L14 and L15 will search *uv.h* and *libuv.so* in the default path which usually is also the system path and doesn't include a directory of WAMR source tree. So I am not sure it will locate the libuv downloaded by CMake.

phated: This is not designed to load libuv downloaded by cmake. It is only downloaded below if there is no system installed libuv

phated: I've updated the original description for clarity

lum1n0us: Got it. It would be helpful if using system providing libuv.so. 

BTW, how about the version of system installed libuv? Should  be worried about the compatibility problem?

wenyongh: And how to install the system packages of libuv and uvwasi? Could you help update the document?

phated: > And how to install the system packages of libuv and uvwasi? Could you help update the document?

@wenyongh this PR is not completed. Is this a good way to proceed? If so, I can make similar changes for uvwasi.

wenyongh: It is good to me if it really resolves the issue that developers met, could you make the changes for uvwasi, and update the document? e.g. update the `Note` of `Configure LIBC` here:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/doc/build_wamr.md#configure-libc

phated: I've been working on this, but when the uvwasi source isn't available, the cmake build process errors that it cannot find any of the symbols. Either I'm incorrectly linking uvwasi or maybe it doesn't support linking.

phated:
I believe this is completed but it relies on https://github.com/nodejs/uvwasi/pull/210

core/iwasm/libraries/libc-uvwasi/FindLIBUV.cmake
```diff
@@ -0,0 +1,25 @@
+# Find libuv library
```
wenyongh, line 1: Should add license header, please refer to other cmake file

core/iwasm/libraries/wasi-nn/test/CMakeLists.txt
```diff
@@ -162,12 +162,12 @@ add_executable (iwasm ${WAMR_ROOT_DIR}/product-mini/platforms/${WAMR_BUILD_PLATF
 
 install (TARGETS iwasm DESTINATION bin)
 
-target_link_libraries (iwasm vmlib ${LLVM_AVAILABLE_LIBS} ${UV_A_LIBS} ${TENSORFLOW_LIB} -lm -ldl -lpthread)
+target_link_libraries (iwasm vmlib ${LLVM_AVAILABLE_LIBS} ${UV_A_LIBS} ${UVWASI_LIBS} ${TENSORFLOW_LIB} -lm -ldl -lpthread)
```
wenyongh, line 165: Can we `set (UV_A_LIBS "${UV_A_LIBS} ${UVWASI_LIBS}")` in lib_uvwasi.cmake so that we don't need to change these cmake files?

core/iwasm/libraries/libc-uvwasi/FindLIBUV.cmake
phated: Done

core/iwasm/libraries/wasi-nn/test/CMakeLists.txt
phated: Done

phated:
Looks like some of the linux tests flaked on trying to download an image

core/iwasm/libraries/libc-uvwasi/libc_uvwasi.cmake
```diff
@@ -9,36 +9,52 @@ add_definitions (-DWASM_ENABLE_LIBC_WASI=1 -DWASM_ENABLE_UVWASI=1)
...
+        GIT_REPOSITORY https://github.com/nodejs/uvwasi.git
+        GIT_TAG main
+    )
+    FetchContent_GetProperties(uvwasi)
+    if (NOT uvwasi_POPULATED)
+        message("-- Fetching uvwasi ..")
+        FetchContent_Populate(uvwasi)
+        include_directories("${uvwasi_SOURCE_DIR}/include")
+        add_subdirectory(${uvwasi_SOURCE_DIR} ${uvwasi_BINARY_DIR} EXCLUDE_FROM_ALL)
+        set (UVWASI_LIBRARIES uvwasi_a)
```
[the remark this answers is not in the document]
wenyongh: typo: add them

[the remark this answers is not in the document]
phated: It looks like this was already added for uv_a so I just added it to `uvwasi_a`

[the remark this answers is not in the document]
wenyongh: The issue in my machine is that package libuv1-dev is installed while no uvwasi related package is installed, and `find_package(LIBUV QUIET)` returns true but uvwasi must be compiled from source code. And when downloading and compiling uvwasi, `it also downloads, compiles and links libuv (static lib)` instead of linking the libuv.so from the installed package. And here the compiler reports error when linking libuv.a. I don't know how to let uvwasi use the installed libuv package directly, if we can, it is unnecessary to add the line. Otherwise, had better add the line. Any idea for it? Thanks.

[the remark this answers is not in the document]
phated: >  I don't know how to let uvwasi use the installed libuv package directly, if we can, it is unnecessary to add the line.

Sorry for the confusion. My PR at https://github.com/nodejs/uvwasi/pull/210 was required to have uvwasi use the same global version. I've now merged the PR into the main branch so you should be able to build with the new source code and it will link the global libuv.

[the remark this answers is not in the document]
wenyongh: OK, thanks a lot. I tested again, it links the correct libuv now. An issue is in Ubuntu-20.04, when package libuv1-dev is installed, compile warning and link error occur:
```bash
product-mini/platforms/linux/build/_deps/uvwasi-src/src/uvwasi.c:1743:7: warning: implicit declaration of function ‘uv_fs_lutime’; did you mean ‘uv_fs_futime’? [-Wimplicit-function-declaration]
 1743 |   r = uv_fs_lutime(NULL, &req, resolved_path, atim, mtim, NULL);
```
```bash
[ 98%] Linking C executable iwasm
/usr/local/bin/ld: _deps/uvwasi-build/libuvwasi_a.a(uvwasi.c.o): in function `uvwasi_path_filestat_set_times':
uvwasi.c:(.text+0x293f): undefined reference to `uv_fs_lutime'
```
Note that the warning/error isn't reported in Ubuntu-22.04, it works fine in Ubuntu-22.04.
I think it is due to the installed libuv.so doesn't have the `uv_fs_lutime` function. It is not the issue of this PR, let's merge it first. And it would be good if we can resolve the issue, e.g. modifying uvwasi by using `uv_fs_futime`?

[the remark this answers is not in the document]
phated: That sounds good. I don't have experience with the libuv function mentioned, but I can raise the issue for the other members of the uvwasi team with more libuv experience.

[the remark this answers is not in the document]
phated: That issue has been raised at https://github.com/nodejs/uvwasi/issues/213

PR #1862, opened 2023-01-04
Reserve TID values for WASI threads

eloparco:
According to the [WASI thread specification](https://github.com/WebAssembly/wasi-threads/pull/16), some thread identifiers are reserved and should not be used. In fact, only IDs between `1` and `0x1FFFFFFF` are valid.
The thread ID allocator has been moved to a separate class to avoid polluting the `lib_wasi_threads_wrapper` logic.

### DIFF
--- a/core/iwasm/libraries/lib-wasi-threads/tid_allocator.c
+++ b/core/iwasm/libraries/lib-wasi-threads/tid_allocator.c
@@ -1,0 +1,87 @@
+/*
+ * Copyright (C) 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+
+#include "tid_allocator.h"
+#include "wasm_export.h"
+#include "bh_log.h"
+
+#define MIN(a, b) (((a) < (b)) ? (a) : (b))
+
+struct tidAllocator {
+    int32 *ids;  // Array used to store the stack
+    uint32 size; // Stack capacity
+    uint32 pos;  // Index of the element after the stack top
+};
+
+TidAllocator *
+tid_allocator_init()
+{
+    TidAllocator *tid_allocator = wasm_runtime_malloc(sizeof(TidAllocator));
+    tid_allocator->size = MIN(TID_ALLOCATOR_INIT_SIZE, TID_MAX - TID_MIN + 1);
+    tid_allocator->pos = tid_allocator->size;
+    tid_allocator->ids =
+        wasm_runtime_malloc(tid_allocator->size * sizeof(int32));
+    if (tid_allocator->ids == NULL)
+        return NULL;
+
+    for (int64 i = tid_allocator->pos - 1; i >= 0; i--)
+        tid_allocator->ids[i] = TID_MIN + (tid_allocator->pos - 1 - i);
+
+    return tid_allocator;
+}
+
+void
+tid_allocator_destroy(TidAllocator *tid_allocator)
+{
+    wasm_runtime_free(tid_allocator->ids);
+    wasm_runtime_free(tid_allocator);
+}
+
+int32
+tid_allocator_get(TidAllocator *tid_allocator)
+{
+    if (tid_allocator->pos == 0) { // Resize stack and push new thread ids
+        if (tid_allocator->size == TID_MAX - TID_MIN + 1) {
+            LOG_ERROR("Maximum thread identifier reached");
+            return -1;
+        }
+
+        uint32 old_size = tid_allocator->size;
+        uint32 new_size = MIN(tid_allocator->size * 2, TID_MAX - TID_MIN + 1);
+        if (new_size != TID_MAX - TID_MIN + 1
+            && new_size / 2 != tid_allocator->size) {
+            LOG_ERROR("Overflow detected during new size calculation");
+            return -1;
+        }
+
+        size_t realloc_size = new_size * sizeof(int32);
+        if (realloc_size / sizeof(int32) != new_size) {
+            LOG_ERROR("Overflow detected during realloc");
+            return -1;
+        }
+        int32 *tmp = wasm_runtime_realloc(tid_allocator->ids, realloc_size);
+        if (tmp == NULL) {
+            LOG_ERROR("Thread ID allocator realloc failed");
+            return -1;
+        }
+
+        tid_allocator->size = new_size;
+        tid_allocator->pos = new_size - old_size;
+        tid_allocator->ids = tmp;
+        for (int64 i = tid_allocator->pos - 1; i >= 0; i--)
+            tid_allocator->ids[i] = TID_MIN + (tid_allocator->size - 1 - i);
+    }
+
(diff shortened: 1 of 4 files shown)


core/iwasm/libraries/lib-wasi-threads/tid_allocator.c
```diff
@@ -0,0 +1,89 @@
...
+#include "wasm_export.h"
+#include "bh_log.h"
+
+#define MIN(a, b) (((a) < (b)) ? (a) : (b))
+
+struct tidAllocator {
+    int32 *ids;
+    uint32 size;    // Stack capacity
+    uint32 pos;     // Index of the position after the stack top
+    uint32 next_id; // To keep track of the ids to put in the stack
```
yamt, line 16: next_id seems redundant but i guess it isn't critical.

yamt (approved):

core/iwasm/libraries/lib-wasi-threads/tid_allocator.c
eloparco: I was using it for readability, but probably it's better to get rid of it, being the next id calculation not that complicated

core/iwasm/libraries/lib-wasi-threads/tid_allocator.c
```diff
@@ -0,0 +1,87 @@
...
+{
+    TidAllocator *tid_allocator = wasm_runtime_malloc(sizeof(TidAllocator));
+    tid_allocator->size = MIN(TID_ALLOCATOR_INIT_SIZE, TID_MAX - TID_MIN + 1);
+    tid_allocator->pos = tid_allocator->size;
+    tid_allocator->ids =
+        wasm_runtime_malloc(tid_allocator->size * sizeof(int32));
+    if (tid_allocator->ids == NULL)
+        return NULL;
+
+    for (int64 i = tid_allocator->pos - 1; i >= 0; i--)
```
yamt, line 29: int32 is enough

eloparco: but `tid_allocator->pos` is `uint32`, it may not fit into `int32`

core/iwasm/libraries/lib-wasi-threads/tid_allocator.h
```diff
@@ -0,0 +1,32 @@
...
+ * Copyright (C) 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+
+#ifndef _TID_ALLOCATOR_H
+#define _TID_ALLOCATOR_H
+
+#include "platform_common.h"
+
+#define TID_ALLOCATOR_INIT_SIZE CLUSTER_MAX_THREAD_NUM
```
loganek, line 11: This is only used in `tid_allocator.c`, so I'd suggest defining it there; if there's a need to use it in other places in the future, we can move it back here.

core/iwasm/libraries/lib-wasi-threads/tid_allocator.h
```diff
@@ -0,0 +1,32 @@
...
+tid_allocator_init();
+
+void
+tid_allocator_destroy(TidAllocator *tid_allocator);
+
+int32
+tid_allocator_get(TidAllocator *tid_allocator);
+
+void
+tid_allocator_put(TidAllocator *tid_allocator, int32 thread_id);
```
loganek, line 30: It might be subjective, but I think more natural function names would be `tid_allocator_(get_tid|release_tid)`. Up to you though.

core/iwasm/libraries/lib-wasi-threads/tid_allocator.c
```diff
@@ -0,0 +1,87 @@
...
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+
+#include "tid_allocator.h"
+#include "wasm_export.h"
+#include "bh_log.h"
+
+#define MIN(a, b) (((a) < (b)) ? (a) : (b))
+
+struct tidAllocator {
```
loganek, line 12: I think struct names should follow `PascalCase` (i.e. `TidAllocator`).

core/iwasm/libraries/lib-wasi-threads/tid_allocator.h
eloparco: As for `TID_MIN` and `TID_MAX`, I put it in the header since they are some parameters we can change to alter the behavior of the id allocator without looking into the `.c` implementation

core/iwasm/libraries/lib-wasi-threads/tid_allocator.h
eloparco: yes, that's better

core/iwasm/libraries/lib-wasi-threads/tid_allocator.h
```diff
@@ -0,0 +1,32 @@
...
+enum {
+    TID_MIN = 1,
+    TID_MAX = 0x1FFFFFFF
+}; // Reserved TIDs (WASI specification)
+
+/* Stack data structure to track available thread identifiers */
+typedef struct tidAllocator TidAllocator;
+
+TidAllocator *
+tid_allocator_init();
```
loganek, line 21: I think slightly better model in this particular case would be to leave allocation to the caller; i.e. the API would be something like:
```
bool tid_allocator_init(TidAllocator *);
void tid_allocator_deinit(TidAllocator *);
```
so then the caller can decide whether to allocate `tid_allocator` on the stack/in the data segment or on the heap (and in our case here, most likely we'd like to avoid dynamic allocations.

core/iwasm/libraries/lib-wasi-threads/tid_allocator.c
eloparco: I was using `TidAllocator` for the opaque struct in the header (i.e. `typedef struct tidAllocator TidAllocator;`) but actually I can use the same name for both

core/iwasm/libraries/lib-wasi-threads/tid_allocator.h
eloparco: good point, I'll have to expose the `TidAllocator` struct in the header, but it's better in our case so we avoid the dynamic allocation

core/iwasm/libraries/lib-wasi-threads/tid_allocator.c
```diff
@@ -7,40 +7,34 @@
...
-    uint32 pos;  // Index of the element after the stack top
-};
-
-TidAllocator *
-tid_allocator_init()
+bool
+tid_allocator_init(TidAllocator *tid_allocator)
 {
-    TidAllocator *tid_allocator = wasm_runtime_malloc(sizeof(TidAllocator));
+    memset(tid_allocator, 0, sizeof(*tid_allocator));
```
loganek, line 16: Not sure if that's really needed if you already set all the fields below?

loganek (approved):
LGTM

yamt (approved):

PR #1863, opened 2023-01-05
Enable more arch for github cached LLVM library

no1wudi:


.github/workflows/build_llvm_libraries.yml
```diff
@@ -35,5 +35,5 @@ jobs:
       - name: Build llvm
         id: build_llvm
         if: ${{ steps.cache_llvm.outputs.cache-hit != 'true' }} 
-        run: /usr/bin/env python3 ./build_llvm.py --arch X86 WebAssembly
+        run: /usr/bin/env python3 ./build_llvm.py --arch AArch64 ARM RISCV X86 WebAssembly
```
lum1n0us, line 38: Adding all five targets will dramatically increase the libraries compilation time. We'd better pass the minimum arch set with a argument.

lum1n0us: I am working on a PR about enhancing LLVM compilation in CI. And I think I can add this requirement in it if you don't mind.

no1wudi: I thinks it's OK since:
1. LLVM will only build once for almost PR
2. ARM and RISCV are used actullay in daily test
3. AArch64 will be used in daily test later

lum1n0us: #1870

PR #1865, opened 2023-01-05
Fix on-heap aux stack allocation

loganek:
Because stack grows from high address towards low address, the value returned by malloc is the end of the stack, not top of the stack. The top of the stack is the end of the allocated space (i.e. address returned by malloc + cluster size).

#1790

### DIFF
--- a/core/iwasm/libraries/thread-mgr/thread_manager.c
+++ b/core/iwasm/libraries/thread-mgr/thread_manager.c
@@ -83,11 +83,14 @@ allocate_aux_stack(WASMExecEnv *exec_env, uint32 *start, uint32 *size)
 #if WASM_ENABLE_HEAP_AUX_STACK_ALLOCATION != 0
     WASMModuleInstanceCommon *module_inst =
         wasm_exec_env_get_module_inst(exec_env);
+    uint32 stack_end;
 
-    *start = wasm_runtime_module_malloc(module_inst, cluster->stack_size, NULL);
+    stack_end =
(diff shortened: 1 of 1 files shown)


eloparco:
We need to remove these two lines https://github.com/bytecodealliance/wasm-micro-runtime/blob/dev/wasi_threads/samples/wasi-threads/wasm-apps/CMakeLists.txt#L32. I had to put them in this [PR](https://github.com/bytecodealliance/wasm-micro-runtime/blob/dev/wasi_threads/samples/wasi-threads/wasm-apps/CMakeLists.txt#L32) since there was a heap corruption problem that was making the pipeline fail.

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -116,15 +119,18 @@ allocate_aux_stack(WASMExecEnv *exec_env, uint32 *start, uint32 *size)
...
 free_aux_stack(WASMExecEnv *exec_env, uint32 start)
 {
+    WASMCluster *cluster = wasm_exec_env_get_cluster(exec_env);
+
 #if WASM_ENABLE_HEAP_AUX_STACK_ALLOCATION != 0
     WASMModuleInstanceCommon *module_inst =
         wasm_exec_env_get_module_inst(exec_env);
 
-    wasm_runtime_module_free(module_inst, start);
+    assert(start >= cluster->stack_size);
```
wenyongh, line 128: Should use bh_assert(), some platforms may not support assert()

yamt (approved):

PR #1867, opened 2023-01-06
Disable aux stack allocations for threads spawned by wasi_thread_start

loganek:
This syscall doesn't need allocating stack or TLS and it's expected from the application to do that instead. E.g. WASI-libc already does this for `pthread_create`.

I also fixed some of the examples to allocate memory for stack and not use stack before the stack pointer is set to a correct value.

### DIFF
--- a/core/iwasm/common/wasm_exec_env.h
+++ b/core/iwasm/common/wasm_exec_env.h
@@ -179,6 +179,13 @@ wasm_exec_env_create(struct WASMModuleInstanceCommon *module_inst,
 void
 wasm_exec_env_destroy(WASMExecEnv *exec_env);
 
+static inline bool
+wasm_exec_env_is_aux_stack_managed_by_runtime(WASMExecEnv *exec_env)
+{
+    return exec_env->aux_stack_boundary.boundary != 0
+           || exec_env->aux_stack_bottom.bottom != 0;
+}
+
 /**
  * Allocate a WASM frame from the WASM stack.
  *
--- a/core/iwasm/interpreter/wasm_interp_classic.c
+++ b/core/iwasm/interpreter/wasm_interp_classic.c
@@ -1783,14 +1783,18 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                 global = globals + global_idx;
                 global_addr = get_global_addr(global_data, global);
                 aux_stack_top = *(uint32 *)(frame_sp - 1);
-                if (aux_stack_top <= exec_env->aux_stack_boundary.boundary) {
-                    wasm_set_exception(module, "wasm auxiliary stack overflow");
-                    goto got_exception;
-                }
-                if (aux_stack_top > exec_env->aux_stack_bottom.bottom) {
-                    wasm_set_exception(module,
-                                       "wasm auxiliary stack underflow");
-                    goto got_exception;
+                if (wasm_exec_env_is_aux_stack_managed_by_runtime(exec_env)) {
+                    if (aux_stack_top
+                        <= exec_env->aux_stack_boundary.boundary) {
+                        wasm_set_exception(module,
+                                           "wasm auxiliary stack overflow");
+                        goto got_exception;
+                    }
+                    if (aux_stack_top > exec_env->aux_stack_bottom.bottom) {
+                        wasm_set_exception(module,
+                                           "wasm auxiliary stack underflow");
+                        goto got_exception;
+                    }
                 }
                 *(int32 *)global_addr = aux_stack_top;
                 frame_sp--;
--- a/samples/wasi-threads/wasm-apps/no_pthread.c
+++ b/samples/wasi-threads/wasm-apps/no_pthread.c
@@ -9,18 +9,20 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <assert.h>
-#include <wasi/api.h>
+
+#include "wasi_thread_start.h"
 
 static const int64_t SECOND = 1000 * 1000 * 1000;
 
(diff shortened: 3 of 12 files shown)


samples/wasi-threads/wasm-apps/no_pthread.c
```diff
@@ -9,18 +9,20 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <assert.h>
-#include <wasi/api.h>
+
+#include "wasi_thread_start.h"
```
yamt, line 13: where is this header?

samples/wasi-threads/wasm-apps/wasi_thread_start.S
```diff
@@ -0,0 +1,20 @@
+	.globaltype	__stack_pointer, i32
+	.functype	__wasi_thread_start_C (i32, i32) -> ()
+
+	.globl	wasi_thread_start
+
+wasi_thread_start:
```
yamt, line 6: if this file is a modified copy from wasi-libc, please mention it.

samples/wasi-threads/wasm-apps/no_pthread.c
loganek: added

samples/wasi-threads/wasm-apps/wasi_thread_start.S
loganek: updated

core/iwasm/common/wasm_exec_env.h
```diff
@@ -179,6 +179,13 @@ wasm_exec_env_create(struct WASMModuleInstanceCommon *module_inst,
 void
 wasm_exec_env_destroy(WASMExecEnv *exec_env);
 
+static inline bool
+wasm_exec_env_is_aux_stack_managed_by_runtime(WASMExecEnv *exec_env)
+{
+    return exec_env->aux_stack_boundary.boundary != 0
+           || exec_env->aux_stack_bottom.bottom != 0;
```
yamt, line 186: i feel it's simpler to use a dedicated flag as 0 is a valid address in wasm. how do you think?

also, i guess aot needs an equivalent around create_aux_stack_info.

loganek: > i feel it's simpler to use a dedicated flag as 0 is a valid address in wasm. how do you think?

Yeah, I did think about it. But I noticed that WAMR actually doesn't always respect 0 as a valid address; for example, it considers `wasm_runtime_malloc()` returning `0` as a failure. Flag was another option but didn't want to increase the size of the object unnecessarily if there was a way to achieve the same without it (if my assumption about WAMR not respecting 0 as a correct value holds).

> also, i guess aot needs an equivalent around create_aux_stack_info.

That's correct. At the moment we're not testing AOT though so I was planning to do that (and other necessary things) in the follow-up PR. Does it make sense?

yamt (approved):

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -1783,14 +1783,18 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
                 aux_stack_top = *(uint32 *)(frame_sp - 1);
-                if (aux_stack_top <= exec_env->aux_stack_boundary.boundary) {
-                    wasm_set_exception(module, "wasm auxiliary stack overflow");
-                    goto got_exception;
-                }
-                if (aux_stack_top > exec_env->aux_stack_bottom.bottom) {
-                    wasm_set_exception(module,
-                                       "wasm auxiliary stack underflow");
-                    goto got_exception;
+                if (wasm_exec_env_is_aux_stack_managed_by_runtime(exec_env)) {
```
wenyongh, line 1786: Do we need to apply similar changes for LLVM AOT/JIT and Fast JIT, refer to:

https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/compilation/aot_emit_variable.c#L192

https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/fast-jit/fe/jit_emit_variable.c#L264-L279

loganek: Yes we do; however, for now we focus on the interpreter and we'll work on JIT/AOT next.

wenyongh: OK, so let's merge this PR first?

loganek: Yes, we can merge this one and next we'll continue working on AOT and JIT

eloparco: @wenyongh @yamt for AOT, should we disable the aux stack globally when WASI threads are enabled? by replacing this 
https://github.com/bytecodealliance/wasm-micro-runtime/blob/21ba730de38672c2eebf1d60c7e98ab2dbf87501/core/iwasm/compilation/aot_llvm.c#L1614

with something like
```
#if WASM_ENABLE_LIB_WASI_THREADS != 0
        comp_ctx->enable_aux_stack_check = false;
#else
        comp_ctx->enable_aux_stack_check = true;
#endif
```

yamt: i guess it can be a workaround.
but i think it's better to avoid making the compiler depend on build-time configurations like this.

eloparco: at this point, should we just specify in the documentation to compile with `--disable-aux-stack-check` when using wasi threads?

PR #1868, opened 2023-01-06
Adding an extension to the Visual Studio Marketplace

fadss:
What 

Add an extension to the visual studio marketplace releasing new versions of the WAMR-IDE to download the extension from the marketplace .

Why

This allows for automation to take place because downloading and installing the extension is a manual process so it'd be  simpler if the extension was published automatically through the visual studio marketplace.

How

Alter the github action in build_wamr_vscode_ext.yml to publish to an account created for the WAMR publisher to maintain and a personal access token(PAT) that publishes to the account below. For every run of the binary release process job the WAMR-IDE is published to the registered account 
 
Record of the PAT is stored and will be sent over. 
 https://dev.azure.com/wamr-vs-marketplace/Wamr%20Extension.

### DIFF
--- a/.github/workflows/build_wamr_vscode_ext.yml
+++ b/.github/workflows/build_wamr_vscode_ext.yml
@@ -32,11 +32,14 @@ jobs:
         working-directory: test-tools/wamr-ide/VSCode-Extension
 
       - name: generate wamr ide vscode extension
+        env: 
+          credentials: ${{ secrets.TOKEN }}
         run: |
           npm install -g vsce
           rm -rf node_modules
           npm install
           vsce package
+          vsce publish -p ${{ secrets.TOKEN }}
         working-directory: test-tools/wamr-ide/VSCode-Extension
 
       - name: compress the vscode extension
--- a/.github/workflows/release_process.yml
+++ b/.github/workflows/release_process.yml
@@ -150,6 +150,7 @@ jobs:
   release_wamr_ide_vscode_ext:
     needs: [create_tag, create_release]
     uses: ./.github/workflows/build_wamr_vscode_ext.yml
+    secrets: inherit
     with:
       upload_url: ${{ needs.create_release.outputs.upload_url }}
       ver_num: ${{ needs.create_tag.outputs.new_ver }}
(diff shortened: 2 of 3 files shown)


loganek:
Many thanks for doing that!

Could we update a readme file with information about the Azure organization that owns the extension and who has access to it? Also, did we agree on the owner of the organization? 

I clicked a link but don't seem to have access: 
> Your account, marcin.kolny@gmail.com is not authorized to view this page. Make sure the URL is correct and your account has access.

.github/workflows/build_wamr_vscode_ext.yml
```diff
@@ -32,11 +32,14 @@ jobs:
...
 
       - name: generate wamr ide vscode extension
+        env: 
+          credentials: ${{ secrets.TOKEN }}
         run: |
           npm install -g vsce
           rm -rf node_modules
           npm install
           vsce package
+          vsce publish -p ${{ secrets.TOKEN }}
```
loganek, line 42: does it mean the new version will be published on every build? If so, in general I like the idea of CI/CD, but I wonder if we need to put some extra checks in the extension to make sure customers use the correct version of the WAMR <-> vscode extension; I guess there might be features we add to unreleased version of WAMR that will break the use of the extension or make it non functioning correctly.

loganek, line 42: Given we now publish the extension to a market place, I wonder if we'll need some mechanism to check compatibility between VSCode version and WAMR version? It might be a problem already if some users use old version of WAMR (and will be a problem if there will be features added to VSCode extension that make it incompatible with older WAMR versions).

fadss: Thanks, I've added your email to the email chain, does the link work now ? I made the project private instead of public.

Who would have access to it, unsure of who specifically.

I believe we haven't agreed on the owner of the organization 

I

wenyongh: @fadss Many thanks for uploading the PR! Could you help grant the access for `Jun1.Xu@intel.com`, `ning4.wang@intel.com` and `tianlong.liang@intel.com`? They are the main maintainers of WAMR-IDE.

fadss: Added to the chain, do you have access now ?

wenyongh: > Added to the chain, do you have access now ?

Yes, thanks a lot!

PR #1869, opened 2023-01-06
Handle thread exit

eloparco:
Implementation of what's described here: https://github.com/WebAssembly/wasi-threads/pull/17:
- a trap or `proc_exit` in the main thread should stop all the threads in the process
- a thread created with wasi_thread_spawn that calls `proc_exit` should terminate all the threads in the process
 
The sample has been modified to test combinations of trap or `proc_exit` for termination and termination forced in the main thread or a spawn thread.

Part of #1790.

### DIFF
--- a/samples/wasi-threads/wasm-apps/thread_termination.c
+++ b/samples/wasi-threads/wasm-apps/thread_termination.c
@@ -15,6 +15,9 @@
 
 #include "wasi_thread_start.h"
 
+#define TEST_TERMINATION_BY_TRAP 0 // Otherwise test `proc_exit` termination
+#define TEST_TERMINATION_IN_MAIN_THREAD 1
+
 #define TIMEOUT_SECONDS 10
 #define NUM_THREADS 3
 static sem_t sem;
@@ -27,7 +30,7 @@ typedef struct {
 void
 run_long_task()
 {
-    // Busy waiting to be interruptible by exception
+    // Busy waiting to be interruptible by trap or `proc_exit`
     for (int i = 0; i < TIMEOUT_SECONDS; i++)
         sleep(1);
 }
@@ -39,18 +42,22 @@ __wasi_thread_start_C(int thread_id, int *start_arg)
 
     if (data->throw_exception) {
         // Wait for all other threads (including main thread) to be ready
-        printf("Waiting before throwing exception\n");
+        printf("Waiting before terminating\n");
         for (int i = 0; i < NUM_THREADS; i++)
             sem_wait(&sem);
 
-        printf("Throwing exception\n");
+        printf("Force termination\n");
+#if TEST_TERMINATION_BY_TRAP == 1
         __builtin_trap();
+#else
+        __wasi_proc_exit(1);
+#endif
     }
     else {
         printf("Thread running\n");
 
         sem_post(&sem);
-        run_long_task(); // Wait to be interrupted by exception
+        run_long_task(); // Wait to be interrupted
         assert(false && "Unreachable");
     }
 }
@@ -74,19 +81,26 @@ main(int argc, char **argv)
         }
     }
 
-    // Create a thread that throws an exception
+// Create a thread that forces termination through trap or `proc_exit`
+#if TEST_TERMINATION_IN_MAIN_THREAD == 1
+    data[0].throw_exception = false;
+#else
     data[0].throw_exception = true;
+#endif
     thread_id = __wasi_thread_spawn(&data[0]);
     if (thread_id < 0) {
         printf("Failed to create thread: %d\n", thread_id);
         return EXIT_FAILURE;
     }
+
     // Create two additional threads to test exception propagation
+    data[1].throw_exception = false;
     thread_id = __wasi_thread_spawn(&data[1]);
     if (thread_id < 0) {
         printf("Failed to create thread: %d\n", thread_id);
         return EXIT_FAILURE;
     }
+    data[2].throw_exception = false;
     thread_id = __wasi_thread_spawn(&data[2]);
     if (thread_id < 0) {
         printf("Failed to create thread: %d\n", thread_id);
@@ -96,8 +110,19 @@ main(int argc, char **argv)
     printf("Main thread running\n");
 
     sem_post(&sem);
-    run_long_task(); // Wait to be interrupted by exception
-    assert(false && "Unreachable");
 
+#if TEST_TERMINATION_IN_MAIN_THREAD == 1
+
+    printf("Force termination (main thread)\n");
+#if TEST_TERMINATION_BY_TRAP == 1
+    __builtin_trap();
+#else  /* TEST_TERMINATION_BY_TRAP */
+    __wasi_proc_exit(1);
+#endif /* TEST_TERMINATION_BY_TRAP */
+
+#else  /* TEST_TERMINATION_IN_MAIN_THREAD */
+    run_long_task(); // Wait to be interrupted
+    assert(false && "Unreachable");
+#endif /* TEST_TERMINATION_IN_MAIN_THREAD */
     return EXIT_SUCCESS;
 }
--- a/samples/wasi-threads/wasm-apps/CMakeLists.txt
+++ b/samples/wasi-threads/wasm-apps/CMakeLists.txt
@@ -36,4 +36,4 @@ function (compile_sample SOURCE_FILE)
 endfunction ()
 
 compile_sample(no_pthread.c wasi_thread_start.S)
-compile_sample(exception_propagation.c wasi_thread_start.S)
+compile_sample(thread_termination.c wasi_thread_start.S)


core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -2157,7 +2157,6 @@ wasm_call_function(WASMExecEnv *exec_env, WASMFunctionInstance *function,
     wasm_exec_env_set_thread_info(exec_env);
 
     interp_call_wasm(module_inst, exec_env, function, argc, argv);
-    (void)clear_wasi_proc_exit_exception(module_inst);
```
wenyongh, line 2160: Not sure why remove this line?

The wasi proc_exit native API is to terminate the process normally, but if runtime exits directly in the native API wasi_proc_exit of libc-wasi, some resources won't be destroyed, and there might be other threads running with the same module instance, normally runtime should wait for all threads of the module instance to terminate before destroy the module instance. So currently runtime just throws exception "wasi proc exit" in the native API, and clear this exception after call wasm funtion to make it exit normally.

For example, when using emsdk to build a "hello world" wasm app, normally it calls proc_exit at the end. If we don't throw "wasi proc exit" exception in the native API, then the wasm app will continue to execute the successive opcodes and "unreachable" exception will be thrown instead.

And if we don't clear "wasi_proc_exception", it will be thrown.

core/iwasm/common/wasm_application.c
```diff
@@ -107,7 +107,12 @@ execute_main(WASMModuleInstanceCommon *module_inst, int32 argc, char *argv[])
        the actual main function. Directly calling main function
        may cause exception thrown. */
     if ((func = wasm_runtime_lookup_wasi_start_function(module_inst))) {
-        return wasm_runtime_call_wasm(exec_env, func, 0, NULL);
+        int ret = wasm_runtime_call_wasm(exec_env, func, 0, NULL);
+#if WASM_ENABLE_THREAD_MGR != 0
+        if (!ret && wasm_runtime_get_exception(module_inst))
+            wasm_cluster_spread_exception(exec_env);
```
wenyongh, line 113: Should we apply the same changes to the main branch?
And should we ignore spreading the exception "wasi proc exit" since it is a normal behavior and throwing this exception is a special handling of WAMR?

eloparco: This change is to propagate `proc_exit` and traps from the main thread, I can put an additional `WASM_ENABLE_LIB_WASI_THREADS != 0` at L111 to avoid changing the normal behavior when wasi threads are not involved

core/iwasm/interpreter/wasm_runtime.c
eloparco: I was commenting it so that after calling `wasm_runtime_call_wasm` we can check if a `proc_exit` occurred to propagate it; since with wasi threads we have one instance per thread, do you think it would be enough to just wrap the changes with `WASM_ENABLE_LIB_WASI_THREADS`?
If not, what's your suggestion? I started contributing to this project not long ago so any clarification/suggestion is much appreciated

core/iwasm/common/wasm_application.c
wenyongh: How about applying changes in wasm_runtime_call_wasm since here it is to call the _start function of wasm app, but developer may call other functions to start the app which may also create new threads. 

https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/common/wasm_runtime_common.c#L1788-L1789

For example, before return false, add:
```C
#if WASM_ENABLE_THREAD_MGR != 0 && WASM_ENABLE_LIB_WASI_THREADS != 0
        if (wasm_runtime_get_exception(module_inst))
            wasm_cluster_spread_exception(exec_env);
#endif
```

core/iwasm/interpreter/wasm_runtime.c
wenyongh: There are some places calling clear_wasi_proc_exit_exception, I think if you need to disable the clear exception action, had better just modify the clear_wasi_proc_exit_exception itself, in both wasm_runtime.c and aot_runtime:

https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_runtime.c#L2037

https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/aot/aot_runtime.c#L903

core/iwasm/common/wasm_application.c
eloparco: yes that makes more sense, thanks

wenyongh (approved):
LGTM

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -2033,7 +2033,9 @@ wasm_lookup_table(const WASMModuleInstance *module_inst, const char *name)
 static bool
 clear_wasi_proc_exit_exception(WASMModuleInstance *module_inst)
 {
-#if WASM_ENABLE_LIBC_WASI != 0
+    /* With WASI threads, proc_exit exception is not cleared because it has to
```
loganek, line 2036: should we then capture the exception and clear it for each module instances?

eloparco: I think I need to modify `wasm_runtime_call_wasm()` to return `true` in case the exception encountered is `proc_exit`

samples/wasi-threads/wasm-apps/thread_termination.c
```diff
@@ -27,7 +30,7 @@ typedef struct {
 void
 run_long_task()
 {
-    // Busy waiting to be interruptible by exception
+    // Busy waiting to be interruptible by trap or `proc_exit`
     for (int i = 0; i < TIMEOUT_SECONDS; i++)
```
eloparco, line 34: @wenyongh btw I notice something while testing: if I replace this for loop with a `__builtin_wasm_memory_atomic_wait32(0, 0, -1);` or even `while(1);` the thread doesn't get interrupted even if the exception is propagated (in the first case it gets stuck here https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/shared/platform/common/posix/posix_thread.c#L192). Is that expected? I'll investigate to understand what's happening

core/iwasm/interpreter/wasm_runtime.c
yamt: my expectation on `proc_exit` is:

* wasm_application_execute_main etc succeeds without an exception
* exit code is propagated here https://github.com/bytecodealliance/wasm-micro-runtime/blob/219713bfbc5196ee7c91fe78d18ddceed069c179/product-mini/platforms/posix/main.c#L669

samples/wasi-threads/wasm-apps/thread_termination.c
yamt: it's what i expect to happen with the current implementation.

to terminate busy looping threads, something like `wasm_cluster_thread_send_signal` would work.
i guess we eventually need to implement something (eg. send a signal on posix-like platforms) to terminate the blocking threads.

wenyongh: > @wenyongh btw I notice something while testing: if I replace this for loop with a `__builtin_wasm_memory_atomic_wait32(0, 0, -1);` or even `while(1);` the thread doesn't get interrupted even if the exception is propagated (in the first case it gets stuck here https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/shared/platform/common/posix/posix_thread.c#L192). Is that expected? I'll investigate to understand what's happening

Normally runtime should terminate all threads when exception was thrown, I remember it is the expected behavior of Java VM. Seems it is an issue of the main branch, I try to fix it with the patch below for interpreter. And another issue of main branch may be that the main thread doesn't spread exception to other thread when exception was thrown, like the change of this PR in wasm_runtime_call_wasm. I will discuss with @xujuntwt95329 for how to fix them.
```diff
diff --git a/core/iwasm/interpreter/wasm_interp_fast.c b/core/iwasm/interpreter/wasm_interp_fast.c
index 3109b0c8..36863bf2 100644
--- a/core/iwasm/interpreter/wasm_interp_fast.c
+++ b/core/iwasm/interpreter/wasm_interp_fast.c
@@ -1054,6 +1054,8 @@ wasm_interp_call_func_import(WASMModuleInstance *module_inst,
 #if WASM_ENABLE_THREAD_MGR != 0
 #define CHECK_SUSPEND_FLAGS()                           \
     do {                                                \
+        if (wasm_get_exception(module))                 \
+            goto got_exception;                         \
         if (exec_env->suspend_flags.flags != 0) {       \
             if (exec_env->suspend_flags.flags & 0x01) { \
```

wenyongh: @eloparco I discussed the issues with @xujuntwt95329, he agreed that they are issues and will submit patch to fix them on main branch. Could you wait until we fix them and merge branch `main` into `dev/wasi_threads`, so as to sync up the code and avoid conflicts in the future?

core/iwasm/interpreter/wasm_runtime.c
loganek: > I get an exception so the runtime doesn't return 0 at the end 

> it seems like the exit code gets already propagated properly

I'm a bit confused, so does it return the exit code being passed to `proc_exit()`?

eloparco: yes, the return code is set when `proc_exit` is called https://github.com/bytecodealliance/wasm-micro-runtime/blob/903f5212f194c2709b32631bd32936fdcfed27af/core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c#L989 and then retrieved here https://github.com/bytecodealliance/wasm-micro-runtime/blob/219713bfbc5196ee7c91fe78d18ddceed069c179/product-mini/platforms/posix/main.c#L669
I tested it by calling `__wasi_proc_exit(exit_code)` with different exit codes and I see the `iwasm` executable returning with the proper exit codes

samples/wasi-threads/wasm-apps/thread_termination.c
eloparco: as proposed by @yamt, I tried adding a `wasm_cluster_send_signal_all(cluster, WAMR_SIG_TERM);` after spreading the exception https://github.com/bytecodealliance/wasm-micro-runtime/blob/903f5212f194c2709b32631bd32936fdcfed27af/core/iwasm/libraries/thread-mgr/thread_manager.c#L995 and it solves in case of a `while(1);` but not for other cases like `sleep(TIMEOUT_SECONDS);` or `__builtin_wasm_memory_atomic_wait32(0, 0, -1);`

yamt: > but not for other cases like `sleep(TIMEOUT_SECONDS);` or `__builtin_wasm_memory_atomic_wait32(0, 0, -1);`

i guess something more complex is necessary for blocking threads.
like use `pthread_kill` and make them handle `EINTR`.

for another runtime i made all i/o non-blocking to handle thread termination requests. but i don't think it's an appropriate approach for WAMR.

xujuntwt95329: @eloparco Yes, `wasm_cluster_send_signal_all` can only tell the interpreter to exit the execution loop, so `while(1)` can be terminated, but for `sleep` and `__builtin_wasm_memory_atomic_wait32`, they are not executed in interpreter, they have already fall into OS kernel, so the `wasm_cluster_send_signal_all` will not work for this situation.

I agree with @yamt that we may need some more complex mechanism to terminate the threads when they are blocked in kernel.

loganek: Is that also a problem in lib_pthread? I wonder if we should fix it in the `main` branch as well.

eloparco: > Is that also a problem in lib_pthread?

yes; it applies in general to threads running "blocking" instructions like `sleep()` or `__builtin_wasm_memory_atomic_wait32()`

> I wonder if we should fix it in the main branch as well.

I think that's what @wenyongh was proposing a few comments above

core/iwasm/aot/aot_runtime.c
```diff
@@ -902,7 +902,9 @@ create_exports(AOTModuleInstance *module_inst, AOTModule *module,
 static bool
 clear_wasi_proc_exit_exception(AOTModuleInstance *module_inst)
 {
-#if WASM_ENABLE_LIBC_WASI != 0
+    /* With WASI threads, proc_exit exception is not cleared because it has to
+     * be detected and propagated to the other threads in the process */
+#if WASM_ENABLE_LIBC_WASI != 0 && WASM_ENABLE_LIB_WASI_THREADS == 0
```
xujuntwt95329, line 907: Seems `lib-pthread` also need to spread the proc exit exception, I'll modify this line to `#if WASM_ENABLE_LIBC_WASI != 0 && WASM_ENABLE_THREAD_MGR == 0` in main branch

samples/wasi-threads/wasm-apps/thread_termination.c
wenyongh: @loganek @eloparco Jun's PR #1889 was merged and I tried to create a PR to merge branch main into dev/wasi_threads, but unfortunately there are conflicts in below files:
```
        .github/workflows/compilation_on_android_ubuntu.yml
        core/iwasm/libraries/thread-mgr/thread_manager.c
        core/iwasm/libraries/thread-mgr/thread_manager.h
```
I can resolve these conflicts, and the diff file of the PR is as below:
[merge_main_into_wasi_threads.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/10462938/merge_main_into_wasi_threads.zip)

Could you help check whether the patch works for you? If yes, I will push it manually under the command line. Thanks.

loganek: Hi @wenyongh , I reviewed the patch and it is mostly correct, except one thing. In function `thread_manager_start_routine()` you moved `free_aux_stack` below `wasm_runtime_deinstantiate_internal`; `free_aux_stack` must be executed first because it uses module instance. 

Also, one minor thing about `wasm_cluster_add_exec_env` - the `ret` variable is no longer needed so we could remove it; but that's not a big deal.

You can either update your merge commit or feel free to merge it as it is, and I'll open a PR to address the problem.

wenyongh: @loganek Thanks for the review, I changed the wasm_runtime_deinstantiate_internal in thread_manager_start_routine. But for the wasm_cluster_add_exec_env, not sure why `ret` can be removed.
I pushed the PR, see: https://github.com/bytecodealliance/wasm-micro-runtime/commit/c7141894fb7b5ef5d39f13748382c9138ee945de

loganek: > But for the wasm_cluster_add_exec_env, not sure why ret can be removed.

It can be replaced with `return` statements (we no longer have a lock to unlock). As I said, it's not a big deal though. The rest looks ok for me.

loganek: > I agree with @yamt that we may need some more complex mechanism to terminate the threads when they are blocked in kernel.

@xujuntwt95329 / @wenyongh does anybody on your side actively working on that? If not, we'll take on that work.

wenyongh: @loganek not yet

eloparco:
We can probably merge this PR and I'll open a separate issue to track the possible race condition that I documented above (and I've never been able to reproduce again).

PR #1873, opened 2023-01-09
Enable running mode control for runtime and module instance

TianlongLiang:
Subtasks:
- CLI options: four options for running mode for iwasm and APIs to set the runtime running mode
- APIs to set the module instance running mode
- CLI options: two options for LLVM JIT

This is draft pr, subtasks three in one.

### DIFF
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -132,7 +132,7 @@ static JitCompOptions jit_options = { 0 };
 static LLVMJITOptions llvm_jit_options = { 3, 3 };
 #endif
 
-static RunningMode runtime_running_mode = { 0 };
+static RunningMode runtime_running_mode = Mode_Default;
 
 #ifdef OS_ENABLE_HW_BOUND_CHECK
 /* The exec_env of thread local storage, set before calling function
@@ -587,7 +587,7 @@ wasm_runtime_full_init(RuntimeInitArgs *init_args)
 bool
 wasm_runtime_is_running_mode_supported(RunningMode running_mode)
 {
-    if (running_mode == 0) {
+    if (running_mode == Mode_Default) {
         return true;
     }
     else if (running_mode == Mode_Interp) {
@@ -1255,14 +1255,16 @@ wasm_runtime_set_running_mode(wasm_module_inst_t module_inst,
         WASMModuleInstance *module_inst_interp =
             (WASMModuleInstance *)module_inst;
 
-        if (running_mode == 0) {
+        if (running_mode == Mode_Default) {
 #if WASM_ENABLE_FAST_JIT == 0 && WASM_ENABLE_JIT == 0
             running_mode = Mode_Interp;
 #elif WASM_ENABLE_FAST_JIT != 0 && WASM_ENABLE_JIT == 0
             running_mode = Mode_Fast_JIT;
-#elif WASM_ENABLE_FAST_JIT == 0 && WASM_ENABLE_JIT != 0
+#elif (WASM_ENABLE_FAST_JIT == 0 || WASM_ENABLE_LAZY_JIT == 0) \
+    && WASM_ENABLE_JIT != 0
             running_mode = Mode_LLVM_JIT;
-#elif WASM_ENABLE_FAST_JIT != 0 && WASM_ENABLE_JIT != 0
+#elif WASM_ENABLE_FAST_JIT != 0 && WASM_ENABLE_JIT != 0 \
+    && WASM_ENABLE_LAZY_JIT != 0
             running_mode = Mode_Multi_Tier_JIT;
 #endif
         }
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -606,7 +606,8 @@ wasm_runtime_is_running_mode_supported(RunningMode running_mode)
 #endif
     }
     else if (running_mode == Mode_Multi_Tier_JIT) {
-#if WASM_ENABLE_FAST_JIT != 0 && WASM_ENABLE_JIT != 0
+#if WASM_ENABLE_FAST_JIT != 0 && WASM_ENABLE_JIT != 0 \
+    && WASM_ENABLE_LAZY_JIT != 0
         return true;
 #endif
     }
@@ -1260,12 +1261,14 @@ wasm_runtime_set_running_mode(wasm_module_inst_t module_inst,
             running_mode = Mode_Interp;
 #elif WASM_ENABLE_FAST_JIT != 0 && WASM_ENABLE_JIT == 0
             running_mode = Mode_Fast_JIT;
-#elif (WASM_ENABLE_FAST_JIT == 0 || WASM_ENABLE_LAZY_JIT == 0) \
-    && WASM_ENABLE_JIT != 0
+#elif WASM_ENABLE_FAST_JIT == 0 && WASM_ENABLE_JIT != 0
             running_mode = Mode_LLVM_JIT;
-#elif WASM_ENABLE_FAST_JIT != 0 && WASM_ENABLE_JIT != 0 \
-    && WASM_ENABLE_LAZY_JIT != 0
+#else /* WASM_ENABLE_FAST_JIT != 0 && WASM_ENABLE_JIT != 0 */
+#if WASM_ENABLE_LAZY_JIT == 0
+            running_mode = Mode_LLVM_JIT;
+#else
             running_mode = Mode_Multi_Tier_JIT;
+#endif
 #endif
         }
         module_inst_interp->e->running_mode = running_mode;
@@ -1284,12 +1287,6 @@ wasm_runtime_set_running_mode(wasm_module_inst_t module_inst,
 RunningMode
 wasm_runtime_get_running_mode(wasm_module_inst_t module_inst)
 {
-
-#if WASM_ENABLE_AOT != 0
-    if (module_inst->module_type == Wasm_Module_AoT)
-        return 0;
-#endif
-
 #if WASM_ENABLE_INTERP != 0
     if (module_inst->module_type == Wasm_Module_Bytecode) {
         WASMModuleInstance *module_inst_interp =
@@ -1298,7 +1295,7 @@ wasm_runtime_get_running_mode(wasm_module_inst_t module_inst)
     }
 #endif
 
-    return 0;
+    return Mode_Default;
 }
 
 void
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -132,9 +132,7 @@ static JitCompOptions jit_options = { 0 };
 static LLVMJITOptions llvm_jit_options = { 3, 3 };
 #endif
 
-#if WASM_ENABLE_JIT != 0 || WASM_ENABLE_FAST_JIT != 0
-static RuntimeOptions runtime_options = { 0 };
-#endif
+static RunningMode runtime_running_mode = { 0 };
 
 #ifdef OS_ENABLE_HW_BOUND_CHECK
 /* The exec_env of thread local storage, set before calling function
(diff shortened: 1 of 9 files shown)


core/iwasm/common/wasm_runtime_common.c
```diff
@@ -554,6 +589,41 @@ wasm_runtime_full_init(RuntimeInitArgs *init_args)
...
+#if WASM_ENABLE_FAST_JIT != 0
+    ret |= (running_mode == Mode_Fast_JIT);
+#endif
+#if WASM_ENABLE_JIT != 0
+    ret |= (running_mode == Mode_LLVM_JIT);
+#endif
+#if WASM_ENABLE_FAST_JIT != 0 && WASM_ENABLE_JIT != 0 \
+    && WASM_ENABLE_LAZY_JIT != 0
+    ret |= (running_mode == Mode_Multi_Tier_JIT);
+#endif
```
wenyongh, line 611: How about check according to running mode:
```C
if (running_mode == 0) {
    return true;
}
elsel if (running_mode == Mode_Interp) {
#if WASM_ENABLE_INTERP != 0
    return true;
#endif
}
else if (running_mode == Mode_Fast_JIT) {
#if WASM_ENABLE_FAST_JIT != 0
    return true;
#endif
}
else if (running_mode == Mode_LLVM_JIT) {
#if WASM_ENABLE_LLVM_JIT != 0
    return true;
#endif
}
else if (running_mode == Mode_Multi_Tier_JIT) {
#if WASM_ENABLE_FAST_JIT != 0 && WASM_ENABLE_LLVM_JIT != 0
    return true;
#endif
}

return false;

File: core/iwasm/common/wasm_runtime_common.c [hunk 2]
```diff
@@ -554,6 +589,41 @@ wasm_runtime_full_init(RuntimeInitArgs *init_args)
...
+    return ret;
+}
+
+bool
+wasm_runtime_set_default_running_mode(RunningMode running_mode)
+{
+#if WASM_ENABLE_FAST_JIT != 0 || WASM_ENABLE_JIT != 0
```
wenyongh, line 618: Remove macro control

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1171,6 +1241,28 @@ wasm_runtime_deinstantiate_internal(WASMModuleInstanceCommon *module_inst,
 #endif
 }
 
+bool
+wasm_runtime_set_running_mode(wasm_module_inst_t module_inst,
+                              RunningMode running_mode)
+{
+#if WASM_ENABLE_FAST_JIT != 0 || WASM_ENABLE_JIT != 0
```
wenyongh, line 1248: Remove macro control

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1171,6 +1241,28 @@ wasm_runtime_deinstantiate_internal(WASMModuleInstanceCommon *module_inst,
...
+    if (wasm_runtime_is_running_mode_supported(running_mode)) {
+        WASMModuleInstance *actual_module_inst =
+            (WASMModuleInstance *)module_inst;
+        actual_module_inst->e->running_mode = running_mode;
```
wenyongh, line 1252: 1. Module instance can be AOTModuleInstance, should not convert to WASMModuleInstance directly.

2. Had better to convert running mode 0 to the actual running mode

Code may be like:
```C
{
#if WASM_ENABLE_AOT != 0
    if (module_inst->type == Wasm_Module_AoT)
        return true;
#endif

#if WASM_ENABLE_INTERP != 0
    if (module_inst->type == Wasm_Module_Bytecode) {
        WASMModuleInstance *module_inst_interp =
            (WASMModuleInstance *)module_inst;

        if (running_mode == 0) {
#if WASM_ENABLE_FAST_JIT == 0 && WASM_ENABLE_LLVM_JIT == 0
            running_mode = Mode_Interp;
#elif WASM_ENABLE_FAST_JIT != 0 && WASM_ENABLE_LLVM_JIT == 0
            running_mode = Mode_Fast_JIT;
#elif WASM_ENABLE_FAST_JIT == 0 && WASM_ENABLE_LLVM_JIT != 0
            running_mode = Mode_LLVM_JIT;
#elif WASM_ENABLE_FAST_JIT != 0 && WASM_ENABLE_LLVM_JIT != 0
            running_mode = Mode_Multi_Tier_JIT;
#endif
        }
        module_inst_interp->e->running_mode = running_mode;

        /* And here we should handle the function pointers, and call this API when instantiating the wasm module in wasm_runtime.c, so we don't need to modify the function pointers in wasm_interp_classic.c, let's do it later */
        return true;
    }
#endif

    return false;
}
```

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1171,6 +1241,28 @@ wasm_runtime_deinstantiate_internal(WASMModuleInstanceCommon *module_inst,
...
+    }
+#endif
+    return false;
+}
+
+RunningMode
+wasm_runtime_get_running_mode(wasm_module_inst_t module_inst)
+{
+    WASMModuleInstance *actual_module_inst = (WASMModuleInstance *)module_inst;
+    return actual_module_inst->e->running_mode;
```
wenyongh, line 1263: Should check whether module inst is AOT or Bytecode, if it is the former, return 0.

core/iwasm/common/wasm_runtime_common.h
```diff
@@ -413,6 +413,19 @@ typedef struct wasm_frame_t {
     const char *func_name_wp;
 } WASMCApiFrame;
 
+#if WASM_ENABLE_JIT != 0 || WASM_ENABLE_FAST_JIT != 0
+typedef struct RuntimeOptions {
+    RunningMode running_mode;
+} RuntimeOptions;
+#endif
```
wenyongh, line 420: Remove the structure.

core/iwasm/common/wasm_runtime_common.h
```diff
@@ -437,10 +450,30 @@ wasm_runtime_get_exec_env_tls(void);
 WASM_RUNTIME_API_EXTERN bool
 wasm_runtime_init(void);
 
+#if WASM_ENABLE_JIT != 0 || WASM_ENABLE_FAST_JIT != 0
```
wenyongh, line 453: Remove the macro control

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -1183,7 +1183,10 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                 goto got_exception;
             }
 
-            HANDLE_OP(WASM_OP_NOP) { HANDLE_OP_END(); }
+            HANDLE_OP(WASM_OP_NOP)
+            {
+                HANDLE_OP_END();
+            }
```
wenyongh, line 1189: Use clang-format-12 to format the code

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -4226,56 +4232,107 @@ wasm_interp_call_wasm(WASMModuleInstance *module_inst, WASMExecEnv *exec_env,
...
+#if WASM_ENABLE_JIT != 0 && WASM_ENABLE_FAST_JIT != 0 \
+    && WASM_ENABLE_LAZY_JIT != 0
+            running_mode = Mode_Multi_Tier_JIT;
+#elif WASM_ENABLE_JIT != 0
+            running_mode = Mode_LLVM_JIT;
+#elif WASM_ENABLE_FAST_JIT != 0
+            running_mode = Mode_Fast_JIT;
+#endif
+        }
+#endif
```
wenyongh, line 4261: running mode should have been resolved and set in wasm_instantiate when instantiating or wasm_runtime_set_running_mode, no need to resolve it again.

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -3027,8 +3027,11 @@ init_llvm_jit_functions_stage1(WASMModule *module, char *error_buf,
     }
 
     option.is_jit_mode = true;
-    option.opt_level = 3;
-    option.size_level = 3;
+
+    LLVMJITOptions llvm_jit_options = wasm_runtime_get_llvm_jit_options();
```
wenyongh, line 3031: Put the variable definition at the beginning of the function.

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -1419,6 +1419,8 @@ wasm_instantiate(WASMModule *module, bool is_sub_inst, uint32 stack_size,
     module_inst->e =
         (WASMModuleInstanceExtra *)((uint8 *)module_inst + extra_info_offset);
 
+    module_inst->e->running_mode = 0;
```
wenyongh, line 1422: Had better `wasm_runtime_set_running_mode(module_inst, wasm_runtime_get_default_running_mode())`, which resolves the running mode 0 to actual running mode.

core/iwasm/interpreter/wasm_runtime.h
```diff
@@ -242,6 +242,7 @@ typedef struct WASMModuleInstanceExtra {
         && WASM_ENABLE_LAZY_JIT != 0)
     WASMModuleInstance *next;
 #endif
+    RunningMode running_mode;
```
wenyongh, line 245: How about moving it after `CApiFuncImport *c_api_func_imports;`, L221

product-mini/platforms/posix/main.c
```diff
@@ -378,6 +394,8 @@ main(int argc, char *argv[])
     int instance_port = 0;
 #endif
 
+    memset(&init_args, 0, sizeof(RuntimeInitArgs));
```
wenyongh, line 397: Define running_mnode = 0 at the begining of the function and don't move the line here.

product-mini/platforms/posix/main.c
```diff
@@ -387,6 +405,49 @@ main(int argc, char *argv[])
             }
             func_name = argv[0];
         }
+#if WASM_ENABLE_JIT != 0 || WASM_ENABLE_FAST_JIT != 0
+        else if (!strcmp(argv[0], "--interp")) {
+            init_args.running_mode = Mode_Interp;
```
wenyongh, line 410: Define running_mnode = 0 at the begining of the function and here `running_mode = Mode_Interp`.

product-mini/platforms/posix/main.c
```diff
@@ -387,6 +405,49 @@ main(int argc, char *argv[])
             }
             func_name = argv[0];
         }
+#if WASM_ENABLE_JIT != 0 || WASM_ENABLE_FAST_JIT != 0
```
wenyongh, line 408: Seems no need to add macro control

product-mini/platforms/posix/main.c
```diff
@@ -387,6 +405,49 @@ main(int argc, char *argv[])
...
+#if WASM_ENABLE_JIT != 0
+        else if (!strncmp(argv[0], "--size-level=", 13)) {
+            if (argv[0][13] == '\0')
+                return print_help();
+            init_args.llvm_jit_size_level = atoi(argv[0] + 13);
+            if (init_args.llvm_jit_size_level < 1
+                || init_args.llvm_jit_size_level > 3) {
+                printf("LLVM JIT size level should in range [1,3]\n");
+                return 1;
+            }
```
wenyongh, line 438: How about setting it to 1 when size_level is smaller than 1?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -554,6 +584,46 @@ wasm_runtime_full_init(RuntimeInitArgs *init_args)
     return true;
 }
 
+bool
+wasm_runtime_is_running_mode_supported(RunningMode running_mode)
+{
+    if (running_mode == 0) {
```
yamt, line 590: what's 0?

product-mini/platforms/posix/main.c
```diff
@@ -387,6 +402,57 @@ main(int argc, char *argv[])
             }
             func_name = argv[0];
         }
+        else if (!strcmp(argv[0], "--interp")) {
+            running_mode = Mode_Interp;
+        }
```
yamt, line 407: i think you can build wamr without interpreter. (ie. aot-only)

core/iwasm/common/wasm_runtime_common.c
yamt: if it means the default, i guess it's better to have `Mode_Default` instead.

TianlongLiang: Yes, `Mode_Default` is more readable. The reason for using a 0 rather than adding a `Mode_Default` in the enum type is that I think it's maybe confusing for the user to use the 'default' running mode (for a different compilation setup, the default may mean a different actual running mode). Maybe I can add an internal macro/enum, `Mode_Default` to prevent users from using it and make it more readable; what do you think?

PR #1875, opened 2023-01-09
fix go language binding build on macos arm64

bonifaido:
Also a question in the meantime, are there any plans to support exposing host functions in the Go binding as in the Python implementation?

### DIFF
--- a/language-bindings/go/build.sh
+++ b/language-bindings/go/build.sh
@@ -7,6 +7,12 @@ PLATFORM=$(uname -s | tr A-Z a-z)
 CUR_DIR=$PWD
 WAMR_DIR=$PWD/../..
 WAMR_GO_DIR=$PWD/wamr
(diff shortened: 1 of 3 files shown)


wenyongh:
> Also a question in the meantime, are there any plans to support exposing host functions in the Go binding as in the Python implementation?

Do you mean to register native APIs written by Go? There is no plan yet. It is a little difficult to convert Go arguments into C arguments, and dispatch Go native API to C native API, and register the latter to runtime, do you have any ideas? Thanks.

bonifaido:
Yes, for example one of the wasm3 bindings has this feature implemented (I haven't tried it): https://github.com/iden3/go-wasm3/blob/master/wasm3.go#L414

wenyongh:
OK, thanks a lot, we may investigate more and try to implement the feature in the future.

PR #1879, opened 2023-01-10
Document some info about estimating memory usage

yamt:
While I've written this for my colleagues and thus it only covers limited configurations, hopefully it can still be useful for others.

### DIFF
--- a/doc/memory_usage.md
+++ b/doc/memory_usage.md
@@ -122,7 +122,7 @@ In some configurations, runtime overflow can be detected using hardware traps.
 
 In some configurations, explicit overflow detection logic can be emitted
 into AOT modules themselves. (cf. `os_thread_get_stack_boundary`,
-`check_stack_boundary`, `wamrc --stack-bounds-checks`)
+`check_stack_boundary`, `wamrc --stack-bounds-checks=1/0`)
 
 Memory profiling
 ================
--- a/doc/memory_usage.md
+++ b/doc/memory_usage.md
@@ -1,0 +1,132 @@
+Memory usage estimation for a module
+====================================
+
+This document aims to provide information useful to make a rough estimation
+of necessary memory to execute a WASM module.
+
+Instead of trying to cover every possible configurations,
+the following configuration is assumed in this document:
+
+* Module is built with `wasi-sdk`
+* Module is loaded with `wasm_runtime_load`
+* AOT is used
+* WASI is used
+* libc heap is used
+* app heap is not used
+* wasi-threads based pthread might be used
+
+Module
+------
+
+The memory to store the module binary is allocated by the embedder and
+passed to `wasm_runtime_load`.
+While WAMR owns the buffer, WAMR might make in-place modifications to
+its contents.
+
+Loaded module and its instances
+-------------------------------
+
+Many of data structures for module and instances are allocated from
+the global heap. (aka. `wasm_runtime_malloc`)
+
+AOT code section
+----------------
+
+Memory to load AOT machine code section.
+
+Because this memory needs to be executable, depending on platforms,
+it's allocated from a separate allocator.
+For example, `mmap` and `mprotect` are used on POSIX-like platforms.
+
+Linear memory
+-------------
+
+A WASM linear memory is either shared or non-shared.
+
+A WASM linear memory has `min` and `max` sizes.
+(They correspond to `wasm-ld`'s `--init-memory` and `--max-memory` options.)
+They are in the number of WASM pages, each of which is of 65536 bytes.
+The `max` is optional for non-shared memory. When omitted, it effectivily
+means unlimited.
+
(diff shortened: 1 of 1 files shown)


doc/memory_usage.md
```diff
@@ -0,0 +1,132 @@
...
+AUX stack (aka. C shadow stack)
+-------------------------------
+
+For the main thread, it's a part of the linear memory,
+between `__data_end` and `__heap_base` symbols.
+You can control the size of this stack with `wasm-ld`'s
+`-z stack-size` option.
+
+For threads created by `pthread_create`, libc allocates the stack for
+them dynamically from the libc heap.
```
wenyongh, line 102: May be not very correct here, the wasi-threads mode (dev/wasi-threads) really behaves by this way, but for the main branch, the current implementation or non wasi-threads mode, runtime divides the whole AUX stack into many regions (count is set by wasm_runtime_set_max_thread_num), each thread can apply a region as its private aux stack, but not allocate the stack from libc heap.

doc/memory_usage.md
```diff
@@ -0,0 +1,132 @@
...
+The size of this stack is inherited from the main thread's one
+unless overwritten with `pthread_attr_setstacksize` etc.
+
+WAMR tries to detect overflow/underflow when updating the stack pointer
+global. For threads created by `pthread_create`, the detection mechanism
+is disabled as of writing this.
```
wenyongh, line 108: The detection mechanism is also enabled for the child threads, since runtime can set the bottom and boundary of each child thread's private aux stack.

doc/memory_usage.md
```diff
@@ -0,0 +1,132 @@
...
+host threads to run those WASM threads. The stack size of these host
+threads are controlled by a build-time configuration.
+(`APP_THREAD_STACK_SIZE_DEFAULT`)
+
+In some configurations, runtime overflow can be detected using hardware traps.
+(`OS_ENABLE_HW_BOUND_CHECK`)
+
+In some configurations, explicit overflow detection logic can be emitted
+into AOT modules themselves. (cf. `os_thread_get_stack_boundary`,
+`check_stack_boundary`, `wamrc --stack-bounds-checks`)
```
wenyongh, line 125: had better `wamrc --stack-bounds-checks=1/0`?

doc/memory_usage.md
yamt: i'm aware of it. as stated at the top of the doc, i'm not trying to cover every configurations, at least for the first version.

doc/memory_usage.md
yamt: as far as i know, it isn't the case for the current version of wasi-threads.

doc/memory_usage.md
yamt: do you mean to add `=1/0`?

doc/memory_usage.md
wenyongh: OK, you mean `wasi-threads based pthread might be used` for this document? That sounds good.

doc/memory_usage.md
wenyongh: Got it.

doc/memory_usage.md
wenyongh: Yes, it looks more complete?

doc/memory_usage.md
yamt: yes.
maybe it's better to say `WASM_ENABLE_LIB_PTHREAD` is not used explicitly.

doc/memory_usage.md
```diff
@@ -0,0 +1,133 @@
...
+Instead of trying to cover every possible configurations,
+the following configuration is assumed in this document:
+
+* Module is built with `wasi-sdk`
+* Module is loaded with `wasm_runtime_load`
+* AOT is used
+* WASI is used
+* libc heap is used
+* app heap is not used
+* wasi-threads based pthread might be used
```
loganek, line 16: worth mentioning `WASM_ENABLE_WASI_THREADS` just like you do in the line below for `lib_pthread`

doc/memory_usage.md
```diff
@@ -0,0 +1,133 @@
+Memory usage estimation for a module
```
loganek, line 1: I think there are some overlaps with the https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/doc/memory_tune.md
Whereas I think both documents add good value, I wonder if we should rather merge them into a single high-quality doc about memory in general? Having said that, I don't mind keeping them separate for now, just a thought for the future.

loganek (approved):
LGTM, thanks for wrting the doc!

doc/memory_usage.md
yamt: done

doc/memory_usage.md
yamt: > I wonder if we should rather merge them into a single high-quality doc about memory in general?

ideally, maybe.

PR #1882, opened 2023-01-11
Add auto set up docker images for `WAMR-IDE`

NingW101:
1. add docker images auto check and setup support
2. fix bug that if user skip install then the code goes on

### DIFF
--- a/test-tools/wamr-ide/VSCode-Extension/src/extension.ts
+++ b/test-tools/wamr-ide/VSCode-Extension/src/extension.ts
@@ -29,6 +29,7 @@ import {
     checkIfDockerImagesExist,
     promptSetupDockerImages,
 } from './utilities/dockerUtilities';
+import { SelectionOfPrompt } from './constants';
 
 let wasmTaskProvider: WasmTaskProvider;
 let wasmDebugConfigProvider: WasmDebugConfigurationProvider;
@@ -329,7 +330,10 @@ export async function activate(context: vscode.ExtensionContext) {
                      *        we should return rather than continue
                      *        the execution
                      */
-                    if ((await promptSetupDockerImages(context)) === 'skip') {
+                    if (
+                        (await promptSetupDockerImages(context)) ===
+                        SelectionOfPrompt.skip
+                    ) {
                         return;
                     }
                 }
@@ -411,7 +415,10 @@ export async function activate(context: vscode.ExtensionContext) {
                      *        we should return rather than continue
                      *        the execution
                      */
-                    if ((await promptInstallLLDB(context)) === 'skip') {
+                    if (
+                        (await promptInstallLLDB(context)) ===
+                        SelectionOfPrompt.skip
+                    ) {
                         return;
                     }
                 }
@@ -423,7 +430,10 @@ export async function activate(context: vscode.ExtensionContext) {
                     /**NOTE - save as above lldb, should return if
                      *        users select to skip set up
                      */
-                    if ((await promptSetupDockerImages(context)) === 'skip') {
+                    if (
+                        (await promptSetupDockerImages(context)) ===
+                        SelectionOfPrompt.skip
+                    ) {
                         return;
                     }
                 }
--- a/test-tools/wamr-ide/VSCode-Extension/src/extension.ts
+++ b/test-tools/wamr-ide/VSCode-Extension/src/extension.ts
@@ -24,6 +24,12 @@ import {
     getWAMRExtensionVersion,
 } from './utilities/lldbUtilities';
 
+import {
+    checkIfDockerStarted,
+    checkIfDockerImagesExist,
+    promptSetupDockerImages,
+} from './utilities/dockerUtilities';
+
 let wasmTaskProvider: WasmTaskProvider;
 let wasmDebugConfigProvider: WasmDebugConfigurationProvider;
 let currentPrjDir = '';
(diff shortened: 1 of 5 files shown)


wenyongh:
@loganek @eloparco Could you help review this PR? Thanks.

test-tools/wamr-ide/VSCode-Extension/src/utilities/lldbUtilities.ts
```diff
@@ -78,7 +78,7 @@ export async function promptInstallLLDB(
     );
 
     if (response === skipPrompt) {
-        return;
+        return Promise.resolve(skipPrompt);
```
loganek, line 81: It's `async` function, so you can just return the value, i.e.:
```
return skipPrompt;
```

test-tools/wamr-ide/VSCode-Extension/src/utilities/lldbUtilities.ts
```diff
@@ -114,7 +114,6 @@ export async function promptInstallLLDB(
     );
 
     // Remove the bundle.zip
-    fs.unlink(lldbZipPath, () => {
-        return;
-    });
+    fs.unlinkSync(lldbZipPath);
+    return Promise.resolve('done');
```
loganek, line 118: same here, this can be simplified to: 
```
return 'done`;
```

test-tools/wamr-ide/VSCode-Extension/src/utilities/dockerUtilities.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    const downloadUrlArray = getDockerImagesDownloadUrl(context);
+
+    const destinationFolder = path.resolve(
+        extensionPath,
+        'resource',
+        DOCKER_IMAGES_TEM_FOLDER_NAME
+    );
+
+    if (!fs.existsSync(destinationFolder)) {
```
loganek, line 49: That technically only checks if file or directory exists, whereas we want to make sure this is a directory. I don't think it's a big problem though, so we can probably leave it as it is for now.

test-tools/wamr-ide/VSCode-Extension/src/extension.ts
```diff
@@ -382,10 +407,29 @@ export async function activate(context: vscode.ExtensionContext) {
             /* we should check again whether the user installed lldb, as this can be skipped during activation */
             try {
                 if (!isLLDBInstalled(context)) {
-                    await promptInstallLLDB(context);
+                    /**NOTE - if users select to skip install,
+                     *        we should return rather than continue
+                     *        the execution
+                     */
+                    if ((await promptInstallLLDB(context)) === 'skip') {
```
loganek, line 414: Maybe we can define an enum, string union type or at least use `boolean` for the return value here and in `promptSetupDockerImages` instead of using raw strings? That'd help avoiding potential bugs in the future.

test-tools/wamr-ide/VSCode-Extension/src/extension.ts
```diff
@@ -461,48 +505,70 @@ export async function activate(context: vscode.ExtensionContext) {
...
+                                e.execution.task.name ===
+                                'Wasm-Container-Before-Run'
+                            ) {
+                                /* make sure that run wasm task will be executed when destroy task finish */
+                                vscode.commands
+                                    .executeCommand(
+                                        'workbench.action.tasks.runTask',
+                                        'Run: Wasm'
+                                    )
+                                    .then(() => {
```
loganek, line 559: Would it be possible to convert this code into `async`? I think this nested code is quite difficult to read and maintain.

NingW101: Thanks so much for taking time to review this PR : )

And yes, the code in this part is not easily to read and may cost unnecessary effort to maintain, I will take it seriously to think about refactoring with `async` and introduce in another patch.

NingW101:
@loganek I updated the patch according to your comments, could you please help to take a look. Thanks~

cimacmillan (approved):

test-tools/wamr-ide/VSCode-Extension/src/utilities/dockerUtilities.ts
```diff
@@ -0,0 +1,123 @@
...
+            if (error) {
+                return reject(error);
+            }
+            return resolve(result);
+        });
+    });
+
+export async function promptSetupDockerImages(
+    context: vscode.ExtensionContext
+): Promise<string> {
```
loganek, line 28: Looks like this function returns `SelectionOfPrompt` or `Status`, but not string. You might consider having a single type being returned (e.g. `Status`) and extend it with `Skip` or `Cancel` field. Also, I wonder why typechecker didn't capture this - is enum treated as string in TypeScript or is the typechecker disabled?

wenyongh:
@loganek Ning has addressed the comment, shall I merge the PR?

test-tools/wamr-ide/VSCode-Extension/src/utilities/lldbUtilities.ts
```diff
@@ -67,18 +68,17 @@ export function isLLDBInstalled(context: vscode.ExtensionContext): boolean {
 
 export async function promptInstallLLDB(
     context: vscode.ExtensionContext
-): Promise<void> {
+): Promise<string> {
```
loganek, line 71: I'd prefer not returning `string` from a function that returns well-defined subset of values. Looking at the usage of this function, my recommendation would be to return `SelectionOfPrompt` (and return `setUp` at the end of this function instead of `Status` object).

loganek (approved):
The change looks good, I left one minor comment that you might want to address, but it's not critical, so approving.

test-tools/wamr-ide/VSCode-Extension/src/utilities/lldbUtilities.ts
NingW101: good suggestion, updated

PR #1883, opened 2023-01-11
Prevent undefined behaviour from c_api_func_imports == NULL

MartinKlang:
We found an Undefined Behaviour, caught by the clang sanitizers, stemming from this line:
https://github.com/bytecodealliance/wasm-micro-runtime/blame/9d52960e4d097d2503d9cda7b36c1716dcec01da/core/iwasm/aot/aot_runtime.c#L1767

The reported error was:

```
/root/src/wasm-micro-runtime/core/iwasm/aot/aot_runtime.c:1767:47: runtime error: applying zero offset to null pointer
110
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /root/src/wasm-micro-runtime/core/iwasm/aot/aot_runtime.c:1767:47
```

It would seem that `module_inst_extra->c_api_func_imports` is NULL under some circumstances. I've not investigated how or why that happens, but checking for NULL makes the Undefined Behaviour go away, in our use cases at least.

Unfortunately the test that brought this up is fairly complicated and not something I can easily share.

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -1787,7 +1787,9 @@ aot_invoke_native(WASMExecEnv *exec_env, uint32 func_idx, uint32 argc,
     AOTModuleInstanceExtra *module_inst_extra =
         (AOTModuleInstanceExtra *)module_inst->e;
     CApiFuncImport *c_api_func_import =
-        module_inst_extra->c_api_func_imports + func_idx;
+        module_inst_extra->c_api_func_imports
+            ? module_inst_extra->c_api_func_imports + func_idx
+            : NULL;
     uint32 *func_type_indexes = module_inst->func_type_indexes;
     uint32 func_type_idx = func_type_indexes[func_idx];
     AOTFuncType *func_type = aot_module->func_types[func_type_idx];
@@ -1803,7 +1805,8 @@ aot_invoke_native(WASMExecEnv *exec_env, uint32 func_idx, uint32 argc,
 
     import_func = aot_module->import_funcs + func_idx;
     if (import_func->call_conv_wasm_c_api)
-        func_ptr = c_api_func_import->func_ptr_linked;
+        func_ptr =
+            c_api_func_import ? c_api_func_import->func_ptr_linked : NULL;
 
     if (!func_ptr) {
         snprintf(buf, sizeof(buf),
--- a/core/iwasm/interpreter/wasm_interp_classic.c
+++ b/core/iwasm/interpreter/wasm_interp_classic.c
@@ -891,7 +891,7 @@ wasm_interp_call_func_native(WASMModuleInstance *module_inst,
     if (!func_import->call_conv_wasm_c_api) {
         native_func_pointer = module_inst->import_func_ptrs[cur_func_index];
     }
-    else {
+    else if (module_inst->e->c_api_func_imports) {
         c_api_func_import = module_inst->e->c_api_func_imports + cur_func_index;
         native_func_pointer = c_api_func_import->func_ptr_linked;
     }
--- a/core/iwasm/interpreter/wasm_interp_fast.c
+++ b/core/iwasm/interpreter/wasm_interp_fast.c
@@ -925,7 +925,7 @@ wasm_interp_call_func_native(WASMModuleInstance *module_inst,
     if (!func_import->call_conv_wasm_c_api) {
         native_func_pointer = module_inst->import_func_ptrs[cur_func_index];
     }
-    else {
+    else if (module_inst->e->c_api_func_imports) {
         c_api_func_import = module_inst->e->c_api_func_imports + cur_func_index;
         native_func_pointer = c_api_func_import->func_ptr_linked;
     }
--- a/core/iwasm/interpreter/wasm_runtime.c
+++ b/core/iwasm/interpreter/wasm_runtime.c
@@ -2908,8 +2908,14 @@ llvm_jit_invoke_native(WASMExecEnv *exec_env, uint32 func_idx, uint32 argc,
 
     import_func = &module->import_functions[func_idx].u.function;
     if (import_func->call_conv_wasm_c_api) {
-        c_api_func_import = module_inst->e->c_api_func_imports + func_idx;
-        func_ptr = c_api_func_import->func_ptr_linked;
+        if (module_inst->e->c_api_func_imports) {
+            c_api_func_import = module_inst->e->c_api_func_imports + func_idx;
+            func_ptr = c_api_func_import->func_ptr_linked;
+        }
+        else {
+            c_api_func_import = NULL;
+            func_ptr = NULL;
+        }
     }
 
     if (!func_ptr) {


core/iwasm/aot/aot_runtime.c
```diff
@@ -1802,7 +1802,7 @@ aot_invoke_native(WASMExecEnv *exec_env, uint32 func_idx, uint32 argc,
     bh_assert(func_idx < aot_module->import_func_count);
 
     import_func = aot_module->import_funcs + func_idx;
-    if (import_func->call_conv_wasm_c_api)
+    if (import_func->call_conv_wasm_c_api && c_api_func_import)
```
yamt, line 1805: are these NULL checks necessary?
as far as i know, when you have a function with call_conv_wasm_c_api, c_api_func_imports can't be NULL.

MartinKlang: > are these NULL checks necessary?

Good question, and frankly I don't know.

> when you have a function with call_conv_wasm_c_api, c_api_func_imports can't be NULL

How do we know this?

wenyongh: Seems had better change to:
```C
if (import_func->call_conv_wasm_c_api)
    func_ptr = c_api_func_import ? c_api_func_import->func_ptr_linked : NULL;
```
Since for c api import, we shouldn't get func_ptr from `func_ptrs[func_idx]`

core/iwasm/aot/aot_runtime.c
```diff
@@ -1814,7 +1816,7 @@ aot_invoke_native(WASMExecEnv *exec_env, uint32 func_idx, uint32 argc,
     }
 
     attachment = import_func->attachment;
-    if (import_func->call_conv_wasm_c_api) {
+    if (import_func->call_conv_wasm_c_api && c_api_func_import) {
```
wenyongh, line 1819: Seems no need to change here if the code above is changed to what I mentioned, since if `import_func->call_conv_wasm_c_api && c_api_func_import == NULL`, func_ptr will be NULL, and exception will be thrown in L1814 and then L1815 will exit, L1819 won't be executed.

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -904,7 +904,7 @@ wasm_interp_call_func_native(WASMModuleInstance *module_inst,
         return;
     }
 
-    if (func_import->call_conv_wasm_c_api) {
+    if (func_import->call_conv_wasm_c_api && c_api_func_import) {
```
wenyongh, line 907: Same as above, seems no need to change here.

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -939,7 +939,7 @@ wasm_interp_call_func_native(WASMModuleInstance *module_inst,
         return;
     }
 
-    if (func_import->call_conv_wasm_c_api) {
+    if (func_import->call_conv_wasm_c_api && c_api_func_import) {
```
wenyongh, line 942: Same as above

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -2907,7 +2907,8 @@ llvm_jit_invoke_native(WASMExecEnv *exec_env, uint32 func_idx, uint32 argc,
     bh_assert(func_idx < module->import_function_count);
 
     import_func = &module->import_functions[func_idx].u.function;
-    if (import_func->call_conv_wasm_c_api) {
+    if (import_func->call_conv_wasm_c_api
+        && module_inst->e->c_api_func_imports) {
         c_api_func_import = module_inst->e->c_api_func_imports + func_idx;
         func_ptr = c_api_func_import->func_ptr_linked;
     }
```
wenyongh, line 2914: Had better change to:
```C
    if (import_func->call_conv_wasm_c_api) {
        if (module_inst->e->c_api_func_imports) {
            c_api_func_import = module_inst->e->c_api_func_imports + func_idx;
            func_ptr = c_api_func_import->func_ptr_linked;
        }
        else {
            c_api_func_import = NULL;
            func_ptr = NULL;
        }
    }
```

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -2921,7 +2922,7 @@ llvm_jit_invoke_native(WASMExecEnv *exec_env, uint32 func_idx, uint32 argc,
...
 
     attachment = import_func->attachment;
-    if (import_func->call_conv_wasm_c_api) {
+    if (import_func->call_conv_wasm_c_api && c_api_func_import) {
```
wenyongh, line 2925: Same above, no need to change here

core/iwasm/aot/aot_runtime.c
MartinKlang: great, changes done -
ty @wenyongh, it looks much better now!

core/iwasm/interpreter/wasm_interp_classic.c
MartinKlang: done

core/iwasm/interpreter/wasm_interp_fast.c
MartinKlang: done

core/iwasm/interpreter/wasm_runtime.c
MartinKlang: done

wenyongh (commented):
LGTM

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -891,7 +891,7 @@ wasm_interp_call_func_native(WASMModuleInstance *module_inst,
     if (!func_import->call_conv_wasm_c_api) {
         native_func_pointer = module_inst->import_func_ptrs[cur_func_index];
     }
-    else {
+    else if (module_inst->e->c_api_func_imports) {
         c_api_func_import = module_inst->e->c_api_func_imports + cur_func_index;
         native_func_pointer = c_api_func_import->func_ptr_linked;
     }
```
lum1n0us, line 897: Guess a more `else` branch about the condition of `func_import->call_conv_wasm_c_api == true && module_inst->e->c_api_func_imports == NULL` would  be better

wenyongh: If another else branch is added, I think what we can do is to set native_func_pointer to NULL inside it:
```C
    if (!func_import->call_conv_wasm_c_api) {
        native_func_pointer = module_inst->import_func_ptrs[cur_func_index];
    }
    else if (module_inst->e->c_api_func_imports) {
        c_api_func_import = module_inst->e->c_api_func_imports + cur_func_index;
        native_func_pointer = c_api_func_import->func_ptr_linked;
    }
    else {
       native_func_pointer = NULL;
    }
```
But the else branch can do nothing since native_func_pointer is set to NULL at the function beginning.

lum1n0us: IIUC, `wasm_instance_new` shall set `call_conv_wasm_c_api` to `true` and bind `c_api_func_imports` if execute successful. So, a warning or error should be raised.

wenyongh: Do you mean that the `module_inst->e->c_api_func_imports` must not be NULL when wasm_instance_new is called and the import func count is not 0? I guess here the issue reported by sanitizers in aot_runtime.c is that the developer uses the APIs in wasm_export.h to load and instantiate the wasm module, and wasm-c-api wasm_instance_new isn't called, so `module_inst->e->c_api_func_imports` in aot_runtime.c is NULL, and `c_api_func_imports + func_idx` triggers sanitizer error.

Seems that we only need to add check in aot_runtime.c:
```C
    CApiFuncImport *c_api_func_import =
        module_inst_extra->c_api_func_imports
            ? module_inst_extra->c_api_func_imports + func_idx
            : NULL;
```
And in other places when call_conv_wasm_c_api is detected as true, no need to change the code, or just add an assertion, e.g.:
```C
    if (import_func->call_conv_wasm_c_api) {
        bh_assert(c_api_func_import);
        func_ptr = c_api_func_import->func_ptr_linked;
    }
```

lum1n0us: Got it. In that case (`wasm_instance_new` isn't called), it does make sense.

LGTM now.

yamt (approved):
lgtm

PR #1888, opened 2023-01-15
Add some macros for int literal.

FromLiQg:
I found the file `stdint.h` (in `libc-builtin-sysroot`) is missing some macro for int leteral, so I add these macros.
I guess this file originally refers to `/usr/include/stdint.h` in unbuntu, and the `__INT64_C` is missing.
So I refer to the implementation of `stdint.h` in 'wasi-sdk'.

### DIFF
--- a/wamr-sdk/app/libc-builtin-sysroot/include/stdint.h
+++ b/wamr-sdk/app/libc-builtin-sysroot/include/stdint.h
@@ -11,6 +11,13 @@ extern "C" {
 #endif
 
 /* clang-format off */
+/* The word size of platform */
+#ifdef __wasm64__
+#define __WORDSIZE 64
+#else
+#define __WORDSIZE 32
+#endif
+
 typedef char            int8_t;
 typedef short int       int16_t;
 typedef int             int32_t;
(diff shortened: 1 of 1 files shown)


wamr-sdk/app/libc-builtin-sysroot/include/stdint.h
```diff
@@ -25,22 +25,44 @@ typedef unsigned long long int  uint64_t;
 typedef __INTPTR_TYPE__		intptr_t;
 typedef __UINTPTR_TYPE__	uintptr_t;
 
+/* Values to test for integral types holding `void *' pointer.  */
+#define INTPTR_MIN      INT32_MIN
+#define INTPTR_MAX      INT32_MAX
+#define UINTPTR_MAX     UINT32_MAX
```
wenyongh, line 31: Had better add macro control, like L55
```C
#if UINTPTR_MAX == UINT64_MAX
#define INTPTR_MIN      INT64_MIN
#define INTPTR_MAX      INT64_MAX
#define UINTPTR_MAX     UINT64_MAX
#else
#define INTPTR_MIN      INT32_MIN
#define INTPTR_MAX      INT32_MAX
#define UINTPTR_MAX     UINT32_MAX
#endif
```

FromLiQg: I got it. These macros refer to the implementation of wasi, but I think your suggestion is correct.

wamr-sdk/app/libc-builtin-sysroot/include/stdint.h
```diff
@@ -11,6 +11,9 @@ extern "C" {
 #endif
 
 /* clang-format off */
+/* The word size of platform */
+#define __WORDSIZE 32
```
wenyongh, line 15: Though currently wasm in spec mvp is 32-bit, but wasm has introduced the [memory64](https://github.com/WebAssembly/memory64) spec proposal, had better not assume wordsize is 32-bit here. Is there a better way to control the code?

FromLiQg: Thank you very much. I got it!

FromLiQg: I use macros like `__wasm32__` to identify `wasm32 or wasm64`. This macro is predefined by `clang`

wenyongh: OK, can you try both wasi-sdk and emsdk to check whether they both define the macro?

FromLiQg: Ok, I got it.

FromLiQg: I haven't used `emsdk` before. I used the following commands and programs to check and got the following output. It seems that emsdk also uses clang. And the macro ` __wasm32__` is valid.
* command
```shell
emcc ./main.c -o main
```
* source code
```c
#include<stdio.h>
#ifdef __wasm32__
#error WASM32 detected!
void wasm32(){
    printf("wasm32");
}
#endif
#ifdef __wasm64__
#error wasm64
void wasm64(){
    printf("wasm64");
}
#endif
int main(){
#ifdef __wasm32__
wasm32();
#endif
#ifdef __wasm64__
wasm64();
#endif
    return 0;
}
```
* result
```
./main.c:3:2: error: WASM32 detected!
#error WASM32 detected!
 ^
1 error generated.
emcc: error: '/home/ljq/test/emsdk/upstream/bin/clang -target wasm32-unknown-emscripten -fignore-exceptions -fvisibility=default -mllvm -combiner-global-alias-analysis=false -mllvm -enable-emscripten-sjlj -mllvm -disable-lsr -DEMSCRIPTEN -Werror=implicit-function-declaration --sysroot=/home/ljq/test/emsdk/upstream/emscripten/cache/sysroot -Xclang -iwithsysroot/include/fakesdl -Xclang -iwithsysroot/include/compat ./main.c -c -o /tmp/emscripten_temp_amzvebaf/main_0.o' failed (returned 1)
```

wenyongh: Got it, thanks.

PR #1889, opened 2023-01-16
Main thread spread

xujuntwt95329:


core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1794,6 +1798,9 @@ wasm_runtime_call_wasm(WASMExecEnv *exec_env,
                                              result_argc, argv)) {
         wasm_runtime_set_exception(exec_env->module_inst,
                                    "the result conversion is failed");
+#if WASM_ENABLE_THREAD_MGR != 0
+        wasm_cluster_spread_exception(exec_env);
```
wenyongh, line 1802: Should also spread exception when wasm_runtime_prepare_call_function fails, L1767?

core/iwasm/aot/aot_runtime.c
```diff
@@ -902,7 +902,7 @@ create_exports(AOTModuleInstance *module_inst, AOTModule *module,
 static bool
 clear_wasi_proc_exit_exception(AOTModuleInstance *module_inst)
 {
-#if WASM_ENABLE_LIBC_WASI != 0
+#if (WASM_ENABLE_LIBC_WASI != 0) && (WASM_ENABLE_THREAD_MGR == 0)
```
wenyongh, line 905: Not sure why skip clear the exception for multi-threads?

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -4021,7 +4021,7 @@ fast_jit_call_func_bytecode(WASMModuleInstance *module_inst,
 static bool
 clear_wasi_proc_exit_exception(WASMModuleInstance *module_inst)
 {
-#if WASM_ENABLE_LIBC_WASI != 0
+#if (WASM_ENABLE_LIBC_WASI != 0) && (WASM_ENABLE_THREAD_MGR == 0)
```
wenyongh, line 4024: Same as above, why skip clear for multi-threads?

core/iwasm/common/wasm_runtime_common.c
xujuntwt95329: according to our discussion, I‘ve moved the spread into wasm_set_exception function, so we don't need to process the spread elsewhere

core/iwasm/aot/aot_runtime.c
xujuntwt95329: Restored, thanks~

yamt:
* i guess you need some kind of a retry loop as other threads can create threads in the meantime.
* i guess traverse_list should require a lock. (cluster->lock?) but it isn't a fault of this PR.

xujuntwt95329:
Thanks for the suggestions.

> * i guess you need some kind of a retry loop as other threads can create threads in the meantime.

I think we can add a status field to record the status of a cluster. When we spreading the exception:
1. get `cluster->lock`
2. set cluster's status to `exception`
3. spread the exception

Then if there are any other threads creating new thread at the meantime, we directly fail it if the cluster's status is `exception`

@yamt @wenyongh how do you think about this solution?

> * i guess traverse_list should require a lock. (cluster->lock?) but it isn't a fault of this PR.

Yes, and `traverse_list` is used for different lists, I‘ll analyze every situation and add the locks if necessary

samples/multi-thread/wasm-apps/main_thread_exception.c
```diff
@@ -0,0 +1,36 @@
...
+typedef struct ThreadArgs {
+    int start;
+    int length;
+} ThreadArgs;
+
+void *
+thread(void *args)
+{
+    while (1) {
+        /* When other thread (including main thread) throw exception,
```
loganek, line 18:
```suggestion
        /* When other threads (including main thread) throw exception,
```

samples/multi-thread/wasm-apps/main_thread_exception.c
```diff
@@ -0,0 +1,36 @@
...
+main()
+{
+    pthread_t tids;
+
+    if (pthread_create(&tids, NULL, thread, NULL) != 0) {
+        printf("pthread_create failed\n");
+    }
+
+    /* Trigger an exception */
+    *(int *)(-1) = 100;
```
loganek, line 33: If the code is only supposed to compile for WASM, you might consider `__builtin_trap` instead.

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -919,14 +926,24 @@ set_exception_visitor(void *node, void *user_data)
...
+    WASMModuleInstance *wasm_inst = (WASMModuleInstance *)module_inst;
     const char *exception = wasm_runtime_get_exception(module_inst);
     /* skip "Exception: " */
     exception += 11;
 
     if (curr_exec_env != exec_env) {
-        curr_module_inst = get_module_inst(curr_exec_env);
-        wasm_runtime_set_exception(curr_module_inst, exception);
+        WASMModuleInstance *curr_wasm_inst;
+        WASMModuleInstanceCommon *curr_module_inst =
```
loganek, line 936:
```suggestion
        WASMModuleInstance *curr_wasm_inst = (WASMModuleInstance*)get_module_inst(curr_exec_env);
```

yamt:
> Thanks for the suggestions.
> 
> > * i guess you need some kind of a retry loop as other threads can create threads in the meantime.
> 
> I think we can add a status field to record the status of a cluster. When we spreading the exception:
> 
>     1. get `cluster->lock`
> 
>     2. set cluster's status to `exception`
> 
>     3. spread the exception
> 
> 
> Then if there are any other threads creating new thread at the meantime, we directly fail it if the cluster's status is `exception`
> 
> @yamt @wenyongh how do you think about this solution?

i think it works.

> 
> > * i guess traverse_list should require a lock. (cluster->lock?) but it isn't a fault of this PR.
> 
> Yes, and `traverse_list` is used for different lists, I‘ll analyze every situation and add the locks if necessary

thank you.

core/iwasm/libraries/thread-mgr/thread_manager.c
xujuntwt95329: Fixed, thanks a lot!

wenyongh:
> Thanks for the suggestions.
> 
> > * i guess you need some kind of a retry loop as other threads can create threads in the meantime.
> 
> I think we can add a status field to record the status of a cluster. When we spreading the exception:
> 
> 1. get `cluster->lock`
> 2. set cluster's status to `exception`
> 3. spread the exception
> 
> Then if there are any other threads creating new thread at the meantime, we directly fail it if the cluster's status is `exception`
> 
> @yamt @wenyongh how do you think about this solution?

It looks good to me.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1743,6 +1743,29 @@ wasm_runtime_finalize_call_function(WASMExecEnv *exec_env,
...
+bool
+clear_wasi_proc_exit_exception(WASMModuleInstanceCommon *module_inst_comm)
+{
+#if WASM_ENABLE_LIBC_WASI != 0
+    WASMModuleInstance *module_inst = (WASMModuleInstance *)module_inst_comm;
+
+    bh_assert(module_inst_comm->module_type == Wasm_Module_Bytecode
+              || module_inst_comm->module_type == Wasm_Module_AoT);
+
+    const char *exception = wasm_get_exception(module_inst);
```
wenyongh, line 1755: Had better declare the variable at the beginning of the function, a concern is that old version compiler might report warning.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -2150,11 +2178,19 @@ wasm_runtime_get_exec_env_singleton(WASMModuleInstanceCommon *module_inst_comm)
...
-    else
+    }
+    else {
         module_inst->cur_exception[0] = '\0';
+    }
+
+#if WASM_ENABLE_THREAD_MGR != 0
+    wasm_cluster_spread_exception(
+        wasm_clusters_search_exec_env((WASMModuleInstanceCommon *)module_inst),
+        exception ? false : true);
```
wenyongh, line 2192: When exception is NULL, does it mean to clear exception of other threads?

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -919,24 +958,33 @@ set_exception_visitor(void *node, void *user_data)
...
 
 void
-wasm_cluster_spread_exception(WASMExecEnv *exec_env)
+wasm_cluster_spread_exception(WASMExecEnv *exec_env, bool clear)
 {
     WASMCluster *cluster = wasm_exec_env_get_cluster(exec_env);
     bh_assert(cluster);
 
+    os_mutex_lock(&cluster->lock);
+    cluster->has_exception = !clear;
```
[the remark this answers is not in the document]
xujuntwt95329: Seems we need to change, because when clearing the exception, we don't want to terminate the thread

core/iwasm/common/wasm_runtime_common.c
xujuntwt95329: Done, thanks

core/iwasm/aot/aot_runtime.c
```diff
@@ -956,7 +941,8 @@ execute_start_function(AOTModuleInstance *module_inst)
     u.f(exec_env);
 
     wasm_exec_env_destroy(exec_env);
-    (void)clear_wasi_proc_exit_exception(module_inst);
+    (void)clear_wasi_proc_exit_exception(
+        (WASMModuleInstanceCommon *)module_inst);
```
wenyongh, line 945: Seems no need to clear wasi proc exit exception here. WASI module doesn't set internal start func index，instead it exports a function named "_start". If here it really call wasi proc exit，we can also let instantiation process failed. @lum1n0us what is your opinion?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1743,6 +1743,30 @@ wasm_runtime_finalize_call_function(WASMExecEnv *exec_env,
 }
 #endif
 
+bool
```
wenyongh, line 1746: can set to static if the above is ok

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -808,15 +840,31 @@ terminate_thread_visitor(void *node, void *user_data)
 void
 wasm_cluster_terminate_all(WASMCluster *cluster)
 {
+    os_mutex_lock(&cluster->lock);
+    cluster->processing = true;
+    os_mutex_unlock(&cluster->lock);
+
     traverse_list(&cluster->exec_env_list, terminate_thread_visitor, NULL);
+
```
wenyongh, line 848: does traverse_list add lock for the list? why not remove L845 and L849

xujuntwt95329: In `terminate_thread_visitor`, we will call os_thread_join to wait for other threads to exit, and the exited thread need to get `cluster->lock` for accessing the list, so we can't hold the lock during `terminate_thread_visitor`

core/iwasm/aot/aot_runtime.c
xujuntwt95329: Agree with that, and updated

core/iwasm/libraries/thread-mgr/thread_manager.c
wenyongh: Got it, thanks.

wenyongh (approved):
LGTM

PR #1890, opened 2023-01-16
Enable running mode control for various compiler/cli options

wenyongh:


core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -1567,9 +1567,6 @@ wasm_instantiate(WASMModule *module, bool is_sub_inst, uint32 stack_size,
     module_inst->e =
         (WASMModuleInstanceExtra *)((uint8 *)module_inst + extra_info_offset);
 
-    wasm_runtime_set_running_mode((wasm_module_inst_t)module_inst,
-                                  wasm_runtime_get_default_running_mode());
-
```
TianlongLiang, line 1424: I add a few lines to set running mode after functions sections instantiate. You can see it in [this commit](https://github.com/TianlongLiang/wasm-micro-runtime/commit/2552dbe05514ad14fbf3fdcff2cfd192e4fa3463) on my branch

in line 1662~1667
```C
   if (!wasm_runtime_set_running_mode(
            (wasm_module_inst_t)module_inst,
            wasm_runtime_get_default_running_mode())) {
        goto fail;
    }
```

wenyongh: It is unnecessary and may cause issue, the running mode will be set in line 1967 before calling functions, no need to set it repeatedly:
```C
    if (!set_running_mode(module_inst, wasm_runtime_get_default_running_mode(),
                          true)) {
                      "set instance running mode failed");
        goto fail;
    }
```

PR #1894, opened 2023-01-16
bump wasi-sdk version to 19

lum1n0us:


.github/workflows/compilation_on_android_ubuntu.yml
```diff
@@ -260,7 +260,7 @@ jobs:
         os: [ubuntu-20.04, ubuntu-22.04]
```
loganek, line 260: You might consider using [variables](https://docs.github.com/en/actions/learn-github-actions/variables) for WASI-SDK version so the next update will require less changes.

test-tools/wamr-ide/WASM-Toolchain/Docker/Dockerfile
```diff
@@ -25,10 +25,10 @@ RUN wget --progress=dot:giga https://github.com/Kitware/CMake/releases/download/
 #    - wamr-sdk
 
 ##  - download wasi-sdk with wget and set up to /opt/wasi-sdk
-RUN wget --progress=dot:giga https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-16/wasi-sdk-16.0-linux.tar.gz \
-    && tar -zxvf wasi-sdk-16.0-linux.tar.gz \
-    && mv wasi-sdk-16.0 /opt/wasi-sdk/ \
-    && rm -f wasi-sdk-16.0-linux.tar.gz
+RUN wget --progress=dot:giga https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-19/wasi-sdk-19.0-linux.tar.gz \
```
loganek, line 28: Probably having `ARG WASI_SDK=19` would simplify future updates.

.github/workflows/compilation_on_android_ubuntu.yml
lum1n0us: @wenyongh  Are we able to set some repository variables in ["Settings"](https://docs.github.com/en/actions/learn-github-actions/variables#creating-configuration-variables-for-a-repository)

wenyongh: I have no access to the `bytecodealliance/wasm-micro-runtime` yet

loganek: for now we can probably define it per-file. I think it might actually be better to have it defined in the code, so it's clear what version do we use.

lum1n0us: I am thinking about a [cmake superbuild](https://www.kitware.com/cmake-superbuilds-git-submodules/) as a "poor package manager" to search and install dependencies. It will let above similar dependencies upgrading more easily.

The rough idea is using CMake to replace "apt" and "wget" to manage those key dependencies like binaryren, emscripten-core, libuv, llvm, wabt, wasi-sdk and so on.  Beside crossing platform, it is also able to simplify the bumpping.

But it seems we still need to maintain a installation script for those developers who use Makefiles only.

Please let me know your opinions.

wenyongh: Seems to be a good idea, but there may be lots of changes, which requires effort to test and may be a risk now. Suggest to keep the current `poor package manager` until we release 1.2.0, and try to upgrade to `cmake superbuild` after that.

loganek: I think superbuild sounds like a good idea but agree with @wenyongh that's a bit of work, so let's stick to something simpler for now.

lum1n0us: track the new idea with https://github.com/bytecodealliance/wasm-micro-runtime/issues/1897

wenyongh (approved):
LGTM

PR #1900, opened 2023-01-19
feat: added tests for socket apis

hritikgupta:


core/iwasm/libraries/lib-socket/test/nslookup.c
```diff
@@ -0,0 +1,16 @@
+#include <assert.h>
+#include <stdio.h>
+#include <string.h>
```
loganek, line 3: do we have to include `stdio.h` and `string.h`?

core/iwasm/libraries/lib-socket/test/nslookup.c
```diff
@@ -0,0 +1,16 @@
...
+#else
+#include <netdb.h>
+#endif
+
+int
+main()
+{
+    struct addrinfo *result;
+    int ret = getaddrinfo("https://google.com", NULL, NULL, &result);
+    assert(ret >= 0);
```
loganek, line 15: can `getaddrinfo` ever return value greater than zero? And if so, is that successful error code? According to the [doc](https://man7.org/linux/man-pages/man3/getaddrinfo.3.html#RETURN_VALUE):

> getaddrinfo() returns 0 if it succeeds, or one of the following nonzero error codes:

So we probably want to change the assertion to:
```suggestion
    assert(ret == 0);
```

core/iwasm/libraries/lib-socket/test/nslookup.c
```diff
@@ -0,0 +1,16 @@
...
+#include <stdio.h>
+#include <string.h>
+#ifdef __wasi__
+#include <wasi_socket_ext.h>
+#else
+#include <netdb.h>
+#endif
+
+int
+main()
```
loganek, line 11: I'm not sure what exactly are we testing here? I'd suggest we test the result of `getaddrinfo`. Unfortunately, we can't control the responses (unless we mock DNS), but I think we can still test if there's any response or whether the filtering (`hints` parameter) works. What do you think?

core/iwasm/libraries/lib-socket/test/tcp.c
```diff
@@ -0,0 +1,90 @@
+#include <unistd.h>
+#include <string.h>
+#include "assert.h"
```
loganek, line 3:
```suggestion
#include <assert.h>
```

core/iwasm/libraries/lib-socket/test/tcp.c
```diff
@@ -0,0 +1,90 @@
...
+init_socket_addr()
+{
+    int sock = socket(AF_INET, SOCK_STREAM, 0);
+    assert(sock > 0);
+
+    struct sockaddr_in addr;
+    memset(&addr, '\0', sizeof(addr));
+    addr.sin_family = AF_INET;
+    addr.sin_port = PORT;
+    addr.sin_addr.s_addr = inet_addr(IP);
```
loganek, line 29: Consider using `htonl(INADDR_LOOPBACK);` instead, so we don't have to hardcode IP address.

core/iwasm/libraries/lib-socket/test/tcp.c
```diff
@@ -0,0 +1,90 @@
...
+};
+
+struct init_socket
+init_socket_addr()
+{
+    int sock = socket(AF_INET, SOCK_STREAM, 0);
+    assert(sock > 0);
+
+    struct sockaddr_in addr;
+    memset(&addr, '\0', sizeof(addr));
```
loganek, line 26:
```suggestion
    memset(&addr, 0, sizeof(addr));
```

core/iwasm/libraries/lib-socket/test/tcp.c
```diff
@@ -0,0 +1,90 @@
...
+    assert(strcmp(buffer, "Message from server.") == 0);
+    assert(close(sock) >= 0);
+    return NULL;
+}
+
+int
+main()
+{
+    pthread_t thread1,
+        thread2; // one thread runs the server, other runs the client
```
loganek, line 82: maybe we can use more descriptive names instead of the comment? e.g. `server_thread` and `client_thread`?

core/iwasm/libraries/lib-socket/test/tcp.c
```diff
@@ -0,0 +1,90 @@
...
+
+void *
+client()
+{
+    struct init_socket init_client_sock = init_socket_addr();
+    int sock = init_client_sock.sock;
+    struct sockaddr_in addr = init_client_sock.addr;
+    socklen_t addr_size;
+    char buffer[1024];
+    assert(connect(sock, (struct sockaddr *)&addr, sizeof(addr)) >= 0);
```
loganek, line 69:
```suggestion
    assert(connect(sock, (struct sockaddr *)&addr, sizeof(addr)) == 0);
```

core/iwasm/libraries/lib-socket/test/tcp.c
```diff
@@ -0,0 +1,90 @@
...
+}
+
+void *
+server()
+{
+    struct init_socket init_server_sock = init_socket_addr();
+    struct sockaddr_in server_addr = init_server_sock.addr, client_addr;
+    int server_sock = init_server_sock.sock, client_sock;
+    socklen_t addr_size;
+    char buffer[1024];
```
loganek, line 40: can this buffer be smaller?

core/iwasm/libraries/lib-socket/test/tcp.c
```diff
@@ -0,0 +1,90 @@
...
+}
+
+int
+main()
+{
+    pthread_t thread1,
+        thread2; // one thread runs the server, other runs the client
+
+    pthread_create(&thread1, NULL, server, NULL);
+    pthread_create(&thread2, NULL, client, NULL);
```
loganek, line 85: There's race condition here - `thread2` can go faster than `thread` and therefore the `connect` will be executed before server listens for connections. We'll have to make sure the server is ready before client tries to connect (or have retries in client in case the server is not up yet).

core/iwasm/libraries/lib-socket/test/tcp.c
```diff
@@ -0,0 +1,90 @@
...
+        bind(server_sock, (struct sockaddr *)&server_addr, sizeof(server_addr));
+    assert(n >= 0);
+
+    listen(server_sock, 1);
+    addr_size = sizeof(client_addr);
+    client_sock =
+        accept(server_sock, (struct sockaddr *)&client_addr, &addr_size);
+    assert(client_sock >= 0);
+
+    bzero(buffer, 1024);
```
loganek, line 52: we probably don't need to zero it if we override it in the line below.

core/iwasm/libraries/lib-socket/test/tcp.c
```diff
@@ -0,0 +1,90 @@
...
+    strcpy(buffer, "Message from server.");
+    sendto(client_sock, buffer, strlen(buffer), 0,
```
loganek, line 54: You can also use `send` and not specify `sockaddr` (unless you wanted to test `sendto` explicitly.

core/iwasm/libraries/lib-socket/test/tcp.c
```diff
@@ -0,0 +1,90 @@
...
+           (struct sockaddr *)&client_addr, addr_size);
+
+    assert(close(client_sock) >= 0);
```
loganek, line 57:
```suggestion
    assert(close(client_sock) == 0);
```

core/iwasm/libraries/lib-socket/test/tcp.c
```diff
@@ -0,0 +1,90 @@
...
+    int sock = init_client_sock.sock;
+    struct sockaddr_in addr = init_client_sock.addr;
+    socklen_t addr_size;
+    char buffer[1024];
+    assert(connect(sock, (struct sockaddr *)&addr, sizeof(addr)) >= 0);
+
+    bzero(buffer, 1024);
+    recv(sock, buffer, sizeof(buffer), 0);
+    assert(strcmp(buffer, "Message from server.") == 0);
+    assert(close(sock) >= 0);
```
loganek, line 74:
```suggestion
    assert(close(sock) == 0);
```

core/iwasm/libraries/lib-socket/test/tcp.c
hritikgupta: Right, I was thinking about approaches to fix it, should we use a lock or set a flag once the server is ready (prevent execution of client thread until the flag is set)?

loganek: We could probably use condition variable and notify other thread when the server is ready.

core/iwasm/libraries/lib-socket/test/nslookup.c
```diff
@@ -0,0 +1,16 @@
+#include <assert.h>
```
wenyongh, line 1: Should add license header.

core/iwasm/libraries/lib-socket/test/tcp.c
```diff
@@ -0,0 +1,90 @@
+#include <unistd.h>
```
wenyongh, line 1: Should add license header.

lum1n0us:
We'd better not check in a binary file (core/iwasm/libraries/lib-socket/test/tcp)

lum1n0us:
@loganek  @wenyongh  we usually keep similar code in *samples/socket-api*, is it better if we also store above code there?

wenyongh:
> We'd better not check in a binary file (core/iwasm/libraries/lib-socket/test/tcp)

Yes, don't check in wasm binary file here.

wenyongh:
> @loganek @wenyongh we usually keep similar code in _samples/socket-api_, is it better if we also store above code there?

No, had better move the source files into samples/socket-api/src, rename the sample files with suitable names, and update document about how to run the samples and what the output should be.

@hritikgupta Could you add the files as samples in `samples/socket-api/src` and update the document and build scripts? Thanks.

loganek:
@wenyongh , @lum1n0us so the intention of this is to run those tests as part of WASI test suite. I don't have strong opinion about where to put those tests, but my preference would be to keep examples and tests separate (even though there's some overlap).

loganek:
@hritikgupta could you also add integration with wasi testsuite to this PR? I think the change might be a bit confusing without that.

wenyongh:
> @wenyongh , @lum1n0us so the intention of this is to run those tests as part of WASI test suite. I don't have strong opinion about where to put those tests, but my preference would be to keep examples and tests separate (even though there's some overlap).

Got it, if they are only tests, it should be good to put them here, and yes, had better add them to CI test.

hritikgupta:
> > We'd better not check in a binary file (core/iwasm/libraries/lib-socket/test/tcp)
> 
> Yes, don't check in wasm binary file here.

Right right, this got committed accidentally while I was testing the script, will remove this as a part of the commit addressing the remaining review comments.

core/iwasm/libraries/lib-socket/test/nslookup.c
hritikgupta: Yes, makes sense - updating!

.github/workflows/compilation_on_android_ubuntu.yml
```diff
@@ -448,7 +448,7 @@ jobs:
         run: echo "::error::can not get prebuilt llvm libraries" && exit 1
 
       - name: run tests
-        run: ./test_wamr.sh ${{ matrix.test_option }} -t ${{ matrix.running_mode }}
+        run: ./core/iwasm/libraries/lib-socket/test/build.sh && ./test_wamr.sh ${{ matrix.test_option }} -t ${{ matrix.running_mode }}
```
wenyongh, line 451: Should be `../../core/iwasm/libraries/lib-socket/test/build.sh`, since the working-directory is `./tests/wamr-test-suites`

loganek: also, you might want to move build to a separate step so it's easier to spot in CI what actually failed (build or test)

core/iwasm/libraries/lib-socket/test/build.sh
```diff
@@ -0,0 +1,13 @@
+
+# Copyright (C) 2019 Intel Corporation.  All rights reserved.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+/opt/wasi-sdk/bin/clang \
```
loganek, line 5: I'd recommend taking compiler from env variable (similarly to https://github.com/WebAssembly/wasi-testsuite/blob/main/tests/c/build.sh#L4) and fall-back to `/opt/was-sdk/bin/clang` if not provided - not everybody install wasi-sdk in that location so there should be a way to override it.

core/iwasm/libraries/lib-socket/test/tcp.c
```diff
@@ -0,0 +1,109 @@
...
+    int sock = init_client_sock.sock;
+    struct sockaddr_in addr = init_client_sock.addr;
+    socklen_t addr_size;
+    char buffer[sizeof(SERVER_MSG) + 1];
+    assert(connect(sock, (struct sockaddr *)&addr, sizeof(addr)) == 0);
+
+    recv(sock, buffer, sizeof(buffer), 0);
+    assert(strcmp(buffer, SERVER_MSG) == 0);
+    assert(close(sock) == 0);
+    pthread_mutex_unlock(&mut);
```
loganek, line 90: can we unlock this mutex in L79?

core/iwasm/libraries/lib-socket/test/tcp.c
```diff
@@ -0,0 +1,109 @@
...
+    addr.sin_family = AF_INET;
+    addr.sin_port = PORT;
+    addr.sin_addr.s_addr = htonl(INADDR_LOOPBACK);
+    return (struct init_socket){ addr, sock };
+}
+
+void *
+server()
+{
+    pthread_mutex_lock(&mut);
```
loganek, line 42: we probably only need to lock the mutex when updating `server_init_complete` variable, i.e. in L53

core/iwasm/libraries/lib-socket/test/nslookup.c
```diff
@@ -0,0 +1,46 @@
...
+main(int argc, char **argv)
+{
+    struct addrinfo hints;
+    memset(&hints, 0, sizeof(hints));
+    int v4count = 0;
+    int v6count = 0;
+
+    char *url = "google-public-dns-a.google.com";
+    struct addrinfo *res;
+    for (int af = 0; af < 2; ++af) {
```
eloparco, line 26: any specific reason for using a loop here? maybe it would be more readable to export the logic into a function and call it twice, once for ipv4 and once for ipv6

.github/workflows/compilation_on_android_ubuntu.yml
```diff
@@ -392,6 +392,10 @@ jobs:
             $THREADS_TEST_OPTIONS,
             $WASI_TEST_OPTIONS,
           ]
+        wasi_sdk_release:
+          [
+            "https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-12/wasi-sdk-12.0-linux.tar.gz",
+          ]
```
lum1n0us, line 398: We have upgraded this dependency to wasi-sdk-19 since 27e7e160af54fe8a74620c4fbfacad480b5cb00c. Please rebase.

hritikgupta: Sure, thanks!

.github/workflows/compilation_on_android_ubuntu.yml
```diff
@@ -441,6 +445,13 @@ jobs:
       - name: checkout
         uses: actions/checkout@v3
 
+      - name: download and install wasi-sdk
+        run: |
+          cd /opt
+          sudo wget ${{ matrix.wasi_sdk_release }}
+          sudo tar -xzf wasi-sdk-12.0-*.tar.gz
```
wenyongh, line 452: 19.0

core/iwasm/libraries/lib-socket/test/build.sh
```diff
@@ -0,0 +1,14 @@
+
+# Copyright (C) 2019 Intel Corporation.  All rights reserved.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+CC="${CC:=/opt/wasi-sdk/bin/clang}"
+$CC \
```
loganek, line 6: to avoid duplication you can iterate through array instead.

core/iwasm/libraries/lib-socket/test/tcp_ipv4.c
```diff
@@ -0,0 +1,106 @@
...
+    assert(connect(sock, (struct sockaddr *)&addr, sizeof(addr)) == 0);
+
+    recv(sock, buffer, sizeof(buffer), 0);
+    assert(strcmp(buffer, SERVER_MSG) == 0);
+    assert(close(sock) == 0);
+    return NULL;
+}
+
+int
+main(int argc, char **argv)
```
loganek, line 92: There seems to be a lot of shared code between tcp_ipv4 and tcp_ipv6 test, we can potentially move the common code to separate file?

loganek:
It'd be good to cover [this](https://github.com/bytecodealliance/wasm-micro-runtime/pull/1981/files) issue as well here.

core/iwasm/libraries/lib-socket/test/tcp_ipv4.c
hritikgupta: done

core/iwasm/libraries/lib-socket/test/build.sh
hritikgupta: done

core/iwasm/libraries/lib-socket/test/nslookup.c
hritikgupta: done

.github/workflows/compilation_on_android_ubuntu.yml
```diff
@@ -526,9 +534,13 @@ jobs:
...
         run: sudo apt-get update && sudo apt install -y jq
 
+      - name: build socket api tests
+        run: ./build.sh
+        working-directory: ./core/iwasm/libraries/lib-socket/test/
+
       - name: run tests
         timeout-minutes: 10
-        run: ./test_wamr.sh ${{ matrix.test_option }} -t ${{ matrix.running_mode }}
+        run: ./core/iwasm/libraries/lib-socket/test/build.sh && ./test_wamr.sh ${{ matrix.test_option }} -t ${{ matrix.running_mode }}
```
eloparco, line 543: the build command here seems duplicated, the `build socket api tests` step already does that

core/iwasm/libraries/lib-socket/test/nslookup.c
```diff
@@ -0,0 +1,42 @@
...
+            address = address->ai_next;
+        }
+    }
+    assert(count > 0);
+}
+
+int
+main(int argc, char **argv)
+{
+    testNsLookup(AF_INET);  // invocation for ipv4
```
eloparco, line 40: maybe the comments here are not necessary, it's clear what's happening from the function parameter names; but in case you want to keep them, better to use c-style comments (i.e. `/* */`)

core/iwasm/libraries/lib-socket/test/tcp_udp.c
```diff
@@ -0,0 +1,190 @@
...
+    assert(strcmp(buffer, SERVER_MSG) == 0);
+    assert(close(sock) == 0);
+    return NULL;
+}
+
+void
+test_protocol(int family, int protocol)
+{
+    pthread_t server_thread, client_thread;
+    pthread_cond_init(&cond, NULL);
```
eloparco, line 165: I would put an assert for the values returned by pthread functions as well, if they don't succeed there's no point in continuing with the test

core/iwasm/libraries/lib-socket/test/build.sh
```diff
@@ -0,0 +1,15 @@
+
```
eloparco, line 1:
```suggestion
#!/bin/bash

#
# Copyright (C) 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
#
```

core/iwasm/libraries/lib-socket/test/nslookup.c
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright (C) 2019 Intel Corporation.  All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
```
eloparco, line 4:
```suggestion
/*
 * Copyright (C) 2023 Amazon.com Inc. or its affiliates. All rights reserved.
 * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 */
```

core/iwasm/libraries/lib-socket/test/nslookup.c
```diff
@@ -0,0 +1,42 @@
...
+    int count = 0;
+    struct addrinfo hints;
+    char *url = "google-public-dns-a.google.com";
+
+    memset(&hints, 0, sizeof(hints));
+    hints.ai_family = af;
+    hints.ai_socktype = SOCK_STREAM;
+    int ret = getaddrinfo(url, 0, &hints, &res);
+    assert(ret == 0);
+    if (ret == 0) {
```
eloparco, line 27: we already have the line above to check for the return code, that should be enough since this is a test (i.e. not meant to be run without the asserts)

core/iwasm/libraries/lib-socket/test/tcp_udp.c
```diff
@@ -0,0 +1,190 @@
...
+{
+    int sock = socket(family, protocol, 0);
+    assert(sock != -1);
+
+    struct socket_info info;
+    if (family == AF_INET) {
+        struct sockaddr_in addr;
+        memset(&addr, 0, sizeof(addr));
+        addr.sin_family = AF_INET;
+        addr.sin_port = PORT;
```
eloparco, line 43:
```suggestion
        addr.sin_port = htons(PORT);
```

core/iwasm/libraries/lib-socket/test/tcp_udp.c
```diff
@@ -0,0 +1,190 @@
...
+    strcpy(buffer, SERVER_MSG);
+
+    struct sockaddr *server_addr = (struct sockaddr *)&init_server_sock.addr;
+    int ret = bind(server_sock, server_addr,
+                   args->family == AF_INET ? sizeof(struct sockaddr_in)
+                                           : sizeof(struct sockaddr_in6));
+    assert(ret == 0);
+    pthread_mutex_lock(&mut);
+    server_init_complete = 1;
+    pthread_mutex_unlock(&mut);
```
eloparco, line 88: I think the `pthread_mutex_unlock()` should go after the `pthread_cond_signal()`

core/iwasm/libraries/lib-socket/test/tcp_udp.c
```diff
@@ -0,0 +1,190 @@
...
+}
+
+void *
+client(void *arg)
+{
+    struct thread_args *args = (struct thread_args *)arg;
+    assert_thread_args(args);
+
+    pthread_mutex_lock(&mut);
+    char buffer[sizeof(SERVER_MSG) + 1];
```
eloparco, line 121: probably this line can be moved at the beginning of the function (or before `assert_thread_args()`)

core/iwasm/libraries/lib-socket/test/build.sh
loganek: Also, please add `set -ueo pipefail` so the script fails if any file fails to compile.

core/iwasm/libraries/lib-socket/test/nslookup.c
```diff
@@ -0,0 +1,42 @@
...
+
+#include <assert.h>
+#include <string.h>
+#ifdef __wasi__
+#include <wasi_socket_ext.h>
+#endif
+#include <netdb.h>
+
+void
+testNsLookup(int af)
```
loganek, line 14: in WAMR we use `sneak_casing` everywhere, so it'd be good to stick to that in tests too.

core/iwasm/libraries/lib-socket/test/nslookup.c
```diff
@@ -0,0 +1,42 @@
...
+{
+
+    struct addrinfo *res;
```
loganek, line 17: even though it's a test, I think we should attempt to free the structure as well (e.g. if we ever run it with sanitizers, it'll give us some hints in case something goes wrong). Use `freeaddrinfo` for that.

core/iwasm/libraries/lib-socket/test/nslookup.c
```diff
@@ -0,0 +1,42 @@
...
+    int count = 0;
+    struct addrinfo hints;
+    char *url = "google-public-dns-a.google.com";
+
+    memset(&hints, 0, sizeof(hints));
+    hints.ai_family = af;
+    hints.ai_socktype = SOCK_STREAM;
```
loganek, line 24: Can we also validate that data that's being returned by `getaddrinfo` also matches the criteria?

core/iwasm/libraries/lib-socket/test/build.sh
```diff
@@ -0,0 +1,15 @@
...
+# Copyright (C) 2019 Intel Corporation.  All rights reserved.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+CC="${CC:=/opt/wasi-sdk/bin/clang}"
```
loganek, line 5: do we have to enable threads?

core/iwasm/libraries/lib-socket/test/tcp_udp.c
loganek: I don't think that really matters in this case.

core/iwasm/libraries/lib-socket/test/tcp_udp.c
```diff
@@ -0,0 +1,190 @@
...
+    strcpy(buffer, SERVER_MSG);
+
+    struct sockaddr *server_addr = (struct sockaddr *)&init_server_sock.addr;
+    int ret = bind(server_sock, server_addr,
+                   args->family == AF_INET ? sizeof(struct sockaddr_in)
+                                           : sizeof(struct sockaddr_in6));
+    assert(ret == 0);
+    pthread_mutex_lock(&mut);
+    server_init_complete = 1;
```
loganek, line 87: should this be set after calling `listen`?

core/iwasm/libraries/lib-socket/test/tcp_udp.c
```diff
@@ -0,0 +1,190 @@
...
+
+    int server_sock = init_server_sock.sock;
+    socklen_t addr_size;
+    struct sockaddr_storage client_addr;
+    char buffer[sizeof(SERVER_MSG) + 1];
+    strcpy(buffer, SERVER_MSG);
+
+    struct sockaddr *server_addr = (struct sockaddr *)&init_server_sock.addr;
+    int ret = bind(server_sock, server_addr,
+                   args->family == AF_INET ? sizeof(struct sockaddr_in)
```
loganek, line 83: you could probably assign it to `addr_size` and not set it again in [L91](https://github.com/bytecodealliance/wasm-micro-runtime/pull/1900/files#diff-ecd54200020a46e3e7f11d8460dd029aa59c389540c5034b011284982182cba8R91)

.github/workflows/compilation_on_android_ubuntu.yml
```diff
@@ -493,6 +497,13 @@ jobs:
       - name: checkout
         uses: actions/checkout@v3
 
+      - name: download and install wasi-sdk
+        run: |
```
wenyongh, line 501: This step and below step "build socket api tests" will be executed in each running mode and each test_option, which may take a lot of time to run the whole CI. Shall we add the test for some situations, for example, $WASI_TEST_OPTIONS test_option for all running modes? Refer to #1963, here and below, add condition:
```yml
if: matrix.test_option == '$WASI_TEST_OPTIONS'
```

hritikgupta:
@wenyongh / @loganek — addressed the changes, does this look good to merge?

core/iwasm/libraries/lib-socket/test/nslookup.c
```diff
@@ -0,0 +1,47 @@
...
+    assert(count > 0);
+    freeaddrinfo(res);
+}
+
+int
+main()
+{
+    test_nslookup(AF_INET);  /* for ipv4 */
+    test_nslookup(AF_INET6); /* for ipv6 */
+}
```
loganek, line 47: `wasi-testsuite` test runner checks for exit code, so I think it'd be better to add explicit `return` here (IIRC not specifying that is UB, although might be wrong about that).

.github/workflows/compilation_on_android_ubuntu.yml
```diff
@@ -457,6 +457,10 @@ jobs:
             $THREADS_TEST_OPTIONS,
             $WASI_TEST_OPTIONS,
           ]
+        wasi_sdk_release:
+          [
+            "https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-19/wasi-sdk-19.0-linux.tar.gz",
```
wenyongh, line 462: The same change was already added in #1963 so now no need to add this line here after it was merged.
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/.github/workflows/compilation_on_android_ubuntu.yml#L462

.github/workflows/compilation_on_android_ubuntu.yml
```diff
@@ -493,6 +497,14 @@ jobs:
       - name: checkout
         uses: actions/checkout@v3
 
+      - name: download and install wasi-sdk
+        if: matrix.test_option == '$WASI_TEST_OPTIONS'
+        run: |
+          cd /opt
+          sudo wget ${{ matrix.wasi_sdk_release }}
+          sudo tar -xzf wasi-sdk-19.0-*.tar.gz
+          sudo mv wasi-sdk-19.0 wasi-sdk
```
wenyongh, line 506: Same as above, see
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/.github/workflows/compilation_on_android_ubuntu.yml#L500-L523

.github/workflows/compilation_on_android_ubuntu.yml
```diff
@@ -526,6 +538,28 @@ jobs:
...
+          # "Rename thread_spawn import" commit on main branch
+          git fetch https://github.com/WebAssembly/wasi-libc \
+            8f5275796a82f8ecfd0833a4f3f444fa37ed4546
+          git checkout FETCH_HEAD
+          make \
+            AR=/opt/wasi-sdk/bin/llvm-ar \
+            NM=/opt/wasi-sdk/bin/llvm-nm \
+            CC=/opt/wasi-sdk/bin/clang \
+            THREAD_MODEL=posix
+        working-directory: core/deps
```
wenyongh, line 556: Same as above

wenyongh:
> @wenyongh / @loganek — addressed the changes, does this look good to merge?

It is good to me, but need to remove the duplicated changes with #1963.

.github/workflows/compilation_on_android_ubuntu.yml
hritikgupta: removed

core/iwasm/libraries/lib-socket/test/nslookup.c
hritikgupta: added

wenyongh (approved):
LGTM

PR #1902, opened 2023-01-19
Improve instantiate-time linking

lum1n0us:


yamt:
how an embedder can know which imports are already linked?

lum1n0us:
run `$ ./iwasm target.wasm` and collect messages like *warning: failed to link import ....* and those failures are required and others are linked by loader.

But we can provide a script to scan and report all need to be linked items in a proper order. Is it useful?

yamt:
are you suggesting to hardcode the list of imports linked by the loader in an embedder?

lum1n0us:
No. users should not repeat those imported items provided by loader and only need to give the "missing" parts. In most of the time, I believe just a few.

yamt:
a program using this api needs to know the list of linked imports. if not hardcoded, how?

lum1n0us:
Two ways:
- run $ ./iwasm target.wasm and collect messages like *warning: failed to link import ....* and those are required. 
- develop a script to scan the wasm and report all requirements.

yamt:
> Two ways:
> 
>     * run $ ./iwasm target.wasm and collect messages like _warning: failed to link import ...._ and those are required.
> 
>     * develop a script to scan the wasm and report all requirements.

do you mean to make the program execute it and parse the result?

lum1n0us:
Yes. the first idea which involve iwasm is. 

Or if it is wired, we can develop a script to analyze the .wasm.

yamt:
it sounds unreasonable to me. what's wrong to invent an api to query which imports are already-linked?

lum1n0us:
Even there is the api to query linked imports, user still need to write a small program to check first. Or they need to preparation all imports and filter out some of them with the query api when running.

yamt:
> Even there is the api to query linked imports, user still need to write a small program to check first. Or they need to preparation all imports and filter out some of them with the query api when running.

it sounds far more usable than actually running the module and parse warnings.
am i missing something?

lum1n0us:
*Develop a script to analyze the .wasm and report all necessaries* ?

yamt:
> _Develop a script to analyze the .wasm and report all necessaries_ ?

sorry, i don't understand what you mean.

lum1n0us:
We both agree it would be helpful and useable if there is a tool can tell developers which imports are required and which imports have been linked by loader. Right?

So I suggest we provide the tool(a script) directly instead of providing an query API and let users to create their own tools.

yamt:
> We both agree it would be helpful and useable if there is a tool can tell developers which imports are required and which imports have been linked by loader. Right?

no.
IMO it isn't usable without a proper api.

many of programs using wasm-c-api have a logic like this:
https://github.com/yamt/garbage/blob/master/wasm/capi/test.c#L77-L113
i can't imagine any reasonable way to adapt it with your external script approach.

lum1n0us:
So the *usable* way mentioned before are:
- First, let users prepare all imports no matter which are already linked.
- Then, Use a query API to filter out and form a proper import list and pass to `wasm_instance_new`.

yamt:
> So the _usable_ way mentioned before are ?
> 
>     * First, let users prepare all imports no matter which are already linked.
> 
>     * Then, Use a query API to filter out and form a proper import list and pass to `wasm_instance_new`.

i'm not sure what's the best api.

a possible api is to return a `wasm_exporttype_vec_t` which contains all builtin symbols including wasi.
and make `wasm_instance_new` skip processing them. (just check they are already linked at load-time.)

lum1n0us:
Maybe I missed something. Because there is a problem in my view. Developers have to prepare all imports (include potential built-ins). Because there is no such knowledge on which import is linked when they are programming unless I put it into a file (like DefinedSymbol.list) and developers can check it in design phase.

And `wasm-wasi` can have a lot of imports and usually the missing part (need developer to provide) is a few. I thought it is not worthy to create many `wasm_xxxtype_t` and `wasm_extern_t` but not use them eventually.

yamt:
> Maybe I missed something. Because there is a problem in my view. Developers have to prepare all imports (include potential built-ins). Because there is no such knowledge on which import is linked when they are programming unless I put it into a file (like DefinedSymbol.list) and developers can check it in design phase.
> 
> And `wasm-wasi` can have a lot of imports and usually the missing part (need developer to provide) is a few. I thought it is not worthy to create many `wasm_xxxtype_t` and `wasm_extern_t` but not use them eventually.

my view is that the set of imports and exports are known to the embedder until you load/instantiate a module.
even the set of available host funcs (both for link-time and instantiate-time) can be changed with something like https://github.com/bytecodealliance/wasm-micro-runtime/tree/main/samples/native-lib.

core/iwasm/common/wasm_c_api.c
```diff
@@ -4711,65 +4633,110 @@ aot_process_export(wasm_store_t *store, const AOTModuleInstance *inst_aot,
...
+                                            wasm_extern_as_func(import));
+                    }
+#endif
+                    if (ret == UNMATCHED) {
+                        LOG_WARNING("link #%d function failed", import_func_i);
+                        goto failed;
+                    }
+
+                    import_func_i++;
+                }
```
wenyongh, line 4682: Confusing about the link algorithm: why `import_func_i++` on each loop and import_func_i isn't set 0 at the beginning of each loop, it will skip the import func 0 to import_func_i when linking the next import function. And how to ensure their module_name/field_name are the same? Seems here we don't check them. And can the input imports be un-ordered?

My understanding is that: for each imports[i], if it is a function, then we should traversing all the import functions in the wasm module (from module->import_functions[0] to module->import_functions[import_function_count-1]), if module->import_functions[j] hasn't been linked and its module_name/field_name is same with imports[i]'s module_name/field_name, then we can link imports[i] to module->import_functions[j].

lum1n0us:
@try-agaaain please join us

IIUC, there are some pre-condition would infect the final design choices.

- Do developers know which *imports* require host implementations and which will be handled by WAMR? Usually, WAMR supports all `(import "wasi_snapshot_preview1"` generated by wasi-sdk when setting `WAMR_BUILD_LIBC_WASI=1`.  Based on that rule, it would be ok for developers to prepare all other imports beside `wasi_snaphost_preview`.

- Do developers know where *required imports* are. For me , there are two ways to find out indexes of an import item. The first way is using a tool (like wasm-objdump) to dump the import section. The second way is using `wasm_module_imports` to get `wasm_importtype_vec_t` and searching in it.  `WASM_API_EXTERN const wasm_name_t* wasm_importtype_name(const wasm_importtype_t*)` can return the field name. And WAMR should make sure the order of `wasm_moudle_imports()` return value is same with the order of the argument `imports` of `wasm_instance_new`.

- Should WAMR ignore or override a loader-linked import item if meets it in the argument `imports` of `wasm_instance_new()`. For example, WAMR would provide a built-in implementation for `wasi_snapshot_preview1.fd_seek`. But a developer passes another one via the `imports` of `wasm_instnace_new()` by accident or on purpose. I tend to ignore it because the developer should use `WAMR_BUILD_LIBC_WASI=0` if wants to use  a private implementation.

My proposal for how to create  the `imports` argument for `wasm_intance_new` would be:
- prepare a host function list covers all import requirements. Don't need to be worried about the order at this step. Will deal with it later. Usually, for a wasm32-wasi module, requirements are quite obviously.  Be sure to provide names.
- get the import list of current .wasm by `wasm_module_imports`. 
- create an `wasm_extern_vec_t* imports`. It should have same length with `wasm_importtype_vec_t` returned at previous step.
- fill the `imports` based on both name information provided by `wasm_importtype_t` and *the host function list*.

wenyongh:
@yamt @lum1n0us The discussion is mainly about how developer should prepare the import list for wasm_instance_new, right? There may be some questions:
 - Whether developer should fill all import functions/globals into the import list? Can the built-in native APIs be ignored in the import list?
 - Can the import functions/globals be unordered?
 - Can developer overwrite the runtime built-in native API with API written by himself?

My opinion is that (1) a full import functions/globals should be filled into the import list, if not, it may be difficult for developer to know which APIs are built-in and can be ignored (e.g. libc-builtin APIs), and also, developer may want to rewrite the native API by himself, (2) The import type can be unordered, but elements of each type should be ordered, for example, all ordered functions followed by all ordered globals, or, all ordered globals followed by all ordered functions, or, mixed elements of all ordered functions and all ordered globals.

I think it is really difficult for developer to create the import list, my suggestion is that we develop a tool to help generate the native API template, somewhat like the Java tool javah, the tool (suppose we name it as wamrh) generates header/source files with empty native APIs, import list and the register process, for example, suppose there is a wasm file below:
```wast
(module
  (import "wasi_snapshot_preview1" "args_get" (func $wasi_args_get (param i32 i32) (result i32)))
  (import "wasi_snapshot_preview1" "args_sizes_get" (func $wasi_args_sizes_get (param i32 i32) (result i32)))
  (import "env" "foo1" (func $foo1 (param i32) (param i64) (result i32)))
  (import "env" "foo2" (func $foo2 (param f32) (param f64) (result i32)))
  (func (export "main") (param $x i32) (param $y i32) (result i32)
        ...
  )
)
```
Note that it imports two wasi built-in APIs which have been implemented by runtime.
The wasm-c-api native API source file generated may be as below:
```C
#include <stdint.h>
#include <stdbool.h>

#include "wasm_c_api.h"

static wasm_trap_t *
foo1_callback(const wasm_val_vec_t *args, wasm_val_vec_t *results)
{
    int32_t arg0 = args->data[0].of.i32;
    int64_t arg1 = args->data[1].of.i64;

    /* TODO */
    return NULL;
}

static wasm_trap_t *
foo2_callback(const wasm_val_vec_t *args, wasm_val_vec_t *results)
{
    float arg0 = args->data[0].of.f32;
    double arg1 = args->data[0].of.f64;

    /* TODO */
    return NULL;
}

static wasm_instance_t *
wasm_instance_create(wasm_store_t *store, wasm_module_t *module)
{
    wasm_functype_t *type_foo1 = NULL;
    wasm_functype_t *type_foo2 = NULL;
    wasm_func_t *func_foo1 = NULL;
    wasm_func_t *func_foo2 = NULL;
    wasm_instance_t *instance = NULL;

    /* Create func types */
    type_foo1 = wasm_functype_new_2_1(
        wasm_valtype_new_i32(), wasm_valtype_new_i64(), wasm_valtype_new_i32());
    type_foo2 =
        wasm_functype_new_1_1(wasm_valtype_new_f32(), wasm_valtype_new_i32());
    if (!type_foo1 || !type_foo2)
        goto fail;

    /* Create functions */
    func_foo1 = wasm_func_new(store, type_foo1, foo1_callback);
    func_foo2 = wasm_func_new(store, type_foo2, foo2_callback);
    if (!func_foo1 || !func_foo2)
        goto fail;

    /* Create instance */
    wasm_extern_t *externs[] = {
        NULL, /* ignore for built-in ("wasi_snapshot_preview1" "args_get") */
        NULL, /* ignore for built-in ("wasi_snapshot_preview1" "args_sizes_get") */
        wasm_func_as_extern(func_foo1), wasm_func_as_extern(func_foo2)
    };
    wasm_extern_vec_t imports = WASM_ARRAY_VEC(externs);
    instance = wasm_instance_new(store, module, &imports, NULL);

fail:
    if (type_foo1)
        wasm_functype_delete(type_foo1);
    if (type_foo2)
        wasm_functype_delete(type_foo1);
    if (func_foo1)
        wasm_func_delete(func_foo1);
    if (func_foo2)
        wasm_func_delete(func_foo2);

    return instance;
}
```
By this way, developer can just **implement the empty APIs and call `wasm_instance_create` to create the wasm instance**, he dont' need to care about the prototype of the native APIs, the import list and the native API register process. It can greatly reduce the effort the developer.

And the template of the wamr runtime API may be like:
```C
#include <stdlib.h>
#include <stdint.h>
#include <stdbool.h>

#include "wasm_export.h"

static int32_t
foo1_wrapper(wasm_exec_env_t exec_env, int32_t arg0, int64_t arg1)
{
    wasm_module_inst_t module_inst = wasm_runtime_get_module_inst(exec_env);

    /* TODO */
    return 0;
}

static int32_t
foo2_wrapper(wasm_exec_env_t exec_env, float arg0, double arg1)
{
    wasm_module_inst_t module_inst = wasm_runtime_get_module_inst(exec_env);

    /* TODO */
    return 0;
}

#define REG_NATIVE_FUNC(func_name, signature) \
    { #func_name, func_name##_wrapper, signature, NULL }

static NativeSymbol native_symbols[] = {
    REG_NATIVE_FUNC(foo1, "(iI)i"),
    REG_NATIVE_FUNC(foo2, "(fF)i")
};

uint32_t
get_native_lib(char **p_module_name, NativeSymbol **p_native_symbols)
{
    *p_module_name = "env";
    *p_native_symbols = native_symbols;
    return sizeof(native_symbols) / sizeof(NativeSymbol);
}
```
Looking forward for your feedbacks.

try-agaaain:
> +  prepare a host function list covers all import requirements. Don't need to be worried about the order at this step. Will deal with it later. Usually, for a wasm32-wasi module, requirements are quite obviously. Be sure to provide names.
> + get the import list of current .wasm by wasm_module_imports.
> + create an wasm_extern_vec_t* imports. It should have same length with wasm_importtype_vec_t returned at previous step.
> + fill the imports based on both name information provided by wasm_importtype_t and the host function list.

@lum1n0us Thank you for your suggestion, it seems to be feasible.

However, when I tried to do this, in step four, I couldn't find a proper way to fill the `imports` with the information from `wasm_importtype_t`. Because for the compiler, `wasm_importtype_t` is an "incomplete type", which makes it impossible for me to access its fields.

Is there an existing API or another way that can complete step four?

yamt:
> @yamt @lum1n0us The discussion is mainly about how developer should prepare the import list for wasm_instance_new, right? There may be some questions:
> 
>     * Whether developer should fill all import functions/globals into the import list? Can the built-in native APIs be ignored in the import list?
> 
>     * Can the import functions/globals be unordered?
> 
>     * Can developer overwrite the runtime built-in native API with API written by himself?
> 
> 
> My opinion is that (1) a full import functions/globals should be filled into the import list, if not, it may be difficult for developer to know which APIs are built-in and can be ignored (e.g. libc-builtin APIs), and also, developer may want to rewrite the native API by himself, (2) The import type can be unordered, but elements of each type should be ordered, for example, all ordered functions followed by all ordered globals, or, all ordered globals followed by all ordered functions, or, mixed elements of all ordered functions and all ordered globals.

are you talking about wasm-c-api in general?
or about WAMR-specific extensions to it?
if the former, i guess (1) everything should be filled and (2) it should be ordered to match the order in module's imports.
if the latter, i guess we should make it explicit.

> 
> I think it is really difficult for developer to create the import list, my suggestion is that we develop a tool to help generate the native API template, somewhat like the Java tool javah, the tool (suppose we name it as wamrh) generates header/source files with empty native APIs, import list and the register process, for example, suppose there is a wasm file below:

IMO, it shouldn't be an offline tool because the list of native symbols can be dynamic.
(wasm_runtime_register_natives/wasm_runtime_unregister_natives)

> 
> ```webassembly
> (module
>   (import "wasi_snapshot_preview1" "args_get" (func $wasi_args_get (param i32 i32) (result i32)))
>   (import "wasi_snapshot_preview1" "args_sizes_get" (func $wasi_args_sizes_get (param i32 i32) (result i32)))
>   (import "env" "foo1" (func $foo1 (param i32) (param i64) (result i32)))
>   (import "env" "foo2" (func $foo2 (param f32) (param f64) (result i32)))
>   (func (export "main") (param $x i32) (param $y i32) (result i32)
>         ...
>   )
> )
> ```
> 
> Note that it imports two wasi built-in APIs which have been implemented by runtime. The wasm-c-api native API source file generated may be as below:
> 
> ```c
> #include <stdint.h>
> #include <stdbool.h>
> 
> #include "wasm_c_api.h"
> 
> static wasm_trap_t *
> foo1_callback(const wasm_val_vec_t *args, wasm_val_vec_t *results)
> {
>     int32_t arg0 = args->data[0].of.i32;
>     int64_t arg1 = args->data[1].of.i64;
> 
>     /* TODO */
>     return NULL;
> }
> 
> static wasm_trap_t *
> foo2_callback(const wasm_val_vec_t *args, wasm_val_vec_t *results)
> {
>     float arg0 = args->data[0].of.f32;
>     double arg1 = args->data[0].of.f64;
> 
>     /* TODO */
>     return NULL;
> }
> 
> static wasm_instance_t *
> wasm_instance_create(wasm_store_t *store, wasm_module_t *module)
> {
>     wasm_functype_t *type_foo1 = NULL;
>     wasm_functype_t *type_foo2 = NULL;
>     wasm_func_t *func_foo1 = NULL;
>     wasm_func_t *func_foo2 = NULL;
>     wasm_instance_t *instance = NULL;
> 
>     /* Create func types */
>     type_foo1 = wasm_functype_new_2_1(
>         wasm_valtype_new_i32(), wasm_valtype_new_i64(), wasm_valtype_new_i32());
>     type_foo2 =
>         wasm_functype_new_1_1(wasm_valtype_new_f32(), wasm_valtype_new_i32());
>     if (!type_foo1 || !type_foo2)
>         goto fail;
> 
>     /* Create functions */
>     func_foo1 = wasm_func_new(store, type_foo1, foo1_callback);
>     func_foo2 = wasm_func_new(store, type_foo2, foo2_callback);
>     if (!func_foo1 || !func_foo2)
>         goto fail;
> 
>     /* Create instance */
>     wasm_extern_t *externs[] = {
>         NULL, /* ignore for built-in ("wasi_snapshot_preview1" "args_get") */
>         NULL, /* ignore for built-in ("wasi_snapshot_preview1" "args_sizes_get") */
>         wasm_func_as_extern(func_foo1), wasm_func_as_extern(func_foo2)
>     };
>     wasm_extern_vec_t imports = WASM_ARRAY_VEC(externs);
>     instance = wasm_instance_new(store, module, &imports, NULL);
> 
> fail:
>     if (type_foo1)
>         wasm_functype_delete(type_foo1);
>     if (type_foo2)
>         wasm_functype_delete(type_foo1);
>     if (func_foo1)
>         wasm_func_delete(func_foo1);
>     if (func_foo2)
>         wasm_func_delete(func_foo2);
> 
>     return instance;
> }
> ```
> 
> By this way, developer can just **implement the empty APIs and call `wasm_instance_create` to create the wasm instance**, he dont' need to care about the prototype of the native APIs, the import list and the native API register process. It can greatly reduce the effort the developer.
> 
> And the template of the wamr runtime API may be like:
> 
> ```c
> #include <stdlib.h>
> #include <stdint.h>
> #include <stdbool.h>
> 
> #include "wasm_export.h"
> 
> static int32_t
> foo1_wrapper(wasm_exec_env_t exec_env, int32_t arg0, int64_t arg1)
> {
>     wasm_module_inst_t module_inst = wasm_runtime_get_module_inst(exec_env);
> 
>     /* TODO */
>     return 0;
> }
> 
> static int32_t
> foo2_wrapper(wasm_exec_env_t exec_env, float arg0, double arg1)
> {
>     wasm_module_inst_t module_inst = wasm_runtime_get_module_inst(exec_env);
> 
>     /* TODO */
>     return 0;
> }
> 
> #define REG_NATIVE_FUNC(func_name, signature) \
>     { #func_name, func_name##_wrapper, signature, NULL }
> 
> static NativeSymbol native_symbols[] = {
>     REG_NATIVE_FUNC(foo1, "(iI)i"),
>     REG_NATIVE_FUNC(foo2, "(fF)i")
> };
> 
> uint32_t
> get_native_lib(char **p_module_name, NativeSymbol **p_native_symbols)
> {
>     *p_module_name = "env";
>     *p_native_symbols = native_symbols;
>     return sizeof(native_symbols) / sizeof(NativeSymbol);
> }
> ```
> 
> Looking forward for your feedbacks.

wenyongh:
> > My opinion is that (1) a full import functions/globals should be filled into the import list, if not, it may be difficult for developer to know which APIs are built-in and can be ignored (e.g. libc-builtin APIs), and also, developer may want to rewrite the native API by himself, (2) The import type can be unordered, but elements of each type should be ordered, for example, all ordered functions followed by all ordered globals, or, all ordered globals followed by all ordered functions, or, mixed elements of all ordered functions and all ordered globals.
> 
> are you talking about wasm-c-api in general? or about WAMR-specific extensions to it? if the former, i guess (1) everything should be filled and (2) it should be ordered to match the order in module's imports. if the latter, i guess we should make it explicit.

It is a WAMR-specific tool and we may publish it with other binary files released. The usage may be like:
```C
wamrh --wasi-c-api -o test.c test.wasm     # generate native API template src file for wasm-c-api mode
wamrc --wamr-api -o test.c test.wasm       #generate native API template src file for wamr-api mode
```
We can name it wamrh or others, and the options can be discussed, more options may be provided.


> 
> > I think it is really difficult for developer to create the import list, my suggestion is that we develop a tool to help generate the native API template, somewhat like the Java tool javah, the tool (suppose we name it as wamrh) generates header/source files with empty native APIs, import list and the register process, for example, suppose there is a wasm file below:
> 
> IMO, it shouldn't be an offline tool because the list of native symbols can be dynamic. (wasm_runtime_register_natives/wasm_runtime_unregister_natives)

I think it can be an offline tool, the goal is to generate the empty import function definitions and the symbols array to register: for wasm-c-api mode, it generates the `wasm_extern_t *externs[]` array and a API `wasm_instance_create` to call `wasm_instance_new` to register the callbacks, developer can just call the `wasm_instance_create` to register the callbacks. For wamr-api mode, it generates `NativeSymbol native_symbols[]` array, and developer can register them with `wasm_runtime_register_natives`. See samples above for details.

It is like javah tool: https://docs.oracle.com/javase/8/docs/technotes/tools/windows/javah.html
IMO, wasm vm is somewhat like java vm, the jvm can provide the tool, why wasm vm can not provide such tool?

For a wasm file, the import functions are pre-specified in the wasm file, and some of them may be implemented by runtime builit-in native APIs, we may generate the empty functions that aren't implemented by runtime, or add options to control whether to generate their empty definitions.

nchamzn:
I'm probably missing something from the discussion but agree with @yamt that there should be an API or some way to generate this dynamically since the host does not know ahead of time what APIs the wasm is going to import from the runtime. A codegen/compile time solution wont work for my use case at least.

Is the suggestion above to prepare a list of *all* possible WASI imports, then *all* host defined native imports, pass it when instantiating the module and then having the runtime ignore imports from the list which are not required?

wenyongh:
> I'm probably missing something from the discussion but agree with @yamt that there should be an API or some way to generate this dynamically since the host does not know ahead of time what APIs the wasm is going to import from the runtime. A codegen/compile time solution wont work for my use case at least.
> 
> Is the suggestion above to prepare a list of _all_ possible WASI imports, then _all_ host defined native imports, pass it when instantiating the module and then having Wamr ignore imports from the list which are not required?

Yes, the suggested tool is to generate a list of both all possible WASI imports and all host defined native imports, but there may be option to control how to fill all possible WASI imports into the list: (1) fill the list with NULL pointers and runtime just ignores them since there are built-in WASI import implementations of them, (2) generate empty WASI import definitions, but the code is commented, and if developer wants to re-write the import, he can uncomment the code and fill the list with the related function pointer.

For the non-wasi (or non-builtin) imports, the tool generates the empty definitions, so developer can just implement them.
For example, if a wasm module imports two wasi APIs and two host defined APIs:
```wast
  (import "wasi_snapshot_preview1" "args_get" (func $wasi_args_get (param i32 i32) (result i32)))
  (import "wasi_snapshot_preview1" "args_sizes_get" (func $wasi_args_sizes_get (param i32 i32) (result i32)))
  (import "env" "foo1" (func $foo1 (param i32) (param i64) (result i32)))
  (import "env" "foo2" (func $foo2 (param f32) (param f64) (result i32)))
```
Then the tool generates the empty definitions of the host defined APIs:
```C
static wasm_trap_t *
foo1_callback(const wasm_val_vec_t *args, wasm_val_vec_t *results)
{
    int32_t arg0 = args->data[0].of.i32;
    int64_t arg1 = args->data[1].of.i64;

    /* developer to implement the API */
    return NULL;
}

static wasm_trap_t *
foo2_callback(const wasm_val_vec_t *args, wasm_val_vec_t *results)
{
    float arg0 = args->data[0].of.f32;
    double arg1 = args->data[0].of.f64;

    /* developer to implement the API */
    return NULL;
}
```
And then it generates an API, which creates a import list and then invokes wasm_instance_new with the import list to register all the imports:
```C
static wasm_instance_t *
wasm_instance_create(wasm_store_t *store, wasm_module_t *module)
{
    ...
    /* Create functions */
    func_foo1 = wasm_func_new(store, type_foo1, foo1_callback); 
    func_foo2 = wasm_func_new(store, type_foo2, foo2_callback);
    if (!func_foo1 || !func_foo2)
        goto fail;

    /* Create instance */
    wasm_extern_t *externs[] = {                       ==> The import list
        NULL, /* ignore for built-in ("wasi_snapshot_preview1" "args_get") */
        NULL, /* ignore for built-in ("wasi_snapshot_preview1" "args_sizes_get") */
        wasm_func_as_extern(func_foo1),           ==> user defined API
        wasm_func_as_extern(func_foo2)            ==> user defined API
    };
    wasm_extern_vec_t imports = WASM_ARRAY_VEC(externs);
    instance = wasm_instance_new(store, module, &imports, NULL);      ==> new the wasm module with the import list
    ...
    return instance;
}
```
Then developer can call this API `wasm_instance_create` to new the wasm module instance. He can integrate this source file with his project.
I think this way should be friendly and clearly, and it can reduce the effort of developer.
Not sure whether it meets the requirements of you?

lum1n0us:
@try-agaaain The way of using `wasm_importype_vect_t` to create an import list likes below.  It only includes the key part but not a fully runnable function.

``` c
static wasm_instance_t *
wasm_instance_create(wasm_store_t *store, wasm_module_t *module) {
    wasm_importtype_vec_t import_types = { 0 };
    wasm_module_imports(module, &import_types);

    wasm_extern_t *externs[4] = { 0 };

    for (unsigned i = 0; i < import_types.num_elems; i++) {
        wasm_importtype_t *import_type = &(import_types.data[i]);
        wasm_byte_vec_t *module_name = wasm_importtype_module(import_type);
        wasm_byte_vec_t *field_name = wasm_importtype_name(import_type);
        char module_name_str[64] = { 0 };
        char field_name_str[64] = { 0 };

        memcpy(module_name_str, module_name->data, module_name->num_elems);
        memcpy(field_name_str, field_name->data, field_name->num_elems);

        /* ignore for built-int. It is WAMR's duty to provide imports of
         * `wasm_snapshot_preview1 */
        if (strncmp(module_name_str, "wasm_snapshot_preview1",
                    strlen("wasm_snapshot_preview1"))
            == 0)
            continue;

        if (strncmp(field_name_str, "foo1", strlen("foo1")) == 0) {
            wasm_functype_t *type_foo1 = wasm_functype_new_2_1(
                wasm_valtype_new_i32(), wasm_valtype_new_i64(),
                wasm_valtype_new_i32());
            wasm_func_t *func_foo1 =
                wasm_func_new(store, type_foo1, foo1_callback);

            externs[i] = wasm_func_as_extern(func_foo1);
        }
        else if (strncmp(field_name_str, "foo2", strlen("foo2")) == 0) {
            wasm_functype_t *type_foo2 = wasm_functype_new_1_1(
                wasm_valtype_new_f32(), wasm_valtype_new_i32());
            wasm_func_t *func_foo2 =
                wasm_func_new(store, type_foo2, foo2_callback);

            externs[i] = wasm_func_as_extern(func_foo2);
        }
    }

    wasm_extern_vec_t imports = WASM_ARRAY_VEC(externs);
    /* Create instance */
    wasm_instance_t *instance =
        wasm_instance_new(store, module, &imports, NULL);
}
```

- use `wasm_moudle_import` to retrieve `wasm_importtype_t` list. The big advantage here is there are name information in `wasm_importtype_t`.
- create a `wasm_externvec_t` as a target to filled in. It should have the same length of `wasm_importtype_vec_t` from previous step.
- traversal `wasm_importtype_t` in a loop. Since it has the same required order with `wasm_importtype_vec_t`, it is convenient to check the name of import. 
- It should be Wasm runtime's duty to provide all import with a module name like *wasm_snapshot_preview1*. It is a big assumption and it is the key point to allow developers to form `imports` without scanning .wasm.

> many of programs using wasm-c-api have a logic like this:
> https://github.com/yamt/garbage/blob/master/wasm/capi/test.c#L77-L113
> i can't imagine any reasonable way to adapt it with your external script approach.

@yamt  IIUC, the only difference is let Wasm runtime handle "standard APIs" and let users handle "private functions"

@wenyongh  The script idea will work well if wasm-c-api users just have a few target .wasm. It would be hard when the amount of target .wasm growing. 

If we are going to suggest wasm-c-api users to create the imports list based on name information from `wasm_importtype_t`, 
- make sure `wasm_importtype_vec_t` from `wasm_module_import` has the proper order which `wasm_instance_new` required.
- make sure always be compatible with wasi-libc. May need a CI to compare `__imported_wasi_snapshot_preview1_XXX` in https://github.com/WebAssembly/wasi-libc/blob/b4814997f61ee352d8c1ae397561a813fb30c701/expected/wasm32-wasi/undefined-symbols.txt with `native_symbols_libc_wasi`?

wenyongh:
> @wenyongh The script idea will work well if wasm-c-api users just have a few target .wasm. It would be hard when the amount of target .wasm growing.

@lum1n0us Not sure what `the amount of target .wasm` means? Does it mean the amount of the import functions? If yes, wondering why it is hard when the amount of import functions grows? For developer, after generating the template file, normally there are two things to do:
- Implement the empty functions
- Integrate the file into his project, and call `wasm_instance_create` to create the instance with the imports

For the first thing, he should do it no matter using the tool or not. For the second thing, it should be simple enough: he doesn't need to prepare the import list and create the function types and the functions for the list, and he doesn't need to care the import function order inside the import list, all these are done by tool.

> 
> If we are going to suggest wasm-c-api users to create the imports list based on name information from `wasm_importtype_t`,
> 
> * make sure `wasm_importtype_vec_t` from `wasm_module_import` has the proper order which `wasm_instance_new` required.

Seems it is difficult (at least not very easy) for developer to prepare the list with proper order, firstly he should know what the is order required by runtime, secondly he should know what the imports are in the wasm file and what the order of them is, normally he should use special tool (e.g. wasm-objdump, wasm2wat) to do it.

yamt:
> > > My opinion is that (1) a full import functions/globals should be filled into the import list, if not, it may be difficult for developer to know which APIs are built-in and can be ignored (e.g. libc-builtin APIs), and also, developer may want to rewrite the native API by himself, (2) The import type can be unordered, but elements of each type should be ordered, for example, all ordered functions followed by all ordered globals, or, all ordered globals followed by all ordered functions, or, mixed elements of all ordered functions and all ordered globals.
> > 
> > 
> > are you talking about wasm-c-api in general? or about WAMR-specific extensions to it? if the former, i guess (1) everything should be filled and (2) it should be ordered to match the order in module's imports. if the latter, i guess we should make it explicit.
> 
> It is a WAMR-specific tool and we may publish it with other binary files released. The usage may be like:
> 
> ```c
> wamrh --wasi-c-api -o test.c test.wasm     # generate native API template src file for wasm-c-api mode
> wamrc --wamr-api -o test.c test.wasm       #generate native API template src file for wamr-api mode
> ```
> 
> We can name it wamrh or others, and the options can be discussed, more options may be provided.
> 
> > > I think it is really difficult for developer to create the import list, my suggestion is that we develop a tool to help generate the native API template, somewhat like the Java tool javah, the tool (suppose we name it as wamrh) generates header/source files with empty native APIs, import list and the register process, for example, suppose there is a wasm file below:
> > 
> > 
> > IMO, it shouldn't be an offline tool because the list of native symbols can be dynamic. (wasm_runtime_register_natives/wasm_runtime_unregister_natives)
> 
> I think it can be an offline tool, the goal is to generate the empty import function definitions and the symbols array to register: for wasm-c-api mode, it generates the `wasm_extern_t *externs[]` array and a API `wasm_instance_create` to call `wasm_instance_new` to register the callbacks, developer can just call the `wasm_instance_create` to register the callbacks. For wamr-api mode, it generates `NativeSymbol native_symbols[]` array, and developer can register them with `wasm_runtime_register_natives`. See samples above for details.
> 
> It is like javah tool: https://docs.oracle.com/javase/8/docs/technotes/tools/windows/javah.html IMO, wasm vm is somewhat like java vm, the jvm can provide the tool, why wasm vm can not provide such tool?
> 
> For a wasm file, the import functions are pre-specified in the wasm file, and some of them may be implemented by runtime builit-in native APIs, we may generate the empty functions that aren't implemented by runtime, or add options to control whether to generate their empty definitions.

it can be an offline tool and we can provide it. maybe it works for some static usage.

however, there are use cases for which such an offline tool is not appropriate.
for my use case, none of following is known ahead of time:
* available set of native symbols
* what a wasm module imports

so, my suggestion is to provide a runtime api for an embedder to query which imports are already linked at load time. the embedder then can/should use a dummy extern entry like `NULL` for them.

wenyongh:
> 
> it can be an offline tool and we can provide it. maybe it works for some static usage.
> 
> however, there are use cases for which such an offline tool is not appropriate. for my use case, none of following is known ahead of time:
> 
> * available set of native symbols
> * what a wasm module imports
> 
> so, my suggestion is to provide a runtime api for an embedder to query which imports are already linked at load time. the embedder then can/should use a dummy extern entry like `NULL` for them when calling wasm_instance_new().

Got it, it sounds reasonable, shall we query each kind of import? How about:
```C
bool wasm_runtime_is_import_func_linked(const char *module_name, const char *field_name);
bool wasm_runtime_is_import_global_linked(const char *module_name, const char *field_name);
```
Import memory and import table are not supported.

BTW, do you agree that for wasm_module_new, developer should provide a full list of the imports, and the order of each import type (functions or globals) should be same as the order of them in the wasm file? And if a NULL func pointer is filled, runtime can just ignore it?

nchamzn:
> my suggestion is to provide a runtime api for an embedder to query which imports are already linked at load time. the embedder then can/should use a dummy extern entry like NULL for them when calling wasm_instance_new().

Just to solidify with a use case, consider for example I have a Wamr application with a plugin system - the plugin system has an API where the plugin can define a set of functions which can be used by Wasm at runtime. Plugins can be developed at any time, including tomorrow after I ship my application, so I don't know what functions might be used in the future. The Wasm run on the device may or may not make use of the plugin for whatever reason.

Further to that, I might also add some code down the line which adds or removes the use of some Wasi API and so I can't know ahead of time which Wasi functions are required from the runtime either.

With the scenario described above:
* I need to be able to dynamically generate a list of imports provided from the application including its plugins.
* I need to be able to know all supported wasi functions from Wamr.

The suggestion above, where we can just get a list of all functions from Wamr makes sense to me, it can then be appended to my list of plugin functions, and passed to Wamr when instantiating the module, so long as unused/unlinked functions are ignored, and order doesn't matter.

Note that this scenario was already kind of working with register_natives. I keep a list of registered native functions which I then register with Wamr using the register_natives function. When instantiating the module I was passing NULL as the imports vector, and Wamr would work everything out, but the functionality broke recently. It would be preferable though if I could use the wasm C api to link all these imports at instantiate time because managing all these functions though the engine singleton means I need to be careful particularly if I have wasm running on multiple threads not to unload plugins.

wenyongh:
> > my suggestion is to provide a runtime api for an embedder to query which imports are already linked at load time. the embedder then can/should use a dummy extern entry like NULL for them when calling wasm_instance_new().
> 
> Just to solidify with a use case, consider for example I have a Wamr application with a plugin system - the plugin system has an API where the plugin can define a set of functions which can be used by Wasm at runtime. Plugins can be developed at any time, including tomorrow after I ship my application, so I don't know what functions might be used in the future. The Wasm run on the device may or may not make use of the plugin for whatever reason.
> 
> Further to that, I might also add some code down the line which adds or removes the use of some Wasi API and so I can't know ahead of time which Wasi functions are required from the runtime either.
> 
> With the scenario described above:
> 
> * I need to be able to dynamically generate a list of imports provided from the application including its plugins.
> * I need to be able to know all supported wasi functions from Wamr.
> 
> The suggestion above, where we can just get a list of all functions from Wamr makes sense to me, it can then be appended to my list of plugin functions, and passed to Wamr when instantiating the module, so long as unused/unlinked functions are ignored, and order doesn't matter.

Do you mean that the API `bool wasm_runtime_is_import_func_linked(const char *module_name, const char *func_name)` makes sense? Or runtime should provide another API to get all the linked functions? The latter API may be a little complex.

BTW, the order of the functions does matter, my understanding is that the order of functions in the import list should be exactly same with the order in the wasm file, since the import list doesn't take the info of import function's module name and function name, runtime cannot know which import func of the instance to new should be linked to for each function in the import list. It seems runtime had better link them by the index in the import list (suppose no globals in import list): `imports[i]` is linked to `instance->imports[i]`, and if `imports[i]'s func_ptr` is NULL, runtime just ignores linking.

nchamzn:
I think that `wasm_runtime_is_import_func_linked` would work fine, if I load the module then ahead of instantiation can get all the function imports and use this API to check if I should set as NULL in the imports vector.

Modifying code form above, but something like:
```
static wasm_instance_t *
wasm_instance_create(wasm_store_t *store, wasm_module_t *module) {
    wasm_importtype_vec_t import_types = { 0 };
    wasm_module_imports(module, &import_types);

    wasm_extern_t *externs[4] = { 0 };

    for (unsigned i = 0; i < import_types.num_elems; i++) {
        wasm_importtype_t *import_type = &(import_types.data[i]);
        wasm_byte_vec_t *module_name = wasm_importtype_module(import_type);
        wasm_byte_vec_t *field_name = wasm_importtype_name(import_type);
        char module_name_str[64] = { 0 };
        char field_name_str[64] = { 0 };

        memcpy(module_name_str, module_name->data, module_name->num_elems);
        memcpy(field_name_str, field_name->data, field_name->num_elems);

        if (wasm_runtime_is_import_func_linked(module_name_str, field_name_str)) {
            continue;
        }

        if (is_function_from_a_plugin(module_name_str, field_name_str)) {
            // Set up import from host function
        }
    }

    wasm_extern_vec_t imports = WASM_ARRAY_VEC(externs);
    wasm_instance_t *instance =
        wasm_instance_new(store, module, &imports, NULL);
}

```

yamt:
> > it can be an offline tool and we can provide it. maybe it works for some static usage.
> > however, there are use cases for which such an offline tool is not appropriate. for my use case, none of following is known ahead of time:
> > 
> > * available set of native symbols
> > * what a wasm module imports
> > 
> > so, my suggestion is to provide a runtime api for an embedder to query which imports are already linked at load time. the embedder then can/should use a dummy extern entry like `NULL` for them when calling wasm_instance_new().
> 
> Got it, it sounds reasonable, shall we query each kind of import? How about:
> 
> ```c
> bool wasm_runtime_is_import_func_linked(const char *module_name, const char *func_name);
> bool wasm_runtime_is_import_global_linked(const char *module_name, const char *global_name);
> ```
> 
> Import memory and import table are not supported.
> 
> BTW, do you agree that for wasm_module_new, developer should provide a full list of the imports, and the order of each import type (functions or globals) should be same as the order of them in the wasm file? And if a NULL func pointer is filled, runtime can just ignore it?

i was thinking something like:
```
bool wasm_importtype_linked(wasm_importtype_t *);
```
to examine the result of `wasm_module_imports`.

i guess your `wasm_runtime_is_import_func_linked`/`wasm_runtime_is_import_global_linked` need to take a type. (function signature incase of func)
spec-wise wasm allows type-based overloading with the same import name. while wamr might or might not want to support it, it's better not to exclude it in the api. how do you think?

nchamzn:
> i was thinking something like:
>
> bool wasm_importtype_linked(wasm_importtype_t *);

This requires a bit less work for the embedder - I guess we can also propose this for addition to [wasm-c-api](https://github.com/WebAssembly/wasm-c-api)?

wenyongh:
> i was thinking something like:
> 
> ```
> bool wasm_importtype_linked(wasm_importtype_t *);
> ```
> 
> to examine the result of `wasm_module_imports`.
> 
> i guess your `wasm_runtime_is_import_func_linked`/`wasm_runtime_is_import_global_linked` need to take a type. (function signature incase of func) spec-wise wasm allows type-based overloading with the same import name. while wamr might or might not want to support it, it's better not to exclude it in the api. how do you think?

Maybe we can provide both `wasm_runtime_is_import_func_linked` and `wasm_is_importtype_linked`, the former is declared in wasm_export.h and the latter is declared in wasm_c_api.h. For developer using WAMR API in wasm_export.h, he can use the former to check whether an import func is linked, for developer using wasm-c-api, he can use the latter. And both of them require signature. The prototype may be:
```C
bool
wasm_runtime_is_import_func_linked(const char *module_name, const char *func_name,
                                                            const char *signature);
bool
wasm_runtime_is_import_global_linked(const char *module_name, const char *global_name);

bool
wasm_is_importtype_linked(wasm_importtype_t *import_type);
```
For the `signature` argument in wasm_runtime_is_import_func_linked, the supported format is like the format of signature in wasm_runtime_register_natives, character `i`/`I`/`f`/`F` mean i32/i64/f32/f64 types, e.g. "(iI)f" mean func argument type (i32, i64) and return type f32.

And if the above is OK, maybe we can implement these APIs and fix the import linking process in wasm-c-api accordingly.

yamt:
> Maybe we can provide both wasm_runtime_is_import_func_linked and wasm_is_importtype_linked, the former is declared in wasm_export.h and the latter is declared in wasm_c_api.h.

do you imply we might introduce instantiate-time linking for wasm_export.h api?
(otherwise this functionality doesn't seem necessary there.)

lum1n0us:
we tend to keep "loading-time-linking" in wasm_export.h and keep "instantiate-time-linking" in wasm_c_api.h both. Hope we can maintain the capability on both sides or else we might be looking at a big U-turn.

core/iwasm/libraries/lib-pthread/lib_pthread_wrapper.c
```diff
@@ -648,6 +685,8 @@ pthread_create_wrapper(wasm_exec_env_t exec_env,
         wasm_runtime_free(info_node);
     if (routine_args)
         wasm_runtime_free(routine_args);
+    if (new_c_api_func_imports)
+        wasm_runtime_free(new_c_api_func_imports);
```
wenyongh, line 689: Should not free again, one is the allocated memory is `*new_c_api_func_imports`, but not `new_c_api_func_imports`, see L634, the other is that the `*new_c_api_func_imports` is `new_module_inst->e->c_api_func_imports`, if it was set, it will have been freed in L683 with `wasm_deinstantiate` and `aot_deinstantiate`, see:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_runtime.c#L2167-L2168

samples/how-to-deal-with-import/wasm/inc/pthread.h
```diff
@@ -0,0 +1,91 @@
+/*
```
wenyongh, line 1: This file already exists in `wamr-sdk/app/libc-builtin-sysroot/include/pthread.h`, we had better only maintain one copy, could we copy it to here in CMakeLists.txt or add `wamr-sdk/app/libc-builtin-sysroot/include` to include dir?

samples/how-to-deal-with-import/host/example1.c
```diff
@@ -0,0 +1,181 @@
+#include <stdlib.h>
```
wenyongh, line 1: Should add license header

samples/how-to-deal-with-import/CMakeLists.txt
```diff
@@ -0,0 +1,95 @@
+# Copyright (C) 2019 Intel Corporation.  All rights reserved.
```
wenyongh, line 1: The sample is only about how to handle imports in wasm-c-api, it doesn't include the part of wamr apis, which has been demonstrated in `samples/native-lib`, how about renaming this folder to `wasm-c-api-imports` and adding README.md?

core/iwasm/common/wasm_c_api.c
```diff
@@ -4711,65 +4650,108 @@ aot_process_export(wasm_store_t *store, const AOTModuleInstance *inst_aot,
...
+    }
+#endif
 
-    for (i = 0; imports && i < imports->num_elems; i++) {
+    /* we have run a module_type check before. */
+
+    /* 1. imported functions */
+    for (i = 0, import_func_i = 0;
+         i < imports->num_elems && import_func_i < import_func_count;
+         i++, import_func_i++) {
```
wenyongh, line 4684: Must the import types of functions and globals in the import list of wasm_instance be ordered? e.g. first goud of functions, and then the group of globals? Could globals be before functions, or mixed of them with ordered functions and ordered globals?

samples/how-to-deal-with-import/wasm/inc/pthread.h
lum1n0us: Will copy *pthread.h* to `wasm/inc` during configuration phase.

Adding *wamr-sdk/app/libc-builtin-sysroot/include* into headers search paths will leads to conflicts likes:

```
warning: call to undeclared function 'UINT16_C'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
error: use of undeclared identifier 'errno'
```

core/iwasm/common/wasm_c_api.c
lum1n0us: The new recommendation is different with before.  

Before, we required developers to make a list of import and keep them in the right order. But now, we only ask developers to prepare all necessary *imports* and follow the order information providing by `wasm_module_imports()`. That is what `build_imports()` does.

samples/how-to-deal-with-import/CMakeLists.txt
lum1n0us: Sure. Renaming the directory firstly, Upload README.md later.

core/iwasm/common/wasm_c_api.c
```diff
@@ -4501,11 +4487,16 @@ aot_link_func(const wasm_instance_t *inst, const AOTModule *module_aot,
...
+    /* it is handled by loader durning validation */
+    if (import_aot_func->is_linked == VALIDATION_LINKED)
+        return true;
+
     /* type comparison */
     if (!wasm_functype_same_internal(import->type, import_aot_func->func_type))
         return false;
 
     import_aot_func->call_conv_wasm_c_api = true;
+    import_aot_func->is_linked = INSTANTIATION_LINKED;
```
wenyongh, line 4499: Shall we check import's func ptr and ignore the link if the func ptr is NULL?

core/iwasm/common/wasm_c_api.c
```diff
@@ -4711,65 +4647,99 @@ aot_process_export(wasm_store_t *store, const AOTModuleInstance *inst_aot,
...
+                }
+#endif
+#if WASM_ENABLE_AOT != 0
+                if ((*module)->module_type == Wasm_Module_AoT) {
+                    ret = aot_link_func(inst, MODULE_AOT(module), import_func_i,
+                                        wasm_extern_as_func(import));
+                }
+#endif
+                if (!ret) {
+                    LOG_WARNING("link #%d function failed", import_func_i);
```
wenyongh, line 4683: How about `"link function #%d failed"

core/iwasm/common/wasm_c_api.c
```diff
@@ -4711,65 +4647,99 @@ aot_process_export(wasm_store_t *store, const AOTModuleInstance *inst_aot,
...
+                }
+#endif
+#if WASM_ENABLE_AOT != 0
+                if ((*module)->module_type == Wasm_Module_AoT) {
+                    ret = aot_link_global(MODULE_AOT(module), import_global_i,
+                                          wasm_extern_as_global(import));
+                }
+#endif
+                if (!ret) {
+                    LOG_WARNING("link #%d global failed", import_global_i);
```
wenyongh, line 4707: How about `link global #%d failed`

core/iwasm/common/wasm_c_api.c
```diff
@@ -4856,18 +4798,24 @@ wasm_instance_new_with_args(wasm_store_t *store, const wasm_module_t *module,
...
 
     /* create the c-api func import list */
-    inst_rt = (WASMModuleInstance *)instance->inst_comm_rt;
+    inst_rt = NULL;
 #if WASM_ENABLE_INTERP != 0
     if (instance->inst_comm_rt->module_type == Wasm_Module_Bytecode) {
-        p_func_imports = &inst_rt->e->c_api_func_imports;
-        import_func_count = inst_rt->module->import_function_count;
+        inst_rt = instance->inst_comm_rt;
+        WASMModuleInstanceExtra *e = ((WASMModuleInstance *)inst_rt)->e;
```
wenyongh, line 4805: Had better declare `WASMModuleInstanceExtra *e` at the beginning of `{ .. }`

core/iwasm/common/wasm_c_api.c
```diff
@@ -4856,18 +4798,24 @@ wasm_instance_new_with_args(wasm_store_t *store, const wasm_module_t *module,
...
+        import_func_count = MODULE_INTERP(module)->import_function_count;
     }
 #endif
 #if WASM_ENABLE_AOT != 0
     if (instance->inst_comm_rt->module_type == Wasm_Module_AoT) {
-        p_func_imports =
-            &((AOTModuleInstanceExtra *)inst_rt->e)->c_api_func_imports;
-        import_func_count = ((AOTModule *)inst_rt->module)->import_func_count;
+        inst_rt = instance->inst_comm_rt;
+        AOTModuleInstanceExtra *e =
```
wenyongh, line 4813: Had better declare `WASMModuleInstanceExtra *e` at the beginning of `{ .. }`

core/iwasm/include/wasm_c_api.h
```diff
@@ -797,6 +798,39 @@ static inline void* wasm_val_ptr(const wasm_val_t* val) {
 
 #define KILOBYTE(n) ((n) * 1024)
 
+// Create placeholders filled in `wasm_externvec_t* imports` for `wasm_instance_new()`
+static inline wasm_func_t *wasm_func_new_dummy(wasm_store_t*store)
+{
+  wasm_functype_t *type = wasm_functype_new_0_0();
+  wasm_func_t *func = wasm_func_new(store, type, (void*)0xDEADDEAD);
```
wenyongh, line 805: `0xDEADDEAD` isn't `must to be`, right? Developer can use `NULL`?

core/iwasm/include/wasm_c_api.h
```diff
@@ -797,6 +798,39 @@ static inline void* wasm_val_ptr(const wasm_val_t* val) {
 
 #define KILOBYTE(n) ((n) * 1024)
 
+// Create placeholders filled in `wasm_externvec_t* imports` for `wasm_instance_new()`
+static inline wasm_func_t *wasm_func_new_dummy(wasm_store_t*store)
+{
+  wasm_functype_t *type = wasm_functype_new_0_0();
```
wenyongh, line 804: Should we check the return value of type?

core/iwasm/include/wasm_c_api.h
```diff
@@ -797,6 +798,39 @@ static inline void* wasm_val_ptr(const wasm_val_t* val) {
...
+  wasm_func_t *func = wasm_func_new(store, type, (void*)0xDEADDEAD);
+  wasm_functype_delete(type);
+  return func;
+}
+
+static inline wasm_global_t *wasm_global_new_dummy(wasm_store_t*store)
+{
+  wasm_valtype_t *valtype = wasm_valtype_new_i32();
+  wasm_globaltype_t *type = wasm_globaltype_new(valtype, true);
```
wenyongh, line 813: Should we check the return value of valtype and type?

core/iwasm/include/wasm_c_api.h
```diff
@@ -797,6 +798,39 @@ static inline void* wasm_val_ptr(const wasm_val_t* val) {
...
+  wasm_val_t init = WASM_I32_VAL(0xDEADDEAD);
```
wenyongh, line 814: `0xDEADDEAD` isn't `must to be`, right? Developer can use 0?

samples/wasm-c-api-imports/README.md
```diff
@@ -0,0 +1,171 @@
+# How to create `import` for wasm_instance_new() properly
```
wenyongh, line 1: how about `import` -> `import list` or `imports`

samples/wasm-c-api-imports/README.md
```diff
@@ -0,0 +1,171 @@
...
+## A natural way
+
+One natural answer is "to create a list which matches every item in _the import
+section_" of the .wasm. Since developers can see the section details of
+a .wasm by tools like _wasm-objdump_, the answer is doable. Most of the time,
+if they also prepare Wasm modules, developers have full control over import
+requirements, and they only need to take a look at the order of _the import
+section_.
+
+Yes, _the order_. A proper `wasm_extern_vec_t *import` includes two things:
```
wenyongh, line 53: imports

samples/wasm-c-api-imports/README.md
```diff
@@ -0,0 +1,171 @@
...
+
+1. how many `wasm_extern_t`
+2. and order of those
+
+Because there is no "name information" in a `wasm_extern_t`. The only way to let
```
wenyongh, line 58: `to let` -> `is to let`?

core/iwasm/common/wasm_c_api.c
```diff
@@ -5202,3 +5198,16 @@ BASIC_FOUR_LIST(WASM_EXTERN_AS_OTHER_CONST)
 
 BASIC_FOUR_LIST(WASM_OTHER_AS_EXTERN_CONST)
 #undef WASM_OTHER_AS_EXTERN_CONST
+
+wasm_extern_t *
+wasm_extern_new_empty(wasm_store_t *store, wasm_externkind_t extern_kind)
+{
+    if (extern_kind == WASM_EXTERN_FUNC)
+        return wasm_func_as_extern(wasm_func_new_empty(store));
```
wenyongh, line 5206: Should we check the return value of `wasm_func_new_empty` since it may be NULL?

core/iwasm/common/wasm_c_api.c
```diff
@@ -5202,3 +5198,16 @@ BASIC_FOUR_LIST(WASM_EXTERN_AS_OTHER_CONST)
...
+
+    if (extern_kind == WASM_EXTERN_GLOBAL)
+        return wasm_global_as_extern(wasm_global_new_empty(store));
```
wenyongh, line 5209: Ditto

wenyongh (commented):
LGTM

core/iwasm/common/wasm_c_api.c
lum1n0us: `wasm_func_as_extern` will return NULL if the input is NULL.

https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/common/wasm_c_api.c#L5178-L5181

core/iwasm/common/wasm_c_api.c
lum1n0us: same as above

core/iwasm/common/wasm_c_api.c
wenyongh: Got it

wenyongh:
> > Maybe we can provide both wasm_runtime_is_import_func_linked and wasm_is_importtype_linked, the former is declared in wasm_export.h and the latter is declared in wasm_c_api.h.
> 
> do you imply we might introduce instantiate-time linking for wasm_export.h api? (otherwise this functionality doesn't seem necessary there.)

@yamt @nchamzn Liang has basically implemented the feature with samples and document provided, could you help review the PR? If there is no issue left, we may merge the PR.

The `wasm_runtime_is_import_func_linked` and `wasm_is_importtype_linked` are added, and another API `wasm_extern_new_empty` is also implemented to create an empty placeholder if the linking of that function/global is to be ignored. The sample code provided by @nchamzn may be changed like:

```C
static wasm_instance_t *
wasm_instance_create(wasm_store_t *store, wasm_module_t *module) {
    wasm_importtype_vec_t import_types = { 0 };
    wasm_module_imports(module, &import_types);

    // should be dynamic
    wasm_extern_t *externs[4] = { 0 };

    for (unsigned i = 0; i < import_types.num_elems; i++) {
        wasm_importtype_t *import_type = &(import_types.data[i]);

        if (wasm_is_importtype_linked(import_type)) {
            // push empty placeholder to externs
            externs[i] = wasm_extern_new_empty(store, wasm_externtype_kind(wasm_importtype_type(importtype)));
        } else if (is_function_from_a_plugin(module_name_str, field_name_str)) {
            // Set up import from host function and add to externs
        } else {
            // warn/err
        }
    }

    wasm_extern_vec_t imports = WASM_ARRAY_VEC(externs);
    wasm_instance_t *instance =
        wasm_instance_new(store, module, &imports, NULL);
}
```

nchamzn:
Thanks @wenyongh and @lum1n0us . As discussed, this should address my issues. To test this is going to require a bit of refactoring since I'm currently using the register_natives apis, but I will have a look at this and report back if I have some problems.

PR #1905, opened 2023-01-20
Document how to use WASI threads in AOT mode

eloparco:
Describe how to use WASI threads in AOT mode (aux stack must be disabled).

### DIFF
--- a/samples/wasi-threads/README.md
+++ b/samples/wasi-threads/README.md
@@ -10,7 +10,7 @@ make \
     THREAD_MODEL=posix
 ```
 
-Build and run the samples
+## Build and run the samples
 
 ```shell
 $ mkdir build
@@ -21,4 +21,11 @@ $ make
 $ ./iwasm wasm-apps/no_pthread.wasm
 ...
 $ ./iwasm wasm-apps/exception_propagation.wasm
+```
+
+## Run samples in AOT mode
+```shell
+$ ../../../wamr-compiler/build/wamrc \
+    -o wasm-apps/no_pthread.aot wasm-apps/no_pthread.wasm
+$ ./iwasm wasm-apps/no_pthread.aot
 ```
--- a/core/iwasm/interpreter/wasm_interp_classic.c
+++ b/core/iwasm/interpreter/wasm_interp_classic.c
@@ -1783,18 +1783,14 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                 global = globals + global_idx;
                 global_addr = get_global_addr(global_data, global);
                 aux_stack_top = *(uint32 *)(frame_sp - 1);
-                if (wasm_exec_env_is_aux_stack_managed_by_runtime(exec_env)) {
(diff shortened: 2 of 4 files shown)


eloparco:
Following the discussion [here](https://github.com/bytecodealliance/wasm-micro-runtime/pull/1867#discussion_r1070268062).

samples/wasi-threads/README.md
```diff
@@ -22,3 +22,11 @@ $ ./iwasm wasm-apps/no_pthread.wasm
 ...
 $ ./iwasm wasm-apps/exception_propagation.wasm
 ```
+
+Run samples in AOT mode (aux stack must be disabled)
+```shell
+$ ../../../wamr-compiler/build/wamrc \
+    --disable-aux-stack-check \
```
loganek, line 29: I wonder if that should be implied when `--enable-multi-thread` is set?

eloparco: That's the thing, it would disable the aux stack not only for WASI threads but also for pthreads. If that's what we want, I'll modify it to use `--enable-multi-thread` and have the aux stack disabled automatically when `--enable-multi-thread` is used. What do you think?

wamr-compiler/main.c
```diff
@@ -231,6 +231,7 @@ main(int argc, char *argv[])
             option.enable_bulk_memory = true;
             option.enable_thread_mgr = true;
             option.enable_ref_types = false;
+            option.enable_aux_stack_check = false;
```
wenyongh, line 234: Should not disable aux stack check for multi-thread, it is enabled for lib-pthread mode, which divides aux stack area into many regions and assign each region to a thread, the bottom and boundary of a region are pre-known and are set to thread exec_env's aux_stack_bottom and aux_stack_boundary fields.

The wasi-threads mode allocates the aux stack from the libc heap of wasm app, the bottom and boundary of the the aux stack are also known when the aux stack is allocated (seems bottom = addr_allocated + size, boundary = addr_allocated), I think we should set them to exec_env's aux_stack_bottom and aux_stack_boundary, but not change code here.

loganek: For wasi-threads the aux stack boundaries are not known for the runtime though. There were some discussions here https://github.com/WebAssembly/wasi-threads/issues/12 and here https://github.com/WebAssembly/wasi-libc/pull/380; there's no final conclusion although I'm leaning towards having this part of C ABI and not runtime ABI, so all the stack check work will most likely be done in the user space code.

wenyongh: If so, could you set `exec_env->aux_stack_bottom` to UINT32_MAX and `exec_env->aux_stack_boundary` to 0? So the aux stack boundary checks will always succeed and no need to change code here.

eloparco: Right, I tried it and it works with AOT.
I kept `wasm_exec_env_is_aux_stack_managed_by_runtime()` to use it in `free_aux_stack()`.

PR #1911, opened 2023-01-23
fix: fixed wasi-libc build steps in readme

hritikgupta:


samples/wasi-threads/README.md
```diff
@@ -1,6 +1,6 @@
 # "WASI threads" sample introduction
 
-Currently, since the `wasi-sdk` does not have thread support in the latest release, make sure to have [wasi-libc](https://github.com/WebAssembly/wasi-libc) installed. Build it with threads enabled, e.g.
+Currently, since the `wasi-sdk` does not have thread support in the latest release, make sure to have [wasi-libc](https://github.com/WebAssembly/wasi-libc) installed. After cloning, build the branch `wasi-sdk-17` with threads enabled, e.g.
```
loganek, line 3: Why do we have to build `wasi-sdk-17` branch specifically? Shouldn't the `main` branch already contain the necessary changes?

hritikgupta:
```
clang-15: warning: argument unused during compilation: '-arch arm64' [-Wunused-command-line-argument]
wasm-ld: error: cannot open crt1.o: No such file or directory
wasm-ld: error: unable to find library -lc
clang-15: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [wasm-apps/no_pthread.wasm] Error 1
make[1]: *** [wasm-apps/CMakeFiles/no_pthread.wasm.dir/all] Error 2
make: *** [all] Error 2
```
was running into the above issue when using the main branch.

loganek: yes, we'll have to update the examples as well, see: https://github.com/WebAssembly/wasi-libc/commit/4362b1885fd369e042a7c0ecd8df3b6cd47fb4e8

eloparco: The annoying thing of specifying the version here is that if we change it in the CI we need to remember to update it here as well. Maybe we can add a comment in the github action.

PR #1907, opened 2023-01-23
Rename thread_spawn import

yamt:
Following the wit-defined ABI:
https://github.com/WebAssembly/wasi-threads/pull/26

cf. https://github.com/WebAssembly/wasi-libc/pull/387

PR #1912, opened 2023-01-24
Update compiler target for wasi-threads examples

loganek:
The target was updated to wasm32-wasi-threads in wasi-libc (and most likely will stay like that for a while) https://github.com/WebAssembly/wasi-libc/pull/381


yamt:
see https://github.com/bytecodealliance/wasm-micro-runtime/pull/1907

PR #1913, opened 2023-01-24
Remove unused variable

pbeza:


loganek (approved):
LGTM. I wonder though if we can enable compiler flags on CI to fail the build for this kind of issues?

pbeza:
It's a good idea, but I don't know your build system so I can't help without investing a little more time into it. Actually the above problem was detected in my project by the compiler where I put `-Wall` (along with some other linters).

PR #1916, opened 2023-01-26
SGX IPFS: fix a segfault and support seeking beyond the end of files while using SEEK_CUR/SEEK_END

JamesMenetrey:
Dear WAMR developers,

This PR is about a fix in the `seek` operation offered by WASI and backed by SGX IPFS:

 - I realized that `sgx_fwrite` require a buffer of _count_ items when specifying the parameter `count` of said function. The current behaviour would read the only set byte and then overflow by reading what is in the stack, which may at worst dump the stack into files as padding and at best, throws a segfault.
 - The `seek` operation only supported `SEEK_SET` when seeking beyond the end of the file. This PR supports all the remaining enumeration values (`SEEK_CUR`/`SEEK_END`).

Note: for the first point, I have chosen to allocate a large buffer of zeroes (32 * 1024). I have selected that value to be big enough to avoid iterating too much the `do .. while` loop when seeking far beyond the end of the file while preserving some spaces in the data segment. Feel free to tell me if this size is not ok for you.

Cheers

### DIFF
--- a/core/shared/platform/linux-sgx/sgx_ipfs.c
+++ b/core/shared/platform/linux-sgx/sgx_ipfs.c
@@ -90,7 +90,7 @@ ipfs_write_zeroes(void *sgx_file, size_t len)
 {
     int min_count;
 
-    do {
+    while (len > 0) {
         min_count = len < ZEROES_PADDING_LENGTH ? len : ZEROES_PADDING_LENGTH;
 
         if (sgx_fwrite(zeroes_padding, 1, min_count, sgx_file) == 0) {
@@ -99,7 +99,7 @@ ipfs_write_zeroes(void *sgx_file, size_t len)
         }
 
         len -= min_count;
-    } while (len > 0);
+    }
 
     return 0;
 }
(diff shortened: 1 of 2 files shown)


core/shared/platform/linux-sgx/sgx_ipfs.c
```diff
@@ -78,6 +83,27 @@ ipfs_file_destroy(void *sgx_file)
     sgx_fclose(sgx_file);
 }
 
+// Writes a given number of zeroes in file at the current offset.
+// The return value is zero if successful; otherwise non-zero.
+static int
+ipfs_write_zeroes(void *sgx_file, size_t len)
+{
+    int min_count;
```
wenyongh, line 91: Shall we check `len == 0` firstly, or `sgx_fwrite(zeroes_padding, 1, min_count, sgx_file)` will be called with `min_count == 0`?
```C
if (len == 0)
    return 0;
```

JamesMenetrey: Hey @wenyongh, thanks for the code review.

I preferred to convert the `do.. while` loop into a `while` loop, which has the same predicate, rather than adding an `if` statement. This should simplify the code and verify that the value of the argument `len` is greater than 0.

Cheers

wenyongh: Good, thanks!

PR #1878, opened 2023-01-10
Upgrade toolkits

wenyongh:
Upgrade the version of related toolkits:
- upgrade llvm to 15.0
- upgrade wasi-sdk to 19.0
- upgrade emsdk to 3.1.28
- upgrade wabt to 1.0.31
- upgrade binaryen to 111

And upgrade the CI scripts, sample workload build scripts, Dockerfiles, and documents.

PR #1917, opened 2023-01-27
Make memory profiling show native stack usage

yamt:
open questions:
* is this worth to have at all?
* ~~should wamrc have a dedicated option? eg. `--enable-memory-profiling`~~


core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1399,6 +1399,22 @@ wasm_runtime_dump_mem_consumption(WASMExecEnv *exec_env)
...
+     * Unlike the aux stack above, we report the amount unused
+     * because we don't know the stack "bottom".
+     *
+     * Note that this is just about what the runtime itself observed.
+     * It doesn't cover host func implementations, signal handlers, etc.
+     */
+    if (exec_env->native_stack_max_used != (void *)UINTPTR_MAX)
+        os_printf("Native stack left: %zd\n",
+                  exec_env->native_stack_max_used
+                      - exec_env->native_stack_boundary);
```
wenyongh, line 1414: We can get the stack bottom for some platforms, e.g. posix platforms, refer to `pthread_attr_getstack` and https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/shared/platform/common/posix/posix_thread.c#L380

Maybe we can print the actual used size for these platforms.

yamt: we can. but i'm not sure if it's worth inventing another `os_` interface only for this purpose.

core/iwasm/common/wasm_exec_env.h
```diff
@@ -165,6 +171,19 @@ typedef struct WASMExecEnv {
...
+#define RECORD_STACK_USAGE(e, p)                \
+    do {                                        \
+        if ((e)->native_stack_max_used > (p)) { \
+            (e)->native_stack_max_used = (p);   \
+        }                                       \
+    } while (0)
+#else
+#define RECORD_STACK_USAGE(e, p) \
+    do {                         \
+    } while (0)
```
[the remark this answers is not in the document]
yamt: why?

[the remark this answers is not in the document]
wenyongh: The `do { } while (0)` also works, using `(void)0` just because it is more concise and we usually use that coding style.

core/iwasm/common/wasm_runtime_common.c
wenyongh: I am not sure yet, maybe we can keep your implementation, and discuss it later.

core/iwasm/common/wasm_exec_env.h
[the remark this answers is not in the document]
yamt: ok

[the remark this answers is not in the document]
yamt: done

core/iwasm/common/wasm_exec_env.c
```diff
@@ -211,6 +211,7 @@ wasm_exec_env_set_thread_info(WASMExecEnv *exec_env)
     exec_env->handle = os_self_thread();
     exec_env->native_stack_boundary =
         stack_boundary ? stack_boundary + WASM_STACK_GUARD_SIZE : NULL;
+    exec_env->native_stack_top_min = (void *)UINTPTR_MAX;
```
wenyongh, line 214: Had better add `#if WASM_ENABLE_MEMORY_PROFILING != 0`?

build-scripts/build_llvm.py
```diff
@@ -45,6 +45,7 @@ def build_llvm(llvm_dir, platform, backends, projects):
         "-DLLVM_ENABLE_IDE:BOOL=OFF",
         "-DLLVM_ENABLE_TERMINFO:BOOL=OFF",
         "-DLLVM_ENABLE_ZLIB:BOOL=OFF",
+        "-DLLVM_ENABLE_LIBEDIT:BOOL=OFF",
```
wenyongh, line 48: Had better not change this file currently, the line is to be added in https://github.com/bytecodealliance/wasm-micro-runtime/pull/1878

core/iwasm/common/wasm_exec_env.c
yamt: as the wamrc part is separately controlled with a cli option, it's safer to initialize unconditionally here.

build-scripts/build_llvm.py
yamt: oops, this was not intentional. i will revert.

yamt: reverted

xujuntwt95329 (commented):
LGTM

Issue #1910, opened 2023-01-23
proc_exit and traps do not stop thread executing blocking instructions


eloparco:
With #1889, `proc_exit` and traps are properly propagated to all the threads in the process. Once a thread traps or calls `proc_exit`, all the other threads are stopped.

This is not true if the thread is executing a blocking operation (e.g. `sleep`, `atomic.wait`). In that case, the blocking operation is not interrupted by the exception propagation process. 

Partial discussion here: https://github.com/bytecodealliance/wasm-micro-runtime/pull/1869#discussion_r1066524792

eloparco:
@wenyongh @xujuntwt95329 @yamt after the initial discussion in https://github.com/bytecodealliance/wasm-micro-runtime/pull/1869#discussion_r1066524792, any suggestion on the possible implementation of such a mechanism?
Initially I was thinking of using `sigsetjmp/siglongjmp` (to resume from [here](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_interp_classic.c#L4245)), but then I would need to keep track of `sigjmp_buf`s for each thread and also I don't know how portable it is for other non-linux platforms.
Otherwise, I'm not sure on what we can do in the signal handlers to stop the threads gracefully (ideally we'd like move to the next instruction after `sleep/atomic_wait` so that the exception flag can be read to stop the execution).

loganek:
I guess another approach would be to make all operations non-blocking in the host, but I'm not sure how much effort would that require (e.g. for WASI calls).

eloparco:
Two possible approaches:
- make all operations non-blocking in the host: it requires effort to modify blocking operations (possibly work to be replicated for each mode i.e. classic interpreter, AOT, ...); it wouldn't work with user-defined functions
- through signal handling: when propagating the exception to other threads, a signal is sent to those threads to stop their execution and return to the runtime (somewhere around [here](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_interp_classic.c#L4245)); 
it would require a global map to track `<thread_id, sigjmp_buf>` and the usage of `sigsetjmp/siglongjmp`.
Disadvantage: possible portability problem because of `sigsetjmp/siglongjmp`

yamt:
> it wouldn't work with user-defined functions

in either approaches, i guess user-defined functions need to be adapted anyway.
eg. resource cleanup on termination.

[...]

PR #1309, opened 2022-07-20
Implement interpreter hw bound check

wenyongh:
Implement boundary check with hardware trap for interpreter on
64-bit platforms:
- To improve the performance of interpreter and Fast JIT
- To prepare for multi-tier compilation for the feature

Linux/MacOS/Windows 64-bit are enabled.

PR #1921, opened 2023-01-29
Interrupt blocking instructions

eloparco:
Following the discussion in #1910, blocking operations (e.g. atomic wait, sleep, user-defined blocking functions) should be interruptible by exceptions (e.g. trap, proc_exit).

To achieve that, before calling a wasm function, the stack context/environment is saved (using `sigsetjmp()` in posix). When an exception is encountered in one thread, a signal is sent to all the other threads, interrupting their execution and passing the control to the registered signal handler. The signal handler then restores (using `siglongjmp()` in posix) the most recent stack context saved for the threads.

### DIFF
--- a/core/iwasm/interpreter/wasm_interp_classic.c
+++ b/core/iwasm/interpreter/wasm_interp_classic.c
@@ -4183,9 +4183,8 @@ wasm_interp_call_wasm(WASMModuleInstance *module_inst, WASMExecEnv *exec_env,
 
     wasm_exec_env_set_cur_frame(exec_env, frame);
 
-    sigjmp_buf *context = os_save_context(exec_env->handle);
-    if (!os_is_exception(exec_env->handle, context)) {
-
+    sigjmp_buf *context = os_create_stack_context();
+    if (!os_is_returning_from_signal(exec_env->handle, context)) {
         if (function->is_import_func) {
 #if WASM_ENABLE_MULTI_MODULE != 0
             if (function->import_module_inst) {
@@ -4263,10 +4262,7 @@ wasm_interp_call_wasm(WASMModuleInstance *module_inst, WASMExecEnv *exec_env,
 #endif
         }
     }
-
-    korp_tid self = pthread_self();
-    printf("exit context=%p %p\n", context, (void *)self);
-    os_rm_context(exec_env->handle, context);
+    os_remove_stack_context(exec_env->handle, context);
 
     /* Output the return value to the caller */
     if (!wasm_get_exception(module_inst)) {
--- a/core/iwasm/interpreter/wasm_interp_classic.c
+++ b/core/iwasm/interpreter/wasm_interp_classic.c
@@ -99,18 +99,6 @@ typedef float64 CellType_F64;
 #define CHECK_WRITE_WATCHPOINT(addr, offset) (void)0
 #endif
 
-static uint32
-thread_handle_hash(void *handle)
-{
-    return (uint32)(uintptr_t)handle;
-}
-
-static bool
-thread_handle_equal(void *h1, void *h2)
-{
-    return (uint32)(uintptr_t)h1 == (uint32)(uintptr_t)h2 ? true : false;
-}
-
 static inline uint32
 rotl32(uint32 n, uint32 c)
 {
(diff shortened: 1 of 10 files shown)


eloparco:
MIssing:
- [ ] test in different modes (only classic interpreter for now)
- [ ] add support for other platforms (Windows at least? add dummy/empty functions for the definitions in `platform_api_extension.h` for the other platforms)
- [ ] try with low level tests in https://github.com/WebAssembly/wasi-threads/pull/19

core/shared/platform/common/posix/posix_thread.c
```diff
@@ -47,6 +52,152 @@ os_thread_wrapper(void *arg)
     return NULL;
 }
 
+static void
+stack_context_sig_handler(int sig)
+{
+    assert(sig == SIGUSR1);
+    korp_tid self = pthread_self();
+
+    os_mutex_lock(&stack_context_handler.lock);
```
yamt, line 61: i don't think pthread mutex is async signal safe.
if you mean to protect from a signal handler, it's more appropriate to mask the signal.

core/shared/platform/common/posix/posix_thread.c
```diff
@@ -47,6 +54,166 @@ os_thread_wrapper(void *arg)
...
+    if (os_mutex_destroy(&stack_context_handler.lock) != BHT_OK)
+        LOG_ERROR("failed to destroy stack contexts");
+}
+
+void
+os_remove_stack_context(korp_tid handle, sigjmp_buf *context)
+{
+    korp_tid self = pthread_self();
+
+    stack_ctx_block_sig_and_lock();
```
yamt, line 157: why don't you make it thread-local?

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -4183,79 +4183,86 @@ wasm_interp_call_wasm(WASMModuleInstance *module_inst, WASMExecEnv *exec_env,
 
     wasm_exec_env_set_cur_frame(exec_env, frame);
 
-    if (function->is_import_func) {
+    sigjmp_buf *context = os_create_stack_context();
+    if (!os_is_returning_from_signal(exec_env->handle, context)) {
```
yamt, line 4187: how about resource cleanup i mentioned in https://github.com/bytecodealliance/wasm-micro-runtime/issues/1910#issuecomment-1408193240 ?

core/shared/platform/common/posix/posix_thread.c
eloparco: Good point, otherwise it would be shared between threads (and would then require access protection).

Actually, I wanted to avoid globals and return the `oldmask` from `stack_ctx_block_sig_and_lock()`, but returning anything from that function was corrupting the stack when used inside `os_is_returning_from_signal()`, I'm still not sure I understand why

core/shared/platform/common/posix/posix_thread.c
```diff
@@ -47,6 +54,164 @@ os_thread_wrapper(void *arg)
...
+    BH_FREE(context);
+
+    stack_ctx_unlock_and_unblock_sig();
+}
+
+bool
+os_is_returning_from_signal(korp_tid handle, sigjmp_buf *context)
+{
+    if (sigsetjmp(*context, 1))
+        return true;
```
yamt, line 171: this looks unsafe to me.
when we return here with a longjmp, the call frame of os_is_returning_from_signal is not valid anymore.

to make this work, os_is_returning_from_signal should always be inlined.
maybe it's more straightforward to use a macro instead.

eloparco: thanks, that might explain some problems that I was seeing when trying to use it with AOT. In any case, I realized that I missed this  PR #1309 that is basically doing something very similar to what I want to do. The approach is better because we don't need to keep a map <thread_id, stack contexts> if we use tls. I'll try to apply a similar approach.

wenyongh: @eloparco Yes, in the memory access boundary check with hardware trap feature, both interpreter and AOT use the setjmp/longjmp mechanism in posix like platforms. The local variable jmpbuf is pushed into exec_env's jmpbuf stack before setjmp, runtime doesn't allocate memory for it.
I think that the principle of this PR is similar to the hw bound check feature: firstly register a signal handler, then setjmp before call, and when running wasm function, if there is a signal raised, the signal handler is trigger and the handler longjmp into the place of setjmp.

I think there may be two ways to implement this PR: one is to extend the hw bound check feature, e.g. reuse the current signal handler, the current setjmp and longjmp process, but add another macro to control them respectively (for example, OS_ENABLE_XXX and OS_ENABLE_HW_BOUND_CHECK). The other is to register new signal handler, add new setjmp/longjmp process, but I suggest not to add setjmp in wasm_interp_class.c, had better add in wasm_runtime.c, before calling wasm function:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_runtime.c#L2163
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_runtime.c#LL2459C3-L2459C3

And note that setjmp is only supported by posix platforms, there should be macro to control the feature, and had better use os_setjmp/os_longjmp.

wenyongh: And could you submit this PR to main branch since the main branch also has the issue?

eloparco: > https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_runtime.c#L2163

yes that works, for aot I'm not sure what to wrap with `os_setjmp` since there are 2 calls, I need to check https://github.com/bytecodealliance/wasm-micro-runtime/blob/fe3347d5d2d92c3a78d77c55877692bc07ac8888/core/iwasm/aot/aot_runtime.c#L1411
https://github.com/bytecodealliance/wasm-micro-runtime/blob/fe3347d5d2d92c3a78d77c55877692bc07ac8888/core/iwasm/aot/aot_runtime.c#L1470

eloparco: > And could you submit this PR to main branch since the main branch also has the issue?

I'll try to have it working here first since I'm testing it with wasi threads, then I'll create a PR in main to merge it there

eloparco: I'm closing this PR in favour of #1930, where I rely on what's already implemented for hardware bound checks.

PR #1924, opened 2023-01-31
Fix data race when terminating or waiting for a thread.

loganek:
There is a possibility that while iterating through a list of threads, the threads finish by themselves, making a list of exec_env invalid.

### DIFF
--- a/core/iwasm/libraries/thread-mgr/thread_manager.c
+++ b/core/iwasm/libraries/thread-mgr/thread_manager.c
@@ -76,6 +76,66 @@ traverse_list(bh_list *l, list_visitor visitor, void *user_data)
     }
 }
 
+/* Assumes cluster->lock is locked */
+static bool
+safe_traverse_exec_env_list(WASMCluster *cluster, list_visitor visitor,
+                            void *user_data)
+{
+    Vector proc_nodes;
+    void *node, *next;
+    bool ret = true;
+
+    if (!bh_vector_init(&proc_nodes, cluster->exec_env_list.len, sizeof(void *),
+                        false)) {
+        ret = false;
+        goto final;
+    }
+
+    node = bh_list_first_elem(&cluster->exec_env_list);
+    cluster->exec_env_list_element_removed = false;
+
+    while (node) {
+        bool already_processed = false;
+        void *proc_node;
+        for (size_t i = 0; i < bh_vector_size(&proc_nodes); i++) {
+            if (!bh_vector_get(&proc_nodes, i, &proc_node)) {
+                ret = false;
+                goto final;
+            }
+            if (proc_node == node) {
+                already_processed = true;
+                break;
+            }
+        }
+        if (already_processed) {
+            node = bh_list_elem_next(node);
+            continue;
+        }
+
+        next = bh_list_elem_next(node);
+        os_mutex_unlock(&cluster->lock);
+        visitor(node, user_data);
+        os_mutex_lock(&cluster->lock);
+        if (!bh_vector_append(&proc_nodes, &node)) {
+            ret = false;
+            goto final;
+        }
+
+        if (cluster->exec_env_list_element_removed) {
+            cluster->exec_env_list_element_removed = false;
+            node = bh_list_first_elem(&cluster->exec_env_list);
+        }
+        else {
+            node = next;
+        }
+    }
+
+final:
+    bh_vector_destroy(&proc_nodes);
+
+    return ret;
+}
+
 /* The caller must lock cluster->lock */
 static bool
 allocate_aux_stack(WASMCluster *cluster, uint32 *start, uint32 *size)
@@ -299,7 +359,7 @@ wasm_cluster_del_exec_env(WASMCluster *cluster, WASMExecEnv *exec_env)
         os_mutex_unlock(&cluster->debug_inst->wait_lock);
     }
 #endif
-
+    cluster->exec_env_list_element_removed = true;
     if (bh_list_remove(&cluster->exec_env_list, exec_env) != 0)
         ret = false;
 
@@ -724,16 +784,22 @@ wasm_cluster_join_thread(WASMExecEnv *exec_env, void **ret_val)
     korp_tid handle;
 
     os_mutex_lock(&cluster_list_lock);
+    os_mutex_lock(&exec_env->cluster->lock);
+
     if (!clusters_have_exec_env(exec_env) || exec_env->thread_is_detached) {
         /* Invalid thread, thread has exited or thread has been detached */
         if (ret_val)
             *ret_val = NULL;
+        os_mutex_unlock(&exec_env->cluster->lock);
         os_mutex_unlock(&cluster_list_lock);
         return 0;
     }
     exec_env->wait_count++;
     handle = exec_env->handle;
+
+    os_mutex_unlock(&exec_env->cluster->lock);
     os_mutex_unlock(&cluster_list_lock);
+
     return os_thread_join(handle, ret_val);
 }
 
@@ -816,14 +882,21 @@ int32
 wasm_cluster_cancel_thread(WASMExecEnv *exec_env)
 {
     os_mutex_lock(&cluster_list_lock);
+    os_mutex_lock(&exec_env->cluster->lock);
+
+    if (!exec_env->cluster) {
+        goto final;
+    }
     if (!clusters_have_exec_env(exec_env)) {
         /* Invalid thread or the thread has exited */
-        os_mutex_unlock(&cluster_list_lock);
-        return 0;
+        goto final;
     }
-    os_mutex_unlock(&cluster_list_lock);
 
     set_thread_cancel_flags(exec_env);
+
+final:
+    os_mutex_unlock(&exec_env->cluster->lock);
+    os_mutex_unlock(&cluster_list_lock);
 
     return 0;
 }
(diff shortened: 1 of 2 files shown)


loganek:
@xujuntwt95329 I'd appreciate your feedback as I think it's related to your recent changes in thread manager: https://github.com/bytecodealliance/wasm-micro-runtime/pull/1889

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -76,6 +76,63 @@ traverse_list(bh_list *l, list_visitor visitor, void *user_data)
...
+        if (already_processed) {
+            node = bh_list_elem_next(node);
+            continue;
+        }
+
+        next = bh_list_elem_next(node);
+        os_mutex_unlock(&cluster->lock);
+        visitor(node, user_data);
+        os_mutex_lock(&cluster->lock);
+        bh_vector_append(&proc_nodes, &node);
```
xujuntwt95329, line 119: This API may fail, should check return value

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -76,6 +76,66 @@ traverse_list(bh_list *l, list_visitor visitor, void *user_data)
...
+            }
+        }
+        if (already_processed) {
+            node = bh_list_elem_next(node);
+            continue;
+        }
+
+        next = bh_list_elem_next(node);
+        os_mutex_unlock(&cluster->lock);
+        visitor(node, user_data);
```
yamt, line 117: what keeps "node" from being freed while we release cluster->lock here?

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -76,6 +76,66 @@ traverse_list(bh_list *l, list_visitor visitor, void *user_data)
...
+            }
+            if (proc_node == node) {
+                already_processed = true;
+                break;
+            }
+        }
+        if (already_processed) {
+            node = bh_list_elem_next(node);
+            continue;
+        }
```
yamt, line 113: i guess for many (all?) of usage it's ok to visit nodes multiple times and thus it isn't worth to have this O(x^2) check.

loganek: Yeah, most likely that won't be an issue; having said that, I don't think `x` is going to be large enough to worry about that (so I'd rather keep the implementation slow but always correct, than making it a bit faster but risk somebody will use it incorrectly). If you think that will have a significant performance impact (because `x` can be large), I can update the code.

xujuntwt95329 (commented):
LGTM

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -76,6 +76,66 @@ traverse_list(bh_list *l, list_visitor visitor, void *user_data)
...
+        goto final;
+    }
+
+    node = bh_list_first_elem(&cluster->exec_env_list);
+    cluster->exec_env_list_element_removed = false;
+
+    while (node) {
+        bool already_processed = false;
+        void *proc_node;
+        for (size_t i = 0; i < bh_vector_size(&proc_nodes); i++) {
```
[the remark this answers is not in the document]
loganek: Ok, I'll update the code. It'd be good to have some sort of CI check to enforce that though (I guess compile with `-std=c89` would do, but is the whole codebase c89 compliant?)

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -76,6 +76,66 @@ traverse_list(bh_list *l, list_visitor visitor, void *user_data)
...
+        os_mutex_unlock(&cluster->lock);
+        visitor(node, user_data);
+        os_mutex_lock(&cluster->lock);
+        if (!bh_vector_append(&proc_nodes, &node)) {
+            ret = false;
+            goto final;
+        }
+
+        if (cluster->exec_env_list_element_removed) {
+            cluster->exec_env_list_element_removed = false;
```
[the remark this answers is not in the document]
loganek: Yeah, good point. I can think of two solutions for that:
1. have a new lock that guards the whole function (so no two functions can execute at the same time)
2. Update the function so it always assumes the element was removed

I feel the 1) is safer as the 2) might cause re-running the same visitor twice which might not be desired (it's used in scenarios where elements are removed from the list), although it adds one more field to the cluster object (maybe that's not a big issue though?). @wenyongh  what are your thoughts?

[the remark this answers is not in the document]
wenyongh: Not sure whether 1) will cause dead lock or not? For example, thread A locks the lock for this function and then unlock cluster->lock, after handling the node, it is waiting the cluster->lock, and thread B locks cluster->lock and is waiting the lock for this function?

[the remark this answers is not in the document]
loganek: I don't think that situation can happen in the current setup, but I see in the future incorrectly using the function might cause some problems. Would you say then 2) would be better approach? I think we can still keep an array of already processed nodes to avoid re-processing elements. That will be less efficient for large number of exec_env's, but I don't think the number will ever grow significantly?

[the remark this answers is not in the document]
wenyongh: What we are afraid is that the node visited is removed by other threads so we cannot use its next again, right?
If yes, after visiting and locking cluster->lock again, can we search the whole list to check whether the node is in the list? If yes, node = node->next, else node = bh_list_first_elem(..), does this way work?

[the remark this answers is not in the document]
loganek: Yeah, so what you're suggesting is a slightly improved (more performant in some cases) version of 2). I think that could work. The only concern is when two threads call the traversal at the same time, although for current usecase it's not the problem because the traversal callback does another lock and make sure the node exists. I'll update the PR with the implementation.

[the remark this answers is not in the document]
loganek: @wenyongh I updated the code, I'd appreciate you  taking a look.

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -76,6 +76,58 @@ traverse_list(bh_list *l, list_visitor visitor, void *user_data)
...
+
+        os_mutex_unlock(&cluster->lock);
+        visitor(node, user_data);
+        os_mutex_lock(&cluster->lock);
+        if (!bh_vector_append(&proc_nodes, &node)) {
+            ret = false;
+            goto final;
+        }
+
+        node = bh_list_first_elem(&cluster->exec_env_list);
```
wenyongh, line 122: This will make it traversing from the first node again, nodes from the list head to current will be tested again whether they are in the vector, it is nearly O(n^2), and besides the top `while (node)` loop, the algorithm's complexity is nearly O(n^3). Not very efficient.

I think that we can check whether the most recent visited node is still in the exec_env list, if yes, we can get its next, and then continue to traverse, if not, traversing from the head of the list:
```C
static bool
safe_traverse_exec_env_list(WASMCluster *cluster, list_visitor visitor,
                            void *user_data)
{
    Vector proc_nodes;
    void *node;
    bool ret = false;

    if (!bh_vector_init(&proc_nodes, cluster->exec_env_list.len, sizeof(void *),
                        false)) {
        goto final;
    }

    node = bh_list_first_elem(&cluster->exec_env_list);

    while (node) {
        void *proc_node;
        bool found;
        int i;

        os_mutex_unlock(&cluster->lock);
        visitor(node, user_data);
        os_mutex_lock(&cluster->lock);

        if (!bh_vector_append(&proc_nodes, &node)) {
            goto final;
        }

        found = false;
        /* Find the most recent visited and still existing node */
        for (i = (int)bh_vector_size(&proc_nodes) - 1; i >= 0; i--) {
            if (!bh_vector_get(&proc_nodes, i, &proc_node)) {
                goto final;
            }
            node = bh_list_first_elem(&cluster->exec_env_list);
            while (node) {
                if (proc_node == node) {
                    found = true;
                    break;
                }
                node = bh_list_elem_next(node);
            }
            if (found)
                break;
            else {
                /* The node has been removed from cluster->exec_env_list */
                (void)bh_vector_remove(&proc_nodes, i, NULL);
            }
        }

        if (found)
            node = bh_list_elem_next(node);
        else
            node = bh_list_first_elem(&cluster->exec_env_list);
    }

    ret = true;
final:
    bh_vector_destroy(&proc_nodes);

    return ret;
}
```
Could you help check whether it works?

wenyongh: It is still not very efficient, seems to be O(n^2).
Wondering why we need to unlock and lock again when visiting the node? Will the visiting lock the cluster->lock again?

loganek: @wenyongh please note the critical operation in the loop is:
```
os_mutex_unlock(&cluster->lock);
visitor(node, user_data);
os_mutex_lock(&cluster->lock);
```
and that in either of the implementations is executed only N times. Also, I think in the current implementation we only go beyond for loop (and therefore, goes at the beginning of the list) if the node is not in the vector, so the actual complexity is quadratic. I think there's a room for improvement (not only in the iteration itself, but e.g. by using hash set instead of vector), but I don't think this will significantly improve performance of this function. 

> Wondering why we need to unlock and lock again when visiting the node? Will the visiting lock the cluster->lock again?

Yes, they lock the cluster's lock internally. Although I think we can improve that. My suggestion is to push the fix as it is (unless you see bugs in it) to not block potential customers (please note the problem is also with the main branch, so I'm updating the PR to point to main) and we can tune the performance of the function later if needed. What do you think?

wenyongh: OK, let's merge this PR first.

PR #1928, opened 2023-02-01
Enable gcc-4.8 compilation

loganek:
Once this PR is merged, I'll think of having a CI job for gcc4.8 too


loganek:
> why do you care that old gcc? (just a curious question)

Unfortunately some of the toolchains still use that version.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/numeric_limits.h
```diff
@@ -39,4 +53,21 @@
              : ULLONG_MAX, default           \
              : (void)0)
 
+#else
```
loganek, line 56: I actually wonder if there's any advantage of keeping the implementation with `_Generic` at all.. any opinions if I remove it?

wenyongh: There is only one place using NUMERIC_MAX, and NUMBERIC_MIN isn't used:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c#L2260-L2261
```C
    out->tv_sec =
        (time_t)in < NUMERIC_MAX(time_t) ? (time_t)in : NUMERIC_MAX(time_t);
```
I remember that time_t is defined as `long int`, we should be able to simplify the code.

yamt: please don't assume long int. we have BH_TIME_T_MAX for this purpose.

wenyongh (commented):
LGTM

PR #1929, opened 2023-02-01
fixed terminating stale threads on trap/proc_exit

hritikgupta:
This is to terminate suspended threads in case an atomic wait occurs with a huge or indefinite (-1) timeout, followed by a proc exit or trap.

### DIFF
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -2195,6 +2195,12 @@ wasm_set_exception(WASMModuleInstance *module_inst, const char *exception)
     if (exec_env) {
         wasm_cluster_spread_exception(exec_env, exception ? false : true);
     }
+#if WASM_ENABLE_SHARED_MEMORY
+    if (exception) {
+        notify_stale_threads_on_exception(
+            (WASMModuleInstanceCommon *)module_inst);
+    }
+#endif
 #else
     (void)exec_env;
 #endif
--- a/core/iwasm/common/wasm_shared_memory.c
+++ b/core/iwasm/common/wasm_shared_memory.c
@@ -30,6 +30,11 @@ typedef struct AtomicWaitNode {
     korp_cond wait_cond;
 } AtomicWaitNode;
 
+typedef struct AtomicWaitAddressArgs {
+    uint32 index;
+    void **addr;
+} AtomicWaitAddressArgs;
+
 /* Atomic wait map */
 static HashMap *wait_map;
 
@@ -85,6 +90,53 @@ search_module(WASMModuleCommon *module)
 
     os_mutex_unlock(&shared_memory_list_lock);
     return NULL;
+}
+
+static void
+wait_map_address_count_callback(void *key, void *value,
+                                void *p_total_elem_count)
+{
+    *(uint32 *)p_total_elem_count = *(uint32 *)p_total_elem_count + 1;
+}
+
+static void
+create_list_of_waiter_addresses(void *key, void *value, void *user_data)
+{
+    AtomicWaitAddressArgs *data = (AtomicWaitAddressArgs *)user_data;
+    data->addr[data->index++] = key;
+}
+
+void
+notify_stale_threads_on_exception(WASMModuleInstanceCommon *module_inst)
+{
+    AtomicWaitAddressArgs args = { 0 };
+    uint32 i = 0, total_elem_count = 0;
+
+    os_mutex_lock(&shared_memory_list_lock);
+
+    /* count number of addresses in wait_map */
+    bh_hash_map_traverse(wait_map, wait_map_address_count_callback,
+                         (void *)&total_elem_count);
+
+    /* allocate memory */
+    if (!(args.addr = wasm_runtime_malloc(sizeof(void *) * total_elem_count))) {
+        LOG_ERROR(
+            "failed to allocate memory for list of atomic wait addresses");
+        os_mutex_unlock(&shared_memory_list_lock);
+        return;
+    }
+
+    /* set values in list of addresses */
+    bh_hash_map_traverse(wait_map, create_list_of_waiter_addresses, &args);
+    os_mutex_unlock(&shared_memory_list_lock);
+
+    /* notify */
+    for (i = 0; i < args.index; i++) {
+        wasm_runtime_atomic_notify(module_inst, args.addr[i], UINT32_MAX);
+    }
+
+    /* free memory allocated to args data */
+    wasm_runtime_free(args.addr);
 }
 
 WASMSharedMemNode *
@@ -321,6 +373,10 @@ wasm_runtime_atomic_wait(WASMModuleInstanceCommon *module, void *address,
 
     bh_assert(module->module_type == Wasm_Module_Bytecode
               || module->module_type == Wasm_Module_AoT);
+
+    if (wasm_get_exception(module_inst)) {
+        return -1;
+    }
 
     /* Currently we have only one memory instance */
     if (!module_inst->memories[0]->is_shared) {
--- a/core/iwasm/common/wasm_shared_memory.h
+++ b/core/iwasm/common/wasm_shared_memory.h
@@ -37,6 +37,9 @@ wasm_shared_memory_init();
 void
 wasm_shared_memory_destroy();
 
+void
+notify_stale_threads_on_exception(WASMModuleInstanceCommon *module);
+
 WASMSharedMemNode *
 wasm_module_get_shared_memory(WASMModuleCommon *module);
 


core/iwasm/common/wasm_shared_memory.c
```diff
@@ -32,6 +32,7 @@ typedef struct AtomicWaitNode {
 
 /* Atomic wait map */
 static HashMap *wait_map;
+static HashMap *wait_map_without_lock;
```
loganek, line 35: Why does it have to be a global variable?

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -87,6 +88,41 @@ search_module(WASMModuleCommon *module)
...
+notify_stale_threads_on_exception(WASMModuleInstanceCommon *module_inst)
+{
+
+    wait_map_without_lock = bh_hash_map_create(
+        32, false, (HashFunc)wait_address_hash,
+        (KeyEqualFunc)wait_address_equal, NULL, destroy_wait_info);
+    /**
+     * create a copy of wait_map with no lock
+     * this allows invoking methods in the callback
+     * which require acquiring lock on the original map
```
loganek, line 116: I don't see there's any lock in the `notify_stale_threads` callback. Should we rather have a lock on the `shared_memory_list_lock` mutex at the beginning of the function and release it at the end? Otherwise the map could be modified in the other thread making the traversal corrupted.

hritikgupta: the `wasm_runtime_atomic_notify` step in callback calls `acquire_wait_info` which tries to acquire `shared_memory_list_lock`, so I guess encapsulating it with that will cause issues?

eloparco: `HashMap` doesn't duplicate the key/values pairs, so `wait_map_without_lock` will point to the same elements of `wait_map`

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -87,6 +87,42 @@ search_module(WASMModuleCommon *module)
...
+     */
+    bh_hash_map_traverse(wait_map, create_copy_of_wait_map,
+                         wait_map_without_lock);
+
+    // terminate stale threads
+    bh_hash_map_traverse(wait_map_without_lock, notify_stale_threads,
+                         module_inst);
+}
```
eloparco, line 124: we should probably call `bh_hash_map_destroy` on `wait_map_without_lock` at the end

core/iwasm/common/wasm_shared_memory.c
loganek: I think you at least want to lock while you copy hashmaps. Also, what happens if after making a copy a new waiter is added to the map?

hritikgupta: While copying the original map, the original map already acquires a lock as it traversal. as inserts happen in the copy map, every insert locks the copy map. 
What could be concerning here is that during the traversal of copy map, the original map could be updated. If I put a lock on the original map during traversal, the execution gets stuck because the callback steps in `notify_stale_threads` try to put a lock on the original map. This might lead to some entries (newly added ones) in the original map not catered to at that point of time, but would be eventually handled, right?

eloparco: either way I'm still not sure it would be safe since you're not cloning the map but copying the pointers iiuc

hritikgupta:
closing and reopening to reinitiate the build

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -345,6 +376,22 @@ wasm_runtime_atomic_wait(WASMModuleInstanceCommon *module, void *address,
 
     os_mutex_lock(&wait_info->wait_list_lock);
 
+    /* add waiter address to atomic wait list */
+    AtomicWaitAddressNode *wait_address_node;
```
loganek, line 380: Does the list contain a copy of the list of keys of `wait_map`? If so, why do we need to maintain that list instead of using `wait_map`'s keys?

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -87,6 +92,43 @@ search_module(WASMModuleCommon *module)
...
+    wait_address_node->addr = key;
+    bh_assert(bh_list_insert(user_data, wait_address_node) == BH_LIST_SUCCESS);
+}
+
+void
+notify_stale_threads_on_exception(WASMModuleInstanceCommon *module_inst)
+{
+    bh_list atomic_wait_address_list_head;
+    bh_list *const atomic_wait_address_list = &atomic_wait_address_list_head;
+    static korp_mutex atomic_wait_address_list_lock;
```
loganek, line 114: should the mutex be initialized?

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -87,6 +92,43 @@ search_module(WASMModuleCommon *module)
...
+        return;
+    }
+    wait_address_node->addr = key;
+    bh_assert(bh_list_insert(user_data, wait_address_node) == BH_LIST_SUCCESS);
+}
+
+void
+notify_stale_threads_on_exception(WASMModuleInstanceCommon *module_inst)
+{
+    bh_list atomic_wait_address_list_head;
```
loganek, line 112: does it have to be a linked list? I'd assume we know the exact size of the `wait_map`, so we can use array and only have a single allocation?

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -87,6 +92,43 @@ search_module(WASMModuleCommon *module)
...
+    bh_list *const atomic_wait_address_list = &atomic_wait_address_list_head;
+    static korp_mutex atomic_wait_address_list_lock;
+    AtomicWaitAddressNode *node;
+
+    bh_list_status ret = bh_list_init(atomic_wait_address_list);
+    bh_assert(ret == BH_LIST_SUCCESS);
+
+    os_mutex_lock(&atomic_wait_address_list_lock);
+    bh_hash_map_traverse(wait_map, create_list_of_waiter_addresses,
```
loganek, line 121: for this traversal, we should probably lock a `shared_memory_list_lock` otherwise the traversal might be invalidated half way through?

core/iwasm/common/wasm_shared_memory.c
loganek: I actually don't think we need this mutex at all. What's the risk of calling this function from two different threads? Is it ok if we call `notify` twice on the same address in this specific instance?

hritikgupta: I was also skeptical about this, just thought would keep it for the draft and get a second opinion; will remove it.

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -87,6 +93,36 @@ search_module(WASMModuleCommon *module)
...
+    AtomicWaitAddressArgs *data = (AtomicWaitAddressArgs *)user_data;
+    data->addr[(data->index++) % WAIT_MAP_SIZE] = key;
+}
+
+void
+notify_stale_threads_on_exception(WASMModuleInstanceCommon *module_inst)
+{
+    AtomicWaitAddressArgs *args = NULL;
+    if (!(args = wasm_runtime_malloc(sizeof(AtomicWaitAddressArgs)))) {
+        return;
```
loganek, line 108: Should we print an error here for debugging purposes?

hritikgupta: added

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -87,6 +93,37 @@ search_module(WASMModuleCommon *module)
...
+create_list_of_waiter_addresses(void *key, void *value, void *user_data)
+{
+    AtomicWaitAddressArgs *data = (AtomicWaitAddressArgs *)user_data;
+    data->addr[(data->index++) % WAIT_MAP_SIZE] = key;
+}
+
+void
+notify_stale_threads_on_exception(WASMModuleInstanceCommon *module_inst)
+{
+    AtomicWaitAddressArgs *args = NULL;
```
loganek, line 106: I'd suggest allocating `AtomicWaitAddressArgs` on the stack, but dynamically allocate `addr` field (currently it's fixed value - 32), so you don't have to guess the size of the array in advance.

loganek (changes requested):

core/iwasm/common/wasm_shared_memory.c
hritikgupta: done

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -5,6 +5,7 @@
 
 #include "bh_log.h"
 #include "wasm_shared_memory.h"
+#define WAIT_MAP_SIZE 32
```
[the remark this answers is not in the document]
hritikgupta: fixed

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -87,6 +92,53 @@ search_module(WASMModuleCommon *module)
...
+    }
+
+    memset(args, 0, sizeof(*args));
+    os_mutex_lock(&shared_memory_list_lock);
+    // create list of addresses
+    bh_hash_map_traverse(wait_map, create_list_of_waiter_addresses, args);
+    os_mutex_unlock(&shared_memory_list_lock);
+
+    // notify
+    for (uint32 i = 0; i < args->len; i++) {
```
hritikgupta, line 134: @loganek do you think `len`'s value can become inconsistent due to a race condn? do you suggest putting an if check before calling notify i.e. `if(args->addr[i])`?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -2190,6 +2190,13 @@ wasm_set_exception(WASMModuleInstance *module_inst, const char *exception)
     }
 
 #if WASM_ENABLE_THREAD_MGR != 0
+
+#if WASM_ENABLE_SHARED_MEMORY
+    if (exception) {
+        notify_stale_threads_on_exception(
+            (WASMModuleInstanceCommon *)module_inst);
```
wenyongh, line 2197: Is this PR to terminating the thread which is atomic waiting, when there is other thread trapping of proc_exiting? Is it duplicated with https://github.com/bytecodealliance/wasm-micro-runtime/pull/1902? Seems #1902 can trigger the signal handler of the atomic waiting thread when os_pthread_kill to it, and in the signal handler, the atomic waiting thread can longjmp into the place of setjmp before, and exit normally?

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -87,6 +92,53 @@ search_module(WASMModuleCommon *module)
     return NULL;
 }
 
+static void
+create_list_of_waiter_addresses(void *key, void *value, void *user_data)
+{
+    AtomicWaitAddressArgs *data = (AtomicWaitAddressArgs *)user_data;
+    if (data->len > 0) {
+        if (!(data->addr = wasm_runtime_realloc(data->addr,
```
wenyongh, line 100: When to free the memory? And why do nothing except LOG_ERROR when malloc/realloc memory failed?

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -87,6 +92,53 @@ search_module(WASMModuleCommon *module)
...
+{
+    AtomicWaitAddressArgs *args = NULL;
+    if (!(args = wasm_runtime_malloc(sizeof(AtomicWaitAddressArgs)))) {
+        LOG_ERROR("failed to create atomic wait address args node");
+        return;
+    }
+
+    memset(args, 0, sizeof(*args));
+    os_mutex_lock(&shared_memory_list_lock);
+    // create list of addresses
```
wenyongh, line 129: Use `/* .. */` to add comments

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -87,6 +92,53 @@ search_module(WASMModuleCommon *module)
...
+    if (data->len > 0) {
+        if (!(data->addr = wasm_runtime_realloc(data->addr,
+                                                data->len * sizeof(uint32)))) {
+            LOG_ERROR("failed to realloc atomic wait address list during "
+                      "wait_map traversal");
+            return;
+        }
+    }
+    else {
+        if (!(data->addr = wasm_runtime_malloc(sizeof(data->len) + 1))) {
```
wenyongh, line 108: sizeof(void *) * (data->len + 1)?

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -87,6 +92,53 @@ search_module(WASMModuleCommon *module)
...
+    AtomicWaitAddressArgs *args = NULL;
+    if (!(args = wasm_runtime_malloc(sizeof(AtomicWaitAddressArgs)))) {
+        LOG_ERROR("failed to create atomic wait address args node");
+        return;
+    }
+
+    memset(args, 0, sizeof(*args));
+    os_mutex_lock(&shared_memory_list_lock);
+    // create list of addresses
+    bh_hash_map_traverse(wait_map, create_list_of_waiter_addresses, args);
```
wenyongh, line 130: Why not traverse first time to get the total size of wait_map, use it to allocate memory for `AtomicWaitAddressArgs *args`  with size `offsetof(AtomicWaitAddressArgs, addr) + sizeof(void *) * total_size` , and then traverse again without allocating memory?

core/iwasm/common/wasm_shared_memory.c
wenyongh, line 134: Please declare `uint32 i` at the beginning of function

core/iwasm/common/wasm_runtime_common.c
wenyongh: Suppose the patch can notify the current atomic wait operation, but if here is another atomic wait operation followed, will the thread be blocked again?

core/iwasm/common/wasm_shared_memory.c
loganek: I see you still allocate it on a heap.

core/iwasm/common/wasm_runtime_common.c
hritikgupta: @wenyongh did you mean #1903?

core/iwasm/common/wasm_shared_memory.c
hritikgupta: just wondering if there is any specific advantage of traversing two times? is it to ensure the final traversal is safe i.e. we don't go out of bounds?

loganek: The purpose of that is to avoid multiple reallocations - knowing the size in advance will let you make only one allocation.

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -87,6 +92,62 @@ search_module(WASMModuleCommon *module)
...
+    /* count number of addresses in wait_map */
+    bh_hash_map_traverse(wait_map, wait_map_address_count_callback,
+                         (void *)&total_elem_count);
+
+    /* allocate memory */
+    uint64 total_size = offsetof(AtomicWaitAddressArgs, addr)
+                        + sizeof(void *) * total_elem_count;
+
+    if (!(args->addr = wasm_runtime_malloc((uint32)total_size))) {
+        LOG_ERROR("failed to allocate memory for list of atomic wait addresses "
```
loganek, line 130: You don't free the memory allocated for `args` if we reach here (although I still advise you to allocate it on allocate it on the stack instead).

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -87,6 +92,62 @@ search_module(WASMModuleCommon *module)
...
+    }
+
+    os_mutex_lock(&shared_memory_list_lock);
+
+    /* count number of addresses in wait_map */
+    bh_hash_map_traverse(wait_map, wait_map_address_count_callback,
+                         (void *)&total_elem_count);
+
+    /* allocate memory */
+    uint64 total_size = offsetof(AtomicWaitAddressArgs, addr)
```
loganek, line 126: Why do you need to add offset of `addr` here?

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -87,6 +92,56 @@ search_module(WASMModuleCommon *module)
...
+create_list_of_waiter_addresses(void *key, void *value, void *user_data)
+{
+    AtomicWaitAddressArgs *data = (AtomicWaitAddressArgs *)user_data;
+    data->addr[data->index++] = key;
+}
+
+void
+notify_stale_threads_on_exception(WASMModuleInstanceCommon *module_inst)
+{
+    AtomicWaitAddressArgs *args = NULL;
```
loganek, line 112:
```suggestion
    AtomicWaitAddressArgs args = { 0 };
```

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -87,6 +92,56 @@ search_module(WASMModuleCommon *module)
...
+    uint32 i = 0, total_elem_count = 0;
+    if (!(args = wasm_runtime_malloc(sizeof(AtomicWaitAddressArgs)))) {
+        LOG_ERROR("failed to create atomic wait address args node");
+        return;
+    }
```
loganek, line 117:
```suggestion
```

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -87,6 +92,56 @@ search_module(WASMModuleCommon *module)
...
+    bh_hash_map_traverse(wait_map, create_list_of_waiter_addresses, args);
+    os_mutex_unlock(&shared_memory_list_lock);
+
+    /* notify */
+    for (i = 0; i < args->index; i++) {
+        wasm_runtime_atomic_notify(module_inst, args->addr[i], UINT32_MAX);
+    }
+
+    /* free memory allocated to args data */
+    wasm_runtime_free(args);
```
loganek, line 142:
```suggestion
    wasm_runtime_free(args->addr);
```

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -87,6 +92,56 @@ search_module(WASMModuleCommon *module)
...
+
+    os_mutex_lock(&shared_memory_list_lock);
+
+    /* count number of addresses in wait_map */
+    bh_hash_map_traverse(wait_map, wait_map_address_count_callback,
+                         (void *)&total_elem_count);
+
+    /* allocate memory */
+    uint32 total_size = sizeof(void *) * total_elem_count;
+    void *temp_addr[total_size];
```
loganek: that's I think just a compiler extension and won't work with all compilers. I think you can delete it, see my suggestion below.

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -87,6 +92,56 @@ search_module(WASMModuleCommon *module)
...
+    memset(temp_addr, 0, sizeof(temp_addr));
+
+    args->addr = temp_addr;
```
loganek, line 130:
```suggestion
    args->addr = wasm_runtime_malloc(sizeof(void *) * total_elem_count;
```

loganek (changes requested):

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -87,6 +92,56 @@ search_module(WASMModuleCommon *module)
...
+
+    /* allocate memory */
+    uint32 total_size = sizeof(void *) * total_elem_count;
+    void *temp_addr[total_elem_count];
+    memset(temp_addr, 0, total_size);
+
+    args->addr = temp_addr;
+
+    /* set values in list of addresses */
+    bh_hash_map_traverse(wait_map, create_list_of_waiter_addresses, args);
```
loganek, line 133:
```suggestion
    bh_hash_map_traverse(wait_map, create_list_of_waiter_addresses, &args);
```

core/iwasm/common/wasm_runtime_common.c
wenyongh: Oh, sorry, I meant #1930, but it only works for posix platforms now, maybe this PR is also necessary.

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -87,6 +92,48 @@ search_module(WASMModuleCommon *module)
...
+    uint32 i = 0, total_elem_count = 0;
+
+    os_mutex_lock(&shared_memory_list_lock);
+
+    /* count number of addresses in wait_map */
+    bh_hash_map_traverse(wait_map, wait_map_address_count_callback,
+                         (void *)&total_elem_count);
+
+    /* allocate memory */
+    args.addr = wasm_runtime_malloc(sizeof(void *) * total_elem_count);
```
wenyongh, line 122: Should check the return value, if args.addr is NULL, then unlock shared_memory_list_lock and return?

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -87,6 +92,48 @@ search_module(WASMModuleCommon *module)
...
+
+    /* notify */
+    for (i = 0; i < args.index; i++) {
+        wasm_runtime_atomic_notify(module_inst, args.addr[i], UINT32_MAX);
+    }
+
+    /* free memory allocated to args data */
+    wasm_runtime_free(args.addr);
+}
+
```
wenyongh, line 136: Another issue is the atomic waiting thread might be blocked again after notified by notify_stale_threads_on_exception, when it encounters an atomic_wait opcode again, suggest to check whether there is exception thrown at the beginning of wasm_runtime_atomic_wait in this file, after
```C
    bh_assert(module->module_type == Wasm_Module_Bytecode
              || module->module_type == Wasm_Module_AoT);
```
add
```C
    if (wasm_get_exception(module_inst)) {
        return -1;
    }
```

hritikgupta: done

core/iwasm/common/wasm_shared_memory.c
hritikgupta: done

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -322,6 +374,9 @@ wasm_runtime_atomic_wait(WASMModuleInstanceCommon *module, void *address,
     bh_assert(module->module_type == Wasm_Module_Bytecode
               || module->module_type == Wasm_Module_AoT);
 
+    if (wasm_get_exception(module_inst)) {
+        return -1;
+    }
```
wenyongh, line 379: Had better add an empty line below.

wenyongh: And found this PR is for `dev/wasi_threads`, could you submit to main first? We will merge main into dev/wasi_threads after several PRs are merged.

loganek (approved):

eloparco:
Does this PR work in AOT mode?

Also, can you try it with the sample [thread_termination.c](https://github.com/bytecodealliance/wasm-micro-runtime/blob/dev/wasi_threads/samples/wasi-threads/wasm-apps/thread_termination.c) after replacing https://github.com/bytecodealliance/wasm-micro-runtime/blob/ee1871d3f857a056593ca0d390ce25b6ae492feb/samples/wasi-threads/wasm-apps/thread_termination.c#L34 with a wait operation?

And try with `TEST_TERMINATION_IN_MAIN_THREAD` set to 0 and 1 https://github.com/bytecodealliance/wasm-micro-runtime/blob/ee1871d3f857a056593ca0d390ce25b6ae492feb/samples/wasi-threads/wasm-apps/thread_termination.c#L19

Just to make sure that the base cases are covered.

core/iwasm/common/wasm_shared_memory.c
hritikgupta: i think changing base to main will cause all commits to dev/wasi_threads added to this PR (since the branch was created from dev/wasi_threads), will create a new PR

loganek: You can keep this PR, but change base of your `stale_threads_term` branch to be `main` instead of `dev/wasi_threads`.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -2190,6 +2190,13 @@ wasm_set_exception(WASMModuleInstance *module_inst, const char *exception)
...
+    }
+#endif
```
wenyongh, line 2199: Should move these lines to be after L2204: spread exception firstly to let other threads get exception, and then notify them, so for other thread, if it is atomic waiting and is notified, and if there is another atomic waiting operation, in `wasm_runtime_atomic_wait`, it can detect the exception and return -1 directly. Or it might atomic wait again.

core/iwasm/common/wasm_shared_memory.c
hritikgupta: done

core/iwasm/common/wasm_runtime_common.c
hritikgupta: makes sense, done

PR #1930, opened 2023-02-02
Allow interrupting blocking instructions

eloparco:
Implementing #1910


eloparco:
I tested different combinations of blocking from main thread or from spawn thread, interpreter or aot, with or without hw bound check enabled.
So far the combination that doesn't seem to work is aot +  hw bound check disabled, where I see that the exit code is not the one expected. I'll keep investigating.

wenyongh:
> I tested different combinations of blocking from main thread or from spawn thread, interpreter or aot, with or without hw bound check enabled. So far the combination that doesn't seem to work is aot + hw bound check disabled, where I see that the exit code is not the one expected. I'll keep investigating. @wenyongh

@eloparco Thanks a lot, the code is much better now, I will look into it soon.

build-scripts/config_common.cmake
```diff
@@ -248,6 +248,12 @@ else ()
     add_definitions (-DWASM_DISABLE_STACK_HW_BOUND_CHECK=0)
   endif ()
 endif ()
+if (WAMR_DISABLE_BLOCK_INSN_INTERRUPT EQUAL 1)
```
loganek, line 251: Should we initialize `WAMR_DISABLE_BLOCK_INSN_INTERRUPT` in `CMakeLists.txt` (similarly to what we do for other variables in `product-mini/platforms/linux/CMakeLists.txt`

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -315,7 +319,35 @@ wasm_runtime_get_exec_env_tls()
...
+    act.sa_handler = os_interrupt_block_insn_sig_handler;
+    sigfillset(&act.sa_mask);
+    if (sigaction(SIGUSR1, &act, NULL) < 0) {
+        LOG_ERROR("failed to set signal handler");
+        return false;
+    }
+
+    return true;
+}
+#endif
```
wenyongh, line 350: Should not call posix API directly in this file, move it into posix_thread.c

core/shared/platform/android/platform_internal.h
```diff
@@ -63,19 +63,17 @@ typedef sem_t korp_sem;
 
 #define bh_socket_t int
 
+#if WASM_DISABLE_BLOCK_INSN_INTERRUPT == 0
+#define OS_ENABLE_BLOCK_INSN_INTERRUPT
+#endif
```
wenyongh, line 68: The feature is only required by multi-threading, had better:
#if WASM_DISABLE_BLOCK_INSN_INTERRUPT == 0 && WASM_ENABLE_THREAD_MGR != 0

core/shared/platform/windows/platform_internal.h
```diff
@@ -89,18 +89,15 @@ os_mem_decommit(void *ptr, size_t size);
 #define strncasecmp _strnicmp
 #define strcasecmp _stricmp
 
+#if WASM_DISABLE_BLOCK_INSN_INTERRUPT == 0
+#define OS_ENABLE_BLOCK_INSN_INTERRUPT
+#endif
```
wenyongh, line 94: Windows doesn't support posix like API, had better not change this file currently

core/iwasm/aot/aot_runtime.c
```diff
@@ -1444,8 +1458,19 @@ aot_call_function(WASMExecEnv *exec_env, AOTFunctionInstance *function,
         }
 #endif
 
-        ret = invoke_native_internal(exec_env, function->u.func.func_ptr,
-                                     func_type, NULL, NULL, argv, argc, argv);
+#ifdef OS_ENABLE_BLOCK_INSN_INTERRUPT
+        wasm_exec_env_push_jmpbuf(exec_env, &jmpbuf_node);
+        wasm_runtime_set_exec_env_tls(exec_env);
+        if (os_setjmp(jmpbuf_node.jmpbuf) == 0)
```
wenyongh, line 1464: Same as above, add else branch to set `ret = false`

core/iwasm/aot/aot_runtime.c
```diff
@@ -1444,8 +1458,19 @@ aot_call_function(WASMExecEnv *exec_env, AOTFunctionInstance *function,
...
+#endif
+            ret =
+                invoke_native_internal(exec_env, function->u.func.func_ptr,
+                                       func_type, NULL, NULL, argv, argc, argv);
```
xujuntwt95329, line 1468: How about add a pair of braces? Otherwise it may be easy to make a mistake when adding new lines

```suggestion
        {
            ret =
                invoke_native_internal(exec_env, function->u.func.func_ptr,
                                       func_type, NULL, NULL, argv, argc, argv);
        }
```

eloparco: not a big deal since the code formatter would show that with the indentation; I'm adding it anyway since it makes the code cleaner with the `else` statement

core/shared/platform/windows/platform_internal.h
eloparco: isn't `setjmp` supported on Windows? https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/setjmp

build-scripts/config_common.cmake
eloparco: I was following what's done for `WAMR_DISABLE_HW_BOUND_CHECK` since `WAMR_DISABLE_BLOCK_INSN_INTERRUPT` is available on the same platforms where HW bound checks is.

Right now, it is enabled by default (as it is for HW bound checks).
Maybe for Windows we can disable them both directly in `product-mini/platforms/windows/CMakeLists.txt` instead of manually in the CI https://github.com/bytecodealliance/wasm-micro-runtime/blob/2eed50b3034b8eec531cce4a782b8c698fea00fd/.github/workflows/compilation_on_windows.yml#L97

eloparco: Actually that's not true, since windows doesn't include `runtime_lib.cmake` as other platforms do https://github.com/bytecodealliance/wasm-micro-runtime/blob/17f3375d530c394536379fc0e0247847ad756de9/product-mini/platforms/linux/CMakeLists.txt#L121
that in turn includes `config_common.cmake` https://github.com/bytecodealliance/wasm-micro-runtime/blob/17f3375d530c394536379fc0e0247847ad756de9/build-scripts/runtime_lib.cmake#L157

core/shared/platform/windows/platform_internal.h
wenyongh: Windows signal is a little more complex, we tried using setjmp/longjmp before, but found sometimes longjmp doesn't work, it cannot jump the place set by setjmp.
Refer to https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/signal?view=msvc-160
It mentions:
`Don't use longjmp unless the interrupt is caused by a floating-point exception (that is, sig is SIGFPE). In this case, first reinitialize the floating-point package by using a call to _fpreset.`

My suggestion is to enable posix platforms first, and enable windows in the future.

eloparco: right, it's also reported in the `longjmp` https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/longjmp

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -348,7 +365,11 @@ wasm_runtime_env_init()
         goto fail5;
     }
 #endif
-
+#ifdef OS_ENABLE_BLOCK_INSN_INTERRUPT
+    if (!os_interrupt_block_insn_init(interrupt_block_insn_sig_handler)) {
+        goto fail6;
```
wenyongh, line 370: No `os_interrupt_block_insn_destroy` function is defined to destroy the signal, normally the below fails should call it.

Suggest to add `os_interrupt_block_insn_init` and `os_interrupt_block_insn_destroy` into `runtime_signal_init` and `runtime_signal_destroy` to avoid changing this function.

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -1064,6 +1064,12 @@ set_exception_visitor(void *node, void *user_data)
         bh_memcpy_s(curr_wasm_inst->cur_exception,
                     sizeof(curr_wasm_inst->cur_exception),
                     wasm_inst->cur_exception, sizeof(wasm_inst->cur_exception));
+
+#ifdef OS_ENABLE_BLOCK_INSN_INTERRUPT
+        bh_assert(curr_exec_env->handle);
+        os_thread_signal(curr_exec_env->handle, SIGUSR1);
```
wenyongh, line 1070: Should not use SIGUSR1 in non-platform related code, suggest to wrap a function `os_thread_killl`, and call that API instead.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -315,7 +319,20 @@ wasm_runtime_get_exec_env_tls()
...
     return exec_env_tls;
 }
-#endif /* end of OS_ENABLE_HW_BOUND_CHECK */
+#endif
+
+#ifdef OS_ENABLE_BLOCK_INSN_INTERRUPT
+static void
+interrupt_block_insn_sig_handler(int sig)
+{
+    bh_assert(sig == SIGUSR1);
```
wenyongh, line 328: Should not use SIGUSR1 in non platform related code, if to assert it, suggest a the signal handler in posix_thread.c and assert there, and call this function inside that signal handler.

core/iwasm/aot/aot_runtime.c
```diff
@@ -1444,8 +1465,26 @@ aot_call_function(WASMExecEnv *exec_env, AOTFunctionInstance *function,
         }
 #endif
 
-        ret = invoke_native_internal(exec_env, function->u.func.func_ptr,
-                                     func_type, NULL, NULL, argv, argc, argv);
+#ifdef OS_ENABLE_BLOCK_INSN_INTERRUPT
+        wasm_exec_env_push_jmpbuf(exec_env, &jmpbuf_node);
+        wasm_runtime_set_exec_env_tls(exec_env);
+        if (os_setjmp(jmpbuf_node.jmpbuf) == 0) {
```
yamt, line 1471: while OS_ENABLE_BLOCK_INSN_INTERRUPT and OS_ENABLE_HW_BOUND_CHECK set up separate jmpbufs.
signal handers seem looking at the only most recent one.

i guess it's simpler to use a single jmpbuf. how do you think?

core/iwasm/aot/aot_runtime.c
```diff
@@ -1444,8 +1465,26 @@ aot_call_function(WASMExecEnv *exec_env, AOTFunctionInstance *function,
...
+#endif
+            ret =
+                invoke_native_internal(exec_env, function->u.func.func_ptr,
+                                       func_type, NULL, NULL, argv, argc, argv);
+#ifdef OS_ENABLE_BLOCK_INSN_INTERRUPT
+        }
+        else {
+            ret = false;
```
yamt, line 1479: who frees eg. "argv1" in wasm_runtime_invoke_native on a termination?

while OS_ENABLE_HW_BOUND_CHECK seems to have the same problem,
it seems much more complex to fix for OS_ENABLE_BLOCK_INSN_INTERRUPT because it's an async event.

core/iwasm/aot/aot_runtime.c
wenyongh: The function which calls setjmp in OS_ENABLE_HW_BOUND_CHECK is really complex and the whole function is controlled by OS_ENABLE_HW_BOUND_CHECK, had better not re-use the setjmp:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/aot/aot_runtime.c#L1239-L1335

Using another signal handler is allowed, since different signal handlers handle different signals, pthread_kill signal should be caught by this PR's new signal handler only.

core/iwasm/aot/aot_runtime.c
wenyongh: Yes, this is a potential issue, maybe there are two ways to resolve it: (1) avoid allocating memory in wasm_runtime_invoke_native, just define a large enough buffer for argv, (2) add a list in exec_env, add the allocating the memory to the list, and free it later.

core/iwasm/aot/aot_runtime.c
yamt: > had better not re-use the setjmp

in that case, the signal handers should not assume it's on the top of the jmpbuf stack.

core/iwasm/aot/aot_runtime.c
yamt: (2) needs to block the signal for OS_ENABLE_BLOCK_INSN_INTERRUPT.
maybe we should invent a convenient api for common cases like memory allocation.

core/iwasm/aot/aot_runtime.c
wenyongh: Right, another jmbuf is pushed in calling function when hw bound check is enabled, seems that we had better add a flag in the WASMJmpBuf, and in signal handler, pop jmpbuf until its flag is matched.

core/iwasm/common/wasm_runtime_common.c
eloparco: What do you mean by "destroy the signal"? For HW bound checks, `thread_signal_inited` is set to false, but in my case I don't see any necessary action to perform https://github.com/bytecodealliance/wasm-micro-runtime/blob/2eed50b3034b8eec531cce4a782b8c698fea00fd/core/shared/platform/common/posix/posix_thread.c#L629

core/iwasm/aot/aot_runtime.c
eloparco: They way it works now: `aot_call_function` saves the jmpbuf for instruction interruption, then the nested `invoke_native_with_hw_bound_check` save the jmpbuf for the hw bound check.
In case of `proc_exit`, the instruction interruption sig handler restores the hw bound check jmpbuf (not the corresponding instruction interruption jmpbuf) and returns to `invoke_native_with_hw_bound_check` that will in turn return to `aot_call_function`.

Not sure if we need to jump directly to `aot_call_function` as you were pointing out, I'm still debugging that since there is one case (i.e. instruction interruption and HW bound check both enabled in AOT mode) that still doesn't work correctly in my tests (the exit code is not the right one).

core/iwasm/aot/aot_runtime.c
eloparco: Are you referring to the call at L1396? Isn't it freed at L1458 as before my changes?

core/iwasm/common/wasm_runtime_common.c
wenyongh: OK, I original thought that we should unset the signal handler, e.g. sigaction(.., NULL), but I found the old code also didn't do that.

core/iwasm/aot/aot_runtime.c
wenyongh: Not sure, seems both ways work.
For the exit code, do you mean the wasi_proc_exit? Is `wasi_ctx->exit_code = rval;` executed after the exception was set?

core/iwasm/aot/aot_runtime.c
```diff
@@ -1385,8 +1388,26 @@ aot_call_function(WASMExecEnv *exec_env, AOTFunctionInstance *function,
...
+        }
+
+        jmpbuf_node_pop = wasm_exec_env_pop_jmpbuf(exec_env);
+        bh_assert(&jmpbuf_node == jmpbuf_node_pop);
+        if (!exec_env->jmpbuf_stack_top) {
+            wasm_runtime_set_exec_env_tls(NULL);
+        }
+#endif
```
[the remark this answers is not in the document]
eloparco: That looks better! Especially in `aot_runtime.c`

core/iwasm/aot/aot_runtime.c
eloparco: > For the exit code, do you mean the wasi_proc_exit? Is wasi_ctx->exit_code = rval; executed after the exception was set?

Exactly, I see the call to `proc_exit` but then it returns 1 at the end in that particular configuration (instruction interruption enabled and HW bound check disabled in AOT mode). I'll debug that tomorrow.

eloparco: I'm testing with the `thread_termination.c` example (updated to use `atomic.wait`) in AOT mode with HW bound check disabled.
I see another exception ("unreachable") being raised after the proc_exit and overriding the first one. Still debugging to understand what's happening there.

core/iwasm/aot/aot_runtime.c
```diff
@@ -1310,11 +1310,53 @@ invoke_native_with_hw_bound_check(WASMExecEnv *exec_env, void *func_ptr,
...
+    wasm_exec_env_push_jmpbuf(exec_env, &jmpbuf_node);
+    wasm_runtime_set_exec_env_tls(exec_env);
+
+    if (os_setjmp(jmpbuf_node.jmpbuf) == 0) {
+        ret = invoke_native_block_insn_interrupt(exec_env, func_ptr, func_type,
+                                                 signature, attachment, argv,
+                                                 argc, argv_ret);
+    }
+    else {
+        ret = false;
```
hritikgupta, line 1348: this line wouldn't be needed, right? as ret is false by default

core/iwasm/aot/aot_runtime.c
eloparco: I made a try with valgrind using interpreter and AOT, with or without HW bound check and no memory leakage is detected. Unfortunately, in case of AOT I see many warnings about invalid memory accesses that may explain the weird behavior that I reported in another comment.

core/iwasm/aot/aot_runtime.c
eloparco: https://github.com/bytecodealliance/wasm-micro-runtime/pull/1930#discussion_r1094289179

core/iwasm/aot/aot_runtime.c
eloparco: Sharing the stack trace, in case I'm missing something obvious.
In AOT with HW bound check disabled, from the thread that calls `proc_exit`, after the exception set because of the `proc_exit` call,
```bash
* thread #2, name = 'iwasm', stop reason = breakpoint 8.1
  * frame #0: 0x000055555557846e iwasm`wasm_cluster_spread_exception(exec_env=0x00005555555f6710, clear=false) at thread_manager.c:1099:28
    frame #1: 0x000055555556243e iwasm`wasm_set_exception(module_inst=0x00005555555f6080, exception="wasi proc exit") at wasm_runtime_common.c:2236:9
    frame #2: 0x0000555555562531 iwasm`wasm_runtime_set_exception(module_inst_comm=0x00005555555f6080, exception="wasi proc exit") at wasm_runtime_common.c:2295:5
  
  frame #3: 0x00005555555b6365 iwasm`wasi_proc_exit(exec_env=0x00005555555f6710, rval=55) at libc_wasi_wrapper.c:989:5
```

I see another "unreachable" exception that overrides the first one and causes problems with the exit code.
(`function.func_name` = `"wasi_thread_start"`)
```bash
thread #2, name = 'iwasm', stop reason = breakpoint 9.1
    frame #0: 0x00005555555623b9 iwasm`wasm_set_exception(module_inst=0x00005555555f6080, exception="unreachable") at wasm_runtime_common.c:2224:13
    frame #1: 0x000055555556247f iwasm`wasm_set_exception_with_id(module_inst=0x00005555555f6080, id=0) at wasm_runtime_common.c:2273:9
    frame #2: 0x00005555555751ec iwasm`aot_set_exception_with_id(module_inst=0x00005555555f6080, id=0) at aot_runtime.c:1554:9
    frame #3: 0x0000555555565e9b iwasm`push_args_end at invokeNative_em64.s:61
    frame #4: 0x00007ffff7fb8a60
    frame #5: 0x0000555555564660 iwasm`wasm_runtime_invoke_native(exec_env=0x00005555555f6710, func_ptr=0x0000000040000834, func_type=0x00005555555e2fc0, signature=0x0000000000000000, attachment=0x0000000000000000, argv=0x00007ffff7fb8dc0, argc=2, argv_ret=0x00007ffff7fb8dc0) at wasm_runtime_common.c:4190:9
    frame #6: 0x0000555555574a3d iwasm`invoke_native_internal(exec_env=0x00005555555f6710, func_ptr=0x0000000040000834, func_type=0x00005555555e2fc0, signature=0x0000000000000000, attachment=0x0000000000000000, argv=0x00007ffff7fb8dc0, argc=2, argv_ret=0x00007ffff7fb8dc0) at aot_runtime.c:1344:15
    frame #7: 0x0000555555575096 iwasm`aot_call_function(exec_env=0x00005555555f6710, function=0x00005555555f66e0, argc=2, argv=0x00007ffff7fb8dc0) at aot_runtime.c:1492:15
    frame #8: 0x00005555555621ff iwasm`wasm_runtime_call_wasm(exec_env=0x00005555555f6710, function=0x00005555555f66e0, argc=2, argv=0x00007ffff7fb8dc0) at wasm_runtime_common.c:1828:15
    frame #9: 0x00005555555b19c9 iwasm`thread_start(arg=0x00005555555f6710) at lib_wasi_threads_wrapper.c:61:10
    frame #10: 0x000055555557783d iwasm`thread_manager_start_routine(arg=0x00005555555f6710) at thread_manager.c:521:11
    frame #11: 0x0000555555580c0c iwasm`os_thread_wrapper(arg=0x00005555556075e0) at posix_thread.c:43:5
    frame #12: 0x00007ffff7d25b43 libc.so.6`start_thread(arg=<unavailable>) at pthread_create.c:442:8
    
frame #13: 0x00007ffff7db7a00 libc.so.6`__clone3 at clone3.S:81

Comment (wenyongh) [reply core/iwasm/aot/aot_runtime.c]: @eloparco Could you upload the wasm file compiled so that we can also debug it? And in interrupt_block_insn_sig_handler, could you try popping the jmpbuf of hw bound check firstly, and then longjmp to the jmpbuf set by invoke_native_internal in aot_runtime.c?

File: core/iwasm/common/wasm_runtime_common.c [hunk 14]
```diff
@@ -315,7 +319,18 @@ wasm_runtime_get_exec_env_tls()
...
     return exec_env_tls;
 }
-#endif /* end of OS_ENABLE_HW_BOUND_CHECK */
+#endif
+
+#ifdef OS_ENABLE_BLOCK_INSN_INTERRUPT
+static void
+interrupt_block_insn_sig_handler()
+{
+    WASMJmpBuf *jmpbuf_node = exec_env_tls->jmpbuf_stack_top;
```
yamt, line 328: jmpbuf_stack_top here can be for another handler.

core/iwasm/aot/aot_runtime.c
yamt: > Are you referring to the call at L1396? Isn't it freed at L1458 as before my changes?

i'm not sure which file and version your line numbers are for.

i meant eg.
https://github.com/bytecodealliance/wasm-micro-runtime/blob/2c8375ee67964a4ef99dde2faa1c5225051adca0/core/iwasm/common/wasm_runtime_common.c#L3506
https://github.com/bytecodealliance/wasm-micro-runtime/blob/2c8375ee67964a4ef99dde2faa1c5225051adca0/core/iwasm/common/wasm_runtime_common.c#L3790

when it escapes with a long jump, it might not be executed.

core/iwasm/aot/aot_runtime.c
eloparco: Thanks, I'll try that.

The easiest way to reproduce is to change [no_pthread.c](https://github.com/bytecodealliance/wasm-micro-runtime/blob/dev/wasi_threads/samples/wasi-threads/wasm-apps/no_pthread.c) to
```c
/*
 * Copyright (C) 2022 Amazon.com Inc. or its affiliates. All rights reserved.
 */
#ifndef __wasi__
#error This example only compiles to WASM/WASI target
#endif

#include <stdlib.h>
#include <stdio.h>
#include <assert.h>

#include "wasi_thread_start.h"

#define STOP_FROM_MAIN 0

static const int64_t SECOND = 1000 * 1000 * 1000;

typedef struct {
    start_args_t base;
    int th_ready;
    int value;
    int thread_id;
} shared_t;

void
__wasi_thread_start_C(int thread_id, int *start_arg)
{
    shared_t *data = (shared_t *)start_arg;

    printf("New thread ID: %d, starting parameter: %d\n", thread_id,
           data->value);

#if STOP_FROM_MAIN == 1
    __builtin_wasm_memory_atomic_wait32(0, 0, -1);
#else
    sleep(2);
    __wasi_proc_exit(55);
#endif
}

int
main(int argc, char **argv)
{
    shared_t data = { { NULL }, 0, 52, -1 };
    int thread_id;
    int ret = EXIT_SUCCESS;

    if (!start_args_init(&data.base)) {
        printf("Stack allocation for thread failed\n");
        return EXIT_FAILURE;
    }

    thread_id = __wasi_thread_spawn(&data);
    if (thread_id < 0) {
        printf("Failed to create thread: %d\n", thread_id);
        ret = EXIT_FAILURE;
        goto final;
    }

#if STOP_FROM_MAIN == 1
    sleep(2);
    __wasi_proc_exit(56);
#else
    __builtin_wasm_memory_atomic_wait32(0, 0, -1);
#endif

final:
    start_args_deinit(&data.base);

    return ret;
}
```
and add `set(WAMR_DISABLE_HW_BOUND_CHECK 1)` in the [CMakeLists.txt](https://github.com/bytecodealliance/wasm-micro-runtime/blob/dev/wasi_threads/samples/wasi-threads/CMakeLists.txt).

core/iwasm/common/wasm_runtime_common.c
eloparco: Yes, if HW bound check is enabled, it will get a "HW bound" `jmpbuf` https://github.com/bytecodealliance/wasm-micro-runtime/pull/1930#discussion_r1096389777, but that should not be a problem unless I'm missing something

yamt: you can even get a signal between wasm_exec_env_push_jmpbuf and os_setjmp.

eloparco: > you can even get a signal between wasm_exec_env_push_jmpbuf and os_setjmp.

In that case, the `jmpbuf` would be invalid, as it is in general before being set up by `os_setjmp`. Isn't that the same for the HW bound check before my PR?

I wonder if we have to disable/re-enable the signals to avoid the signal handler being called in the middle. We want to avoid signal handlers being called between `wasm_exec_env_push_jmpbuf` and `os_setjmp` as they would use a non-initialized `jmpbuf` iiuc.

core/iwasm/aot/aot_runtime.c
eloparco: > And in interrupt_block_insn_sig_handler, could you try popping the jmpbuf of hw bound check firstly, and then longjmp to the jmpbuf set by invoke_native_internal in aot_runtime.c?

Actually I have the reported problem when `WAMR_DISABLE_HW_BOUND_CHECK=1`, so that's not the cause.

core/iwasm/aot/aot_runtime.c
eloparco: Got it, for this one https://github.com/bytecodealliance/wasm-micro-runtime/blob/2c8375ee67964a4ef99dde2faa1c5225051adca0/core/iwasm/common/wasm_runtime_common.c#L4130 in my tests the `if` is never entered and so there's no memory allocation, that's why I was not observing any memory leakage in my configuration.

But, right, there's a potential memory leakage.

core/iwasm/aot/aot_runtime.c
eloparco: Just realized that there are problems in general with AOT. When I enabled it in #1905 I only tried with the `no_pthread.c` example. Running it again (from the HEAD of `dev/wasi_threads`) I see that the `proc_exit` doesn't behave correctly, not stopping even with non-blocking instructions.
So the problem is not specific to this PR.

I'll address the other review comments and close the PR once it's ready, AOT will be fixed in a separate PR.

core/iwasm/aot/aot_runtime.c
eloparco: (3) what if we allocate it in the parent function?

core/iwasm/aot/aot_runtime.c
wenyongh: Good, how about submitting a PR to main branch firstly since the same issue also exists in main branch? I will merge main into dev/wasi_threads after several PRs are merged into main: #1924, #1929, #1902.

core/iwasm/aot/aot_runtime.c
wenyongh: @yamt How about limiting the max argument count to 32 and reporting error if argument count is larger than that? So there is no memory allocation in wasm_runtime_invoke_native.
I think in normal usage, there isn't a function taking so many arguments, 32 is big enough: `uint64 argv_buf[32] = { 0 }`

wenyongh: Have a test on the spec cases, need more size, considering 36 or 48

yamt: well, `argv1` is just an example. it isn't the only problem or the biggest problem.
the native functions themselves likely have more important things to cleanup.
eg. locks used in wasi implementation need to be released when it's terminated with the signal.

the only simple solution i can think of right now involves block/unblock of the signal.
like:
```
block signal
get a resource (allocate memory / acquire lock / etc)
register cleanup handler for the resource
unblock signal
```
how do you think?

core/iwasm/common/wasm_runtime_common.c
yamt: it isn't a problem for hw bound check because the signal for it is basically synchronous to the thread execution. ie. the thread itself can control where it can happen.

core/iwasm/aot/aot_runtime.c
wenyongh: The issue is only introduced by this PR which may raise the signal to kill other threads, right?

yamt: right.
well, there might be similar issues in the hw bound check code. i dunno. but at least this PR makes far much code affected because the signal in question is sent by other threads asynchronously.

core/iwasm/common/wasm_runtime_common.c
eloparco: Can we maybe invert the order of `wasm_exec_env_push_jmpbuf` and `os_setjmp`? So that the `jmpbuf` is only pushed after being initialized by `os_setjmp`

core/iwasm/aot/aot_runtime.c
wenyongh: It is really complex, we just don't know the possible behaviors of user implemented native API, there are some questions:
- How many resources do we want to protect? The memory allocated from wasm_runtime_malloc, the initialized lock/cond? And if developer allocates memory from non wasm_runtime_malloc?
- What the blocking situations include? `while (1)` in native, sleep, timedwait, and lock? Is it reasonable to terminate them?
- How to protect the resources, the process may be like what you proposed, but need to break down them. Do we want to implement them in this PR, or in another PR?

core/iwasm/common/wasm_runtime_common.c
eloparco: > you can even get a signal between wasm_exec_env_push_jmpbuf and os_setjmp.

Apart from this, do you think there are other cases that are not covered?

core/iwasm/aot/aot_runtime.c
eloparco: If we decide to implement it in another PR we can probably disable the feature by default.

wenyongh: Yes, I think we had better create another PR for the further enhancement. How about we creating another branch `dev/interrup_block_insn` from main, submitting this PR to that branch, and then we can submit following PRs to it. And after the feature is tested more and is relatively stable, we can merge it to main.
@yamt What's you opinion?

yamt: i'm afraid that it needs to be more like code audit than testing.
i suppose disabling it by default and enabling it on per-function basis can help.

i feel having another branch is too much as far as it's disabled by default. but i don't have strong opinions either way.

eloparco: Should we focus on avoiding memory allocations in the function called after `os_setjmp`? E.g. 
https://github.com/eloparco/wasm-micro-runtime/blob/9d6dc29a2d4dc85cf759dc3066741b2a832423f3/core/iwasm/interpreter/wasm_runtime.c#L2142
and
https://github.com/eloparco/wasm-micro-runtime/blob/9d6dc29a2d4dc85cf759dc3066741b2a832423f3/core/iwasm/interpreter/wasm_runtime.c#L2144

Because if a signal is received before `os_setjmp`, it will be ignored in the signal handler. If it's received after, [this](https://github.com/eloparco/wasm-micro-runtime/blob/9d6dc29a2d4dc85cf759dc3066741b2a832423f3/core/iwasm/interpreter/wasm_runtime.c#L2156) is the function that could leak memory. 

After https://github.com/eloparco/wasm-micro-runtime/blob/9d6dc29a2d4dc85cf759dc3066741b2a832423f3/core/iwasm/interpreter/wasm_runtime.c#L2160, there is no `jmpbuf`, so the signal handler would [return](https://github.com/eloparco/wasm-micro-runtime/blob/9d6dc29a2d4dc85cf759dc3066741b2a832423f3/core/iwasm/common/wasm_runtime_common.c#L330) without jumping.

wenyongh: > i'm afraid that it needs to be more like code audit than testing. i suppose disabling it by default and enabling it on per-function basis can help.
> 
> i feel having another branch is too much as far as it's disabled by default. but i don't have strong opinions either way.

Two reasons to merge it into a development branch:
(1) avoid merging a not so mature PR to the main branch quickly, the PR changes the vmcore part, I think we had better be cautious and do more test on it. And actually the feature is enabled by default for posix platforms, see https://github.com/bytecodealliance/wasm-micro-runtime/pull/1930/files#diff-77a81caccc9f723fd431bfc253006da5129f55a3c93b6275b01ec1f642f9b6ee
```C
#if WASM_DISABLE_BLOCK_INSN_INTERRUPT == 0 && WASM_ENABLE_THREAD_MGR != 0
#define OS_ENABLE_BLOCK_INSN_INTERRUPT
```
(2) we can then try to resolve the memory allocation issue: protect the allocated memory after os_setjmp, and disable the signal before/after calling wasm_runtime_malloc/free.

I will merge it to main branch as soon as possible when (2) is done and well tested.

wenyongh: > Should we focus on avoiding memory allocations in the function called after `os_setjmp`?

Agree, let's resolve the memory allocation issue firstly.

eloparco: > I will merge it to main branch as soon as possible when (2) is done and well tested.

At this point, we can probably just have the feature disabled by default until it's well tested.

> the only simple solution i can think of right now involves block/unblock of the signal.

The tricky part is that it's not always possible to just block signals before the malloc and unblock after the free, because the blocking instruction can be in the middle. Apart from `argv1 ` that we could avoid allocating dynamically, the problem is in the native function calls.

Not sure what's the best solution there, maybe the approach of keeping a list in `exec_env` is not that bad, but only for native calls otherwise we only release the entire allocated memory at the end of the execution. But then how do we decide when to use the exec_env list and when to free normally? It seems quite error prone.

wenyongh: Yes, suggest to add a list of allocated memories after setjmp in the exec_env.
@eloparco I have created branch `dev/interrupt_block_insn`, could you upload another patch to it? Or do you mind I refine some code and upload the PR by myself?

eloparco: Thanks @wenyongh, I pushed the changes I made so far in #1946.

wenyongh: @eloparco I submitted PR #1947 to refactor the feature, please help review, thanks.

wenyongh: Sorry, the source branch and dest branch are invalid, closed it and submitted #1948 instead

wenyongh:
Test this feature with branch `dev/interrupt_block_insn`, and PR #1946, #1948 were merged.

PR #1931, opened 2023-02-02
Runtime memory allocation in managed heap when user import alloc/free

linzang:
In our case, we want to use managed heap provided by wamr for "wasm runtime data", such as loaded wasm code, resolved wasm module etc. And our case also use external allocate/free in assembly script for the data in wasm heap.  However, the current wamr seems only support to allocate runtime data in system heap by using os_malloc/os_free when wasm heap is controlled by external allocate/free..

So we made this change, expecting it could satisfy our cases and help improve wamr.


linzang:
Close this one, it is submitted by mistake.

PR #1932, opened 2023-02-02
Loosen the wasi abi compatibility check in loader

wenyongh:
Change an error to warning when checking wasi abi compatibility in loader, for rust case below:
#[no_mangle]
pub extern "C" fn main() {
    println!("foo");
}
compile it with `cargo build --target wasm32-wasi`, a wasm file is generated with wasi apis imported
and a "void main(void)" function exported.

Other runtime e.g. wasmtime allows to load it and execute the main function with `--invoke` option.


lum1n0us (approved):
LGTM

PR #1937, opened 2023-02-03
tests: Save .wasm/.aot file to report dir if test failed

no1wudi:


tests/wamr-test-suites/spec-test-script/runtest.py
```diff
@@ -200,6 +200,8 @@ def assert_prompt(runner, prompts, timeout, is_need_execute_result):
         help="Use direct pipes instead of pseudo-tty")
 parser.add_argument('--log-file', type=str,
         help="Write messages to the named file in addition the screen")
+parser.add_argument('--log-dir', type=str,
+        help="Write messages to the named file in addition the screen")
```
wenyongh, line 204: Invalid help prompt

PR #1941, opened 2023-02-07
Enable specifying out-of-source platform configuration cmake file

### LINKED ISSUE #1935: Building the runtime with an out-of-repository platform is hacky
I want to build WAMR's runtime with custom, internal platforms, but our team does not want to maintain a patch set or fork of the repository.

Here's how `runtime_lib.cmake` uses some key CMake variables in the current release (1.1.2):
```cmake
include (${SHARED_DIR}/platform/${WAMR_BUILD_PLATFORM}/shared_platform.cmake)
include (${SHARED_DIR}/mem-alloc/mem_alloc.cmake)
include (${IWASM_DIR}/common/iwasm_common.cmake)
include (${SHARED_DIR}/utils/shared_utils.cmake)
```

I could touch `SHARED_DIR`, but this would require us to also maintain `mem_alloc.cmake` and `shared_utils.cmake` in our own repository.

I could touch `WAMR_BUILD_PLATFORM` to use a relative path. For example: 
```cmake
set (WAMR_BUILD_PLATFORM ../../../../MyRepo/WAMR)
```
But, this is a hack and breaks if the directory structure changes in depth anywhere. It also breaks and string parsing on `WAMR_BUILD_PLATFORM` for configuration.

I talked with the team, and one solution we propose is setting another variable to set the absolute directory path to `shared_platform.cmake` independent of both `WAMR_BUILD_PLATFORM` and `SHARED_DIR`.

loganek:
resolves #1935

### DIFF
--- a/build-scripts/runtime_lib.cmake
+++ b/build-scripts/runtime_lib.cmake
@@ -19,6 +19,11 @@ endif ()
 if (NOT DEFINED DEPS_DIR)
     set (DEPS_DIR ${WAMR_ROOT_DIR}/core/deps)
 endif ()
+if (NOT DEFINED SHARED_PLATFORM_CONFIG)
+    # CMake file for platform configuration. The PLATFORM_SHARED_SOURCE varable
+    # should point to a list of platform-specfic source files to compile.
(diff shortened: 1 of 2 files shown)


wenyongh (commented):
LGTM

PR #1942, opened 2023-02-07
added gh api call for fetching llvm version

### LINKED ISSUE #1877: Speed up by querying Github API
https://github.com/bytecodealliance/wasm-micro-runtime/blob/b92075e283bd574ea8b809845031f86b1879c67a/.github/workflows/build_llvm_libraries.yml#L29-L32

Instead of cloning llvm-project repo, use [a Github API](https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#list-commits) to list the latest commit.

hritikgupta:
Fixes #1877


build-scripts/build_llvm.py
```diff
@@ -7,6 +7,7 @@
 import argparse
 import os
 import pathlib
+import requests
```
loganek, line 10: By adding this we probably also need to add `requirements.txt` as this is not a built in module.

build-scripts/build_llvm.py
```diff
@@ -232,14 +244,17 @@ def main():
         "arc": {
             "repo": "https://github.com/llvm/llvm-project.git",
             "branch": "release/15.x",
+            "owner_project": "llvm/llvm-project",
```
loganek, line 247: do we need to replicate that information? It's already available in the `repo` field.

.github/workflows/build_llvm_libraries.yml
```diff
@@ -28,7 +28,7 @@ jobs:
```
lum1n0us, line 28: may need to install python dependencies(from requirements.txt) before

build-scripts/build_llvm.py
lum1n0us: @loganek shall we create a global requirements.txt under */repo* or just create a local requirements.txt for *build_llvm.py* only.

build-scripts/build_llvm.py
```diff
@@ -27,11 +28,22 @@ def clone_llvm(dst_dir, llvm_repo, llvm_branch):
...
 
 
-def query_llvm_version(llvm_dir):
-    GIT_LOG_CMD = f"git log --format=format:'%h' -1"
-    return subprocess.check_output(
-        shlex.split(GIT_LOG_CMD), cwd=llvm_dir, universal_newlines=True, text=True
-    )
+def query_llvm_version(llvm_info):
+    github_token = os.environ['GH_TOKEN']
+    url = f"https://api.github.com/repos/{llvm_info['owner_project']}/commits/{llvm_info['branch']}"
```
lum1n0us, line 33: just a minor one: `owner_project` -> `ownwer`

build-scripts/build_llvm.py
hritikgupta: done

build-scripts/build_llvm.py
hritikgupta: the string is a combination of owner and project, isn't it?

build-scripts/build_llvm.py
hritikgupta: done, but i've added it locally under build-scripts

loganek: I think we can keep the file local; looks like that's what we do for `wasi-nn` tests and `language-bindings` for python as well.

build-scripts/requirements.txt
```diff
@@ -0,0 +1 @@
+requests
```
loganek, line 1: no minimum version?

build-scripts/build_llvm.py
hritikgupta: 👍 , kept it local for the same reason

build-scripts/requirements.txt
hritikgupta: added latest

loganek (commented):
LGTM

loganek (approved):

PR #1945, opened 2023-02-08
fix: explicit casts and types in espidf_socket.c

cookpate:
In the WAMR esp-idf platform, Xtensa GCC 8.4.0 reported incompatible pointer warnings when building with the lwip component.

Berkeley (POSIX) sockets uses composition in combination with type punning to handle many protocol families, including IPv4 & IPv6. The type punning just has to be made explicit with pointer casts from `sockaddr_in` for IPv4 to the generic `sockaddr`.


wenyongh (approved):
LGTM

PR #1950, opened 2023-02-10
Fix build failure when using clang

### LINKED ISSUE #1949: Fail to build iwasm after #1928 merged
I followed the instructions in [WASR Document - Docker Environment Preparation](https://wamr.gitbook.io/document/basics/getting-started/devcontainer). In the container, my toolchain version is as follows:

- cmake version 3.25.2
- gcc version 9.4.0 (Ubuntu 9.4.0-1ubuntu1~20.04.1)
- clang version 15.0.7

When I try to build iwasm vmcore, I get this error:

``` plain
refcount.h:132:2: error: "Reference counter isn't implemented"
```

I tried reverting commit 17f3375, which temporarily fixed the issue. I don't understand why this problem occurs, can someone help me?

Full cmake log attached: [cmake.log](https://github.com/bytecodealliance/wasm-micro-runtime/files/10702665/cmake.log)

CSZongzi:
After some investigation, I found that #1928 will cause clang to fail to compile.

Fix #1949

### DIFF
--- a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/ssp_config.h
+++ b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/ssp_config.h
@@ -113,7 +113,7 @@ so we have to handle this case specially */
 #if defined(__clang__)
 /* Clang provides stdatomic.h since 3.6.0
 See https://releases.llvm.org/3.6.0/tools/clang/docs/ReleaseNotes.html */
-#if __clang_major__ > 3 || (__clang_minor__ == 3 && __clang_minor__ >= 6)
+#if __clang_major__ > 3 || (__clang_major__ == 3 && __clang_minor__ >= 6)
 #define CONFIG_HAS_STD_ATOMIC 1
 #else
 #define CONFIG_HAS_STD_ATOMIC 0
--- a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/ssp_config.h
+++ b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/ssp_config.h
@@ -113,7 +113,7 @@ so we have to handle this case specially */
 #if defined(__clang__)
 /* Clang provides stdatomic.h since 3.6.0
 See https://releases.llvm.org/3.6.0/tools/clang/docs/ReleaseNotes.html */
-#if __clang_major__ >= 3 || (__clang_minor__ == 3 && __clang_minor__ >= 6)
+#if __clang_major__ > 3 || (__clang_minor__ == 3 && __clang_minor__ >= 6)
 #define CONFIG_HAS_STD_ATOMIC 1
 #else
 #define CONFIG_HAS_STD_ATOMIC 0
--- a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/ssp_config.h
+++ b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/ssp_config.h
@@ -108,11 +108,15 @@
 #endif
 
 #if !defined(BH_PLATFORM_LINUX_SGX)
+/* Clang's __GNUC_PREREQ macro has a different meaning than GCC one,
+so we have to handle this case specially */
 #if defined(__clang__)
 /* Clang provides stdatomic.h since 3.6.0
 See https://releases.llvm.org/3.6.0/tools/clang/docs/ReleaseNotes.html */
 #if __clang_major__ >= 3 || (__clang_minor__ == 3 && __clang_minor__ >= 6)
 #define CONFIG_HAS_STD_ATOMIC 1
+#else
+#define CONFIG_HAS_STD_ATOMIC 0
 #endif
 #elif defined(__GNUC_PREREQ)
 /* Even though older versions of GCC support C11, atomics were
--- a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/ssp_config.h
+++ b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/ssp_config.h
@@ -108,7 +108,13 @@
 #endif
 
 #if !defined(BH_PLATFORM_LINUX_SGX)
-#if defined(__GNUC_PREREQ)
+#if defined(__clang__)
+/* Clang provides stdatomic.h since 3.6.0
+See https://releases.llvm.org/3.6.0/tools/clang/docs/ReleaseNotes.html */
+#if __clang_major__ >= 3 || (__clang_minor__ == 3 && __clang_minor__ >= 6)
+#define CONFIG_HAS_STD_ATOMIC 1
+#endif
+#elif defined(__GNUC_PREREQ)
 /* Even though older versions of GCC support C11, atomics were
 not implemented until 4.9. See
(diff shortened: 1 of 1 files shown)


core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/ssp_config.h
```diff
@@ -108,7 +108,13 @@
 #endif
 
 #if !defined(BH_PLATFORM_LINUX_SGX)
-#if defined(__GNUC_PREREQ)
+#if defined(__clang__)
+/* Clang provides stdatomic.h since 3.6.0
+See https://releases.llvm.org/3.6.0/tools/clang/docs/ReleaseNotes.html */
+#if __clang_major__ >= 3 || (__clang_minor__ == 3 && __clang_minor__ >= 6)
```
loganek, line 114: Is `__GNUC_PREREQ` defined for clang? If so, I guess we should be able to slightly simplify it by updating line 121:
```
#if (defined(__clang__) && __GNUC_PREREQ(3, 6)) || __GNUC_PREREQ(4,9)
```

CSZongzi: Unfortunately, clang doesn't have a `__GNUC_PREREQ` macro, so I can't simplify it. See [Clang Language Extensions - Builtin Macros](https://clang.llvm.org/docs/LanguageExtensions.html#builtin-macros).

loganek: hmm, in that case I wonder why the previous code didn't work? If `__GNUC_PREREQ` is not defined, it should go to `#define CONFIG_HAS_STD_ATOMIC 1`?
I added indentation for visibility below
```c
1.  #if !defined(BH_PLATFORM_LINUX_SGX) // true for clang
2.  #  if defined(__GNUC_PREREQ)  // false for clang, go to line 8
3.  #    if __GNUC_PREREQ(4, 9)
4.  #      define CONFIG_HAS_STD_ATOMIC 1
5.  #    else /* else of __GNUC_PREREQ(4, 9) */
6.  #      define CONFIG_HAS_STD_ATOMIC 0
7.  #    endif /* end of __GNUC_PREREQ(4, 9) */
8.  #  else  /* else of defined(__GNUC_PREREQ) */  // true for clang
9.  #    define CONFIG_HAS_STD_ATOMIC 1  // <------
10. #  endif /* end of defined(__GNUC_PREREQ) */
11. #else  /* else of !defined(BH_PLATFORM_LINUX_SGX) */
12. #  define CONFIG_HAS_STD_ATOMIC 0
13. #endif /* end of !defined(BH_PLATFORM_LINUX_SGX) */
```

CSZongzi: I make a test, and it shows a strange situation:

``` c
#include <stdio.h>

int
main()
{
#ifdef __GNUC_PREREQ
    printf("__GNUC_PREREQ\n"); // This line is compiled and executed
#endif

#if __GNUC_PREREQ(4, 9)
    printf("__GNUC_PREREQ(4, 9)\n"); // This line is not
#endif

#if __GNUC_PREREQ(4, 2)
    printf("__GNUC_PREREQ(4, 2)\n"); // This line is compiled and executed
#endif
}
```

After reading the [Clang Compiler User’s Manual](https://clang.llvm.org/docs/UsersManual.html), I found this:

> Note that because clang pretends to be like GCC 4.2, and...

and this:

> -fgnuc-version=
> This flag controls the value of `__GNUC__` and related macros. This flag does not enable or disable any GCC extensions implemented in Clang. Setting the version to zero causes Clang to leave `__GNUC__` and other GNU-namespaced macros, such as `__GXX_WEAK__`, undefined.

Then, I found a stackoverflow question about this issue: [How to tell Clang to stop pretending to be other compilers?](https://stackoverflow.com/questions/38499462/how-to-tell-clang-to-stop-pretending-to-be-other-compilers)

Now I understand why it happens. Clang pretends to be like GCC 4.2, so `__GNUC_PREREQ` is defined, but `__GNUC__`.`__GNUC_MINOR__` will always be `4`.`2` until clang support all GNU C 4.3 features (currently not).

Well, the problem is more complicated than I thought.

loganek: Thanks for sharing that! I think we have two options then:
1. We keep your code as it is but we also need to set `CONFIG_HAS_STD_ATOMIC` to `0` for clang older than 3.6
2. We do the following change
```
#if defined(__GNUC_PREREQ)
```
to be:
```
#if defined(__GNUC_PREREQ) && !defined(__clang__)
```
as that branch is for GCC exclusively, so handling clang 


up to you

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/ssp_config.h
```diff
@@ -108,7 +108,13 @@
...
+#define CONFIG_HAS_STD_ATOMIC 1
```
wenyongh, line 115: Agree with @loganek, here had better add
```C
#else
#define CONFIG_HAS_STD_ATOMIC 0
```

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/ssp_config.h
CSZongzi: Thanks for pointing this out, I chose the first way and added an else statement.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/ssp_config.h
CSZongzi: The corresponding statement has been added, thanks for your review.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/ssp_config.h
```diff
@@ -108,7 +108,17 @@
 #endif
 
 #if !defined(BH_PLATFORM_LINUX_SGX)
-#if defined(__GNUC_PREREQ)
+/* Clang's __GNUC_PREREQ macro has a different meaning than GCC one,
+so we have to handle this case specially */
+#if defined(__clang__)
+/* Clang provides stdatomic.h since 3.6.0
+See https://releases.llvm.org/3.6.0/tools/clang/docs/ReleaseNotes.html */
+#if __clang_major__ >= 3 || (__clang_minor__ == 3 && __clang_minor__ >= 6)
```
wenyongh, line 116: This line is invalid? Should change
```C
#if __clang_major__ >= 3 || (__clang_minor__ == 3 && __clang_minor__ >= 6)
```
to:
```C
#if __clang_major__ > 3 || (__clang_major__ == 3 && __clang_minor__ >= 6)
```

CSZongzi: Uh, The original code works, but your version is better. Thanks for your careful inspection.

wenyongh (commented):
LGTM

loganek (approved):
LGTM, thanks.

PR #1951, opened 2023-02-10
modified poll_oneoff to make it interruptible

hritikgupta:


core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -951,6 +951,25 @@ wasi_path_remove_directory(wasm_exec_env_t exec_env, wasi_fd_t fd,
...
+get_timeout_for_poll_oneoff(const wasi_subscription_t *in)
+{
+    int timeout = 0;
+    const __wasi_subscription_t *clock_subscription = NULL;
+    if (clock_subscription == NULL
+        && (in->u.u.clock.flags & __WASI_SUBSCRIPTION_CLOCK_ABSTIME) == 0) {
+        clock_subscription = in;
+    }
+    if (clock_subscription != NULL) {
+        __wasi_timestamp_t ts = clock_subscription->u.u.clock.timeout / 1e9;
```
hritikgupta, line 964: should this be 1e9 or 1e6 as written [here](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c#L2693)?

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -951,6 +951,25 @@ wasi_path_remove_directory(wasm_exec_env_t exec_env, wasi_fd_t fd,
     return wasmtime_ssp_path_remove_directory(curfds, fd, path, path_len);
 }
 
+static int
```
loganek, line 954: is int large enough for that? Or should we rather use `__wasi_timestamp_t` instead?

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
loganek: the timeout is in `nanoseconds`, so if you want to convert it to seconds, it should be `1e9`.

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -969,9 +988,26 @@ wasi_poll_oneoff(wasm_exec_env_t exec_env, const wasi_subscription_t *in,
...
+        if (timeout != -1 && elapsed > timeout)
+            break;
+
+        in_copy[0].u.u.clock.timeout = time_quant * 1e9;
+        err = wasmtime_ssp_poll_oneoff(curfds, in_copy, out, nsubscriptions,
+                                       &nevents);
+        if (err)
+            return err;
+
+        if (wasm_runtime_get_exception(module_inst) || nevents_app >= 0)
```
loganek, line 1008: isn't the condition `nevents_app >= 0` always true?

loganek (changes requested):

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
hritikgupta: right, was unsure why it was 1e6 above because of the same reason

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
hritikgupta: was trying to keep it consistent with the data type [here](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c#L2688)

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
hritikgupta: was a bit unsure of this here, should the interrupt be dependent on the number of events and just rely on the exception occurrence?

loganek: We should stop the loop when the event happened (we shouldn't call `poll_oneof` again, it would be incorrect behavior). Having said that, when there's an event, we shouldn't return `EINTR` error code.

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -969,9 +988,26 @@ wasi_poll_oneoff(wasm_exec_env_t exec_env, const wasi_subscription_t *in,
...
         return (wasi_errno_t)-1;
 
-    err = wasmtime_ssp_poll_oneoff(curfds, in, out, nsubscriptions, &nevents);
-    if (err)
-        return err;
+    int elapsed = 0;
+    /* in seconds */
+    const int timeout = get_timeout_for_poll_oneoff(in), time_quant = 1;
+    __wasi_subscription_t in_copy[nsubscriptions];
+    memcpy(in_copy, in, nsubscriptions * sizeof(wasi_subscription_t));
```
wenyongh, line 995: Please use bh_memcpy_s to improve safety

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
wenyongh: Agree with @loganek, should stop the loop here when the event(s) occured.
And here we should use `nevents > 0` instead of `nevents_app >= 0`, nevents_app is the address pointing to some place of the linear memory space, it will always `>= 0` and is a valid address since it is checked by L988.
nevents is the returned count of events occured

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
hritikgupta: done

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -969,9 +988,30 @@ wasi_poll_oneoff(wasm_exec_env_t exec_env, const wasi_subscription_t *in,
         || !validate_native_addr(nevents_app, sizeof(uint32)))
         return (wasi_errno_t)-1;
 
-    err = wasmtime_ssp_poll_oneoff(curfds, in, out, nsubscriptions, &nevents);
-    if (err)
-        return err;
+    int elapsed = 0;
+    /* in seconds */
+    const int timeout = get_timeout_for_poll_oneoff(in), time_quant = 1;
+    __wasi_subscription_t in_copy[nsubscriptions];
```
hritikgupta, line 994: convert this to malloc

loganek, line 994: This a [gcc extension](https://gcc.gnu.org/onlinedocs/gcc/Variable-Length.html); normally you can't use variable for the size as the size needs to be known at compile time. Better to allocate it explicitly on the heap.

hritikgupta: done

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -119,7 +119,9 @@ notify_stale_threads_on_exception(WASMModuleInstanceCommon *module_inst)
                          (void *)&total_elem_count);
 
     /* allocate memory */
-    if (!(args.addr = wasm_runtime_malloc(sizeof(void *) * total_elem_count))) {
+    const uint32 total_elem_count_size = sizeof(void *) * total_elem_count;
```
loganek, line 122: that probably should be `size_t`? Also, should we check for overflow?

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -119,7 +119,9 @@ notify_stale_threads_on_exception(WASMModuleInstanceCommon *module_inst)
...
+    if (total_elem_count_size
```
loganek, line 123: should we just return from this function if `total_elem_count` is zero?

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -951,6 +951,25 @@ wasi_path_remove_directory(wasm_exec_env_t exec_env, wasi_fd_t fd,
     return wasmtime_ssp_path_remove_directory(curfds, fd, path, path_len);
 }
 
+static __wasi_timestamp_t
+get_timeout_for_poll_oneoff(const wasi_subscription_t *in)
+{
+    __wasi_timestamp_t timeout = 0;
+    const __wasi_subscription_t *clock_subscription = NULL;
+    if (clock_subscription == NULL
```
loganek, line 959: this condition is always true, since you set `clock_subscription` to `NULL` in the line above.

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -951,6 +951,25 @@ wasi_path_remove_directory(wasm_exec_env_t exec_env, wasi_fd_t fd,
...
+        && (in->u.u.clock.flags & __WASI_SUBSCRIPTION_CLOCK_ABSTIME) == 0) {
+        clock_subscription = in;
+    }
+    if (clock_subscription != NULL) {
+        __wasi_timestamp_t ts = clock_subscription->u.u.clock.timeout / 1e9;
+        timeout = ts > INT_MAX ? -1 : (int)ts;
+    }
+    else {
+        timeout = -1;
```
loganek, line 968: should we have an explicit cast here, since `timeout` is unsigned?

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -951,6 +951,25 @@ wasi_path_remove_directory(wasm_exec_env_t exec_env, wasi_fd_t fd,
...
+get_timeout_for_poll_oneoff(const wasi_subscription_t *in)
+{
+    __wasi_timestamp_t timeout = 0;
+    const __wasi_subscription_t *clock_subscription = NULL;
+    if (clock_subscription == NULL
+        && (in->u.u.clock.flags & __WASI_SUBSCRIPTION_CLOCK_ABSTIME) == 0) {
+        clock_subscription = in;
+    }
+    if (clock_subscription != NULL) {
+        __wasi_timestamp_t ts = clock_subscription->u.u.clock.timeout / 1e9;
```
loganek, line 964: So this function returns value in seconds? But then here: https://github.com/bytecodealliance/wasm-micro-runtime/pull/1951/files#diff-18a1c12794da572224835cdd9491ff7f93bc7ff81249c75ead3052e7f815f9f8R1006 we compare it to `elapsed` which is in `time_quant` unit - luckily, it's exactly a second now, but we might accidentally change the quant to be e.g. 0.5s and forget updating other places. My recommendation would be to stick to nanoseconds like we do everywhere unless we have a good reason not to do that.

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -969,9 +988,36 @@ wasi_poll_oneoff(wasm_exec_env_t exec_env, const wasi_subscription_t *in,
...
+        return (wasi_errno_t)-1;
+    }
+    bh_memcpy_s(in_copy, size_to_copy, in, size_to_copy);
+
+    while (1) {
+        elapsed += time_quant;
+        if ((int)timeout != -1 && elapsed > timeout)
+            break;
+
+        in_copy[0].u.u.clock.timeout = time_quant * 1e9;
```
loganek, line 1009: What if zero-th subscription is not a clock subscription (but it's Read or Write)?

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -969,9 +988,36 @@ wasi_poll_oneoff(wasm_exec_env_t exec_env, const wasi_subscription_t *in,
...
-    if (err)
-        return err;
+    __wasi_timestamp_t elapsed = 0;
+    /* in seconds */
+    const __wasi_timestamp_t timeout = get_timeout_for_poll_oneoff(in),
+                             time_quant = 1;
+    const uint32 size_to_copy = nsubscriptions * sizeof(wasi_subscription_t);
+    __wasi_subscription_t *in_copy;
+    if (size_to_copy
+        && !(in_copy =
```
loganek, line 998: Where do we free `in_copy`?

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
hritikgupta: yes, that's exactly why I had kept `timestamp` as int earlier, but now it it seems we're good to get rid of the division given the datatype is changed to `__wasi_timestamp_t`

core/iwasm/common/wasm_shared_memory.c
hritikgupta: done

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
hritikgupta: done

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
hritikgupta: iterating on subscriptions and updated timeout of all clock subscriptions only instead

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
hritikgupta: done

core/iwasm/common/wasm_shared_memory.c
hritikgupta: done

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -951,6 +951,43 @@ wasi_path_remove_directory(wasm_exec_env_t exec_env, wasi_fd_t fd,
...
+        const __wasi_subscription_t *s = &in[i];
+        if (s->u.type == __WASI_EVENTTYPE_CLOCK
+            && (s->u.u.clock.flags & __WASI_SUBSCRIPTION_CLOCK_ABSTIME) == 0) {
+            clock_subscription = s;
+            break;
+        }
+    }
+    if (clock_subscription != NULL) {
+        __wasi_timestamp_t ts = clock_subscription->u.u.clock.timeout;
+        timeout = ts >= UINT64_MAX ? (__wasi_timestamp_t)-1 : ts;
```
wenyongh, line 971: Seems we could change it to `timeout = ts` directly?
__wasi_timestamp_t is defined as uint64_t:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/include/wasmtime_ssp.h#L302
 when `ts >= UINT64_MAX`, ts must be `UINT64_MAX`, same as `(__wasi_timestamp_t)-1`.

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -969,9 +1006,41 @@ wasi_poll_oneoff(wasm_exec_env_t exec_env, const wasi_subscription_t *in,
...
         return (wasi_errno_t)-1;
 
-    err = wasmtime_ssp_poll_oneoff(curfds, in, out, nsubscriptions, &nevents);
-    if (err)
-        return err;
+    __wasi_timestamp_t elapsed = 0;
+    const __wasi_timestamp_t timeout = get_timeout_for_poll_oneoff(
+                                 in, nsubscriptions),
+                             time_quant = 1e9;
+    const uint32 size_to_copy = nsubscriptions * sizeof(wasi_subscription_t);
```
wenyongh, line 1013: `uint32` to `uint64`,
had better
`const uint64 size_to_copy = nsubscriptions * (uint64)sizeof(wasi_subscription_t);` to avoid integer overflow.

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -969,9 +1006,41 @@ wasi_poll_oneoff(wasm_exec_env_t exec_env, const wasi_subscription_t *in,
...
+    __wasi_subscription_t *in_copy;
+    if (size_to_copy >= UINT32_MAX
+        || (size_to_copy
+            && !(in_copy = (__wasi_subscription_t *)wasm_runtime_malloc(
+                     size_to_copy)))) {
```
wenyongh, line 1018: wasm_runtime_malloc(
                     (uint32)size_to_copy)

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -118,8 +118,13 @@ notify_stale_threads_on_exception(WASMModuleInstanceCommon *module_inst)
...
                          (void *)&total_elem_count);
 
+    if (!total_elem_count)
+        return;
+
     /* allocate memory */
-    if (!(args.addr = wasm_runtime_malloc(sizeof(void *) * total_elem_count))) {
+    const size_t total_elem_count_size = sizeof(void *) * total_elem_count;
+    if (total_elem_count_size >= UINT32_MAX
+        || !(args.addr = wasm_runtime_malloc(total_elem_count_size))) {
```
wenyongh, line 127: wasm_runtime_malloc((uint32)total_elem_count_size)

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -969,9 +1006,41 @@ wasi_poll_oneoff(wasm_exec_env_t exec_env, const wasi_subscription_t *in,
...
+        return (wasi_errno_t)-1;
+    }
+    bh_memcpy_s(in_copy, size_to_copy, in, size_to_copy);
+
+    while (1) {
+        elapsed += time_quant;
+        if ((int)timeout != -1 && elapsed > timeout)
```
wenyongh, line 1025: Not sure why `(int)timeout != -1`, should be `timeout !=(__wasi_timestamp_t)-1`?

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -969,9 +1006,41 @@ wasi_poll_oneoff(wasm_exec_env_t exec_env, const wasi_subscription_t *in,
...
 
-    err = wasmtime_ssp_poll_oneoff(curfds, in, out, nsubscriptions, &nevents);
-    if (err)
-        return err;
+    __wasi_timestamp_t elapsed = 0;
+    const __wasi_timestamp_t timeout = get_timeout_for_poll_oneoff(
+                                 in, nsubscriptions),
+                             time_quant = 1e9;
+    const uint32 size_to_copy = nsubscriptions * sizeof(wasi_subscription_t);
+    __wasi_subscription_t *in_copy;
```
wenyongh, line 1014: Should initialize it as NULL

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -969,9 +1006,41 @@ wasi_poll_oneoff(wasm_exec_env_t exec_env, const wasi_subscription_t *in,
...
+        elapsed += time_quant;
+        if ((int)timeout != -1 && elapsed > timeout)
+            break;
+
+        /* update timeout for clock subscription events */
+        update_clock_subscription_data(&in_copy, nsubscriptions, time_quant);
+        err = wasmtime_ssp_poll_oneoff(curfds, in_copy, out, nsubscriptions,
+                                       &nevents);
+        if (err) {
+            wasm_runtime_free(in_copy);
```
wenyongh, line 1033:
```C
if (in_copy)
    wasm_runtime_free(in_copy);
```

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -969,9 +1006,41 @@ wasi_poll_oneoff(wasm_exec_env_t exec_env, const wasi_subscription_t *in,
...
+            return err;
+        }
+
+        if (wasm_runtime_get_exception(module_inst) || nevents > 0) {
+            wasm_runtime_free(in_copy);
```
wenyongh, line 1038:
```C
if (in_copy)
    wasm_runtime_free(in_copy);
```

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -969,9 +1006,41 @@ wasi_poll_oneoff(wasm_exec_env_t exec_env, const wasi_subscription_t *in,
...
+            if (nevents)
+                return 0;
```
wenyongh, line 1040: Should `*nevents_app = (uint32)nevents` before return 0?

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -969,9 +1006,41 @@ wasi_poll_oneoff(wasm_exec_env_t exec_env, const wasi_subscription_t *in,
...
+            return EINTR;
+        }
+    }
 
     *nevents_app = (uint32)nevents;
```
wenyongh, line 1045: Should free in_copy?
```C
if (in_copy)
    wasm_runtime_free(in_copy);
```

wenyongh: And shall we init nevents as 0 at the beginning? Is there a case that nevents wasn't set before and this func returns 0?

hritikgupta:
[wip]

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -111,15 +111,21 @@ notify_stale_threads_on_exception(WASMModuleInstanceCommon *module_inst)
...
+    uint64 total_elem_count_size = 0;
 
     os_mutex_lock(&shared_memory_list_lock);
 
     /* count number of addresses in wait_map */
     bh_hash_map_traverse(wait_map, wait_map_address_count_callback,
                          (void *)&total_elem_count);
 
+    if (!total_elem_count)
+        return;
```
loganek, line 123: should you unlock `shared_memory_list_lock` here?

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -951,6 +951,43 @@ wasi_path_remove_directory(wasm_exec_env_t exec_env, wasi_fd_t fd,
...
+    for (size_t i = 0; i < nsubscriptions; ++i) {
+        const __wasi_subscription_t *s = &in[i];
+        if (s->u.type == __WASI_EVENTTYPE_CLOCK
+            && (s->u.u.clock.flags & __WASI_SUBSCRIPTION_CLOCK_ABSTIME) == 0) {
+            clock_subscription = s;
+            break;
+        }
+    }
+    if (clock_subscription != NULL) {
+        __wasi_timestamp_t ts = clock_subscription->u.u.clock.timeout;
```
loganek, line 970:
```suggestion
        timeout = clock_subscription->u.u.clock.timeout;
```

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -951,6 +951,43 @@ wasi_path_remove_directory(wasm_exec_env_t exec_env, wasi_fd_t fd,
...
+        timeout = ts;
```
loganek, line 971:
```suggestion
        timeout = clock_subscription->u.u.clock.timeout;
```

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -969,9 +1006,49 @@ wasi_poll_oneoff(wasm_exec_env_t exec_env, const wasi_subscription_t *in,
...
+    __wasi_subscription_t *in_copy = NULL;
+    if (size_to_copy >= UINT32_MAX
+        || (size_to_copy
+            && !(in_copy = (__wasi_subscription_t *)wasm_runtime_malloc(
+                     (uint32)size_to_copy)))) {
+        return (wasi_errno_t)-1;
+    }
+    bh_memcpy_s(in_copy, size_to_copy, in, size_to_copy);
+
+    while (1) {
```
loganek, line 1023: This is just a personal opinion so feel free to ignore - I'm not a big fan of infinite loops (`while(1)`) and prefer to avoid them; instead, I prefer have a clear stop condition in the loop declaration. In this case, we could have:
```
while (timeout != (__wasi_timestamp_t)-1 && elapsed > timeout)
```

In fact, that change actually fixes the bug - currently, if the initial `timeout` is smaller than `time_quant`, we'll never execute `wasmtime_ssp_poll_oneoff`. I think we should always wait for the reminder of `timeout % time_quant`.

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -969,9 +1006,49 @@ wasi_poll_oneoff(wasm_exec_env_t exec_env, const wasi_subscription_t *in,
...
+        elapsed += time_quant;
+        if (timeout != (__wasi_timestamp_t)-1 && elapsed > timeout)
+            break;
+
+        /* update timeout for clock subscription events */
+        update_clock_subscription_data(in_copy, nsubscriptions, time_quant);
```
loganek, line 1029: This should not be `time_quant`; if the timeout e.g. is 2.5s, we should pass 1s for the first 2 iterations, but for the last iteration it should be 0.5s.

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
hritikgupta: but don't we want to loop infinitely if timeout is -1?

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
hritikgupta: good catch, we can instead pass `min(time_quant, timeout - elapsed)`

core/iwasm/common/wasm_shared_memory.c
hritikgupta: missed, done

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
hritikgupta: converted to ternary assignment instead

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
hritikgupta: i think it should instead be the ! of the above, changing it

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -969,9 +1007,48 @@ wasi_poll_oneoff(wasm_exec_env_t exec_env, const wasi_subscription_t *in,
...
         return (wasi_errno_t)-1;
 
-    err = wasmtime_ssp_poll_oneoff(curfds, in, out, nsubscriptions, &nevents);
-    if (err)
-        return err;
+    __wasi_timestamp_t elapsed = 0;
+    const __wasi_timestamp_t timeout = get_timeout_for_poll_oneoff(
+                                 in, nsubscriptions),
+                             time_quant = 1e9;
+    const uint64 size_to_copy = nsubscriptions * sizeof(wasi_subscription_t);
```
wenyongh, line 1014: Had better use `nsubscriptions * (uint64)sizeof(wasi_subscription_t)` to avoid possible integer overflow.

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -969,9 +1007,48 @@ wasi_poll_oneoff(wasm_exec_env_t exec_env, const wasi_subscription_t *in,
...
+    __wasi_subscription_t *in_copy = NULL;
+    if (size_to_copy >= UINT32_MAX
+        || (size_to_copy
+            && !(in_copy = (__wasi_subscription_t *)wasm_runtime_malloc(
+                     (uint32)size_to_copy)))) {
+        return (wasi_errno_t)-1;
+    }
+    bh_memcpy_s(in_copy, size_to_copy, in, size_to_copy);
```
wenyongh, line 1022: Code is not easy to understand, had better change to:
```C
if (size_to_copy > 0) {
    if (size_to_copy >= UINT32_MAX
        || !(in_copy = (__wasi_subscription_t *)wasm_runtime_malloc(
                     (uint32)size_to_copy))) {
        return (wasi_errno_t)-1;
    }
    bh_memcpy_s(in_copy, size_to_copy, in, size_to_copy);
}
```

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -969,9 +1007,48 @@ wasi_poll_oneoff(wasm_exec_env_t exec_env, const wasi_subscription_t *in,
...
+                return 0;
+            }
+            return EINTR;
+        }
+    }
+
+    if (in_copy)
+        wasm_runtime_free(in_copy);
 
     *nevents_app = (uint32)nevents;
```
wenyongh, line 1053: Could we ensure that `nevents` was set while setting it to `*nevents_app`? Had better initialize it to 0 while declaring it in L999? `size_t nevents = 0;`

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -969,9 +1007,48 @@ wasi_poll_oneoff(wasm_exec_env_t exec_env, const wasi_subscription_t *in,
         || !validate_native_addr(nevents_app, sizeof(uint32)))
         return (wasi_errno_t)-1;
 
-    err = wasmtime_ssp_poll_oneoff(curfds, in, out, nsubscriptions, &nevents);
-    if (err)
-        return err;
```
wenyongh, line 974: The changes of this PR are only for multi-threading mode, can we keep the original implementation for non multi-threading mode? Or, use WASM_ENABLE_THREAD_MGR to control the code:
```C
    err = wasmtime_ssp_poll_oneoff(curfds, in, out, nsubscriptions, &nevents);
    if (err)
        return err;
#else
...
#endif
```

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -63,6 +63,12 @@ typedef struct WASIContext {
 wasi_ctx_t
 wasm_runtime_get_wasi_ctx(wasm_module_inst_t module_inst);
 
+static inline size_t
+min(size_t a, size_t b)
```
loganek, line 67: we already have the same function defined in `posix.c`. Maybe worth moving it to `bh_common`?

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
loganek: or could we add at the beginning:
```c
if (nsubscriptions == 0) {
    return 0;
}
```

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
wenyongh: It is a simple sentence and the function is defined as inline, my suggestion is just keep it here.

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
wenyongh: That's good, in that case, here in_copy must be not NULL, and we don't need to check whether it is NULL when freeing it below.

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
hritikgupta: done

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
hritikgupta: done

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -951,14 +957,104 @@ wasi_path_remove_directory(wasm_exec_env_t exec_env, wasi_fd_t fd,
...
+{
+    __wasi_timestamp_t timeout = 0;
+    const __wasi_subscription_t *clock_subscription = NULL;
+
+    for (size_t i = 0; i < nsubscriptions; ++i) {
+        const __wasi_subscription_t *s = &in[i];
+        if (s->u.type == __WASI_EVENTTYPE_CLOCK
+            && (s->u.u.clock.flags & __WASI_SUBSCRIPTION_CLOCK_ABSTIME) == 0) {
+            clock_subscription = s;
+            break;
```
hritikgupta, line 972: @loganek / @wenyongh does it make sense to return the largest timeout amongst all clock subscriptions for final poll_oneoff timeout?

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -951,14 +957,106 @@ wasi_path_remove_directory(wasm_exec_env_t exec_env, wasi_fd_t fd,
...
+                *nevents_app = (uint32)(*nevents);
+                return 0;
+            }
+            return EINTR;
+        }
+    }
+
+    wasm_runtime_free(in_copy);
+    return 0;
+}
```
wenyongh, line 1050: Had better wrap the above code (L960 to L1050) with `#if WASM_ENABLE_THREAD_MGR != 0 ... #endif` to avoid compile warning when multi-thread is disabled.

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
wenyongh: It is really an issue, the timeout of each subscription may be different, not sure whether `poll_oneoff` will always return when the smallest timeout of all subscriptions is elapsed or not? If yes, I think here we should use the smallest timeout of all subscriptions.

It not, the algorithm is complex, here we should use the largest timeout, and record each subscription's left timeout (its original timeout - elapsed time), update them after `poll_oneoff` (L1030), and if the left timeout is < 0, remove it from the subscription list.

Seems the first one is reasonable, had better have a test for poll_oneoff under Linux to get the detail.

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
hritikgupta: yes, realised that later, else I'll have to add this new function declaration to the wasmtime_ssp.h file. done!

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
hritikgupta: iiuc, if there are multiple clock subscriptions and file descriptor (fd), [poll](https://linux.die.net/man/2/poll) waits until min timeout of all clock subscription has elapsed or when the fd is ready for I/O, whichever happens first. so I guess, it makes sense to take minima of all timeouts and return from this method (and use it in our while loop for timeout chaining).

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -951,14 +957,105 @@ wasi_path_remove_directory(wasm_exec_env_t exec_env, wasi_fd_t fd,
...
+
+    for (i = 0; i < nsubscriptions; ++i) {
+        const __wasi_subscription_t *s = &in[i];
+        if (s->u.type == __WASI_EVENTTYPE_CLOCK
+            && (s->u.u.clock.flags & __WASI_SUBSCRIPTION_CLOCK_ABSTIME) == 0) {
+            timeout = min(timeout, s->u.u.clock.timeout);
+        }
+    }
+    return timeout == UINT64_MAX ? (__wasi_timestamp_t)-1
+                                 : timeout; /* returns timeout in ns */
```
wenyongh, line 976: __wasi_timestamp_t is defined as uint64_t, `(__wasi_timestamp_t)-1` is also UINT64_MAX, seems we can just `return timeout;`

hritikgupta: done

wenyongh (commented):
LGTM

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -951,14 +957,104 @@ wasi_path_remove_directory(wasm_exec_env_t exec_env, wasi_fd_t fd,
...
+                                 in, nsubscriptions),
+                             time_quant = 1e9;
+    const uint64 size_to_copy =
+        nsubscriptions * (uint64)sizeof(wasi_subscription_t);
+    __wasi_subscription_t *in_copy = NULL;
+
+    if (size_to_copy >= UINT32_MAX
+        || !(in_copy = (__wasi_subscription_t *)wasm_runtime_malloc(
+                 (uint32)size_to_copy))) {
+        return (wasi_errno_t)-1;
```
loganek, line 1016: Probably better to return `__WASI_ENOMEM` here.

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -951,14 +957,104 @@ wasi_path_remove_directory(wasm_exec_env_t exec_env, wasi_fd_t fd,
...
+            if (*nevents) {
+                *nevents_app = (uint32)(*nevents);
+                return 0;
+            }
+            return EINTR;
+        }
+    }
+
+    wasm_runtime_free(in_copy);
+    return 0;
```
loganek, line 1046:
```suggestion
    return __WASI_ESUCCESS;
```

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -951,14 +957,104 @@ wasi_path_remove_directory(wasm_exec_env_t exec_env, wasi_fd_t fd,
     return wasmtime_ssp_path_remove_directory(curfds, fd, path, path_len);
 }
 
+#if WASM_ENABLE_THREAD_MGR != 0
+static __wasi_timestamp_t
+get_timeout_for_poll_oneoff(const wasi_subscription_t *in,
+                            uint32 nsubscriptions)
+{
+    __wasi_timestamp_t timeout = UINT64_MAX;
```
loganek, line 965: For consistency, I'd recommend either using `(__wasi_timestamp_t)-1` here or `UINT64_MAX` in the [L1021](https://github.com/bytecodealliance/wasm-micro-runtime/pull/1951/files#diff-18a1c12794da572224835cdd9491ff7f93bc7ff81249c75ead3052e7f815f9f8R1021)

loganek (approved):
LGTM, left some minor comments.

eloparco:
@wenyongh can we merge `main` into `dev/wasi_threads` once this PR is merged?

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
hritikgupta: done

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
hritikgupta: done, on a second thought, does it conflict with [L1042](https://github.com/bytecodealliance/wasm-micro-runtime/blob/70117ed41be1b6b5fa495965fb6280aee0977a11/core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c#L1042) which also returns 0 (or __WASI_ESUCCESS) ?

wenyongh:
> @wenyongh can we merge `main` into `dev/wasi_threads` once this PR is merged?

Sure, will do that soon.

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
wenyongh: Had better change `return 0` of L1001 and L1039 into `return __WASI_ESUCCESS`.

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -951,14 +957,104 @@ wasi_path_remove_directory(wasm_exec_env_t exec_env, wasi_fd_t fd,
...
+execute_interruptible_poll_oneoff(wasm_module_inst_t module_inst,
+#if !defined(WASMTIME_SSP_STATIC_CURFDS)
+                                  struct fd_table *curfds,
+#endif
+                                  const __wasi_subscription_t *in,
+                                  __wasi_event_t *out, size_t nsubscriptions,
+                                  size_t *nevents, uint32 *nevents_app)
+{
+    if (nsubscriptions == 0)
+        return 0;
```
wenyongh, line 1001: Had better set `*nevents_app = 0` before return.

hritikgupta: wouldn't that anyway happen on L1077?

wenyongh: Oh, that's really an issue, not sure why the prototype of execute_interruptible_poll_oneoff isn't same as wasmtime_ssp_poll_oneoff. My understanding is that we should define them as same prototype.

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -951,14 +957,104 @@ wasi_path_remove_directory(wasm_exec_env_t exec_env, wasi_fd_t fd,
...
+}
+
+static wasi_errno_t
+execute_interruptible_poll_oneoff(wasm_module_inst_t module_inst,
+#if !defined(WASMTIME_SSP_STATIC_CURFDS)
+                                  struct fd_table *curfds,
+#endif
+                                  const __wasi_subscription_t *in,
+                                  __wasi_event_t *out, size_t nsubscriptions,
+                                  size_t *nevents, uint32 *nevents_app)
```
wenyongh, line 998: Had better remove argument `uint32 *nevents_app`

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
hritikgupta: it's mostly the same (that's what my intent was too) except for one argument i.e. `module_inst` which needs to be passed to check if there are any exceptions that occurred

wenyongh: Yes, adding module_inst is OK, but how about removing nevents_app? It is duplicated set inside execute_interruptible_poll_oneoff and after calling it.

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
hritikgupta: makes sense, as we are anyway setting it post this fn returns, done!

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -951,14 +957,104 @@ wasi_path_remove_directory(wasm_exec_env_t exec_env, wasi_fd_t fd,
...
+{
+    if (nsubscriptions == 0)
+        return __WASI_ESUCCESS;
```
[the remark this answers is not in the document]
hritikgupta: done

wenyongh (approved):
LGTM

loganek (approved):

PR #1952, opened 2023-02-12
Fix deadlock when destroying exec env in thread manager

eloparco:
Fix deadlock in case of error while creating a new thread in the thread manager.

The function `wasm_exec_env_destroy()` calls `wasm_cluster_terminate_all_except_self()` that tries to acquire the lock `cluster->lock`, but, in case of error while creating a new thread, that lock was already acquired before.

### DIFF
--- a/core/iwasm/libraries/thread-mgr/thread_manager.c
+++ b/core/iwasm/libraries/thread-mgr/thread_manager.c
@@ -425,6 +425,7 @@ wasm_cluster_spawn_exec_env(WASMExecEnv *exec_env)
     os_mutex_lock(&cluster->lock);
 
     if (cluster->has_exception || cluster->processing) {
+        os_mutex_unlock(&cluster->lock);
         goto fail1;
     }
 
@@ -444,6 +445,7 @@ wasm_cluster_spawn_exec_env(WASMExecEnv *exec_env)
 
     if (!(new_module_inst = wasm_runtime_instantiate_internal(
               module, true, stack_size, 0, NULL, 0))) {
+        os_mutex_unlock(&cluster->lock);
         goto fail1;
     }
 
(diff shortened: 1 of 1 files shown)


eloparco:
I run into the error while trying to created more threads than `CLUSTER_MAX_THREAD_NUM`. In that case, we terminate all the threads and exit. Is that the expected behavior? Why don't report the error and continue with the execution of the other threads?
I can push a fix in case we decide to go with the second option.

loganek:
> I run into the error while trying to created more threads than CLUSTER_MAX_THREAD_NUM. In that case, we terminate all the threads and exit.  Is that the expected behavior? 

IMO that's not an expected behavior. I think we should probably call `wasm_exec_env_destroy_internal()` instead, but I'd let @xujuntwt95329 comment on that.

@xujuntwt95329 was the behavior intentional or should we rather not kill other threads when spawning a new thread fails or is the current behavior (assuming the deadlock is fixed) correct?

xujuntwt95329:
> > I run into the error while trying to created more threads than CLUSTER_MAX_THREAD_NUM. In that case, we terminate all the threads and exit.  Is that the expected behavior?
> 
> IMO that's not an expected behavior. I think we should probably call `wasm_exec_env_destroy_internal()` instead, but I'd let @xujuntwt95329 comment on that.
> 
> @xujuntwt95329 was the behavior intentional or should we rather not kill other threads when spawning a new thread fails or is the current behavior (assuming the deadlock is fixed) correct?

Well, I agree with you!

We shouldn't call `wasm_exec_env_destroy` when thread creation failed (I forgot to modify this when we introduce `wasm_exec_env_destroy_internal`). 
Using `wasm_exec_env_destroy_internal` should get the expected behavior back, and deadlock issue is avoid.

xujuntwt95329:
@eloparco I've submitted a PR https://github.com/bytecodealliance/wasm-micro-runtime/pull/1955 to recover the expected behavior, could you please check if it recover the expected behaviour?

eloparco:
I was planning on updating this PR depending on the result of the discussion, but no problem. I had already tested it after replacing `wasm_exec_env_destroy` with `wasm_exec_env_destroy_internal` and it works correctly.

eloparco:
Closing after the fix in #1955

PR #1960, opened 2023-02-14
Use shared memory lock for threads generated from same module

### LINKED ISSUE #1958: Atomic add result is not correct with multiple threads
I was running some tests and I found this problem after adding the following sample in `samples/multi-thread/wasm-apps/main_global_var.c`.

```c
/*
 * Copyright (C) 2023 Amazon.com Inc. or its affiliates. All rights reserved.
 */

#include <stdio.h>
#include <pthread.h>
#include <assert.h>

enum CONSTANTS {
    MAX_NUM_THREADS = 4,
    NUM_ITER = 1000,
};

#define USE_MUTEX 0

int g_count = 0;
pthread_mutex_t mutex;

static void *
thread(void *arg)
{
    for (int i = 0; i < NUM_ITER; i++) {
#if USE_MUTEX == 0
        __atomic_fetch_add(&g_count, 1, __ATOMIC_SEQ_CST);
#else
        pthread_mutex_lock(&mutex);
        g_count++;
        pthread_mutex_unlock(&mutex);
#endif
    }

    return NULL;
}

int
main(int argc, char **argv)
{
    pthread_t tids[MAX_NUM_THREADS];

    if (pthread_mutex_init(&mutex, NULL) != 0) {
        printf("Failed to init mutex.\n");
        return -1;
    }

    for (int i = 0; i < MAX_NUM_THREADS; i++) {
        if (pthread_create(&tids[i], NULL, thread, NULL) != 0) {
            printf("Thread creation failed\n");
        }
    }

    for (int i = 0; i < MAX_NUM_THREADS; i++) {
        if (pthread_join(tids[i], NULL) != 0) {
            printf("Thread join failed\n");
        }
    }

    printf("Value of count after update: %d (expected=%d)\n", g_count,
           MAX_NUM_THREADS * NUM_ITER);

    return 0;
}
```

The final result is not correct when `__atomic_fetch_add()` is used. I see that the code in `WASM_OP_ATOMIC_RMW_I32_ADD` is executed but the final result is wrong:

```bash
Value of count after update: 2651 (expected=4000)
```

eloparco:
Multiple threads should use the same lock to protect atomic operations. Fix #1958

Before this PR, each thread was using a different lock to protect atomic operations (e.g. atomic add), making the lock ineffective.

### DIFF
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -1111,21 +1111,32 @@ wasm_runtime_load(uint8 *buf, uint32 size, char *error_buf,
                   uint32 error_buf_size)
 {
     WASMModuleCommon *module_common = NULL;
+    WASMModule *module = NULL;
 
     if (get_package_type(buf, size) == Wasm_Module_Bytecode) {
 #if WASM_ENABLE_INTERP != 0
         module_common =
             (WASMModuleCommon *)wasm_load(buf, size, error_buf, error_buf_size);
-        return register_module_with_null_name(module_common, error_buf,
-                                              error_buf_size);
+        module_common = register_module_with_null_name(module_common, error_buf,
+                                                       error_buf_size);
+
+        if (module_common == NULL) {
+            return NULL;
+        }
+        goto success;
 #endif
     }
     else if (get_package_type(buf, size) == Wasm_Module_AoT) {
 #if WASM_ENABLE_AOT != 0
         module_common = (WASMModuleCommon *)aot_load_from_aot_file(
             buf, size, error_buf, error_buf_size);
-        return register_module_with_null_name(module_common, error_buf,
-                                              error_buf_size);
+        module_common = register_module_with_null_name(module_common, error_buf,
+                                                       error_buf_size);
+
+        if (module_common == NULL) {
+            return NULL;
+        }
+        goto success;
 #endif
     }
 
@@ -1136,6 +1147,32 @@ wasm_runtime_load(uint8 *buf, uint32 size, char *error_buf,
         set_error_buf(error_buf, error_buf_size,
                       "WASM module load failed: magic header not detected");
     return NULL;
+
+success:
+#if WASM_ENABLE_SHARED_MEMORY != 0
+    module = (WASMModule *)module_common;
+    if (os_mutex_init(&module->mem_lock) != 0) {
+        set_error_buf(error_buf, error_buf_size,
+                      "WASM module load failed: unable to init memory mutex");
+
+#if WASM_ENABLE_INTERP != 0
+        if (module->module_type == Wasm_Module_Bytecode) {
+            wasm_unload((WASMModule *)module);
+            return NULL;
+        }
+#endif
+
+#if WASM_ENABLE_AOT != 0
+        if (module->module_type == Wasm_Module_AoT) {
+            aot_unload((AOTModule *)module);
+            return NULL;
+        }
+#endif
+        return NULL;
+    }
+#endif /* end of WASM_ENABLE_SHARED_MEMORY != 0 */
+
+    return module_common;
 }
 
 WASMModuleCommon *
@@ -1171,6 +1208,13 @@ wasm_runtime_load_from_sections(WASMSection *section_list, bool is_aot,
 void
 wasm_runtime_unload(WASMModuleCommon *module)
 {
+#if WASM_ENABLE_SHARED_MEMORY != 0
+    WASMModule *wasm_module = (WASMModule *)module;
+    if (os_mutex_destroy(&wasm_module->mem_lock) != 0) {
+        LOG_ERROR("WASM module load failed: unable to destroy memory mutex");
+    }
+#endif
+
 #if WASM_ENABLE_MULTI_MODULE != 0
     /**
      * since we will unload and free all module when runtime_destroy()
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -1111,20 +1111,21 @@ wasm_runtime_load(uint8 *buf, uint32 size, char *error_buf,
                   uint32 error_buf_size)
 {
     WASMModuleCommon *module_common = NULL;
-    WASMModule *module = NULL;
 
     if (get_package_type(buf, size) == Wasm_Module_Bytecode) {
 #if WASM_ENABLE_INTERP != 0
         module_common =
             (WASMModuleCommon *)wasm_load(buf, size, error_buf, error_buf_size);
-        goto register_module;
+        return register_module_with_null_name(module_common, error_buf,
+                                              error_buf_size);
 #endif
     }
     else if (get_package_type(buf, size) == Wasm_Module_AoT) {
 #if WASM_ENABLE_AOT != 0
         module_common = (WASMModuleCommon *)aot_load_from_aot_file(
             buf, size, error_buf, error_buf_size);
-        goto register_module;
+        return register_module_with_null_name(module_common, error_buf,
+                                              error_buf_size);
 #endif
     }
 
@@ -1135,38 +1136,6 @@ wasm_runtime_load(uint8 *buf, uint32 size, char *error_buf,
         set_error_buf(error_buf, error_buf_size,
                       "WASM module load failed: magic header not detected");
     return NULL;
-
-register_module:
-    module_common = register_module_with_null_name(module_common, error_buf,
-                                                   error_buf_size);
-    if (module_common == NULL) {
-        return NULL;
-    }
-
-#if WASM_ENABLE_SHARED_MEMORY != 0
-    module = (WASMModule *)module_common;
-    if (os_mutex_init(&module->mem_lock) != 0) {
-        set_error_buf(error_buf, error_buf_size,
-                      "WASM module load failed: unable to init memory mutex");
-
-#if WASM_ENABLE_INTERP != 0
-        if (module->module_type == Wasm_Module_Bytecode) {
-            wasm_unload((WASMModule *)module);
-            return NULL;
-        }
-#endif
-
-#if WASM_ENABLE_AOT != 0
-        if (module->module_type == Wasm_Module_AoT) {
-            aot_unload((AOTModule *)module);
-            return NULL;
-        }
-#endif
-        return NULL;
-    }
-#endif /* end of WASM_ENABLE_SHARED_MEMORY != 0 */
-
-    return module_common;
 }
 
 WASMModuleCommon *
@@ -1202,13 +1171,6 @@ wasm_runtime_load_from_sections(WASMSection *section_list, bool is_aot,
 void
 wasm_runtime_unload(WASMModuleCommon *module)
 {
-#if WASM_ENABLE_SHARED_MEMORY != 0
-    WASMModule *wasm_module = (WASMModule *)module;
-    if (os_mutex_destroy(&wasm_module->mem_lock) != 0) {
-        LOG_ERROR("WASM module load failed: unable to destroy memory mutex");
-    }
-#endif
-
 #if WASM_ENABLE_MULTI_MODULE != 0
     /**
      * since we will unload and free all module when runtime_destroy()
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -1117,26 +1117,14 @@ wasm_runtime_load(uint8 *buf, uint32 size, char *error_buf,
 #if WASM_ENABLE_INTERP != 0
         module_common =
             (WASMModuleCommon *)wasm_load(buf, size, error_buf, error_buf_size);
-        module_common = register_module_with_null_name(module_common, error_buf,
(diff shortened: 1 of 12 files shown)


eloparco:
#1958

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1111,21 +1111,32 @@ wasm_runtime_load(uint8 *buf, uint32 size, char *error_buf,
...
+        goto success;
 #endif
     }
     else if (get_package_type(buf, size) == Wasm_Module_AoT) {
 #if WASM_ENABLE_AOT != 0
         module_common = (WASMModuleCommon *)aot_load_from_aot_file(
             buf, size, error_buf, error_buf_size);
-        return register_module_with_null_name(module_common, error_buf,
-                                              error_buf_size);
+        module_common = register_module_with_null_name(module_common, error_buf,
```
loganek, line 1133: I presume the registration could be moved down to a common code?

eloparco: good point, I didn't notice it

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1136,6 +1135,38 @@ wasm_runtime_load(uint8 *buf, uint32 size, char *error_buf,
...
+                                                   error_buf_size);
+    if (module_common == NULL) {
+        return NULL;
+    }
+
+#if WASM_ENABLE_SHARED_MEMORY != 0
+    module = (WASMModule *)module_common;
+    if (os_mutex_init(&module->mem_lock) != 0) {
+        set_error_buf(error_buf, error_buf_size,
+                      "WASM module load failed: unable to init memory mutex");
```
wenyongh, line 1150: Don't initialize `module->mem_lock` here, since the `module_common` here may be aot module (AOTModule), which hasn't field mem_lock, mem_lock is only used in interpreter mode.
Had better init it in `create_module` in wasm_loader.c/wasm_mini_loader.c instead:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_loader.c#L3877
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_mini_loader.c#L2736
And destroy it in wasm_loader_unload:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_loader.c#L4249
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_mini_loader.c#L2942

But another solution is to reuse the main instance's mem_lock for child instance while instantiating the child instance in thread creating, similar to the wasi_ctx (note the `is_sub_inst` flag, false for main instance, true means child instance created by new thread):
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_runtime.c#L1954-L1966
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_runtime.c#L2127

My personal preference is to use the latter method, had better isolate the instance data as long as we can, so two different instances (created by wasm_runtime_instantiate but not pthread_create callback) won't share the same mem_lock.

eloparco: Thanks for the review. For the second approach I need to check, the problem is that (at least for WASI threads) each thread uses a separate module instance and the lock should be shared.

eloparco: I went for the first approach.
PS: I'm working on a separate PR to fix some other race conditions I found while testing.

core/iwasm/interpreter/wasm_mini_loader.c
```diff
@@ -2734,6 +2734,13 @@ create_module(char *error_buf, uint32 error_buf_size)
     }
 #endif
 
+#if WASM_ENABLE_SHARED_MEMORY != 0
+    if (os_mutex_init(&module->mem_lock) != 0) {
+        set_error_buf(error_buf, error_buf_size, "init memory lock failed");
+        return NULL;
```
wenyongh, line 2740: Should destroy resources before return NULL:
```C
#if WASM_ENABLE_FAST_JIT != 0 && WASM_ENABLE_JIT != 0 && WASM_ENABLE_LAZY_JIT != 0
        os_mutex_destroy(&module->instance_list_lock);
#endif
        wasm_runtime_free(module);
```

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -3875,6 +3875,13 @@ create_module(char *error_buf, uint32 error_buf_size)
     }
 #endif
 
+#if WASM_ENABLE_SHARED_MEMORY != 0
+    if (os_mutex_init(&module->mem_lock) != 0) {
+        set_error_buf(error_buf, error_buf_size, "init memory lock failed");
+        return NULL;
```
wenyongh, line 3881: Should destroy resources before `return NULL`:
```C
#if WASM_ENABLE_DEBUG_INTERP != 0                    \
    || (WASM_ENABLE_FAST_JIT != 0 && WASM_ENABLE_JIT != 0 \
        && WASM_ENABLE_LAZY_JIT != 0)
        os_mutex_destroy(&module->instance_list_lock);
#endif
        wasm_runtime_free(module);
```

wenyongh: And found that L3868 has typo, could you help to modify:
`WASM_ENABLE_FAST_JIT != 0 && WASM_ENABLE_JIT`
to
`WASM_ENABLE_FAST_JIT != 0 && WASM_ENABLE_JIT != 0`

core/iwasm/interpreter/wasm_mini_loader.c
wenyongh: Found that L2728 has typo, could you help to modify:
WASM_ENABLE_FAST_JIT != 0 && WASM_ENABLE_JIT
to
WASM_ENABLE_FAST_JIT != 0 && WASM_ENABLE_JIT != 0

core/iwasm/common/wasm_runtime_common.c
xujuntwt95329: I agree with @wenyongh's second approach. The module may be instantiated multiple times (not a thread of main instance, directly instantiate from a module), then these instances are independent and should not rely on the same lock.
If the lock is defined in module, then atomic operation in one instance will influence another independence instance.

xujuntwt95329: I checked the code, seems in current implementation the shared memory is shared globally (all instances from all clusters share the same memory). But I checked the proposal again, seems it should be better to only share the memory inside a cluster.
If we want to modify the sharing scope to cluster later, had better not put the lock inside module

xujuntwt95329: I opened an issue to track the issue about memory is shared globally: https://github.com/bytecodealliance/wasm-micro-runtime/issues/1962

eloparco: I understand what you say, if a module is instantiated multiple times, the lock (and with the current implementation the memory too https://github.com/bytecodealliance/wasm-micro-runtime/blob/1c17665f68e118cd9c6bc9b8782f1d0cda5994c9/core/iwasm/interpreter/wasm_runtime.h#L265) would be shared as well. That's good for threads but not in the general case.

I'll see if I can modify this PR to handle that.

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -4253,7 +4268,8 @@ wasm_loader_unload(WASMModule *module)
     if (!module)
         return;
 
-#if WASM_ENABLE_FAST_JIT != 0 && WASM_ENABLE_JIT && WASM_ENABLE_LAZY_JIT != 0
+#if WASM_ENABLE_FAST_JIT != 0 && WASM_ENABLE_JIT != 0 \
```
wenyongh, line 4271: Could you also change L3868?
```C
#if WASM_ENABLE_DEBUG_INTERP != 0                    \
    || (WASM_ENABLE_FAST_JIT != 0 && WASM_ENABLE_JIT \
        && WASM_ENABLE_LAZY_JIT != 0)
```

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -4425,6 +4442,11 @@ wasm_loader_unload(WASMModule *module)
     }
 #endif
 
+#if WASM_ENABLE_SHARED_MEMORY != 0
+    WASMModule *wasm_module = (WASMModule *)module;
+    os_mutex_destroy(&wasm_module->mem_lock);
```
wenyongh, line 4447: Had better not declare variable here, here we can just `os_mutex_destroy(&module->mem_lock);`

core/iwasm/interpreter/wasm_mini_loader.c
```diff
@@ -3079,6 +3097,11 @@ wasm_loader_unload(WASMModule *module)
     }
 #endif
 
+#if WASM_ENABLE_SHARED_MEMORY != 0
+    WASMModule *wasm_module = (WASMModule *)module;
+    os_mutex_destroy(&wasm_module->mem_lock);
```
wenyongh, line 3102: Same as above, `os_mutex_destroy(&module->mem_lock);`

core/iwasm/common/wasm_runtime_common.c
eloparco: Actually, what I can do in this PR is to just move the lock to `WASMMemoryInstance` instead of `WASMModule`. That should be compatible with what we want to do in #1962. What do you think?

core/iwasm/interpreter/wasm_loader.c
eloparco: I'll update it depending on the outcome of https://github.com/bytecodealliance/wasm-micro-runtime/pull/1960#discussion_r1106963947

core/iwasm/interpreter/wasm_loader.c
eloparco: I found a couple of other cases with the same issue, fixed them

core/iwasm/interpreter/wasm_runtime.h
```diff
@@ -112,6 +112,10 @@ struct WASMMemoryInstance {
     MemBound mem_bound_check_8bytes;
     MemBound mem_bound_check_16bytes;
 #endif
+
+#if WASM_ENABLE_SHARED_MEMORY != 0
+    korp_mutex lock;
+#endif
```
wenyongh, line 118: Putting a new field here will break the AOT ABI, or, the layout of WASM/AOT module instance will change and the AOT file compiled by old wamrc won't be able to run on new wamr runtime, see:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_runtime.h#L330-L338
The offset of global data and table instance will change.

Could we add the mem_lock to `WASMSharedMemory`:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/common/wasm_shared_memory.h#L31
Seems it is more reasonable.

eloparco: Right, I can add another lock there, in addition to the already existing one https://github.com/bytecodealliance/wasm-micro-runtime/blob/1c17665f68e118cd9c6bc9b8782f1d0cda5994c9/core/iwasm/common/wasm_shared_memory.h#L24

eloparco: Actually, maybe I can use the same; the already existing one is only used to protect the `ref_count` https://github.com/bytecodealliance/wasm-micro-runtime/blob/1c17665f68e118cd9c6bc9b8782f1d0cda5994c9/core/iwasm/common/wasm_shared_memory.h#L31

core/iwasm/common/wasm_runtime_common.c
xujuntwt95329: > Actually, what I can do in this PR is to just move the lock to `WASMMemoryInstance` instead of `WASMModule`. That should be compatible with what we want to do in #1962. What do you think? @xujuntwt95329

Yes, sounds reasonable~

eloparco: After the suggestion from @wenyongh, I put it in `WASMSharedMemory`. We should be good to merge.

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -696,28 +696,28 @@ trunc_f64_to_int(WASMModuleInstance *module, uint32 *frame_sp, float64 src_min,
             CHECK_BULK_MEMORY_OVERFLOW(addr + offset, 1, maddr);     \
             CHECK_ATOMIC_MEMORY_ACCESS();                            \
                                                                      \
-            os_mutex_lock(&module->e->mem_lock);                     \
+            os_mutex_lock(&node->atomic_lock);                       \
```
xujuntwt95329, line 699: How about rename to `shared_mem_node`?

eloparco: What about `shared_mem_lock`?

wenyongh (commented):
LGTM

xujuntwt95329:
LGTM

PR #1963, opened 2023-02-15
Add internal tests for WASI threads

eloparco:
Add internal tests for WASI threads. These tests are run in addition to the ones in the proposal https://github.com/WebAssembly/wasi-threads/tree/main/test/testsuite. 

The purpose is to test additional and more complex scenarios.


.github/workflows/compilation_on_android_ubuntu.yml
```diff
@@ -521,6 +548,10 @@ jobs:
         if: env.USE_LLVM == 'true' && steps.retrieve_llvm_libs.outputs.cache-hit != 'true'
         run: echo "::error::can not get prebuilt llvm libraries" && exit 1
 
+      - name: Build WASI thread tests
```
wenyongh, line 551: Running the test will download wasi-sdk, build and make wasi-libc, and run build.sh, which may take a lot of time, shall we add the test for some situations, for example, `$DEFAULT_TEST_OPTIONS` test_option for all running modes?

loganek: Why do we have to build tests? Tests are already compiled and published, so there's no need to rebuild them on CI

loganek: Can we instead use the wasi-sdk prerelease that already includes threading? https://github.com/WebAssembly/wasi-sdk/releases/tag/wasi-sdk-20%2Bthreads

tests/wamr-test-suites/test_wamr.sh
```diff
@@ -176,6 +176,10 @@ readonly CLASSIC_INTERP_COMPILE_FLAGS="\
     -DWAMR_BUILD_INTERP=1 -DWAMR_BUILD_FAST_INTERP=0 \
     -DWAMR_BUILD_JIT=0 -DWAMR_BUILD_AOT=0 \
     -DWAMR_BUILD_SPEC_TEST=1 \
+    -DWAMR_BUILD_LIBC_WASI=1 \
+    -DWAMR_BUILD_LIBC_BUILTIN=1 \
+    -DWAMR_BUILD_LIB_WASI_THREADS=1 \
+    -DWASI_SYSROOT=${WAMR_DIR}/core/deps/wasi-libc/sysroot \
```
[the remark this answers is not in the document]
eloparco: In #1985 I only added `-DWAMR_BUILD_LIB_WASI_THREADS=1`, do we also want to do the same there and only set it if a test option is provided (let's say `-w`)?

.github/workflows/compilation_on_android_ubuntu.yml
eloparco: > Running the test will download wasi-sdk, build and make wasi-libc, and run build.sh, which may take a lot of time, shall we add the test for some situations, for example, $WASI_TEST_OPTIONS test_option for all running modes?

Sure, this was just a draft, as you say it's only needed with `$WASI_TEST_OPTIONS`, I'll add the condition.

> Can we instead use the wasi-sdk prerelease that already includes threading? https://github.com/WebAssembly/wasi-sdk/releases/tag/wasi-sdk-20%2Bthreads

Yes, I saw that a few days ago. I was gonna wait for the release, do we want to use the pre-release?

loganek: I think we could use pre-release and move to release once it's available. Also, if we experience any problems with the pre-release, we can migrate to your current proposal.

eloparco: > I think we could use pre-release and move to release once it's available. Also, if we experience any problems with the pre-release, we can migrate to your current proposal.

Not sure if it makes more sense to do it in a separate PR.

tests/wamr-test-suites/test_wamr.sh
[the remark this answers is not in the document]
eloparco: In the end, only `-DWAMR_BUILD_LIB_WASI_THREADS=1` needs to be added. Do you think it's better to use a separate option (`-w`) to add it on demand?

.github/workflows/compilation_on_android_ubuntu.yml
eloparco: I think it would be better to do it in a separate PR, where we would update to `wasi-sdk-20` (pre-release) in all the CI jobs, not only the ones using WASI threads.

tests/wamr-test-suites/test_wamr.sh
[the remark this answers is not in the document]
eloparco: Yes we need it, otherwise the `-DWAMR_BUILD_LIB_WASI_THREADS=1` flag will compromize other tests (e.g. spec tests).

wenyongh:
CI reported error:
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/4324003969/jobs/7548386927#step:13:232
```C
Test nonmain_proc_exit_sleep failed
  [exit_code] 33 == 250
STDOUT:

STDERR:
double free or corruption (fasttop)
```
Any idea about that?

eloparco:
> Any idea about that?

I rebased the PR on top of `main`. The problem seems to be caused by the exec_env being freed by the thread after finishing but still used by the main thread for the join operation. Apart from that, there are some edge cases that still cause problems for the tests in this PR.
Instead, after #1995, WASi tests from the [proposal](https://github.com/WebAssembly/wasi-threads/tree/main/test/testsuite) run fine without race conditions.

For the tests in this PR, I mainly see two king of issues:
- some remaining race conditions generated in the thread manager and shared memory handling
- many other data races coming from wasm load/store operations, like 
https://github.com/bytecodealliance/wasm-micro-runtime/blob/e6a01847971747824e7ed6dd27d4d3b6bdbf9ece/core/iwasm/interpreter/wasm_interp_classic.c#L1840
in race condition with both
https://github.com/bytecodealliance/wasm-micro-runtime/blob/e6a01847971747824e7ed6dd27d4d3b6bdbf9ece/core/iwasm/interpreter/wasm_interp_classic.c#L3663 and
https://github.com/bytecodealliance/wasm-micro-runtime/blob/e6a01847971747824e7ed6dd27d4d3b6bdbf9ece/core/iwasm/interpreter/wasm_interp_classic.c#L2013
My guess is that they're caused by the internal implementation of some functions (like the pthread ones for sync primitives) in wasi-libc. For instance, I don't see any data race here https://github.com/eloparco/wasm-micro-runtime/blob/eloparco/test-wasi-threads/core/iwasm/libraries/lib-wasi-threads/test/global_atomic.c, but I see store/load data races in this other test https://github.com/eloparco/wasm-micro-runtime/blob/eloparco/test-wasi-threads/core/iwasm/libraries/lib-wasi-threads/test/global_lock.c, where the only difference is in the usage of a lock instead of atomic variable.

The investigation is still on.

wenyongh:
> > Any idea about that?
> 
> I rebased the PR on top of `main`. The problem seems to be caused by the exec_env being freed by the thread after finishing but still used by the main thread for the join operation. Apart from that, there are some edge cases that still cause problems for the tests in this PR. Instead, after #1995, WASi tests from the [proposal](https://github.com/WebAssembly/wasi-threads/tree/main/test/testsuite) run fine without race conditions.
> 

Yes, seems that the "double free or corruption (fasttop)" error had been reported by CI when running `nonmain_proc_exit_sleep` and `nonmain_trap_wait`, and I found it ever occurred when running `thread_termination.wasm` in `samples/wasi-threads`. But I manually tested them again and cannot reproduce the issues, instead, I found that `nonmain_proc_exit_wait` sometimes got hang.

> For the tests in this PR, I mainly see two kind of issues:
> 
> * some remaining race conditions generated in the thread manager and shared memory handling
> * many other data races coming from wasm load/store operations, like
>   https://github.com/bytecodealliance/wasm-micro-runtime/blob/e6a01847971747824e7ed6dd27d4d3b6bdbf9ece/core/iwasm/interpreter/wasm_interp_classic.c#L1840
>   
>   in race condition with both
>   https://github.com/bytecodealliance/wasm-micro-runtime/blob/e6a01847971747824e7ed6dd27d4d3b6bdbf9ece/core/iwasm/interpreter/wasm_interp_classic.c#L3663
>    and
>   https://github.com/bytecodealliance/wasm-micro-runtime/blob/e6a01847971747824e7ed6dd27d4d3b6bdbf9ece/core/iwasm/interpreter/wasm_interp_classic.c#L2013
>   
>   My guess is that they're caused by the internal implementation of some functions (like the pthread ones for sync primitives) in wasi-libc. For instance, I don't see any data race here https://github.com/eloparco/wasm-micro-runtime/blob/eloparco/test-wasi-threads/core/iwasm/libraries/lib-wasi-threads/test/global_atomic.c, but I see store/load data races in this other test https://github.com/eloparco/wasm-micro-runtime/blob/eloparco/test-wasi-threads/core/iwasm/libraries/lib-wasi-threads/test/global_lock.c, where the only difference is in the usage of a lock instead of atomic variable.
> 
> The investigation is still on.

OK, seems the first issue is caused by runtime and the second is caused by wasi-libc itself.

wenyongh:
> > Any idea about that?
> 
> I rebased the PR on top of `main`. The problem seems to be caused by the exec_env being freed by the thread after finishing but still used by the main thread for the join operation.

I found an issue in thread manager wasm_cluster_create_thread which may result in invalid join operation, and submitted PR #2004 to fix it. Maybe you can try whether it fixes some issues found?

g0djan:
Tried with the fix, for me it started harder to reproduce the failure but `main_proc_exit_wait` still fails the same with `double free or corruption (fasttop)` and `exit code: 134`

Core dump backtraces of the threads https://gist.github.com/g0djan/0a6910cb03f24f251a02bee2514689df

eloparco:
> OK, seems the first issue is caused by runtime and the second is caused by wasi-libc itself.

Exactly, my plan is to first fix the ones in the runtime and then, if the ones in wasi-libc still seem to create problem, manually re-implement sync primitives (mutexes and barriers) using webassembly atomic primitives. Otherwise, just ignore the wasi-libc sanitizer warnings.

After #1995, there are very few data races in the runtime (detected by sanitizers) and, together with failures and deadlocks, they require many runs to be reproduced.

I'll rebase on top of main and check what happens. Thanks for #2004.

eloparco:
> Maybe you can try whether it fixes some issues found?

It didn't fix the data race that I was looking at, I still see it between these two calls
https://github.com/bytecodealliance/wasm-micro-runtime/blob/04616d398d017002f3dffb3665abc3893ddec966/core/iwasm/common/wasm_memory.c#L218
https://github.com/bytecodealliance/wasm-micro-runtime/blob/04616d398d017002f3dffb3665abc3893ddec966/core/iwasm/libraries/thread-mgr/thread_manager.c#L870

In addition, from time to time, I still see (even before your PR) problems related to unlocking an unlocked mutex, usage of invalid mutexes and heap-use-after-free.

If you want to try to reproduce, add the thread sanitizer here https://github.com/bytecodealliance/wasm-micro-runtime/blob/04616d398d017002f3dffb3665abc3893ddec966/samples/wasi-threads/CMakeLists.txt#L79
```cmake
set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fsanitize=thread")
set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fsanitize=thread")
```
and then run the tests like
```bash
/wasm-micro-runtime/samples/wasi-threads/build $ ./iwasm -v=5 ../../../core/iwasm/libraries/lib-wasi-threads/test/nonmain_trap_wait.wasm; echo $?
```
Usually I have I have a script to run all the tests in sequence and I may have to run it multiple times to get useful warnings from the sanitizer.

I won't have much time in the next couple of weeks, so feel free to take a look if you want.

wenyongh:
> > Maybe you can try whether it fixes some issues found?
> 
> It didn't fix the data race that I was looking at, I still see it between these two calls
> 
> https://github.com/bytecodealliance/wasm-micro-runtime/blob/04616d398d017002f3dffb3665abc3893ddec966/core/iwasm/common/wasm_memory.c#L218
> 
> 
> https://github.com/bytecodealliance/wasm-micro-runtime/blob/04616d398d017002f3dffb3665abc3893ddec966/core/iwasm/libraries/thread-mgr/thread_manager.c#L870
> 
> In addition, from time to time, I still see (even before your PR) problems related to unlocking an unlocked mutex, usage of invalid mutexes and heap-use-after-free.
> 
> If you want to try to reproduce, add the thread sanitizer here
> 
> https://github.com/bytecodealliance/wasm-micro-runtime/blob/04616d398d017002f3dffb3665abc3893ddec966/samples/wasi-threads/CMakeLists.txt#L79
> 
> ```cmake
> set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fsanitize=thread")
> set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fsanitize=thread")
> ```
> 
> and then run the tests like
> 
> ```shell
> /wasm-micro-runtime/samples/wasi-threads/build $ ./iwasm -v=5 ../../../core/iwasm/libraries/lib-wasi-threads/test/nonmain_trap_wait.wasm; echo $?
> ```
> 
> Usually I have I have a script to run all the tests in sequence and I may have to run it multiple times to get useful warnings from the sanitizer.
> 
> I won't have much time in the next couple of weeks, so feel free to take a look if you want.

OK, thanks, I will try looking into it.

wenyongh:
@eloparco I found some multi-threading related issues and uploaded PR #2013 to fix them, I think it should be able to fix some of the issues you reported, could you help review it and have a try?
BTW: The PR will lead to compile error in Windows when thread manager is enabled, I will fix it later.

eloparco:
@Weining2019 Great! I just tried the changes, and they seem to fix the remaining race conditions I was seeing.
The only one that I spot a couple of times is with
https://github.com/bytecodealliance/wasm-micro-runtime/blob/7a3d2bfab6cfd63d518c0ac45aebf267838ab8e8/core/iwasm/common/wasm_runtime_common.c#L3147
https://github.com/bytecodealliance/wasm-micro-runtime/blob/7a3d2bfab6cfd63d518c0ac45aebf267838ab8e8/core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c#L1095
But that should be fairly easy to fix.

I tried many runs in classic interpreter mode, I still need to check with AOT.

wenyongh:
> @wenyongh Great! I just tried the changes, and they seem to fix the remaining race conditions I was seeing. The only one that I spot a couple of times is with
> 
> https://github.com/bytecodealliance/wasm-micro-runtime/blob/7a3d2bfab6cfd63d518c0ac45aebf267838ab8e8/core/iwasm/common/wasm_runtime_common.c#L3147
> 
> 
> https://github.com/bytecodealliance/wasm-micro-runtime/blob/7a3d2bfab6cfd63d518c0ac45aebf267838ab8e8/core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c#L1095
> 
> 
> But that should be fairly easy to fix.
> I tried many runs in classic interpreter mode, I still need to check with AOT.

OK, I have merged that PR, maybe you can rebase this PR with main branch, and if CIs run OK, I will merge it.

.github/workflows/compilation_on_android_ubuntu.yml
```diff
@@ -493,6 +497,31 @@ jobs:
...
+        run: |
+          cd /opt
+          sudo wget ${{ matrix.wasi_sdk_release }}
+          sudo tar -xzf wasi-sdk-*.tar.gz
+          sudo mv wasi-sdk-19.0 wasi-sdk
+
+      - name: build wasi-libc (needed for wasi-threads)
+        if: matrix.test_option == '$WASI_TEST_OPTIONS'
+        run: |
+          mkdir wasi-libc
```
loganek, line 511: Could we upgrade to wasi-sdk 20, so we don't need to build libc from git?

g0djan:
https://github.com/bytecodealliance/wasm-micro-runtime/pull/2016

Fixed a data race when deleting wait_info in wasm_shared_memory, spotted with main_proc_exit_wait.c from this PR

.github/workflows/compilation_on_android_ubuntu.yml
eloparco: Sure, I'm addressing that in a separate PR

eloparco:
> OK, I have merged that PR, maybe you can rebase this PR with main branch, and if CIs run OK, I will merge it.

Yes, let's do that so it's easier to use the tests without applying patches or cherry-picking. They seem to succeed now (I tried many runs) and the data races seem gone (apart from the ones related to load/store operations).
I tried only in classic interpreter mode for now, I'll have to check in AOT mode.

.github/workflows/compilation_on_android_ubuntu.yml
eloparco: Do you think I should update it everywhere in the CI or only for the parts where WASI threads are used?

wenyongh: Had better update here only since we claimed in the document that we support the wasi-sdk 19.0+, we had better use it by default in the CI files.
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/doc/build_wasm_app.md

wenyongh:
> > OK, I have merged that PR, maybe you can rebase this PR with main branch, and if CIs run OK, I will merge it.
> 
> Yes, let's do that so it's easier to use the tests without applying patches or cherry-picking. They seem to succeed now (I tried many runs) and the data races seem gone (apart from the ones related to load/store operations). I tried only in classic interpreter mode for now, I'll have to check in AOT mode.

OK, I merged it.
BTW, @loganek added a comment to suggest to use the wasi-sdk-20.0 instead in CI, please submit another PR to do it if you want.

wenyongh:
CI run failure:
```bash
./test_wamr.sh -m x86_32 -P $WASI_TEST_OPTIONS -t classic-interp
...
Test nonmain_proc_exit_sleep failed
  [exit_code] 33 == 0
STDOUT:

STDERR:
```

.github/workflows/compilation_on_android_ubuntu.yml
eloparco: I tried it here https://github.com/eloparco/wasm-micro-runtime/commit/006534cfe3b853068b1f67cecb82efab3170229d. It works on MacOS but not on Ubuntu 20.04, as we observed in the past (because of the glibc version installed there).

https://github.com/eloparco/wasm-micro-runtime/actions/runs/4379797818/jobs/7666263974
https://github.com/eloparco/wasm-micro-runtime/actions/runs/4379796629/jobs/7666151109

We can use wasi-sdk-20 pre-release in case of Ubuntu 22.04 and keep using wasi-libc for wasi-sdk-19 on Ubuntu 20.04, making sure that in case of wasi-sdk-19 we specify `WASI_SYSROOT`. What do you think?

Maybe I should create a PR anyway and we can discuss there.

eloparco:
> CI run failure:

Locally, in classic interpreter mode, I tried multiple runs of `test_wamr.sh` and it was always succeeding. I tried multiple runs of `nonmain_proc_exit_sleep` with the thread sanitizer but I couldn't find anything (apart from the usual race conditions in load/store operations).

I didn't try with `x86_32` though since I don't have the necessary packages installed, do you think the problem could be `x86_32`-specific?

.github/workflows/compilation_on_android_ubuntu.yml
wenyongh: It's good, at least we don't need to build wasi-libc in Ubuntu 22.04. And we can update CI again when the wasi-sdk-20 is actually released. Not sure whether the glibc version issue on Ubuntu 20.04 will be resolved in wasi-sdk-20.0? Or should we file an issue in wasi-sdk community, or upload a PR to help resolve the issue?

wenyongh:
> Locally, in classic interpreter mode, I tried multiple runs of `test_wamr.sh` and it was always succeeding. I tried multiple runs of `nonmain_proc_exit_sleep` with the thread sanitizer but I couldn't find anything (apart from the usual race conditions in load/store operations).
> 
> I didn't try with `x86_32` though since I don't have the necessary packages installed, do you think the problem could be `x86_32`-specific?

I am not sure, I tested it many times with both iwasm x86_64 and x86_32, but also cannot reproduce the issue.

But for `update_shared_data_and_alloc_heap.wasm` in `core/iwasm/libraries/lib-wasi-threads/test`, seems it is only related to x86_32 when calling `malloc` function:
```bash
Failed to spawn a new thread
Exception: out of bounds memory access
```
I didn't reproduce it with iwasm x86_64, but it easily occurs with iwasm x86_32.

eloparco:
> But for update_shared_data_and_alloc_heap.wasm in core/iwasm/libraries/lib-wasi-threads/test, seems it is only related to x86_32 when calling malloc function:

Ok, that may be caused by the fact that the `malloc` implementation in `wasi-libc` is not thread-safe  iirc and no lock is being used in the test.

wenyongh:
OK, could we add lock to malloc for this case? And not sure whether there will be a thread-safe malloc in future version of wasi-libc?

eloparco:
> OK, could we add lock to malloc for this case? And not sure whether there will be a thread-safe malloc in future version of wasi-libc?

Done #2022

PR #1966, opened 2023-02-16
Merge wasi_threads to main

loganek:
I think the work is mainly completed. We're still testing the work but bugs found so far were not specific to wasi threads, and they're also problems for pthreads. I think the remaining work can be continued in main.


wenyongh:
@loganek I manually merged `main` into `dev/wasi_threads` to resolve the conflicts of several CI files, and now this PR is ready to merge. Wondering whether to `Squash and merge` or `Create a merge commit` it into main branch? The former keeps all the commits, but some may be duplicated in main branch since I used `Rebase and merge` to merge main into dev/wasi_threads last time. The latter will ignore the commit logs in dev/wasi, and only create a main log in main branch.

loganek:
Hi @wenyongh , thanks a lot. I think the very ideal case would be to rebase the `dev/wasi_threads` branch onto main and forcepush it before merging, so the history is clean and straight. However, that seems to be quite a bit of work and not sure if the benefits overweight that. So my preference would be to have `Create a merge commit` to keep the history (even though that means we'll have some duplications).

On a side note, I wonder, if for future dev branches we could possibly avoid merging back from main and instead of that periodically rebase onto main and force push (I hope the developers working on that dev branch wouldn't mind) so then the final rebase wouldn't be that painful. What do you think?

wenyongh:
> Hi @wenyongh , thanks a lot. I think the very ideal case would be to rebase the `dev/wasi_threads` branch onto main and forcepush it before merging, so the history is clean and straight. However, that seems to be quite a bit of work and not sure if the benefits overweight that. So my preference would be to have `Create a merge commit` to keep the history (even though that means we'll have some duplications).

OK, let's merge it first.

> 
> On a side note, I wonder, if for future dev branches we could possibly avoid merging back from main and instead of that periodically rebase onto main and force push (I hope the developers working on that dev branch wouldn't mind) so then the final rebase wouldn't be that painful. What do you think?

`periodically rebase onto main and force push` may be an option, but I am not so familiar with the commands. I think if the commits of merging main to that branch are not large (e.g. 1-3 commits), we may `Create a merge commit` to main to keep all histories, and if the commits are large, I may ask colleagues to help `rebase onto main and force push`.

loganek:
> but I am not so familiar with the commands

FYI, the commands would be:
```
git checkout <dev branch>
git fetch origin
git rebase origin/main
git push -f origin <dev branch>
```

wenyongh:
> > but I am not so familiar with the commands
> 
> FYI, the commands would be:
> 
> ```
> git checkout <dev branch>
> git fetch origin
> git rebase origin/main
> git push -f origin <dev branch>
> ```

OK, many thanks!

PR #1967, opened 2023-02-17
Don't call start/initialize in child thread's instantiation

wenyongh:
The start/initialize functions of wasi module are to do some initialization work
during instantiation, which should be only called one time in the instantiation
of main instance. For example, they may initialize the data in linear memory,
if the data was changed later by main instance, and re-initialized again by
sub instance, unexpected behaviors may occur.

And clear a shadow warning in classic interpreter.

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -901,6 +901,19 @@ create_exports(AOTModuleInstance *module_inst, AOTModule *module,
     return create_export_funcs(module_inst, module, error_buf, error_buf_size);
 }
 
+#if WASM_ENABLE_LIBC_WASI != 0
+static bool
+execute_initialize_function(AOTModuleInstance *module_inst)
+{
+    AOTFunctionInstance *initialize =
+        aot_lookup_function(module_inst, "_initialize", NULL);
+
+    return !initialize
+           || aot_create_exec_env_and_call_function(module_inst, initialize, 0,
+                                                    NULL);
+}
+#endif
+
 static bool
 execute_post_inst_function(AOTModuleInstance *module_inst)
 {
@@ -1121,11 +1134,27 @@ aot_instantiate(AOTModule *module, bool is_sub_inst, uint32 stack_size,
     }
 #endif
 
-    /* Execute __post_instantiate function and start function*/
-    if (!execute_post_inst_function(module_inst)
-        || !execute_start_function(module_inst)) {
-        set_error_buf(error_buf, error_buf_size, module_inst->cur_exception);
-        goto fail;
+    if (!is_sub_inst) {
+        if (
+#if WASM_ENABLE_LIBC_WASI != 0
+            /*
+             * reactor instances may assume that _initialize will be called by
+             * the environment at most once, and that none of their other
+             * exports are accessed before that call.
+             *
+             * let the loader decide how to act if there is no _initialize
+             * in a reactor
+             */
+            !execute_initialize_function(module_inst) ||
+#endif
+            /* Execute __post_instantiate function */
+            !execute_post_inst_function(module_inst)
+            /* Execute the function in "start" section */
+            || !execute_start_function(module_inst)) {
+            set_error_buf(error_buf, error_buf_size,
+                          module_inst->cur_exception);
+            goto fail;
+        }
     }
 
 #if WASM_ENABLE_BULK_MEMORY != 0
--- a/core/iwasm/interpreter/wasm_interp_classic.c
+++ b/core/iwasm/interpreter/wasm_interp_classic.c
@@ -1367,7 +1367,7 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
 
             HANDLE_OP(EXT_OP_BR_TABLE_CACHE)
             {
-                BrTableCache *node =
+                BrTableCache *node_cache =
                     bh_list_first_elem(module->module->br_table_cache_list);
                 BrTableCache *node_next;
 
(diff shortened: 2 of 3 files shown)


lum1n0us (approved):
LGTM

yamt:
while it seems reasonable for `_initialize`, i'm not sure about the start function.
is it specified in some standards?

wenyongh:
> while it seems reasonable for `_initialize`, i'm not sure about the start function. is it specified in some standards?

The [threads](https://github.com/WebAssembly/threads/blob/main/proposals/threads/Overview.md) spec proposal mentions [Initializing Memory Only Once](https://github.com/WebAssembly/threads/blob/main/proposals/threads/Overview.md#initializing-memory-only-once). For the start function generated by wasi-sdk when `-Wl,--shared-memory` is added, it may be used to initialize the linear memory:
```wast
(start $__wasm_init_memory)

(func $__wasm_init_memory (type $7)
    i32.const 5348
    i32.const 0
    i32.const 1
    i32.atomic.rmw.cmpxchg
    if $if
      i32.const 5348
      i32.const 1
      i64.const -1
      memory.atomic.wait32
      drop
    else
```
If child thread calls start function again, the shared memory will be duplicated initialized.

yamt:
> > while it seems reasonable for `_initialize`, i'm not sure about the start function. is it specified in some standards?
> 
> The [threads](https://github.com/WebAssembly/threads/blob/main/proposals/threads/Overview.md) spec proposal mentions [Initializing Memory Only Once](https://github.com/WebAssembly/threads/blob/main/proposals/threads/Overview.md#initializing-memory-only-once). For the start function generated by wasi-sdk when `-Wl,--shared-memory` is added, it may be used to initialize the linear memory:
> 
> ```webassembly
> (start $__wasm_init_memory)
> 
> (func $__wasm_init_memory (type $7)
>     i32.const 5348
>     i32.const 0
>     i32.const 1
>     i32.atomic.rmw.cmpxchg
>     if $if
>       i32.const 5348
>       i32.const 1
>       i64.const -1
>       memory.atomic.wait32
>       drop
>     else
> ```
> 
> If child thread calls start function again, the shared memory will be duplicated initialized.


as you can see in the code you cited,
`__wasm_init_memory` itself prevents duplicated initialization.
it should be safe to be called multiple times.
cf. https://reviews.llvm.org/D65783

wenyongh:
> > > while it seems reasonable for `_initialize`, i'm not sure about the start function. is it specified in some standards?
> > 
> > 
> > The [threads](https://github.com/WebAssembly/threads/blob/main/proposals/threads/Overview.md) spec proposal mentions [Initializing Memory Only Once](https://github.com/WebAssembly/threads/blob/main/proposals/threads/Overview.md#initializing-memory-only-once). For the start function generated by wasi-sdk when `-Wl,--shared-memory` is added, it may be used to initialize the linear memory:
> > ```webassembly
> > (start $__wasm_init_memory)
> > 
> > (func $__wasm_init_memory (type $7)
> >     i32.const 5348
> >     i32.const 0
> >     i32.const 1
> >     i32.atomic.rmw.cmpxchg
> >     if $if
> >       i32.const 5348
> >       i32.const 1
> >       i64.const -1
> >       memory.atomic.wait32
> >       drop
> >     else
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If child thread calls start function again, the shared memory will be duplicated initialized.
> 
> as you can see in the code you cited, `__wasm_init_memory` itself prevents duplicated initialization. it should be safe to be called multiple times. cf. https://reviews.llvm.org/D65783

Yes, but what if it doesn't prevent that, can we ensure that all compilers generate the similar code? And it doesn't introduce side effect if child thread doesn't call start again, instead, it may result in unexpected behavior if child thread calls start again.

wenyongh:
> > as you can see in the code you cited, `__wasm_init_memory` itself prevents duplicated initialization. it should be safe to be called multiple times. cf. https://reviews.llvm.org/D65783
> 
> Yes, but what if it doesn't prevent that, can we ensure that all compilers generate the similar code? And it doesn't introduce side effect if child thread doesn't call start again, instead, it may result in unexpected behavior if child thread calls start again.

@yamt I am not very sure yet, but it looks a little strange, my personal preference is not to call start again.

yamt:
to me, it seems wrong to assume that memory initialization is the only thing a start function can do.

btw, i noticed we call `_initialize` before the start function. is it expected? my expectation was the opposite.

lum1n0us:
for a commander, `_start` is an export entry point. And all exports will call `wasm_call_ctor`(`_initialize` will call) before execution.

IMO, the sub_inst entry is actually `run` instead of `start`, right?

wenyongh:
> to me, it seems wrong to assume that memory initialization is the only thing a start function can do.

Yes, that's confusing, the start function may initialize the shared linear memory, non-shared table and global. For the shared memory, it should be initialized only once, but for table/global, each thread seems should call start function if start function initializes them. I think it really depends the behavior of compiler.
Let's suppose the compiler will prevent duplicated initialization of shared memory in start function, and let runtime calls it in each thread instance's instantiation?

Another issue is that when hw bound check is enabled, in the instantiation of child thread, the exec_enc used to call the start function may be the exec_env of main thread (first argument of pthread_create_wrapper) but not a newly created exec_env, so the module_inst in that exec_env is the module_inst of the main thread, and is different from the newly instantiated module_inst, and the start function will initialize the original main thread's module_inst again:
```C
static bool
execute_start_function(WASMModuleInstance *module_inst)
{
    ...
    return wasm_create_exec_env_and_call_function(module_inst, func, 0, NULL);
}

bool
wasm_create_exec_env_and_call_function(WASMModuleInstance *module_inst,
                                       WASMFunctionInstance *func,
                                       unsigned argc, uint32 argv[])
{
    WASMExecEnv *exec_env = NULL, *existing_exec_env = NULL;
    bool ret;

#if defined(OS_ENABLE_HW_BOUND_CHECK)
    existing_exec_env = exec_env = wasm_runtime_get_exec_env_tls();
    => the exec_env searched from exec_env_tls is exec_env of main thread,
          and contains the module_inst of main thread, but not the new instance
    ...
}
```

> 
> btw, i noticed we call `_initialize` before the start function. is it expected? my expectation was the opposite.

Agree, had better call `_initialize` after `start`

wenyongh:
> to me, it seems wrong to assume that memory initialization is the only thing a start function can do.
> 
> btw, i noticed we call `_initialize` before the start function. is it expected? my expectation was the opposite.

@yamt I uploaded PR #1972 to re-org/refine the code, please help review, thanks.

yamt:
> for a commander, `_start` is an export entry point. And all exports will call `wasm_call_ctor`(`_initialize` will call) before execution.
> 
> IMO, the sub_inst entry is actually `run` instead of `start`, right?

i suspect you are confused between wasm start function and WASI `_start`.

PR #1969, opened 2023-02-17
Expose wasm_runtime_call_indirect

yamt:
The function has been there for long.
While what it does looks a bit unsafe as it calls a function which is not wasm-wise exported explicitly, it's useful and widely used when implemention callback-taking APIs, including our pthread implementation. We even have a sample to use it.

### DIFF
--- a/core/iwasm/common/wasm_runtime_common.h
+++ b/core/iwasm/common/wasm_runtime_common.h
@@ -617,7 +617,7 @@ wasm_runtime_call_wasm_v(WASMExecEnv *exec_env,
 
 /* See wasm_export.h for description */
 WASM_RUNTIME_API_EXTERN bool
-wasm_runtime_call_indirect(WASMExecEnv *exec_env, uint32 element_indices,
+wasm_runtime_call_indirect(WASMExecEnv *exec_env, uint32 element_index,
                            uint32 argc, uint32 argv[]);
 
 #if WASM_ENABLE_DEBUG_INTERP != 0
--- a/core/iwasm/common/wasm_runtime_common.h
+++ b/core/iwasm/common/wasm_runtime_common.h
@@ -615,6 +615,11 @@ wasm_runtime_call_wasm_v(WASMExecEnv *exec_env,
                          uint32 num_results, wasm_val_t *results,
                          uint32 num_args, ...);
 
+/* See wasm_export.h for description */
+WASM_RUNTIME_API_EXTERN bool
+wasm_runtime_call_indirect(WASMExecEnv *exec_env, uint32 element_indices,
+                           uint32 argc, uint32 argv[]);
+
 #if WASM_ENABLE_DEBUG_INTERP != 0
 /* See wasm_export.h for description */
 WASM_RUNTIME_API_EXTERN uint32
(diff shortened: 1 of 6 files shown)


core/iwasm/include/wasm_export.h
```diff
@@ -800,6 +800,27 @@ wasm_runtime_call_wasm_v(wasm_exec_env_t exec_env,
                          uint32_t num_results, wasm_val_t results[],
                          uint32_t num_args, ...);
 
+/**
+ * Call a function reference of a given WASM runtime instance with
+ * arguments.
```
wenyongh, line 805: Could we mention that it may be unsafe since the function is not wasm-wise exported explicitly?

core/iwasm/include/wasm_export.h
```diff
@@ -800,6 +800,27 @@ wasm_runtime_call_wasm_v(wasm_exec_env_t exec_env,
...
+ *
+ * @param exec_env the execution environment to call the function
+ *   which must be created from wasm_create_exec_env()
+ * @param element_indices the function ference indicies, usually
```
wenyongh, line 809: Not sure whether element_indice is better?
And `the function ference indicies` to `the function reference indice`?

core/iwasm/include/wasm_export.h
yamt: done

core/iwasm/include/wasm_export.h
yamt: i just moved this comment from another file. anyway, fixed.

core/iwasm/include/wasm_export.h
```diff
@@ -800,6 +800,31 @@ wasm_runtime_call_wasm_v(wasm_exec_env_t exec_env,
                          uint32_t num_results, wasm_val_t results[],
                          uint32_t num_args, ...);
 
+/**
+ * Call a function reference of a given WASM runtime instance with
+ * arguments.
+ *
+ * Note: this can be used to call a function which is not exported
+ * by the module's explicitly. You might consider it as an abstraction
```
wenyongh, line 808: by the module?

core/iwasm/common/wasm_runtime_common.h
```diff
@@ -615,6 +615,11 @@ wasm_runtime_call_wasm_v(WASMExecEnv *exec_env,
                          uint32 num_results, wasm_val_t *results,
                          uint32 num_args, ...);
 
+/* See wasm_export.h for description */
+WASM_RUNTIME_API_EXTERN bool
+wasm_runtime_call_indirect(WASMExecEnv *exec_env, uint32 element_indices,
```
wenyongh, line 620: Had better `element_index`?

wenyongh (commented):
LGTM

core/iwasm/include/wasm_export.h
wenyongh: yes, it was committed by the some community developer before, we should review the code more strictly and let him change :)

core/iwasm/include/wasm_export.h
yamt: yes. typo. i will fix.

core/iwasm/common/wasm_runtime_common.h
yamt: i fill fix.

loganek (approved):

core/iwasm/include/wasm_export.h
yamt: done

Issue #1962, opened 2023-02-15
Shared memory should only be shared inside cluster


xujuntwt95329:
In current implementation, the shared memory is shared globally, instances from different clusters shares the same memory, which is not expected behavior.

We should only share the memory inside cluster, and when new clusters are created, there should be new memory instances.

The problem is: the cluster is created during creating exec_env rather than module instance, so during instantiating there may be no cluster information.

xujuntwt95329:
Possible solution:
### Data structures:
- In `WASMSharedMemNode`, add one field: `main_inst_wp`, weak pointer to main instance
- In `WASMCluster`, add one field: `shared_memory_wp`, weak pointer to the shared `MemoryInstance`
### Logic:
- When creating shared memory instance, store the pointer of main instance in `main_inst_wp` field
- When creating new cluster, if the module's memory is shared, search the `WASMSharedMemNode` according to the `module` and `main_instance`, and set the shared `MemoryInstance` to `shared_memory_wp` field
- Change `is_sub_inst` parameter of `wasm_runtime_instantiate_internal` to `WASMCluster *cluster`
    - If cluster is NULL, always create new memory instance rather than search the shared memory list.
    - If cluster is not NULL and module's memory is shared, use the `cluster->shared_memory_wp` directly

PR #1971, opened 2023-02-18
Fix data races

eloparco:
I ran into multiple data races when executing samples that uses threads (pthreads or WASI threads).
This PR protect the access to the shared `GOTO_TABLE` and memory used by atomic wait and notify operations.
Tested using thread sanitizers.


eloparco:
Missing data races to fix:
- [ ] `memory` used in `wasm_shared_memory.c` and `wasm_memory.c` (obtained with `wasm_get_default_memory()`)
  - related with #1962
- [ ] data race in thread manager
- [ ] data race in store/load operations (root cause still not clear)

core/iwasm/interpreter/wasm_opcode.h
```diff
@@ -686,230 +686,230 @@ typedef enum WASMAtomicEXTOpcode {
...
-        HANDLE_OPCODE(EXT_OP_BR_TABLE_CACHE),        /* 0xd6 */ \
-    };                                                          \
-    do {                                                        \
-        _name[WASM_OP_MISC_PREFIX] =                            \
-            HANDLE_OPCODE(WASM_OP_MISC_PREFIX); /* 0xfc */      \
-        _name[WASM_OP_ATOMIC_PREFIX] =                          \
-            HANDLE_OPCODE(WASM_OP_ATOMIC_PREFIX); /* 0xfe */    \
-        DEF_DEBUG_BREAK_HANDLE(_name)                           \
+#define DEFINE_GOTO_TABLE(type, _name)                                    \
+    static os_thread_local_attribute type _name[WASM_INSTRUCTION_NUM] = { \
```
loganek, line 690: is there a data race risk here? I thought it's only initialized once before first access, and it's read-only.

core/iwasm/common/wasm_memory.c
```diff
@@ -16,6 +16,7 @@ typedef enum Memory_Mode {
 } Memory_Mode;
 
 static Memory_Mode memory_mode = MEMORY_MODE_UNKNOWN;
+static korp_mutex mem_mode_lock;
```
loganek, line 19: could we make `memory_mode` atomic instead?

eloparco: right, that's better since `memory_mode` is a simple enum

core/iwasm/interpreter/wasm_opcode.h
eloparco: It's not done only once right now https://github.com/bytecodealliance/wasm-micro-runtime/blob/739acfc9082e63967ee91fca85dc01ab1630a5cc/core/iwasm/interpreter/wasm_interp_classic.c#L1170
but I'll see if I can move the initialization somewhere else (also because I noticed from the CI that TLS is not supported on all platforms)

core/iwasm/common/wasm_memory.c
```diff
@@ -39,7 +39,7 @@ wasm_memory_init_with_pool(void *mem, unsigned int bytes)
     mem_allocator_t _allocator = mem_allocator_create(mem, bytes);
 
     if (_allocator) {
-        memory_mode = MEMORY_MODE_POOL;
+        __atomic_store_n(&memory_mode, MEMORY_MODE_POOL, __ATOMIC_SEQ_CST);
```
wenyongh, line 42: I remember that the global variable `memory_mode` is set only when wasm_runtime_init/wasm_runtime_full_init, and won't be set again after that. Is there any data race risk on it?

core/iwasm/interpreter/wasm_opcode.h
wenyongh: It is a `static const void * handle_table[256]` table, its data isn't changed after it is defined, is there any data race risk on it?

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -157,13 +157,13 @@ int32
 shared_memory_inc_reference(WASMModuleCommon *module)
 {
     WASMSharedMemNode *node = search_module(module);
+    uint32 ref_count = -1;
     if (node) {
         os_mutex_lock(&node->lock);
-        node->ref_count++;
+        ref_count = node->ref_count++;
```
wenyongh, line 163: Should be `++node->ref_count` according to the original code

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -417,8 +417,14 @@ wasm_runtime_atomic_wait(WASMModuleInstanceCommon *module, void *address,
...
     os_mutex_lock(&wait_info->wait_list_lock);
 
-    if ((!wait64 && *(uint32 *)address != (uint32)expect)
-        || (wait64 && *(uint64 *)address != expect)) {
+    WASMSharedMemNode *node =
+        search_module((WASMModuleCommon *)module_inst->module);
+    os_mutex_lock(&node->shared_mem_lock);
+    bool no_wait = (!wait64 && *(uint32 *)address != (uint32)expect)
+                   || (wait64 && *(uint64 *)address != expect);
+    os_mutex_unlock(&node->shared_mem_lock);
```
wenyongh, line 425: The whole operation is already locked by wait_info->wait_list_lock, shall we add the lock node->shared_mem_lock again?

wenyongh:
> Missing data races to fix:
> 
> * [ ]  `memory` used in `wasm_shared_memory.c` and `wasm_memory.c` (obtained with `wasm_get_default_memory()`)
>   
>   * probably related with [Shared memory should only be shared inside cluster #1962](https://github.com/bytecodealliance/wasm-micro-runtime/issues/1962)
>   * not sure where to put the lock or if we can reuse an existing one
> * [ ]  data race in thread manager
> * [ ]  data race in store/load operations (root cause still not clear)

@eloparco Can this PR resolve the issue you found? Could you describe how to reproduce the issue and upload the related source code?

eloparco:
> @eloparco Can this PR resolve the issue you found? Could you describe how to reproduce the issue and upload the related source code?

I wanted to fix as many of them as possible in this PR.

To reproduce, what I do is, from the branch `eloparco/test-wasi-threads` (#1963) I run the samples there + `samples/wasi-threads/wasm-apps/no_pthread.c` and `samples/wasi-threads/wasm-apps/thread_termination.c` using the thread sanitizer.
To use the sanitizer, after this line
https://github.com/bytecodealliance/wasm-micro-runtime/blob/739acfc9082e63967ee91fca85dc01ab1630a5cc/samples/wasi-threads/CMakeLists.txt#L78
I add the following two lines:
```
set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fsanitize=thread")
set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fsanitize=thread")
```

And that detects multiple data races, that I was trying to fix.

core/iwasm/common/wasm_shared_memory.c
eloparco: For other atomic operations (for instance the atomic store), we acquired a lock before accessing the memory `maddr`
https://github.com/bytecodealliance/wasm-micro-runtime/blob/739acfc9082e63967ee91fca85dc01ab1630a5cc/core/iwasm/interpreter/wasm_interp_classic.c#L3557
But here (atomic wait)
https://github.com/bytecodealliance/wasm-micro-runtime/blob/739acfc9082e63967ee91fca85dc01ab1630a5cc/core/iwasm/interpreter/wasm_interp_classic.c#L3420
we don't do that

core/iwasm/common/wasm_memory.c
eloparco: this is probably the only exception (done when destroying the memory)
https://github.com/bytecodealliance/wasm-micro-runtime/blob/739acfc9082e63967ee91fca85dc01ab1630a5cc/core/iwasm/common/wasm_memory.c#L131
and it's causing the data race

core/iwasm/interpreter/wasm_opcode.h
eloparco:
```bash
WARNING: ThreadSanitizer: data race (pid=1397466)
  Read of size 8 at 0x55f7ae9d2570 by main thread:
    #0 wasm_interp_call_func_bytecode wasm-micro-runtime/core/iwasm/interpreter/wasm_interp_classic.c:2123 (iwasm+0x4f629)
    #1 wasm_interp_call_wasm wasm-micro-runtime/core/iwasm/interpreter/wasm_interp_classic.c:4207 (iwasm+0x5a062)
    #2 call_wasm_with_hw_bound_check wasm-micro-runtime/core/iwasm/interpreter/wasm_runtime.c:2261 (iwasm+0x1c01e)
[...]

  Previous write of size 8 at 0x55f7ae9d2570 by thread T1:
    #0 wasm_interp_call_func_bytecode wasm-micro-runtime/core/iwasm/interpreter/wasm_interp_classic.c:1170 (iwasm+0x49feb)
    #1 wasm_interp_call_wasm wasm-micro-runtime/core/iwasm/interpreter/wasm_interp_classic.c:4207 (iwasm+0x5a062)
    #2 call_wasm_with_hw_bound_check wasm-micro-runtime/core/iwasm/interpreter/wasm_runtime.c:2261 (iwasm+0x1c01e)
[...]
```
I guess the problem is with what is done after the static initialization, in the `do ... while` https://github.com/bytecodealliance/wasm-micro-runtime/blob/739acfc9082e63967ee91fca85dc01ab1630a5cc/core/iwasm/interpreter/wasm_opcode.h#L908

eloparco: I moved everything into the initialization, that seems to fix the problem

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -417,8 +417,14 @@ wasm_runtime_atomic_wait(WASMModuleInstanceCommon *module, void *address,
 
     os_mutex_lock(&wait_info->wait_list_lock);
 
-    if ((!wait64 && *(uint32 *)address != (uint32)expect)
-        || (wait64 && *(uint64 *)address != expect)) {
+    WASMSharedMemNode *node =
```
wenyongh, line 420: Could you declare `node` and `no_wait` below at the beginning of the function? We usually keep this C coding style to avoid compile warning for some compilers.

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -493,13 +501,22 @@ wasm_runtime_atomic_notify(WASMModuleInstanceCommon *module, void *address,
     bh_assert(module->module_type == Wasm_Module_Bytecode
               || module->module_type == Wasm_Module_AoT);
 
-    if ((uint8 *)address < module_inst->memories[0]->memory_data
-        || (uint8 *)address + 4 > module_inst->memories[0]->memory_data_end) {
+    WASMSharedMemNode *node =
```
wenyongh, line 504: Same as above for `node` and `out_of_bounds`

core/iwasm/interpreter/wasm_opcode.h
```diff
@@ -675,242 +675,239 @@ typedef enum WASMAtomicEXTOpcode {
 } WASMAtomicEXTOpcode;
 
 #if WASM_ENABLE_DEBUG_INTERP != 0
-#define DEF_DEBUG_BREAK_HANDLE(_name) \
-    _name[DEBUG_OP_BREAK] = HANDLE_OPCODE(DEBUG_OP_BREAK); /* 0xd7 */
+#define DEF_DEBUG_BREAK_HANDLE \
```
wenyongh, line 678: Had better `DEF_DEBUG_BREAK_HANDLE()`, a macro without `()` looks a little strange.

core/iwasm/interpreter/wasm_opcode.h
```diff
@@ -675,242 +675,239 @@ typedef enum WASMAtomicEXTOpcode {
...
+    [DEBUG_OP_BREAK] = HANDLE_OPCODE(DEBUG_OP_BREAK), /* 0xd7 */
 #else
-#define DEF_DEBUG_BREAK_HANDLE(_name)
+#define DEF_DEBUG_BREAK_HANDLE NULL
```
wenyongh, line 681: Had better `#define DEF_DEBUG_BREAK_HANDLE()` to avoid adding code in goto table?

core/iwasm/common/wasm_memory.c
wenyongh: The root cause of `memory_mode` data race is not this file, it is due to main thread doesn't wait the destroying of sub thread's exec env, and start to destroy the runtime environment (only when all child threads exit and all exec envs are destroyed can the wasm_runtime_destroy be called). And then when sub thread's exec env is destroyed, it reads an re-initialized `memory_mode`.
I uploaded PR #1973 to fix it, could you help try?

core/iwasm/interpreter/wasm_opcode.h
```diff
@@ -675,242 +675,239 @@ typedef enum WASMAtomicEXTOpcode {
...
+        HANDLE_OPCODE(WASM_OP_REF_NULL),                            /* 0xd0 */ \
+        HANDLE_OPCODE(WASM_OP_REF_IS_NULL),                         /* 0xd1 */ \
+        HANDLE_OPCODE(WASM_OP_REF_FUNC),                            /* 0xd2 */ \
+        HANDLE_OPCODE(EXT_OP_BLOCK),                                /* 0xd3 */ \
+        HANDLE_OPCODE(EXT_OP_LOOP),                                 /* 0xd4 */ \
+        HANDLE_OPCODE(EXT_OP_IF),                                   /* 0xd5 */ \
+        HANDLE_OPCODE(EXT_OP_BR_TABLE_CACHE),                       /* 0xd6 */ \
+        [WASM_OP_MISC_PREFIX] = HANDLE_OPCODE(WASM_OP_MISC_PREFIX), /* 0xfc */ \
+        [WASM_OP_ATOMIC_PREFIX] =                                              \
+            HANDLE_OPCODE(WASM_OP_ATOMIC_PREFIX), /* 0xfe */                   \
```
wenyongh, line 908: Adding the two lines causes the change of many lines, how about defining a macro to avoid that?
```C
#define SET_GOTO_TABLE_ELEM(opcode) \
[opcode] = HANDLE_OPCODE(opcode)
```
And here:
```C
SET_GOTO_TABLE_ELEM(WASM_OP_MISC_PREFIX),
SET_GOTO_TABLE_ELEM(WASM_OP_ATOMIC_PREFIX),
```

core/iwasm/common/wasm_memory.c
```diff
@@ -54,7 +54,7 @@ wasm_memory_init_with_allocator(void *_user_data, void *_malloc_func,
                                 void *_realloc_func, void *_free_func)
 {
     if (_malloc_func && _free_func && _malloc_func != _free_func) {
-        memory_mode = MEMORY_MODE_ALLOCATOR;
+        __atomic_store_n(&memory_mode, MEMORY_MODE_ALLOCATOR, __ATOMIC_SEQ_CST);
```
loganek, line 57: This won't work with e.g. msvc compiler, although I'm not sure if wasi-libc is supposed to work on windows? Anyway, as @wenyongh pointed out, we might not need this change if his PR fixes the problem.

core/iwasm/common/wasm_memory.c
eloparco: Yes, that was my guess too, but I didn't have time to try it. I'll check the change and let you know in #1973

core/iwasm/interpreter/wasm_opcode.h
eloparco: sure, that looks better

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -489,17 +497,27 @@ wasm_runtime_atomic_notify(WASMModuleInstanceCommon *module, void *address,
...
+    WASMSharedMemNode *node;
+    bool out_of_bounds;
 
     bh_assert(module->module_type == Wasm_Module_Bytecode
               || module->module_type == Wasm_Module_AoT);
 
-    if ((uint8 *)address < module_inst->memories[0]->memory_data
-        || (uint8 *)address + 4 > module_inst->memories[0]->memory_data_end) {
+    node = search_module((WASMModuleCommon *)module_inst->module);
+    os_mutex_lock(&node->shared_mem_lock);
```
eloparco, line 507: this line is creating problems in the spec tests, I'm checking why

wenyongh: > This line is creating problems in the spec tests, I'm checking why.
> 
> [EDIT]: The problem is that, according to the specification, `atomic.notify` is supposed to also work without shared memory (in contrast with `atomic.wait`). But, without shared memory, `node->shared_mem_lock` is not initialized, making the program crash.

The `node->shared_mem_lock` is initialized when node is created, I think what you mean should be that `node` may be NULL when `atomic.notify`. How about adding checks for it:
```C
    if (node)
    ut_of_bounds =
        ((uint8 *)address < module_inst->memories[0]->memory_data
         || (uint8 *)address + 4 > module_inst->memories[0]->memory_data_end);
    if (node)
```
```

File: core/iwasm/common/wasm_shared_memory.c [hunk 13]
```diff
@@ -489,17 +497,27 @@ wasm_runtime_atomic_notify(WASMModuleInstanceCommon *module, void *address,
...
+    os_mutex_unlock(&node->shared_mem_lock);
+
+    if (out_of_bounds) {
         wasm_runtime_set_exception(module, "out of bounds memory access");
         return -1;
     }
 
+    os_mutex_lock(&node->shared_mem_lock);
     wait_info = acquire_wait_info(address, false);
+    os_mutex_unlock(&node->shared_mem_lock);
```
wenyongh, line 520: Same as above:
```C
    if (node)
    wait_info = acquire_wait_info(address, false);
    if (node)
```
I have a test, after adding checks, the spec cases can run success.

wenyongh: And maybe here we can avoid locking twice:
```C
    node = search_module((WASMModuleCommon *)module_inst->module);
    if (node)
    out_of_bounds =
        ((uint8 *)address < module_inst->memories[0]->memory_data
         || (uint8 *)address + 4 > module_inst->memories[0]->memory_data_end);

    if (out_of_bounds) {
        wasm_runtime_set_exception(module, "out of bounds memory access");
        if (node)
        return -1;
    }

    wait_info = acquire_wait_info(address, false);
    if (node)
```

core/iwasm/common/wasm_shared_memory.c
eloparco: Right! I was mislead since I was using an assert to check that `node` was not null without realizing that spec tests were executed in release mode. Thanks

core/iwasm/common/wasm_shared_memory.c
eloparco: Not sure what's failing now, the error is different; I can reproduce it locally, but I see the same error even if I run the spec tests from the main branch

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -489,17 +497,27 @@ wasm_runtime_atomic_notify(WASMModuleInstanceCommon *module, void *address,
...
+    node = search_module((WASMModuleCommon *)module_inst->module);
+    if (node)
+        os_mutex_lock(&node->shared_mem_lock);
+    out_of_bounds =
+        ((uint8 *)address < module_inst->memories[0]->memory_data
+         || (uint8 *)address + 4 > module_inst->memories[0]->memory_data_end);
+
+    if (out_of_bounds) {
         wasm_runtime_set_exception(module, "out of bounds memory access");
         return -1;
```
wenyongh, line 515: Should also `if (node) os_mutex_unlock(&node->shared_mem_lock)` before return -1.

core/iwasm/common/wasm_shared_memory.c
wenyongh: Missed adding unlock before return -1, seems that causes the spec test failure.

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -489,17 +497,29 @@ wasm_runtime_atomic_notify(WASMModuleInstanceCommon *module, void *address,
...
+    if (node)
+        os_mutex_lock(&node->shared_mem_lock);
+    out_of_bounds =
+        ((uint8 *)address < module_inst->memories[0]->memory_data
+         || (uint8 *)address + 4 > module_inst->memories[0]->memory_data_end);
+
+    if (out_of_bounds) {
         wasm_runtime_set_exception(module, "out of bounds memory access");
+        if (node)
+            os_mutex_unlock(&node->shared_mem_lock);
```
wenyongh, line 516: Seems that we should put the unlock operation before L514 `wasm_runtime_set_exception`, or spec cases will run failed. Not sure whether it is caused by the spreading of exception or the termination of other threads?

core/iwasm/common/wasm_shared_memory.c
eloparco: thanks, that worked!

wenyongh (approved):
LGTM

PR #1972, opened 2023-02-20
Re-org calling functions after instantiation

wenyongh:
- Use execute_post_instantiate_functions to call start, _initialize,
  __post_instantiate, __wasm_call_ctors functions
- Always call start function for both main instance and sub instance
- Only call _initialize and __post_instantiate for main instance
- Only call ___wasm_call_ctors for main instance and when bulk memory
  is enabled and wasi import functions are not found
- When hw bound check is enabled, use the existing exec_env_tls
  to call func for sub instance, and switch exec_env_tls's module inst
  to current module inst to avoid checking failure and passing the wrong
  module inst

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -901,76 +901,148 @@ create_exports(AOTModuleInstance *module_inst, AOTModule *module,
     return create_export_funcs(module_inst, module, error_buf, error_buf_size);
 }
 
-#if WASM_ENABLE_LIBC_WASI != 0
-static bool
-execute_initialize_function(AOTModuleInstance *module_inst)
+static AOTFunctionInstance *
+lookup_post_instantiate_func(AOTModuleInstance *module_inst,
+                             const char *func_name)
 {
-    AOTFunctionInstance *initialize =
-        aot_lookup_function(module_inst, "_initialize", NULL);
+    AOTFunctionInstance *func;
+    AOTFuncType *func_type;
 
-    return !initialize
-           || aot_create_exec_env_and_call_function(module_inst, initialize, 0,
-                                                    NULL);
-}
-#endif
-
-static bool
-execute_post_inst_function(AOTModuleInstance *module_inst)
-{
-    AOTFunctionInstance *post_inst_func =
-        aot_lookup_function(module_inst, "__post_instantiate", "()");
-
-    if (!post_inst_func)
+    if (!(func = aot_lookup_function(module_inst, func_name, NULL)))
         /* Not found */
-        return true;
+        return NULL;
 
-    return aot_create_exec_env_and_call_function(module_inst, post_inst_func, 0,
-                                                 NULL);
+    func_type = func->u.func.func_type;
+    if (!(func_type->param_count == 0 && func_type->result_count == 0))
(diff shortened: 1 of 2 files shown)


lum1n0us:
a `_start` of *wasm32-wasi* function may look like below. It won't return unless `main()` finish. It may cause a re-entering issue if *Always call start function for sub instance*. It is also a little un-appropriated considering that WAMR tends to use execution context instead of the real instantiate.

```
0004fc func[29] <_start>:
 0004fd: 01 7f                      | local[0] type=i32
 0004ff: 02 40                      | block
 000501: 02 40                      |   block
 000503: 41 00                      |     i32.const 0
 000505: 28 02 90 a3 80 80 00       |     i32.load 2 4496
 00050c: 0d 00                      |     br_if 0
 00050e: 41 00                      |     i32.const 0
 000510: 41 01                      |     i32.const 1
 000512: 36 02 90 a3 80 80 00       |     i32.store 2 4496
 000519: 10 9b 80 80 80 00          |     call 27 <__wasm_call_ctors>
 00051f: 10 bd 80 80 80 00          |     call 61 <__main_void>
 000525: 21 00                      |     local.set 0
 000527: 10 d0 80 80 80 00          |     call 80 <__wasm_call_dtors>
 00052d: 20 00                      |     local.get 0
 00052f: 0d 01                      |     br_if 1
 000531: 0f                         |     return
 000532: 0b                         |   end
 000533: 00                         |   unreachable
 000534: 00                         |   unreachable
 000535: 0b                         | end
 000536: 20 00                      | local.get 0
 000538: 10 c4 80 80 80 00          | call 68 <__wasi_proc_exit>
 00053e: 00                         | unreachable
 00053f: 0b                         | end
```

wenyongh:
> a `_start` of _wasm32-wasi_ function may look like below. It won't return unless `main()` finish. It may cause a re-entering issue if _Always call start function for sub instance_. It is also a little un-appropriated considering that WAMR tends to use execution context instead of the real instantiate.
> 

Here the `start` function is the wasm start function but not the `_start` function exposed by the wasi module:
https://webassembly.github.io/spec/core/syntax/modules.html#syntax-start

lum1n0us:
> The start function is intended for initializing the state of a module. The module and its exports are not accessible externally before this initialization has completed.

But in our "sub instance" case, IMO, we want to inherent variables from "the main instance" instead of using variables with initialized values, right?

yamt:
is it intended to call the start function first for interpreter but last for aot? why?

wenyongh:
> is it intended to call the start function first for interpreter but last for aot? why?

No, for aot, runtime also calls start first, the calling sequence is same as interpreter (start, _initialize, __post_instantiate and __wasm_call_ctors). Could you check the code again?

yamt:
> > is it intended to call the start function first for interpreter but last for aot? why?
> 
> No, for aot, runtime also calls start first, the calling sequence is same as interpreter (start, _initialize, __post_instantiate and __wasm_call_ctors). Could you check the code again?

ok. i misread the code. thank you.

i feel it's a bit confusing that aot execute_post_instantiate_functions handles the start function
while interpreter execute_post_instantiate_functions doesn't.
otherwise lgtm.

yamt (approved):

PR #1973, opened 2023-02-20
Fix possible data race in thread manager

wenyongh:
Destroy child thread's exec_env before destroying its module instance and
add the process into cluster's lock to avoid possible data race: if exec_env
is removed from custer's exec_env_list and destroyed later, the main thread
may not wait it and start to destroy the wasm runtime, and the destroying
of the sub thread's exec_env may free or overwritten an destroyed or 
re-initialized resource.

And fix an issue in wasm_cluster_cancel_thread.


loganek (approved):
LGTM

eloparco:
Thanks, I just tested it and it fixes the problem I was seeing in #1971, let's merge it so that I can rebase my PR

wenyongh:
> Thanks, I just tested it and it fixes the problem I was seeing in #1971, let's merge it so that I can rebase my PR

Done.

PR #1975, opened 2023-02-20
Do not fetch the next instruction when proc_exit is propagated

eloparco:
Without the change in this PR, upon receiving the `proc_exit` exception propagation while on an `atomic.atomic` or `poll_oneoff`, the next instruction is fetched instead of stopping the executing.

That results in the exception being overridden (by an `unreachable` exception) when using the tests in https://github.com/WebAssembly/wasi-threads/tree/main/test/testsuite, causing the exit code to be wrong.
Continuation of #1929 #1951.

### DIFF
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -194,11 +194,9 @@ runtime_signal_handler(void *sig_addr)
         else if (exec_env_tls->exce_check_guard_page <= (uint8 *)sig_addr
                  && (uint8 *)sig_addr
                         < exec_env_tls->exce_check_guard_page + page_size) {
-            bh_assert(wasm_get_exception(module_inst)
-#if WASM_ENABLE_THREAD_MGR != 0
-                      || wasm_cluster_is_thread_terminated(exec_env_tls)
+#if WASM_ENABLE_THREAD_MGR == 0
+            bh_assert(wasm_get_exception(module_inst));
 #endif
-            );
             os_longjmp(jmpbuf_node->jmpbuf, 1);
         }
     }
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -194,7 +194,6 @@ runtime_signal_handler(void *sig_addr)
         else if (exec_env_tls->exce_check_guard_page <= (uint8 *)sig_addr
                  && (uint8 *)sig_addr
                         < exec_env_tls->exce_check_guard_page + page_size) {
-            bh_assert(wasm_get_exception(module_inst));
             os_longjmp(jmpbuf_node->jmpbuf, 1);
         }
     }
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -194,6 +194,11 @@ runtime_signal_handler(void *sig_addr)
         else if (exec_env_tls->exce_check_guard_page <= (uint8 *)sig_addr
                  && (uint8 *)sig_addr
                         < exec_env_tls->exce_check_guard_page + page_size) {
+            bh_assert(wasm_get_exception(module_inst)
+#if WASM_ENABLE_THREAD_MGR != 0
+                      || wasm_cluster_is_thread_terminated(exec_env_tls)
+#endif
+            );
             os_longjmp(jmpbuf_node->jmpbuf, 1);
         }
     }
@@ -249,7 +254,11 @@ runtime_exception_handler(EXCEPTION_POINTERS *exce_info)
             else if (exec_env_tls->exce_check_guard_page <= (uint8 *)sig_addr
                      && (uint8 *)sig_addr
                             < exec_env_tls->exce_check_guard_page + page_size) {
-                bh_assert(wasm_get_exception(module_inst));
+                bh_assert(wasm_get_exception(module_inst)
+#if WASM_ENABLE_THREAD_MGR != 0
+                          || wasm_cluster_is_thread_terminated(exec_env_tls)
+#endif
+                );
                 if (module_inst->module_type == Wasm_Module_Bytecode) {
                     return EXCEPTION_CONTINUE_SEARCH;
                 }
--- a/core/iwasm/compilation/aot_emit_function.c
+++ b/core/iwasm/compilation/aot_emit_function.c
@@ -1003,7 +1003,7 @@ aot_compile_op_call(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
     /* Insert suspend check point */
     if (comp_ctx->enable_thread_mgr) {
         if (!check_suspend_flags(comp_ctx, func_ctx))
-            return false;
+            goto fail;
     }
 #endif
 
@@ -1657,7 +1657,7 @@ aot_compile_op_call_indirect(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
     /* Insert suspend check point */
     if (comp_ctx->enable_thread_mgr) {
         if (!check_suspend_flags(comp_ctx, func_ctx))
-            return false;
+            goto fail;
     }
 #endif
 
--- a/core/iwasm/compilation/aot_emit_function.c
+++ b/core/iwasm/compilation/aot_emit_function.c
@@ -857,14 +857,6 @@ aot_compile_op_call(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
                 goto fail;
             }
 
-#if WASM_ENABLE_THREAD_MGR != 0
-            /* Insert suspend check point */
-            if (comp_ctx->enable_thread_mgr) {
-                if (!check_suspend_flags(comp_ctx, func_ctx))
-                    return false;
-            }
-#endif
-
             /* Check whether there was exception thrown when executing
                the function */
             if (!check_exception_thrown(comp_ctx, func_ctx)) {
@@ -1004,6 +996,14 @@ aot_compile_op_call(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
     if (comp_ctx->enable_aux_stack_frame) {
         if (!call_aot_free_frame_func(comp_ctx, func_ctx))
             goto fail;
+    }
+#endif
+
+#if WASM_ENABLE_THREAD_MGR != 0
+    /* Insert suspend check point */
+    if (comp_ctx->enable_thread_mgr) {
+        if (!check_suspend_flags(comp_ctx, func_ctx))
+            return false;
     }
 #endif
 
@@ -1650,6 +1650,14 @@ aot_compile_op_call_indirect(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
     if (comp_ctx->enable_aux_stack_frame) {
         if (!call_aot_free_frame_func(comp_ctx, func_ctx))
             goto fail;
+    }
+#endif
+
+#if WASM_ENABLE_THREAD_MGR != 0
+    /* Insert suspend check point */
+    if (comp_ctx->enable_thread_mgr) {
+        if (!check_suspend_flags(comp_ctx, func_ctx))
+            return false;
     }
 #endif
 
--- a/core/iwasm/compilation/aot_emit_function.c
+++ b/core/iwasm/compilation/aot_emit_function.c
@@ -857,6 +857,14 @@ aot_compile_op_call(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
                 goto fail;
             }
 
+#if WASM_ENABLE_THREAD_MGR != 0
+            /* Insert suspend check point */
+            if (comp_ctx->enable_thread_mgr) {
+                if (!check_suspend_flags(comp_ctx, func_ctx))
+                    return false;
+            }
+#endif
+
             /* Check whether there was exception thrown when executing
                the function */
             if (!check_exception_thrown(comp_ctx, func_ctx)) {
--- a/core/iwasm/interpreter/wasm_interp_classic.c
+++ b/core/iwasm/interpreter/wasm_interp_classic.c
@@ -3419,6 +3419,10 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                         ret = wasm_runtime_atomic_wait(
                             (WASMModuleInstanceCommon *)module, maddr,
                             (uint64)expect, timeout, false);
+
+#if WASM_ENABLE_THREAD_MGR != 0
+                        CHECK_SUSPEND_FLAGS();
+#endif
                         if (ret == (uint32)-1)
                             goto got_exception;
 
@@ -3439,6 +3443,10 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                         ret = wasm_runtime_atomic_wait(
                             (WASMModuleInstanceCommon *)module, maddr, expect,
                             timeout, true);
+
+#if WASM_ENABLE_THREAD_MGR != 0
+                        CHECK_SUSPEND_FLAGS();
+#endif
                         if (ret == (uint32)-1)
                             goto got_exception;
 
@@ -3894,10 +3902,10 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
             PUSH_CSP(LABEL_TYPE_FUNCTION, 0, cell_num, frame_ip_end - 1);
 
             wasm_exec_env_set_cur_frame(exec_env, frame);
-#if WASM_ENABLE_THREAD_MGR != 0
-            CHECK_SUSPEND_FLAGS();
-#endif
         }
+#if WASM_ENABLE_THREAD_MGR != 0
+        CHECK_SUSPEND_FLAGS();
+#endif
         HANDLE_OP_END();
     }
 
--- a/core/iwasm/compilation/aot_emit_memory.c
+++ b/core/iwasm/compilation/aot_emit_memory.c
@@ -7,6 +7,7 @@
 #include "aot_emit_exception.h"
 #include "../aot/aot_runtime.h"
 #include "aot_intrinsic.h"
+#include "aot_emit_control.h"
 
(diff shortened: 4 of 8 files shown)


core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -3716,7 +3716,9 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                           value: readv * 0 + sval */
                         DEF_ATOMIC_RMW_OPCODE(XCHG, *0 +);
                 }
-
+#if WASM_ENABLE_THREAD_MGR != 0
```
loganek, line 3719: isn't that a problem for the fast interpreter as well?

eloparco: of course, I forgot

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -3716,7 +3716,9 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+                CHECK_SUSPEND_FLAGS();
```
wenyongh, line 3720: Adding `CHECK_SUSPEND_FLAGS` here impacts each opcode of shared memory (atomic load/store, atomic wait/notify, rmw cmpxchg, etc.), it may greatly degrade the performance.

My understanding is that when main thread encounters proc_exit, it will throw and spread the exception, and notify all the atomic waiting nodes of other thread, and when the other thread receives the notification, it doesn't check the exception again, and runs into unreachable opcode, and overwrites the original proc_exit exception with unreachable exception.

We may add extra checks instead, e.g., in the handling of atomic.wait32/wait64 in interpreter:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_interp_fast.c#L3266-L3267
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_interp_fast.c#L3286-L3287
Change
```C
    if (ret == (uint32)-1)
        goto got_exception;
```
to
```C
    if (ret == (uint32)-1 || wasm_get_exception(module))
        goto got_exception;
```
Or directly modify wasm_runtime_atomic_notify:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/common/wasm_shared_memory.c#L482
Change `return is_timeout ? 2 : 0;` to
```C
if (wasm_runtime_get_exception(module))
    return -1;
return is_timeout ? 2 : 0;
```

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -3898,6 +3900,9 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
             CHECK_SUSPEND_FLAGS();
 #endif
         }
+#if WASM_ENABLE_THREAD_MGR != 0
+        CHECK_SUSPEND_FLAGS();
```
wenyongh, line 3904: Had better remove L3899 to L3901 to avoid duplicated check after calling non-import function.

core/iwasm/interpreter/wasm_interp_classic.c
eloparco: When the other thread receives the notification, we just want to check if the exception flag was set and, if that's the case, we want to return instead of fetching the next instruction using `HANDLE_OP_END`.
As you were pointing out, putting `CHECK_SUSPEND_FLAGS()` after each atomic instruction is too much overhead, we can just move it into the atomic wait opcode handling.

About `unreachable`, it just happened to be the next instruction in my test case
```wasm
i32.const 0
i32.const 0
i64.const -1
memory.atomic.wait32
unreachable
```

wenyongh: OK, had you tested the AOT mode? Seems that we should also add the suspend flag check after calling wasm_runtime_atomic_wait in AOT code:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/compilation/aot_emit_memory.c#L1370

Refer to:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/compilation/aot_emit_control.c#L753-L759

eloparco: Added a commit for that, but it doesn't work. Any suggestion for debugging AOT in general?

wenyongh (commented):
LGTM

eloparco:
> LGTM

Let me complete it before merging, it's not done yet since I need to fix AOT (and JIT as well probably)

core/iwasm/interpreter/wasm_interp_classic.c
wenyongh: Not a simple way to debug the AOT code now, normally we step into the assembler instructions to debug it.

Did you add `--enable-multi-thread` for wamrc to generate the AOT file?

eloparco: Do I need it? I'm testing using this sample but I replace the busy waiting with a `__builtin_wasm_memory_atomic_wait32(0, 0, -1)`
https://github.com/bytecodealliance/wasm-micro-runtime/blob/92c4bbebadb1589bda96d4365300723e06082dde/samples/wasi-threads/wasm-apps/thread_termination.c#L34

Without `--enable-multi-thread`, I see that it exits the wait but fetches the instruction after (i.e. what I want to avoid with this PR). The next instruction being this assert https://github.com/bytecodealliance/wasm-micro-runtime/blob/92c4bbebadb1589bda96d4365300723e06082dde/samples/wasi-threads/wasm-apps/thread_termination.c#L43

With `--enable-multi-thread` it crashes here https://github.com/bytecodealliance/wasm-micro-runtime/blob/92c4bbebadb1589bda96d4365300723e06082dde/core/iwasm/common/wasm_runtime_common.c#L197
But I don't think I need `--enable-multi-thread`, at least I wasn't using it with WASI threads in the past.

eloparco: Actually, I see problems in general when using AOT: this line
https://github.com/bytecodealliance/wasm-micro-runtime/blob/92c4bbebadb1589bda96d4365300723e06082dde/samples/wasi-threads/wasm-apps/no_pthread.c#L65
never gets executed. After some debugging, it looks like, after the atomic wait, the main thread ends and start waiting for the spawn thread https://github.com/bytecodealliance/wasm-micro-runtime/blob/92c4bbebadb1589bda96d4365300723e06082dde/core/iwasm/common/wasm_application.c#L218 without executing the instructions after the wait.

All this happens without the changes in `core/iwasm/compilation/aot_emit_memory.c` that I added with this PR.
Not sure how long the problem has been there.

eloparco: I think I found the problem, the atomic wait returns `-1` in case of error, instead here we check for `> 0`.
https://github.com/bytecodealliance/wasm-micro-runtime/blob/92c4bbebadb1589bda96d4365300723e06082dde/core/iwasm/compilation/aot_emit_memory.c#L1347
I quickly fixed it by using `LLVMIntSGE` instead of `LLVMIntSGT` and it seems to work. Now I need to do it properly by using a cast like here https://github.com/bytecodealliance/wasm-micro-runtime/blob/92c4bbebadb1589bda96d4365300723e06082dde/core/iwasm/interpreter/wasm_interp_classic.c#L3422

wenyongh: Can use `BUILD_ICMP(LLVMIntNE, ret_value, I32_NEG_ONE, cmp, "atomic_wait_ret");` for `if (ret == (uint32)-1`

wenyongh: > https://github.com/bytecodealliance/wasm-micro-runtime/blob/92c4bbebadb1589bda96d4365300723e06082dde/samples/wasi-threads/wasm-apps/thread_termination.c#L43
> 
> With `--enable-multi-thread` it crashes here
> 
> https://github.com/bytecodealliance/wasm-micro-runtime/blob/92c4bbebadb1589bda96d4365300723e06082dde/core/iwasm/common/wasm_runtime_common.c#L197
> 
The exception should have been set before accessing the guard page:
[wasm_runtime_access_exce_check_guard_page](https://github.com/bytecodealliance/wasm-micro-runtime/blob/92c4bbebadb1589bda96d4365300723e06082dde/core/iwasm/common/wasm_runtime_common.c#L1584)
Not sure whether it was cleared by `clear_wasi_proc_exit_exception`?

> 
> But I don't think I need `--enable-multi-thread`, at least I wasn't using it with WASI threads in the past.
> [EDIT] It is failing already at compilation time, `if (comp_ctx->enable_thread_mgr)` evaluates to false (without `--enable-multi-thread`), I'm debugging that.

It should be set, or `if (comp_ctx->enable_thread_mgr)` will be false.

eloparco: > The exception should have been set before accessing the guard page:
[wasm_runtime_access_exce_check_guard_page](https://github.com/bytecodealliance/wasm-micro-runtime/blob/92c4bbebadb1589bda96d4365300723e06082dde/core/iwasm/common/wasm_runtime_common.c#L1584)
Not sure whether it was cleared by clear_wasi_proc_exit_exception?

Yes, that's exactly what's happening. Depending on the run, sometimes the exception gets cleared before and the assert makes it crash. Sometimes it doesn't and the behavior is correct.

eloparco: Also, what's the equivalent of https://github.com/bytecodealliance/wasm-micro-runtime/blob/63273a1673d27df64159c0511f30e3047806a13b/core/iwasm/interpreter/wasm_interp_classic.c#L3821 in AOT mode? Because I need to check for termination there as well.

eloparco: I removed the assert and added the termination check in `aot_emit_function.c`. I ran a few test programs and it seems to work.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -194,7 +194,6 @@ runtime_signal_handler(void *sig_addr)
         else if (exec_env_tls->exce_check_guard_page <= (uint8 *)sig_addr
                  && (uint8 *)sig_addr
                         < exec_env_tls->exce_check_guard_page + page_size) {
-            bh_assert(wasm_get_exception(module_inst));
```
wenyongh, line 197: Had better not remove and modify to:
```C
    bh_assert(wasm_get_exception(module_inst)
              || wasm_cluster_is_thread_terminated(exec_env_tls));
#endif
);
```
And also change the similar place for Windows (this file, L253):
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/common/wasm_runtime_common.c#L253

core/iwasm/compilation/aot_emit_function.c
```diff
@@ -857,6 +857,14 @@ aot_compile_op_call(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
                 goto fail;
             }
 
+#if WASM_ENABLE_THREAD_MGR != 0
+            /* Insert suspend check point */
+            if (comp_ctx->enable_thread_mgr) {
+                if (!check_suspend_flags(comp_ctx, func_ctx))
+                    return false;
+            }
+#endif
```
wenyongh, line 866: If to make the similar changes like interpreter for AOT, the place to add the check should be:
the end of aot_compile_op_call (after returning from the jit function):
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/compilation/aot_emit_function.c#L1001
and the end of aot_compile_op_call_indirect (after returning from the jit function):
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/compilation/aot_emit_function.c#L1647

Note that the suspend flags was also checked at the beginning of the two functions, or, before calling the jit function:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/compilation/aot_emit_function.c#L641
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/compilation/aot_emit_function.c#L1447

But adding the new checks may impact the performance of AOT/JIT when multi-thread is enabled, can we make sure that we exactly need to add these checks (both for interpreter and JIT/AOT)?

eloparco: > Note that the suspend flags was also checked at the beginning of the two functions, or, before calling the jit function:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/compilation/aot_emit_function.c#L641
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/compilation/aot_emit_function.c#L1447

> But adding the new checks may impact the performance of AOT/JIT when multi-thread is enabled, can we make sure that we exactly need to add these checks (both for interpreter and JIT/AOT)?

I tested it and it doesn't work without adding the checks at L1001 and L1647. We need to check for termination right after the function call, otherwise the next instruction will be fetched and we want to avoid that.

wenyongh: OK, let's add them.

core/iwasm/common/wasm_runtime_common.c
eloparco: Right, I'm debugging to understand why `exec_env_tls->suspend_flags` is not set properly once I get there.

core/iwasm/interpreter/wasm_interp_classic.c
eloparco: Ok, the problem was that we don't set the suspend flag for the thread that originally raises the exception.

core/iwasm/common/wasm_runtime_common.c
eloparco: Ok, the problem was that we don't set the suspend flag for the thread that originally raises the exception.

eloparco: The problem is that the assert would be executed also on the thread that originally raises the exception and that thread doesn't have the suspend flag set. I tried setting the suspend flag on that initial thread too but it break things since it forces the initial thread to exit prematurely (after a `CHECK_SUSPEND_FLAGS();`) while we don't want that.

So I was thinking of just having a
```c
    bh_assert(wasm_get_exception(module_inst));
#endif
```

wenyongh: Not sure how to reproduce the issue? The thread_terminate sample with TEST_TERMINATION_BY_TRAP == 1 and TEST_TERMINATION_IN_MAIN_THREAD == 1?

core/iwasm/compilation/aot_emit_function.c
```diff
@@ -999,6 +999,14 @@ aot_compile_op_call(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
     }
 #endif
 
+#if WASM_ENABLE_THREAD_MGR != 0
+    /* Insert suspend check point */
+    if (comp_ctx->enable_thread_mgr) {
+        if (!check_suspend_flags(comp_ctx, func_ctx))
+            return false;
```
wenyongh, line 1006: Should `goto fail` to destroy the resource

core/iwasm/compilation/aot_emit_function.c
```diff
@@ -1645,6 +1653,14 @@ aot_compile_op_call_indirect(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
...
```
wenyongh, line 1660: goto fail;

core/iwasm/common/wasm_runtime_common.c
eloparco: I updated the thread termination sample, you should be able to reproduce the problem (in AOT mode) by running it (no changes required). Let me know if you have issues reproducing it.

wenyongh: Thanks @eloparco, I can reproduce the issue now. The process is that:
the main thread met wasi_proc_exit, raised that exception, and spread it to child threads
the main thread accessed the guard page to trigger signal handler, but hadn't entered the signal handler
the child thread got exception and ended running, and it founds that the exception is `wasi proc exit` excpeiton, so it tried to clear it, and during the clearing, **it spread null exception to other threads including main thread**
the main thread run into signal handler, but the exception had been cleared, so bh_assert failed

So the root cause is that the main thread's `wasi proc exit` exception is cleared by child thread.
It seems a better solution is not to spread the exception `wasi proc exit` to other threads, but just set the suspend flags of other threads to let them exit.

I think we may merge this PR firstly if you are OK and we will submit another PR to fix that issue.

eloparco: > I think we may merge this PR firstly if you are OK and we will submit another PR to fix that issue.

Let's do it, so that I can move to #1985 that hopefully should succeed after we merge this.

wenyongh: @eloparco I tried again, and think this is not a best way to resolve the issue. The root cause is that the main thread's "wasi proc exit" exception is cleared by the sub thread after the latter gets the spread exception from main thread and then clear exception for all threads.
It is strange to spread the "wasi proc exit" exception to other threads and then clear it in other threads. I think we can just ignore spreading of this exception and no need to clear it again. I uploaded PR #1988 to fix the issues, I tested the cases and didn't find the issues you mentioned (unreachable exception was thrown and bh_assert failure), could you help review and try? Thanks.

eloparco: Can we merge this one and rebase #1988? I'll then try #1988 in a few different cases to see if we don't have regressions.

wenyongh: My personal preference is we had better not merge a temporarily PR which doesn't fix the root cause, #1988 should be better and have resolved the issues mentioned in this PR. I tend to merge #1988 directly.

eloparco: Right, it makes sense. Let's close this one once #1988 gets merged

eloparco:
Closing since the changes here have been applied to #1988

PR #1980, opened 2023-02-21
Fix timeout in poll_oneoff

eloparco:
Two fixes in the `poll_oneoff` implementation:
-  `wasm_runtime_get_exception()` was being called after the `proc_exit` exception was cleared https://github.com/bytecodealliance/wasm-micro-runtime/blob/bb5629811fd8f22825a935f25f4af721cef6051e/core/iwasm/common/wasm_runtime_common.c#L1870
therefore it was always returning `NULL`; `suspend_flags` should be used instead to check for termination;
- `elapsed` was incremented before the call to `wasmtime_ssp_poll_oneoff()`

Follow-up on #1951. Tested with multiple timeout values, with and without interruption and measuring the time spent sleeping.

### DIFF
--- a/core/iwasm/include/wasm_export.h
+++ b/core/iwasm/include/wasm_export.h
@@ -717,6 +717,15 @@ WASM_RUNTIME_API_EXTERN wasm_module_inst_t
 wasm_runtime_get_module_inst(wasm_exec_env_t exec_env);
 
 /**
+ * Get suspend flags from the execution environment
+ * 
+ * @param exec_env the execution environment to get the suspend flags from
+ * @return the suspend flags 
+ */
+WASM_RUNTIME_API_EXTERN uint32_t
+wasm_runtime_get_suspend_flags(wasm_exec_env_t exec_env);
+
+/**
  * Set WASM module instance of execution environment
  * Caution:
  *   normally the module instance is bound with the execution
--- a/core/iwasm/include/wasm_export.h
+++ b/core/iwasm/include/wasm_export.h
@@ -717,15 +717,6 @@ WASM_RUNTIME_API_EXTERN wasm_module_inst_t
 wasm_runtime_get_module_inst(wasm_exec_env_t exec_env);
 
 /**
- * Get suspend flags from the execution environment
- * 
- * @param exec_env the execution environment to get the suspend flags from
- * @return the suspend flags 
- */
-WASM_RUNTIME_API_EXTERN uint32_t
-wasm_runtime_get_suspend_flags(wasm_exec_env_t exec_env);
-
-/**
  * Set WASM module instance of execution environment
  * Caution:
  *   normally the module instance is bound with the execution
--- a/core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
+++ b/core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
@@ -10,7 +10,31 @@
 #if WASM_ENABLE_THREAD_MGR != 0
 #include "../../../thread-mgr/thread_manager.h"
 typedef struct WASIContext *wasi_ctx_t;
+#else
+typedef struct WASIContext {
+    struct fd_table *curfds;
+    struct fd_prestats *prestats;
+    struct argv_environ_values *argv_environ;
+    struct addr_pool *addr_pool;
+    char *ns_lookup_buf;
+    char **ns_lookup_list;
+    char *argv_buf;
+    char **argv_list;
+    char *env_buf;
+    char **env_list;
+    uint32_t exit_code;
+} * wasi_ctx_t;
 #endif
+
+typedef struct wasi_prestat_app {
+    wasi_preopentype_t pr_type;
+    uint32 pr_name_len;
+} wasi_prestat_app_t;
+
+typedef struct iovec_app {
+    uint32 buf_offset;
+    uint32 buf_len;
+} iovec_app_t;
 
 void
 wasm_runtime_set_exception(wasm_module_inst_t module, const char *exception);
@@ -40,33 +64,6 @@ wasm_runtime_set_exception(wasm_module_inst_t module, const char *exception);
 #define module_free(offset) \
     wasm_runtime_module_free(module_inst, offset)
 /* clang-format on */
-
-typedef struct wasi_prestat_app {
-    wasi_preopentype_t pr_type;
-    uint32 pr_name_len;
-} wasi_prestat_app_t;
-
-typedef struct iovec_app {
-    uint32 buf_offset;
-    uint32 buf_len;
-} iovec_app_t;
-
-/* If the thread manager is enabled, WASIContext is already defined */
-#if WASM_ENABLE_THREAD_MGR == 0
-typedef struct WASIContext {
-    struct fd_table *curfds;
-    struct fd_prestats *prestats;
-    struct argv_environ_values *argv_environ;
-    struct addr_pool *addr_pool;
-    char *ns_lookup_buf;
-    char **ns_lookup_list;
-    char *argv_buf;
-    char **argv_list;
-    char *env_buf;
-    char **env_list;
-    uint32_t exit_code;
-} * wasi_ctx_t;
-#endif
 
 wasi_ctx_t
 wasm_runtime_get_wasi_ctx(wasm_module_inst_t module_inst);
--- a/core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
+++ b/core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
@@ -1033,9 +1033,11 @@ execute_interruptible_poll_oneoff(
         }
         else if (*nevents > 0) {
             all_outs_are_type_clock = true;
-            for (i = 0; i < nsubscriptions; i++) {
-                if (out[i].type != __WASI_EVENTTYPE_CLOCK)
+            for (i = 0; i < *nevents; i++) {
+                if (out[i].type != __WASI_EVENTTYPE_CLOCK) {
                     all_outs_are_type_clock = false;
+                    break;
+                }
             }
 
             if (!all_outs_are_type_clock) {
--- a/core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
+++ b/core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
@@ -1042,6 +1042,12 @@ execute_interruptible_poll_oneoff(wasm_module_inst_t module_inst,
             wasm_runtime_free(in_copy);
             return EINTR;
         }
+        else if (*nevents > 0
+                 && !(nsubscriptions == 1
+                      && in[0].u.type == __WASI_EVENTTYPE_CLOCK)) {
+            wasm_runtime_free(in_copy);
+            return __WASI_ESUCCESS;
+        }
     }
 
     wasm_runtime_free(in_copy);
--- a/core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
+++ b/core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
@@ -7,6 +7,11 @@
 #include "bh_platform.h"
 #include "wasm_export.h"
 
+#if WASM_ENABLE_THREAD_MGR != 0
+#include "../../../thread-mgr/thread_manager.h"
+typedef struct WASIContext *wasi_ctx_t;
+#endif
+
 void
 wasm_runtime_set_exception(wasm_module_inst_t module, const char *exception);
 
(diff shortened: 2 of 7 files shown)


core/iwasm/include/wasm_export.h
```diff
@@ -716,6 +716,15 @@ wasm_runtime_thread_env_inited(void);
 WASM_RUNTIME_API_EXTERN wasm_module_inst_t
 wasm_runtime_get_module_inst(wasm_exec_env_t exec_env);
 
+/**
+ * Get suspend flags from the execution environment
+ * 
+ * @param exec_env the execution environment to get the suspend flags from
+ * @return the suspend flags 
+ */
+WASM_RUNTIME_API_EXTERN uint32_t
```
wenyongh, line 725: Is it a little heavy to exposing the API?
How about adding API in `core/iwasm/libraries/thread-mgr/thread_manager.h/.c`:

```C
bool
wasm_cluster_is_thread_terminated(WASMExecEnv *exec_env)
{
    return (exec_env->suspend_flags.flags & 0x01) ? true : false;
}
```
And call it in libc_wasi_wrapper.c:
```C
#include "../../../thread-mgr/thread_manager.h"
#endif
```

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -1021,24 +1024,20 @@ execute_interruptible_poll_oneoff(wasm_module_inst_t module_inst,
...
                                        nevents);
+        elapsed += time_quant;
+
         if (err) {
             wasm_runtime_free(in_copy);
             return err;
         }
 
-        if (wasm_runtime_get_exception(module_inst) || *nevents > 0) {
+        if (get_suspend_flags(exec_env) & 0x01) {
```
wenyongh, line 1039: Should also handle `*nevents > 0`:
```
if (thread is terminated) {
    ...
}
else if (*nevents > 0) {
    wasm_runtime_free(in_copy);
    return __WASI_ESUCCESS;
}

File: core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c [hunk 3]
```diff
@@ -46,6 +51,8 @@ typedef struct iovec_app {
     uint32 buf_len;
 } iovec_app_t;
 
+/* If the thread manager is enabled, WASIContext is already defined */
+#if WASM_ENABLE_THREAD_MGR == 0
```
eloparco, line 55: In alternative, we can put L55-L69 in an else at L13 to make things clearer. I didn't do it to keep it in the same place as the other `typedef`s.

loganek: could we just delete those typedefs and unconditionally include '`wasm_runtime_common.h`?

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -7,6 +7,11 @@
 #include "bh_platform.h"
 #include "wasm_export.h"
 
+#if WASM_ENABLE_THREAD_MGR != 0
+#include "../../../thread-mgr/thread_manager.h"
```
loganek, line 11: why do we need to include thread manager here?

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
loganek: > I removed it on purpose since *nevents evaluates to 1 already at the first iteration and that ends the loop immediately

If that is the case, that means there was an event and we should indeed end the loop. That's how `poll()` works in general - it only blocks until either it reaches the timeout OR it detects an event.

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
eloparco: I was referring to the two typedefs here (that are libc_wasi_wrapper specific) https://github.com/bytecodealliance/wasm-micro-runtime/blob/d2772c4153a5ca27b33df3e3d5f0aa4c804234a1/core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c#L39
Maybe I can just move them after the includes so that I can put L55-L69 in an else at L13.

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
eloparco: It seems like it's always set to `1`
https://github.com/bytecodealliance/wasm-micro-runtime/blob/d2772c4153a5ca27b33df3e3d5f0aa4c804234a1/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c#L2615

wenyongh: > It seems like it's always set to `1`
> 
> https://github.com/bytecodealliance/wasm-micro-runtime/blob/d2772c4153a5ca27b33df3e3d5f0aa4c804234a1/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c#L2615

It is set to `1` only when `if (nsubscriptions == 1 && in[0].u.type == __WASI_EVENTTYPE_CLOCK) {`:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/d2772c4153a5ca27b33df3e3d5f0aa4c804234a1/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c#LL2542C5-L2542C73

If `nsubscriptions > 1`, it will run into L2624:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/d2772c4153a5ca27b33df3e3d5f0aa4c804234a1/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c#L2624

and `*nevents` may be increased
https://github.com/bytecodealliance/wasm-micro-runtime/blob/d2772c4153a5ca27b33df3e3d5f0aa4c804234a1/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c#L2738

https://github.com/bytecodealliance/wasm-micro-runtime/blob/d2772c4153a5ca27b33df3e3d5f0aa4c804234a1/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c#L2752

eloparco: so the condition should be something like:
```c
else if (*nevents > 0 && (nsubscriptions > 1 || in[0].u.type != __WASI_EVENTTYPE_CLOCK)) {
    wasm_runtime_free(in_copy);
    return __WASI_ESUCCESS;
}
```

wenyongh: Had better `else if (*nevents > 0 && !(nsubscriptions == 1 && in[0].u.type == __WASI_EVENTTYPE_CLOCK))`

wenyongh: > It seems like it's always set to `1`
> 
> https://github.com/bytecodealliance/wasm-micro-runtime/blob/d2772c4153a5ca27b33df3e3d5f0aa4c804234a1/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c#L2615

Seems that it is a bug of posix.c when `if (nsubscriptions == 1 && in[0].u.type == __WASI_EVENTTYPE_CLOCK) {`, it always returns 0 even when error occurs, which is different from the handling of the non `if (nsubscriptions == 1 && in[0].u.type == __WASI_EVENTTYPE_CLOCK) {`,  the latter sets `error` to 0 first, and converts `errno` to `error` when poll returns -1, and eventually returns `error`:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/d2772c4153a5ca27b33df3e3d5f0aa4c804234a1/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c#L2712-L2715

My personal understanding is that we should also apply similar changes in the former, for example, change
https://github.com/bytecodealliance/wasm-micro-runtime/blob/d2772c4153a5ca27b33df3e3d5f0aa4c804234a1/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c#L2615-L2616
to:
```C
    *nevents = 1;
    if (out[0].error != 0)
        return convert_error(out[0].error);
    return 0;
```

And here, restore to
```C
if (thread is terminated) {
    ...
}
else if (*nevents > 0) {
    wasm_runtime_free(in_copy);
    return __WASI_ESUCCESS;
}
```
Not sure it is right, could you help review and test?

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -1070,7 +1077,7 @@ wasi_poll_oneoff(wasm_exec_env_t exec_env, const wasi_subscription_t *in,
     err = wasmtime_ssp_poll_oneoff(curfds, in, out, nsubscriptions, &nevents);
 #else
     err = execute_interruptible_poll_oneoff(module_inst, curfds, in, out,
-                                            nsubscriptions, &nevents);
+                                            nsubscriptions, &nevents, exec_env);
```
hritikgupta, line 1080: we probably don't need to pass `module_inst` if using `wasm_cluster_is_thread_terminated`

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
eloparco: It makes sense, but how does that make the `!(nsubscriptions == 1 && in[0].u.type == __WASI_EVENTTYPE_CLOCK)` part in the else unnecessary? With your proposed changes we modify the return code, but `*nevents` is still set to `1`, making the loop always return `__WASI_ESUCCESS` at the first iteration

wenyongh: Seems you are right, we divide total timeout into many pieces, for the case of `nsubscriptions == 1 && in[0].u.type == __WASI_EVENTTYPE_CLOCK`, it should wait many times until the total timeout is reached.
Seems we should change the first part of `wasmtime_ssp_poll_oneoff` to:
```C
    *nevents = 1;
    if (out[0].error != 0)
        return convert_error(out[0].error);
    return 0;
```
and here keep:
```C
if (thread is terminated) {
    ...
}
else if (*nevents > 0 && !(nsubscriptions == 1 && in[0].u.type == __WASI_EVENTTYPE_CLOCK)) {
    wasm_runtime_free(in_copy);
    return __WASI_ESUCCESS;
}
```
Is that right?

eloparco: like this?
```c
else if (*nevents > 0) {
    all_outs_are_type_clock = true;
    for (i = 0; i < nsubscriptions; i++) {
        if (out[i].type != __WASI_EVENTTYPE_CLOCK)
            all_outs_are_type_clock = false;
    }

    if (!all_outs_are_type_clock) {
        wasm_runtime_free(in_copy);
        return __WASI_ESUCCESS;
    }
}
```

eloparco: Ok, I added that

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
eloparco: > could we just delete those typedefs and unconditionally include 'wasm_runtime_common.h

right, done

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -1021,25 +1015,33 @@ execute_interruptible_poll_oneoff(wasm_module_inst_t module_inst,
...
         }
 
-        if (wasm_runtime_get_exception(module_inst) || *nevents > 0) {
+        if (wasm_cluster_is_thread_terminated(exec_env)) {
             wasm_runtime_free(in_copy);
+            return EINTR;
+        }
+        else if (*nevents > 0) {
+            all_outs_are_type_clock = true;
+            for (i = 0; i < nsubscriptions; i++) {
```
wenyongh, line 1036: Should be `i < *nevents`

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -1021,25 +1015,33 @@ execute_interruptible_poll_oneoff(wasm_module_inst_t module_inst,
...
+                if (out[i].type != __WASI_EVENTTYPE_CLOCK)
+                    all_outs_are_type_clock = false;
```
wenyongh, line 1038: Had better `break` when found:
```C
                if (out[i].type != __WASI_EVENTTYPE_CLOCK) {
                    all_outs_are_type_clock = false;
                    break;
                }
```

wenyongh (commented):
LGTM

PR #1981, opened 2023-02-22
fix random bind socket

xujuntwt95329:
Previously, the *port is assigned before `getsockname`, can't get actual port number assigned by system.

This PR assign the *port after calling `getsockname`, then the caller can get the actual port number

### DIFF
--- a/core/shared/platform/common/posix/posix_socket.c
+++ b/core/shared/platform/common/posix/posix_socket.c
@@ -156,17 +156,6 @@ os_socket_bind(bh_socket_t socket, const char *host, int *port)
         goto fail;
     }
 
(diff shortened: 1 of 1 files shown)


loganek:
Looks good, thanks. Just to clarify - this is only a problem when the user doesn't explicitly request a port (i.e. set port to 0), right?

loganek (approved):

xujuntwt95329:
> Looks good, thanks. Just to clarify - this is only a problem when the user doesn't explicitly request a port (i.e. set port to 0), right?

@loganek  Yes, user can set port to 0, in this case the API should return the actual port assigned by OS

PR #1988, opened 2023-02-24
thread-mgr: Fix spread "wasi proc exit" exception and atomic.wait issues

wenyongh:
Raising "wasi proc exit" exception, spreading it to other threads and then
clearing it in all threads may result in unexpected behavior: the sub thread
may end first, handle the "wasi proc exit" exception and clear exceptions
of other threads, including the main thread. And when main thread's
exception is cleared, it may continue to run and throw "unreachable"
exception. This also leads to some assertion failed.

Ignore exception spreading for "wasi proc exit" and don't clear exception
of other threads  to resolve the issue.

And add suspend flag check after atomic wait since the atomic wait may
be notified by other thread when exception occurs.

### DIFF
--- a/core/iwasm/compilation/aot_emit_memory.c
+++ b/core/iwasm/compilation/aot_emit_memory.c
@@ -1345,14 +1345,6 @@ aot_compile_op_atomic_wait(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
         return false;
     }
 
-#if WASM_ENABLE_THREAD_MGR != 0
-    /* Insert suspend check point */
-    if (comp_ctx->enable_thread_mgr) {
-        if (!check_suspend_flags(comp_ctx, func_ctx))
-            return false;
-    }
-#endif
-
     BUILD_ICMP(LLVMIntNE, ret_value, I32_NEG_ONE, cmp, "atomic_wait_ret");
 
     ADD_BASIC_BLOCK(wait_fail, "atomic_wait_fail");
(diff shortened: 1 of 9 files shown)


core/iwasm/compilation/aot_emit_memory.c
```diff
@@ -1368,6 +1369,14 @@ aot_compile_op_atomic_wait(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
 
     PUSH_I32(ret_value);
 
+#if WASM_ENABLE_THREAD_MGR != 0
```
eloparco, line 1372: We also need this change in `aot_emit_function.c`, as in #1975.

wenyongh: As the main thread's exception won't be cleared by sub thread, main thread will exit after returning the function to the caller, and won't fetch the `unreachable` opcode and throw "unreachable" exception again. It works for the test cases, for the performance reason, I didn't add the check for AOT.

eloparco: The change in `aot_emit_function.c` is not needed for the main thread, but for the case in which a spawned thread is waiting on a `poll_oneoff`. You can reproduce it with this example https://github.com/WebAssembly/wasi-threads/blob/main/test/testsuite/wasi_threads_exit_main_wasi.wat. We want to exit after the `poll_oneoff` gets interrupted without fetching the next instruction (that is `unreachable`) https://github.com/WebAssembly/wasi-threads/blob/c1b421c2158b745c840b5bdca98ed870c2a6d11a/test/testsuite/wasi_threads_exit_main_wasi.wat#L26

wenyongh: OK, I added the checks in aot_emit_function.c.

eloparco (approved):

eloparco:
LGTM

loganek (approved):

eloparco:
It seems like the pipeline got stuck, should we restart it? @wenyongh

PR #1985, opened 2023-02-23
Update WASI thread proposal tests in CI

eloparco:
Update CI to run extended WASI threads tests included in the proposal https://github.com/WebAssembly/wasi-testsuite/tree/prod/testsuite-all/tests/proposals/wasi-threads

PR #1991, opened 2023-02-26
Protect access to exec env

eloparco:
Fix concurrent access to exec env: it is accessed by the thread setting the suspend flags (to stop other threads) and threads (in the same cluster) reading it to check for termination. 

Tested it by using the thread sanitizer and running WASI tests in the proposal https://github.com/WebAssembly/wasi-threads/tree/main/test/testsuite.

### DIFF
--- a/core/iwasm/common/wasm_exec_env.c
+++ b/core/iwasm/common/wasm_exec_env.c
@@ -208,10 +208,20 @@ void
 wasm_exec_env_set_thread_info(WASMExecEnv *exec_env)
 {
     uint8 *stack_boundary = os_thread_get_stack_boundary();
+
+#if WASM_ENABLE_THREAD_MGR != 0
+    WASMCluster *cluster = wasm_exec_env_get_cluster(exec_env);
+    if (cluster)
+        os_mutex_lock(&cluster->lock);
+#endif
     exec_env->handle = os_self_thread();
     exec_env->native_stack_boundary =
         stack_boundary ? stack_boundary + WASM_STACK_GUARD_SIZE : NULL;
     exec_env->native_stack_top_min = (void *)UINTPTR_MAX;
+#if WASM_ENABLE_THREAD_MGR != 0
+    if (cluster)
+        os_mutex_unlock(&cluster->lock);
+#endif
 }
 
 #if WASM_ENABLE_THREAD_MGR != 0
--- a/core/iwasm/common/wasm_exec_env.c
+++ b/core/iwasm/common/wasm_exec_env.c
@@ -210,17 +210,14 @@ wasm_exec_env_set_thread_info(WASMExecEnv *exec_env)
     uint8 *stack_boundary = os_thread_get_stack_boundary();
 
 #if WASM_ENABLE_THREAD_MGR != 0
-    WASMCluster *cluster = wasm_exec_env_get_cluster(exec_env);
-    if (cluster)
-        os_mutex_lock(&cluster->lock);
+    os_mutex_lock(&exec_env->wait_lock);
 #endif
     exec_env->handle = os_self_thread();
     exec_env->native_stack_boundary =
         stack_boundary ? stack_boundary + WASM_STACK_GUARD_SIZE : NULL;
     exec_env->native_stack_top_min = (void *)UINTPTR_MAX;
 #if WASM_ENABLE_THREAD_MGR != 0
-    if (cluster)
-        os_mutex_unlock(&cluster->lock);
+    os_mutex_unlock(&exec_env->wait_lock);
 #endif
 }
 
--- a/core/iwasm/interpreter/wasm_interp_classic.c
+++ b/core/iwasm/interpreter/wasm_interp_classic.c
@@ -12,10 +12,8 @@
 #if WASM_ENABLE_SHARED_MEMORY != 0
 #include "../common/wasm_shared_memory.h"
 #endif
-#if WASM_ENABLE_THREAD_MGR != 0
-#include "../libraries/thread-mgr/thread_manager.h"
-#endif
 #if WASM_ENABLE_THREAD_MGR != 0 && WASM_ENABLE_DEBUG_INTERP != 0
+#include "../libraries/thread-mgr/thread_manager.h"
 #include "../libraries/debug-engine/debug_engine.h"
 #endif
 #if WASM_ENABLE_FAST_JIT != 0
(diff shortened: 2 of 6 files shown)


core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -1036,27 +1038,41 @@ wasm_interp_call_func_import(WASMModuleInstance *module_inst,
 #if WASM_ENABLE_DEBUG_INTERP != 0
 #define CHECK_SUSPEND_FLAGS()                                          \
     do {                                                               \
+        WASMCluster *cluster = wasm_exec_env_get_cluster(exec_env);    \
+        if (cluster)                                                   \
+            os_mutex_lock(&cluster->lock);                             \
```
wenyongh, line 1043: `cluster->lock` is used to lock the whole exec_env list in the cluster, using it here is a little heavy and may cause dead lock (e.g., other thread is suspending all other threads in future GC feature, which may lock `cluster->lock` firstly).

Suggest to use `exec_env->wait_lock` instead. And had better add `exec_env->wait_lock` for the related read/write operations of suspend_flag/signal_flag?

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -1027,7 +1027,18 @@ execute_interruptible_poll_oneoff(
             return err;
         }
 
-        if (wasm_cluster_is_thread_terminated(exec_env)) {
+#if WASM_ENABLE_THREAD_MGR != 0
+        WASMCluster *cluster = wasm_exec_env_get_cluster(exec_env);
+        if (cluster)
+            os_mutex_lock(&cluster->lock);
+#endif
+        bool is_thread_terminated = wasm_cluster_is_thread_terminated(exec_env);
```
wenyongh, line 1035: Suggest to lock `exec_env->wait_lock` in wasm_cluster_is_thread_terminated and keep this file unchanged.

core/iwasm/common/wasm_exec_env.c
```diff
@@ -208,10 +208,20 @@ void
 wasm_exec_env_set_thread_info(WASMExecEnv *exec_env)
 {
     uint8 *stack_boundary = os_thread_get_stack_boundary();
+
+#if WASM_ENABLE_THREAD_MGR != 0
+    WASMCluster *cluster = wasm_exec_env_get_cluster(exec_env);
+    if (cluster)
+        os_mutex_lock(&cluster->lock);
```
wenyongh, line 215: Shall we add lock for these operations? It seems these field are only read later by this thread.

eloparco: These two are the operations manipulating it in different threads
https://github.com/bytecodealliance/wasm-micro-runtime/blob/38c67b3f4815a628ce0f9bc22186bface4678089/core/iwasm/libraries/thread-mgr/thread_manager.c#L858

https://github.com/bytecodealliance/wasm-micro-runtime/blob/38c67b3f4815a628ce0f9bc22186bface4678089/core/iwasm/common/wasm_exec_env.c#L211

wenyongh: OK, could we use `exec_env->wait_lock` to lock it in the two places?

eloparco: Yes, pushed a commit with the change, it makes more sense to use the more specific/fine-grained lock in the exec_env instead of the cluster one.

I tested the change in classic interpreter, fast interpreter and aot modes running the tests in #1985 with the thread sanitizer and no data race is reported anymore.

core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
```diff
@@ -123,14 +123,12 @@ thread_spawn_wrapper(wasm_exec_env_t exec_env, uint32 start_arg)
...
     thread_start_arg->start_func = start_func;
 
-    os_mutex_lock(&exec_env->wait_lock);
     ret = wasm_cluster_create_thread(exec_env, new_module_inst, false,
                                      thread_start, thread_start_arg);
     if (ret != 0) {
         LOG_ERROR("Failed to spawn a new thread");
         goto thread_spawn_fail;
     }
-    os_mutex_unlock(&exec_env->wait_lock);
```
wenyongh, line 133: Is there data race here?
Had better not remove the lock, instead, add a cond_wait like pthread_create_wrapper in lib_pthread_wrapper.c:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/38c67b3f4815a628ce0f9bc22186bface4678089/core/iwasm/libraries/lib-pthread/lib_pthread_wrapper.c#L665-L668

And in the `thread_start`, add cond_signal like pthread_start_routine:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/38c67b3f4815a628ce0f9bc22186bface4678089/core/iwasm/libraries/lib-pthread/lib_pthread_wrapper.c#L509

eloparco: The reason why I removed it is that, after getting the `exec_env` lock there, the `cluster` lock gets acquired https://github.com/bytecodealliance/wasm-micro-runtime/blob/38c67b3f4815a628ce0f9bc22186bface4678089/core/iwasm/libraries/thread-mgr/thread_manager.c#L623
but then an attempt to get the `exec_env` lock is made again at https://github.com/eloparco/wasm-micro-runtime/blob/5550b5647037a7cc9674dda03413a95e3c06cba3/core/iwasm/libraries/thread-mgr/thread_manager.c#L946 generating a warning in the sanitizer for potential deadlock.

Let me see how it can be fixed.

wenyongh: It seems that the thread isn't actually started, but it is already added to the cluster's exec_env list, and another thread (seems main thread) tries to terminate it? Maybe after adding the similar logic like pthread_create_wrapper, the issue can be resolved since the main thread will wait unit the child thread actually started.
Here try adding cond_wait:
```C
    ret = wasm_cluster_create_thread(exec_env, new_module_inst, false,
                                     thread_start, thread_start_arg);
    ...
    os_cond_wait(&exec_env->wait_cond, &exec_env->wait_lock);
    os_mutex_unlock(&exec_env->wait_lock);
```
And in thread_start:
```
    os_mutex_lock(&parent_exec_env->wait_lock);
    ...
    os_cond_signal(&parent_exec_env->wait_cond, &parent_exec_env->wait_lock);
    os_mutex_unlock(&parent_exec_env->wait_lock);
```

eloparco: It actually only happens when it's the spawn thread that calls `proc_exit`.
https://github.com/bytecodealliance/wasm-micro-runtime/blob/38c67b3f4815a628ce0f9bc22186bface4678089/core/iwasm/libraries/lib-pthread/lib_pthread_wrapper.c#L665-L668
I see that for pthreads that lock is used to wait for the update of `ThreadInfoNode`.
In `lib_wasi_threads_wrapper.c` we don't use `ThreadInfoNode` so I don't think the lock there is even needed.

wenyongh: OK, got it, thanks.

wenyongh (approved):
LGTM

core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
eloparco: Great, once this one is merged, we can merge #1985 too

PR #1995, opened 2023-03-01
Add/reorganize locks for thread synchronization

eloparco:
Attempt to fix data races when using threads.
- Protect access (from multiple threads) to exception and memory
- Fix shared memory lock usage

### DIFF
--- a/core/iwasm/common/wasm_memory.c
+++ b/core/iwasm/common/wasm_memory.c
@@ -721,3 +721,39 @@ wasm_enlarge_memory(WASMModuleInstance *module, uint32 inc_page_count)
 
     return ret;
 }
+
+#if !defined(OS_ENABLE_HW_BOUND_CHECK)              \
+    || WASM_CPU_SUPPORTS_UNALIGNED_ADDR_ACCESS == 0 \
+    || WASM_ENABLE_BULK_MEMORY != 0
+uint32
+get_num_bytes_per_page(WASMMemoryInstance *memory, void *node)
+{
+    uint32 num_bytes_per_page;
+#if WASM_ENABLE_SHARED_MEMORY != 0
+    if (node)
+        os_mutex_lock(&((WASMSharedMemNode *)node)->shared_mem_lock);
+#endif
+    num_bytes_per_page = memory->num_bytes_per_page;
+#if WASM_ENABLE_SHARED_MEMORY != 0
+    if (node)
+        os_mutex_unlock(&((WASMSharedMemNode *)node)->shared_mem_lock);
+#endif
+    return num_bytes_per_page;
+}
+
+uint32
+get_linear_memory_size(WASMMemoryInstance *memory, void *node)
+{
+    uint32 linear_mem_size;
+#if WASM_ENABLE_SHARED_MEMORY != 0
+    if (node)
+        os_mutex_lock(&((WASMSharedMemNode *)node)->shared_mem_lock);
+#endif
+    linear_mem_size = memory->num_bytes_per_page * memory->cur_page_count;
+#if WASM_ENABLE_SHARED_MEMORY != 0
+    if (node)
+        os_mutex_unlock(&((WASMSharedMemNode *)node)->shared_mem_lock);
+#endif
+    return linear_mem_size;
+}
+#endif
--- a/core/iwasm/common/wasm_memory.c
+++ b/core/iwasm/common/wasm_memory.c
@@ -726,7 +726,7 @@ wasm_enlarge_memory(WASMModuleInstance *module, uint32 inc_page_count)
     || WASM_CPU_SUPPORTS_UNALIGNED_ADDR_ACCESS == 0 \
     || WASM_ENABLE_BULK_MEMORY != 0
 uint32
-get_num_bytes_per_page(WASMMemoryInstance *memory, void *node)
+wasm_get_num_bytes_per_page(WASMMemoryInstance *memory, void *node)
 {
     uint32 num_bytes_per_page;
 #if WASM_ENABLE_SHARED_MEMORY != 0
@@ -742,7 +742,7 @@ get_num_bytes_per_page(WASMMemoryInstance *memory, void *node)
 }
 
 uint32
-get_linear_memory_size(WASMMemoryInstance *memory, void *node)
+wasm_get_linear_memory_size(WASMMemoryInstance *memory, void *node)
 {
     uint32 linear_mem_size;
 #if WASM_ENABLE_SHARED_MEMORY != 0
--- a/core/iwasm/common/wasm_memory.c
+++ b/core/iwasm/common/wasm_memory.c
@@ -8,6 +8,10 @@
 #include "bh_platform.h"
 #include "mem_alloc.h"
 
+#if WASM_ENABLE_SHARED_MEMORY != 0
+#include "../common/wasm_shared_memory.h"
+#endif
+
 typedef enum Memory_Mode {
     MEMORY_MODE_UNKNOWN = 0,
     MEMORY_MODE_POOL,
@@ -506,7 +510,7 @@ wasm_get_default_memory(WASMModuleInstance *module_inst)
 
 #ifndef OS_ENABLE_HW_BOUND_CHECK
 bool
-wasm_enlarge_memory(WASMModuleInstance *module, uint32 inc_page_count)
+wasm_enlarge_memory_internal(WASMModuleInstance *module, uint32 inc_page_count)
 {
     WASMMemoryInstance *memory = wasm_get_default_memory(module);
     uint8 *memory_data_old, *memory_data_new, *heap_data_old;
@@ -624,7 +628,7 @@ wasm_enlarge_memory(WASMModuleInstance *module, uint32 inc_page_count)
 }
 #else
 bool
-wasm_enlarge_memory(WASMModuleInstance *module, uint32 inc_page_count)
+wasm_enlarge_memory_internal(WASMModuleInstance *module, uint32 inc_page_count)
 {
     WASMMemoryInstance *memory = wasm_get_default_memory(module);
     uint32 num_bytes_per_page, total_size_old;
@@ -697,3 +701,23 @@ wasm_enlarge_memory(WASMModuleInstance *module, uint32 inc_page_count)
     return true;
 }
 #endif /* end of OS_ENABLE_HW_BOUND_CHECK */
+
+bool
+wasm_enlarge_memory(WASMModuleInstance *module, uint32 inc_page_count)
+{
+    bool ret = false;
+
+#if WASM_ENABLE_SHARED_MEMORY != 0
+    WASMSharedMemNode *node =
+        wasm_module_get_shared_memory((WASMModuleCommon *)module->module);
+    if (node)
+        os_mutex_lock(&node->shared_mem_lock);
+#endif
+    ret = wasm_enlarge_memory_internal(module, inc_page_count);
+#if WASM_ENABLE_SHARED_MEMORY != 0
+    if (node)
+        os_mutex_unlock(&node->shared_mem_lock);
+#endif
+
+    return ret;
+}
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -2387,7 +2387,7 @@ wasm_get_exception(WASMModuleInstance *module_inst)
 bool
 wasm_copy_exception(WASMModuleInstance *module_inst, char *exception_buf)
 {
-    int has_exception = false;
+    bool has_exception = false;
 
 #if WASM_ENABLE_SHARED_MEMORY != 0
     WASMSharedMemNode *node =
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -2397,7 +2397,8 @@ wasm_copy_exception(WASMModuleInstance *module_inst, char *exception_buf)
 #endif
     if (module_inst->cur_exception[0] != '\0') {
         /* NULL is passed if the caller is not interested in getting the
-         * exception name, but only in knowing if an exception has been raised
+         * exception content, but only in knowing if an exception has been
+         * raised
          */
         if (exception_buf != NULL)
             bh_memcpy_s(exception_buf, sizeof(module_inst->cur_exception),
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -1870,14 +1870,15 @@ static bool
 clear_wasi_proc_exit_exception(WASMModuleInstanceCommon *module_inst_comm)
 {
 #if WASM_ENABLE_LIBC_WASI != 0
-    const char *exception;
+    bool has_exception;
+    char exception[EXCEPTION_BUF_LEN];
     WASMModuleInstance *module_inst = (WASMModuleInstance *)module_inst_comm;
 
     bh_assert(module_inst_comm->module_type == Wasm_Module_Bytecode
               || module_inst_comm->module_type == Wasm_Module_AoT);
 
-    exception = wasm_get_exception(module_inst);
-    if (exception && !strcmp(exception, "Exception: wasi proc exit")) {
+    has_exception = wasm_copy_exception(module_inst, exception);
+    if (has_exception && !strcmp(exception, "Exception: wasi proc exit")) {
         /* The "wasi proc exit" exception is thrown by native lib to
            let wasm app exit, which is a normal behavior, we clear
            the exception here. And just clear the exception of current
@@ -2374,11 +2375,19 @@ wasm_set_exception_with_id(WASMModuleInstance *module_inst, uint32 id)
         wasm_set_exception(module_inst, "unknown exception");
 }
 
-// TODO(eloparco): Change signature to receive exception buffer
 const char *
 wasm_get_exception(WASMModuleInstance *module_inst)
 {
-    char *exception = NULL;
+    if (module_inst->cur_exception[0] == '\0')
+        return NULL;
+    else
+        return module_inst->cur_exception;
+}
+
+bool
+wasm_copy_exception(WASMModuleInstance *module_inst, char *exception_buf)
+{
+    int has_exception = false;
 
 #if WASM_ENABLE_SHARED_MEMORY != 0
     WASMSharedMemNode *node =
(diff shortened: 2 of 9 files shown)


core/iwasm/common/wasm_runtime_common.c
```diff
@@ -2364,13 +2374,31 @@ wasm_set_exception_with_id(WASMModuleInstance *module_inst, uint32 id)
         wasm_set_exception(module_inst, "unknown exception");
 }
 
+// TODO(eloparco): Change signature to receive exception buffer
 const char *
```
eloparco, line 2378: The thread sanitizer complains since this function is used inside the signal handler and it's not safe because it uses `malloc`.

It should be replace by something like
```c
bool
wasm_get_exception(WASMModuleInstance *module_inst, char *exception)
{
    int ret = true;

    WASMSharedMemNode *node =
        wasm_module_get_shared_memory((WASMModuleCommon
        *)module_inst->module);
    if (node)
#endif
    if (module_inst->cur_exception[0] != '\0') {
        if (exception != NULL)
            bh_memcpy_s(exception, sizeof(module_inst->cur_exception),
                        module_inst->cur_exception,
                        sizeof(module_inst->cur_exception));
    }
    else {
        ret = false;
    }
    if (node)
#endif

    return ret;
}
```

but that would require to change multiple APIs, e.g.
```c
WASM_RUNTIME_API_EXTERN bool
wasm_runtime_get_exception(wasm_module_inst_t module_inst, char *exception);

bool
wasm_get_exception(WASMModuleInstance *module, char *exception);

bool
aot_get_exception(AOTModuleInstance *module_inst, char *exception);
```
so I'm waiting to get some feedback before proceeding.

wenyongh: Not sure which signal handler uses this function? It is used widely used and had better not change it.
If needed, we can add another API for thread manager, e.g.:
```C
void wasm_copy_exception(WASMModuleInstance *module_inst,
                         char *exception_buf, uint32 exception_buf_size);
```

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -52,11 +53,17 @@ wasm_shared_memory_init()
 {
     if (os_mutex_init(&shared_memory_list_lock) != 0)
         return false;
-    /* wait map not exists, create new map */
+
+    if (os_mutex_init(&wait_map_lock) != 0) {
```
wenyongh, line 57: The hash map already has its lock if it is created with `bh_hash_map_create(.., true, ..), and its related operations will be locked, could you check whether we should add the lock again?

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -113,14 +121,14 @@ notify_stale_threads_on_exception(WASMModuleInstanceCommon *module_inst)
     uint32 i = 0, total_elem_count = 0;
     uint64 total_elem_count_size = 0;
 
-    os_mutex_lock(&shared_memory_list_lock);
+    os_mutex_lock(&wait_map_lock);
```
wenyongh, line 124: bh_hash_map_traverse is already locked:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/shared/utils/bh_hashmap.c#L317

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -292,13 +302,13 @@ acquire_wait_info(void *address, bool create)
     AtomicWaitInfo *wait_info = NULL;
     bh_list_status ret;
 
-    os_mutex_lock(&shared_memory_list_lock);
+    os_mutex_lock(&wait_map_lock);
```
wenyongh, line 305: Same as above

core/iwasm/common/wasm_runtime_common.c
eloparco: Yes, I came to the same conclusion in the end, better to add a new API.

eloparco: > Not sure which signal handler uses this function?

here
https://github.com/bytecodealliance/wasm-micro-runtime/blob/9f0c4b63ac21c975aee7aae30236e306c8e49d13/core/iwasm/common/wasm_runtime_common.c#L197

core/iwasm/common/wasm_shared_memory.c
eloparco: Thanks, I didn't know about that. I suppose I still need it since I'm also using it to make the removal from list and hashmap atomic.
https://github.com/eloparco/wasm-micro-runtime/blob/0c8da385a8dc696e46b4e93457da42da73cdc965/core/iwasm/common/wasm_shared_memory.c#L477

list removal:
https://github.com/eloparco/wasm-micro-runtime/blob/0c8da385a8dc696e46b4e93457da42da73cdc965/core/iwasm/common/wasm_shared_memory.c#L483
map removal:
https://github.com/eloparco/wasm-micro-runtime/blob/0c8da385a8dc696e46b4e93457da42da73cdc965/core/iwasm/common/wasm_shared_memory.c#L490
https://github.com/eloparco/wasm-micro-runtime/blob/0c8da385a8dc696e46b4e93457da42da73cdc965/core/iwasm/common/wasm_shared_memory.c#L382

So I think I can pass `false` to the constructor to avoid using the built-in lock and use the one I created instead.

core/iwasm/common/wasm_runtime_common.c
eloparco:
```c
void wasm_copy_exception(WASMModuleInstance *module_inst,
                         char *exception_buf, uint32 exception_buf_size);
```

I only changed it to return `bool` so that `false` is returned if there is no exception. And `NULL` can be passed as `exception_buf` if the caller is only interested in the returned value (i.e. if an exception has been raised or not). I wanted to avoid creating a separate API like `wasm_has_exception()` to do the same.

core/iwasm/common/wasm_shared_memory.c
eloparco: The wait_map lock was not necessary in the end, I pushed a change to get rid of it

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -440,48 +427,46 @@ wasm_runtime_atomic_wait(WASMModuleInstanceCommon *module, void *address,
 
         if (0 != os_mutex_init(&wait_node->wait_lock)) {
             wasm_runtime_free(wait_node);
-            os_mutex_unlock(&wait_info->wait_list_lock);
```
wenyongh, line 443: Should also remove L436 (original file)

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -440,48 +427,46 @@ wasm_runtime_atomic_wait(WASMModuleInstanceCommon *module, void *address,
...
 
     os_cond_reltimedwait(&wait_node->wait_cond, &wait_node->wait_lock,
                          timeout < 0 ? BHT_WAIT_FOREVER
                                      : (uint64)timeout / 1000);
 
+    is_timeout = wait_node->status == S_WAITING ? true : false;
     os_mutex_unlock(&wait_node->wait_lock);
 
-    /* Check the wait node status */
+    /* Unlocked in release_wait_info() */
```
wenyongh, line 457: Code may be not easy to understand if releasing the lock in release_wait_info(), and the lock/unlock order is not clear (lock1, lock2, unlock1, unlock2, but not lock1, lock2, unlock2, unlock1), how about moving the `os_mutex_unlock(&wait_info->wait_list_lock)` to be after L472:
```C

    check_ret = is_wait_node_exists(wait_info->wait_list, wait_node);
    bh_assert(check_ret);
    ...

    release_wait_info(wait_map, wait_info, address);

    os_mutex_unlock(&wait_info->wait_list_lock)
```

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -2102,6 +2111,10 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                 read_leb_uint32(frame_ip, frame_ip_end, reserved);
                 delta = (uint32)POP_I32();
 
+#if WASM_ENABLE_SHARED_MEMORY != 0
+                if (node)
+                    os_mutex_lock(&node->shared_mem_lock);
```
wenyongh, line 2116: Should we also add the lock for aot_enlarge_memory:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/aot/aot_runtime.c#L1833

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -1128,13 +1128,27 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+    if (node)
+        os_mutex_lock(&node->shared_mem_lock);
+#endif
     uint32 num_bytes_per_page = memory ? memory->num_bytes_per_page : 0;
     uint32 linear_mem_size =
         memory ? num_bytes_per_page * memory->cur_page_count : 0;
+#if WASM_ENABLE_SHARED_MEMORY != 0
+    if (node)
+        os_mutex_unlock(&node->shared_mem_lock);
 #endif
```
wenyongh, line 1147: This makes the coding style not so good, how about add a function to get the num_bytes_per_page and linear_mem_size:
```C
static uint32
get_num_bytes_per_page(WASMMemoryInstance *memory, void *node)
{
    if (node)
        os_mutex_lock(&((WASMSharedMemNode *)node)->shared_mem_lock);
#endif
   ...
}
static uint32
get_linear_memory_size(WASMMemoryInstance *memory, void *node)
{
    ...
}
#endif

And here:
    WASMSharedMemNode *node =
        wasm_module_get_shared_memory((WASMModuleCommon *)module->module);
#else
    void *node =NULL;
#endif
    uint32 num_bytes_per_page = get__num_bytes_per_page(memory, node);
    uint32 linear_mem_size = get_linear_mem_size(memory, node);
```

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -1155,13 +1155,27 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
 {
     WASMMemoryInstance *memory = wasm_get_default_memory(module);
+
 #if !defined(OS_ENABLE_HW_BOUND_CHECK)              \
     || WASM_CPU_SUPPORTS_UNALIGNED_ADDR_ACCESS == 0 \
     || WASM_ENABLE_BULK_MEMORY != 0
+#if WASM_ENABLE_SHARED_MEMORY != 0
+    WASMSharedMemNode *node =
+        wasm_module_get_shared_memory((WASMModuleCommon *)module->module);
+    if (node)
```
wenyongh, line 1165: Same as above

core/iwasm/common/wasm_shared_memory.c
eloparco: The problem there is that `release_wait_info()` may call `destroy_wait_info()`
https://github.com/eloparco/wasm-micro-runtime/blob/65ce28c8085eb45d8a3032062658b35175ee2100/core/iwasm/common/wasm_shared_memory.c#L364
 so when you call `os_mutex_unlock(&wait_info->wait_list_lock)` the wait info is invalid.

I refactored the code to address that.

core/iwasm/interpreter/wasm_interp_classic.c
eloparco: The disadvantage is that we have to lock/unlock twice

eloparco:
While the changes in this PR seem to fix most of the race conditions shown by the thread sanitizer when running the tests in #1963, the test here is still failing  https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/4308819396/jobs/7515771517 (and it's failing quite consistently also on my local machine). I though it was related to race conditions, but apparently that's not the case.

core/iwasm/interpreter/wasm_interp_classic.c
wenyongh: Yes, but usually we declare local variables at the beginning of `{ .. }`. Or you can declare the variables and get the node here, and set them after the variable declarations?

core/iwasm/common/wasm_shared_memory.c
wenyongh: OK, that looks better.

core/iwasm/interpreter/wasm_interp_classic.c
eloparco: Or maybe I can move the lock in `wasm_enlarge_memory()` directly

core/iwasm/interpreter/wasm_interp_classic.c
wenyongh: > The disadvantage is that we have to lock/unlock twice

The num_bytes_per_page isn't changed in enlarge memory, no need to add lock for it. Only need to add lock for linear mem size.

eloparco: But the value still gets written (even if not changed)
https://github.com/bytecodealliance/wasm-micro-runtime/blob/9f0c4b63ac21c975aee7aae30236e306c8e49d13/core/iwasm/common/wasm_memory.c#L552 so that should still cause undefined behavior in case of a race condition. I'll apply the first suggestion.

core/iwasm/common/wasm_memory.c
```diff
@@ -697,3 +701,59 @@ wasm_enlarge_memory(WASMModuleInstance *module, uint32 inc_page_count)
...
+#endif
+
+    return ret;
+}
+
+#if !defined(OS_ENABLE_HW_BOUND_CHECK)              \
+    || WASM_CPU_SUPPORTS_UNALIGNED_ADDR_ACCESS == 0 \
+    || WASM_ENABLE_BULK_MEMORY != 0
+uint32
+get_num_bytes_per_page(WASMMemoryInstance *memory, void *node)
```
wenyongh, line 729: If added in wasm_interp_classic.c as a static function, it is good to name like this. For here, could you rename it as `wasm_get_num_bytes_per_page`, adding prefix `wasm_` like other functions?

core/iwasm/common/wasm_memory.c
```diff
@@ -697,3 +701,59 @@ wasm_enlarge_memory(WASMModuleInstance *module, uint32 inc_page_count)
...
+    num_bytes_per_page = memory->num_bytes_per_page;
+#if WASM_ENABLE_SHARED_MEMORY != 0
+    if (node)
+        os_mutex_unlock(&((WASMSharedMemNode *)node)->shared_mem_lock);
+#endif
+    return num_bytes_per_page;
+}
+
+uint32
+get_linear_memory_size(WASMMemoryInstance *memory, void *node)
```
wenyongh, line 745: Same as above, `wasm_get_linear_memory_size`

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -130,13 +139,13 @@ notify_stale_threads_on_exception(WASMModuleInstanceCommon *module_inst)
...
             "failed to allocate memory for list of atomic wait addresses");
-        os_mutex_unlock(&shared_memory_list_lock);
+        os_mutex_unlock(&wait_map_lock);
         return;
     }
 
     /* set values in list of addresses */
     bh_hash_map_traverse(wait_map, create_list_of_waiter_addresses, &args);
-    os_mutex_unlock(&shared_memory_list_lock);
+    os_mutex_unlock(&wait_map_lock);
```
wenyongh, line 148: Is it better to also lock the notify process? Or, move this line to be after L153?

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -1126,14 +1127,21 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
     WASMMemoryInstance *memory = wasm_get_default_memory(module);
     uint8 *global_data = module->global_data;
 #if !defined(OS_ENABLE_HW_BOUND_CHECK)              \
     || WASM_CPU_SUPPORTS_UNALIGNED_ADDR_ACCESS == 0 \
     || WASM_ENABLE_BULK_MEMORY != 0
-    uint32 num_bytes_per_page = memory ? memory->num_bytes_per_page : 0;
-    uint32 linear_mem_size =
-        memory ? num_bytes_per_page * memory->cur_page_count : 0;
+    uint32 num_bytes_per_page =
+        memory ? get_num_bytes_per_page(memory, node) : 0;
```
wenyongh, line 1143: Had better `wasm_get_num_bytes_per_page`

core/iwasm/interpreter/wasm_runtime.h
```diff
@@ -444,6 +446,14 @@ wasm_set_exception_with_id(WASMModuleInstance *module_inst, uint32 id);
 const char *
 wasm_get_exception(WASMModuleInstance *module);
 
+/**
+ * @brief Copy exception in buffer passed as parameter. Thread-safe version of
```
wenyongh, line 450: Had better mention that the size of exception_buf must be no smaller than EXCEPTION_BUF_LEN.

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -1142,13 +1146,23 @@ set_exception_visitor(void *node, void *user_data)
         WASMModuleInstance *curr_wasm_inst =
             (WASMModuleInstance *)get_module_inst(curr_exec_env);
 
-        /* Only spread non "wasi proc exit" exception */
+/* Only spread non "wasi proc exit" exception */
```
wenyongh, line 1149: Indent, not sure why CI didn't report error.

wenyongh (commented):
LGTM, except several minor change suggestion.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -2373,6 +2384,35 @@ wasm_get_exception(WASMModuleInstance *module_inst)
...
+
+#if WASM_ENABLE_SHARED_MEMORY != 0
+    WASMSharedMemNode *node =
+        wasm_module_get_shared_memory((WASMModuleCommon *)module_inst->module);
+    if (node)
+        os_mutex_lock(&node->shared_mem_lock);
+#endif
+    if (module_inst->cur_exception[0] != '\0') {
+        /* NULL is passed if the caller is not interested in getting the
+         * exception name, but only in knowing if an exception has been raised
```
xujuntwt95329, line 2400:
```suggestion
         * exception content, but only in knowing if an exception has been raised
```

core/iwasm/common/wasm_shared_memory.c
eloparco: It goes into deadlock since then `wasm_runtime_atomic_notify()` calls `acquire_wait_info()` that in turn tries to get `wait_map_lock`

eloparco:
I think we can merge it. I still see some edge cases remaining, but this PR should fix most data races.

PR #2004, opened 2023-03-06
Fix wasm_cluster_create_thread issue

wenyongh:
In wasm_cluster_create_thread, the new_exec_env is added into the cluster's
exec_env list before the thread is created, so other threads can access the
fields of new_exec_env once the cluster->lock is unlocked, while the
new_exec_env's handle is set later inside the thread routine. This may result
in the new_exec_env's handle may be invalidly accessed by other threads.


loganek (approved):
LGTM

PR #2006, opened 2023-03-06
Enhance/Fix sample socket-api and workload

wenyongh:
Add python script to test socket-api cases
Fix issues in send_recv wasm app
Fix issues in build meshoptimizer
Enhance build script of sample workload


hritikgupta:
@wenyongh it seems this PR covers the cases I was trying to add in #1900, should I close mine?

wenyongh:
> @wenyongh it seems this PR covers the cases I was trying to add in #1900, should I close mine?

Suggest to keep #1900, since this PR (#2006) just uses the original cases under samples/socket-api and adds python script to auto run the cases, it didn't introduce new cases. In the future, we can add more test cases for socket-api based on #1900, I think we had better merge it after #1963 is merged, since the changes of CI may be duplicated.

samples/socket-api/sample_test_run.py
```diff
@@ -0,0 +1,143 @@
...
+
+    test_aot = False
+    suffix = ".wasm"
+    if not args.aot:
+        print("Test with interpreter mode")
+    else:
+        print("Test with AOT mode")
+        test_aot = True
+        suffix = ".aot"
+        wasm_files = [ "addr_resolve", "multicast_server", "socket_opts",
```
loganek, line 47: I wonder if we can get a list of files automatically (e.g. using `glob('*.wasm')`)? If we need to exclude any particular file for any reason, I presume we could add exclusion list in the future.

samples/socket-api/sample_test_run.py
```diff
@@ -0,0 +1,143 @@
...
+                       "tcp_server", "timeout_server", "udp_server",
+                       "multicast_client", "send_recv", "tcp_client",
+                       "timeout_client",  "udp_client" ]
+        wasm_apps_dir = args.working_directory
+        print("Compile wasm app into aot files")
+        for wasm_file in wasm_files:
+            CMD = []
+            CMD.append(WAMRC_CMD)
```
loganek, line 55: [minor] not sure what's the benefit of using `append` here if the list can be constructed in place, i.e.
```python
CMD = [
	WAMRC_CMD,
	'-o',
	# ...
```

Also, I think upper case variable names are used for global variables, so consider renaming it to `cmd`

samples/socket-api/sample_test_run.py
```diff
@@ -0,0 +1,143 @@
+#!/usr/bin/env python3
+#
+# Copyright (C) 2019 Intel Corporation.  All rights reserved.
```
loganek, line 3:
```suggestion
# Copyright (C) 2023 Intel Corporation.  All rights reserved.
```

loganek:
> > @wenyongh it seems this PR covers the cases I was trying to add in #1900, should I close mine?
> 
> Suggest to keep #1900, since this PR (#2006) just uses the original cases under samples/socket-api and adds python script to auto run the cases, it didn't introduce new cases. In the future, we can add more test cases for socket-api based on #1900, I think we had better merge it after #1963 is merged, since the changes of CI may be duplicated.

Could you actually give some context for this script? Is it supposed to be run as part of test automation? If so, I think we already have that in: https://github.com/bytecodealliance/wasm-micro-runtime/pull/1900
Or is it to be able to run examples automatically? If that's the case, do you think instructions that we have in README are not enough?

samples/socket-api/sample_test_run.py
wenyongh: OK, I will have a try.

wenyongh:
> > > @wenyongh it seems this PR covers the cases I was trying to add in #1900, should I close mine?
> > 
> > 
> > Suggest to keep #1900, since this PR (#2006) just uses the original cases under samples/socket-api and adds python script to auto run the cases, it didn't introduce new cases. In the future, we can add more test cases for socket-api based on #1900, I think we had better merge it after #1963 is merged, since the changes of CI may be duplicated.
> 
> Could you actually give some context for this script? Is it supposed to be run as part of test automation? If so, I think we already have that in: #1900 Or is it to be able to run examples automatically? If that's the case, do you think instructions that we have in README are not enough?

Just want to auto run these cases with interpreter mode and AOT mode respectively, as you know, it is not so convenient to run the command one by one. Now it can run all the instructions in the README. Do you want to add more?

samples/socket-api/sample_test_run.py
wenyongh: Done, code is better now.

PR #2013, opened 2023-03-07
Fix multi-threading issues

wenyongh:
- Implement atomic.fence to ensure a proper memory synchronization order
- Destroy exec_env_singleton first in wasm/aot deinstantiation
- Fix detach thread in thread_manager_start_routine
- Fix duplicated lock cluster->lock in wasm_cluster_cancel_thread
- Add lib-pthread and lib-wasi-threads compilation to Windows CI


core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -595,14 +595,24 @@ thread_manager_start_routine(void *arg)
 
     /* Routine exit */
 
-    /* Detach the native thread here to ensure the resources are freed */
-    wasm_cluster_detach_thread(exec_env);
 #if WASM_ENABLE_DEBUG_INTERP != 0
     wasm_cluster_thread_exited(exec_env);
 #endif
 
+    os_mutex_lock(&cluster_list_lock);
```
xujuntwt95329, line 602: Is this lock required by `wasm_cluster_del_exec_env`?
Had better update the comment for `wasm_cluster_del_exec_env`:

``` C
/* The caller should lock cluster->lock and 
    cluster_list_lock for thread safety */
```

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -981,8 +1006,6 @@ wasm_cluster_cancel_thread(WASMExecEnv *exec_env)
         return 0;
     }
 
-    os_mutex_lock(&exec_env->cluster->lock);
```
xujuntwt95329, line 984: Had better add comment for this function to ask the caller to ensure thread safety

core/iwasm/libraries/thread-mgr/thread_manager.c
wenyongh: No, this lock here is to lock
```C
if (exec_env->wait_count == 0 && !exec_env->thread_is_detached) {
    ...
}
```

wasm_cluster_del_exec_env only needs to lock cluster->lock

core/iwasm/libraries/thread-mgr/thread_manager.c
wenyongh: Removing the code here since clusters_have_exec_env will lock again, which leads to hang.

eloparco (approved):

xujuntwt95329:
LGTM

PR #2016, opened 2023-03-08
Fix wait_info data race for deletion + fix atomic_wait logic

g0djan:
Fix a data race for test main_proc_exit_wait.c from https://github.com/bytecodealliance/wasm-micro-runtime/pull/1963

```
   #2 wasm_runtime_atomic_wait /workspaces/wasm-micro-runtime/core/iwasm/common/wasm_shared_memory.c:461:9 (iwasm+0xe7a00) (BuildId: 1368de2fad9dd8fefb4d8ce54eb6927bc45154b7)
    #3 wasm_interp_call_func_bytecode /workspaces/wasm-micro-runtime/core/iwasm/interpreter/wasm_interp_classic.c:3429:31 (iwasm+0x1516cb) (BuildId: 1368de2fad9dd8fefb4d8ce54eb6927bc45154b7)
    #4 wasm_interp_call_wasm /workspaces/wasm-micro-runtime/core/iwasm/interpreter/wasm_interp_classic.c:4226:13 (iwasm+0x13bb03) (BuildId: 1368de2fad9dd8fefb4d8ce54eb6927bc45154b7)
    #5 call_wasm_with_hw_bound_check /workspaces/wasm-micro-runtime/core/iwasm/interpreter/wasm_runtime.c:2276:9 (iwasm+0xed92f) (BuildId: 1368de2fad9dd8fefb4d8ce54eb6927bc45154b7)
    #6 wasm_call_function /workspaces/wasm-micro-runtime/core/iwasm/interpreter/wasm_runtime.c:2340:5 (iwasm+0xed6fb) (BuildId: 1368de2fad9dd8fefb4d8ce54eb6927bc45154b7)
    #7 wasm_runtime_call_wasm /workspaces/wasm-micro-runtime/core/iwasm/common/wasm_runtime_common.c:1927:15 (iwasm+0xddd00) (BuildId: 1368de2fad9dd8fefb4d8ce54eb6927bc45154b7)
    #8 thread_start /workspaces/wasm-micro-runtime/core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c:60:10 (iwasm+0x1745de) (BuildId: 1368de2fad9dd8fefb4d8ce54eb6927bc45154b7)
    #9 thread_manager_start_routine /workspaces/wasm-micro-runtime/core/iwasm/libraries/thread-mgr/thread_manager.c:587:11 (iwasm+0x1096e5) (BuildId: 1368de2fad9dd8fefb4d8ce54eb6927bc45154b7)
    #10 os_thread_wrapper /workspaces/wasm-micro-runtime/core/shared/platform/common/posix/posix_thread.c:43:5 (iwasm+0x10b438) (BuildId: 1368de2fad9dd8fefb4d8ce54eb6927bc45154b7)

  Previous write of size 8 at 0x7b1000006000 by thread T3 (mutexes: write M0):
    #0 free <null> (iwasm+0x53d24) (BuildId: 1368de2fad9dd8fefb4d8ce54eb6927bc45154b7)
    #1 wasm_runtime_free_internal /workspaces/wasm-micro-runtime/core/iwasm/common/wasm_memory.c:218:9 (iwasm+0xdaca8) (BuildId: 1368de2fad9dd8fefb4d8ce54eb6927bc45154b7)
    #2 wasm_runtime_free /workspaces/wasm-micro-runtime/core/iwasm/common/wasm_memory.c:250:5 (iwasm+0xdaba5) (BuildId: 1368de2fad9dd8fefb4d8ce54eb6927bc45154b7)
    #3 destroy_wait_info /workspaces/wasm-micro-runtime/core/iwasm/common/wasm_shared_memory.c:375:9 (iwasm+0xe6ae3) (BuildId: 1368de2fad9dd8fefb4d8ce54eb6927bc45154b7)
    #4 wasm_runtime_atomic_wait /workspaces/wasm-micro-runtime/core/iwasm/common/wasm_shared_memory.c:492:9 (iwasm+0xe7bd5) (BuildId: 1368de2fad9dd8fefb4d8ce54eb6927bc45154b7)
    #5 wasm_interp_call_func_bytecode /workspaces/wasm-micro-runtime/core/iwasm/interpreter/wasm_interp_classic.c:3429:31 (iwasm+0x1516cb) (BuildId: 1368de2fad9dd8fefb4d8ce54eb6927bc45154b7)
    #6 wasm_interp_call_wasm /workspaces/wasm-micro-runtime/core/iwasm/interpreter/wasm_interp_classic.c:4226:13 (iwasm+0x13bb03) (BuildId: 1368de2fad9dd8fefb4d8ce54eb6927bc45154b7)
    #7 call_wasm_with_hw_bound_check /workspaces/wasm-micro-runtime/core/iwasm/interpreter/wasm_runtime.c:2276:9 (iwasm+0xed92f) (BuildId: 1368de2fad9dd8fefb4d8ce54eb6927bc45154b7)
    #8 wasm_call_function /workspaces/wasm-micro-runtime/core/iwasm/interpreter/wasm_runtime.c:2340:5 (iwasm+0xed6fb) (BuildId: 1368de2fad9dd8fefb4d8ce54eb6927bc45154b7)
    #9 wasm_runtime_call_wasm /workspaces/wasm-micro-runtime/core/iwasm/common/wasm_runtime_common.c:1927:15 (iwasm+0xddd00) (BuildId: 1368de2fad9dd8fefb4d8ce54eb6927bc45154b7)
    #10 thread_start /workspaces/wasm-micro-runtime/core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c:60:10 (iwasm+0x1745de) (BuildId: 1368de2fad9dd8fefb4d8ce54eb6927bc45154b7)
    #11 thread_manager_start_routine /workspaces/wasm-micro-runtime/core/iwasm/libraries/thread-mgr/thread_manager.c:587:11 (iwasm+0x1096e5) (BuildId: 1368de2fad9dd8fefb4d8ce54eb6927bc45154b7)
    #12 os_thread_wrapper /workspaces/wasm-micro-runtime/core/shared/platform/common/posix/posix_thread.c:43:5 (iwasm+0x10b438) (BuildId: 1368de2fad9dd8fefb4d8ce54eb6927bc45154b7)
```
And fix of `atomic_wait` logic that was wrong before:
- a thread 1 started executing wasm instruction `wasm_atomic_wait` but hasn't reached waiting on condition variable
- a main thread calls `proc_exit` and notifies all the threads that reached waiting on condition variable
-> thread 1 hangs on waiting on condition variable after that

Now it's atomically checked whether `proc_exit` was already called. 

Checked that this code fix the both problems described above which are reproducable on the main branch


eloparco:
It may have been fixed already in #2013, I wasn't getting any race condition in `wasm_shared_memory.c` when I tried it

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -380,11 +383,20 @@ static bool
 map_remove_wait_info(HashMap *wait_map_, AtomicWaitInfo *wait_info,
                      void *address)
 {
+    os_mutex_lock(&wait_map_lock);
+    --wait_info->count_acquisition;
+
     if (wait_info->wait_list->len > 0) {
```
g0djan, line 389: Not sure whether length ever can be bigger than number of acquirements though

g0djan:
@eloparco it couldn't be fixed there, this file wasn't modified in that PR

wenyongh:
> @eloparco it couldn't be fixed there, this file wasn't modified in that PR

@g0djan Thanks for uploading the PR!

Another issue I found several times is that (Also the issue that the CI ran failed for this PR):
```bash
Run ./test_wamr.sh $WASI_TEST_OPTIONS -t classic-interp
...
Test thread_spawn-simple failed
  [exit_code] 0 == 1
STDOUT:
Exception: unreachable

STDERR:
Assertion failed: __builtin_wasm_memory_atomic_wait32(&data[i].th_done, 0, SECOND) != 2 (testsuite/thread_spawn-simple.c: main: 67)
```
Refer to https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/4363883798/jobs/7630640507

@g0djan @eloparco Any idea about it?

g0djan:
No idea, but will look into it

eloparco:
> Another issue I found several times is that (Also the issue that the CI ran failed for this PR)

Did that happen after your recent PR #2013? Otherwise it might be fixed already.

eloparco:
@g0djan Did you rebase on top of main? I'm a bit skeptical about merging it if there is no proven data race being fixed and with the latest changes I didn't see the race conditions you reported.

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -380,11 +384,23 @@ static bool
 map_remove_wait_info(HashMap *wait_map_, AtomicWaitInfo *wait_info,
                      void *address)
 {
+    os_mutex_lock(&wait_map_lock);
+    --wait_info->count_acquisition;
```
eloparco, line 388: what's the difference between using `count_acquisition` and `wait_info->wait_list->len` directly?

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -307,6 +308,8 @@ acquire_wait_info(void *address, bool create)
 
     if (address)
         wait_info = (AtomicWaitInfo *)bh_hash_map_find(wait_map, address);
+    if (wait_info)
+        ++wait_info->count_acquisition;
```
eloparco, line 312: Here we're accessing `count_acquisition` without a lock on `wait_info`

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -380,11 +384,23 @@ static bool
 map_remove_wait_info(HashMap *wait_map_, AtomicWaitInfo *wait_info,
                      void *address)
 {
+    os_mutex_lock(&wait_map_lock);
```
eloparco, line 387: `wait_map_` is initialized with true as a second parameter https://github.com/bytecodealliance/wasm-micro-runtime/blob/289fc5efbf5cb2f75f1993ad0131dada78400d3b/core/iwasm/common/wasm_shared_memory.c#L63
which means it already uses a lock internally. Therefore, the lock here should not be necessary.

Currently, `wait_map_lock` is used to make multiple operations on the map atomic (see comments)
https://github.com/bytecodealliance/wasm-micro-runtime/blob/289fc5efbf5cb2f75f1993ad0131dada78400d3b/core/iwasm/common/wasm_shared_memory.c#L125
https://github.com/bytecodealliance/wasm-micro-runtime/blob/289fc5efbf5cb2f75f1993ad0131dada78400d3b/core/iwasm/common/wasm_shared_memory.c#L306

core/iwasm/common/wasm_shared_memory.c
g0djan: Fair point, I'll try to reorder stuff to use the existing list then

core/iwasm/common/wasm_shared_memory.c
g0djan: Because `wait_info->wait_list` is empty when we acquire and it causes the problem, but I can try to reorder it so we can use `wait_info->wait_list->len directly`

core/iwasm/common/wasm_shared_memory.c
g0djan: Just remembered it's like that because count_acquisition memory accessed only under wait_map_lock. But I will try to change to using the existing list

g0djan:
@eloparco just reproduced the data race on top of the main in the loop of 100 iterations

g0djan:
Made a fix without introducing counter and with reorder of calls, but pushed it to a separate branch because I'm not sure whether it's better https://github.com/bytecodealliance/wasm-micro-runtime/compare/main...g0djan:wasm-micro-runtime:godjan/fix-wait-info-data-race2?expand=1


Current status of the data race:
- reproduces on main
- doesn't reproduce with [this fix](https://github.com/bytecodealliance/wasm-micro-runtime/pull/2016)
- doesn't reproduce with [the fix without introducing counter](https://github.com/bytecodealliance/wasm-micro-runtime/compare/main...g0djan:wasm-micro-runtime:godjan/fix-wait-info-data-race2?expand=1) as well

But on all 3 branches I can catch a deadlock if I try enough times

g0djan:
CHecked with the debugger that deadlock is the same for all 3 branches: main, g0djan:godjan/fix-wait-info-data-race and g0djan:godjan/fix-wait-info-data-race2 so I updated the PR to avoid using the counter

eloparco:
> Made a fix without introducing counter and with reorder of calls, but pushed it to a separate branch because I'm not sure whether it's better https://github.com/bytecodealliance/wasm-micro-runtime/compare/main...g0djan:wasm-micro-runtime:godjan/fix-wait-info-data-race2?expand=1

As discussed offline, I prefer the second approach since the counter here seems redundant and it's not in sync with the `wait_list` length.

g0djan:
The updated version is already in this PR, pushed right after we discussed it.

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -298,7 +298,7 @@ notify_wait_list(bh_list *wait_list, uint32 count)
 }
 
 static AtomicWaitInfo *
-acquire_wait_info(void *address, bool create)
+acquire_wait_info(void *address, bool create, AtomicWaitNode *wait_node)
```
wenyongh, line 304: The function is called with either `acquire_wait_info(address, true, wait_node)` or `acquire_wait_info(address, false, NULL)`, seems we can remove argument `bool create`, only keep `wait node`.

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -380,11 +387,17 @@ static bool
...
                      void *address)
 {
+    os_mutex_lock(&wait_map_lock);
+    os_mutex_lock(&wait_info->wait_list_lock);
     if (wait_info->wait_list->len > 0) {
+        os_mutex_unlock(&wait_info->wait_list_lock);
+        os_mutex_unlock(&wait_map_lock);
         return false;
     }
+    os_mutex_unlock(&wait_info->wait_list_lock);
```
wenyongh, line 400: Had better move this line to be after L399 `bh_hash_map_remove` to avoid other thread inserting `wait_node` into `wait_info->wait_list_lock` when it is to be removed?

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -435,40 +444,48 @@ wasm_runtime_atomic_wait(WASMModuleInstanceCommon *module, void *address,
...
-        wait_node->status = S_WAITING;
-        os_mutex_lock(&wait_info->wait_list_lock);
-        ret = bh_list_insert(wait_info->wait_list, wait_node);
-        os_mutex_unlock(&wait_info->wait_list_lock);
-        bh_assert(ret == BH_LIST_SUCCESS);
-        (void)ret;
+    wait_node->status = S_WAITING;
+
+    /* acquire the wait info, create new one if not exists */
+    wait_info = acquire_wait_info(address, true, wait_node);
```
wenyongh, line 468: acquire_wait_info(address, wait_node)

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -435,40 +444,48 @@ wasm_runtime_atomic_wait(WASMModuleInstanceCommon *module, void *address,
...
     }
 
     /* condition wait start */
     os_mutex_lock(&wait_node->wait_lock);
 
-    os_cond_reltimedwait(&wait_node->wait_cond, &wait_node->wait_lock,
-                         timeout < 0 ? BHT_WAIT_FOREVER
-                                     : (uint64)timeout / 1000);
+#if WASM_ENABLE_THREAD_MGR != 0
+    if (!wasm_cluster_is_thread_terminated(exec_env)) {
```
wenyongh, line 481: Suggest to also re-read the value in address again, which may have changed after we read it last time:
```C
    no_wait = (!wait64 && *(uint32 *)address != (uint32)expect)
              || (wait64 && *(uint64 *)address != expect);
    if (!no_wait
        && !wasm_cluster_is_thread_terminated(exec_env)
#endif
       ) {
         os_cond_reltimedwait(&wait_node->wait_cond, &wait_node->wait_lock,
                             timeout < 0 ? BHT_WAIT_FOREVER
                                         : (uint64)timeout / 1000);
      }
```
And in the end of this function:
`return no_wait ? 1 : is_timeout ? 2 : 0;`

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -380,17 +390,24 @@ static bool
...
     bh_hash_map_remove(wait_map_, address, NULL, NULL);
+    os_mutex_unlock(&wait_map_lock);
     return true;
 }
 
 uint32
 wasm_runtime_atomic_wait(WASMModuleInstanceCommon *module, void *address,
-                         uint64 expect, int64 timeout, bool wait64)
+                         uint64 expect, int64 timeout, bool wait64,
+                         WASMExecEnv *exec_env)
```
wenyongh, line 410: Had better not add the new argument since the API is also called from AOT code, we can use `exec_env = wasm_clusters_search_exec_env((WASMModuleInstanceCommon *)module_inst)`. If the performance is not so good, we may consider improving it future by saving the `exec_env` into module_inst.

core/iwasm/libraries/lib-wasi-threads/test/common.h
```diff
@@ -36,12 +36,11 @@ void
...
 {
     if (blocking_task_type == BLOCKING_TASK_BUSY_WAIT) {
-        for (int i = 0; i < TIMEOUT_SECONDS; i++)
+        for (;;)
             sleep(1);
     }
     else if (blocking_task_type == BLOCKING_TASK_ATOMIC_WAIT) {
-        __builtin_wasm_memory_atomic_wait32(
-            0, 0, TIMEOUT_SECONDS * 1000 * 1000 * 1000);
+        __builtin_wasm_memory_atomic_wait32(0, 0, -1);
```
eloparco, line 43: Good call, better to use infinite waits in this test otherwise failures may pass undetected in the CI.

core/iwasm/libraries/lib-wasi-threads/test/common.h
```diff
@@ -36,12 +36,11 @@ void
 run_long_task()
 {
     if (blocking_task_type == BLOCKING_TASK_BUSY_WAIT) {
-        for (int i = 0; i < TIMEOUT_SECONDS; i++)
+        for (;;)
             sleep(1);
```
eloparco, line 40: we can probably avoid the sleep and just have `for (;;) {};`

core/iwasm/common/wasm_shared_memory.c
g0djan: It can't happen because we insert only inside of `acquire_wait_info` and in the beginning of it we lock `wait_map_lock` which is as well locked here in L393.

A deletion won't happen as well because we already checked that the list is empty.

core/iwasm/libraries/lib-wasi-threads/test/common.h
```diff
@@ -36,12 +36,11 @@ void
...
     }
     else if (blocking_task_type == BLOCKING_TASK_ATOMIC_WAIT) {
-        __builtin_wasm_memory_atomic_wait32(
-            0, 0, TIMEOUT_SECONDS * 1000 * 1000 * 1000);
+        __builtin_wasm_memory_atomic_wait32(0, 0, -1);
     }
     else {
         sleep(TIMEOUT_SECONDS);
```
eloparco, line 46:
```suggestion
        sleep(UINT_MAX);
```
and then remove `#define TIMEOUT_SECONDS 10ll`

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -486,8 +503,8 @@ wasm_runtime_atomic_wait(WASMModuleInstanceCommon *module, void *address,
     wasm_runtime_free(wait_node);
 
     /* Release wait info if no wait nodes attached */
-    removed_from_map = map_remove_wait_info(wait_map, wait_info, address);
     os_mutex_unlock(&wait_info->wait_list_lock);
+    removed_from_map = map_remove_wait_info(wait_map, wait_info, address);
     if (removed_from_map)
         destroy_wait_info(wait_info);
```
eloparco, line 509: We can probably move this line inside `map_remove_wait_info` and rename `map_remove_wait_info` to `release_wait_info`. I made the reverse change in a previous PR, but with your modifications it makes more sense to restore it as it was before.

core/iwasm/common/wasm_shared_memory.c
g0djan, line 481: Concerned about `if` and not `while` here: [spurious wakeup](https://en.wikipedia.org/wiki/Spurious_wakeup)
@wenyongh @eloparco any thought on what `ret` should be considered as appropriate one to exit loop or what codes should be considered a `false` wakeup?

eloparco:
General comment: initially I wasn't able to reproduce the problem that this PR was trying to fix, but @g0djan showed me how to do it: instead of running all the tests multiple times (as I was doing), just run the single `main_proc_exit_wait` many times in a loop. The race conditions appear in the sanitizer and, after the changes in `common.h` (in this PR) we can also see a deadlock that before was hidden by the 10-second timeout.

g0djan:
Yes, this one("deadlock) we also sorted out in [this commit](https://github.com/bytecodealliance/wasm-micro-runtime/pull/2016/commits/bc9fa8b2889ff40923954bb3046e5fcdeb378f32) and now doing a check atomically(@eloparco thanks for the idea with condition_variable instead of wrapping it with another lock).

core/iwasm/common/wasm_shared_memory.c
eloparco: You mean the flag being set between the call to `wasm_cluster_is_thread_terminated` and `os_cond_reltimedwait`, right? That would cause the termination to be missed and the `os_cond_reltimedwait` to wait forever in case of `BHT_WAIT_FOREVER`

g0djan: No, the problem you talking about I solved with your idea of condition variable:
thread 1
- lock `wait_node->wait_lock`
- check whether it's been already terminated
- call wait on the condition variable that releases the lock
thread 2
- we lock `wait_node->wait_lock`
- we call notify on condition variable
- we unlock `wait_node->wait_lock`

To check I ran the tests and it never hangs any more, no tsan warnings as well(besides unrelated LOAD-STORE one)

g0djan: Here I'm concerned about `spurious wakeup`, on some systems thread can actually wake up from wait even there was no other thread in this program that notified it. Usually it's prevented by checking exit condition rather in a loop than just with an `if`.
https://en.wikipedia.org/wiki/Spurious_wakeup

g0djan: It's done in L440 already, since `address` and `expect` don't change in this function then there's no need to call it here.

core/iwasm/common/wasm_shared_memory.c
wenyongh: OK, got it.

core/iwasm/common/wasm_shared_memory.c
wenyongh: Though address don't change, the value of `*address` may be changed by other thread after L440?

core/iwasm/common/wasm_shared_memory.c
g0djan: Fixed

core/iwasm/common/wasm_shared_memory.c
g0djan: Fixed

core/iwasm/common/wasm_shared_memory.c
g0djan: Fixed

core/iwasm/libraries/lib-wasi-threads/test/common.h
g0djan: Fixed, will heat CPU but not for long

core/iwasm/libraries/lib-wasi-threads/test/common.h
g0djan: Fixed

core/iwasm/common/wasm_shared_memory.c
g0djan: Fixed

g0djan:
Removing timeouts showing up in the other tests already
```
Test main_trap_sleep passed
Error: The action has timed out.
```

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -21,6 +24,8 @@ typedef struct AtomicWaitInfo {
     korp_mutex wait_list_lock;
     bh_list wait_list_head;
     bh_list *wait_list;
+    // WARNING: insert to the list allowed only in acquire_wait_info
+    // otherwise there will be data race as described in PR #2016
```
wenyongh, line 28: Had better use `/* .. */` to add the comment, here it is C source code, we usually do like that (at least in several core folders) to avoid possible compilation warning.

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -308,7 +313,7 @@ acquire_wait_info(void *address, bool create)
     if (address)
         wait_info = (AtomicWaitInfo *)bh_hash_map_find(wait_map, address);
 
-    if (!create) {
+    if (!wait_node && !wait_info) {
```
wenyongh, line 316: Not sure why need to add `&& !wait_info`? That means we will create the wait_info even when this function is called from `wasm_runtime_atomic_notify`, not sure whether it is necessary?

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -435,40 +446,53 @@ wasm_runtime_atomic_wait(WASMModuleInstanceCommon *module, void *address,
...
+    if (!wait_info) {
+        os_mutex_destroy(&wait_node->wait_lock);
+        wasm_runtime_free(wait_node);
+        wasm_runtime_set_exception(module, "failed to acquire wait_info");
+        return -1;
     }
 
+#if WASM_ENABLE_THREAD_MGR != 0
+    exec_env =
+        wasm_clusters_search_exec_env((WASMModuleInstanceCommon *)module_inst);
```
wenyongh, line 481: Had better add `bh_assert(exec_env)`.

core/iwasm/common/wasm_shared_memory.c
g0djan: Okay I will add it, I believe I don't understand yet how it might be used and purpose of `*address` in general. Can you give an example why/how it can be changed?

wenyongh: Another thread may store data into the address with atomic.i32/i64.store opcodes after this thread read it in L442:

https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_interp_classic.c#L3550-L3623

https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_interp_classic.c#L3577

core/iwasm/common/wasm_shared_memory.c
g0djan: Fixed

core/iwasm/common/wasm_shared_memory.c
g0djan: Fixed

core/iwasm/common/wasm_shared_memory.c
g0djan: Fixed

core/iwasm/common/wasm_shared_memory.c
g0djan: Fixed

eloparco: Then I presume we want to add a `while` with a check on the `wait_node->status`

core/iwasm/common/wasm_shared_memory.c
eloparco: Let's maybe add a comment near `wait_map_lock` to specify that the lock is not used for single operations on the map since those are already protected by the internal lock

wenyongh:
> Removing timeouts showing up in the other tests already
> 
> ```
> Test main_trap_sleep passed
> Error: The action has timed out.
> ```

@g0djan @eloparco I fixed several issues based on this PR, I tested the cases many times and it seems they run good, I will merge this PR first and submit another PR, let's discuss more in that PR.

wenyongh:
> > Removing timeouts showing up in the other tests already
> > ```
> > Test main_trap_sleep passed
> > Error: The action has timed out.
> > ```
> 
> @g0djan @eloparco I fixed several issues based on this PR, I tested the cases many times and it seems they run good, I will merge this PR first and submit another PR, let's discuss more in that PR.

Refer to #2024.

PR #2021, opened 2023-03-10
Use wasi-sdk 20 pre-release for tests with threads

eloparco:
`wasi-sdk-20` pre-release can be used to avoid building `wasi-libc` to enable threads.
It's not possible to use `wasi-sdk-20` pre-release on Ubuntu 20.04 because of incompatibility with the glibc version:
```bash
/opt/wasi-sdk/bin/clang: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /opt/wasi-sdk/bin/clang)
/opt/wasi-sdk/bin/clang: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /opt/wasi-sdk/bin/clang)
/opt/wasi-sdk/bin/clang: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /opt/wasi-sdk/bin/clang)
```

### DIFF
--- a/.github/workflows/compilation_on_android_ubuntu.yml
+++ b/.github/workflows/compilation_on_android_ubuntu.yml
@@ -337,9 +337,13 @@ jobs:
           - os: ubuntu-20.04
             wasi_sdk_release: "https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-19/wasi-sdk-19.0-linux.tar.gz"
             wabt_release: "https://github.com/WebAssembly/wabt/releases/download/1.0.31/wabt-1.0.31-ubuntu.tar.gz"
+            wasi_sdk_folder_name: "wasi-sdk-19.0"
+            wasi_sysroot_option: "-DWASI_SYSROOT=`pwd`/../../../core/deps/wasi-libc/sysroot"
           - os: ubuntu-22.04
-            wasi_sdk_release: "https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-19/wasi-sdk-19.0-linux.tar.gz"
+            wasi_sdk_release: "https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-20%2Bthreads/wasi-sdk-20.0.threads-linux.tar.gz"
             wabt_release: "https://github.com/WebAssembly/wabt/releases/download/1.0.31/wabt-1.0.31-ubuntu.tar.gz"
+            wasi_sdk_folder_name: "wasi-sdk-20.0+threads"
+            wasi_sysroot_option: ""
     steps:
       - name: checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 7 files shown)


eloparco:
I opened an issue for the wasi-sdk problem on Ubuntu 20.04 https://github.com/WebAssembly/wasi-sdk/issues/303

.github/workflows/compilation_on_android_ubuntu.yml
```diff
@@ -431,13 +436,13 @@ jobs:
...
           mkdir build && cd build
-          cmake -DWASI_SYSROOT=`pwd`/../../../core/deps/wasi-libc/sysroot ..
+          cmake ${{ matrix.wasi_sysroot_option }} ..
           cmake --build . --config Release --parallel 4
           ./iwasm wasm-apps/no_pthread.wasm
 
   test:
     needs: [build_iwasm, build_llvm_libraries_on_ubuntu_2004, build_wamrc]
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
wenyongh, line 445: It may be a risk to upgrade the default test OS to Ubuntu-22.04. Maybe we can keep this PR open, and wait until the wasi-sdk-20.0 for threads works for Ubuntu-20.04?

eloparco: No strong opinion, let's wait for the decision on the issue I opened. Or, as an alternative, we could run those tests both with 20.04 and 22.04.

abrown: What's the risk? In https://github.com/WebAssembly/wasi-sdk/issues/303, I mentioned that I wouldn't mind if someone else (@eloparco?) works on getting wasi-sdk to build on older machines but I wasn't personally too interested. If this workflow is the only reason to do so, I wonder if we could resolve whatever risks are here? I assume this workflow would have to be updated at some point anyways since I wouldn't assume it to pin to `20.04` forever.

wenyongh: I was just worried about the spec cases and wasi-threads cases are not tested on Ubuntu-20.04, and we don't know whether there are errors reported on it. I discussed with @lum1n0us, he also agreed to run these tests on both Ubuntu-20.04 and Ubuntu-22.04, it seems better. So @eloparco, could you help enable that?

.github/workflows/compilation_on_android_ubuntu.yml
```diff
@@ -515,7 +536,7 @@ jobs:
           git fetch https://github.com/WebAssembly/wasi-libc \
             8f5275796a82f8ecfd0833a4f3f444fa37ed4546
           git checkout FETCH_HEAD
-          make -j \
+          make \
```
wenyongh, line 539: Using `make -j` can speed up the build process, did you find error when building with `-j` flag?

wenyongh (commented):
LGTM

.github/workflows/compilation_on_android_ubuntu.yml
eloparco: I must have removed it by mistake, I put it back. The `multi-tier-jit` test is failing in the CI, it doesn't look related to this PR.

PR #2022, opened 2023-03-11
Fix malloc non-thread-safe usage in test

eloparco:
In the WASI thread test modified in this PR, `malloc` was used in multiple threads without a lock. But wasi-libc implementation of malloc is not thread-safe.


wenyongh (approved):
LGTM

PR #2028, opened 2023-03-15
Implement sync primitives instead of using pthread ones

eloparco:
Attempt to replace pthread sync primitives since they seem to cause data races when running with the thread sanitizer. Those data races appear as concurrent accesses to load/store operations when running in classic interpreter mode.

In particular, this PR implements `mutex` and `barrier` to replace `pthread_mutex_t` and `pthread_barrier_t` in the tests with threads.


core/iwasm/libraries/lib-wasi-threads/test/sync_primitives.h
```diff
@@ -0,0 +1,75 @@
...
+    bool no_wait;
+
+    mutex_lock(&barrier->mutex);
+    barrier->count++;
+    no_wait = (barrier->count >= barrier->num_threads);
+    mutex_unlock(&barrier->mutex);
+
+    if (no_wait) {
+        __atomic_store_n(&barrier->ready, 1, __ATOMIC_SEQ_CST);
+        __builtin_wasm_memory_atomic_notify(&barrier->ready, 1);
```
g0djan, line 70: Why do you notify only 1 thread?

g0djan: Okay, it looked like it works with notifying only 1 thread because in such sample it works wrong on main_proc_exit_wait:
-> main thread starts actually waiting
-> 2 threads enter atomic.wait, but not actually waiting yet
->  3 thread increases the counter and notifies main thread
-> main thread wakes up and calls proc_exit before any other thread actually started waiting

I don't think it's possible to make a barrier without a cond_var

eloparco: > Why do you notify only 1 thread?

Nice catch, thanks! It got lost during the copy and paste.

> Okay, it looked like it works with notifying only 1 thread because in such sample it works wrong on main_proc_exit_wait:

No, I think it should be fine, it's an `atomic.wait` not a `pthread_cond_wait`, the condition variable is part of its implementation (one of the argument is `expect` in fact).

core/iwasm/libraries/lib-wasi-threads/test/sync_primitives.h
```diff
@@ -0,0 +1,76 @@
...
+    int num_threads;
+    int mutex;
+    int ready;
+} barrier_t;
+
+void
+barrier_init(barrier_t *barrier, int num_threads)
+{
+    barrier->count = 0;
+    barrier->num_threads = num_threads;
```
hritikgupta, line 53: why not just use one of these?

core/iwasm/libraries/lib-wasi-threads/test/sync_primitives.h
```diff
@@ -0,0 +1,76 @@
...
+    barrier->mutex = 0;
+    barrier->ready = 0;
+}
+
+void
+barrier_wait(barrier_t *barrier)
+{
+    bool no_wait;
```
hritikgupta, line 61: this variable looks redundant, can just rely on `num_threads`?

core/iwasm/libraries/lib-wasi-threads/test/sync_primitives.h
eloparco: Right, the point of having both is to reset the count to 0 once it reaches `num_threads`. I'll update the code.

core/iwasm/libraries/lib-wasi-threads/test/sync_primitives.h
eloparco: I don't think so, because we want to avoid accessing the shared variable once we release the lock.

eloparco:
After this PR I don't see the load/store race conditions anymore. I need to fix the failure on the specification test for atomics (that's breaking in the CI) and then do some more testing.

core/iwasm/libraries/lib-wasi-threads/test/sync_primitives.h
g0djan: Okay, you fixed number of notifying threads. But there is a possibility for deadlock with this logic.
- all threads leave section under lock
- counter has reached number of threads, so no_wait set to true in 1 of the threads
- this thread notifies all the other threads before the threads reached waiting call
-> all other threads hang infinetely

g0djan: Managed to hang it without changes from #2024, but with this changes it doesn't seem to be a problem

g0djan:
> After this PR I don't see the load/store race conditions anymore. I need to fix the failure on the specification test for atomics (that's breaking in the CI) and then do some more testing.

Me neither, no TSAN warnings

wenyongh:
@eloparco Wondering why we replace the `pthread_barrier_wait` with self-implemented `barrier_wait`, can we confirm that the hang issue is caused by wasi-libc `pthread_barrier_wait` or by runtime? If not, could we keep the implementation of both, and add a macro to control whether using it or self-implemented `barrier_wait`?

BTW, I tried to debug the hang issue and found that the last thread didn't notify others after it entered `pthread_barrier_wait`, seems there is memory overwritten. Need to check more. One thing to confirm, the aux stack of the thread created is allocated previously in the bytecode, no need to be allocated by runtime, right?

eloparco:
> @eloparco Wondering why we replace the pthread_barrier_wait with self-implemented barrier_wait, can we confirm that the hang issue is caused by wasi-libc pthread_barrier_wait or by runtime? If not, could we keep the implementation of both, and add a macro to control whether using it or self-implemented barrier_wait?

This PR is part of the investigation to understand if the hang problems are coming from the usage of pthread primitives from `wasi-libc`. Using the self-implemented mutex and barrier the TSAN warnings and deadlocks disappear.
Yes, we can keep both using a macro, I don't think it's worth running each test with both.

> BTW, I tried to debug the hang issue and found that the last thread didn't notify others after it entered pthread_barrier_wait, seems there is memory overwritten. Need to check more. One thing to confirm, the aux stack of the thread created is allocated previously in the bytecode, no need to be allocated by runtime, right?

I need to check, but it may just be the consequence of the data race since they cause undefined behavior?

g0djan:
@wenyongh prints stoped working for me as well at some point. I wanted to use WAMR source debugger so I can attach when it stuck, but the lld patch is out of date. I opened #2035 for that

g0djan:
Didn't manage to build wasmtime with thread sanitizer but managed to run tests that hang for wamr on `pthread_barrier_wait` and on wasmtime it doesn't hang after 10k runs, the same .wasm binary file easily gets hang after 500 iterations

wenyongh:
> Didn't manage to build wasmtime with thread sanitizer but managed to run tests that hang for wamr on `pthread_barrier_wait` and on wasmtime it doesn't hang after 10k runs, the same .wasm binary file easily hangs. on wamr after 500 iterations
> 
> So it might mean that problem is not with `pthread_barrier` actually in that case

Yes, I also found that it doesn't hang when running with WAMR AOT mode, so it should doesn't matter with `pthread_barrier_wait`.

The root cause might be the CPU runtime memory ordering:
https://en.wikipedia.org/wiki/Memory_ordering#Runtime_memory_ordering

For example, the data written by thread A is in cache, but not actually written into memory yet, and the data read by thread B is from memory and may be invalid. We may need to add some memory barrier operations in interpreter like what AOT does: I tried to added some, and it didn't hang so often, I uploaded a rough patch, maybe you can have a further investigation:
[mem_order.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/11007126/mem_order.zip)

wenyongh (approved):
LGTM

core/iwasm/libraries/lib-wasi-threads/test/common.h
```diff
@@ -6,11 +6,16 @@
...
 #include <assert.h>
-#include <pthread.h>
 #include <stdbool.h>
 #include <unistd.h>
 #include <limits.h>
 
+#if USE_PTHREAD_SYNC_PRIMITIVES != 0
+#include <pthread.h>
+#else
+#include "sync_primitives.h"
```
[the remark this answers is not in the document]
wenyongh: Yes, had better use the `pthread_barrier_wait` after the hang issue is resolved. But note that there will be some normal load/store data races in wasi-libc implementation, e.g. a_swap:
https://github.com/WebAssembly/wasi-libc/blob/main/libc-top-half/musl/src/internal/atomic.h#L106-L115
```C
	int old;
	do old = *p;
	while (a_cas(p, old, v) != old);
	return old;
```
Here `old = *p` is not atomic, and a_cas(p, old, v) is atomic, if two threads operate on the same address `p`, data races may be reported by sanitizer.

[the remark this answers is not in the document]
eloparco: > Yes, had better use the pthread_barrier_wait after the hang issue is resolved.

So should we merge this one as it is then? Until we get it fixed in wasi-libc.

> Here old = *p is not atomic, and a_cas(p, old, v) is atomic, if two threads operate on the same address p, data races may be reported by sanitizer.

Nice catch, that's probably what the thread sanitizer is complaining about for the load/store data races. It may be the same for other operations as well, I see the `a_fetch_add` a few lines later doing something similar.

loganek:
How much effort would be to fix wasi-libc? I personally would prefer to do it there rather than re-implementing those primitives; this is extra maintenance effort + WASI community won't benefit from your changes here as they're WAMR internal.

wenyongh:
> How much effort would be to fix wasi-libc? I personally would prefer to do it there rather than re-implementing those primitives; this is extra maintenance effort + WASI community won't benefit from your changes here as they're WAMR internal.

Just to remove a line in `libc-top-half/musl/arch/wasm32/atomic_arch.h`, I submitted a PR to it:
https://github.com/WebAssembly/wasi-libc/pull/403

ttrenner:
> How much effort would be to fix wasi-libc? I personally would prefer to do it there rather than re-implementing those primitives; this is extra maintenance effort + WASI community won't benefit from your changes here as they're WAMR internal.

The question can be raised in general: which things should be part of wasi and which should be part of wamr. Often this results in a chicken & egg problem: no one is implementing it as long as it is not standardized, but on the other hand no one pushes standardization if no one wants it to have/implement it. So I would propose to go forward and push it/make it visible at the community (with the risk of doing some waste).

loganek:
> > How much effort would be to fix wasi-libc? I personally would prefer to do it there rather than re-implementing those primitives; this is extra maintenance effort + WASI community won't benefit from your changes here as they're WAMR internal.
> 
> The question can be raised in general: which things should be part of wasi and which should be part of wamr. Often this results in a chicken & egg problem: no one is implementing it as long as it is not standardized, but on the other hand no one pushes standardization if no one wants it to have/implement it. So I would propose to go forward and push it/make it visible at the community (with the risk of doing some waste).

I don't think the question is which bits should be a part of wasi-libc and which should not be - those primitives are already part of wasi-libc, but they don't work (at least according to this PR). From what I understand, we re-implement those sync primitives to not use the (probably) broken ones from wasi-libc, but what I'm suggesting is that we should rather fix wasi-libc instead.

wenyongh:
@eloparco The PR for wasi-libc to fix the hang issue was merged, I think we may update CI to use that commit for both Ubuntu-20.04 and Ubuntu-22.04, and here use `pthread_barrier_wait` by default, so as to better test the wasi-threads cases. What's your opinion?

.github/workflows/compilation_on_android_ubuntu.yml
```diff
@@ -368,9 +368,9 @@ jobs:
           mkdir wasi-libc
           cd wasi-libc
           git init
-          # "Rename thread_spawn import" commit on main branch
+          # "Fix a_store operation in atomic.h" commit on main branch
           git fetch https://github.com/WebAssembly/wasi-libc \
-            8f5275796a82f8ecfd0833a4f3f444fa37ed4546
+            1dfe5c302d1c5ab621f7abf04620fae92700fd22
```
wenyongh, line 373: Had better also compile wasi-libc in Ubuntu-22.04 instead using the pre-release wasi-sdk-thread version since there is issue in it:
Merge L337 to L346, and remove  L366 and

.github/workflows/compilation_on_android_ubuntu.yml
```diff
@@ -532,9 +532,9 @@ jobs:
...
```
wenyongh, line 537: Same as above

.github/workflows/compilation_on_android_ubuntu.yml
eloparco: I thought of that, but it would mean that now we remove the usage wasi-sdk 20 pre-release from the ci, basically reverting all the changes previously added in #2021. When wasi-sdk 20 final release comes out we will have to add those changes again.

I was leaving the wasi-sdk 20 part as it is since in practice we didn't encounter problems in the main branch, the hanging issues, only reproducible by running each test many times, always passed unnoticed in the CI.

wenyongh: Got it, so let's keep it.

PR #2040, opened 2023-03-19
fix dead lock in source debugger

xujuntwt95329:


xujuntwt95329:
@eloparco This PR is based on your previous one: https://github.com/bytecodealliance/wasm-micro-runtime/pull/1991
Could you help review it?

Thanks

eloparco:
I ran the `main` branch locally and I could see the problem, with the lock already acquired in https://github.com/bytecodealliance/wasm-micro-runtime/blob/5c37ddfbca2350d82ff71d60cc71974fb7ef55c6/core/iwasm/interpreter/wasm_interp_classic.c#L1040 and then again in `wasm_cluster_thread_waiting_run`.

I hadn't tried the debugger since that PR so I didn't notice. Unfortunately, I don't think there is any simple way to automatically detect similar issues in the CI.

eloparco (approved):
LGTM

xujuntwt95329:
@eloparco Thanks a lot!
Yes, there seems not an easy way to validate source debugger in CI. 
A possible approach may be using lldb scripts, which can execute some commands without terminal input, the problem is how to validate the result of every command.

PR #2043, opened 2023-03-21
set gc reclaim in wasm_exec_env_create

yviansu:


core/iwasm/common/wasm_exec_env.c
```diff
@@ -154,6 +163,9 @@ wasm_exec_env_create(struct WASMModuleInstanceCommon *module_inst,
         exec_env->aux_stack_bottom.bottom = module->aux_stack_bottom;
         exec_env->aux_stack_boundary.boundary =
             module->aux_stack_bottom - module->aux_stack_size;
+#if WASM_ENABLE_GC != 0
+        gc_heap_handle = ((AOTModuleInstance *)module_inst)->e->gc_heap_handle;
```
wenyongh, line 167: The gc_heap_handle for AOTModuleInstanceExtra isn't available now, had better comment this line to avoid compile/run error.

yviansu: Get~

PR #2047, opened 2023-03-23
Use pre-created exec_env for instantiation and module_malloc/free

wenyongh:
Use pre-created exec_env for instantiation and module_malloc/free,
use the same exec_env of the current thread to avoid potential
unexpected behavior.

And remove unnecessary shared_mem_lock in wasm_module_free,
which may cause dead lock.


yamt:
> Use pre-created exec_env for instantiation and module_malloc/free, use the same exec_env of the current thread to avoid potential unexpected behavior.

@wenyongh 
what unexpected behavior did this fix?

wenyongh:
> > Use pre-created exec_env for instantiation and module_malloc/free, use the same exec_env of the current thread to avoid potential unexpected behavior.
> 
> @wenyongh what unexpected behavior did this fix?

Hi, I remember a major issue is when hw bound check is enabled, the thread local variable `exec_env_tls` is set when calling wasm/aot function, and before the calling ends, if there is another function call to instantiate or malloc/free memroy (from native API or in creating child thread), and if we use another exec_env to call the function, then the exception "invalid exec env" will be thrown:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/aot/aot_runtime.c#L1364

And if we reset the `exec_env_tls` to another exec_env, I am not sure how to restore it after instantiation or malloc/free. `exec_env_tls` is used in signal handler to get the jmpbuf and then longjmp. Maybe we can temporarily save exec_env_tls in execute_post_instantiate_functions/module_malloc/module_free and then restore it if needed?

wenyongh:
Another issue is it may impact the dump call stack feature, there are two exec_envs used in the function call chain, we can only dump the call stack in the latest exec_env if exception is thrown.

yamt:
> > > Use pre-created exec_env for instantiation and module_malloc/free, use the same exec_env of the current thread to avoid potential unexpected behavior.
> > 
> > 
> > @wenyongh what unexpected behavior did this fix?
> 
> Hi, I remember a major issue is when hw bound check is enabled, the thread local variable `exec_env_tls` is set when calling wasm/aot function, and before the calling ends, if there is another function call to instantiate or malloc/free memroy (from native API or in creating child thread), and if we use another exec_env to call the function, then the exception "invalid exec env" will be thrown: https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/aot/aot_runtime.c#L1364

i see. thank you for explanation.

does `wasm_interp_call_func_import` avoid switching exec_env for the same reason?

> And if we reset the `exec_env_tls` to another exec_env, I am not sure how to restore it after instantiation or malloc/free. `exec_env_tls` is used in signal handler to get the jmpbuf and then longjmp. Maybe we can temporarily save exec_env_tls in execute_post_instantiate_functions/module_malloc/module_free and then restore it if needed?

sure. it makes sense to update exec_env_tls to sync with the "current effective" exec_env more frequently.

yamt:
> Another issue is it may impact the dump call stack feature, there are two exec_envs used in the function call chain, we can only dump the call stack in the latest exec_env if exception is thrown.

i guess it makes sense to always use a single exec_env for a thread and pass it all over places as a function argument, instead of module_inst.
(except the places where it's impossible, where we use exec_env_tls.)
i did something similar for another engine and it works fine so far.
but for wamr it's a major surgery involving a lot of api/abi changes...

wenyongh:
> > > > Use pre-created exec_env for instantiation and module_malloc/free, use the same exec_env of the current thread to avoid potential unexpected behavior.
> > > 
> > > 
> > > @wenyongh what unexpected behavior did this fix?
> > 
> > 
> > Hi, I remember a major issue is when hw bound check is enabled, the thread local variable `exec_env_tls` is set when calling wasm/aot function, and before the calling ends, if there is another function call to instantiate or malloc/free memroy (from native API or in creating child thread), and if we use another exec_env to call the function, then the exception "invalid exec env" will be thrown: https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/aot/aot_runtime.c#L1364
> 
> i see. thank you for explanation.
> 
> does `wasm_interp_call_func_import` avoid switching exec_env for the same reason?
> 

Not exactly, it is a reason, but another reason is for passing function results to the caller's frame (in caller's exec_env) from callee's frame (in callee's exec_env), `wasm_interp_call_func_import` calls `wasm_interp_call_func_bytecode(...,exec_env, prev_frame)` with `exec_env = curent exec_env, prev_frame = current exec_env's current frame`, and in `wasm_interp_call_func_bytecode`, when returning from function call, it writes results into `prev_frame->sp[..]`:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_interp_classic.c#L1435-L1437

If using different exec_env, then before calling `wasm_interp_call_func_bytecode` in `wasm_interp_call_func_import`, we should allocate a frame as prev_frame for `wasm_interp_call_func_bytecode`, like what we do in `wasm_interp_call_wasm`:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_interp_classic.c#L4238

PR #2049, opened 2023-03-23
add architecture document for wasm export

xwang98:


core/iwasm/doc/wasm_exports.MD
```diff
@@ -0,0 +1,22 @@
...
+
+A `WASMExport` item contains three elements that map the Wasm file exports section structure:
+- name: the name shown to external
+- kind: total 4 export types: function, globals, memory, table
+- index: As all the 4 export types are organized in array, this refers to index in target export type 
+
+## Function exports
+### use function exports
+function exports are often used in two situations:  
+    1. **call by host**: runtime API `wasm_runtime_lookup_function` will walk through the array of `WASMModuleInstance::export_functions` and compare the exported name with given target symbol name in the function parameter. If any array item matches the name, then return the value of field `function` which points to associated function instance (WASMFunctionInstance)
```
wenyongh, line 16: Had better?
If any array item matches the name, then return the value of field
->
If any array item matches the name, it then returns the value of field

xwang98: ok, I will update it according to your comments

PR #2052, opened 2023-03-23
Release 1.1.3

g0djan:
Considering number of bug-fixes and recent elimination of known problems with WASI threads we think it would be a good idea to release a new version

### DIFF
--- a/core/version.h
+++ b/core/version.h
@@ -7,5 +7,5 @@
 #define _WAMR_VERSION_H_
 #define WAMR_VERSION_MAJOR 1
 #define WAMR_VERSION_MINOR 1
-#define WAMR_VERSION_PATCH 2
+#define WAMR_VERSION_PATCH 3
 #endif


wenyongh:
> @wenyongh We would appreciate if it can be released this week, so it will make it possible for us to use a stable release version that supports WASI threads in our internal release in the next few days. Thank you!

@g0djan We are preparing to release the new version, but it really takes time and effort to do more tests, normally we hope to finish the test and release the new version before the end of next week. Is it OK for you? If not, we may create a pre-release version this week for you.
BTW, the next version should be 1.2.0, since there are big features implemented, e.g. Multi-tier JIT and wasi-threads.

g0djan:
If pre-release also means some tests, then we will be happy with it as well. We aim for 27th, thank you.

wenyongh:
> If pre-release also means some tests, then we will be happy with it as well. We aim for 27th, thank you.

@g0djan WAMR-1.2.0 was created, thanks.
https://github.com/bytecodealliance/wasm-micro-runtime/releases/tag/WAMR-1.2.0

PR #2072, opened 2023-03-28
Fix module_malloc/module_free issues

wenyongh:
Try using existing exec_env to execute wasm app's malloc/free func and
execute post instantiation functions. Create a new exec_env only when
no existing exec_env was found.


xujuntwt95329 (approved):
LGTM

PR #2095, opened 2023-04-02
Upgrade to wasi-sdk 20

eloparco:
With respect to wasi-sdk 20 pre-release, wasi-sdk 20 supports older versions of glibc, allowing us to use it in the CI with Ubuntu 20.04.

See #2021 for previous upgrade to wasi-sdk 20 pre-release.

### DIFF
--- a/.github/workflows/compilation_on_android_ubuntu.yml
+++ b/.github/workflows/compilation_on_android_ubuntu.yml
@@ -274,7 +274,7 @@ jobs:
         os: [ubuntu-20.04, ubuntu-22.04]
         wasi_sdk_release:
           [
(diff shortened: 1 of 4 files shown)


eloparco:
I upgraded only the Ubuntu and MacOS CI for now, should we update everything (nuttx, sgx, release) to use version 20?

wenyongh:
> I upgraded only the Ubuntu and MacOS CI for now, should we update everything (nuttx, sgx, release) to use version 20?

Had better not, we haven't tested it a lot yet. Seems someone found and fixed an issue related to malloc in wasi-libc:
  https://github.com/WebAssembly/wasi-libc/pull/404
Maybe we can upgrade to a more stable version in the future.

PR #2107, opened 2023-04-05
User can provide a different vprintf() at runtime

Photosounder:
I needed a better mechanism for printing WAMR's messages than defining `BH_VPRINTF` (not a good mechanism if you want to build the library once and dynamically link to it). So now there's a global called `ptr_vprintf` which by default is equal to `vprintf`, and in all the versions of `os_printf()` and `os_vprintf()` calls to `vprintf()` are now replaced by calls to `ptr_vprintf()`, which if the user does nothing is exactly the same.

However if the user calls `bh_set_vprintf(my_vprintf);` then all calls to `ptr_vprintf` are now calls to the user's function, which can allow for things such as logging to a buffer than can be displayed in the same program's GUI or logged to a file. So my change breaks nothing but adds the possibility to change at runtime whether `vprintf()` or another function is used by `os_printf()`.

I'm just not sure if `bh_log.[ch]` were the right files to put this, but there was no better non-platform specific file to put this in. Btw there's an `os_printf()` and `os_vprintf()` for every platform but they're all the same, you might want to consider grouping them into one. Also I haven't added this mechanism to platforms that just use `#define os_printf printf`, but if anything is to be done about this it would probably to make one `os_printf()` for all platforms.

### DIFF
--- a/core/shared/utils/bh_log.c
+++ b/core/shared/utils/bh_log.c
@@ -11,9 +11,10 @@
  */
 static uint32 log_verbose_level = BH_LOG_LEVEL_WARNING;
 
-int (*ptr_vprintf)(const char*,va_list) = vprintf;
+int (*ptr_vprintf)(const char *, va_list) = vprintf;
 
-void bh_set_vprintf(int (*ptr)(const char*,va_list))
+void
+bh_set_vprintf(int (*ptr)(const char *, va_list))
 {
     ptr_vprintf = ptr;
 }
--- a/core/shared/utils/bh_log.c
+++ b/core/shared/utils/bh_log.c
@@ -4,8 +4,6 @@
  */
 
 #include "bh_log.h"
-#include <stdio.h>
-#include <stdarg.h>
 
 /**
  * The verbose level of the log system.  Only those verbose logs whose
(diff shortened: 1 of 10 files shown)


core/shared/utils/bh_log.c
```diff
@@ -11,6 +11,18 @@
  */
 static uint32 log_verbose_level = BH_LOG_LEVEL_WARNING;
 
+#ifndef BH_PLATFORM_LINUX_SGX
+int (*ptr_vprintf)(const char *, va_list) = vprintf;
```
xujuntwt95329, line 15: This will make the platform layer depends on the utils, which is not reasonable.

Had better put the `ptr_vprintf` to every platform.

Also the `bh_set_vprintf` should be moved to every platform's directory and rename to `os_set_vprintf`

Of course this will introduce some duplicate code, but avoid breaking the architecture

wenyongh: Yes, had better declare `os_set_vprintf` in `platform_api_vmcore.h`, and implement it in platform layer besides with `os_printf` and `os_vprintf`.

wenyongh:
Close this PR as it hasn't been update for a long time.

PR #2119, opened 2023-04-10
Remove the reference for binaryen.Module for DataSegmentContext

coderebot:
BUG=N/A

### DIFF
--- a/ts2wasm/src/backend/index.ts
+++ b/ts2wasm/src/backend/index.ts
@@ -9,3 +9,95 @@ export abstract class Ts2wasmBackend {
     public abstract emitText(options?: any): string;
     public abstract dispose(): void;
 }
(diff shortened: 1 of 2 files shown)


ts2wasm/src/backend/index.ts
```diff
@@ -9,3 +9,95 @@ export abstract class Ts2wasmBackend {
     public abstract emitText(options?: any): string;
     public abstract dispose(): void;
 }
+
+
+export interface SegmentInfo {
```
xujuntwt95329, line 14: This file is used to describe the common base class for all backends, it shouldn't contain specific implementations.

The `DataSegmentContext` is only used in `binaryen` backend, it would be better to keep this inside `binaryen` folder

PR #2121, opened 2023-04-11
Support execution on WAMR and refactor test cases

xujuntwt95329:
* re-org test cases
* move semantic checker as a separate visitor
* enable execution on WAMR
* fix interface method access issue
* fix semantic checking for operation between any and primitive type
* implement string_slice function
* update feature list


brucehuke:
LGTM

PR #2122, opened 2023-04-11
Updated request.ts wasm_response_send signature

sax1johno:
Updated wasm_response_send function signature to match api by returning bool instead of void (was causing assemblyscript example to break)


xujuntwt95329 (approved):
LGTM

PR #2132, opened 2023-04-18
build_wamr_lldb.yml: sync lldb build options between ubuntu and macos

yamt:
Notably, this disables python scripting support on ubuntu:

* It isn't necessary for the primary purpose of this build. (that is, the vscode extension)
* It has been disabled in the macOS counterpart.
* It allows to use the same binary for 22.04 and 20.04. (note: 22.04 and 20.04 provide different versions of libpython)


yamt:
assets generated with identical patch: https://github.com/yamt/wasm-micro-runtime/releases/tag/WAMR-1.2.999

xujuntwt95329 (commented):
LGTM

PR #2139, opened 2023-04-20
VSCode-Extension: download lldb built for ubuntu 20.04

yamt:
This should allow users to use the vscode extension on ubuntu 20.04.

After https://github.com/bytecodealliance/wasm-micro-runtime/pull/2132 , our lldb binary for 20.04 works on ubuntu 22.04 as well.

On the other hand, lldb for 22.04 has no chance to work on ubuntu 20.04. (because of glibc version requirement.)


yamt:
does anyone know a good way to test this kind of changes w/o actually making a release?

PR #2140, opened 2023-04-20
Fix ems allocator unaligned memory access on riscv64

wenyongh:
Add padding bytes to ensure the left/right/parent fields in struct hmu_tree_node
are 8-byte aligned on the target which doesn't support unaligned memory access.

### DIFF
--- a/core/shared/mem-alloc/ems/ems_gc_internal.h
+++ b/core/shared/mem-alloc/ems/ems_gc_internal.h
@@ -207,6 +207,12 @@ set_hmu_normal_node_next(hmu_normal_node_t *node, hmu_normal_node_t *next)
 typedef struct hmu_tree_node {
     hmu_t hmu_header;
     gc_size_t size;
+#if UINTPTR_MAX == UINT64_MAX && WASM_CPU_SUPPORTS_UNALIGNED_ADDR_ACCESS == 0
+    /* each hmu node is 4-byte aligned and not 8-byte aligned,
+       add 4 bytes padding to make below left/right/parent fields
+       8-byte aligned, so that we can directly access them */
+    uint32 __padding;
+#endif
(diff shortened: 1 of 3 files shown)


core/shared/mem-alloc/ems/ems_gc_internal.h
```diff
@@ -207,6 +207,12 @@ set_hmu_normal_node_next(hmu_normal_node_t *node, hmu_normal_node_t *next)
 typedef struct hmu_tree_node {
     hmu_t hmu_header;
     gc_size_t size;
+#if UINTPTR_MAX == UINT64_MAX && WASM_CPU_SUPPORTS_UNALIGNED_ADDR_ACCESS == 0
+    /* each hmu node is 4-byte aligned and not 8-byte aligned,
+       add 4 bytes padding to make below left/right/parent fields
+       8-byte aligned, so that we can directly access them */
+    uint32 __padding;
```
Zzzabiyaka, line 214: Hi, 

Cannot find where you set this variable but generally very good change for the project as currently ubsan wamr build fails with unaligned access

Zzzabiyaka (approved):

core/shared/mem-alloc/ems/ems_gc_internal.h
wenyongh: Do you mean that before this change, ubsan reports unaligned access issue, and after this PR, the issue isn't reported again? Maybe I should apply the change for all the targets, including x86-64.

Zzzabiyaka: Hi 

Now it reports unaligned access but unfortunately I didn't try to fix padding for x86-64 now, and I hope something like that could help

PR #2142, opened 2023-04-20
Update sample workload xnnpack

TianlongLiang:
Add a shell script for building xnnpack examples and run it

### DIFF
--- a/samples/workload/XNNPACK/build.sh
+++ b/samples/workload/XNNPACK/build.sh
@@ -26,23 +26,25 @@ if [ -d "$xnnpack_out_dir" ]; then
   file_count=$(ls "$xnnpack_out_dir" | grep ".wasm$" | wc -l)
   # TODO: currently should be 116 files, may changes in the future
   if [ $file_count -eq 116 ]; then
-    echo "Fully build XNNPACK benchmark wasm files"
-    need_build=false
+    echo "Have fully built XNNPACK benchmark wasm files before, rebuild it to update"
+    need_fresh_build=false
+    cmake -B build
+    cmake --build build
   else
-    echo "Partially build XNNPACK wasm files, rebuild again"
-    need_build=true
+    echo "Partially build XNNPACK wasm files, start a fresh build"
+    need_fresh_build=true
   fi
 else
-  echo "XNNPACK build directory does not exist"
-  need_build=true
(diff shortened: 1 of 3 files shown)


samples/workload/XNNPACK/build.sh
```diff
@@ -0,0 +1,127 @@
...
+
+# 1. build xnnpack wasm files
+# 1.1 to decide whether needs to build xnnpack wasm files
+cd ${WAMR_DIR}/samples/workload/XNNPACK
+xnnpack_out_dir=xnnpack/bazel-bin
+if [ -d "$xnnpack_out_dir" ]; then
+  echo "XNNPACK build directory exists"
+  file_count=$(ls "$xnnpack_out_dir" | grep ".wasm$" | wc -l)
+  # TODO: currently should be 116 files, may changes in the future
+  if [ $file_count -eq 116 ]; then
```
lum1n0us, line 28: `egrep "\/\/:.*\.wasm" CMakeLists.txt | wc` can get the amount of .wasm

samples/workload/XNNPACK/build.sh
```diff
@@ -0,0 +1,127 @@
...
+  cmake --build build
+fi
+
+# 1.3 copy files to out/
+rm -rf ${OUT_DIR}
+mkdir ${OUT_DIR}
+cp xnnpack/bazel-bin/*.wasm out/
+
+# 2. compile xnnpack benchmark wasm files to benchmark aot files with wamrc
+# 2.1 build wamr-compiler if needed
```
lum1n0us, line 56:
```suggestion
# 2.1 build wamr-compiler and llvm libraries if needed
```

samples/workload/XNNPACK/build.sh
```diff
@@ -0,0 +1,127 @@
...
+cd ${WAMRC_DIR}
```
lum1n0us, line 57: maybe some `test` first to achieve "build if needed"

lum1n0us: And if *.wasm -> *.aot is necessary, we shall add it in *CMakeLists.txt*

TianlongLiang: SIMD feature is only supported on aot mode, so maybe it's a good idea to add steps `building wamrc` and `translate wasm to aot` to CMakeLists.txt?

PR #2147, opened 2023-04-24
Add ubsan checks to WAMR

Zzzabiyaka:
this PR adds ubsan checks with no aligment enabled to basic CI tests.

### DIFF
--- a/samples/wasm-c-api/CMakeLists.txt
+++ b/samples/wasm-c-api/CMakeLists.txt
@@ -24,7 +24,7 @@ if (APPLE)
   add_definitions(-DBH_PLATFORM_DARWIN)
 endif ()
 
-# Resetdefault linker flags
+# Reset default linker flags
 set(CMAKE_SHARED_LIBRARY_LINK_C_FLAGS "")
 set(CMAKE_SHARED_LIBRARY_LINK_CXX_FLAGS "")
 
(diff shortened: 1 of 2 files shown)


lum1n0us:
LGTM

samples/wasm-c-api/CMakeLists.txt
```diff
@@ -87,6 +87,16 @@ endif()
...
 include (${WAMR_ROOT_DIR}/build-scripts/runtime_lib.cmake)
 
+if (NOT DEFINED SANITIZER)
+  set(SANITIZER "")
+endif()
+
+if (SANITIZER STREQUAL "ubsan")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -g -O2 -fno-omit-frame-pointer -fsanitize=undefined -fno-sanitize-recover=all -fno-sanitize=alignment" )
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fsanitize=undefined")
+endif()
```
loganek, line 97: could we possibly add an `else` branch to warn about unsupported / invalid sanitizer?

samples/wasm-c-api/CMakeLists.txt
```diff
@@ -87,6 +87,16 @@ endif()
 set(WAMR_ROOT_DIR ${CMAKE_CURRENT_LIST_DIR}/../..)
 include (${WAMR_ROOT_DIR}/build-scripts/runtime_lib.cmake)
 
+if (NOT DEFINED SANITIZER)
+  set(SANITIZER "")
+endif()
+
+if (SANITIZER STREQUAL "ubsan")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -g -O2 -fno-omit-frame-pointer -fsanitize=undefined -fno-sanitize-recover=all -fno-sanitize=alignment" )
```
loganek, line 95: Do we also need to check for a compiler here, or are we ok with the compiler (sliently) failing if those flags are not supported?

Zzzabiyaka: I believe that you won't set these flags for the wrong the wrong compiler on the CI side. 

But if you think it's better add here I'll add

samples/wasm-c-api/CMakeLists.txt
Zzzabiyaka: yeah, we probably can. 

I mean, I'm not sure it's harmful in any way, but let's add for easier debug in case of any errors

samples/wasm-c-api/CMakeLists.txt
```diff
@@ -87,6 +87,15 @@ endif()
 set(WAMR_ROOT_DIR ${CMAKE_CURRENT_LIST_DIR}/../..)
 include (${WAMR_ROOT_DIR}/build-scripts/runtime_lib.cmake)
 
+if (NOT DEFINED SANITIZER)
+  set(SANITIZER "")
```
loganek, line 91: is that variable going to be used outside of this if..else block that we set it here?

loganek (approved):
looks good, thanks.

PR #2152, opened 2023-04-25
Fix poll_oneoff, size_t < 8 bytes on some platforms

g0djan:
> According to the 1999 ISO C standard (C99), size_t is an unsigned integer type of at least 16 bit (see sections 7.17 and 7.18.3)
https://en.cppreference.com/w/cpp/types/size_t

### DIFF
--- a/core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
+++ b/core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
@@ -56,8 +56,14 @@ typedef struct WASIContext *wasi_ctx_t;
 wasi_ctx_t
 wasm_runtime_get_wasi_ctx(wasm_module_inst_t module_inst);
 
-static inline size_t
-min(size_t a, size_t b)
+static inline uint64_t
+min_uint64(uint64_t a, uint64_t b)
+{
+    return a > b ? b : a;
+}
+
+static inline uint32_t
+min_uint32(uint32_t a, uint32_t b)
 {
     return a > b ? b : a;
 }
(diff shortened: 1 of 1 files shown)


core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -961,8 +961,9 @@ get_timeout_for_poll_oneoff(const wasi_subscription_t *in,
     for (i = 0; i < nsubscriptions; ++i) {
         const __wasi_subscription_t *s = &in[i];
         if (s->u.type == __WASI_EVENTTYPE_CLOCK
-            && (s->u.u.clock.flags & __WASI_SUBSCRIPTION_CLOCK_ABSTIME) == 0) {
-            timeout = min(timeout, s->u.u.clock.timeout);
+            && (s->u.u.clock.flags & __WASI_SUBSCRIPTION_CLOCK_ABSTIME) == 0
+            && s->u.u.clock.timeout < timeout) {
+            timeout = s->u.u.clock.timeout;
```
loganek, line 966: Good catch; I wonder if other use of min (https://github.com/bytecodealliance/wasm-micro-runtime/pull/2152/files#diff-18a1c12794da572224835cdd9491ff7f93bc7ff81249c75ead3052e7f815f9f8R1021) would potentially be invalid too as it can go over `size_t` type?

Zzzabiyaka: From my point of view, looks like the answer is "no" 

It's defined in the same file and defined to compare two `size_t` which is not the safest way of doing that 

Probably we can refactor it to compare two `uint64_t`/`int64_t` . It'll be much safer I think because cast to a  wider type is always ok

wenyongh (approved):
LGTM

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
g0djan: Fair concern, fixed

loganek (approved):

PR #2161, opened 2023-04-27
Add ASAN and nightly build to WAMR CI

Zzzabiyaka:
Hi 

This CR adds nightly (UTC time) checks with asan and ubsan. Currently, asan works only with aot so adding it for other modes (as well as adding tsan) is tbd. 

Also I've put gcc4.8 build to nightly run as I think we don't need to run it with every PR


eloparco:
Are we running `tsan` with `WASI_TEST_OPTIONS`? I couldn't find it in the jobs. And I would except that to report errors (coming from `wasi-libc`).

wenyongh:
> Asking for help with CI in this PR.
> 
> If you can take a look and suggest anything that will help to fix this CI issues (see asan and tsan build in github actions) I will really appreciate that.
> 
> Also feel free to share you opinion about what the optimization should we use for san build (even `-O2` can hide some errors because of compiler optimizations so now it's set to `-O0` in testing purposes)

Hi, it may be not so convenient to view the github actions to check the issues reported, and also it is complex to resolve all the issues for both asan and tsan. I think we had better file new issues to provide more details (e.g. one issue for asan and the other for tsan, list some errors reported and describe how to reproduce them), and then others can discuss and check how to resolve the issues one by one.

product-mini/platforms/linux/CMakeLists.txt
```diff
@@ -130,6 +130,23 @@ set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra -Wformat -Wformat-security -W
 
 set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wformat -Wformat-security -Wno-unused")
 
+if (NOT DEFINED SANITIZER)
```
lum1n0us, line 133: maybe `SANITIZER` -> `WAMR_BUILD_SANITIZER`

@wenyongh what's your suggestion?

wenyongh: Agree, had better use `WAMR_BUILD_SANITIZER`, and apply these changes to `build-scripts/config_common.cmake`, no need to add similar lines in three files: `product-mini/platforms/linux/CMakeLists.txt`, `samples/wasi-threads/CMakeLists.txt` and `samples-wasm-c-api/CMakeLists.txt`.

wenyongh: And this PR adds more sanitizer modes in `compilation_on_android_ubuntu.yml`, which introduces lots of CI jobs (from 515 to 675) and increases the test time when PR is created/merged, my suggestion is to add another CI to daily test them.

Zzzabiyaka: done

.github/workflows/compilation_on_android_ubuntu.yml
```diff
@@ -115,84 +115,6 @@ jobs:
           cmake --build . --config Release --parallel 4
         working-directory: wamr-compiler
 
-  build_iwasm_linux_gcc4_8:
```
wenyongh, line 118: The building with gcc-4.8 is also moved to nightly build, right? @loganek, is it OK for you?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -4212,6 +4212,9 @@ static V128FuncPtr invokeNative_V128 = (V128FuncPtr)(uintptr_t)invokeNative;
           || defined(BUILD_TARGET_RISCV64_LP64) */
 #endif /* end of defined(_WIN32) || defined(_WIN32_) */
 
+#if defined(__GNUC__)
+__attribute__((no_sanitize_address))
```
wenyongh, line 4216: Not sure why add `no_sanitize_address` here? Maybe we can resolve the issue, do you know how to reproduce the issue?

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -3986,6 +3986,9 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
 }
 
 #if WASM_ENABLE_FAST_JIT != 0
+#if defined(__GNUC__)
+__attribute__((no_sanitize_address))
```
wenyongh, line 3990: Same as above

.github/workflows/compilation_on_android_ubuntu.yml
Zzzabiyaka: yeah, I think that's right thing to do

loganek: I think it's fine to move it to nightly.

wenyongh (commented):
LGTM

lum1n0us:
LGTM

build-scripts/config_common.cmake
```diff
@@ -127,6 +127,29 @@ else ()
...
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -g -O0 -fno-omit-frame-pointer -fsanitize=undefined -fno-sanitize-recover=all -fno-sanitize=alignment" )
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fsanitize=undefined")
+elseif (WAMR_BUILD_SANITIZER STREQUAL "asan")
+  if (NOT WAMR_BUILD_JIT EQUAL 1)
+    set (ASAN_OPTIONS "verbosity=2 debug=true ")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -g -O0 -fno-omit-frame-pointer -fsanitize=address -fno-sanitize-recover=all" )
+    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fsanitize=address")
+  endif()
+elseif (WAMR_BUILD_SANITIZER STREQUAL "tsan") 
+  # set (ASAN_OPTIONS "verbosity=2 debug=true ")
```
loganek, line 146: This should probably be removed or uncommented.

core/shared/platform/common/posix/posix_thread.c
```diff
@@ -492,6 +492,9 @@ destroy_stack_guard_pages()
 }
 #endif /* end of WASM_DISABLE_STACK_HW_BOUND_CHECK == 0 */
 
+#if defined(__GNUC__)
+__attribute__((no_sanitize_address))
```
loganek, line 496: could we please have a comment in the code explaining why do we disable sanitizer for this function?

core/shared/platform/common/posix/posix_thread.c
```diff
@@ -506,6 +509,9 @@ mask_signals(int how)
 static os_thread_local_attribute struct sigaction prev_sig_act_SIGSEGV;
 static os_thread_local_attribute struct sigaction prev_sig_act_SIGBUS;
 
+#if defined(__GNUC__)
+__attribute__((no_sanitize_address))
```
loganek, line 513: same here.

loganek (approved):
LGTM, left a few minor comments though.

core/iwasm/common/wasm_runtime_common.c
wenyongh: Got it, thanks.

build-scripts/config_common.cmake
Zzzabiyaka: done

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -4212,6 +4212,12 @@ static V128FuncPtr invokeNative_V128 = (V128FuncPtr)(uintptr_t)invokeNative;
           || defined(BUILD_TARGET_RISCV64_LP64) */
 #endif /* end of defined(_WIN32) || defined(_WIN32_) */
 
+// ASAN is not designed to work with custom stack unwind or other low-level things.
+// > Ignore a function that does some low-level magic.
+// (e.g. walking through the thread's stack bypassing the frame boundaries)
+#if defined(__GNUC__)
```
wenyongh, line 4218: Could you use `/* .. */` to add comments? We usually do that since it is C source file.

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -3986,6 +3986,12 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
 }
 
 #if WASM_ENABLE_FAST_JIT != 0
+// ASAN is not designed to work with custom stack unwind or other low-level things.
+// > Ignore a function that does some low-level magic.
+// (e.g. walking through the thread's stack bypassing the frame boundaries)
```
wenyongh, line 3991: Same as above

core/shared/platform/common/posix/posix_thread.c
```diff
@@ -492,6 +492,12 @@ destroy_stack_guard_pages()
 }
 #endif /* end of WASM_DISABLE_STACK_HW_BOUND_CHECK == 0 */
 
+// ASAN is not designed to work with custom stack unwind or other low-level things.
+// > Ignore a function that does some low-level magic.
+// (e.g. walking through the thread's stack bypassing the frame boundaries)
```
wenyongh, line 497: Same as above

core/iwasm/common/wasm_runtime_common.c
wenyongh: And CI reported code guideline check failure, the length of a line should be no larger than 80.

.github/workflows/compilation_on_android_ubuntu.yml
```diff
@@ -513,6 +434,7 @@ jobs:
       ]
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
```
eloparco, line 437: What is this for?

.github/workflows/nightly_run.yml
```diff
@@ -0,0 +1,622 @@
+# Copyright (C) 2019 Intel Corporation.  All rights reserved.
```
eloparco, line 1:
```suggestion
# Copyright (C) 2023 Intel Corporation.  All rights reserved.
```

eloparco, line 1: It looks like most of the code here is in common with other jobs, any easy way to avoid this duplication?

core/iwasm/common/wasm_runtime_common.c
Zzzabiyaka: Thank you, tried clang-format but it didn't help for some reason. Will fix

.github/workflows/nightly_run.yml
Zzzabiyaka: will try

Zzzabiyaka: i've checked about github actions reusage, the only thing is https://github.com/actions/starter-workflows/issues/245
for me the only thins is they allow now, is to have a "template" in other file to call it with different inputs (eg the other file is one/multiples job which is called with different inputs)
we could do this like
```
prepare:
  uses ./.github/common.yml
  ```
but then we lose needs functionality
if later you write
```
stepN:
  needs[build_llvm]
```
you will get the error because this job is not actually in this file and it can't find it's name
as I know "waterfall" (or fallback) syntaxis is not supported, otherwise we could do
```
j1:
j2:
j3:
j4:
  uses ./.github/common.yml
```
In short, don't see any pretty enough way

Please correct me if others have any good ideas about it

core/iwasm/interpreter/wasm_interp_classic.c
Zzzabiyaka: done

core/shared/platform/common/posix/posix_thread.c
Zzzabiyaka: done

.github/workflows/nightly_run.yml
lum1n0us: we used to apply [Reusing workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows) in binary releasing scripts. https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/.github/workflows/release_process.yml#L56

It can help to handle some of duplications but not a silver bullet for all of them.

Zzzabiyaka: yeah, it applies there because there's just one job triggered with different input. I don't think it will add much sense for us unfortunately (it's nice if you have only one job in file, if you have 2+ every `uses` instruction will trigger them all) so I'd not do that in this PR at least

wenyongh (commented):
LGTM

PR #2162, opened 2023-04-28
Add external delegation to support several NPU/GPU

ayakoakasaka:
Add VX delegation as an external delegation of TFLite. So that several NPU/GPU (from VeriSilicon, NXP, Amlogic) can be controlled via WASI-NN. Test Code can work with the X86 simulator.


core/iwasm/libraries/wasi-nn/src/wasi_nn_tensorflowlite.cpp
```diff
@@ -206,6 +210,17 @@ tensorflowlite_init_execution_context(void *tflite_ctx, graph g,
                 NN_ERR_PRINTF("Error when enabling GPU delegate.");
                 use_default = true;
             }
+#elif defined(WASI_NN_ENABLE_EXTERNAL_DELEGATE)
+            NN_WARN_PRINTF("external delegation enabled.");
+            auto options =
+                TfLiteExternalDelegateOptionsDefault(WASI_NN_EXT_DELEGATE_PATH);
+            auto *delegate = TfLiteExternalDelegateCreate(&options);
```
wenyongh, line 217: Had better check the return value? And same to do for L206?

wenyongh:
@tonibofarull Could you help review this PR? Thanks.

core/iwasm/libraries/wasi-nn/src/wasi_nn_tensorflowlite.cpp
tonibofarull: +1

wenyongh (commented):
LGTM

tonibofarull (approved):

PR #2169, opened 2023-05-02
Return error when exception encountered after main thread finishes

eloparco:
Currently, if a thread is spawned and raises an exception after the main thread has finished already, `iwasm` returns with success, instead of returning `1` (i.e. error).

Since `wasm_runtime_get_wasi_exit_code` waits for all threads to finished, the exception check has to performed again after all threads have finished.


product-mini/platforms/posix/main.c
```diff
@@ -745,8 +745,12 @@ main(int argc, char *argv[])
 
 #if WASM_ENABLE_LIBC_WASI != 0
     if (ret == 0) {
-        /* propagate wasi exit code. */
+        /* wait for threads to finish and propagate wasi exit code. */
         ret = wasm_runtime_get_wasi_exit_code(wasm_module_inst);
+        if (wasm_runtime_get_exception(wasm_module_inst)) {
+            /* got an exception in spawned thread */
+            ret = 1;
+        }
```
wenyongh, line 753: Should also apply similar changes in file `product-mini/platforms/windows/main.c`

eloparco: Right, done

wenyongh (commented):
LGTM

Issue #2149, opened 2023-04-24
c_api_func_imports are not passed to new threads


cpetig:
While combining wasi-threads with custom functions defined by the executing program I found that on newly created threads the call fails with "Exception: failed to call unlinked import function (...)".

c_api_func_imports is not properly passed to new threads, copying the following code snippet from lib_pthread_wrapper.c into wasm_runtime.c fixes the problem for me:
[c_api_imports.patch](https://github.com/bytecodealliance/wasm-micro-runtime/files/11315198/c_api_imports.patch)

lum1n0us:
Would you mind sharing the test code or the specific scenario can trigger the issue?

cpetig:
I was able to reproduce the problem with wasm-c-api-imports. if you use sdk-20 and replace --target=wasm32-wasi with --target=wasm32-wasi-threads the old pthread_foo symbols get replaced by the standardized wasi.thread-spawn. 

Running this with example1 yields:
Extracting export...
Calling _start ...
[...]: .../wasm-micro-runtime/core/iwasm/interpreter/wasm_interp_classic.c, line 4310, meet an exception Exception: failed to call unlinked import function (env, log)

cpetig:
I found that my patch is slightly wrong, the size_in_bytes should be sizeof(CApiFuncImport)*import_func_count (no star, thus size of struct not the pointer)

eloparco:
I managed to reproduce. Depending on the run, I get an unaligned exception
```
Initializing...
Loading binary...
> Load .wasm
Compiling module...
[17:04:39:988 - 209814140]: Load type section success.

[17:04:39:118 - 209814140]: Load import section success.

[17:04:39:573 - 209814140]: Load function section success.

[17:04:39:586 - 209814140]: Load table section success.

[17:04:39:591 - 209814140]: Load memory section success.

[17:04:39:595 - 209814140]: Load global section success.

[17:04:39:600 - 209814140]: Load export section success.

[17:04:39:602 - 209814140]: Load start section success.

[17:04:39:608 - 209814140]: Load table segment section success.

[17:04:39:611 - 209814140]: Load datacount section success.

[17:04:39:613 - 209814140]: Load code segment section success.

[17:04:39:620 - 209814140]: Load data segment section success.

[17:04:39:623 - 209814140]: Ignore custom section [name].
[17:04:39:625 - 209814140]: Ignore custom section [producers].
[...]

PR #2170, opened 2023-05-03
Fix import error when using wasm-c-api with WASI threads

eloparco:
Similar to what already in place for `pthreads`: https://github.com/bytecodealliance/wasm-micro-runtime/blob/47fcb8506f7f1ee66fadeef4de8a3481b44633a7/core/iwasm/libraries/lib-pthread/lib_pthread_wrapper.c#L597-L635

#2149

TODO:
- [ ] refactor common code
- [ ] check missing free for `new_c_api_func_imports` in `lib_pthread_wrapper.c`

### DIFF
--- a/core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
+++ b/core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
@@ -83,6 +83,7 @@ thread_spawn_wrapper(wasm_exec_env_t exec_env, uint32 start_arg)
 #if WASM_ENABLE_LIBC_WASI != 0
     WASIContext *wasi_ctx;
 #endif
+    CApiFuncImport **new_c_api_func_imports = NULL;
 
     bh_assert(module);
     bh_assert(module_inst);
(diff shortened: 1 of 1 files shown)


core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
```diff
@@ -102,6 +103,46 @@ thread_spawn_wrapper(wasm_exec_env_t exec_env, uint32 start_arg)
...
+
+            e = (AOTModuleInstanceExtra *)((AOTModuleInstance *)module_inst)->e;
+            c_api_func_imports = e->c_api_func_imports;
+
+            import_func_count = ((AOTModule *)module)->import_func_count;
+        }
+#endif
+
+        if (import_func_count != 0 && c_api_func_imports) {
+            size_in_bytes = sizeof(CApiFuncImport *) * import_func_count;
```
cpetig, line 136: This star is wrong, it needs the size of the struct, not the size of one pointer. Corrected in https://github.com/bytecodealliance/wasm-micro-runtime/compare/main...cpetig:wasm-micro-runtime:main

Also there is more issues with threads which I will describe in a separate issue.

cpetig: (Note: I was the one introducing this error first)

cpetig: Also this error is still present in the code I copied the fix from: https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/libraries/lib-pthread/lib_pthread_wrapper.c#L627

core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c
```diff
@@ -136,6 +177,7 @@ thread_spawn_wrapper(wasm_exec_env_t exec_env, uint32 start_arg)
     deallocate_thread_id(thread_id);
 
 thread_preparation_fail:
+    wasm_runtime_free(*new_c_api_func_imports);
```
cpetig, line 180: If I understand correctly this is already implied by the next two lines - as deinstantiate frees c_api_func_imports in line 2265.

eloparco:
Closing since the same changes are being applied in #2173

Issue #2172, opened 2023-05-03
wasm_func_call uses the wrong exec_env when multithreaded, gives "invalid exec env"


cpetig:
If you call wasm_func_call from a callback which executes on a separate thread you get "invalid exec env" exception.

The check in wasm_runtime.c:2398 fails.

cpetig:
Testcase (adapted from samples/wasm-c-api-imports)
[wamr_test_2172.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/11391563/wamr_test_2172.zip)

cpetig:
Fixed by above MR. I found that the C API is unsuitable for multi-threaded host implementations, and much less ergonomic than the native API.

PR #2173, opened 2023-05-03
Fix three threading+callback issues

cpetig:
Testcase is available at #2172 , fixes also #2149 and an unreported mistake in the original code I copied the fix from.

### DIFF
--- a/core/iwasm/interpreter/wasm_runtime.c
+++ b/core/iwasm/interpreter/wasm_runtime.c
@@ -22,7 +22,7 @@
 #if WASM_ENABLE_FAST_JIT != 0
 #include "../fast-jit/jit_compiler.h"
 #endif
-#if WASM_ENABLE_JIT != 0
+#if WASM_ENABLE_JIT != 0 || WASM_ENABLE_AOT != 0
 #include "../aot/aot_runtime.h"
 #endif
 
--- a/core/iwasm/interpreter/wasm_runtime.c
+++ b/core/iwasm/interpreter/wasm_runtime.c
@@ -2078,7 +2078,10 @@ wasm_instantiate(WASMModule *module, bool is_sub_inst,
             &module_inst->e->functions[module->retain_function];
     }
 
-    if (exec_env_main && ((WASMModuleInstance const*)(exec_env_main->module_inst))->e->c_api_func_imports != NULL) {
+    if (exec_env_main
+        && ((WASMModuleInstance const *)(exec_env_main->module_inst))
+                   ->e->c_api_func_imports
+               != NULL) {
         /* workaround about passing instantiate-linking information */
         {
             CApiFuncImport **new_c_api_func_imports = NULL;
@@ -2087,12 +2090,15 @@ wasm_instantiate(WASMModule *module, bool is_sub_inst,
             uint32 size_in_bytes = 0;
 
 #if WASM_ENABLE_INTERP != 0
-            if (exec_env_main->module_inst->module_type == Wasm_Module_Bytecode) {
+            if (exec_env_main->module_inst->module_type
+                == Wasm_Module_Bytecode) {
                 new_c_api_func_imports = &(
                     ((WASMModuleInstance *)module_inst)->e->c_api_func_imports);
                 c_api_func_imports =
-                    ((WASMModuleInstance *)exec_env_main->module_inst)->e->c_api_func_imports;
-                import_func_count = ((WASMModule *)module)->import_function_count;
+                    ((WASMModuleInstance *)exec_env_main->module_inst)
+                        ->e->c_api_func_imports;
+                import_func_count =
+                    ((WASMModule *)module)->import_function_count;
             }
 #endif
 #if WASM_ENABLE_AOT != 0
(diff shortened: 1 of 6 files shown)


cpetig:
This is a superset of https://github.com/bytecodealliance/wasm-micr4o-runtime/pull/2170 and misses the mentioned future refactor and both cleanup codes (including the one included there).

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -2078,6 +2078,57 @@ wasm_instantiate(WASMModule *module, bool is_sub_inst,
             &module_inst->e->functions[module->retain_function];
     }
 
+    if (exec_env_main
+        && ((WASMModuleInstance const *)(exec_env_main->module_inst))
+                   ->e->c_api_func_imports
+               != NULL) {
+        /* workaround about passing instantiate-linking information */
+        {
```
wenyongh, line 2086: @cpetig It is to fix the issue of `c_api_func_imports are not passed to new threads` reported in #2149 , right? My suggestion is to apply the changes to `lib_wasi_threads_wrapper.c` like #2170 since (1) it is only related to multi-threading, adding code here may increase the footprint, (2) the fixing for AOT mode is invalid, the related code is in `aot_instantiate`: https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/aot/aot_runtime.c#L1089

@eloparco How about we merging the fixing of #2170 to this PR? Just use this PR to fix all these issues? Thanks.

wenyongh: BTW, there may be duplicated code in `lib_wasi_threads_wrapper.c` and `lib_pthread_wrapper.c`, after discussing with @lum1n0us, we suggest to add an API in `core/iwasm/libraries/thread-mgr/thread_manager.{h|c}`, e.g.,
```C
bool
wasm_cluster_dup_c_api_imports(WASMModuleInstance *module_inst_dst,
                               const WASMModuleInstance *module_inst_src);
```
@cpetig @eloparco How about that?

lum1n0us: *lib_pthread_wrapper.c* has a same workaround. It doesn't seem clever to write the same workaround twice. Shall we merge them together and store it somewhere public, maybe *thread_manager*?

eloparco: > How about we merging the fixing of https://github.com/bytecodealliance/wasm-micro-runtime/pull/2170 to this PR? Just use this PR to fix all these issues? Thanks.

Makes sense, I'll close the other PR.

> lib_pthread_wrapper.c has a same workaround. It doesn't seem clever to write the same workaround twice. Shall we merge them together and store it somewhere public, maybe thread_manager?

Yes, that's what I had in mind as well when in my PR I wrote "refactor common code".

eloparco: > Do I understand you correctly that https://github.com/cpetig/wasm-micro-runtime/blob/main/core/iwasm/libraries/thread-mgr/thread_manager.c#L758 is not necessary because this is taken care of in aot_runtime.c ?

I think it's probably the other way around, at the beginning you were putting your changes in `wasm_runtime.c` so you would have needed to change `aot_runtime.c` as well. If the changes are in the pthread and wasi thread files no change in `aot_runtime.c` should be required.

eloparco:
I closed the other PR, let's discuss here instead. For https://github.com/bytecodealliance/wasm-micro-runtime/pull/2170#discussion_r1184676549, yes, no need for the additional `free`, it's done already in https://github.com/bytecodealliance/wasm-micro-runtime/blob/47fcb8506f7f1ee66fadeef4de8a3481b44633a7/core/iwasm/interpreter/wasm_runtime.c#L2265

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -733,6 +733,56 @@ wasm_cluster_create_thread(WASMExecEnv *exec_env,
     return -1;
 }
 
+bool
+wasm_cluster_dup_c_api_imports(wasm_module_inst_t module_inst_dst,
+                               wasm_module_inst_t module_inst_src,
+                               wasm_module_t module)
```
wenyongh, line 739: Had better use `WASMModuleInstanceCommon *` instead of `wasm_module_inst_t`, we normally use the former internally.
And no need to add argument module, we can get it from module_inst->module

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -733,6 +733,56 @@ wasm_cluster_create_thread(WASMExecEnv *exec_env,
...
+{
+    if (((const WASMModuleInstance *)module_inst_src)->e->c_api_func_imports
+        != NULL) {
```
wenyongh, line 742: Should not use assume that the module instance is type of bytecode here, it may be type of AOT, the `module_inst_src->e` may be `WASMModuleInstanceExtra` or `AOTModuleInstanceExtra`, had better:
```C
        CApiFuncImport **new_c_api_func_imports = NULL;
        uint32 import_func_count = 0;
        uint32 size_in_bytes = 0;

#if WASM_ENABLE_INTERP != 0
        if (module_inst_src->module_type == Wasm_Module_Bytecode) {
             WASMModuleInstanceExtra *e_src =
                (WASMModuleInstanceExtra *)((WASMModuleInstance *)module_inst_src)
                    ->e;
              if (e_src->c_api_func_imports != NULL) {
                   ...
              }
         }
#endif
#if WASM_ENABLE_AOT != 0
        if (module_inst_src->module_type == Wasm_Module_AoT) {
             AOTModuleInstanceExtra *e_src =
                (AOTModuleInstanceExtra *)((AOTModuleInstance *)module_inst_src)
                    ->e;
             if (e_src->c_api_func_imports != NULL) {
                 ...
             }
        }
#endif
```

wenyongh (commented):
LGTM

lum1n0us:
LGTM

PR #2176, opened 2023-05-04
Use a manual flag to disable clock_nanosleep on platforms where it's not implemented

g0djan:
For some platforms WAMR getting compiled with CONFIG_HAS_CLOCK_NANOSLEEP=1 when clock_nanosleep is not present at the platform.

Use [try_compile](https://cmake.org/cmake/help/latest/command/try_compile.html) to make sure clock_nanosleep is actually present on the platform

Tested that it correctly set flag on linux and on an arm platform without clock_nanosleep

### DIFF
--- a/build-scripts/runtime_lib.cmake
+++ b/build-scripts/runtime_lib.cmake
@@ -13,12 +13,32 @@ file(WRITE "${PROJECT_BINARY_DIR}/try_compile_clock_nanosleep.c"
 	int main () { return 0; }
 ")
 
-try_compile(WAMR_HAVE_CLOCK_NANOSLEEP
+try_compile(WAMR_HAS_CLOCK_NANOSLEEP
   ${PROJECT_BINARY_DIR}/try_compile_clock_nanosleep
   ${PROJECT_BINARY_DIR}/try_compile_clock_nanosleep.c)
 
-if (WAMR_HAVE_CLOCK_NANOSLEEP)
+if (WAMR_HAS_CLOCK_NANOSLEEP)
     add_definitions(-DCLOCK_NANOSLEEP_COMPILES)
+endif()
+
+file(WRITE "${PROJECT_BINARY_DIR}/try_compile_nanosleep.c"
+"	#include <time.h>
+    void mysleep_ms(int milisec)
+    {
+        struct timespec res;
+        res.tv_sec = milisec/1000;
+        res.tv_nsec = (milisec*1000000) % 1000000000;
+        nanosleep(&res, NULL);
+    }
+	int main () { return 0; }
+")
+
+try_compile(WAMR_HAS_NANOSLEEP
+  ${PROJECT_BINARY_DIR}/try_compile_nanosleep
+  ${PROJECT_BINARY_DIR}/try_compile_nanosleep.c)
+
+if (WAMR_HAS_NANOSLEEP)
+    add_definitions(-DNANOSLEEP_COMPILES)
 endif()
 
 if (NOT DEFINED WAMR_ROOT_DIR)
--- a/build-scripts/runtime_lib.cmake
+++ b/build-scripts/runtime_lib.cmake
@@ -1,6 +1,26 @@
 # Copyright (C) 2019 Intel Corporation. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+file(WRITE "${PROJECT_BINARY_DIR}/try_compile_clock_nanosleep.c"
+"	#include <time.h>
+    void mysleep_ms(int milisec)
+    {
+        struct timespec res;
+        res.tv_sec = milisec/1000;
+        res.tv_nsec = (milisec*1000000) % 1000000000;
+        clock_nanosleep(CLOCK_MONOTONIC, 0, &res, NULL);
+    }
+	int main () { return 0; }
+")
+
+try_compile(WAMR_HAVE_CLOCK_NANOSLEEP
+  ${PROJECT_BINARY_DIR}/try_compile_clock_nanosleep
+  ${PROJECT_BINARY_DIR}/try_compile_clock_nanosleep.c)
+
+if (WAMR_HAVE_CLOCK_NANOSLEEP)
+    add_definitions(-DCLOCK_NANOSLEEP_COMPILES)
+endif()
+
 if (NOT DEFINED WAMR_ROOT_DIR)
     set (WAMR_ROOT_DIR ${CMAKE_CURRENT_LIST_DIR}/../)
 endif ()
--- a/build-scripts/runtime_lib.cmake
+++ b/build-scripts/runtime_lib.cmake
@@ -1,46 +1,6 @@
 # Copyright (C) 2019 Intel Corporation. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-file(WRITE "${PROJECT_BINARY_DIR}/try_compile_clock_nanosleep.c"
-"   #include <time.h>
-    void mysleep_ms(int milisec)
-    {
-        struct timespec res;
-        res.tv_sec = milisec/1000;
-        res.tv_nsec = (milisec*1000000) % 1000000000;
-        clock_nanosleep(CLOCK_MONOTONIC, 0, &res, NULL);
-    }
-    int main () { return 0; }
-")
-
-try_compile(WAMR_HAS_CLOCK_NANOSLEEP
-  ${PROJECT_BINARY_DIR}/try_compile_clock_nanosleep
-  ${PROJECT_BINARY_DIR}/try_compile_clock_nanosleep.c)
-
-if (WAMR_HAS_CLOCK_NANOSLEEP)
-    add_definitions(-DCLOCK_NANOSLEEP_COMPILES)
-endif()
-
-file(WRITE "${PROJECT_BINARY_DIR}/try_compile_nanosleep.c"
-"   #include <time.h>
-    void mysleep_ms(int milisec)
-    {
-        struct timespec res;
-        res.tv_sec = milisec/1000;
-        res.tv_nsec = (milisec*1000000) % 1000000000;
-        nanosleep(&res, NULL);
-    }
-    int main () { return 0; }
-")
-
-try_compile(WAMR_HAS_NANOSLEEP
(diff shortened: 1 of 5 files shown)


build-scripts/runtime_lib.cmake
```diff
@@ -1,6 +1,46 @@
...
+  ${PROJECT_BINARY_DIR}/try_compile_clock_nanosleep
+  ${PROJECT_BINARY_DIR}/try_compile_clock_nanosleep.c)
+
+if (WAMR_HAS_CLOCK_NANOSLEEP)
+    add_definitions(-DCLOCK_NANOSLEEP_COMPILES)
+endif()
+
+file(WRITE "${PROJECT_BINARY_DIR}/try_compile_nanosleep.c"
+"   #include <time.h>
+    void mysleep_ms(int milisec)
```
hritikgupta, line 26: [nit] millisec

build-scripts/runtime_lib.cmake
```diff
@@ -1,6 +1,46 @@
...
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+file(WRITE "${PROJECT_BINARY_DIR}/try_compile_clock_nanosleep.c"
+"   #include <time.h>
+    void mysleep_ms(int milisec)
+    {
+        struct timespec res;
+        res.tv_sec = milisec/1000;
+        res.tv_nsec = (milisec*1000000) % 1000000000;
+        clock_nanosleep(CLOCK_MONOTONIC, 0, &res, NULL);
```
hritikgupta, line 11: should add error handling ?

loganek: I don't think this code is intended to run, just compile.

build-scripts/runtime_lib.cmake
```diff
@@ -1,6 +1,46 @@
 # Copyright (C) 2019 Intel Corporation. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+file(WRITE "${PROJECT_BINARY_DIR}/try_compile_clock_nanosleep.c"
+"   #include <time.h>
+    void mysleep_ms(int milisec)
```
loganek, line 6: Is it possible that the code compiles but it fails.at runtime? According to the doc: https://man7.org/linux/man-pages/man2/clock_nanosleep.2.html it may return enotsup.

build-scripts/runtime_lib.cmake
```diff
@@ -1,6 +1,46 @@
 # Copyright (C) 2019 Intel Corporation. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+file(WRITE "${PROJECT_BINARY_DIR}/try_compile_clock_nanosleep.c"
```
wenyongh, line 4: Compiling the files here may be a little heavy, it will be performed even when libc-wasi isn't enabled. How about moving the changes to `core/iwasm/libraries/libc-wasi/libc_wasi.cmake`?

build-scripts/runtime_lib.cmake
g0djan: According to this man indeed it is possible. However this change only makes the check stricter even it can't catch this tricky case but it wouldn't be caught before either.

build-scripts/runtime_lib.cmake
g0djan: Fixed

build-scripts/runtime_lib.cmake
g0djan: Sure, moved it

wenyongh (commented):
LGTM

core/iwasm/libraries/libc-wasi/libc_wasi.cmake
```diff
@@ -11,3 +11,23 @@ include_directories(${LIBC_WASI_DIR}/sandboxed-system-primitives/include
...
+    {
+        struct timespec res;
+        res.tv_sec = millisec/1000;
+        res.tv_nsec = (millisec*1000000) % 1000000000;
+        clock_nanosleep(CLOCK_MONOTONIC, 0, &res, NULL);
+    }
+    int main () { return 0; }
+")
+
+try_compile(WAMR_HAS_CLOCK_NANOSLEEP
```
wenyongh, line 27: Hi, it seems that cmake already has `check_function_exists` function to do the same work, no need to use `try_compile` (which generates files under build folder and is not so friendly): https://cmake.org/cmake/help/v3.14/module/CheckFunctionExists.html

And by using `check_function_exists`, we only need to modify `libc_wasi.cmake` and `sandboxed-system-primitives/src/ssp_config.h`, no need to modify other two files. It works on building iwasm for linux and linux-sgx platforms. Here is the patch I tested, could you test whether it works for you:
```diff
diff --git a/core/iwasm/libraries/libc-wasi/libc_wasi.cmake b/core/iwasm/libraries/libc-wasi/libc_wasi.cmake
index d72c42a0..8549ae98 100644
--- a/core/iwasm/libraries/libc-wasi/libc_wasi.cmake
+++ b/core/iwasm/libraries/libc-wasi/libc_wasi.cmake
@@ -11,3 +11,10 @@ include_directories(${LIBC_WASI_DIR}/sandboxed-system-primitives/include
 file (GLOB_RECURSE source_all ${LIBC_WASI_DIR}/*.c )

 set (LIBC_WASI_SOURCE ${source_all})
+
+include (CheckFunctionExists)
+check_function_exists(clock_nanosleep WAMR_HAS_CLOCK_NANOSLEEP)
+
+if (WAMR_HAS_CLOCK_NANOSLEEP)
+  add_definitions(-DCLOCK_NANOSLEEP_COMPILES)
+endif()
diff --git a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/ssp_config.h b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/ssp_config.h
index d5babd02..094d52e4 100644
--- a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/ssp_config.h
+++ b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/ssp_config.h
@@ -41,7 +41,7 @@
 #endif

 #if !defined(__APPLE__) && !defined(__FreeBSD__) && !defined(__EMSCRIPTEN__) \
-    && !defined(ESP_PLATFORM)
+    && !defined(ESP_PLATFORM) && defined(CLOCK_NANOSLEEP_COMPILES)
 #define CONFIG_HAS_CLOCK_NANOSLEEP 1
 #else
 #define CONFIG_HAS_CLOCK_NANOSLEEP 0
```

g0djan: Seems better, I checked on linux and it seems fine. I'll check it out on a device without clock_nanosleep tommorrow in the office

yamt:
> For some platforms WAMR getting compiled with CONFIG_HAS_CLOCK_NANOSLEEP=1 when clock_nanosleep is not present at the platform.
> 
> Use [try_compile](https://cmake.org/cmake/help/latest/command/try_compile.html) to make sure clock_nanosleep is actually present on the platform
> 
> Tested that it correctly set flag on linux and on an arm platform without clock_nanosleep

which platforms specifically?

i'm a bit hesitate to have try-compile based solutions.

core/iwasm/libraries/libc-wasi/libc_wasi.cmake
g0djan: I checked and it worked this way. BUT I found that I was wrong in the first place, because clock_nanosleep was available on the platform but it used netbsd https://man.netbsd.org/clock_nanosleep.2 and even function was defined it returned ENOSYS

So it means that try_compile and checkFunctionExists detected absense of the function wrong, so I decided to go with a flag because its simplier and more explicit solution even it's less elegant

g0djan:
I test it on Roku 4200X, it's ARM based device and it uses NetBSD apparently https://man.netbsd.org/clock_nanosleep.2

And I also don't think that using try_compile is a good idea anymore

build-scripts/runtime_lib.cmake
g0djan: You were right, it's what was happening all along

core/iwasm/libraries/libc-wasi/libc_wasi.cmake
g0djan: But now I don't see any logs in CI failing jobs :thinking:

yamt: > I checked and it worked this way. BUT I found that I was wrong in the first place, because clock_nanosleep was available on the platform but it used netbsd https://man.netbsd.org/clock_nanosleep.2 and even function was defined it returned ENOSYS

while i have no idea about Roku 4200X,
i'm sure NetBSD has had clock_nanosleep for a decade.

yamt: > But now I don't see any logs in CI failing jobs 🤔

github is broken today.
https://www.githubstatus.com/

g0djan: > i'm sure NetBSD has had clock_nanosleep for a decade.
And it does have it, it compiles. And that's why idea of using `try_compile` actually wasn't solving the problem I met.
clock_nanosleep on this platform returns ENOSYS, which is correct behavior according to the NetBSD doc, but it means that it doesn't work.

Now I just switched to a simplier solution

yamt: > > i'm sure NetBSD has had clock_nanosleep for a decade.
> 
> And it does have it, it compiles. And that's why idea of using `try_compile` actually wasn't solving the problem I met. `clock_nanosleep` on this platform returns ENOSYS, which is correct behavior according to the NetBSD doc, but it means that implementation on this platform is stub and doesn't do anything.

i meant NetBSD has had working clock_nanosleep.
i guess it's just listing theoretical values.
it isn't even NetBSD-specific. cf. https://pubs.opengroup.org/onlinepubs/007908799/xsh/nanosleep.html

> 
> Now I just switched to a simplier solution

ok.

g0djan:
@wenyongh anything to address or good to merge?

wenyongh:
> @wenyongh anything to address or good to merge?

Merged.

g0djan:
Thank you!

PR #2190, opened 2023-05-09
Add --enable-gc option to WAMRC

no1wudi:
The runtime instance memory layout changed with GC enabled. With this patch GC enabled for WAMRC, but keep compatible for iwasm wether it enabled GC or not for non-GC wasm module.

It may waste some memory for iwasm without GC support since the GC relative fileds are always here, let's do optimization until   
AOT fully support GC.

### DIFF
--- a/wamr-compiler/CMakeLists.txt
+++ b/wamr-compiler/CMakeLists.txt
@@ -46,6 +46,12 @@ add_definitions(-DWASM_ENABLE_PERF_PROFILING=1)
 add_definitions(-DWASM_ENABLE_LOAD_CUSTOM_SECTION=1)
 add_definitions(-DWASM_ENABLE_LIB_WASI_THREADS=1)
 
+if (WAMR_BUILD_GC_BINARYEN EQUAL 1)
+  add_definitions (-DWASM_ENABLE_GC_BINARYEN=1)
+else ()
+  add_definitions (-DWASM_ENABLE_GC=1)
+endif ()
+
 if (WAMR_BUILD_LLVM_LEGACY_PM EQUAL 1)
   add_definitions(-DWASM_ENABLE_LLVM_LEGACY_PM=1)
 endif()
@@ -213,6 +219,7 @@ endif()
 include (${IWASM_DIR}/libraries/lib-pthread/lib_pthread.cmake)
 include (${IWASM_DIR}/libraries/lib-wasi-threads/lib_wasi_threads.cmake)
 include (${IWASM_DIR}/common/iwasm_common.cmake)
+include (${IWASM_DIR}/common/gc/iwasm_gc.cmake)
 include (${IWASM_DIR}/interpreter/iwasm_interp.cmake)
 include (${IWASM_DIR}/aot/iwasm_aot.cmake)
 include (${IWASM_DIR}/compilation/iwasm_compl.cmake)
@@ -269,7 +276,8 @@ add_library (vmlib
              ${LIB_WASI_THREADS_SOURCE}
              ${IWASM_COMMON_SOURCE}
              ${IWASM_INTERP_SOURCE}
-             ${IWASM_AOT_SOURCE})
+             ${IWASM_AOT_SOURCE}
+             ${IWASM_GC_SOURCE})
 
 add_library (aotclib ${IWASM_COMPL_SOURCE})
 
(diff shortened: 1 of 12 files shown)


wamr-compiler/CMakeLists.txt
```diff
@@ -45,6 +45,7 @@ add_definitions(-DWASM_ENABLE_DUMP_CALL_STACK=1)
 add_definitions(-DWASM_ENABLE_PERF_PROFILING=1)
 add_definitions(-DWASM_ENABLE_LOAD_CUSTOM_SECTION=1)
 add_definitions(-DWASM_ENABLE_LIB_WASI_THREADS=1)
+add_definitions(-DWASM_ENABLE_GC=1)
```
wenyongh, line 48: According to the design, we want to add control to enable gc or enable gc binaryen, the change should be like:
```cmake
if (WAMR_BUILD_GC_BINARYEN EQUAL 1)
  add_definitions (-DWASM_ENABLE_GC_BINARYEN=1)
else ()
  add_definitions (-DWASM_ENABLE_GC=1)
endif ()
```
Then we can build wamrc for gc/gc_binaryen version with: `cmake -DWAMR_BUILD_GC_BINARYEN=0/1`

wamr-compiler/main.c
```diff
@@ -153,6 +154,7 @@ main(int argc, char *argv[])
     option.enable_aux_stack_check = true;
     option.enable_bulk_memory = true;
     option.enable_ref_types = true;
+    option.enable_gc = false;
```
wenyongh, line 157: Should also add related code to control the feature for LLVM JIT:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_loader.c#L3053
After it, add:
```C
#if WASM_ENABLE_GC != 0 || WASM_ENABLE_GC_BINARYEN != 0
    option.enable_gc = true;
    option.enable_ref_types = false;
#endif
```

wamr-compiler/main.c
```diff
@@ -266,6 +268,9 @@ main(int argc, char *argv[])
         else if (!strcmp(argv[0], "--enable-indirect-mode")) {
             option.is_indirect_mode = true;
         }
+        else if (!strcmp(argv[0], "--enable-gc")) {
+            option.enable_gc = true;
```
wenyongh, line 272: Should disable ref_types when gc is enabled?
```C
option.enable_ref_types = false;
```

And when handling `else if (!strcmp(argv[0], "--enable-ref-types")) {`
```C
option.enable_ref_types = true;
option.enable_gc = false;
```

no1wudi: OK, this case handled in line 333 by:
```c
    if (option.enable_gc) {
        option.enable_ref_types = false;
    }
```

wamr-compiler/main.c
```diff
@@ -57,12 +57,13 @@ print_help()
     printf("  --disable-simd            Disable the post-MVP 128-bit SIMD feature:\n");
     printf("                              currently 128-bit SIMD is supported for x86-64 and aarch64 targets,\n");
     printf("                              and by default it is enabled in them and disabled in other targets\n");
-    printf("  --disable-ref-types       Disable the MVP reference types feature\n");
+    printf("  --disable-ref-types       Disable the MVP reference types feature, it will be force disabled if GC enabled\n");
```
wenyongh, line 60: How about `it will be disabled forcibly`?

wenyongh: `be force disabled` should be a grammar mistake

wamr-compiler/main.c
```diff
@@ -57,12 +57,13 @@ print_help()
     printf("  --disable-simd            Disable the post-MVP 128-bit SIMD feature:\n");
     printf("                              currently 128-bit SIMD is supported for x86-64 and aarch64 targets,\n");
     printf("                              and by default it is enabled in them and disabled in other targets\n");
-    printf("  --disable-ref-types       Disable the MVP reference types feature\n");
+    printf("  --disable-ref-types       Disable the MVP reference types feature, it will be disabled forcibly if GC enabled\n");
```
wenyongh, line 60:
```C
printf(" --disable-ref-types Disable the MVP reference types feature, it will be disabled forcibly if\n");
printf("                     GC is enabled\n");

wenyongh (approved):
LGTM

PR #2197, opened 2023-05-09
Enable `wamrc --enable-segue=<flags>` to provide detailed options

wenyongh:


yamt:
what does "segue" stands for?

wenyongh:
> what does "segue" stands for?

Hi, the segue is an optimization technology to use the x86 segment register to improve performance, currently it is under testing, we may make it as an option in the feature.

PR #2202, opened 2023-05-10
.github/workflows/build_wamr_lldb.yml: use a bit more specific cache key

yamt:
the current key looks like "x86_64-ubuntu-20.04-lldb_build". it's a way too generic and can even hit the cache from the previous releases.

this commit makes it contain the WAMR version number.


lum1n0us:
Previous key without `ver_num` tends to reuse llvm-libraries between different WAMR releases especially when wamr-lldb always comes from a specific commit *1f27fe6128769f00197925c3b8f6abb9d0e5cd2e*.

PR #2204, opened 2023-05-11
fix GC issues

xujuntwt95329:
- don't mark non-ref array elements
- trace ref returned from native function

### DIFF
--- a/core/iwasm/common/gc/gc_object.c
+++ b/core/iwasm/common/gc/gc_object.c
@@ -684,7 +684,7 @@ wasm_object_get_ref_list(WASMObjectRef obj, bool *p_is_compact_mode,
         static uint16 externref_obj_ref_list[] = { (uint16)offsetof(
             WASMExternrefObject, internal_obj) };
         *p_is_compact_mode = false;
-        *p_ref_num = 0;
+        *p_ref_num = 1;
         *p_ref_list = externref_obj_ref_list;
         return true;
     }
@@ -712,9 +712,18 @@ wasm_object_get_ref_list(WASMObjectRef obj, bool *p_is_compact_mode,
     }
     else if (rtt_type->defined_type->type_flag == WASM_TYPE_ARRAY) {
         /* array object */
-        *p_is_compact_mode = true;
-        *p_ref_num = (uint16)wasm_array_obj_length((WASMArrayObjectRef)obj);
-        *p_ref_start_offset = (uint16)offsetof(WASMArrayObject, elem_data);
+        WASMArrayType *type = (WASMArrayType *)rtt_type->defined_type;
+        if (wasm_is_type_reftype(type->elem_type)) {
+            *p_is_compact_mode = true;
+            *p_ref_num = (uint16)wasm_array_obj_length((WASMArrayObjectRef)obj);
+            *p_ref_start_offset = (uint16)offsetof(WASMArrayObject, elem_data);
+        }
+        else {
+            *p_is_compact_mode = false;
+            *p_ref_num = 0;
+            *p_ref_list = NULL;
+        }
+
         return true;
     }
     else {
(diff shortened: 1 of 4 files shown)


core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -4903,6 +4903,20 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
             {
                 wasm_interp_call_func_native(module, exec_env, cur_func,
                                              prev_frame);
+#if WASM_ENABLE_GC != 0
```
wenyongh, line 4906: There may be several places calling wasm_interp_call_func_native, is it better to apply the changes in it? And had better set frame_ref only when the exception isn't thrown?

core/shared/mem-alloc/ems/ems_gc.c
```diff
@@ -368,8 +368,8 @@ reclaim_instance_heap(gc_heap_t *heap)
 
             if (is_compact_mode) {
                 for (j = 0; j < (int)ref_num; j++) {
-                    offset = ref_start_offset + j * 4;
-                    bh_assert(offset + 4 < size);
+                    offset = ref_start_offset + j * sizeof(void *);
+                    bh_assert(offset + sizeof(void *) < size);
```
wenyongh, line 372: Same for L387?

xujuntwt95329: done

core/iwasm/interpreter/wasm_interp_classic.c
xujuntwt95329: Yes, updated.
But seems there is some issue with fast interpreter, I'm not sure if I calculate the right frame_ref address for fast interpreter

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -1093,6 +1094,20 @@ wasm_interp_call_func_native(WASMModuleInstance *module_inst,
     if (!ret)
         return;
 
+#if WASM_ENABLE_GC != 0
+    func_type = cur_func->u.func_import->func_type;
+    if (func_type->result_count
+        && wasm_is_type_reftype(func_type->types[cur_func->param_count])) {
+        frame_ref = (uint8 *)prev_frame->csp_boundary
+                    + (unsigned)(prev_frame->sp - prev_frame->lp);
```
wenyongh, line 1102: Had better `(unsigned)(uintptr_t)(prev_frame->sp - prev_frame->lp)`, or compilation warning may be reported?

core/iwasm/common/gc/gc_object.c
```diff
@@ -712,9 +712,18 @@ wasm_object_get_ref_list(WASMObjectRef obj, bool *p_is_compact_mode,
     }
     else if (rtt_type->defined_type->type_flag == WASM_TYPE_ARRAY) {
         /* array object */
-        *p_is_compact_mode = true;
-        *p_ref_num = (uint16)wasm_array_obj_length((WASMArrayObjectRef)obj);
-        *p_ref_start_offset = (uint16)offsetof(WASMArrayObject, elem_data);
+        WASMArrayType *type = (WASMArrayType *)rtt_type->defined_type;
+        if (wasm_is_type_reftype(type->elem_type)) {
+            *p_is_compact_mode = true;
+            *p_ref_num = (uint16)wasm_array_obj_length((WASMArrayObjectRef)obj);
```
wenyongh, line 718: A concern for the current design is that uint16 only supports maximum 65535 array elements, will it be too small for some scenarios? Should we extend it?

core/iwasm/interpreter/wasm_interp_classic.c
xujuntwt95329: done

core/iwasm/interpreter/wasm_interp_classic.c
wenyongh: Seems right, `frame->frame_ref` in fast interpreter isn't changed after initialized, if the ret_offset is the offset to the `frame->lp`, then it should be right.

core/iwasm/common/gc/gc_object.c
xujuntwt95329: Maybe enhance in another patch?
This patch still have some issue in fast interpreter, in `wasm_interp_traverse_gc_rootset`, some non-gc managed objects are added into gc root, seems the frame_ref not work well

wenyongh: Yes, let's merge this PR first.

xujuntwt95329: updated in #2206

PR #2205, opened 2023-05-12
[dpgo] compilation options for dynamic PGO

lum1n0us:
only enable dynamic PGO in multi-tire JIT mode. otherwise, raise a warning and disable the compilation flag.

### DIFF
--- a/build-scripts/config_common.cmake
+++ b/build-scripts/config_common.cmake
@@ -127,6 +127,14 @@ else ()
   unset (LLVM_AVAILABLE_LIBS)
 endif ()
 
+# Only enable dynamic-PGO(WAMR_BUILD_DYNAMIC_PGO) in multi-tier JIT mode if set it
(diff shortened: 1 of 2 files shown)


build-scripts/config_common.cmake
```diff
@@ -166,6 +174,10 @@ endif ()
 if (WAMR_BUILD_FAST_JIT EQUAL 1 AND WAMR_BUILD_JIT EQUAL 1
     AND WAMR_BUILD_LAZY_JIT EQUAL 1)
   message ("     Multi-tier JIT enabled")
+  if (WAMR_BUILD_DYNAMIC_PGO EQUAL 1)
+    add_definitions( -DWASM_ENABLE_DYNAMIC_PGO=1)
```
wenyongh, line 178: Also add lines in core/config.h
```C
#ifndef WASM_ENABLE_DYNAMIC_PGO
#define WASM_ENABLE_DYNAMIC_PGO 0
#endif
```

PR #2215, opened 2023-05-15
add packed type for type equal

zhenweijin:


core/iwasm/common/gc/gc_common.c
```diff
@@ -354,6 +354,11 @@ wasm_ref_type_equal(const wasm_ref_type_t *ref_type1,
     type1 = ref_type1_norm.value_type;
     type2 = ref_type2_norm.value_type;
 
+    if ((type1 == VALUE_TYPE_I8 && type2 == VALUE_TYPE_I8)
```
xujuntwt95329, line 357: This will not work, if value type is set to `VALUE_TYPE_I8` or `I16`, seems in L348 the `wasm_ref_type_normalize` will fail and this function returned directly?

I think we should modify the implementation of `wasm_ref_type_normalize` to process `VALUE_TYPE_I8` and `VALUE_TYPE_I16`?

zhenweijin: thanks for advice, seems in `wasm_ref_type_normalize`, on condition doesn't contain VALUE_TYPE_I16/I8, I'll fix it

PR #2219, opened 2023-05-16
Add Rust Formatters to Debugger (Vector, Map etc.)

MrSarius:
This PR adds LLDB formatters so that variables are human-readable when debugging Rust code in VS Code. This includes Tuple, Slice, String, Vector, Map, Enum etc.

### Python in LLDB
For performing more complex calculations on debugger variables (as required in this PR), LLDB uses Python. This requires building LLDB with Python support, but there are some hurdles to overcome.

LLDB links to the Python C ABI, which is not compatible between minor versions of Python. Although a [stable Python ABI](https://docs.python.org/3/c-api/stable.html) exists, LLDB does not stick to this subset. Means we have to ensure that the exact version of Python that we are using in our pipeline is available on the user's machine. Even if the version is correct, LLDB cannot link against Python from [different distributions](https://lldb.llvm.org/resources/caveats.html#id1) (e.g. homebrew and python.org).

Therefore I decided to distribute a [stand-alone Python](https://github.com/indygreg/python-build-standalone) version with LLDB. This solution enables high portability. This way, Ubuntu 20.04 and 22.04 can for example still be supported with the same build since glibc is statically linked in the Python build, also making it easier to support more operating systems in the future.

This approach also seems to be the GOTO solution in comparable cases (e.g. XCode also uses LLDB and distributes it with a own Python build with each version).

### Swig3 MacOs
Inside the MacOs pipeline I had to downgrade from swig4 to swig3 because swig4 [causes the LLDB build to fail](https://bugzilla.redhat.com/show_bug.cgi?id=2128646). Perhaps this can be removed at a later date.

### Tests
This was tested on MacOs13-aarch64 and Ubuntu22.04-x86.

I've also written unit and integration tests for my implementation, however for anything related to the VS Code extension there haven't been any tests at all yet, I'll open another PR for that.

### Known Issues
**Enum types are not displayed correctly.** For example, for optionals LLDB resolves this type: `core::option::Option<>` instead of `core::option::Option<&str>...` (see screenshot). This must be an issue in the patched LLDB build, the formatters themselves work fine with native LLDB. 

I'm currently investigating this, but it should not be related to the changes from this PR.

github2

### Test changes

<details>
<summary>Click me</summary>

Build iwasm with source debugging feature.
```
cd ${WAMR_ROOT}/product-mini/platforms/${YOUR_PLATFORM}
mkdir build && cd build
cmake .. -DWAMR_BUILD_DEBUG_INTERP=1
make
```

> Note: On MacOS M1 environment, pass the additional -DWAMR_DISABLE_HW_BOUND_CHECK=1 cmake configuration.

Compile your rust file with debug symbols and no optimisation to avoid inlining.
```
rustc --target wasm32-wasi ./test.rs -g -C opt-level=0
```

Download the LLDB build (with python) for your platform from [my pipeline](https://github.com/MrSarius/wasm-micro-runtime/releases/tag/WAMR-1.2.2) and move it to the extension's resource folder.
```
cd ${WAMR_ROOT}/test-tools/wamr-ide/VSCode-Extension/resource/debug/${LLDB_FOR_YOUR_PLATFORM}
wget ${LLDB_FOR_YOUR_PLATFORM}.tar.gz
tar -xvf ${LLDB_FOR_YOUR_PLATFORM}.tar.gz
```

Execute iwasm with debug engine enabled.
```
product-mini/platforms/darwin/build/iwasm -g=127.0.0.1:1234 test.wasm
```

Open vs code in `${WAMR_ROOT}/test-tools/wamr-ide/VSCode-Extension`, go to the debug tab and launch `Launch Extension`. In this new window, the extension is now enabled. 

Set your breakpoints in vs code and launch the debugger with this `launch.json`
```
{
    "configurations": [
        {
            "type": "wamr-debug",
            "request": "attach",
            "name": "Attach Debugger",
            "stopOnEntry": true,
            "attachCommands": [
                "process connect -p wasm connect://127.0.0.1:1234"
            ]
        }
    ]
}
```
</details>

### DIFF
--- a/.github/workflows/build_wamr_lldb.yml
+++ b/.github/workflows/build_wamr_lldb.yml
@@ -73,7 +73,7 @@ jobs:
           brew link --overwrite swig@3
           sudo rm -rf /Library/Developer/CommandLineTools
 
-      - name: intsall utils ubuntu
+      - name: install utils ubuntu
         if: steps.lldb_build_cache.outputs.cache-hit != 'true' && contains(inputs.runner, 'ubuntu')
         run: sudo apt update && sudo apt-get install -y lld ninja-build
 
@@ -102,7 +102,7 @@ jobs:
           tar -xvf python.tar.gz
         working-directory: core/deps
 
-      - name: get stand-alone pyhton macos
+      - name: get stand-alone python macos
         if: steps.lldb_build_cache.outputs.cache-hit != 'true' && contains(inputs.runner, 'macos')
         run: |
           wget ${{ env.PYTHON_MACOS_STANDALONE_BUILD }} -O python.tar.gz
--- a/.github/workflows/build_wamr_lldb.yml
+++ b/.github/workflows/build_wamr_lldb.yml
@@ -23,6 +23,10 @@ on:
         type: string
         required: true
 
+# If you want to update the version of Python distributed with LLDB, update the 'get stand-alone Python' tasks and then change this variable accordingly.
+env:
+  PYTHON_VERSION: '3.10'
+
 jobs:
   try_reuse:
     uses: ./.github/workflows/reuse_latest_release_binaries.yml
@@ -63,7 +67,9 @@ jobs:
       - name: install utils macos
         if: steps.lldb_build_cache.outputs.cache-hit != 'true' && contains(inputs.runner, 'macos')
         run: |
-          brew install swig cmake ninja libedit
+          brew remove swig
+          brew install swig@3 cmake ninja libedit
+          brew link --overwrite swig@3
           sudo rm -rf /Library/Developer/CommandLineTools
 
       - name: intsall utils ubuntu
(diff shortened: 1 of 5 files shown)


.github/workflows/build_wamr_lldb.yml
```diff
@@ -63,7 +63,8 @@ jobs:
       - name: install utils macos
         if: steps.lldb_build_cache.outputs.cache-hit != 'true' && contains(inputs.runner, 'macos')
         run: |
-          brew install swig cmake ninja libedit
+          brew remove swig
```
MrSarius, line 66: Is there a better way to remove a pre-installed dependency from a GitHub image?

.github/workflows/build_wamr_lldb.yml
```diff
@@ -111,7 +112,7 @@ jobs:
             -DLLVM_INCLUDE_TESTS:BOOL=OFF \
             -DLLVM_ENABLE_BINDINGS:BOOL=OFF \
             -DLLVM_ENABLE_LIBXML2:BOOL=ON \
-            -DLLDB_ENABLE_PYTHON:BOOL=OFF \
+            -DLLDB_ENABLE_PYTHON:BOOL=ON \
```
MrSarius, line 115: Do we need to include a full version of python together with the LLDB download or what what is the common way of dealing with this dependency when shipping LLDB binaries?

.github/workflows/build_wamr_lldb.yml
TianlongLiang: As far as I know, there isn't a better way(. I think removing the existing version and reinstalling the version needed is a solid way.

.github/workflows/build_wamr_lldb.yml
MrSarius: @wenyongh @lum1n0us 
Any ideas on how to create a distributable version of LLDB with python scripting support?

lum1n0us:
https://lldb.llvm.org/resources/build.html

IIUC, just need to enable `LLDB_ENABLE_PYTHON`

MrSarius:
If I do so and execute the resulting binary on a different machine. I get this:
```
./lldb
dyld[17796]: Library not loaded: /opt/homebrew/opt/python@3.11/Frameworks/Python.framework/Versions/3.11/Python
...
```
I think I have to bundle Python with LLDB, so that version and the path are the same.

lum1n0us:
How about using the default configuration? 
`$ cmake -DCMAKE_BUILD_TYPE=Release -DLLVM_ENABLE_PROJECTS="clang;lldb" ../llvm`

test-tools/wamr-ide/VSCode-Extension/formatters/rust.py
```diff
@@ -0,0 +1,746 @@
+'''
+Copyright (c) 2016 Vadim Chugunov
```
wenyongh, line 2: Should also add another line:
`SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception`

.github/workflows/build_wamr_lldb.yml
```diff
@@ -88,31 +94,47 @@ jobs:
...
-            -DCMAKE_INSTALL_PREFIX=../wamr-lldb \
+            -G Ninja -DCMAKE_INSTALL_PREFIX=../wamr-lldb \
             -DCMAKE_BUILD_TYPE:STRING="Release" \
             -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
             -DLLVM_ENABLE_PROJECTS="clang;lldb" \
             -DLLVM_TARGETS_TO_BUILD:STRING="X86;WebAssembly" \
-            -DLLVM_BUILD_BENCHMARKS:BOOL=OFF \
-            -DLLVM_BUILD_DOCS:BOOL=OFF \
-            -DLLVM_BUILD_EXAMPLES:BOOL=OFF  \
+            -DLLVM_BUILD_BENCHMARKS:BOOL=OFF -DLLVM_BUILD_DOCS:BOOL=OFF \
```
wenyongh, line 122: Had better put `-DLLVM_BUILD_DOCS:BOOL=OFF` on another line

.github/workflows/build_wamr_lldb.yml
```diff
@@ -122,18 +144,16 @@ jobs:
           echo "start to build lldb..."
           mkdir -p wamr-lldb
           cmake -S ./llvm -B build \
-            -G Ninja \
-            -DCMAKE_INSTALL_PREFIX=../wamr-lldb \
+            -G Ninja -DCMAKE_INSTALL_PREFIX=../wamr-lldb \
             -DCMAKE_BUILD_TYPE:STRING="Release" \
-            -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
-            -DLLVM_ENABLE_PROJECTS="clang;lldb" \
+            -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DLLVM_ENABLE_PROJECTS="clang;lldb" \
```
wenyongh, line 149: Had better put `-DLLVM_ENABLE_PROJECTS="clang;lldb"` on another line

test-tools/wamr-ide/VSCode-Extension/formatters/rust.py
MrSarius: done

.github/workflows/build_wamr_lldb.yml
MrSarius: done

.github/workflows/build_wamr_lldb.yml
MrSarius: done

TianlongLiang:
Hi, @MrSarius I think it's really cool to better format the container data structure in the debugger panel of VS Code. I wanted to give the new lldb a try, however, as a VS Code rookie I got stuck when writing launch.json. The only way I know how to use a customized lldb is to set `miDebuggerPath` but it requires an `lldb-mi` rather than `lldb`. Could you please help me out? It would be great if you can share your launch.json and what kind of extension and configuration type is needed(I am currently using cppdbg)

MrSarius:
> Hi, @MrSarius I think it's really cool to better format the container data structure in the debugger panel of VS Code. I wanted to give the new lldb a try, however, as a VS Code rookie I got stuck when writing launch.json. The only way I know how to use a customized lldb is to set `miDebuggerPath` but it requires an `lldb-mi` rather than `lldb`. Could you please help me out? It would be great if you can share your launch.json and what kind of extension and configuration type is needed(I am currently using cppdbg)

I have added a "Test changes" at the bottom section to the PR.

eloparco (approved):

TianlongLiang:
LGTM

PR #2221, opened 2023-05-17
AOT: Emit string with '\0' as terminator

no1wudi:


core/iwasm/aot/aot_loader.c
```diff
@@ -363,17 +362,11 @@ load_string(uint8 **p_buf, const uint8 *buf_end, AOTModule *module,
...
     else if (is_load_from_file_buf) {
-        /* As the file buffer can be referred to after loading,
-           we use the 2 bytes of size to adjust the string:
-           move string 2 byte backward and then append '\0' */
-        str = (char *)(p - 2);
-        bh_memmove_s(str, (uint32)(str_len + 1), p, (uint32)str_len);
-        str[str_len] = '\0';
+        /* The string is always terminated with '\0', use it directly.
+         * In this case, the file buffer can be reffered to after loading.
+         */
```
wenyongh, line 368: Had better add `bh_assert(p[str_len - 1] == '\0')` or add check (in case that the AOT file is modified).

PR #2228, opened 2023-05-22
The python-binding needs python>=3.9

FromLiQg:
* Exceptions have been added to the `.py` file that require execution with `python>=3.9`.
* Add explanation in the markdown document
* The following command was executed on `compile.sh`: `chmod +x compile.sh`


lum1n0us:
Plus, https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#python-requires

FromLiQg:
> Plus, https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#python-requires

@lum1n0us  Thank you. I couldn't find a way to control Python version on Google and chatGPT this afternoon.

lum1n0us:
LGTM

PR #2233, opened 2023-05-24
aot_compile_op_call: remove a wrong optimization

yamt:
Unlike a tail-call, the caller of an ordinary recursive call doesn't necessarily return immediately.


wenyongh (commented):
LGTM

PR #2237, opened 2023-05-25
[dpgo] add descriptions in fast-jit IR dump

lum1n0us:


core/iwasm/fast-jit/jit_ir.c
```diff
@@ -1425,3 +1425,16 @@ jit_lock_reg_in_insn(JitCompContext *cc, JitInsn *the_insn, JitReg reg_to_lock)
         jit_set_last_error(cc, "generate insn failed");
     return ret;
 }
+
+#if WASM_ENABLE_FAST_JIT_DUMP != 0
+void
+jit_insn_add_comments(JitInsn *insn, const char *formatter, ...)
```
wenyongh, line 1431: Had better change `formatter` to `format`, like printf/snprintf:
int printf(const char *format, ...);
int snprintf(char *str, size_t size, const char *format, ...);

core/iwasm/fast-jit/jit_ir.h
```diff
@@ -1886,6 +1891,11 @@ jit_cc_update_cfg(JitCompContext *cc);
     for ((I) = (CC)->_ann._label_num; (I) > 0; (I)--)  \
         if (((B) = (CC)->_ann._label_basic_block[(I)-1]))
 
+#if WASM_ENABLE_FAST_JIT_DUMP != 0
+void
+jit_insn_add_comments(JitInsn *insn, const char *formatter, ...);
```
wenyongh, line 1896: Same as above

core/iwasm/fast-jit/jit_ir.c
```diff
@@ -1425,3 +1425,16 @@ jit_lock_reg_in_insn(JitCompContext *cc, JitInsn *the_insn, JitReg reg_to_lock)
...
+{
+    va_list ap;
```
wenyongh, line 1433: Shall we check whether insn is NULL? GEN_INSN may return return.

Issue #2242, opened 2023-05-29
aot_loader RtlAddFunctionTable logic seems to assume a particular order of functions


yamt:
the aot_loader logic to build a table for RtlAddFunctionTable seems to have an assumption that functions are ordered by their code addresses.
https://github.com/bytecodealliance/wasm-micro-runtime/blob/76be848ec34ebd3285fa5840a1db1e81c9e0f941/core/iwasm/aot/aot_loader.c#L1675-L1681
however, as far as understand, LLVMAddFunction order isn't guaranteed to be preserved in the object file.
(LLVM is free to change the order for optimizations, pc-relative addressing, etc.)
and i couldn't find anything is our aot compiler which ensures the expected order.

yamt:
as we specifiy the same UnwindInfoAddress for all entries, i wonder if it's ok to use a single entry covering the whole code section instead.

yamt:
> as we specify the same UnwindInfoAddress for all entries, i wonder if it's ok to use a single entry covering the whole code section instead.

@wenyongh any idea?

wenyongh:
@yamt Hi, yes, currently the implementation of AOT hw bound check for windows 64-bit has defects:

(1) Using RtlAddFunctionTable to dynamically register the AOT functions to the system is the ideal way if we can unwind the AOT functions in the calling stack, and the exception of accessing guard pages can be caught in the aot_runtime.c `invoke_native_with_hw_bound_check` by using `__try { ... } __except(..) { .. }` sentence:
  https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/aot/aot_runtime.c#L1387-L1411

Note that in the interpreter mode, we use `__try { ... } __except(..) { .. }` and it works:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_runtime.c#L2372-L2379

But here I am not sure how to make the AOT functions be unwound successfully, the exception can be caught in `runtime_exception_handler` (registered by AddVectoredExceptionHandler):
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/common/wasm_runtime_common.c#L204

while it cannot be forwarded to the `except` sentence of `__try { .. } __except(..) { .. }` when we return EXCEPTION_CONTINUE_SEARCH in runtime_exception_handler. So I changed the `RIP` and let aot code continue to execute:
[...]

PR #2216, opened 2023-05-15
experiment to query stack sizes for jit

yamt:

Issue #2105, opened 2023-04-04
"native stack overflow" detection is sometimes inappropriate


yamt:
`check_stack_boundary` uses the number of parameters and locals to calculate the required stack: https://github.com/bytecodealliance/wasm-micro-runtime/blob/6af87855b44d092082d43fc756a1dd37fb87a914/core/iwasm/compilation/aot_emit_function.c#L945-L950

unfortunately, the estimation is not always appropriate.

for example, the code attached to this issue is compiled to wasm bytecode like:
```
/opt/wasi-sdk-20.0/bin/clang++ -O0 -fno-exceptions a.cxx
```
(don't ask me why `-O0` is used here. how to build a wasm module is not important to the rest of this issue.)
```
000d7b func[14] <main>:
 000d7c: c1 21 7f                   | local[3..4291] type=i32
 000d7f: 23 80 80 80 80 00          | global.get 0 <__stack_pointer>
 000d85: 21 02                      | local.set 2
 000d87: 41 10                      | i32.const 16
 000d89: 21 03                      | local.set 3
 000d8b: 20 02                      | local.get 2
 000d8d: 20 03                      | local.get 3
 000d8f: 6b                         | i32.sub
 000d90: 21 04                      | local.set 4
 000d92: 20 04                      | local.get 4
 000d94: 24 80 80 80 80 00          | global.set 0 <__stack_pointer>
```
note that the function has a ton of locals.
`check_stack_boundary` uses the number of locals to estimate the stack usage.

at least there are a few issues:

* the estimation seems to assume that explicit `alloca`s are the major source of stack consumption. actually, it depends. if there are not enough machine registers, llvm places temporary variables on stack. you can see it in the `--opt-level=0` output below.
* the estimation assumes later optimization passes don't change the stack consumption much. actually it does. in the above example code, the optimized output of the function only consumes a small amount of stack. you can see it in the `--opt-level=3` output below.

amd64 `wamrc --opt-level=0` output of the function:
```
0000000000000270 <aot_func#3>:
     270: 48 81 ec 98 4e 01 00          subq    $85656, %rsp            # imm = 0x14E98
     277: 41 89 f0                      movl    %esi, %r8d
     27a: 48 89 bc 24 70 4e 01 00       movq    %rdi, 85616(%rsp)
     282: 48 89 f8                      movq    %rdi, %rax
     285: 48 83 c0 10                   addq    $16, %rax
```

[...]

PR #2244, opened 2023-05-30
aot/jit native stack bound check improvement

yamt:
summary:

Move the native stack overflow check from the caller to the callee because the former doesn't work for call_indirect and imported functions.

Make the stack usage estimation more accurate.
Instead of making a guess from the number of wasm locals in the function, use the LLVM's idea of the stack size of each MachineFunction. The former is inaccurate because a) it doesn't reflect optimization passes and b) wasm locals are not the only reason to use stack.

To use the post-compilation stack usage information without requiring 2-pass compilation or machine-code imm rewrites, introduce a global array to store stack consumption of each functions.
for JIT, use a custom IRCompiler with an extra pass to fill the array.
for AOT, use `clang -fstack-usage` equivalent instead because we support external llc.

Re-implement function call stack usage estimation to reflect the real calling conventions better.
(`aot_estimate_stack_usage_for_function_call`)

Re-implement stack estimation logic (`--enable-memory-profiling`) based on the new machinery.

discussions:
https://github.com/bytecodealliance/wasm-micro-runtime/issues/2105

todo/known issues/open questions:

* ~~implement 32-bit case~~
* ~~fill the stack_sizes array for jit (use something similar to https://github.com/bytecodealliance/wasm-micro-runtime/pull/2216)~~
* ~~fix jit tier up (or confirm it isn't broken)~~ reading the code, i couldn't find anything broken.
* ~~ensure appropriate jit partitioning (ensure to compile the function body before executing the corresponding wrapper)~~
* ~~account caller-side stack consumption (cf https://github.com/bytecodealliance/wasm-micro-runtime/issues/2105#issuecomment-1543533575)~~
    * ~~what to do for native function calls?~~ do nothing special, at least within this PR
* ~~fix external llc. pass -fstack-usage to the external command?~~
* ~~re-implement `enable_stack_estimation` based on the new machinary~~
* what to do for RtlAddFunctionTable?
    * it seems broken regardless of this PR. but this PR might break it further. 
    * i'm not even sure how i can test it. is it for AddVectoredExceptionHandler?
    * see also: https://github.com/bytecodealliance/wasm-micro-runtime/issues/2242
    * references:
        * https://learn.microsoft.com/en-us/windows/win32/api/winnt/nf-winnt-rtladdfunctiontable
        * https://learn.microsoft.com/en-us/windows/win32/debug/pe-format?redirectedfrom=MSDN#the-pdata-section
* test
    * ~~this test module consumes about 8MB of stack (wamr aot with llvm 14, amd64) https://github.com/yamt/toywasm/blob/1cc6d551b0fcd10cc8c8b3516c48ba08015e6ad6/wat/many_stack.wat.jinja#L37-L38~~ worked as expected
    * ~~investigate assertion failure seen with js.wasm~~
    * ~~investigate app heap corruptions seen with aot~~ https://github.com/bytecodealliance/wasm-micro-runtime/issues/2275
    * non x86 archs
       * xtensa looks ok
       * i have no plan to test others right now 
* ~~benchmark. `noinline` can have severe implications for certain type of modules. however, as wasm is usually a compiler target, hopefully fundamental inlining has already been done before aot/jit compilation.~~ https://github.com/bytecodealliance/wasm-micro-runtime/pull/2244#issuecomment-1588769538
* ~~remove/disable debug code~~
* ~~do something for func_ctx->debug_func. just disable for the wrapper func?~~
* ~~add missing error checks~~
* ~~reduce code dup. probably make aot_create_func_context use create_basic_func_context.~~
* ~~fix errors caused by empty function~~
* ~~look at x86_32 failure on the ci~~
    * ~~Segv https://github.com/bytecodealliance/wasm-micro-runtime/pull/2260~~
    *~~Nan related issue: an `i32.reinterpret_f32` test in conversions.wast was failing. it's x87 flds/fstp which doesn't preserve sNaN. the problem is not specific to this PR. it seems working on main branch just by luck. (it would fail if you disable optimizations.) i don't think there's a simple way to fix it w/o changing the aot ABI. https://github.com/bytecodealliance/wasm-micro-runtime/pull/2269~~


yamt:
a quick benchmark with coremark
coremark

```
# wamr versions
# base: 7ec77598dd5c62eafdbe03eca883bc42781f097e
# new: be166e8a4fdeef4b941a979d62ab81e62cdf4ddf (https://github.com/bytecodealliance/wasm-micro-runtime/pull/2244)

# wamrc options
# bc0: --bounds-checks=0
# bc1: --bounds-checks=1
# bc1-emp: --bounds-checks=1 --enable-memory-profiling
```

script: https://gist.github.com/yamt/78de859809694a893b7c7732a1025722

wenyongh (commented):
@yamt Do we need to add extra option if we want to enable the new stack overflow check? Or we just use it as normal, e.g. `wamrc --target=i386 -o test.aot test.wasm`?

yamt:
> @yamt Do we need to add extra option if we want to enable the new stack overflow check? Or we just use it as normal, e.g. `wamrc --target=i386 -o test.aot test.wasm`?

the same usage as before.

wenyongh:
> > @yamt Do we need to add extra option if we want to enable the new stack overflow check? Or we just use it as normal, e.g. `wamrc --target=i386 -o test.aot test.wasm`?
> 
> the same usage as before.

Got it, thanks. I just had a quick review, it looks good but it is a little complex, I need to read more carefully about the `aot_llvm.c` and `aot_emit_function.c`, and do some tests.

core/iwasm/compilation/aot_emit_function.c
```diff
@@ -610,6 +473,88 @@ check_app_addr_and_convert(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
...
+
+    /* pointers to results */
+    nb = comp_ctx->pointer_size;
+    for (i = 1; i < result_count; i++) {
+        size = ((size + nb - 1) & -nb) + nb;
+    }
+
+    /* return address */
+    nb = comp_ctx->pointer_size;
+    size = ((size + nb - 1) & -nb) + nb;
```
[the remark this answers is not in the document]
yamt: i meant the caller's pc here. result_count is not relevant.

core/iwasm/compilation/aot_compiler.c
```diff
@@ -2815,6 +2826,51 @@ aot_emit_llvm_file(AOTCompContext *comp_ctx, const char *file_name)
...
+    do {
+        rsz = fread(buf, sizeof(buf), 1, sfp);
+        if (rsz > 0) {
+            size_t wsz = fwrite(buf, rsz, 1, dfp);
+            if (wsz < rsz) {
+                LOG_DEBUG("write error");
+                goto fail;
+            }
+        }
+        if (rsz < sizeof(buf)) {
```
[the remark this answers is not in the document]
yamt: i bumped the buffer size.

core/iwasm/compilation/aot_emit_aot_file.c
```diff
@@ -2418,6 +2418,237 @@ aot_resolve_object_data_sections(AOTObjectData *obj_data)
...
+            continue;
+        }
+        sizes[func_idx] = sz;
+        found++;
+    }
+    fclose(fp);
+    if (precheck_found != count) {
+        LOG_ERROR("%" PRIu32 " precheck entries found where %" PRIu32
+                  " entries are expected",
+                  precheck_found, count);
```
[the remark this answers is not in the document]
yamt: sure

core/iwasm/compilation/aot_emit_function.c
```diff
@@ -610,6 +473,87 @@ check_app_addr_and_convert(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
...
+
+    /* return address */
+    nb = comp_ctx->pointer_size;
+    size = align_uint(size, nb) + nb;
+
+    /*
+     * some extra for possible arch-dependent things like
+     * 16-byte alignment for x86_64.
+     */
+    size += 16;
```
wenyongh, line 545: How about `size = align_uint(size, 16)`?

yamt: i'm not sure how likely `size` here at this point match the alignment of the real calls.
i feel it's safer to use the max value.

yamt:
while this works for x86. it doesn't seem working well for xtensa. let me investigate a bit.

yamt:
> while this works for x86. it doesn't seem working well for xtensa. let me investigate a bit.

there are at least two problems:

* with its windowed abi, tail call elimination is difficult. the xtensa version of llvm doesn't implement it.
* it allocates the area of function call arguments as a part of caller's stack frame. unlike x86, it seems that it's already included the stack size reported by MFI->getStackSize().

wenyongh:
> > while this works for x86. it doesn't seem working well for xtensa. let me investigate a bit.
> 
> there are at least two problems:
> 
> * with its windowed abi, tail call elimination is difficult. the xtensa version of llvm doesn't implement it.
> * it allocates the area of function call arguments as a part of caller's stack frame. unlike x86, it seems that it's already included by the stack size reported by MFI->getStackSize().

Thanks, do you mean changing `size += 16;` to `size = align_uint(size, 16)` doesn't work for xtensa, or this PR has issue for xtensa?

yamt:
> > > while this works for x86. it doesn't seem working well for xtensa. let me investigate a bit.
> > 
> > 
> > there are at least two problems:
> > 
> > * with its windowed abi, tail call elimination is difficult. the xtensa version of llvm doesn't implement it.
> > * it allocates the area of function call arguments as a part of caller's stack frame. unlike x86, it seems that it's already included by the stack size reported by MFI->getStackSize().
> 
> Thanks, do you mean changing `size += 16;` to `size = align_uint(size, 16)` doesn't work for xtensa, or this PR has issue for xtensa?

this PR.
the approach with a wrapper function somehow assumes efficient tail call.

yamt:
> > while this works for x86. it doesn't seem working well for xtensa. let me investigate a bit.
> 
> there are at least two problems:
> 
>     * with its windowed abi, tail call elimination is difficult. the xtensa version of llvm doesn't implement it.

as a tail call in general seems impossible in xtensa windowed abi, i suspect there is no simple solution.

also, riscv seems to prevent tail call optimization in some cases. (when a function have too many parameters to pass via registers?)

a possible workaround is to tweak our aot abi to make >N parameters via a pointer like the following.
that way the stack consumption of the wrapper functions will not be too large even w/o tail call optimization. i feel it's a bit too intrusive though.
```
struct func1_stack_params {
     arg3
     arg4
};

func1(exec_env, arg1, arg2, struct func1_stack_params *)

caller()
{
    struct func1_stack_params params;
    func1(exec_env, arg1, arg2, &params);
}
```

>     * it allocates the area of function call arguments as a part of caller's stack frame. unlike x86, it seems that it's already included by the stack size reported by MFI->getStackSize().

this is just a matter of adding some target dependent code. (eg `if (xtensa)`)

wenyongh:
It is good to me if it is only for xtensa 32-bit. For xtensa 64-bit linux/macos, we can also use stack hw boundary check, right?

yamt:
i fixed xtensa case. it's still inefficient, but not broken.
while i haven't tested on a real hardware yet, the wamrc output looks reasonable.

core/iwasm/compilation/aot_llvm.c
```diff
@@ -38,17 +47,356 @@ wasm_type_to_llvm_type(const AOTLLVMTypes *llvm_types, uint8 wasm_type)
...
+                          uint32 func_index, uint32 orig_param_count,
+                          LLVMTypeRef func_type, LLVMValueRef wrapped_func)
+{
+    LLVMValueRef precheck_func;
+    LLVMBasicBlockRef begin;
+    LLVMBasicBlockRef check_top_block;
+    LLVMBasicBlockRef update_top_block;
+    LLVMBasicBlockRef stack_bound_check_block;
+    LLVMBasicBlockRef call_wrapped_func_block;
+    LLVMValueRef *params = NULL;
```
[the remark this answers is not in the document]
yamt: done

core/iwasm/compilation/aot.h
```diff
@@ -19,6 +19,15 @@ extern "C" {
 #define AOT_FUNC_PREFIX "aot_func#"
 #endif
 
+#ifndef AOT_FUNC_PREFIX2
+#define AOT_FUNC_PREFIX2 "aot_func2#"
```
wenyongh, line 23: "aot_func2#n" is very similar to "aot_func#n", it makes a little confusing to read the llvm-ir file and the generated machine code. Could we change to other name, how about:
```C
#define AOT_FUNC_INTERNAL_PREFIX "aot_func_internal#"
```
let `aot_func#n` call `aot_func_interanl#n`?

core/iwasm/compilation/aot_compiler.c
```diff
@@ -2761,6 +2761,17 @@ aot_compile_wasm(AOTCompContext *comp_ctx)
             aot_handle_llvm_errmsg("failed to addIRModule", err);
             return false;
         }
+
+        if (comp_ctx->stack_sizes != NULL) {
+            LLVMOrcJITTargetAddress addr;
+            if ((err = LLVMOrcLLLazyJITLookup(comp_ctx->orc_jit, &addr,
+                                              aot_stack_sizes_name))) {
+                LLVMOrcDisposeThreadSafeModule(orc_thread_safe_module);
```
wenyongh, line 2769: No need to LLVMOrcDisposeThreadSafeModule again, see the comment above, now orc_jit will destroy the thread safe module when it is destroyed.

core/iwasm/compilation/aot_emit_aot_file.c
```diff
@@ -2418,6 +2442,293 @@ aot_resolve_object_data_sections(AOTObjectData *obj_data)
...
+
+    if (!(sym_itr = LLVMObjectFileCopySymbolIterator(obj_data->binary))) {
+        aot_set_last_error("llvm get symbol iterator failed.");
+        return false;
+    }
+
+    while (!LLVMObjectFileIsSymbolIteratorAtEnd(obj_data->binary, sym_itr)) {
+        if ((name = LLVMGetSymbolName(sym_itr))
+            && !strcmp(name, aot_stack_sizes_name)) {
+            uint64_t sz = LLVMGetSymbolSize(sym_itr);
```
wenyongh, line 2604: Had better use uint64

core/iwasm/compilation/aot_emit_aot_file.c
```diff
@@ -1634,7 +1638,27 @@ aot_emit_object_data_section_info(uint8 *buf, uint8 *buf_end, uint32 *p_offset,
...
+            uint32 ss_offset = obj_data->stack_sizes_offset;
+            uint32 ss_size =
+                obj_data->func_count * sizeof(*obj_data->stack_sizes);
+            LOG_VERBOSE("Replacing stack_sizes in %s section, offset %" PRIu32
+                        ", size %" PRIu32,
+                        obj_data->stack_sizes_section_name, ss_offset, ss_size);
+            if (ss_offset > 0) {
+                EMIT_BUF(data_section->data, ss_offset);
+            }
+            EMIT_BUF(obj_data->stack_sizes, ss_size);
```
wenyongh, line 1653: Seems that `ss_offset + ss_size` may be larger than `data_section->size` according to L1654?
Had better add check here?
```C
EMIT_BUF(obj_data->stack_sizes, ss_size <= data_section->size - ss_offset ? ss_size : data_section->size - ss_offset);
```

core/iwasm/compilation/aot_llvm.c
```diff
@@ -1483,6 +1868,44 @@ create_target_machine_detect_host(AOTCompContext *comp_ctx)
...
+
+    bh_assert(comp_ctx != NULL);
+    bh_assert(comp_ctx->jit_stack_sizes != NULL);
+
+    if (namelen >= sizeof(buf)) {
+        LOG_DEBUG("too long name: %.*s", (int)namelen, name);
+        return;
+    }
+    /* ensure NUL termination */
+    memcpy(buf, name, namelen);
```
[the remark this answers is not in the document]
yamt: ok

core/iwasm/compilation/aot.h
yamt: ok

core/iwasm/compilation/aot_compiler.c
yamt: ok

core/iwasm/compilation/aot_emit_aot_file.c
yamt: ok

core/iwasm/compilation/aot_emit_aot_file.c
yamt: > Seems that `ss_offset + ss_size` may be larger than `data_section->size` according to L1654? Had better add check here?

no, it can't.

> 
> ```c
> EMIT_BUF(obj_data->stack_sizes, ss_size <= data_section->size - ss_offset ? ss_size : data_section->size - ss_offset);
> ```

i added an assertion instead.

wenyongh (commented):
LGTM

yamt:
> i fixed xtensa case. it's still inefficient, but not broken. while i haven't tested on a real hardware yet, the wamrc output looks reasonable.

lightly tested on esp32-devkitc. it worked as expected so far.

xujuntwt95329 (approved):
LGTM

wenyongh:
> > i fixed xtensa case. it's still inefficient, but not broken. while i haven't tested on a real hardware yet, the wamrc output looks reasonable.
> 
> lightly tested on esp32-devkitc. it worked as expected so far.

OK, it seems there is no comment from other developers, let's merge this PR?

yamt:
> > > i fixed xtensa case. it's still inefficient, but not broken. while i haven't tested on a real hardware yet, the wamrc output looks reasonable.
> > 
> > 
> > lightly tested on esp32-devkitc. it worked as expected so far.
> 
> OK, it seems there is no comment from other developers, let's merge this PR?

i have no problem with it

PR #2250, opened 2023-06-01
Never ignore PTHREAD_STACK_MIN

g0djan:
We encountered that some toolchains define PTHREAD_STACK_MIN without PTHREAD_STACK_DEFAULT.
It makes impossible to use to spawn threads because WAMR uses a fallback to default values that are lower than PTHREAD_STACK_MIN.


g0djan:
Hi @wenyongh I hoped setting `APP_THREAD_STACK_SIZE_DEFAULT` to `PTHREAD_STACK_MIN` will be harmless and will solve the problem for toolchains that don't set `PTHREAD_STACK_DEFAULT` but CI doesn't seem to pass at all.

Do you have any suggestion how APP_THREAD_STACK_SIZE_DEFAULT can be inferred or whether it's possible at all?

wenyongh:
> Hi @wenyongh I hoped setting `APP_THREAD_STACK_SIZE_DEFAULT` to `PTHREAD_STACK_MIN` will be harmless and will solve the problem for toolchains that don't set `PTHREAD_STACK_DEFAULT` but CI doesn't seem to pass at all.
> 
> Do you have any suggestion how APP_THREAD_STACK_SIZE_DEFAULT can be inferred or whether it's possible at all?

PTHREAD_STACK_MIN may be very large and or a little small:
https://xenomai.org/pipermail/xenomai-git/2015-October/005373.html

I tested in Ubuntu, it is 16KB, it is too small to set APP_THREAD_STACK_SIZE_DEFAULT with PTHREAD_STACK_MIN. Not sure why not add macro definitions in your own CMakeLists.txt? e.g. 
```cmake
add_definitions(-DAPP_THREAD_STACK_SIZE_DEFAULT=n1)
add_definitions(-DAPP_THREAD_STACK_SIZE_MIN=n2)
```
My understanding is that APP_THREAD_STACK_SIZE_DEFAULT has no direct relationship with PTHREAD_STACK_MIN, setting APP_THREAD_STACK_SIZE_DEFAULT according to PTHREAD_STACK_MIN seems unreasonable.

PR #2254, opened 2023-06-04
[dpgo] Use condbr counters to set branch_weight

lum1n0us:


core/iwasm/compilation/aot_compiler.c
```diff
@@ -2615,14 +2615,20 @@ verify_module(AOTCompContext *comp_ctx)
...
+        if (msg) {
+            if (msg[0] != '\0')
+                aot_set_last_error(msg);
+            else
+                aot_set_last_error("LLVM IR Module is broken");
             LLVMDisposeMessage(msg);
-            return false;
         }
-        LLVMDisposeMessage(msg);
+        aot_set_last_error("LLVM IR Module is broken");
```
wenyongh, line 2630: Had better `else aot_set_last_error("LLVM IR Module is broken");`, or it may overwrite L2625

core/iwasm/compilation/aot_emit_numberic.c
```diff
@@ -387,7 +387,8 @@ compile_rems(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
              LLVMValueRef left, LLVMValueRef right, LLVMValueRef overflow_cond,
              bool is_i32)
 {
-    LLVMValueRef phi, no_overflow_value, zero = is_i32 ? I32_ZERO : I64_ZERO;
+    LLVMValueRef phi, no_overflow_value, zero = is_i32 ? I32_ZERO : I64_ZERO,
+                                         cond_br;
```
wenyongh, line 391: Code looks a little confused, how about define cond_br in a new line:
```C
LLVMValueRef cond_br;
```

core/iwasm/compilation/aot_llvm_extra.cpp
```diff
@@ -381,28 +411,53 @@ wasm_dpgo_set_function_entry_count(LLVMValueRef function, uint32 count)
...
     I->setMetadata(LLVMContext::MD_prof, MDB.createBranchWeights(Weights));
 }
 
+void
+wasm_dpgo_unlike_true_branch(LLVMContextRef context, LLVMValueRef cond_br)
+{
+    bh_assert(LLVMIsABranchInst(cond_br));
+    bh_assert(LLVMIsConditional(cond_br));
+
+    uint32 weights[2] = { 1, 1000 };
```
wenyongh, line 439: We will update the values after profiling, right?

core/iwasm/interpreter/wasm.h
```diff
@@ -867,14 +876,59 @@ wasm_dpgo_get_cur_prof_cnt_info(WASMModule *module, uint32 func_idx,
...
 {
     snprintf(buf, buf_size, "  OP:0x%x,OFFSET:%u,CNT:%u,1ST_IDX:%u",
              cnt->opcode, cnt->offset, cnt->counter_amount,
              cnt->first_counter_idx);
 }
 
+/* search in `prof_cnts_info[i]` with the given `offset` of `struct
+ * WASMProfCntInfo` and return `counter_amount` and `first_counter_idx`
+ */
+static inline bool
```
wenyongh, line 890: Function body may be a little big, had better not use inline? Will we put the implementation into core/iwasm/compilation later?

core/iwasm/interpreter/wasm.h
```diff
@@ -867,14 +876,59 @@ wasm_dpgo_get_cur_prof_cnt_info(WASMModule *module, uint32 func_idx,
...
+            *first_counter_idx = cnt_info->first_counter_idx;
+            return true;
+        }
+        elem = bh_list_elem_next(elem);
+    }
+
+    return false;
+}
+
+static inline void
```
wenyongh, line 914: Same as above

core/iwasm/compilation/aot_llvm_extra.cpp
lum1n0us: Actually no. both `wasm_dpgo_unlike_true_branch()` and `wasm_dpgo_unlike_false_branch()` kinds similar with `unlikely()` and `likely()`. The programmer is very sure about the most likely path during coding.

We are using both functions for exception handling. We will give the exception handler target a very low weight, usually 1, and the normal path a high weight, usually 1000.

Since those exception handling llvm jit generated is different from fast jit generated, there is no counter for those *condbr* in exception handling. But lucky for us, we don't need that.

core/iwasm/interpreter/wasm.h
lum1n0us: You are right. I will do that in the later code refractory PR.

core/iwasm/interpreter/wasm.h
lum1n0us: You are right. I will do that in the later code refractory PR.

PR #2260, opened 2023-06-05
aot/jit: set module layout

yamt:
LLVM 15 and later sometimes performs wrong optimizations without this.


yamt:
> LLVM 15 and later sometimes performs wrong optimizations without this.

i observed wrong optimizations with https://github.com/bytecodealliance/wasm-micro-runtime/pull/2244 .
but i don't think the problem is specific to the PR.

i dunno if it can actually happen with the IR generated by the aot compiler in main branch.

wenyongh (commented):
LGTM

PR #2261, opened 2023-06-05
Fix compile error of wamrc with llvm-13/llvm-14

wenyongh:


yamt (approved):
lgtm. can you consider to add a CI job to test these?

wenyongh:
> lgtm. can you consider to add a CI job to test these?

It needs to build llvm-13/llvm-14 and add them to cache, which may take some efforts, maybe we can test them manually now and add the CI jobs sometime later.

PR #2271, opened 2023-06-07
GC: Make Func/Struct/Array type inherited from WASMType

no1wudi:


core/iwasm/interpreter/wasm.h
```diff
@@ -273,28 +273,14 @@ typedef struct WASMType {
     struct WASMType *parent_type;
     uint32 parent_type_idx;
 
-    uint32 data[1];
+    uint32 data[0];
```
wenyongh, line 276: How about removing this line? It is unused and may cause compile warning for some compilers.

no1wudi: OK

xujuntwt95329:
LGTM

PR #2277, opened 2023-06-08
appease unused warning on min_uint64

yamt:


lum1n0us:
LGTM. But just wondering why not use a macro, like `#define MIN ...`, to simpilfy both `min_uint64()` and `min_uint32()`.

wenyongh (commented):
LGTM

PR #2279, opened 2023-06-08
fix a heap corruption bug in ems realloc

yamt:


wenyongh (commented):
LGTM

samples/mem_allocator/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+cmake_minimum_required(VERSION 3.0)
```
wenyongh, line 1: Had better add license header.

samples/mem_allocator/main.c
```diff
@@ -0,0 +1,53 @@
+#include <stdio.h>
```
wenyongh, line 1: Should add license header

samples/mem_allocator/CMakeLists.txt
yamt: ok. i will add.

yamt: done

samples/mem_allocator/main.c
yamt: done

Issue #2235, opened 2023-05-24
Run AOT failed on windows


Waitforomething2023:
Os:windows 10
wamrc_ver:latest x86
toolchain:emcc
llvm_ver:15.X

```cpp
#include <iostream>

int main()
{

    std::cout<<"Hello, world!\n";
    return 0;



}

```

cmooand:wamrc --target-abi=msvc --target=i386 -o test.aot test.wasm

```log
Create AoT compiler with:
  target:        i386
  target cpu:
  target triple: i386-pc-windows-msvc
  cpu features:
  opt level:     3
  size level:    1
  output format: AoT file
Compile success, file test.aot was generated.


iwasm.exe test.aot
Load relocation section failed: invalid relocation type 20.

**Comment by wenyongh** [comment]
Hi, the relocations for Windows x86-32 haven't been implemented yet, I tried to enable the relocation type 20 as you mentioned, see [win_x86_32.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/11561566/win_x86_32.zip), could you have a try?

**Comment by Waitforomething2023** [comment]
> Hi, the relocations for Windows x86-32 haven't been implemented yet, I tried to enable the relocation type 20 as you mentioned, see [win_x86_32.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/11561566/win_x86_32.zip), could you have a try?


Hi I've tried,buf it failed.
```log

AOT module load failed: resolve symbol _aot_func#384 failed
```

wenyongh:
It seems that the function symbol generated in Windows 32bit starts with '_', could you modify the aot_loader.c:
```diff
diff --git a/core/iwasm/aot/aot_loader.c b/core/iwasm/aot/aot_loader.c
index 5345fb2d..9a431d02 100644
--- a/core/iwasm/aot/aot_loader.c
+++ b/core/iwasm/aot/aot_loader.c
@@ -1947,6 +1947,19 @@ do_text_relocation(AOTModule *module, AOTRelocationGroup *group,
             }
             symbol_addr = module->func_ptrs[func_index];
         }
+#if defined(BH_PLATFORM_WINDOWS) && defined(BUILD_TARGET_X86_32)
+        else if (!strncmp(symbol, "_" AOT_FUNC_PREFIX,
+                          strlen("_" AOT_FUNC_PREFIX))) {
+            p = symbol + strlen("_" AOT_FUNC_PREFIX);
[...]

PR #2285, opened 2023-06-12
Enable windows x86-32 AOT relocations

wenyongh:
Implement relocation types IMAGE_REL_I386_DIR32 and IMAGE_REL_I386_REL32,
fix find symbol of AOT function, and implement symbol __aulldiv.


yamt:
the changes look reasonable to me.
(i'm ignorant of these windows specific things though)

wenyongh:
> the changes look reasonable to me. (i'm ignorant of these windows specific things though)

Yes, I think it should be good, I just collect the changes mentioned in #2235, it should have been tested by the author.

PR #2289, opened 2023-06-13
Make boundary check behavior configurable

no1wudi:
Notice: This is a non standard behavior, in this patch, boundary check can be disabled, for a memory access operation, if address A is not in linear memory space, then read/write to it directly as this:
···c
    do {                                                        \
        uint64 offset1 = (uint64)offset + (uint64)addr;         \
        if (offset1 + bytes <= (uint64)get_linear_mem_size())   \
            /* If offset1 is in valid range, maddr must also    \
               be in valid range, no need to check it again. */ \
            maddr = memory->memory_data + offset1;              \
        else {                                                  \
            if (module->disable_boundary_check)                 \
                maddr = (uint8_t *)(uintptr_t)offset1;          \
            else                                                \
                goto out_of_bounds;                             \
        }                                                       \
    } while (0)
···


core/iwasm/interpreter/wasm_runtime.h
```diff
@@ -320,7 +320,9 @@ struct WASMModuleInstance {
 
     /* Default WASM operand stack size */
     uint32 default_wasm_stack_size;
-    uint32 reserved[3];
+    /* Use one of reserved element to implement disable_boundary_check */
+    uint32 disable_boundary_check;
```
no1wudi, line 324: Consider to move this filed to WASMModule, don't use this filed to restore reserved bytes.

core/iwasm/compilation/aot_llvm.h
```diff
@@ -179,6 +179,7 @@ typedef struct AOTFuncContext {
     LLVMBasicBlockRef func_return_block;
     LLVMValueRef exception_id_phi;
     LLVMValueRef func_type_indexes;
+    LLVMValueRef disable_boundary_check;
```
no1wudi, line 182: This member can be removed

core/iwasm/common/wasm_memory.c
```diff
@@ -274,7 +274,8 @@ wasm_runtime_validate_app_addr(WASMModuleInstanceCommon *module_inst_comm,
         goto fail;
     }
 
-    /* integer overflow check */
+    /* integer overflow check, and if memory boundary check enabled, app_offset
```
no1wudi, line 277: Consider add a unchecked version of these functions for AOT, to improve performance (dont't need do if in each memory access operation)

yamt:
what's the purpose of this?

no1wudi:
> what's the purpose of this?

Let the wasm app access the host memory directly, in some application we bypass the libc to the host side, raw pointer from host maybe accessed in the wasm app.

For this case, @wenyongh noticed that maybe we can implement https://github.com/WebAssembly/memory-control/blob/main/proposals/memory-control/Overview.md proposal.

yamt:
> > what's the purpose of this?
> 
> Let the wasm app access the host memory directly, in some application we bypass the libc to the host side, raw pointer from host maybe accessed in the wasm app.

ok

> 
> For this case, @wenyongh noticed that maybe we can implement https://github.com/WebAssembly/memory-control/blob/main/proposals/memory-control/Overview.md proposal.

the "preferred" option there is based on multi-memory.
it means you can't access it via ordinary C pointers in wasm.
is it ok for your purpose?

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -1764,6 +1764,10 @@ wasm_instantiate(WASMModule *module, bool is_sub_inst,
     module_inst->e->function_count =
         module->import_function_count + module->function_count;
 
+#if WASM_CONFIGUABLE_BOUNDS_CHECKS != 0
+    module_inst->e->disable_bounds_checks = false;
```
wenyongh, line 1768: Seems as above, no need to set again.

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -35,23 +35,24 @@ typedef float64 CellType_F64;
 #define CHECK_MEMORY_OVERFLOW(bytes)                             \
     do {                                                         \
         uint64 offset1 = (uint64)offset + (uint64)addr;          \
-        if (offset1 + bytes <= (uint64)get_linear_mem_size())    \
+        if (disable_bounds_checks                                \
```
wenyongh, line 38: Will it impact the performance when `WASM_CONFIGUABLE_BOUNDS_CHECKS == 0`?

wamr-compiler/CMakeLists.txt
```diff
@@ -45,6 +45,7 @@ add_definitions(-DWASM_ENABLE_DUMP_CALL_STACK=1)
 add_definitions(-DWASM_ENABLE_PERF_PROFILING=1)
 add_definitions(-DWASM_ENABLE_LOAD_CUSTOM_SECTION=1)
 add_definitions(-DWASM_ENABLE_LIB_WASI_THREADS=1)
+add_definitions(-DWASM_CONFIGUABLE_BOUNDS_CHECKS=1)
```
wenyongh, line 48: Seems no need to change here as no file is changed under `core/iwasm/compilation`.

core/iwasm/common/wasm_runtime_common.h
```diff
@@ -593,6 +593,12 @@ wasm_runtime_set_user_data(WASMExecEnv *exec_env, void *user_data);
 WASM_RUNTIME_API_EXTERN void *
 wasm_runtime_get_user_data(WASMExecEnv *exec_env);
 
+#if WASM_CONFIGUABLE_BOUNDS_CHECKS != 0
+/* See wasm_export.h for description */
+WASM_RUNTIME_API_EXTERN void
+wasm_runtime_set_bounds_checks(wasm_module_inst_t module_inst, bool enable);
```
wenyongh, line 599: Had better use `WASMModuleInstanceCommon *`

core/iwasm/interpreter/wasm_interp_fast.c
no1wudi: No impact here, if WASM_CONFIGUABLE_BOUNDS_CHECKS  == 0, the disable_bounds_checks is always false, compiler is smart enough to handle it.

core/config.h
```diff
@@ -456,4 +456,9 @@
 #define WASM_DISABLE_WRITE_GS_BASE 0
 #endif
 
+/* Configurable bounds check */
+#ifndef WASM_CONFIGURABLE_BOUNDS_CHECK
+#define WASM_CONFIGURABLE_BOUNDS_CHECK 0
```
wenyongh, line 461: Should be WASM_CONFIGURABLE_BOUNDS_CHECK**S**?

core/config.h
```diff
@@ -456,4 +456,9 @@
 #define WASM_DISABLE_WRITE_GS_BASE 0
 #endif
 
+/* Configurable bounds check */
```
wenyongh, line 459: Had better `bounds checks`

core/iwasm/aot/aot_runtime.h
```diff
@@ -89,6 +89,10 @@ typedef struct AOTFunctionInstance {
 
 typedef struct AOTModuleInstanceExtra {
     CApiFuncImport *c_api_func_imports;
+#if WASM_CONFIGUABLE_BOUNDS_CHECKS != 0
+    /* Disable bounds check or not */
```
wenyongh, line 93: Had better `bounds checks`

core/iwasm/common/wasm_memory.c
```diff
@@ -354,19 +367,32 @@ wasm_runtime_addr_app_to_native(WASMModuleInstanceCommon *module_inst_comm,
...
 
-    if (memory_inst->memory_data <= addr && addr < memory_inst->memory_data_end)
+    if (bounds_checks) {
+        if (memory_inst->memory_data <= addr
+            && addr < memory_inst->memory_data_end) {
+
+            return addr;
+        }
+    }
+    /* If bound checks is disabled, and app_offset is not NULL,
```
wenyongh, line 391: Had better `bounds checks`

core/iwasm/common/wasm_memory.c
```diff
@@ -378,17 +404,28 @@ wasm_runtime_addr_native_to_app(WASMModuleInstanceCommon *module_inst_comm,
...
         return 0;
     }
 
-    if (memory_inst->memory_data <= addr && addr < memory_inst->memory_data_end)
+    if (bounds_checks) {
+        if (memory_inst->memory_data <= addr
+            && addr < memory_inst->memory_data_end)
+            return (uint32)(addr - memory_inst->memory_data);
+    }
+    /* If bound checks is disabled, and addr is not NULL,
```
wenyongh, line 424: `bounds checks`

core/iwasm/common/wasm_memory.c
```diff
@@ -378,17 +404,28 @@ wasm_runtime_addr_native_to_app(WASMModuleInstanceCommon *module_inst_comm,
...
+       return the address directly */
```
wenyongh, line 425: `return the offset`?

core/iwasm/common/wasm_memory.c
```diff
@@ -460,13 +497,29 @@ wasm_check_app_addr_and_convert(WASMModuleInstance *module_inst, bool is_str,
...
         goto fail;
     }
 
+    bounds_checks =
+        wasm_runtime_is_bounds_checks_enabled((wasm_module_inst_t)module_inst);
+
     native_addr = memory_inst->memory_data + app_buf_addr;
 
+    if (!bounds_checks) {
+        /* If bound checks is disabled, and app_buf_addr is not NULL,
```
wenyongh, line 512: bounds checks?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -2482,6 +2482,50 @@ wasm_runtime_get_custom_data(WASMModuleInstanceCommon *module_inst_comm)
...
 
+#if WASM_CONFIGUABLE_BOUNDS_CHECKS != 0
+void
+wasm_runtime_set_bounds_checks(WASMModuleInstanceCommon *module_inst,
+                               bool enable)
+{
+#if WASM_ENABLE_INTERP != 0
+    if (module_inst->module_type == Wasm_Module_Bytecode) {
+        ((WASMModuleInstanceExtra *)((WASMModuleInstance *)module_inst)->e)
+            ->disable_bounds_checks = enable ? false : true;
```
wenyongh, line 2493: An issue is that whether we should consider the `OS_ENABLE_HW_BOUND_CHECK`: when it is defined and hw bound check is enabled, shall we always set disable_bounds_checks true here?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -2482,6 +2482,50 @@ wasm_runtime_get_custom_data(WASMModuleInstanceCommon *module_inst_comm)
...
+
+bool
+wasm_runtime_is_bounds_checks_enabled(WASMModuleInstanceCommon *module_inst)
+{
+#if WASM_CONFIGUABLE_BOUNDS_CHECKS != 0
+#if WASM_ENABLE_INTERP != 0
+    if (module_inst->module_type == Wasm_Module_Bytecode) {
+        return !((WASMModuleInstanceExtra *)((WASMModuleInstance *)module_inst)
+                     ->e)
+                    ->disable_bounds_checks;
```
wenyongh, line 2514: Same as above, when OS_ENABLE_HW_BOUND_CHECK is defined and hw bound check is enabled, shall we always return false here?

core/iwasm/include/wasm_export.h
```diff
@@ -914,6 +914,25 @@ wasm_runtime_set_custom_data(wasm_module_inst_t module_inst,
 WASM_RUNTIME_API_EXTERN void *
 wasm_runtime_get_custom_data(wasm_module_inst_t module_inst);
 
+/**
+ * Set the memory bounds check flag of a WASM module instance.
```
wenyongh, line 918: bounds checks

core/iwasm/include/wasm_export.h
```diff
@@ -914,6 +914,25 @@ wasm_runtime_set_custom_data(wasm_module_inst_t module_inst,
...
+ * 
+ * @param module_inst the WASM module instance
+ * @param enable the flag to enable/disable the memory bounds check
+ */
+WASM_RUNTIME_API_EXTERN void
+wasm_runtime_set_bounds_checks(wasm_module_inst_t module_inst,
+                               bool enable);
+/**
+ * Check if the memory bounds check flag is enabled for a WASM module instance.
```
wenyongh, line 927: bounds checks

core/iwasm/include/wasm_export.h
```diff
@@ -914,6 +914,25 @@ wasm_runtime_set_custom_data(wasm_module_inst_t module_inst,
...
+ * 
+ * @param module_inst the WASM module instance
+ *
+ * @return true if the memory bounds check flag is enabled, false otherwise
```
wenyongh, line 931: bounds checks

core/iwasm/interpreter/wasm_runtime.h
```diff
@@ -241,6 +241,10 @@ typedef struct WASMModuleInstanceExtra {
         && WASM_ENABLE_LAZY_JIT != 0)
     WASMModuleInstance *next;
 #endif
+#if WASM_CONFIGUABLE_BOUNDS_CHECKS != 0
+    /* Disable bounds check or not */
```
wenyongh, line 245: bounds checks

core/iwasm/include/wasm_export.h
```diff
@@ -914,6 +914,25 @@ wasm_runtime_set_custom_data(wasm_module_inst_t module_inst,
 WASM_RUNTIME_API_EXTERN void *
 wasm_runtime_get_custom_data(wasm_module_inst_t module_inst);
 
+/**
+ * Set the memory bounds check flag of a WASM module instance.
+ * 
+ * @param module_inst the WASM module instance
+ * @param enable the flag to enable/disable the memory bounds check
```
wenyongh, line 921: bounds checks

product-mini/platforms/posix/main.c
```diff
@@ -65,6 +65,9 @@ print_help()
 #endif
     printf("  --repl                   Start a very simple REPL (read-eval-print-loop) mode\n"
            "                           that runs commands in the form of \"FUNC ARG...\"\n");
+#if WASM_CONFIGUABLE_BOUNDS_CHECKS != 0
+    printf("  --disable-bounds-checks  Disable bounds checking for memory accesses\n");
```
wenyongh, line 69: How about `Disable boundary checks for memory accesses`?

core/iwasm/common/wasm_runtime_common.c
no1wudi: I think it's ok, maybe this work with OS_ENABLE_HW_BOUND_CHECK will make ther performance better.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -2482,6 +2482,61 @@ wasm_runtime_get_custom_data(WASMModuleInstanceCommon *module_inst_comm)
...
+    }
+#endif
+#endif
+
+    /* Disable bounds checks and enable hw bounds checks maybe make
+       the performance better. */
+#ifndef OS_ENABLE_HW_BOUND_CHECK
+    return true;
```
wenyongh, line 2534: Should return false? And Had better move these lines before L2513?

no1wudi: OK

core/iwasm/common/wasm_runtime_common.c
no1wudi: @wenyongh Emm, In this patch there is a extra behavior that map the app addr 0 to host addr NULL, this is a non standard option, so can't pass the spec test with OS_ENABLE_HW_BOUND_CHECK.

Should we add a extra option to control this behavior?

no1wudi: Maybe we should do the mapping in wrapper, I'll remove it later.

wenyongh: OK, so don't add control for OS_ENABLE_HW_BOUND_CHECK in this function.

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -1174,6 +1176,8 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
     uint8 local_type, *global_addr;
     uint32 cache_index, type_index, param_cell_num, cell_num;
     uint8 value_type;
+    bool disable_bounds_checks = !wasm_runtime_is_bounds_checks_enabled(
+        (WASMModuleInstanceCommon *)module);
```
wenyongh, line 1180: Not sure why not use WASM_CONFIGURABLE_BOUNDS_CHECKS to control the code? Now no matter WASM_CONFIGURABLE_BOUNDS_CHECKS is 1 or 0, `disable_bounds_checks` will be read from a function, and compiler cannot optimize the code in CHECK_MEMORY_OVERFLOW/CHECK_BULK_MEMORY_OVERFLOW, it impacts the performance.

no1wudi: If WASM_CONFIGUABLE_BOUNDS_CHECKS is disabled, wasm_runtime_is_bounds_checks_enabled  will just return true, so the compiler can do the optmizations.

wenyongh: Not very sure, the function isn't inline, my understanding is that some compilers may not be able to optimize it.

How about using WASM_CONFIGUABLE_BOUNDS_CHECKS to control the whole wasm_runtime_is_bounds_checks_enabled:
```C
#if WASM_CONFIGUABLE_BOUNDS_CHECKS  != 0
bool
wasm_runtime_is_bounds_checks_enabled()
{
   ...
}
void
wasm_runtime_set_bounds_checks()
{
    ...
}
```
And in interpreter and wasm_memory.c:

```C
static inline bool
is_bounds_checks_enabled()
{
#if WASM_CONFIGUABLE_BOUNDS_CHECKS  != 0
    return wasm_runtime_is_bounds_checks_enabled();
#else
    return true;
#endif
}
```
And call is_bounds_checks_enabled instead of wasm_runtime_is_bounds_checks_enabled.

no1wudi: Or, how about make them to be inline function directly ?

wenyongh: Make `is_bounds_checks_enabled` static inline and `wasm_runtime_is_bounds_checks_enabled` non static/inline since the latter may be called as library API?

no1wudi: OK

wenyongh (approved):
LGTM

PR #2290, opened 2023-06-13
Reformat wasm_interp_classic.c

no1wudi:


yamt:
i guess you should somehow tell it to clang-format-12 by modifying the top-level .clang-format file.
i'm not sure if it's possible though.

lum1n0us:
Might need to discuss the modification of *.clang-format* firstly

no1wudi:
I'm not very familiar with clang-format options, I've tried `AllowShortBlocksOnASingleLine: SBS_Always`, but have many unexpected change with this option.
```
AllowShortBlocksOnASingleLine (ShortBlockStyle) clang-format 3.5 [¶](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#allowshortblocksonasingleline)
Dependent on the value, while (true) { continue; } can be put on a single line.

Possible values:

SBS_Never (in configuration: Never) Never merge blocks into a single line.

while (true) {
}
while (true) {
  continue;
}
SBS_Empty (in configuration: Empty) Only merge empty blocks.

while (true) {}
while (true) {
  continue;
}
SBS_Always (in configuration: Always) Always merge short blocks into a single line.

while (true) {}
while (true) { continue; }
```

PR #2291, opened 2023-06-13
fix dockerfile linter warnings

TianlongLiang:
Should we also consider adding a CI that uses hadolint to check Dockerfile?

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -81,7 +81,7 @@ RUN mkdir /opt/bazelisk \
 #
 # install clang+llvm
 ARG LLVM_VER=14
-RUN apt-get purge -y clang-10 llvm-10 && apt autoremove -y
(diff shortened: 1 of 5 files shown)


yamt:
is "replace curl with wget" change a recommendation from hadolint?
do you still need to install curl after the change?
i feel a little sad because i personally prefer curl over wget. :-)

TianlongLiang:
Yes, it's a [warning](https://github.com/hadolint/hadolint/wiki/DL4001) in the hadolint.
Since it's a devdocker, I think maybe should keep it installed in case anyone wants to use curl inside the container.

wenyongh:
Hi @tonibofarull, this PR changes some docker file of wasi-nn, could you help check whether it is good to you? Thanks.

tonibofarull (approved):

wenyongh (commented):
LGTM

PR #2296, opened 2023-06-16
Reduce jobs number in WAMR CI 

### LINKED ISSUE #2267: Build iwasm with ubuntu 20 only nightly
We need to build ubuntu 20 only nightly and ubuntu 22 per every PR to reduce the number of jobs and duplication for both pipelines as for #2135

Zzzabiyaka:
Fixes #2267 

This PR doesn't decrease coverage, because every job is tested either per PR or nightly (instead of 2 times as it was before). Actually, it even increases it because Android is tested with Ubuntu 20 now (it was disabled before)

### DIFF
--- a/.github/workflows/nightly_run.yml
+++ b/.github/workflows/nightly_run.yml
@@ -4,6 +4,14 @@
 name: nightly_run
 
 on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+    #running nightly pipeline if you're changing it 
+    paths:
+      - ".github/workflows/nightly_run.yml"
+      
   # midnight UTC
   schedule:
     - cron: "0 0 * * *"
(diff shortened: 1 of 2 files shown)


TianlongLiang (approved):

.github/workflows/nightly_run.yml
```diff
@@ -4,6 +4,14 @@
 name: nightly_run
 
 on:
+  pull_request:
+    types:
+      - opened
```
wenyongh, line 9: Will this CI be triggered when each PR is opened and updated?

.github/workflows/nightly_run.yml
```diff
@@ -39,24 +47,16 @@ jobs:
...
-  build_llvm_libraries_on_ubuntu_2204:
-    uses: ./.github/workflows/build_llvm_libraries.yml
-    with:
-      os: "ubuntu-22.04"
-      arch: "X86"
   
   build_wamrc:
     needs:
-      [build_llvm_libraries_on_ubuntu_2004, build_llvm_libraries_on_ubuntu_2204]
+      [build_llvm_libraries_on_ubuntu_2004]
```
wenyongh, line 53: Some of the tests on ubuntu-22.04 are removed, are they already tested in the compilation_on_android_ubuntu.yml?

Zzzabiyaka: can you elaborate a bit more?

They're used to be the same workflow (nightly one and per-PR one) but nightly had sanitizers tests. 

So for me the only thing that we lose if sanitizers tests on ubuntu 22, which are unlikely to fail if they pass for ubuntu 20

.github/workflows/nightly_run.yml
wenyongh: Seems OK, I missed the `paths:` expression, L13.
Can refer to other CI for how it is triggered:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/.github/workflows/compilation_on_android_ubuntu.yml#L13-L15

.github/workflows/nightly_run.yml
wenyongh: Got it, it is good to me.

wenyongh (commented):
LGTM

lum1n0us:
LGTM

PR #2302, opened 2023-06-20
Specify wasm features supported by feature_flags

no1wudi:


core/iwasm/aot/aot_loader.c
```diff
@@ -522,6 +523,56 @@ load_target_info_section(const uint8 *buf, const uint8 *buf_end,
         return false;
     }
 
+    /* Check feature flags */
```
wenyongh, line 526: Had better add a function to check the flags, or the code may be too long here:
```C
static bool
check_feature_flags(uint32 feature_flags);
```

core/iwasm/aot/aot_runtime.h
```diff
@@ -298,6 +312,8 @@ typedef struct AOTTargetInfo {
     uint32 e_version;
     /* Processor-specific flags */
     uint32 e_flags;
+    /* Specify wasm features supported */
+    uint32 feature_flags;
```
wenyongh, line 316: Not sure whether uint32 is not enough for all features in the feature? Should we use uint64?

core/iwasm/aot/aot_loader.c
no1wudi: OK

core/iwasm/aot/aot_runtime.h
no1wudi: Changed to uint64 now

PR #2303, opened 2023-06-20
Update to Node >=16 in extension release process

MrSarius:
Since building the extension now depends on node >=16 (see [this](https://github.com/bytecodealliance/wasm-micro-runtime/pull/2292#issuecomment-1596761347)), build_wamr_vscode_ext.yml will fail on the next release without this change.

### DIFF
--- a/.github/workflows/build_wamr_vscode_ext.yml
+++ b/.github/workflows/build_wamr_vscode_ext.yml
@@ -20,10 +20,10 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: Use Node.js 14.x
+      - name: Use Node.js 16.x
         uses: actions/setup-node@v3
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/build_wamr_vscode_ext.yml
```diff
@@ -20,10 +20,10 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: Use Node.js 14.x
```
MrSarius, line 23: What was the reason for specifying a particular node version here in the first place? @wenyongh

wenyongh: I am not sure, the code was written by me, but I think it is to specify a runnable version here. Upgrading to 16.x should be OK since14.x is a little old.

PR #2314, opened 2023-06-26
Dev/zephyr esp32 fix

TianlongLiang:
Provide a Dockerfile.old for fixing ESP32 custom linker scripts not working properly with the newer version of Zephyr
Provide a Dockerfile with newer Zephyr for other boards
Update the corresponding readme


wenyongh (approved):
LGTM

PR #2320, opened 2023-06-27
Implement threads suspension/resume

wenyongh:


core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1154,6 +1154,18 @@ wasm_runtime_set_max_thread_num(uint32 num)
...
 }
+
+void
+wasm_runtime_enter_safe_state()
+{
+    wasm_cluster_change_curr_thread_to_safe();
+}
+
+void
+jeff_runtime_exit_safe_state()
```
yamt, line 1165: copy and paste mistake?

wenyongh: Yes, thanks, fixed.

Issue #2316, opened 2023-06-26
an alternate "indirect" way to find "stack_sizes" is necessary for xip


yamt:

yamt:
the CI failure looks unrelated
```
Err:1 http://azure.archive.ubuntu.com/ubuntu jammy/universe amd64 kconfig-frontends-nox amd64 4.11.0.1+dfsg-6
  Could not connect to azure.archive.ubuntu.com:80 (52.154.174.208), connection timed out [IP: 52.154.174.208 80]
Err:2 http://azure.archive.ubuntu.com/ubuntu jammy/universe amd64 genromfs amd64 0.5.2-4
  Unable to connect to azure.archive.ubuntu.com:http: [IP: 52.154.174.208 80]
E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/universe/k/kconfig-frontends/kconfig-frontends-nox_4.11.0.1%2bdfsg-6_amd64.deb  Could not connect to azure.archive.ubuntu.com:80 (52.154.174.208), connection timed out [IP: 52.154.174.208 80]
E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/universe/g/genromfs/genromfs_0.5.2-4_amd64.deb  Unable to connect to azure.archive.ubuntu.com:http: [IP: 52.154.174.208 80]
```

yamt:
"spec test on nuttx" seems still failing. @wenyongh can you reopen this?

yamt:
```
============> run br_table failed with a non-zero return code 101

==================== LAST LOG of br_table ====================

  expected: 'SystemExit(1)'
  got: 'iwasm --heap-size=0 -v=0 --repl /tmp/tmprdma9ep7.aot
AOT module load failed: cannot apply relocation to text section for aot file generated with "--enable-indirect-mode" flag
```

br_table test case has a large br_table, which happens to be translated into relocations like:
```
000003dc R_RISCV_HI20             .Lswitch.table.aot_func_internal#15
000003e0 R_RISCV_LO12_I           .Lswitch.table.aot_func_internal#15
```
i'm not sure why it was not a problem in the previous code.

i'm not sure how to fix either.
maybe we can disable switch lowering as a workaround?

any idea?

wenyongh:
Maybe it is due to some passes added for XIP mode don't work again with LLVM-15.0, as issue #2328 reported by @no1wudi, I think we can wait until the PR is merged.

[...]

Issue #2144, opened 2023-04-23
GC (Garbage Collection) AOT support


wenyongh:
# GC (Gargbage Collection) AOT support

## Motivation
In recent days we had refactored the WAMR GC interpreter implementation to support most features of the latest [GC MVP](https://github.com/WebAssembly/gc/blob/master/proposals/gc/MVP.md) spec proposal, and support the bytecodes generate by the [binaryen](https://github.com/WebAssembly/binaryen) toolchain as its bytecode definition is a little different from the definition of GC MVP, so as to run the workloads compiled by the [ts2wasm compiler](https://github.com/bytecodealliance/wasm-micro-runtime/issues/2037), which uses binaryen as its backend. To gain better performance and to reduce the footprint, we started to enable the GC AOT support.

## Design Goals
- Support platform linux, macos, windows, android, linux-sgx, nuttx and so on
- Support target x86-64, x86-32, aarch64, arm32, thumb, riscv64, riscv32
- Support the same functionalities as interpreter 
- Support the bytecode definition of both GC MVP and GC binaryen
- Pass the spec cases and the cases compiled by ts2wasm compiler
- Support GC multi-threading
- Export necessary GC APIs
- Standardize the interface between runtime and GC allocator

## Non Design Goals now
- Improve the performance of GC allocator and GC reclaim process, and resolve the fragment issue of GC heap
  There may be some shortcomings of the current ems heap allocator, we hope we can replace it with other GC heap allocator in the feature. And the interfaces between runtime and GC allocator will be standardized, so we can easily integrate a new GC allocator into runtime.

## Changes of the AOT file format
There will be many changes in the AOT file format, after discussion, some decisions were made:
- The AOT file version will be upgraded from 1 to 2
- The new version won’t be compatible with the old version since it is complex and difficult to main the compatibility with old version, in other words, the AOT file generated by the old wamrc won’t be able to be run by the new AOT runtime.
- Remove the current handles to keep the compatibility in AOT runtime, so as to make the code clear and concise
- Consider the support for multiple memories, exception handling and component model
- Strings in the AOT file are always terminated by ‘\0’, so runtime can directly use them and doesn’t need to create a copy

### Some detail changes
- Enable feature flags, each bit denotes whether a feature is support or not
- Extend the structures which have value type field, add extra heap type field if the value type is a reference type that requires heap type info
- Extend the function type to function/struct/array/sub/sub_final/recursive types, and add a flag to specify which defined type it is
[...]

PR #2322, opened 2023-06-27
aot: avoid possible relocations around "stack_sizes" for indirect mode

yamt:
Fixes https://github.com/bytecodealliance/wasm-micro-runtime/issues/2316

Lightly tested on riscv64 qemu.

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -39,10 +39,10 @@ bh_static_assert(offsetof(AOTModuleInstance, func_type_indexes)
                  == 6 * sizeof(uint64));
 bh_static_assert(offsetof(AOTModuleInstance, cur_exception)
                  == 13 * sizeof(uint64));
-bh_static_assert(offsetof(AOTModuleInstance, aot_stack_sizes)
-                 == 13 * sizeof(uint64) + 128 + 11 * sizeof(uint64));
 bh_static_assert(offsetof(AOTModuleInstance, global_table_data)
-                 == 13 * sizeof(uint64) + 128 + 12 * sizeof(uint64));
+                 == 13 * sizeof(uint64) + 128 + 11 * sizeof(uint64));
+
+bh_static_assert(offsetof(AOTModuleInstanceExtra, stack_sizes) == 0);
 
 static void
 set_error_buf(char *error_buf, uint32 error_buf_size, const char *string)
@@ -1212,7 +1212,7 @@ aot_instantiate(AOTModule *module, bool is_sub_inst, WASMExecEnv *exec_env_main,
 #endif
     module_inst->default_wasm_stack_size = stack_size;
 
-    module_inst->aot_stack_sizes =
+    ((AOTModuleInstanceExtra *)module_inst->e)->stack_sizes =
         aot_get_data_section_addr(module, AOT_STACK_SIZES_SECTION_NAME, NULL);
 
 #if WASM_ENABLE_PERF_PROFILING != 0
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -39,8 +39,10 @@ bh_static_assert(offsetof(AOTModuleInstance, func_type_indexes)
                  == 6 * sizeof(uint64));
 bh_static_assert(offsetof(AOTModuleInstance, cur_exception)
                  == 13 * sizeof(uint64));
-bh_static_assert(offsetof(AOTModuleInstance, global_table_data)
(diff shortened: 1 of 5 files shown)


core/iwasm/aot/aot_runtime.c
```diff
@@ -39,8 +39,10 @@ bh_static_assert(offsetof(AOTModuleInstance, func_type_indexes)
                  == 6 * sizeof(uint64));
 bh_static_assert(offsetof(AOTModuleInstance, cur_exception)
                  == 13 * sizeof(uint64));
-bh_static_assert(offsetof(AOTModuleInstance, global_table_data)
+bh_static_assert(offsetof(AOTModuleInstance, aot_stack_sizes)
                  == 13 * sizeof(uint64) + 128 + 11 * sizeof(uint64));
+bh_static_assert(offsetof(AOTModuleInstance, global_table_data)
+                 == 13 * sizeof(uint64) + 128 + 12 * sizeof(uint64));
```
yamt, line 45: i'm not happy to break offsets.
should i add the new member after the variable parts of the structure?

wenyongh: Adding new fields in the WASMModuleInstance/AOTModuleInstance may cause compatibility issue: the AOT code also use pre-calculated offsets to access the global data and table instance (which are at the end of AOTModuleInstance and are allocated together with AOTModuleInstance), these offsets will change if new field is inserted, and the old AOT file (generated by older wamrc) won't run well with the new runtime.

How about adding the `aot_stack_sizes` into exec_env, after field `native_stack_top_min`:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/common/wasm_exec_env.h#L91

yamt: is there a longer term plan about compatibility breakage?
like having a flag day for https://github.com/bytecodealliance/wasm-micro-runtime/issues/2144 when we can clean up these things?

yamt: i moved it to aot inst extra.

wenyongh: > is there a longer term plan about compatibility breakage? like having a flag day for #2144 when we can clean up these things?

Yes, we hope to upgrade the AOT file format in the GC AOT feature and don't support the old AOT file format again after the feature is implemented. The GC AOT is supposed to be finished in this year, and we may merge it into main branch sometime this year.

core/iwasm/aot/aot_runtime.h
```diff
@@ -88,6 +88,7 @@ typedef struct AOTFunctionInstance {
 } AOTFunctionInstance;
 
 typedef struct AOTModuleInstanceExtra {
+    const uint32 *stack_sizes;
```
wenyongh, line 91: Had better use `DefPointer(const uint32 *, stack_sizes)`? Suppose the aot code may access more fields here in the future, we can ensure the offsets of these fields are fixed, and are the same in compilation time and execution time.

core/iwasm/compilation/aot_llvm.c
```diff
@@ -327,9 +339,32 @@ aot_add_precheck_function(AOTCompContext *comp_ctx, LLVMModuleRef module,
...
      */
+    LLVMValueRef stack_sizes;
+    if (comp_ctx->is_indirect_mode) {
+        uint32 offset_u32 = get_inst_extra_offset(comp_ctx);
+        offset_u32 += offsetof(AOTModuleInstanceExtra, stack_sizes);
+        LLVMValueRef offset = I32_CONST(offset_u32);
+        if (!offset) {
+            goto fail;
+        }
+        LLVMValueRef stack_sizes_p =
```
wenyongh, line 350: Had better declare the variables at the beginning of code piece `{ ... }`. And had better add empty lines if needed. We usually keep that coding style to make the code easier to read.
For example, there is no empty line from L342 to L402, a little intensive:)

yamt: i generally feel lesser lines are easier to read. but ok. i will follow your style.

core/iwasm/aot/aot_runtime.h
yamt: ok

core/iwasm/compilation/aot_llvm.c
yamt: done

core/iwasm/aot/aot_runtime.h
yamt: done

wenyongh (commented):
LGTM

no1wudi:
LGTM

Issue #2315, opened 2023-06-26
POSITION_INDEPENDENT_CODE issue


yamt:
where iwasm is POSITION_INDEPENDENT_CODE set, vmlib should be POSITION_INDEPENDENT_CODE set as well.
otherwise, for targets like armhf, link can fail.
```
/usr/bin/arm-linux-gnueabihf-ld: libvmlib.a(posix.c.o): relocation R_ARM_MOVW_ABS_
NC against `a local symbol' can not be used when making a shared object; recompile
 with -fPIC                                                                       
```

yamt:
fixed

PR #2323, opened 2023-06-27
product-mini/platforms/linux: mark vmlib POSITION_INDEPENDENT_CODE

yamt:
This fixes armhf build for me.
https://github.com/bytecodealliance/wasm-micro-runtime/issues/2315

Note: This is not the only place which seems to have the same problem. For example, many of examples sharing the similar structure probably have the same problem. This commit leaves them for now.


wenyongh (commented):
LGTM

PR #2324, opened 2023-06-28
[dpgo] fix bugs

lum1n0us:


core/iwasm/compilation/iwasm_compl.cmake
```diff
@@ -14,7 +14,10 @@ else()
                     ${IWASM_COMPL_DIR}/*.cpp)
 endif()
 
-if (WAMR_BUILD_DYNAMIC_PGO EQUAL 1)
+if (WAMR_BUILD_FAST_JIT EQUAL 1
+      AND WAMR_BUILD_JIT EQUAL 1
+      AND WAMR_BUILD_DYNAMIC_PGO EQUAL 1)
+  message("involve DPGO sources")
```
wenyongh, line 20: Maybe no need to add the prompt? `message ("     Dynamic PGO enabled")` was already added.

PR #2332, opened 2023-06-30
Increase default native stack size

loganek:
Calling `__wasi_sock_addr_resolve` syscall causes native stack overflow. Given this is a standard function available in WAMR, we should have at least the default stack size large enough to handle this case.

I've updated the socket tests so they also run in separate thread, but the simple retro program is:

```c
#include <wasi_socket_ext.h>
#include <pthread.h>

void *th(void *p)
{
    struct addrinfo *res;
    getaddrinfo("amazon.com", NULL, NULL, &res);
    return NULL;
}

int main(int argc, char **argv)
{
    pthread_t pt;
    pthread_create(&pt, NULL, th, NULL);
    pthread_join(pt, NULL);
    return 0;
}
```
when run on WAMR that was built with stack size 32kB (default), it fails due to native stack overflow.


yamt:
> when run on WAMR that was built with stack size 32kB (default), it fails due to native stack overflow.

is it with linux glibc?

loganek:
> > when run on WAMR that was built with stack size 32kB (default), it fails due to native stack overflow.
> 
> is it with linux glibc?

Yes, for example, see this commit that fixes the `native stack overflow` problem when running on `ubuntu-latest` docker image: https://github.com/loganek/webassembly-curl-ssl/commit/b9a8133a5ae5c0fe774c041fe11ea0b1c49993bc

wenyongh:
@yamt, @no1wudi Is this PR OK to you? Shall we merge it?

no1wudi:
> @yamt, @no1wudi Is this PR OK to you? Shall we merge it?

Yes, it's OK .

yamt:
ok for me

yamt (approved):

PR #2336, opened 2023-07-03
Add a shorthand "--xip" to wamrc

no1wudi:
A shorthand of "--enalbe-indirect-mode --disable-llvm-intrinsics"


wenyongh (commented):
Could you also update the document:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/doc/xip.md

no1wudi:
> Could you also update the document: https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/doc/xip.md

OK

PR #2339, opened 2023-07-04
Avoid switch lowering to lookup tables for XIP

yamt:
Because it involves relocations for the table. (.Lswitch.table.XXX)

Discussions: https://github.com/bytecodealliance/wasm-micro-runtime/issues/2316


yamt:
i tested that this avoids switch-table relocations in the "br_table" test case for riscv32.
(merely tested with https://github.com/bytecodealliance/wasm-micro-runtime/pull/2340, not with riscv32 iwasm.)

no1wudi (approved):

PR #2348, opened 2023-07-06
bring up wamr on esp32-s3 device

dongsheng28849455:
esp32-s3's instruction memory and data memory can be accessed through mutual mirroring way, so we define a feature named as WASM_MEM_DUAL_BUS_MIRROR


core/shared/platform/nuttx/nuttx_platform.c
```diff
@@ -55,6 +83,14 @@ os_mmap(void *hint, size_t size, int prot, int flags)
 
     if ((uint64)size >= UINT32_MAX)
         return NULL;
+
+#if (WASM_MEM_DUAL_BUS_MIRROR != 0)
+    if ((prot & MMAP_PROT_EXEC) != 0) {
+        d_addr = malloc((uint32)size);
+        i_addr = d_addr + MEM_DUAL_BUS_OFFSET;
```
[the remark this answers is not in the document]
wenyongh: Had better not `(void *) + offset`, it may cause compiler error or warning.
Had better `i_addr = (void *)((uint8 *)d_addr + MEM_DUAL_BUS_OFFSET);`

core/iwasm/aot/arch/aot_reloc_xtensa.c
```diff
@@ -219,6 +222,9 @@ apply_relocation(AOTModule *module, uint8 *target_section_addr,
             initial_addend = *(int32 *)insn_addr;
             *(uintptr_t *)insn_addr = (uintptr_t)symbol_addr + initial_addend
                                       + (intptr_t)reloc_addend;
+#if (WASM_MEM_DUAL_BUS_MIRROR != 0)
+            os_dcache_flush();
```
donghengqaz, line 226: Don't need to flush cache in each relocation, when this done, `load_from_sections` will do this when loading sections is done.

core/iwasm/aot/arch/aot_reloc_xtensa.c
```diff
@@ -286,6 +296,9 @@ apply_relocation(AOTModule *module, uint8 *target_section_addr,
 #pragma GCC diagnostic pop
 #endif
 
+#if (WASM_MEM_DUAL_BUS_MIRROR != 0)
+            os_dcache_flush();
```
donghengqaz, line 300: The same as upper comment.

core/shared/platform/nuttx/nuttx_platform.c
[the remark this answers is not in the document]
donghengqaz: By a way, not only external RAM(PSRAM) use mirror bus address, but also internal RAM(SRAM) is the same. So maybe you use this solution for internal RAM, too.

core/shared/platform/nuttx/nuttx_platform.c
```diff
@@ -55,6 +83,17 @@ os_mmap(void *hint, size_t size, int prot, int flags)
...
         return NULL;
+
+#if (WASM_MEM_DUAL_BUS_MIRROR != 0)
+    if ((prot & MMAP_PROT_EXEC) != 0) {
+        d_addr = malloc((uint32)size);
+        if (d_addr == NULL) {
+            return NULL;
+        }
+        i_addr = d_addr + MEM_DUAL_BUS_OFFSET;
+        return in_ibus_ext(i_addr) ? i_addr : d_addr;
```
donghengqaz, line 94: By a way, not only external RAM(PSRAM) use mirror bus address, but also internal RAM(SRAM) is the same. So maybe you use this solution for internal RAM, too.

dongsheng28849455: you mean we should indicate from where to allocate before we do it? and does they have different allocator?

core/iwasm/aot/arch/aot_reloc_xtensa.c
wenyongh: How about adding some comments in this PR? developer may be confused about the code.

core/shared/platform/nuttx/nuttx_platform.c
donghengqaz: Yes, to make nuttx platform easier to use, we could make it support to load AOT to sram and run.

dongsheng28849455:
@donghengqaz @wenyongh @yamt any other comments?

core/iwasm/aot/aot_loader.c
```diff
@@ -3071,8 +3074,16 @@ create_sections(AOTModule *module, const uint8 *buf, uint32 size,
                     bh_assert((uintptr_t)aot_text < INT32_MAX);
 #endif
 #endif
+
+#if (WASM_MEM_DUAL_BUS_MIRROR != 0)
+                    mirrored_text = os_get_dbus_mirror(aot_text);
```
wenyongh, line 3079: Had better add `bh_assert(mirrored_text != NULL)`?

core/iwasm/aot/arch/aot_reloc_xtensa.c
```diff
@@ -207,6 +207,9 @@ apply_relocation(AOTModule *module, uint8 *target_section_addr,
         case R_XTENSA_32:
         {
             uint8 *insn_addr = target_section_addr + reloc_offset;
+#if (WASM_MEM_DUAL_BUS_MIRROR != 0)
+            insn_addr = os_get_dbus_mirror((void *)insn_addr);
```
wenyongh, line 211: Same as above

core/iwasm/aot/arch/aot_reloc_xtensa.c
```diff
@@ -265,6 +268,10 @@ apply_relocation(AOTModule *module, uint8 *target_section_addr,
                 return false;
             }
 
+#if (WASM_MEM_DUAL_BUS_MIRROR != 0)
+            insn_addr = os_get_dbus_mirror((void *)insn_addr);
```
wenyongh, line 272: Same as above

core/shared/platform/nuttx/nuttx_platform.c
```diff
@@ -10,6 +10,46 @@
 #include <nuttx/arch.h>
 #endif
 
+#if defined(CONFIG_ARCH_CHIP_ESP32S3) && (CONFIG_ARCH_CHIP_ESP32S3 != 0)
```
no1wudi, line 13: "#ifdef CONFIG_ARCH_CHIP_ESP32S3" is enough

dongsheng28849455: what about the case of "#define CONFIG_ARCH_CHIP_ESP32S3 0"?

no1wudi: This definition is generated from Kconfig automatically, if ARCH_CHIP_ESP32S3 not selected in Kconfig, it should delete this line instead of define it as 0, I guess.

donghengqaz:
When this PR is merged, I plan to add WAMR to espressif official [components management](https://components.espressif.com/), this can help more people to use WAMR more easily.

shreeve:
@donghengqaz - Phenomenal work on this! You mentioned you would be adding WAMR to espressif official [components management](https://components.espressif.com/). Do you have an update on that?

donghengqaz:
> @donghengqaz - Phenomenal work on this! You mentioned you would be adding WAMR to espressif official [components management](https://components.espressif.com/). Do you have an update on that?

Hi @shreeve  I have been busy with some other development work, so have not started to do this. And after these things are done, I will start this.

yamt:
@donghengqaz can you share nuttx kconfig you used for this?

yamt:
> @donghengqaz can you share nuttx kconfig you used for this?

well, i meant @dongsheng28849455 . wrong completion, sorry.

dongsheng28849455:
> > @donghengqaz can you share nuttx kconfig you used for this?
> 
> well, i meant @dongsheng28849455 . wrong completion, sorry.

it is CONFIG_INTERPRETERS_WAMR_MEM_DUAL_BUS_MIRROR

yamt:
> > > @donghengqaz can you share nuttx kconfig you used for this?
> > 
> > 
> > well, i meant @dongsheng28849455 . wrong completion, sorry.
> 
> it is CONFIG_INTERPRETERS_WAMR_MEM_DUAL_BUS_MIRROR

there is no such a config in https://github.com/apache/nuttx-apps.
i guess you are talking about your private fork?

also, i want to see the whole config.
(because i was struggling to find out a combination of memory configs which can work.)

dongsheng28849455:
> i guess you are talking about your private fork?

It's defined in https://github.com/midokura/incubator-nuttx-apps fork
> because i was struggling to find out a combination of memory configs which can work

What is whole config you mean? is that something could enable several hidden APIs in esp32-s3?
```
    extern void cache_writeback_all(void);
    extern uint32_t Cache_Disable_ICache(void);
    extern void Cache_Enable_ICache(uint32_t autoload);
```

yamt:
> > i guess you are talking about your private fork?
> 
> It's defined in https://github.com/midokura/incubator-nuttx-apps fork

ok. it's a private fork.
do you have any plan to upstream necessary changes?

> 
> > because i was struggling to find out a combination of memory configs which can work
> 
> What is whole config you mean? is that something could enable several hidden APIs in esp32-s3?
> 
> ```
>     extern void cache_writeback_all(void);
>     extern uint32_t Cache_Disable_ICache(void);
>     extern void Cache_Enable_ICache(uint32_t autoload);
> ```

i meant a complete working ".config" file.

specifically i wanted to know values for:
CONFIG_ARCH_USE_TEXT_HEAP
CONFIG_XTENSA_IMEM_USE_SEPARATE_HEAP
CONFIG_ESP32S3_RTC_HEAP
CONFIG_MM_KERNEL_HEAP
CONFIG_ESP32S3_SPIRAM
CONFIG_BUILD_FLAT
CONFIG_BUILD_PROTECTED
CONFIG_ESP32S3_SPIRAM_COMMON_HEAP
etc

but now i suspect this PR is not expected to work with upstream nuttx. is it the case?

dongsheng28849455:
the feature quite depends on esp32-s3's implementation, including the config. that might or not delivered into nuttx upstream. I don't have a plan to abstract an interface for Nuttx to use here.

yamt:
> the feature quite depends on esp32-s3's implementation, including the config. that might or not delivered into nuttx upstream. I don't have a plan to abstract an interface for Nuttx to use here.

i needed the following changes to make it work on esp32s3-devkit.
https://github.com/bytecodealliance/wasm-micro-runtime/pull/3421
https://github.com/apache/nuttx/pull/12328
https://github.com/apache/nuttx-apps/pull/2385

xiaoxiang781216:
> the feature quite depends on esp32-s3's implementation, including the config. that might or not delivered into nuttx upstream. I don't have a plan to abstract an interface for Nuttx to use here.

why adding esp32s3 specific code to the common wamr runtime? @no1wudi please review the arch specific change carefully, we need ensure ALL chips on nuttx get the equal support.

PR #2357, opened 2023-07-13
Fix Enclave Interfaces of WAMR SGX version; Set pointer to NULL after free, otherwise cause UAF/DF

LeoneChen:
1. Fix Enclave Interfaces of WAMR SGX version
2. Set pointer to NULL after free, otherwise cause UAF/DF

### DIFF
--- a/product-mini/platforms/linux-sgx/enclave-sample/Enclave/Enclave.cpp
+++ b/product-mini/platforms/linux-sgx/enclave-sample/Enclave/Enclave.cpp
@@ -7,6 +7,7 @@
 #include <string.h>
 #include <inttypes.h>
 #include <stdbool.h>
+#include <unordered_map>
 
 #include "Enclave_t.h"
 #include "wasm_export.h"
@@ -32,28 +33,6 @@ enclave_print(const char *message)
     return bytes_written;
 }
 }
-
-typedef enum EcallCmd {
-    CMD_INIT_RUNTIME = 0,     /* wasm_runtime_init/full_init() */
-    CMD_LOAD_MODULE,          /* wasm_runtime_load() */
-    CMD_INSTANTIATE_MODULE,   /* wasm_runtime_instantiate() */
-    CMD_LOOKUP_FUNCTION,      /* wasm_runtime_lookup_function() */
-    CMD_CREATE_EXEC_ENV,      /* wasm_runtime_create_exec_env() */
-    CMD_CALL_WASM,            /* wasm_runtime_call_wasm */
-    CMD_EXEC_APP_FUNC,        /* wasm_application_execute_func() */
-    CMD_EXEC_APP_MAIN,        /* wasm_application_execute_main() */
-    CMD_GET_EXCEPTION,        /* wasm_runtime_get_exception() */
-    CMD_DEINSTANTIATE_MODULE, /* wasm_runtime_deinstantiate() */
-    CMD_UNLOAD_MODULE,        /* wasm_runtime_unload() */
-    CMD_DESTROY_RUNTIME,      /* wasm_runtime_destroy() */
-    CMD_SET_WASI_ARGS,        /* wasm_runtime_set_wasi_args() */
-    CMD_SET_LOG_LEVEL,        /* bh_log_set_verbose_level() */
-    CMD_GET_VERSION,          /* wasm_runtime_get_version() */
-#if WASM_ENABLE_STATIC_PGO != 0
-    CMD_GET_PGO_PROF_BUF_SIZE,  /* wasm_runtime_get_pro_prof_data_size() */
-    CMD_DUMP_PGO_PROF_BUF_DATA, /* wasm_runtime_dump_pgo_prof_data_to_buf() */
-#endif
-} EcallCmd;
 
 typedef struct EnclaveModule {
     wasm_module_t module;
@@ -85,6 +64,109 @@ static korp_mutex enclave_module_list_lock = OS_THREAD_MUTEX_INITIALIZER;
 static char global_heap_buf[WASM_GLOBAL_HEAP_SIZE] = { 0 };
 #endif
 
+class PointerManager
+{
+  public:
+    bool add(void *ptr, uint16_t *idx_output)
+    {
+        bool success = false;
+        pthread_rwlock_wrlock(&mRwlock);
+        for (uint16_t i = mCurrentIdx; i < std::numeric_limits<uint16_t>::max();
+             i++) {
+            if (mMap.count(i) == 0) {
+                mMap[i] = ptr;
+                *idx_output = i;
+                success = true;
+                mCurrentIdx++;
+                break;
+            }
+        }
+        pthread_rwlock_unlock(&mRwlock);
+        return success;
+    }
+
+    void *get(uint16_t idx)
+    {
+        void *ptr = nullptr;
+        pthread_rwlock_rdlock(&mRwlock);
+        if (mMap.count(idx)) {
+            ptr = mMap[idx];
+        }
+        pthread_rwlock_unlock(&mRwlock);
+        return ptr;
+    }
+
+    bool remove(uint16_t idx)
+    {
+        bool success = false;
+        pthread_rwlock_wrlock(&mRwlock);
+        if (mMap.erase(idx) == 1) {
+            success = true;
+        }
+        mCurrentIdx = idx;
+        pthread_rwlock_unlock(&mRwlock);
+        return success;
+    }
+
+  private:
+    std::unordered_map<uint16_t, void *> mMap;
+    uint16_t mCurrentIdx = 0;
+    pthread_rwlock_t mRwlock = PTHREAD_RWLOCK_INITIALIZER;
+};
+PointerManager gEnclaveModuleMgr, gWasmModuleInstMgr;
+
+/// @brief Deep copy an array of char string
+/// @param SrcCStrArray Source array of char string
+/// @param DstCStrArray Dest. array of char string, if null, then
+/// automatically malloced
+/// @param Length Array length. If 0, cause fail
+/// @return \p DstCStrArray. If failed, return null, char string elements will
+/// be auto free-ed, if \p DstCStrArray is feed with null, it will be auto
+/// free-ed before return, but if \p DstCStrArray is input with not null, it's
+/// caller duty to free \p DstCStrArray. If success, user need to free all char
+/// string elements before free this \p DstCStrArray.
+char **
+DeepCopyCStrArray(char **SrcCStrArray, char **DstCStrArray, size_t Length)
+{
+    if ((SrcCStrArray == nullptr) or (Length == 0)) {
+        return nullptr;
+    }
+    bool DstIsNull = false;
+    if (DstCStrArray == nullptr) {
+        DstIsNull = true;
+        size_t AllocSize = Length * sizeof(char *);
+        if ((AllocSize <= Length /* Int Overflow */)
+            or (AllocSize
+                > std::numeric_limits<uint32_t>::max() /* Too large */)
+            or (!(DstCStrArray = (char **)wasm_runtime_malloc(
+                      AllocSize)) /* Malloc Fail */)) {
+            return nullptr;
+        }
+    }
+
+    for (size_t i = 0; i < Length; i++) {
+        size_t AllocSize = (strlen(SrcCStrArray[i]) + 1) * sizeof(char);
+        if ((AllocSize > std::numeric_limits<uint32_t>::max() /* Too large */)
+            or (!(DstCStrArray[i] = (char *)wasm_runtime_malloc(
+                      AllocSize)) /* Malloc Fail */)) {
+            for (size_t j = 0; j < i; j++) {
+                wasm_runtime_free(DstCStrArray[j]);
+                DstCStrArray[j] = nullptr;
+            }
+            if (DstIsNull) {
+                // DstCStrArray is allocated by myself, free it
+                wasm_runtime_free(DstCStrArray);
+                DstCStrArray = nullptr;
+            }
+            return nullptr;
+        }
+
+        strcpy(DstCStrArray[i], SrcCStrArray[i]);
+    }
+
+    return DstCStrArray;
+}
+
 static void
 set_error_buf(char *error_buf, uint32 error_buf_size, const char *string)
 {
@@ -92,17 +174,13 @@ set_error_buf(char *error_buf, uint32 error_buf_size, const char *string)
         snprintf(error_buf, error_buf_size, "%s", string);
 }
 
-static void
-handle_cmd_init_runtime(uint64 *args, uint32 argc)
+bool
+ecall_handle_cmd_init_runtime(uint32_t max_thread_num)
 {
-    uint32 max_thread_num;
+    bool ret = false;
     RuntimeInitArgs init_args;
 
-    bh_assert(argc == 1);
-
     os_set_print_function(enclave_print);
-
-    max_thread_num = (uint32)args[0];
 
     memset(&init_args, 0, sizeof(RuntimeInitArgs));
     init_args.max_thread_num = max_thread_num;
@@ -118,17 +196,18 @@ handle_cmd_init_runtime(uint64 *args, uint32 argc)
     /* initialize runtime environment */
     if (!wasm_runtime_full_init(&init_args)) {
         LOG_ERROR("Init runtime environment failed.\n");
-        args[0] = false;
-        return;
+        ret = false;
+        goto exit;
     }
 
-    args[0] = true;
-
+    ret = true;
     LOG_VERBOSE("Init runtime environment success.\n");
+exit:
+    return ret;
 }
 
-static void
-handle_cmd_destroy_runtime()
+void
+ecall_handle_cmd_destroy_runtime()
 {
     wasm_runtime_destroy();
 
@@ -201,18 +280,19 @@ is_xip_file(const uint8 *buf, uint32 size)
     return false;
 }
 
-static void
-handle_cmd_load_module(uint64 *args, uint32 argc)
+bool
+ecall_handle_cmd_load_module(char *wasm_file, uint32_t wasm_file_size,
+                             char *error_buf, uint32_t error_buf_size,
+                             uint16_t *enclave_module_idx)
 {
-    uint64 *args_org = args;
-    char *wasm_file = *(char **)args++;
-    uint32 wasm_file_size = *(uint32 *)args++;
-    char *error_buf = *(char **)args++;
-    uint32 error_buf_size = *(uint32 *)args++;
+    bool ret = false;
     uint64 total_size = sizeof(EnclaveModule) + (uint64)wasm_file_size;
     EnclaveModule *enclave_module;
 
-    bh_assert(argc == 4);
+    if (wasm_file == nullptr or enclave_module_idx == nullptr) {
+        ret = false;
+        goto exit;
+    }
 
     if (!is_xip_file((uint8 *)wasm_file, wasm_file_size)) {
         if (total_size >= UINT32_MAX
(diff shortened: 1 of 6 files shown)


product-mini/platforms/linux-sgx/enclave-sample/Enclave/Enclave.edl
```diff
@@ -20,11 +20,71 @@ enclave {
...
+        uint32_t wasi_argc,
+        [in, count=addr_pool_list_size] const char **addr_pool_list,
+        uint32_t addr_pool_list_size
+      );
+      public void ecall_handle_cmd_get_version(
+        [out] uint32_t *major,
+        [out] uint32_t *minor,
+        [out] uint32_t *patch
+      );
+      public uint32_t ecall_handle_cmd_get_pgo_prof_buf_size(uint16_t wasm_module_inst_idx);
```
TianlongLiang, line 78: maybe use macro `#if WASM_ENABLE_STATIC_PGO != 0` for those two functions?

TianlongLiang:
Hi, I see that there some CI tests are failing, seems not able to call `ecall_handle_cmd_exec_app_func()`?

BTW, what is the purpose of using idx rather than raw pointer for module_inst and other data? Is it for easier management with the help of the newly introduced pointer manager?

LeoneChen:
> BTW, what is the purpose of using idx rather than raw pointer for module_inst and other data? Is it for easier management with the help of the newly introduced pointer manager?

Use raw pointer is very dangerous, since it can point to in-Enclave memory, it's hard to check pointed memory is a valid context content or invalid sensitive in-Enclave data. So use indexes to replace the raw pointers when passing `enclave_module` and `wasm_module_inst` across the SGX boundary. This method is decribed in [TeeRex](https://www.usenix.org/system/files/sec20-cloosters.pdf), and at the end of section 5.3

> The developers of the Rust SGX SDK acknowledged the problem and promptly updated their code. Akin to our suggestions to the developers, the enclave code now utilizes session identifiers instead of pointers to identify TLS sessions; similar to using file descriptors on Unix-like systems. Upon session creation in tls_client_new, the pointer to the TLS session object is now inserted into a hashmap, which is then used to map the identifier in subsequent ECALLs. Hence, no pointers are passed on the host-to-enclave boundary. This drastically reduces the attack surface of the enclave and eradicates both the vulnerability pattern Returning pointers to enclave memory (P2) and Pointers to Overlapping Memory (P3).

product-mini/platforms/linux-sgx/enclave-sample/Enclave/Enclave.edl
LeoneChen: > maybe use macro `#if WASM_ENABLE_STATIC_PGO != 0` for those two functions?

Since EDL file is processed by sgx_edger8r, it can't pass like `-DWASM_ENABLE_STATIC_PGO` to preprocess of this file, so when we add `#if WASM_ENABLE_STATIC_PGO != 0` to EDL file, we must add `#define WASM_ENABLE_STATIC_PGO 1 or 0` in EDL file. So I think maybe we can always exposure this two interfaces and add macro check in Enclave.cpp, when `WASM_ENABLE_STATIC_PGO == 0`, keep this two functions empty.

TianlongLiang:
Thanks! That makes sense

LeoneChen:
> Hi, I see that there some CI tests are failing, seems not able to call ecall_handle_cmd_exec_app_func()?

Do you means CI tests successfully call `ecall_handle_cmd_exec_app_func`, but crash or return error inside this function? Or CI tests can't successfully call `ecall_handle_cmd_exec_app_func`？If it's the first situation, maybe the check added by myself in `ecall_handle_cmd_exec_app_func` result in checking fault

product-mini/platforms/linux-sgx/enclave-sample/Enclave/Enclave.edl
TianlongLiang: I guess that would be okay since the command line options are controlled by macro, wouldn't call those two functions anyway when the macro is disabled.

LeoneChen:
> Thanks! That makes sense

This part of code maybe can be optimized. I think it's better to avoid using lock

LeoneChen:
> > Hi, I see that there some CI tests are failing, seems not able to call ecall_handle_cmd_exec_app_func()?
> 
> Do you means CI tests successfully call `ecall_handle_cmd_exec_app_func`, but crash or return error inside this function? Or CI tests can't successfully call `ecall_handle_cmd_exec_app_func`？If it's the first situation, maybe the check added by myself in `ecall_handle_cmd_exec_app_func` result in checking fault

Can you provide more detail about it? Then I can figure out why it failed

TianlongLiang:
> Do you means CI tests successfully call `ecall_handle_cmd_exec_app_func`, but crash or return error inside this function? Or CI tests can't successfully call `ecall_handle_cmd_exec_app_func`？If it's the first situation, maybe the check added by myself in `ecall_handle_cmd_exec_app_func` result in checking fault

https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/5548246017/job/15029123379?pr=2357

Some spec tests are failing, output is something like:

```shell
Starting interpreter for module '/tmp/tmp9j2olynn.aot'
Running: ../../../product-mini/platforms/linux-sgx/enclave-sample/iwasm --heap-size=0 -v=0 --repl /tmp/tmp9j2olynn.aot
Testing(return) i32_add  = 0x102:i32
THE FINAL EXCEPTION IS out should be in a form likes numbers:type, but Call ecall_handle_cmd_exec_app_func() failed.
```
You can run a hello world aot file with similar cmd option to reproduce it to double-check whether it's due to the newly added check

TianlongLiang:
> Can you provide more detail about it? Then I can figure out why it failed

On your local machine, you can also run spec tests to check:

```shell
# assume you install the SDK in /opt/intel/sgxsdk/
source /opt/intel/sgxsdk/environment
# in WAMR root directory
cd tests/wamr-test-suites
# the shell script will automatically compile needed component and run spec test on sgx iwasm
./test_wamr.sh -b -x -t aot -s spec -p -P
```

LeoneChen:
> > Do you means CI tests successfully call `ecall_handle_cmd_exec_app_func`, but crash or return error inside this function? Or CI tests can't successfully call `ecall_handle_cmd_exec_app_func`？If it's the first situation, maybe the check added by myself in `ecall_handle_cmd_exec_app_func` result in checking fault
> 
> https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/5548246017/job/15029123379?pr=2357
> 
> Some spec tests are failing, output is something like:
> 
> ```shell
> Starting interpreter for module '/tmp/tmp9j2olynn.aot'
> Running: ../../../product-mini/platforms/linux-sgx/enclave-sample/iwasm --heap-size=0 -v=0 --repl /tmp/tmp9j2olynn.aot
> Testing(return) i32_add  = 0x102:i32
> THE FINAL EXCEPTION IS out should be in a form likes numbers:type, but Call ecall_handle_cmd_exec_app_func() failed.
> ```
> 
> You can run a hello world aot file with similar cmd option to reproduce it to double-check whether it's due to the newly added check

Ok, I'll try aot mode. I currently is preparing thesis proposal, so maybe late to fix it.

LeoneChen:
> > Can you provide more detail about it? Then I can figure out why it failed
> 
> On your local machine, you can also run spec tests to check:
> 
> ```shell
> # assume you install the SDK in /opt/intel/sgxsdk/
> source /opt/intel/sgxsdk/environment
> # in WAMR root directory
> cd tests/wamr-test-suites
> # the shell script will automatically compile needed component and run spec test on sgx iwasm
> ./test_wamr.sh -b -x -t aot -s spec -p -P
> ```

Thanks

TianlongLiang:
No hurry, good luck with your thesis!

LeoneChen:
> No hurry, good luck with your thesis!

Thanks!

LeoneChen:
@TianlongLiang Hello, I fixed the parameters check in 1306d68ebb982754a4dd7e4612cf857b4b1b6dbf, and all CI tests passed

TianlongLiang:
LGTM

product-mini/platforms/linux-sgx/enclave-sample/Enclave/Enclave.cpp
```diff
@@ -85,25 +64,124 @@ static korp_mutex enclave_module_list_lock = OS_THREAD_MUTEX_INITIALIZER;
...
+    }
+
+    bool remove(uint16_t idx)
+    {
+        bool success = false;
+        pthread_rwlock_wrlock(&mRwlock);
+        if (mMap.erase(idx) == 1) {
+            success = true;
+        }
+        mCurrentIdx = idx;
```
wenyongh, line 106: How about changing to:
```C
        if (mMap.erase(idx) == 1) {
            success = true;
            if (idx < mCurrentIdx)
                mCurrentIdx = idx;
        }
```

product-mini/platforms/linux-sgx/enclave-sample/Enclave/Enclave.cpp
```diff
@@ -118,17 +196,18 @@ handle_cmd_init_runtime(uint64 *args, uint32 argc)
     /* initialize runtime environment */
     if (!wasm_runtime_full_init(&init_args)) {
         LOG_ERROR("Init runtime environment failed.\n");
-        args[0] = false;
-        return;
+        ret = false;
```
wenyongh, line 199: No need to set ret to false again since it is initialized as false.

product-mini/platforms/linux-sgx/enclave-sample/Enclave/Enclave.cpp
```diff
@@ -201,18 +280,19 @@ is_xip_file(const uint8 *buf, uint32 size)
...
-    uint32 wasm_file_size = *(uint32 *)args++;
-    char *error_buf = *(char **)args++;
-    uint32 error_buf_size = *(uint32 *)args++;
+    bool ret = false;
     uint64 total_size = sizeof(EnclaveModule) + (uint64)wasm_file_size;
     EnclaveModule *enclave_module;
 
-    bh_assert(argc == 4);
+    if (wasm_file == nullptr or enclave_module_idx == nullptr) {
+        ret = false;
```
wenyongh, line 293: Same as above

product-mini/platforms/linux-sgx/enclave-sample/Enclave/Enclave.cpp
```diff
@@ -254,11 +334,14 @@ handle_cmd_load_module(uint64 *args, uint32 argc)
...
-        *(void **)args_org = NULL;
-        return;
+        ret = false;
+        goto exit;
     }
 
-    *(EnclaveModule **)args_org = enclave_module;
+    if (!gEnclaveModuleMgr.add(enclave_module, enclave_module_idx)) {
+        ret = false;
+        goto exit;
```
wenyongh, line 343: Should destroy resources before exit, like:
```C
        wasm_runtime_unload(enclave_module->module);
        if (!enclave_module->is_xip_file)
            wasm_runtime_free(enclave_module);
        else
            os_munmap(enclave_module, (uint32)total_size);
```

product-mini/platforms/linux-sgx/enclave-sample/Enclave/Enclave.cpp
```diff
@@ -341,152 +439,223 @@ wasm_runtime_get_module_hash(wasm_module_t module)
...
-        *(void **)args_org = NULL;
-        return;
+        ret = false;
+        goto exit;
     }
 
-    *(wasm_module_inst_t *)args_org = module_inst;
+    if (!gWasmModuleInstMgr.add(module_inst, wasm_module_inst_idx)) {
+        ret = false;
+        goto exit;
```
wenyongh, line 468: Should deinstantiate module_inst before goto exit:
`wasm_runtime_deinstantiate(module_inst)`

product-mini/platforms/linux-sgx/enclave-sample/Enclave/Enclave.cpp
```diff
@@ -341,152 +439,223 @@ wasm_runtime_get_module_hash(wasm_module_t module)
...
+    if (!DeepCopyCStrArray(u_app_argv, app_argv, app_argc)) {
+        wasm_runtime_free(app_argv);
+        app_argv = nullptr;
+        ret = false;
+        goto exit;
     }
 
     wasm_application_execute_main(module_inst, app_argc - 1, app_argv + 1);
 
+    for (uint32 i = 0; i < app_argc; i++) {
```
wenyongh, line 563: The return value may be written back into app_argv[0..1], had better copy them back into u_app_argv

product-mini/platforms/linux-sgx/enclave-sample/Enclave/Enclave.cpp
LeoneChen: I keep `ret = false` to avoid further maintainer may initialize `ret` with other value and forget to modify here

LeoneChen: > No need to set ret to false again since it is initialized as false.

There are case developer forget to set it 
https://github.com/intel/ehsm/issues/265#issuecomment-1610803400

product-mini/platforms/linux-sgx/enclave-sample/Enclave/Enclave.cpp
LeoneChen: You're right, it's my fault

product-mini/platforms/linux-sgx/enclave-sample/Enclave/Enclave.cpp
wenyongh: OK, got it, thanks.

product-mini/platforms/linux-sgx/enclave-sample/Enclave/Enclave.cpp
LeoneChen: Oops, I didn't notice these. Have to check all problems like this

product-mini/platforms/linux-sgx/enclave-sample/Enclave/Enclave.cpp
LeoneChen: Yes

product-mini/platforms/linux-sgx/enclave-sample/Enclave/Enclave.cpp
LeoneChen: OK

LeoneChen: It need some time

LeoneChen:
CI tests failed again. What is nuttx. How can I reproduce this problem in my local machine

wenyongh:
> CI tests failed again. What is nuttx. How can I reproduce this problem in my local machine

Nuttx CI issue was fixed with PR #2414, could you rebase again?

LeoneChen:
> > CI tests failed again. What is nuttx. How can I reproduce this problem in my local machine
> 
> Nuttx CI issue was fixed with PR #2414, could you rebase again?

OK, I try it

product-mini/platforms/linux-sgx/enclave-sample/Enclave/Enclave.cpp
LeoneChen: > The return value may be written back into app_argv[0..1], had better copy them back into u_app_argv

Hello I want to confirm, it means 1. put the return value (`bool` type) into app_argv[0..1] or 2. content in `app_argv` can be modified by wasm_application_execute_main, and you want this modification propagates to outside of Enclave

wenyongh: Hi, I checked again, wasm_application_execute_main puts the return value into `*(int *)argv`, here should be `*(int *)(app_argv + 1)`, see:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/include/wasm_export.h#L836-L838

So my understanding is that we had better write`*(int *)(app_argv + 1)` back into `*(int *)u_app_argv`?

LeoneChen: I got it, so it's case 2, and I have to let the content in `*(int*)argv` propagates to outside Enclave

wenyongh:
@LeoneChen Many thanks for the fixing, it really improves the security of WAMR linux-sgx. We are preparing to release 1.2.3, since the changes of this PR are a little big and complex, we (and some customers) need to more time to test, I think we had better merge this PR after the new release is created, and currently we can create another PR to fix the issue of `wait_map = NULL;` in `wasm_shared_memory_destroy`. Is it OK to you?

LeoneChen:
> @LeoneChen Many thanks for the fixing, it really improves the security of WAMR linux-sgx. We are preparing to release 1.2.3, since the changes of this PR are a little big and complex, we (and some customers) need to more time to test, I think we had better merge this PR after the new release is created, and currently we can create another PR to fix the issue of `wait_map = NULL;` in `wasm_shared_memory_destroy`. Is it OK for you?

Yes, it's OK. Fix `wasm_shared_memory_destroy` is urgent.

This PR is complex and needs more review, and I also need to think about how to write back content to `u_app_argv`.

LeoneChen:
Fix it by yourself or I create a new PR?

wenyongh:
> Fix it by yourself or I create a new PR?

Let me fix it by myself😊

LeoneChen:
> Hi, I checked again, wasm_application_execute_main puts the return value into *(int *)argv, here should be *(int *)(app_argv + 1), see: https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/include/wasm_export.h#L836-L838
 So my understanding is that we had better write*(int *)(app_argv + 1) back into *(int *)u_app_argv?

`wasm_application_execute_main` call `execute_main`, which write back `int` type return value that use `char**` type `argv` as `int *`.

I think it's not a good code practice... 

And in current situation, `app_argv + 1` is malloc-ed in Enclave with memory clone (for security), when `execute_main` store `int` type return value to `app_argv + 1`, it will cause address of malloc-ed char string be overwrite, and then we can't free this deep copied char string.

Maybe can use stack memory

wenyongh:
> > Hi, I checked again, wasm_application_execute_main puts the return value into *(int *)argv, here should be *(int _)(app_argv + 1), see: https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/include/wasm_export.h#L836-L838
> > So my understanding is that we had better write_(int *)(app_argv + 1) back into *(int *)u_app_argv?
> 
> `wasm_application_execute_main` call `execute_main`, which write back `int` type return value that use `char**` type `argv` as `int *`.
> 
> I think it's not a good code practice...
> 
> And in current situation, `app_argv + 1` is malloc-ed in Enclave with memory clone (for security), when `execute_main` store `int` type return value to `app_argv + 1`, it will cause address of malloc-ed char string be overwrite, and then we can't free this deep copied char string.
> 
> Maybe can use stack memory

Yes, how about adding an extra argument in `ecall_handle_cmd_exec_app_main` to output the result? e.g.
```C
public bool ecall_handle_cmd_exec_app_main(
        uint16_t wasm_module_inst_idx,
        [in, count=app_argc] char **u_app_argv,
        uint32_t app_argc,
        [out, count=1] int *ret_main
      );
```

wenyongh:
> Fix it by yourself or I create a new PR?

@LeoneChen I found more similar issues in Enclave, and uploaded PR #2416 to fix them, could you help review it? Thanks.

LeoneChen:
> > Fix it by yourself or I create a new PR?
> 
> @LeoneChen I found more similar issues in Enclave, and uploaded PR #2416 to fix them, could you help review it? Thanks.

OK

lum1n0us:
@LeoneChen   We've created a [new PR](https://github.com/bytecodealliance/wasm-micro-runtime/pull/4828) to enhance the enclave. Please take a look and share your comments.

LeoneChen:
@lum1n0us OK, let me fuzz the patched version (https://github.com/bytecodealliance/wasm-micro-runtime/pull/4828) again.

lum1n0us:
Just checking in - do you still intend to rebase this PR to main or provide any updates? If there's no further activity, I'll close the PR.

LeoneChen:
No further activity.

Have a nice day!

PR #2361, opened 2023-07-14
Implement suspend flags as atomic variable

loganek:
We have observed a significant performance degradation after merging https://github.com/bytecodealliance/wasm-micro-runtime/pull/1991 Instead of protecting suspend flags with a mutex, we implement the flags as atomic variable and only use mutex when atomics are not available on a given platform.


loganek:
@wenyongh the build is passing now and the PR is ready for review. I'd appreciate your team's review. Many thanks

core/iwasm/common/wasm_suspend_flags.h
```diff
@@ -0,0 +1,79 @@
+/*
+ * Copyright (C) 2023 Amazon Inc.  All rights reserved.
```
yamt, line 2: * most of the logic in this file seems not specific to suspend flags. how about using a bit more generic name?
* i guess the default implementation should be c11 atomics.

loganek: Yes, initially I've had an implementation based on stdatomics, but there are two problems with that:
* The size of the atomic variable is not guaranteed to be the size of the base type - and this is a requirement for JIT. We could potentially workaround it with set of ifs on code generation phase though
* this file is included (indirectly) by some c++ files (fast jit) and removing the dependency would require major refactoring. C11 stdatomics are not available in c++.

For those reasons I decided to go for a simple solution for now, but happy to iterate on that.

nchamzn (approved):
Have profiled this change and performance is now comparable to before the locks were added.

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -1062,21 +1062,33 @@ wasm_interp_call_func_import(WASMModuleInstance *module_inst,
...
+            & WASM_SUSPEND_FLAG_TERMINATE) {                          \
+            /* terminate current thread */                            \
+            WASM_SUSPEND_FLAGS_UNLOCK(exec_env->wait_lock);           \
+            return;                                                   \
+        }                                                             \
+        while (WASM_SUSPEND_FLAGS_GET(exec_env->suspend_flags)        \
+               & WASM_SUSPEND_FLAG_SUSPEND) {                         \
+            /* suspend current thread */                              \
+            SUSPENSION_LOCK()                                         \
+            os_cond_wait(&exec_env->wait_cond, &exec_env->wait_lock); \
```
yamt, line 1088: check WASM_SUSPEND_FLAG_SUSPEND with the lock held.
otherwise wasm_cluster_resume_thread can fail to wake up the thread.

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -1214,7 +1219,8 @@ wasm_cluster_suspend_all_except_self(WASMCluster *cluster,
 void
 wasm_cluster_resume_thread(WASMExecEnv *exec_env)
 {
-    exec_env->suspend_flags.flags &= ~0x02;
+    WASM_SUSPEND_FLAGS_FETCH_AND(exec_env->suspend_flags,
+                                 ~WASM_SUSPEND_FLAG_SUSPEND);
     os_cond_signal(&exec_env->wait_cond);
```
yamt, line 1224: forgot to take wait_lock? (not a fault of this PR though)

core/iwasm/common/wasm_suspend_flags.h
```diff
@@ -0,0 +1,79 @@
...
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+
+#ifndef _WASM_SUSPEND_FLAGS_H
+#define _WASM_SUSPEND_FLAGS_H
+
+#include "bh_platform.h"
+#include "gnuc.h"
```
wenyongh, line 10: Could you include "gnuc.h" in "bh_platform.h" instead? We often only include "bh_platfrom.h" to include all the required APIs in `core/shared/platform` and `core/shared/utils`.

core/iwasm/common/wasm_suspend_flags.h
```diff
@@ -0,0 +1,79 @@
...
+#else /* else of WASM_UINT32_IS_ATOMIC */
+#define WASM_SUSPEND_FLAGS_IS_ATOMIC 0
+#endif /* WASM_UINT32_IS_ATOMIC */
+
+#endif
+
+#if WASM_SUSPEND_FLAGS_IS_ATOMIC != 0
+#define WASM_SUSPEND_FLAGS_LOCK(lock) \
+    do {                              \
+    } while (0)
```
wenyongh, line 66: How about `(void)0`?

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -1344,7 +1350,7 @@ wasm_cluster_is_thread_terminated(WASMExecEnv *exec_env)
 {
     os_mutex_lock(&exec_env->wait_lock);
     bool is_thread_terminated =
-        (exec_env->suspend_flags.flags & 0x01) ? true : false;
+        (WASM_SUSPEND_FLAGS_GET(exec_env->suspend_flags) & 0x01) ? true : false;
```
wenyongh, line 1353: Had better change `0x0` to `WASM_SUSPEND_FLAG_TERMINATE`

core/iwasm/interpreter/wasm_interp_classic.c
loganek: not updating it for now as per @wenyongh 's comment

core/iwasm/common/wasm_suspend_flags.h
loganek: done

core/iwasm/common/wasm_suspend_flags.h
loganek: done

core/iwasm/libraries/thread-mgr/thread_manager.c
loganek: done

core/iwasm/common/wasm_suspend_flags.h
yamt: how about "most of the logic in this file seems not specific to suspend flags. how about using a bit more generic name?"

loganek: The file right now includes the following:
* definitions of the suspend flag values (`WASM_SUSPEND_FLAG_*` definitions)
* `WASMSuspendFlags` data structure
* macros for operating on the suspend flag data structure (`WASM_SUSPEND_FLAGS_*` macros)
* information whether the suspend flag is atomic (`WASM_SUSPEND_FLAGS_IS_ATOMIC` definition)
* macros for locking the flag if it's not atomic (`WASM_SUSPEND_FLAGS_(UN)LOCK` macros)

They all seem pretty related to suspend flags.

yamt: availability check and wrappers on atomic builtins are generic.

yamt: https://github.com/bytecodealliance/wasm-micro-runtime/pull/2393

PR #2364, opened 2023-07-17
add initial stress test

Zzzabiyaka:
We need to make a test that runs longer than the tests we had before to check some problems that might happen after running for some time (e.g. memory corruption or something else)

### DIFF
--- a/core/iwasm/libraries/lib-wasi-threads/test/build.sh
+++ b/core/iwasm/libraries/lib-wasi-threads/test/build.sh
@@ -9,10 +9,13 @@ set -eo pipefail
 CC=${CC:=/opt/wasi-sdk/bin/clang}
 WAMR_DIR=../../../../..
 
+# Stress tests names
+thread_start_file_exclusions=("spawn_stress_test.wasm" "linear_memory_size_update.wasm")
+
 for test_c in *.c; do
     test_wasm="$(basename $test_c .c).wasm"
 
-    if [ $test_wasm = "linear_memory_size_update.wasm" ]; then
+    if [[ " ${thread_start_file_exclusions[@]} " =~ " ${test_wasm} " ]] ; then
         thread_start_file=""
     else
         thread_start_file=$WAMR_DIR/samples/wasi-threads/wasm-apps/wasi_thread_start.S
--- a/core/iwasm/libraries/lib-wasi-threads/test/spawn_stress_test.c
+++ b/core/iwasm/libraries/lib-wasi-threads/test/spawn_stress_test.c
@@ -1,0 +1,114 @@
+/*
+ * Copyright (C) 2023 Amazon.com Inc. or its affiliates. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+
+#ifndef __wasi__
+#error This example only compiles to WASM/WASI target
+#endif
+
+#include <assert.h>
+#include <errno.h>
+#include <math.h>
+#include <pthread.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+enum CONSTANTS {
+    NUM_ITER = 100000,
+    NUM_RETRY = 5,
+    MAX_NUM_THREADS = 8,
+};
+
+unsigned prime_numbers_count = 0;
+
+bool
+is_prime(unsigned int num)
+{
+    for (unsigned int i = 2; i <= (unsigned int)(sqrt(num)); ++i) {
+        if (num % i == 0) {
+            return false;
+        }
+    }
+
+    return true;
+}
+
+void *
+check_if_prime(void *value)
+{
+    unsigned int *num = (unsigned int *)(value);
+    usleep(10000);
+    if (is_prime(*num)) {
+        __atomic_fetch_add(&prime_numbers_count, 1, __ATOMIC_SEQ_CST);
+    }
+    return NULL;
+}
+
+unsigned int
+validate()
+{
+    unsigned int counter = 0;
+    for (unsigned int i = 2; i <= NUM_ITER; ++i) {
+        counter += is_prime(i);
+    }
+
+    return counter;
+}
+
+void
+spawn_thread(pthread_t *thread, unsigned int *arg)
+{
+    int status_code = -1;
+    for (int tries = 0; status_code != 0 && tries < NUM_RETRY; ++tries) {
+        status_code = pthread_create(thread, NULL, &check_if_prime, arg);
+        assert(status_code == 0 || status_code == EAGAIN);
+        if (status_code == EAGAIN) {
+            usleep(2000);
+        }
+    }
+
+    assert(status_code == 0 && "Thread creation should succeed");
+}
+
+int
+main(int argc, char **argv)
+{
+    pthread_t threads[MAX_NUM_THREADS];
+    unsigned int args[MAX_NUM_THREADS];
+    double percentage = 0.1;
+
+    for (unsigned int factorised_number = 2; factorised_number < NUM_ITER;
+         ++factorised_number) {
+        if (factorised_number > NUM_ITER * percentage) {
+            fprintf(stderr, "Stress test is %d%% finished\n",
+                    (unsigned int)(percentage * 100));
+            percentage += 0.1;
+        }
+
+        unsigned int thread_num = factorised_number % MAX_NUM_THREADS;
+        if (threads[thread_num] != 0) {
+            assert(pthread_join(threads[thread_num], NULL) == 0);
+        }
+
+        args[thread_num] = factorised_number;
+
+        usleep(2000);
+        spawn_thread(&threads[thread_num], &args[thread_num]);
+        assert(threads[thread_num] != 0);
+    }
+
+    for (int i = 0; i < MAX_NUM_THREADS; ++i) {
+        assert(threads[i] == 0 || pthread_join(threads[i], NULL) == 0);
+    }
+
+    // Check the test results
+    assert(
+        prime_numbers_count == validate()
+        && "Answer mismatch between tested code and reference implementation");
+
+    fprintf(stderr, "Stress test finished successfully\n");
+    return 0;
+}
--- a/tests/wamr-test-suites/wasi-test-script/run_wasi_tests.sh
+++ b/tests/wamr-test-suites/wasi-test-script/run_wasi_tests.sh
@@ -14,6 +14,7 @@ readonly WAMR_DIR="${WORK_DIR}/../../../.."
 readonly IWASM_CMD="${WORK_DIR}/../../../../product-mini/platforms/${PLATFORM}/build/iwasm \
     --allow-resolve=google-public-dns-a.google.com \
     --addr-pool=::1/128,127.0.0.1/32"
+readonly IWASM_CMD_STRESS="${IWASM_CMD} --max-threads=8"
 readonly WAMRC_CMD="${WORK_DIR}/../../../../wamr-compiler/build/wamrc"
 readonly C_TESTS="tests/c/testsuite/"
 readonly ASSEMBLYSCRIPT_TESTS="tests/assemblyscript/testsuite/"
@@ -24,6 +25,11 @@ readonly LIB_SOCKET_TESTS="${WAMR_DIR}/core/iwasm/libraries/lib-socket/test/"
 run_aot_tests () {
     local tests=("$@")
     for test_wasm in ${tests[@]}; do
+        local extra_stress_flags=""
+        if [[ "$test_wasm" =~ "stress" ]]; then
+            extra_stress_flags="--max-threads=8"
+        fi
+
         test_aot="${test_wasm%.wasm}.aot"
         test_json="${test_wasm%.wasm}.json"
 
@@ -41,7 +47,7 @@ run_aot_tests () {
             expected=$(jq .exit_code ${test_json})
         fi
 
-        ${IWASM_CMD} $test_aot
+        ${IWASM_CMD} $extra_stress_flags $test_aot
 
         ret=${PIPESTATUS[0]}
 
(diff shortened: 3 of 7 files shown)


core/iwasm/libraries/lib-wasi-threads/test/spawn_stress_test.c
```diff
@@ -0,0 +1,117 @@
...
+unsigned int g_count = 2;
+unsigned int prime_numbers = 0;
+pthread_mutex_t mutex;
+
+void
+__wasi_thread_start_C(int thread_id, int *start_arg)
+{
+    shared_t *data = (shared_t *)start_arg;
+
+    pthread_mutex_lock(&mutex);
```
loganek, line 46: What's the point of this mutex? Looks like there's only one thread running at a time.

core/iwasm/libraries/lib-wasi-threads/test/spawn_stress_test.c
```diff
@@ -0,0 +1,161 @@
...
+            assert(__builtin_wasm_memory_atomic_wait32(&threads[j].th_done, 0, TIMEOUT) != 2);
+                // Deinit the expired thread
+            {
+                start_args_deinit(&threads[j].base);
+                thread_init_delta--;
+            }
+            
+            // Invalidated the expired thread
+            {
+                threads[j].th_done = 0;
```
loganek, line 122: consider using `memset(&threads[j], 0, sizeof(threads[j]))` so if we ever add any new fields to the structure, they'll automatically be zero-ed. thread ID 0 is invalid too, no need to set it to -1.

core/iwasm/libraries/lib-wasi-threads/test/spawn_stress_test.c
```diff
@@ -0,0 +1,161 @@
...
+
+    return counter;
+}
+
+int
+main(int argc, char **argv)
+{
+    pthread_mutex_init(&mutex, NULL);
+
+    for (int i = 0; i < MAX_NUM_THREADS; ++i) {
```
loganek, line 87: Should this loop be moved inside the while loop to avoid dupplication?

loganek (changes requested):

core/iwasm/libraries/lib-wasi-threads/test/spawn_stress_test.c
```diff
@@ -0,0 +1,161 @@
...
+        pthread_mutex_lock(&mutex);
+
+        // Init a new thread
+        {
```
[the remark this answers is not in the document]
Zzzabiyaka: It was only to separate logical parts of the code. I don't mind removing it

core/iwasm/libraries/lib-wasi-threads/test/spawn_stress_test.c
```diff
@@ -0,0 +1,117 @@
...
+#endif
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <assert.h>
+#include <stdbool.h>
+#include <unistd.h>
+#include <math.h>
+
+#if USE_CUSTOM_SYNC_PRIMITIVES != 0
```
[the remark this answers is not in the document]
Zzzabiyaka: Thanks

core/iwasm/libraries/lib-wasi-threads/test/spawn_stress_test.c
Zzzabiyaka: I'd say they're a bit different anyway. 

e.g. here we don't need to cleanup the resources, wait for the thread, etc. 
I think moving it inside the next loop might introduce even more complexity

core/iwasm/libraries/lib-wasi-threads/test/build.sh
```diff
@@ -9,6 +9,36 @@ set -eo pipefail
...
+            ;;
+        *)
+            display_usage
+            exit 1
+            ;;
+    esac
+    shift
+done
+
+# Stress tests names (to include only in nightly build)
```
loganek, line 39: why is that? Are they taking too long to be executed as part of regular builds?

core/iwasm/libraries/lib-wasi-threads/test/build.sh
```diff
@@ -18,18 +48,22 @@ for test_c in *.c; do
...
-        -Wl,--export=malloc \
-        -Wl,--export=free \
-        -I $WAMR_DIR/samples/wasi-threads/wasm-apps \
-        $thread_start_file \
-        $test_c -o $test_wasm
+    if [[   "$enable_stress_tests" = true   ]] || ! [[ " ${stress_tests[@]} " =~ " ${test_wasm} " ]]; then
+        echo "Compiling $test_c to $test_wasm"
+        $CC \
+            -target wasm32-wasi-threads \
+            -pthread -ftls-model=local-exec \
```
loganek, line 55: is `-ftls-model=local-exec` needed here?

core/iwasm/libraries/lib-wasi-threads/test/build.sh
```diff
@@ -9,6 +9,36 @@ set -eo pipefail
...
+function display_usage() {
+    echo "Usage: $0 [--enable_stress_tests]"
+    echo "Options:"
+    echo "  --enable_stress_tests   Enable stress tests"
+}
+
+# Parse command line arguments
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        --enable_stress_tests)
```
loganek, line 28: Why do we need it? I see you already exclude tests by defining them in `skip.json` file so it's probably fine to compile them unconditionally?

loganek (changes requested):

core/iwasm/libraries/lib-wasi-threads/test/build.sh
Zzzabiyaka: it was not written by me and works like that for quite some time so I can't really say now. 

I will check with the author, but it shouldn't be related to this PR as this part is executed for all the existing tests

core/iwasm/libraries/lib-wasi-threads/test/build.sh
Zzzabiyaka: for now they are not but I was expecting that if we plan to add any new stress-tests later we won't include them into the regular build. 

How do you think, should this test now be a part of regular build?

core/iwasm/libraries/lib-wasi-threads/test/build.sh
loganek: I think we can control tests that are/aren't included using wasi tests' filters

core/iwasm/libraries/lib-wasi-threads/test/build.sh
loganek: I'd keep them in the regular build until it's taking too long time and becomes a problem.

core/iwasm/libraries/lib-wasi-threads/test/spawn_stress_test.c
```diff
@@ -0,0 +1,78 @@
...
+    NUM_RETRY = 5,
+    SECOND = 1000 * 1000 * 1000, /* 1 second */
+    TIMEOUT = 50LL * SECOND,
+    MAX_NUM_THREADS = 8,
+};
+
+unsigned int factorised_number = 2;
+unsigned int prime_numbers = 0;
+double percentage = 0.1;
+pthread_t threads[MAX_NUM_THREADS];
```
loganek, line 31: this probably doesn't have to be global

core/iwasm/libraries/lib-wasi-threads/test/spawn_stress_test.c
```diff
@@ -0,0 +1,78 @@
...
+            return false;
+        }
+    }
+
+    return true;
+}
+
+void* check_if_prime(void* value) {
+    unsigned int* num = (unsigned int*)(value);
+    __builtin_wasm_memory_atomic_wait32(NULL, 0, 0.01 * SECOND);
```
loganek, line 46: consider using `usleep()`

core/iwasm/libraries/lib-wasi-threads/test/spawn_stress_test.c
```diff
@@ -0,0 +1,78 @@
...
+{
+    for (; factorised_number < NUM_ITER; ++factorised_number) {
+        unsigned int thread_num = factorised_number % MAX_NUM_THREADS;
+        args[thread_num] = factorised_number;
+        if (threads[thread_num] != 0) {
+            assert (pthread_join(threads[thread_num], NULL) == 0);
+        }
+        assert (pthread_create(&threads[thread_num], NULL, &check_if_prime, &args[thread_num]) == 0 && "Thread creation should succeed");
+    }
+
```
loganek, line 73: you should probably `join` here on the threads in progress?

core/iwasm/libraries/lib-wasi-threads/test/spawn_stress_test.c
```diff
@@ -0,0 +1,78 @@
...
+    return counter;
+}
+
+int
+main(int argc, char **argv)
+{
+    for (; factorised_number < NUM_ITER; ++factorised_number) {
+        unsigned int thread_num = factorised_number % MAX_NUM_THREADS;
+        args[thread_num] = factorised_number;
```
loganek, line 67: shouldn't that assignment happen after `join`? Otherwise you're overriding a memory used by already running thread.

core/iwasm/libraries/lib-wasi-threads/test/spawn_stress_test.c
```diff
@@ -0,0 +1,78 @@
...
+#ifndef __wasi__
+#error This example only compiles to WASM/WASI target
+#endif
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <assert.h>
+#include <stdbool.h>
+#include <unistd.h>
+#include <math.h>
```
loganek, line 15: what's this include for?

loganek (changes requested):

core/iwasm/libraries/lib-wasi-threads/test/spawn_stress_test.c
```diff
@@ -0,0 +1,78 @@
...
+        counter += is_prime(i);
+    }
+
+    return counter;
+}
+
+int
+main(int argc, char **argv)
+{
+    for (; factorised_number < NUM_ITER; ++factorised_number) {
```
loganek, line 65: Just a note - if you try to maximize the number of threads running all the times, this code does not do that. Imagine situation where you have 3 threads (1, 2, 3) and for some reason 1 is really slow, but 2 and 3 are fast. Once 1, 2, and 3 are spawned, and then 2 and 3 complete, they'll have to wait for thread 1 to complete before 2 and 3 is spawned again. If you want to achieve maximum concurrency at most possible times, you should probably just maintain a list of thread ids and keep track on the number of active threads.

core/iwasm/libraries/lib-wasi-threads/test/spawn_stress_test.c
```diff
@@ -0,0 +1,78 @@
...
+        unsigned int thread_num = factorised_number % MAX_NUM_THREADS;
+        args[thread_num] = factorised_number;
+        if (threads[thread_num] != 0) {
+            assert (pthread_join(threads[thread_num], NULL) == 0);
+        }
+        assert (pthread_create(&threads[thread_num], NULL, &check_if_prime, &args[thread_num]) == 0 && "Thread creation should succeed");
```
loganek, line 71: Because both `pthread_create` and `pthread_join` are implemented in userspace, it's possible that a new native thread will start before the other finishes. I'd suggest adding retry mechanism for `pthread_create` and try a few times on the `EAGAIN` error (with some interval) before failing on assertion.

Zzzabiyaka: yeah, that makes sense. Thank you

core/iwasm/libraries/lib-wasi-threads/test/spawn_stress_test.c
Zzzabiyaka: sqrt()

core/iwasm/libraries/lib-wasi-threads/test/spawn_stress_test.c
```diff
@@ -0,0 +1,108 @@
+/*
+ * Copyright (C) 2023 Amazon.com Inc. or its affiliates. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+
+#ifndef __wasi__
+#error This example only compiles to WASM/WASI target
```
loganek, line 7: are we using any specific WASM/WASI instruction here?

core/iwasm/libraries/lib-wasi-threads/test/spawn_stress_test.c
```diff
@@ -0,0 +1,108 @@
...
+            percentage += 0.1;
+        }
+
+        unsigned int thread_num = factorised_number % MAX_NUM_THREADS;
+        if (threads[thread_num] != 0) {
+            assert(pthread_join(threads[thread_num], NULL) == 0);
+        }
+
+        args[thread_num] = factorised_number;
+        threads[thread_num] = 0;
```
loganek, line 82: this seems unnecessary

core/iwasm/libraries/lib-wasi-threads/test/spawn_stress_test.c
```diff
@@ -0,0 +1,108 @@
...
+                usleep(10000);
+            }
+        }
+
+        assert(status_code == 0 && "Thread creation should succeed");
+        assert(threads[thread_num] != 0);
+    }
+
+    for (int i = 0; i < MAX_NUM_THREADS; ++i) {
+        assert(pthread_join(threads[i], NULL) == 0);
```
loganek, line 100: probably worth checking if `threads[i]` is not 0 before the `pthread_join()`

Zzzabiyaka: they're guaranteed not to be if ITER_NUM is bigger than MAX_NUM_THREAD because of line 96, aren't they?

core/iwasm/libraries/lib-wasi-threads/test/spawn_stress_test.c
Zzzabiyaka: looks like we're not but I'm not sure we should allow anyone running this test without WASI define, should we? 

I mean, I checked this code only with WASI sdk, where this is defined. Without it, you won't be able to run `build.sh` etc

core/iwasm/libraries/lib-wasi-threads/test/spawn_stress_test.c
Zzzabiyaka: But I'll add this check for visibility

loganek: A simple example - if you set MAX_NUM_THREAD to 8 and ITER_NUM to 1, then most of the thread ids will be 0.

core/iwasm/libraries/lib-wasi-threads/test/build.sh
```diff
@@ -9,10 +9,13 @@ set -eo pipefail
 CC=${CC:=/opt/wasi-sdk/bin/clang}
 WAMR_DIR=../../../../..
 
+# Stress tests names
+stress_tests=("spawn_stress_test.wasm")
```
loganek, line 13: is it only used to decide whether the `thread_start_file` is needed or not?  If so, maybe worth renaming this array and include `linear_memory_size_update` in it.

tests/wamr-test-suites/wasi-test-script/run_wasi_tests.sh
```diff
@@ -53,14 +59,22 @@ run_aot_tests () {
 if [[ $MODE != "aot" ]];then
     python3 -m venv wasi-env && source wasi-env/bin/activate
     python3 -m pip install -r test-runner/requirements.txt
+
+    # Stress test requires max-threads=8 so it's run separately
+    if [[ -e "${THREAD_INTERNAL_TESTS}spawn_stress_test.wasm" ]]; then 
+        ${IWASM_CMD_STRESS} ${THREAD_INTERNAL_TESTS}spawn_stress_test.wasm
```
loganek, line 65: How do we check if the test failed? We don't seem to check the exit code.

loganek (changes requested):

core/iwasm/libraries/lib-wasi-threads/test/spawn_stress_test.c
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    return true;
+}
+
+void *
+check_if_prime(void *value)
+{
+    unsigned int *num = (unsigned int *)(value);
+    usleep(10000);
```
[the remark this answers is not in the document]
Zzzabiyaka: Yeah I also feel like another test is good for it

tests/wamr-test-suites/wasi-test-script/run_wasi_tests.sh
Zzzabiyaka: Thank you. That's a good point

Zzzabiyaka:
Thanks Marcin!

It was really a long but useful review for me!

Zzzabiyaka:
This PR is currently blocked on #2389

loganek (approved):
LGTM, thanks!

PR #2374, opened 2023-07-20
Update Docker image for latest version of external libraries & tools

cngzhnp:
Devcontainer seems to old to catch up the latest versions of dependencies. So, I updated the libraries & tools if WASM environment requires for Docker image.

**Impediments:**
- LLVM 16 could not be retrieved directly by using shell command that was provided before. So, I need to upgrade all tools manually but it is workaround and I just put LLVM issue URL as a comment in the script. 

**Improvements:**
- Use `Debian 12 (Bookworm)` as a base image instead of `Ubuntu 20.04`.
- `GCC9` upgraded to `GCC12`.
- `LLVM14` upgraded to `LLVM16`.
- `Clang10` upgraded to `Clang10`.
- `Binaryen111` upgraded to `Binaryen114`.
- `WASI-19` upgraded to `WASI-20`
- `WABT-1.0.29` upgraded to `WABT-1.0.33`
- `Bazelisk-1.12.0` upgraded to `Bazelisk-1.17.0`
- `GithubCLI-2.20.2` upgraded to `GithubCLI-2.32.0`
- `NodeJS-19.x` upgraded to `NodeJS-20.x`
- `EMSDK-3.0.0` upgraded to `EMSDK-3.1.43`

**Notes:**
`Python 2.7` was removed due to no support anymore and not found in the repository.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -10,8 +10,8 @@ ARG DEBIAN_FRONTEND=noninteractive
 ENV TZ=Asian/Shanghai
 
 # hadolint ignore=DL3008
-RUN apt-get update
-RUN apt upgrade -y \
+RUN apt-get update \
+  && apt-get upgrade -y \
   && apt-get install -y apt-transport-https apt-utils build-essential \
   ca-certificates ccache cmake curl g++-multilib git gnupg \
   libgcc-12-dev lib32gcc-12-dev lsb-release \
@@ -93,15 +93,18 @@ WORKDIR /tmp
 #  && ./llvm.sh ${LLVM_VER} all
 
 # Workaround due to https://github.com/llvm/llvm-project/issues/62475
+# hadolint ignore=DL3008
 RUN set -ex \
     && echo "deb http://apt.llvm.org/bookworm/ llvm-toolchain-bookworm-${LLVM_VER} main" > /etc/apt/sources.list.d/apt.llvm.org.list \
     && wget -qO- https://apt.llvm.org/llvm-snapshot.gpg.key | tee /etc/apt/trusted.gpg.d/apt.llvm.org.asc \
-    && apt update \
+    && apt-get update \
     && apt-get install -y \
     clang-${LLVM_VER} lldb-${LLVM_VER} lld-${LLVM_VER} clangd-${LLVM_VER} clang-tidy-${LLVM_VER} clang-format-${LLVM_VER} clang-tools-${LLVM_VER} \
     llvm-${LLVM_VER}-dev lld-${LLVM_VER} lldb-${LLVM_VER} llvm-${LLVM_VER}-tools libomp-${LLVM_VER}-dev libc++-${LLVM_VER}-dev libc++abi-${LLVM_VER}-dev \
     libclang-common-${LLVM_VER}-dev libclang-${LLVM_VER}-dev libclang-cpp${LLVM_VER}-dev libunwind-${LLVM_VER}-dev \
-    libclang-rt-${LLVM_VER}-dev libpolly-${LLVM_VER}-dev
+    libclang-rt-${LLVM_VER}-dev libpolly-${LLVM_VER}-dev --no-install-recommends \
+    && apt-get clean -y \
+    && rm -rf /var/lib/apt/lists/*
 
 #
 # [Optional]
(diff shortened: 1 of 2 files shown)


.devcontainer/Dockerfile
```diff
@@ -3,44 +3,45 @@
 
 # See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.195.0/containers/cpp/.devcontainer/base.Dockerfile
 # [Choice] Debian / Ubuntu version (use Debian 11/9, Ubuntu 18.04/21.04 on local arm64/Apple Silicon): debian-11, debian-10, debian-9, ubuntu-21.04, ubuntu-20.04, ubuntu-18.04
```
wenyongh, line 5: Could we also update the comment to mention Debian 12?

.devcontainer/Dockerfile
```diff
@@ -3,44 +3,45 @@
...
-ARG VARIANT=ubuntu-20.04
-FROM mcr.microsoft.com/vscode/devcontainers/cpp:0-${VARIANT}
+ARG VARIANT=debian-12
```
wenyongh, line 6: Not sure whether it is a good practice to change ubuntu to debian. Some developers may like to use ubuntu? Could we add another file for ubuntu, e.g., Dockerfile.ubuntu?

.devcontainer/Dockerfile
cngzhnp: Yes, I could update it too, thanks for the warning.

lum1n0us:
Please make sure to pass *hadolint* check.

TianlongLiang:
Hi, I run hadolint check on the Dockerfile for you, there are some warnings, you can check too by first pulling the docker image of hadolint, then running the check in the `.devcontainer` directory:

```shell
# pull docker images
docker pull hadolint/hadolint:latest
# hadolint check
git ls-files --exclude="*Dockerfile*" --cached --ignored | xargs -I filename zsh -c "echo filename; docker run --rm -i hadolint/hadolint:2.10.0 < filename; echo \"\n\""
# output
-:13 DL3009 info: Delete the apt-get lists after installing something
-:14 DL3008 warning: Pin versions in apt get install. Instead of `apt-get install <package>` use `apt-get install <package>=<version>`
-:14 DL3027 warning: Do not use apt as it is meant to be a end-user tool, use apt-get or apt-cache instead
-:96 DL3008 warning: Pin versions in apt get install. Instead of `apt-get install <package>` use `apt-get install <package>=<version>`
-:96 DL3027 warning: Do not use apt as it is meant to be a end-user tool, use apt-get or apt-cache instead
-:96 DL3015 info: Avoid additional packages by specifying `--no-install-recommends`
```

cngzhnp:
> Hi, I run hadolint check on the Dockerfile for you, there are some warnings, you can check too by first pulling the docker image of hadolint, then running the check in the `.devcontainer` directory:
> 
> ```shell
> # pull docker images
> docker pull hadolint/hadolint:latest
> # hadolint check
> git ls-files --exclude="*Dockerfile*" --cached --ignored | xargs -I filename zsh -c "echo filename; docker run --rm -i hadolint/hadolint:2.10.0 < filename; echo \"\n\""
> # output
> -:13 DL3009 info: Delete the apt-get lists after installing something
> -:14 DL3008 warning: Pin versions in apt get install. Instead of `apt-get install <package>` use `apt-get install <package>=<version>`
> -:14 DL3027 warning: Do not use apt as it is meant to be a end-user tool, use apt-get or apt-cache instead
> -:96 DL3008 warning: Pin versions in apt get install. Instead of `apt-get install <package>` use `apt-get install <package>=<version>`
> -:96 DL3027 warning: Do not use apt as it is meant to be a end-user tool, use apt-get or apt-cache instead
> -:96 DL3015 info: Avoid additional packages by specifying `--no-install-recommends`
> ```

Thanks for the output, I fixed Dockerfile linter warnings/infos. I may suggest that maybe it would be more helpful to run Hadolint check on Github actions when someone has changed one of Dockerfiles in the repository. Because I was not aware of that linter was run before changing anything on Dockerfile.

TianlongLiang:
Yes, a CI for hadolint linter to scan over Dockerfiles is definitely worth consideration, I will be working on it

.devcontainer/Dockerfile
wenyongh: OK, got it, thanks.

TianlongLiang:
LGTM

PR #2380, opened 2023-07-22
Add more fixes for wasi libc on windows

loganek:
* disable translations of errno codes that aren't defined on Windows
* undef `min()` macro if it is defined to not conflict with the `min()` function we define
* implement `shed_yield` wasi call
* disable some of the features in the config for windows by default


wenyongh (approved):
LGTM

PR #2386, opened 2023-07-24
Temporarily disable code that does not compile on windows

loganek:
and return ENOSYS, so wasi-libc. We do that so we can at least compile the code on CI. We'll be gradually enabling more and more functions.

Also, enabled `proc_raise()` for windows


wenyongh (approved):
LGTM

Issue #2390, opened 2023-07-25
A crash invoked by the instruction `table.fill`


erxiaozhou:
## Test case
[table.fill_247.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/12162621/table.fill_247.zip)

## Build commands
I compile the code with commit id 1cafa375680b6ea8b120e0d4d6006521d858ae2d.
Platform: Ubuntu 20.04
CPU: amd64 

## compile:
```shell
export CC=/usr/bin/clang-14
export CXX=/usr/bin/clang++-14
cd product-mini/platforms/linux/;rm -rf build
cmake  -DWAMR_BUILD_INTERP=1 -DWAMR_BUILD_REF_TYPES=1 -DWASM_ENABLE_SIMD=1 -DWASM_ENABLE_BULK_MEMORY=1 -DWAMR_BUILD_AOT=0 -DCMAKE_BUILD_TYPE=Release -DWAMR_BUILD_FAST_INTERP=0 -DWAMR_BUILD_LIBC_WASI=0 -DWAMR_BUILD_LIBC_BUILTIN=1 -Bbuild
cd build
make -j40
```
## execute:
```shell
iwasm --heap-size=0 -f to_test <tc_name>
```

## Behavior:
Crash 
```
Segmentation fault (core dumped)
```

## Addition
I also compiled a WAMR-classic-interpreter with commit id 6820af621265413b677e450cf126bb66c9f77077 in debug mode.
The build commands are:
```shell
cmake  -DWAMR_BUILD_INTERP=1 -DWAMR_BUILD_REF_TYPES=1 -DWAMR_BUILD_AOT=0 -DCMAKE_BUILD_TYPE=Debug -DWAMR_BUILD_FAST_INTERP=0 -Bbuild
```

The exception information is
```
UndefinedBehaviorSanitizer:DEADLYSIGNAL
==2197185==ERROR: UndefinedBehaviorSanitizer: SEGV on unknown address 0x55ca8b292d3c (pc 0x55c888d32051 bp 0x7ffe5475b490 sp 0x7ffe54752b10 T2197185)
==2197185==The signal is caused by a WRITE memory access.
    #0 0x55c888d32051  (/home/std_runtime_test/dump_iwasm_interp_classic/product-mini/platforms/linux/build/iwasm+0x110051)
[...]

Issue #2096, opened 2023-04-02
Unexpected execution


erxiaozhou:
# Build commands
I compile the code with commit id aaf671d688592784d1f8e000531cd98c2c22bfac.
Platform: Ubuntu 20.04
CPU: amd64 
compile:
```shell
export CC=/usr/bin/clang-14
export CXX=/usr/bin/clang++-14
cd product-mini/platforms/linux/;rm -rf build
cmake  -DWAMR_BUILD_INTERP=1 -DWAMR_BUILD_REF_TYPES=1 -DWAMR_BUILD_AOT=0 -DCMAKE_BUILD_TYPE=Debug -DWAMR_BUILD_FAST_INTERP=1 -Bbuild
cd build
make -j40
```
execute:
```shell
iwasm --heap-size=0 -f to_test <tc_name>
```

[iwasm_table_OOB.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/11132202/iwasm_table_OOB.zip)

# Actual behavior:
No exception

# Expected behavior:
An exception indicating that "out of bounds table access"

wenyongh:
Hi, the opcode `table.init` in this case takes the operand n, src, dst with 0, 0, -2147483648, though the destination is "out of bounds", the size to copy is 0, per my understanding, it should be ok to just ignore the initialization.

erxiaozhou:
> Hi, the opcode `table.init` in this case takes the operand n, src, dst with 0, 0, -2147483648, though the destination is "out of bounds", the size to copy is 0, per my understanding, it should be ok to just ignore the initialization.

Thank you for your nice reply!

erxiaozhou:
> Hi, the opcode `table.init` in this case takes the operand n, src, dst with 0, 0, -2147483648, though the destination is "out of bounds", the size to copy is 0, per my understanding, it should be ok to just ignore the initialization.

Hi~.   Maybe it is better to fix this "bug". 
According to the specification, there should be a trap, (lines 16, 17; the runtime should first check whether d+n greater than the length of the table and then check whether n equals 0).  Furthermore, I have tried this test case on other runtimes and they all failed due to "Table OOB".  So I think fixing this behavior might be a better option. Thank you~

![image](https://user-images.githubusercontent.com/32102519/232321888-5f2cbc87-47c2-4e65-8c5a-584b5f6b7d97.png)

yamt:
i agree it should trap.
[...]

PR #2392, opened 2023-07-26
Fix some check issues on table operations

wenyongh:
Fix some check issues on table.init, table.fill and table.copy, and unify the check method
for all running modes.
Fix issue #2390 and #2096.


yamt (commented):
lgtm.

i guess it's better to have a function or a macro like the following
rather than repeating the mostly same logic.
```
bool limit_check(uint32 off, uint32 len, uint32 lim);
```

wenyongh:
> lgtm.
> 
> i guess it's better to have a function or a macro like the following rather than repeating the mostly same logic.
> 
> ```
> bool limit_check(uint32 off, uint32 len, uint32 lim);
> ```

Good point, thanks, I defined an inline function `offset_len_out_of_bounds` to refine the code.

yamt (approved):
lgtm

PR #2393, opened 2023-07-26
move generic parts of wasm_suspend_flags.h to bh_atomic.h

yamt:


loganek:
I'd expect bh_atomic to use stdatomics when possible; suspend_flag is an exception and should not use stdatomics (due to reasons explained in the linked PR). So my preference would be to keep those two separate, even if that means having some code duplicated (unless we do some ifdef's magic to only use builtin atomics in suspend_flag, but then I'm not sure if the complexity of code is worth the benefit we'll get).

yamt:
> I'd expect bh_atomic to use stdatomics when possible; suspend_flag is an exception and should not use stdatomics (due to reasons explained in the linked PR). So my preference would be to keep those two separate, even if that means having some code duplicated (unless we do some ifdef's magic to only use builtin atomics in suspend_flag, but then I'm not sure if the complexity of code is worth the benefit we'll get).

given how our aot compilation works, many of our runtime structures are
very sensitive to size/offset changes. it isn't only suspend_flags for which the "fixed-sized"
property can be convenient.

core/shared/utils/bh_atomic.h
```diff
@@ -0,0 +1,53 @@
...
+#define CLANG_GCC_HAS_ATOMIC_BUILTIN
+#endif
+#elif defined(__clang__)
+#if __clang_major__ > 3 || (__clang_major__ == 3 && __clang_minor__ >= 0)
+#define CLANG_GCC_HAS_ATOMIC_BUILTIN
+#endif
+#endif
+
+#if defined(CLANG_GCC_HAS_ATOMIC_BUILTIN)
+#define BH_ATOMIC_32_IS_ATOMIC 1
```
wenyongh, line 28: Using `BH_ATOMIC_32_IS_ATOMIC` is a little confusing, how about renaming it to `BH_UINT32_IS_ATOMIC`?

core/shared/utils/bh_atomic.h
```diff
@@ -0,0 +1,53 @@
...
+    __atomic_fetch_and(&(v), (val), __ATOMIC_SEQ_CST)
+#else /* else of defined(CLANG_GCC_HAS_ATOMIC_BUILTIN) */
+#define BH_ATOMIC_32_LOAD(v) (v)
+#define BH_ATOMIC_32_FETCH_OR(v, val) ((v) |= (val))
+#define BH_ATOMIC_32_FETCH_AND(v, val) ((v) &= (val))
+
+/* The flag can be defined by the user if the platform
+   supports atomic access to uint32 aligned memory. */
+#ifdef WASM_UINT32_IS_ATOMIC
+#define BH_ATOMIC_32_IS_ATOMIC 1
```
wenyongh, line 42: Same as above, how about renaming it to `BH_UINT32_IS_ATOMIC`?

core/shared/utils/bh_atomic.h
wenyongh: PS: It is good if it refers to type `bh_atomic_32_t`.

yamt: `BH_ATOMIC_32_IS_ATOMIC` means if `bh_atomic_32_t` is really atomic.
that is, if it can be accessed atomically with macros like `BH_ATOMIC_32_FETCH_OR` without extra lock.

wenyongh: OK, got it.

wenyongh (approved):
LGTM

PR #2398, opened 2023-07-28
bh_atomic.h: add comments

yamt:


core/shared/utils/bh_atomic.h
```diff
@@ -12,6 +12,38 @@
...
+ * - It's available for C++.
+ *   Although C++23 will have C-compatible stdatomics.h, it isn't widely
+ *   available yet.
+ */
+
+/*
+ * Note about BH_ATOMIC_32_IS_ATOMIC
+ *
+ * If BH_ATOMIC_32_IS_ATOMIC == 0, BH_ATOMIC_xxx operations defined below
+ * are not really atomic and requires an external lock.
```
wenyongh, line 32: Had better change `requires` to `require`

yamt: fixed. thank you.

PR #2411, opened 2023-07-31
add another wamr test

Zzzabiyaka:
follows up #2364 where we discussed that we might want to have a test which has really short thread function and creates many threads

### DIFF
--- a/.github/workflows/compilation_on_android_ubuntu.yml
+++ b/.github/workflows/compilation_on_android_ubuntu.yml
@@ -537,7 +537,7 @@ jobs:
         working-directory: ./core/iwasm/libraries/lib-socket/test/
 
       - name: run tests
-        timeout-minutes: 20
+        timeout-minutes: 30
         run: ./test_wamr.sh ${{ matrix.test_option }} -t ${{ matrix.running_mode }}
         working-directory: ./tests/wamr-test-suites
 
@@ -553,7 +553,7 @@ jobs:
           sudo apt install -y g++-multilib lib32gcc-9-dev
 
       - name: run tests x86_32
-        timeout-minutes: 20
+        timeout-minutes: 30
         if: env.TEST_ON_X86_32 == 'true'
         run: ./test_wamr.sh ${{ env.X86_32_TARGET_TEST_OPTIONS }} ${{ matrix.test_option }} -t ${{ matrix.running_mode }}
         working-directory: ./tests/wamr-test-suites
--- a/core/iwasm/libraries/lib-wasi-threads/test/build.sh
+++ b/core/iwasm/libraries/lib-wasi-threads/test/build.sh
@@ -34,7 +34,7 @@ while [[ $# -gt 0 ]]; do
 done
 
 # Stress tests names
-thread_start_file_exclusions=("spawn_stress_test.wasm" "linear_memory_size_update.wasm")
+thread_start_file_exclusions=("spawn_stress_test.wasm" "linear_memory_size_update.wasm" "stress_test_threads_creation.wasm")
 
 for test_c in *.c; do
     test_wasm="$(basename $test_c .c).wasm"
@@ -56,6 +56,7 @@ for test_c in *.c; do
     echo "Compiling $test_c to $test_wasm"
     $CC \
         -target wasm32-wasi-threads \
+        -O2 \
         -pthread -ftls-model=local-exec \
         -z stack-size=32768 \
         -Wl,--export=__heap_base \
--- a/core/iwasm/libraries/lib-wasi-threads/test/skip.json
+++ b/core/iwasm/libraries/lib-wasi-threads/test/skip.json
@@ -1,5 +1,6 @@
 {
     "lib-wasi-threads tests": {
-        "spawn_stress_test": "Stress tests are incompatible with the other part and executed differently"
+        "spawn_stress_test": "Stress tests are incompatible with the other part and executed differently",
+        "stress_test_threads_creation": "Stress tests are incompatible with the other part and executed differently"
     }
 }
--- a/core/iwasm/libraries/lib-wasi-threads/test/spawn_stress_test.c
+++ b/core/iwasm/libraries/lib-wasi-threads/test/spawn_stress_test.c
@@ -18,8 +18,9 @@
 
 enum CONSTANTS {
     NUM_ITER = 100000,
-    NUM_RETRY = 5,
+    NUM_RETRY = 8,
     MAX_NUM_THREADS = 8,
+    RETRY_SLEEP_TIME_US = 2000,
 };
 
 unsigned prime_numbers_count = 0;
@@ -62,11 +63,13 @@ void
 spawn_thread(pthread_t *thread, unsigned int *arg)
 {
     int status_code = -1;
+    int timeout_us = RETRY_SLEEP_TIME_US;
     for (int tries = 0; status_code != 0 && tries < NUM_RETRY; ++tries) {
         status_code = pthread_create(thread, NULL, &check_if_prime, arg);
         assert(status_code == 0 || status_code == EAGAIN);
         if (status_code == EAGAIN) {
-            usleep(2000);
+            usleep(timeout_us);
+            timeout_us *= 2;
         }
     }
 
@@ -95,7 +98,7 @@ main(int argc, char **argv)
 
         args[thread_num] = factorised_number;
 
-        usleep(2000);
+        usleep(RETRY_SLEEP_TIME_US);
         spawn_thread(&threads[thread_num], &args[thread_num]);
         assert(threads[thread_num] != 0);
     }
--- a/core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
+++ b/core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
@@ -1,0 +1,93 @@
+/*
+ * Copyright (C) 2023 Amazon.com Inc. or its affiliates. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+
+#include <assert.h>
+#include <errno.h>
+#include <pthread.h>
+#include <stdio.h>
+#include <unistd.h>
+
+enum CONSTANTS {
+    NUM_ITER = 200000,
+    NUM_RETRY = 8,
+    MAX_NUM_THREADS = 8,
+    RETRY_SLEEP_TIME_US = 4000,
+    SECOND = 1000 * 1000 * 1000
+};
+
+int threads_executed = 0;
+unsigned int threads_creation_tried = 0;
+unsigned int threads_in_use = 0;
+
+void *
+thread_func(void *arg)
+{
+    (void)(arg);
+    __atomic_fetch_add(&threads_executed, 1, __ATOMIC_RELAXED);
+    __atomic_fetch_sub(&threads_in_use, 1, __ATOMIC_SEQ_CST);
+    return NULL;
+}
+
+void
+spawn_thread(pthread_t *thread)
+{
+    int status_code = -1;
+    int timeout_us = RETRY_SLEEP_TIME_US;
+    for (int tries = 0; status_code != 0 && tries < NUM_RETRY; ++tries) {
+        status_code = pthread_create(thread, NULL, &thread_func, NULL);
+        __atomic_fetch_add(&threads_creation_tried, 1, __ATOMIC_RELAXED);
+
+        assert(status_code == 0 || status_code == EAGAIN);
+        if (status_code == EAGAIN) {
+            usleep(timeout_us);
+            timeout_us *= 2;
+        }
+    }
+
+    assert(status_code == 0 && "Thread creation should succeed");
+}
+
+int
+main(int argc, char **argv)
+{
+    double percentage = 0.1;
+
+    for (int iter = 0; iter < NUM_ITER; ++iter) {
+        if (iter > NUM_ITER * percentage) {
+            fprintf(stderr, "Spawning stress test is %d%% finished\n",
+                    (unsigned int)(percentage * 100));
+            percentage += 0.1;
+        }
+        while (__atomic_load_n(&threads_in_use, __ATOMIC_SEQ_CST)
+               == MAX_NUM_THREADS) {
+            usleep(100);
+        }
+
+        __atomic_fetch_add(&threads_in_use, 1, __ATOMIC_SEQ_CST);
+        pthread_t tmp;
+        spawn_thread(&tmp);
+        pthread_detach(tmp);
+    }
+
+    while ((__atomic_load_n(&threads_in_use, __ATOMIC_SEQ_CST) != 0)) {
+        __builtin_wasm_memory_atomic_wait32(&threads_in_use, 0, SECOND);
+    }
+
+    assert(__atomic_load_n(&threads_in_use, __ATOMIC_SEQ_CST) == 0);
+
+    // Validation
+    assert(threads_creation_tried >= threads_executed
+           && "Test executed more threads than were created");
+    assert((1. * threads_creation_tried) / threads_executed < 2.5
+           && "Ensuring that we're retrying thread creation less than 2.5 "
+              "times on average ");
+
+    fprintf(stderr,
+            "Spawning stress test finished successfully executed %d threads "
+            "with retry ratio %f\n",
+            threads_creation_tried,
+            (1. * threads_creation_tried) / threads_executed);
+    return 0;
+}
--- a/tests/wamr-test-suites/wasi-test-script/run_wasi_tests.sh
+++ b/tests/wamr-test-suites/wasi-test-script/run_wasi_tests.sh
@@ -14,6 +14,7 @@ readonly WAMR_DIR="${WORK_DIR}/../../../.."
 readonly IWASM_CMD="${WORK_DIR}/../../../../product-mini/platforms/${PLATFORM}/build/iwasm \
     --allow-resolve=google-public-dns-a.google.com \
     --addr-pool=::1/128,127.0.0.1/32"
+
 readonly IWASM_CMD_STRESS="${IWASM_CMD} --max-threads=8"
 readonly WAMRC_CMD="${WORK_DIR}/../../../../wamr-compiler/build/wamrc"
 readonly C_TESTS="tests/c/testsuite/"
@@ -21,18 +22,22 @@ readonly ASSEMBLYSCRIPT_TESTS="tests/assemblyscript/testsuite/"
 readonly THREAD_PROPOSAL_TESTS="tests/proposals/wasi-threads/"
 readonly THREAD_INTERNAL_TESTS="${WAMR_DIR}/core/iwasm/libraries/lib-wasi-threads/test/"
 readonly LIB_SOCKET_TESTS="${WAMR_DIR}/core/iwasm/libraries/lib-socket/test/"
+readonly STRESS_TESTS=("spawn_stress_test.wasm" "stress_test_threads_creation.wasm")
 
 run_aot_tests () {
     local tests=("$@")
+    local iwasm="${IWASM_CMD}"
     for test_wasm in ${tests[@]}; do
         local extra_stress_flags=""
-        if [[ "$test_wasm" =~ "stress" ]]; then
-            extra_stress_flags="--max-threads=8"
-        fi
+        for stress_test in "${STRESS_TESTS[@]}"; do
+            if [ "$test_wasm" == "$stress_test" ]; then
+                iwasm="${IWASM_CMD_STRESS}"
+            fi  
+        done
 
         test_aot="${test_wasm%.wasm}.aot"
         test_json="${test_wasm%.wasm}.json"
-
+ 
         if [ -f ${test_wasm} ]; then
             expected=$(jq .exit_code ${test_json})
         fi
@@ -48,7 +53,6 @@ run_aot_tests () {
         fi
 
         ${IWASM_CMD} $extra_stress_flags $test_aot
-
         ret=${PIPESTATUS[0]}
 
         echo "expected=$expected, actual=$ret"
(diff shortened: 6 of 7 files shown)


.github/workflows/compilation_on_android_ubuntu.yml
```diff
@@ -527,7 +527,7 @@ jobs:
         working-directory: ./core/iwasm/libraries/lib-socket/test/
 
       - name: run tests
-        timeout-minutes: 20
```
loganek, line 530: How long does it take to run the stress test? If that all adds up to 60 minutes now, I think it's reasonable to keep for nightly only.

core/iwasm/libraries/lib-wasi-threads/test/spawn_stress_test.c
```diff
@@ -20,6 +20,7 @@ enum CONSTANTS {
     NUM_ITER = 100000,
     NUM_RETRY = 5,
     MAX_NUM_THREADS = 8,
+    RETRY_SLEEP_TIME = 8000,
```
loganek, line 23: worth adding a unit to the name, so the reader doesn't have to look for the usage to know the meaning of the number.
```suggestion
    RETRY_SLEEP_TIME_US = 8000,
```

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
```diff
@@ -0,0 +1,89 @@
+/*
+ * Copyright (C) 2023 Amazon.com Inc. or its affiliates. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+
+#ifndef __wasi__
+#error This example only compiles to WASM/WASI target
```
loganek, line 7: not needed I guess?

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
```diff
@@ -0,0 +1,89 @@
...
+void
+spawn_thread(pthread_t *thread)
+{
+    int status_code = -1;
+    for (int tries = 0; status_code != 0 && tries < NUM_RETRY; ++tries) {
+        status_code = pthread_create(thread, NULL, &thread_func, NULL);
+        ++threads_creation_tried;
+        assert(status_code == 0 || status_code == EAGAIN);
+        if (status_code == EAGAIN) {
+            usleep(RETRY_SLEEP_TIME);
```
loganek, line 43: I wonder if exponential backoff wouldn't be better here; if we always assume 8ms sleep time (which is quite a lot IMO for this type of test), we might not be stressing the implementation enough.

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
```diff
@@ -0,0 +1,89 @@
...
+
+    for (int iter = 0; iter < NUM_ITER; ++iter) {
+        if (iter > NUM_ITER * percentage) {
+            fprintf(stderr, "Spawning stress test is %d%% finished\n",
+                    (unsigned int)(percentage * 100));
+            percentage += 0.1;
+        }
+
+        unsigned int thread_num = iter % MAX_NUM_THREADS;
+        if (threads[thread_num] != 0) {
```
loganek, line 64: For this test I think we should spawn threads as quickly as possible - here we wait for a specific thread ID to be released, even though other IDs might have been freed already and could be used. I think this was discussed for the other test too, but I don't know if that was eventually addressed? If so, the code probably could be reused.

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    for (int i = 0; i < MAX_NUM_THREADS; ++i) {
+        assert(threads[i] == 0 || pthread_join(threads[i], NULL) == 0);
+    }
+
+    // Validation
+    assert(threads_creation_tried >= threads_executed
+           && "Test executed more threads than were created");
+    assert((1. * threads_creation_tried) / threads_executed < 1.05
```
loganek, line 80: where does the `1.05` number come from? Have you tested it on different configurations? What if we run the test on some low-end devices - would the test fail?

core/iwasm/libraries/lib-wasi-threads/test/skip.json
```diff
@@ -1,5 +1,6 @@
 {
     "lib-wasi-threads tests": {
-        "spawn_stress_test": "Stress tests are incompatible with the other part and executed differently"
+        "spawn_stress_test": "Stress tests are incompatible with the other part and executed differently",
+        "stress_test_threads_creation": "Stress tests are incompatible with the other part and executed differently"
```
loganek, line 4: more tests we add here more convinced I am this should all be ran as part of the test runner. What would it take to run those tests just like we run other thread tests?

.github/workflows/compilation_on_android_ubuntu.yml
Zzzabiyaka: it's a bit tricky 

actually it should be 2 times less. This constraint applies because `jit` is basically 2 bundle runs 

here's the ticket to set timeouts based on the platform. So, all platforms <= 30 minutes, jit <= 60 minutes

core/iwasm/libraries/lib-wasi-threads/test/skip.json
Zzzabiyaka: for AOT - nothing 

for non-aot I don't even know how hard will it be (because aot is executed differently)

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
Zzzabiyaka: No I didn't 

1.05 is meant to be "a bit higher than 1", e.g. to allow some retries but not that much relatively to created threads. 
 
Maybe I need to double-check on some low devices

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
Zzzabiyaka: I'd expect that we use this part of pthread_join

```  
If that thread has already terminated, then
       pthread_join() returns immediately. 
```

I'm curious whether there's any better solution. If we have mutex, thread function is going to be much longer I suppose. 

If we have atomic, we most likely need SEQ order, also not the fastest solution. 

I think that `thread_execution_time` is going to be less than `7 * thread_creation_time` for this tiny thread function. 

But it's just my assumation, tell me if I'm wrong

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
Zzzabiyaka: hm, interesting idea, I will think about it. Thanks!

Zzzabiyaka: less than 8ms was introducing some flakiness where test could randomly fail

core/iwasm/libraries/lib-wasi-threads/test/spawn_stress_test.c
Zzzabiyaka: Thanks will do

.github/workflows/compilation_on_android_ubuntu.yml
loganek: Even if

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
loganek: So imagine we have three threads (call it 1, 2 and 3). Potential sequence can be:
```
Spawn thread 1
Spawn thread 2
Spawn thread 3
Finish thread 2
Finish thread 3
Spawn thread 1 <- block here
```
Because the first thread is still alive (e.g. due to scheduling), we can't spawn new thread even though slot 2 and 3 are freed already, so we wait unnecessarily.
Not sure how likely or severe the issue is, but it feels like we can completely remove the problem with a low overhead by not relying on the order.

.github/workflows/compilation_on_android_ubuntu.yml
wenyongh: Seems it took long time to run the stress cases:
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/5761023933/job/15618324568
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/5761023933/job/15618325167
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/5761023933/job/15618325916
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/5761023933/job/15618326547
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/5761023933/job/15618326990

Almost close to 30 min for each running mode, but it is only 4-5 min now:
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/5791716284/job/15697007728
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/5791716284/job/15697008239
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/5791716284/job/15697009623

Shall we put the stress test in nightly run?

Zzzabiyaka: we can either decrease the constants (what should be harmless, because if it did 100'000 threads it's really likely to do 10 times more without errors or put to nightly

wenyongh: How about decreasing the constants to ensure it doesn't increase too much time in this PR, and then adding another PR for stress test?
We had better not make it take too long to run the CIs of a PR, developer may be eager to see the running result.

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
Zzzabiyaka: Fixed

core/iwasm/libraries/lib-wasi-threads/test/skip.json
Zzzabiyaka: Maybe we can go the opposite way and scope threads that depend on max_threads=4 (there should be only one/a couple of them)

.github/workflows/compilation_on_android_ubuntu.yml
Zzzabiyaka: I've changed the constants. I will wait for execution and will adjust timeouts respectively

wenyongh: OK, the nightly_run CI reported failure:

https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/5801072533/job/15724690363#step:12:488

```
stress_test_threads_creation.wasm is a stress test
Assertion failed: status_code == 0 && "Thread creation should succeed" (stress_test_threads_creation.c: spawn_thread: 55)
Exception: unreachable
Stress test stress_test_threads_creation.wasm FAILED with code  1
```

Zzzabiyaka: for some reason `multitier-jit` looks really problematic in terms of threads creation. 

I'm not sure whether it's by design or it can be a potential sign of some bug

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
```diff
@@ -0,0 +1,123 @@
...
+                threads_executed);
+    }
+    assert(status_code == 0 && "Thread creation should succeed");
+}
+
+int
+main(int argc, char **argv)
+{
+    pthread_mutex_init(&mutex, NULL);
+    pthread_t threads[MAX_NUM_THREADS];
```
loganek, line 67: Since you detach the thread, you probably don't need this array any more?

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
```diff
@@ -0,0 +1,123 @@
...
+    NUM_ITER = 90000,
+    NUM_RETRY = 5,
+    MAX_NUM_THREADS = 8,
+    RETRY_SLEEP_TIME_US = 4000,
+};
+
+unsigned int threads_executed = 0;
+unsigned int threads_creation_tried = 0;
+
+pthread_mutex_t mutex;
```
loganek, line 23:
```suggestion
pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
```

and then you can remove init/destroy calls in the main function.

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
```diff
@@ -0,0 +1,123 @@
...
+
+bool is_finished[MAX_NUM_THREADS];
+
+void *
+thread_func(void *arg)
+{
+    __atomic_fetch_add(&threads_executed, 1, __ATOMIC_RELAXED);
+
+    pthread_mutex_lock(&mutex);
+    int num = *(int *)(arg);
```
loganek, line 33: won't that line be removed by compiler's optimizer?

core/iwasm/libraries/lib-wasi-threads/test/build.sh
```diff
@@ -10,7 +10,7 @@ CC=${CC:=/opt/wasi-sdk/bin/clang}
 WAMR_DIR=../../../../..
 
 # Stress tests names
-thread_start_file_exclusions=("spawn_stress_test.wasm" "linear_memory_size_update.wasm")
+thread_start_file_exclusions=("spawn_stress_test.wasm" "linear_memory_size_update.wasm" "stress_test_threads_creation.wasm")
```
loganek, line 13: for this new test we probably want to enable compiler optimizations so we can spawn as many threads as we can.

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
```diff
@@ -0,0 +1,123 @@
...
+                    (unsigned int)(percentage * 100));
+            percentage += 0.1;
+        }
+
+        int thread_num = -1;
+
+        while (thread_num == -1) {
+            pthread_mutex_lock(&mutex);
+            for (int i = 0; i < MAX_NUM_THREADS; ++i) {
+                if (is_finished[i]) {
```
loganek, line 88: do we need to know which thread finished? Wouldn't be enough to know how many threads is running?

loganek (changes requested):

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
Zzzabiyaka: I want to reuse `pthread_t` array so I need to know which of `threads[i]` I need to rewrite

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
Zzzabiyaka: I don't think so

I mean, num var is used in the next line so you can't remove it's initialising, if I got your question right

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
Zzzabiyaka: hm, 

I had a concern that `pthread_t` var assigned to thread should be alive until thread finishes. 

I'm not a posix expert so I'm not 100% so tell me if I'm wrong

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
Zzzabiyaka: thanks

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
Zzzabiyaka: I don't think so

I mean, num var is used in the next line so you can't remove it's initialising, if I got your question right

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
Zzzabiyaka: hm,

I had a concern that pthread_t var assigned to thread should be alive until thread finishes.

I'm not a posix expert so I'm not 100% sure so tell me if I'm wrong

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
loganek: Ah yes, sorry I didn't provide a full context in this comment. I also assumed we don't need the is_finished array

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
loganek: You don't need to keep track on the thread I'd if you don't use it. `pthread_t` in wasi libc (and many other implementations) is simply an integer. Also, C doesn't have RAII by default soeven if it was a complex data structure allowing thread management, worse case you'd have a memory leak, but thread wouldn't be affected unless some function was explicitly called

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
loganek: As said in the comment above, you don't need to track it

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
Zzzabiyaka: thanks. Got it

tests/wamr-test-suites/wasi-test-script/run_wasi_tests.sh
```diff
@@ -14,21 +14,26 @@ readonly WAMR_DIR="${WORK_DIR}/../../../.."
 readonly IWASM_CMD="${WORK_DIR}/../../../../product-mini/platforms/${PLATFORM}/build/iwasm \
     --allow-resolve=google-public-dns-a.google.com \
     --addr-pool=::1/128,127.0.0.1/32"
-readonly IWASM_CMD_STRESS="${IWASM_CMD} --max-threads=8"
+readonly IWASM_CMD_STRESS="${IWASM_CMD} --max-threads=8 -v=5"
```
wenyongh, line 17: Must we add `-v=5`? It may print a lot of logs.

wenyongh (approved):
LGTM

@Zzzabiyaka @loganek Is it OK merge this PR now?

core/iwasm/libraries/lib-wasi-threads/test/skip.json
Zzzabiyaka: let's do this in another PR

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
Zzzabiyaka: now it's 2.5 

i'll merge a fix if it's not right for low-end devices after testing. I think we shouldn't block this PR on that

core/iwasm/libraries/lib-wasi-threads/test/build.sh
Zzzabiyaka: that makes sense and I think we can safely set -O2 for all the tests here. Does it make sense to you?

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
```diff
@@ -0,0 +1,90 @@
...
+
+    for (int iter = 0; iter < NUM_ITER; ++iter) {
+        if (iter > NUM_ITER * percentage) {
+            fprintf(stderr, "Spawning stress test is %d%% finished\n",
+                    (unsigned int)(percentage * 100));
+            percentage += 0.1;
+        }
+        while (__atomic_load_n(&threads_in_use, __ATOMIC_SEQ_CST)
+               == MAX_NUM_THREADS) {
+            usleep(100);
```
loganek, line 68: did you consider using conditionals or atomic waits here instead of sleep?

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        __atomic_fetch_add(&threads_in_use, 1, __ATOMIC_SEQ_CST);
+        pthread_t tmp;
+        spawn_thread(&tmp);
+        pthread_detach(tmp);
+    }
+
```
loganek, line 76: should we wait here for all the threads to complete?

tests/wamr-test-suites/wasi-test-script/run_wasi_tests.sh
```diff
@@ -62,15 +66,18 @@ if [[ $MODE != "aot" ]];then
...
-    if [[ -e "${THREAD_INTERNAL_TESTS}spawn_stress_test.wasm" ]]; then 
-        ${IWASM_CMD_STRESS} ${THREAD_INTERNAL_TESTS}spawn_stress_test.wasm
-        ret=${PIPESTATUS[0]}
-        if [ "${ret}" -ne 0 ]; then
-            echo "Stress test spawn_stress_test FAILED with code " ${ret}
-            exit_code=${ret}
+    # Stress tests require max-threads=8 so they're executed separately
+    for stress_test in "${STRESS_TESTS[@]}"; do
+        echo "${stress_test}" is a stress test
+        if [[ -e "${THREAD_INTERNAL_TESTS}${stress_test}" ]]; then
```
loganek, line 72: when is this condition not true?

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
Zzzabiyaka: we can't probably wait for them, as they're detached, but at least we can wait until threads_in_use is zero

loganek: yes, I meant waiting for `threads_in_use` to be zero

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
Zzzabiyaka: we can do that I think (especially atomic wait) but I'm not sure what the benefit will we get comparing to just busy wait

tests/wamr-test-suites/wasi-test-script/run_wasi_tests.sh
Zzzabiyaka: We're just not executing the test as the wasm doesn't exist. 

we can just move `echo` inside the if

Zzzabiyaka: if you don't want to run stress tests for any reasons , the fastest way for you is just to delete .wasm files.
You don't have any other easy ways to avoid running it, so until you're able to disable them explicitly (e.g. until test-runner and refactoring test running system in wamr ) I'd probably leave an opportunity to avoid running stress tests. 

But generally I don't mind and can delete this line

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
Zzzabiyaka: done, thanks

core/iwasm/libraries/lib-wasi-threads/test/stress_test_threads_creation.c
Zzzabiyaka: and how can we do that actually? 
we can introduce a cond var that will be notifying when thread finishes but I think it's just making the test a bit more complex without any benefit (and most likely slows it as well because I don't think it will be any faster using cond var

loganek: the advantage is that you won't have to wait for 0.1ms for the thread to complete if it completes sooner than that. But it's up to you whether to keep it like that or not, I don't insist to change it.

loganek (approved):

Issue #2407, opened 2023-07-31
broken locking wrt exceptions


yamt:
https://github.com/bytecodealliance/wasm-micro-runtime/pull/1995 started to use shared_mem_lock to protect exceptions.
however, it seems wrong by design because a module might not have a memory.
(or, multiple memories in future.)

PR #2415, opened 2023-08-01
Fix memory sharing

yamt:
- inherit shared memory from the parent instance, instead of
  trying to look it up by the underlying module. the old method     
  works correctly only when every clusters use different modules.

- reference count WASMMemoryInstance/AOTMemoryInstance directly

- retire WASMSharedMemNode

- for atomic instructions in interpreters, use a global lock for now.

- update users (wasi-threads, lib-pthread, wasm_runtime_spawn_thread)

fixes https://github.com/bytecodealliance/wasm-micro-runtime/issues/1962

todo:
- ~~aot~~
- ~~classic interpreter~~
- ~~lib-pthread~~
- ~~test wasm_runtime_spawn_thread~~
- ~~test lib-pthread~~

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -339,9 +339,8 @@ memories_deinstantiate(AOTModuleInstance *module_inst)
         memory_inst = module_inst->memories[i];
         if (memory_inst) {
 #if WASM_ENABLE_SHARED_MEMORY != 0
-            if (memory_inst->is_shared) {
-                int32 ref_count = shared_memory_dec_reference(
-                    (WASMModuleCommon *)module_inst->module);
+            if (shared_memory_is_shared(memory_inst)) {
+                uint32 ref_count = shared_memory_dec_reference(memory_inst);
                 bh_assert(ref_count >= 0);
 
                 /* if the reference count is not zero,
@@ -373,9 +372,10 @@ memories_deinstantiate(AOTModuleInstance *module_inst)
 }
 
 static AOTMemoryInstance *
-memory_instantiate(AOTModuleInstance *module_inst, AOTModule *module,
-                   AOTMemoryInstance *memory_inst, AOTMemory *memory,
-                   uint32 heap_size, char *error_buf, uint32 error_buf_size)
+memory_instantiate(AOTModuleInstance *module_inst, AOTModuleInstance *parent,
+                   AOTModule *module, AOTMemoryInstance *memory_inst,
+                   AOTMemory *memory, uint32 memory_idx, uint32 heap_size,
+                   char *error_buf, uint32 error_buf_size)
 {
     void *heap_handle;
     uint32 num_bytes_per_page = memory->num_bytes_per_page;
@@ -396,23 +396,16 @@ memory_instantiate(AOTModuleInstance *module_inst, AOTModule *module,
     bool is_shared_memory = memory->memory_flags & 0x02 ? true : false;
 
     /* Shared memory */
-    if (is_shared_memory) {
+    if (is_shared_memory && parent != NULL) {
         AOTMemoryInstance *shared_memory_instance;
-        WASMSharedMemNode *node =
-            wasm_module_get_shared_memory((WASMModuleCommon *)module);
-        /* If the memory of this module has been instantiated,
-            return the memory instance directly */
-        if (node) {
-            uint32 ref_count;
-            ref_count = shared_memory_inc_reference((WASMModuleCommon *)module);
-            bh_assert(ref_count > 0);
-            shared_memory_instance =
-                (AOTMemoryInstance *)shared_memory_get_memory_inst(node);
-            bh_assert(shared_memory_instance);
-
-            (void)ref_count;
-            return shared_memory_instance;
-        }
+        uint32 ref_count;
+        bh_assert(memory_idx == 0);
+        bh_assert(parent->memory_count > memory_idx);
+        shared_memory_instance = parent->memories[memory_idx];
+        ref_count = shared_memory_inc_reference(shared_memory_instance);
+        bh_assert(ref_count > 0);
+        (void)ref_count;
+        return shared_memory_instance;
     }
 #endif
 
@@ -609,23 +602,12 @@ memory_instantiate(AOTModuleInstance *module_inst, AOTModule *module,
 
 #if WASM_ENABLE_SHARED_MEMORY != 0
     if (is_shared_memory) {
-        memory_inst->is_shared = true;
-        if (!shared_memory_set_memory_inst(
-                (WASMModuleCommon *)module,
-                (WASMMemoryInstanceCommon *)memory_inst)) {
-            set_error_buf(error_buf, error_buf_size, "allocate memory failed");
-            goto fail3;
-        }
+        memory_inst->ref_count = 1;
     }
 #endif
 
     return memory_inst;
 
-#if WASM_ENABLE_SHARED_MEMORY != 0
-fail3:
-    if (heap_size > 0)
-        mem_allocator_destroy(memory_inst->heap_handle);
-#endif
 fail2:
     if (heap_size > 0)
         wasm_runtime_free(memory_inst->heap_handle);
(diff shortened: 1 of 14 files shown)


core/iwasm/common/wasm_runtime_common.c
```diff
@@ -2310,10 +2311,8 @@ wasm_set_exception(WASMModuleInstance *module_inst, const char *exception)
     WASMExecEnv *exec_env = NULL;
 
 #if WASM_ENABLE_SHARED_MEMORY != 0
-    WASMSharedMemNode *node =
-        wasm_module_get_shared_memory((WASMModuleCommon *)module_inst->module);
-    if (node)
-        os_mutex_lock(&node->shared_mem_lock);
+    if (module_inst->memory_count > 0)
+        shared_memory_lock(module_inst->memories[0]);
```
xujuntwt95329, line 2315: @wenyongh Do you remember why we need to lock the shared memory when setting exception?

xujuntwt95329, line 2315: @wenyongh Do you remember why we need to lock the shared memory when setting exception?

yamt: see https://github.com/bytecodealliance/wasm-micro-runtime/issues/2407

wenyongh: yes, I remember that one instance's exception may be accessed by other threads when spreading non "wasi proc exit" exception. It was introduced in this PR https://github.com/bytecodealliance/wasm-micro-runtime/pull/1995

In thread_manager.c, function set_exception_visitor:
```C
        /* Only spread non "wasi proc exit" exception */
        WASMSharedMemNode *shared_mem_node = wasm_module_get_shared_memory(
            (WASMModuleCommon *)curr_wasm_inst->module);
        if (shared_mem_node)
            os_mutex_lock(&shared_mem_node->shared_mem_lock);
#endif
        if (!strstr(wasm_inst->cur_exception, "wasi proc exit")) {
            bh_memcpy_s(curr_wasm_inst->cur_exception,
                        sizeof(curr_wasm_inst->cur_exception),
                        wasm_inst->cur_exception,
                        sizeof(wasm_inst->cur_exception));
        }
        if (shared_mem_node)
            os_mutex_unlock(&shared_mem_node->shared_mem_lock);
#endif
```

eloparco: > yes, I remember that one instance's exception may be accessed by other threads when spreading non "wasi proc exit" exception. It was introduced in this PR https://github.com/bytecodealliance/wasm-micro-runtime/pull/1995

Yes, that's the reason, I was about to comment. When a thread runs into an exception it spreads it to other threads by using `set_exception_visitor` and that other threads could try to read it at the same time, causing concurrency issues.

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -710,28 +710,28 @@ trunc_f64_to_int(WASMModuleInstance *module, uint32 *frame_sp, float64 src_min,
             CHECK_BULK_MEMORY_OVERFLOW(addr + offset, 1, maddr);     \
             CHECK_ATOMIC_MEMORY_ACCESS();                            \
                                                                      \
-            os_mutex_lock(&node->shared_mem_lock);                   \
+            shared_memory_lock(memory);                              \
```
wenyongh, line 713: An concern is about the performance, calling `shared_memory_lock/unlock` adds two function calls, it may impact the performance. How about getting the lock pointer, and here `os_mutex_{lock|unlock}(shared_mem_lock)`?

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -122,9 +122,8 @@ memories_deinstantiate(WASMModuleInstance *module_inst,
                 }
 #endif
 #if WASM_ENABLE_SHARED_MEMORY != 0
-                if (memories[i]->is_shared) {
-                    int32 ref_count = shared_memory_dec_reference(
-                        (WASMModuleCommon *)module_inst->module);
+                if (shared_memory_is_shared(memories[i])) {
+                    uint32 ref_count = shared_memory_dec_reference(memories[i]);
```
wenyongh, line 126: Same as above, had better use `int32`?

core/iwasm/interpreter/wasm_runtime.h
```diff
@@ -79,7 +80,7 @@ struct WASMMemoryInstance {
     /* Module type */
     uint32 module_type;
     /* Shared memory flag */
-    bool is_shared;
+    bh_atomic_32_t ref_count; /* 0: non-shared, >0: reference count */
```
wenyongh, line 83: Coding style, suggest to change to `> 0`?

core/iwasm/aot/aot_runtime.c
```diff
@@ -339,9 +339,8 @@ memories_deinstantiate(AOTModuleInstance *module_inst)
         memory_inst = module_inst->memories[i];
         if (memory_inst) {
 #if WASM_ENABLE_SHARED_MEMORY != 0
-            if (memory_inst->is_shared) {
-                int32 ref_count = shared_memory_dec_reference(
-                    (WASMModuleCommon *)module_inst->module);
+            if (shared_memory_is_shared(memory_inst)) {
+                uint32 ref_count = shared_memory_dec_reference(memory_inst);
```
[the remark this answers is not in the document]
yamt: maybe i will move these assertions to shared_memory_dec_reference/shared_memory_inc_reference.

core/iwasm/interpreter/wasm_interp_classic.c
yamt: does it matter? if we really care the performance here, i guess we should use atomic ops as we do for aot.

core/iwasm/interpreter/wasm_runtime.h
yamt: ok.

core/iwasm/common/wasm_runtime_common.c
xujuntwt95329: Thanks a lot!

core/iwasm/interpreter/wasm_interp_classic.c
wenyongh: I am not very sure, the performance may be not so important here, I just see an opportunity to refine the code😊. Using atomic operations may be also a choice in the future.
@loganek, any idea?

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -48,128 +55,92 @@ destroy_wait_info(void *wait_info);
...
 }
 
-WASMSharedMemNode *
-wasm_module_get_shared_memory(WASMModuleCommon *module)
+uint32
+shared_memory_dec_reference(WASMMemoryInstance *memory)
 {
-    return search_module(module);
+    bh_assert(shared_memory_is_shared(memory));
+    uint32_t old;
```
xujuntwt95329, line 96:
```suggestion
    uint32 old;
```

core/iwasm/common/wasm_shared_memory.c
```diff
@@ -48,128 +55,92 @@ destroy_wait_info(void *wait_info);
...
 {
-    WASMSharedMemNode *node = search_module(module);
-    uint32 ref_count = -1;
-    if (node) {
-        os_mutex_lock(&node->lock);
-        ref_count = ++node->ref_count;
-        os_mutex_unlock(&node->lock);
-    }
-    return ref_count;
+    uint32_t old;
```
xujuntwt95329, line 110:
```suggestion
    uint32 old;
```

core/iwasm/aot/aot_runtime.c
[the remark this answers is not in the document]
yamt: done

core/iwasm/interpreter/wasm_runtime.h
yamt: fixed

core/iwasm/common/wasm_shared_memory.c
yamt: oops. it seems that my fingers automatically type `_t`. fixed.

xujuntwt95329 (commented):
LGTM

wenyongh (commented):
LGTM

Issue #2252, opened 2023-06-03
[BUG] GS register shouldn't be modified in linux-sgx platform hardware mode (64 bit)


LeoneChen:
Hello~

`get_thread_data` use `READ_TD_DATA` assembly macro [here](https://github.com/intel/linux-sgx/blob/1efe23c20e37f868498f8287921eedfbcecdc216/sdk/trts/linux/trts_pic.S#L78)
<img width="211" alt="image" src="https://github.com/bytecodealliance/wasm-micro-runtime/assets/35480440/ea08dc94-d003-4c24-92bd-7a7015a97087">

And in SGX hardware mode, `READ_TD_DATA` use `GS` register in 64bit mode as base address of thread data. [here](https://github.com/intel/linux-sgx/blob/1efe23c20e37f868498f8287921eedfbcecdc216/sdk/trts/linux/trts_pic.h#L106)
(Note in hardware mode `SE_SIM` is undefined)
<img width="330" alt="image" src="https://github.com/bytecodealliance/wasm-micro-runtime/assets/35480440/0b54e60d-2b62-4cf6-b65a-38ea316596dc">

Thus it shouldn't modify `GS` register in Enclave.[here for example](https://github.com/bytecodealliance/wasm-micro-runtime/blob/f1e9029ebcca71c357bfabfc0376c73c3b1d1d0d/core/iwasm/interpreter/wasm_interp_fast.c#L3987)
<img width="428" alt="image" src="https://github.com/bytecodealliance/wasm-micro-runtime/assets/35480440/16ada231-4e58-408b-bfac-41734bd53bb2">

`os_writegsbase` macro is at [here](https://github.com/bytecodealliance/wasm-micro-runtime/blob/f1e9029ebcca71c357bfabfc0376c73c3b1d1d0d/core/shared/platform/linux-sgx/platform_internal.h#L63)

<img width="410" alt="image" src="https://github.com/bytecodealliance/wasm-micro-runtime/assets/35480440/a884d3f2-5aed-4dea-86dc-ec938ee27b0f">

wenyongh:
@LeoneChen Thanks for pointing out this! I only tested in simulator mode and thought that GS register is available. It is a pity that we cannot use the register to enable the segue optimization for linux-sgx platform, I submitted PR #2255 to fix the issue.

LeoneChen:
My pleasure~

LeoneChen:
Hello~ We find another bug
# Heap OOB
[...]

PR #2416, opened 2023-08-02
Add runtime inited checks in Enclave command handlings to improve security

wenyongh:
As reported in https://github.com/bytecodealliance/wasm-micro-runtime/issues/2252#issuecomment-1634940219,
`wait_map` may be referred again after free in linux-sgx ecall command if it isn't set NULL.

### DIFF
--- a/core/iwasm/common/wasm_shared_memory.c
+++ b/core/iwasm/common/wasm_shared_memory.c
@@ -66,6 +66,7 @@ void
 wasm_shared_memory_destroy()
 {
     bh_hash_map_destroy(wait_map);
+    wait_map = NULL;
     os_mutex_destroy(&shared_memory_list_lock);
 }
 
--- a/core/iwasm/common/wasm_shared_memory.c
+++ b/core/iwasm/common/wasm_shared_memory.c
@@ -66,7 +66,6 @@ void
 wasm_shared_memory_destroy()
 {
     bh_hash_map_destroy(wait_map);
-    wait_map = NULL;
     os_mutex_destroy(&shared_memory_list_lock);
 }
 
--- a/product-mini/platforms/linux-sgx/enclave-sample/Enclave/Enclave.cpp
+++ b/product-mini/platforms/linux-sgx/enclave-sample/Enclave/Enclave.cpp
@@ -92,6 +92,8 @@ set_error_buf(char *error_buf, uint32 error_buf_size, const char *string)
         snprintf(error_buf, error_buf_size, "%s", string);
 }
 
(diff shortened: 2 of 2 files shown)


yamt:
are you trying to make wasm_runtime_destroy idempotent?
otherwise, i guess it should be fixed in Enclave.cpp instead.

wenyongh:
> are you trying to make wasm_runtime_destroy idempotent? otherwise, i guess it should be fixed in Enclave.cpp instead.

Good point, I added more checks in Enclave instead to fix some similar issues.

yamt (approved):
lgtm

LeoneChen:
Hello, I think `wait_map = NULL;` is necessary. And I found other functions use `wait_map` already have check `wait_map` is `NULL` at function entrypoint like `bh_hash_map_find`.

If you only fix at `linux-sgx` platform, other platform may also face same UAF bug

yamt:
> Hello, I think `wait_map = NULL;` is necessary. And I found other functions use `wait_map` already have check `wait_map` is `NULL` at function entrypoint like `bh_hash_map_find`, then this problem can be fixed.
> 
> If you only fix at `linux-sgx` platform, other platform may also face same UAF bug

the bug (calling `wasm_runtime_destroy` twice) was in linux-sgx and it has been fixed.

if other platforms have similar bugs, they should be fixed too. (have they?)

having double-free checks everywhere is not viable, IMO.

LeoneChen:
It's similar to blow opinion.

It's different from ``free` since the caller needs to specify which pointer to free, e.g. `free(ptr)`, it's the caller's duty to set ptr to null.

However, in WAMR, API is `wasm_runtime_destroy()` and it free `wait_map` internal, caller (library user) can't access the `wait_map` variable, then it's duty of `wasm_runtime_destroy()` to set `wait_map` to NULL

yamt:
> It's different from `free` since the caller needs to specify which pointer to free, e.g. `free(ptr)`, it's the caller's duty to set ptr to null.
> 
> However, in WAMR, API is `wasm_runtime_destroy()` and it free `wait_map` internal, caller (library user) can't access the `wait_map` variable, then it's duty of `wasm_runtime_destroy()` to set `wait_map` to NULL

i disagree. while it's ok to set wait_map to NULL, it's far from a "duty".
calling `wasm_runtime_destroy` twice as sgx was doing is an undefined behavior as double free is.
what you are saying is something like it's `free`'s duty to prepare to handle double free.

LeoneChen:
All right

PR #2421, opened 2023-08-03
Fix typo in test_wamr.sh

Zzzabiyaka:
I will close this PR if it is done on purpose. 

Another question: can I split "jit" test type to "eager_jit" and "lazy_jit" so that the ci is more consistent (e.g. one test type would always mean that test bundle will run once. It's already true right now with jit being the only exception)

### DIFF
--- a/tests/wamr-test-suites/test_wamr.sh
+++ b/tests/wamr-test-suites/test_wamr.sh
@@ -824,7 +824,7 @@ function trigger()
                 collect_coverage llvm-jit
 
                 echo "work in orc jit lazy compilation mode"
-                BUILD_FLAGS="$ORC_EAGER_JIT_COMPILE_FLAGS $EXTRA_COMPILE_FLAGS"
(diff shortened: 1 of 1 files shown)


wenyongh:
> I will close this PR if it is done on purpose.
> 
> Another question: can I split "jit" test type to "eager_jit" and "lazy_jit" so that the ci is more consistent (e.g. one test type would always mean that test bundle will run once. It's already true right now with jit being the only exception)

@Zzzabiyaka Thanks for fixing, it is really an issue! For the new jit options, how about adding "llvm-lazy-jit" and "llvm-eager-jit" options to make it more clear? Using "-" other than "_" is because we already have "fast-interp"/"classic-interp"/"fast-jit".
And had better just keep current "jit" option which tests both "llvm-lazy-jit" and "llvm-eager-jit", as it is used in some other scripts and it is also an shortcut to test two llvm jit modes.

Zzzabiyaka:
Hi, ok, shouldn't be a problem.

I will just set a different (2x more) timeout for jit in the ci

PR #2423, opened 2023-08-04
Enable running spec tests on Windows

loganek:
We don't enable those tests in CI yet as not all of them are passing

### DIFF
--- a/tests/wamr-test-suites/spec-test-script/runtest.py
+++ b/tests/wamr-test-suites/spec-test-script/runtest.py
@@ -5,22 +5,21 @@ from __future__ import print_function
 import argparse
 import array
 import atexit
-import fcntl
 import math
 import os
-# Pseudo-TTY and terminal manipulation
-import pty
 import re
 import shutil
 import struct
 import subprocess
 import sys
 import tempfile
-import termios
 import time
+import threading
 import traceback
 from select import select
+from queue import Queue
 from subprocess import PIPE, STDOUT, Popen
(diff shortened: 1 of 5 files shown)


loganek:
@wenyongh looks like the build failure is not related to this change. Could we possibly rebase the `dev/wasi-libc-windows` branch onto the `main` branch? I guess there were some changes in main that fix the issue.

wenyongh:
> @wenyongh looks like the build failure is not related to this change. Could we possibly rebase the `dev/wasi-libc-windows` branch onto the `main` branch? I guess there were some changes in main that fix the issue.

@loganek, done with PR #2426.

wenyongh (commented):
LGTM

tests/wamr-test-suites/spec-test-script/runtest.py
```diff
@@ -101,35 +135,53 @@ def __init__(self, args, no_pty=False):
             self.stdin = os.fdopen(master, 'r+b', 0)
             self.stdout = self.stdin
 
+        if platform.system().lower() == "windows":
```
lum1n0us, line 138: Just be curious, is it possible to have a cross-platform solution for reading stdout?

loganek: Good question. The only cross-platform I could think of was by using a thread and a queue (just like it was implemented in this PR for Windows); but I did not use this technique for other operating systems because it turned to be significantly (2 - 3 x) slower, so I kept the approach with `select` for non-windows platforms.

PR #2428, opened 2023-08-06
Remove unnecessary and extra zero length check in mem functions' macro

cngzhnp:
In these 3 functions, no need to do extra check for length is equal zero or not because it was already done inside of these functions. 

PS : I added this check into `b_memcpy_wa` function which was missing.


wenyongh (approved):
LGTM

PR #2431, opened 2023-08-07
Stripping static and shared libraries of iwasm to reduce binary size

cngzhnp:
Reduce iwasm static and shared library size when build type is "Release". It remains as it is when build type was selected Debug.

Before stripping the shared and static library size:

> -rwxr-xr-x  1 546184 Aug  7 15:10 libiwasm.so
> -rw-r--r--  1 932658 Aug  7 15:10 libvmlib.a

After stripping the shared and static library size:

> -rwxr-xr-x  1 485536 Aug  7 15:09 libiwasm.so
> -rw-r--r--  1 489030 Aug  7 15:09 libvmlib.a

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -109,8 +109,7 @@ set (WAMR_ROOT_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 
 # Set the strip command based on the system (GNU or Clang)
 if (CMAKE_STRIP)
-    set(CMAKE_C_STRIP_FLAGS "${CMAKE_STRIP}")
-    set(CMAKE_CXX_STRIP_FLAGS "${CMAKE_STRIP}")
+    set (CMAKE_STRIP_FLAGS "--strip-all")
 endif ()
 
 include (${WAMR_ROOT_DIR}/build-scripts/runtime_lib.cmake)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -141,6 +147,15 @@ endif ()
 
 install (TARGETS iwasm_static ARCHIVE DESTINATION lib)
 
+# If it's a Release build, strip the static library
+if (CMAKE_BUILD_TYPE STREQUAL "Release")
+  # Strip static library
+  message(STATUS "Stripping static library after build!")
+  add_custom_command(TARGET iwasm_static POST_BUILD
+      COMMAND ${CMAKE_STRIP} ${CMAKE_C_STRIP_FLAGS} $<TARGET_FILE:iwasm_static>
```
wenyongh, line 155: Here `${CMAKE_C_STRIP_FLAGS}` is also set to "${CMAKE_STRIP}", see L112, so the command is `${CMAKE_STRIP} ${CMAKE_STRIP} $<TARGET_FILE:iwasm_static>`, seems strange? Suppose that ${CMAKE_STRIP} is `strip`

CMakeLists.txt
```diff
@@ -141,6 +147,15 @@ endif ()
 
 install (TARGETS iwasm_static ARCHIVE DESTINATION lib)
 
+# If it's a Release build, strip the static library
+if (CMAKE_BUILD_TYPE STREQUAL "Release")
```
wenyongh, line 151: Had better also check CMAKE_STRIP, e.g. `if (CMAKE_STRIP AND CMAKE_BUILD_TYPE STREQUAL "Release")`?

CMakeLists.txt
```diff
@@ -162,3 +177,12 @@ install (FILES
     ${WAMR_ROOT_DIR}/core/iwasm/include/wasm_export.h
     ${WAMR_ROOT_DIR}/core/iwasm/include/lib_export.h
     DESTINATION include)
+
+# If it's a Release build, strip the shared library
+if (CMAKE_BUILD_TYPE STREQUAL "Release")
```
wenyongh, line 182: Same as above, had better also check CMAKE_STRIP?

CMakeLists.txt
cngzhnp: You are right sorry, then I set `CMAKE_STRIP_FLAGS ` as `"--strip-all"` and remove language dependency naming convention. Could you please review it again?

CMakeLists.txt
cngzhnp: Yes, I added `CMAKE_STRIP` check both conditions that I also check Release type.

CMakeLists.txt
cngzhnp: Yes, I added `CMAKE_STRIP` check both conditions that I also check Release type.

wenyongh (commented):
LGTM

PR #2436, opened 2023-08-08
module instance context api

yamt:
* introduce module instance context api
* make wasi use it
* add a sample

https://github.com/bytecodealliance/wasm-micro-runtime/issues/2460

todo:
* ~~rename functions:~~
  ```
  wasm_runtime_create_context_key
  wasm_runtime_destroy_context_key
  wasm_runtime_set_context
  wasm_runtime_set_context_spread
  wasm_runtime_get_context
  ```
* ~~make it a build-time option:~~
  ```
  WAMR_BUILD_MODULE_INST_CONTEXT
  WASM_ENABLE_MODULE_INST_CONTEXT
  ```


yamt:
should be unified with wasm_runtime_set_custom_data?

wenyongh:
@yamt The code change is a little big and not sure whether it is good to use the module instance context api, shall we keep this PR unmerged before we release 1.2.3? And could you open an issue to describe module instance context api and why we need to use it? Had better let more people can know that and participate the discussion.

yamt:
> @yamt The code change is a little big and not sure whether it is good to use the module instance context api, shall we keep this PR unmerged before we release 1.2.3? And could you open an issue to describe module instance context api and why we need to use it? Had better let more people can know that and participate the discussion.

ok!

yamt:
> > @yamt The code change is a little big and not sure whether it is good to use the module instance context api, shall we keep this PR unmerged before we release 1.2.3? And could you open an issue to describe module instance context api and why we need to use it? Had better let more people can know that and participate the discussion.
> 
> ok!

i created an issue. https://github.com/bytecodealliance/wasm-micro-runtime/issues/2460

core/iwasm/interpreter/wasm_runtime.h
```diff
@@ -212,6 +212,7 @@ typedef struct CApiFuncImport {
 
 /* The common part of WASMModuleInstanceExtra and AOTModuleInstanceExtra */
 typedef struct WASMModuleInstanceExtraCommon {
+    void *contexts[WASM_MAX_INSTANCE_CONTEXTS];
```
wenyongh, line 215: Does it mean that the wasi_ctx always occupies the first slot with key g_wasi_context_key, and user's contexts occupy the following?

core/iwasm/common/wasm_native.c
```diff
@@ -394,6 +407,155 @@ wasm_native_unregister_natives(const char *module_name,
     return false;
 }
 
+static uint32
+context_key_to_idx(void *key)
```
wenyongh, line 411: Suggest to make this feature configurable with cmake variable and compiler macro, and it is enabled only when libc wasi is enabled? Or just using `cmake -DWAMR_BUILID_LIBC_WASI/UVWASI=1` to control it? Since the feature may introduce more memory usage and binary size, some host embedder may expect the footprint as small as possible.

core/iwasm/common/wasm_native.c
```diff
@@ -394,6 +407,155 @@ wasm_native_unregister_natives(const char *module_name,
...
+{
```
wenyongh, line 412: And not very sure what is the difference from wasm_runtime_set_custom_data, which also binds a custom pointer to the module instance? Is it because that `wasm_runtime_module_instance_{set|get}_context` can set/get context with key while `wasm_runtime_{set|get}_custom_data` can not?

core/iwasm/common/wasm_runtime_common.h
```diff
@@ -939,6 +939,27 @@ WASM_RUNTIME_API_EXTERN bool
 wasm_runtime_unregister_natives(const char *module_name,
                                 NativeSymbol *native_symbols);
 
+/* See wasm_export.h for description */
+WASM_RUNTIME_API_EXTERN void *
+wasm_runtime_module_instance_context_key_create(
```
wenyongh, line 944: The API name may be a little long and confusing? How about:
```C
wasm_runtime_create_custom_ctx_key
wasm_runtime_destroy_custom_ctx_key
wasm_runtime_set_custom_context   (or set_custom_ctx)
wasm_runtime_set_custom_context_spread
wasm_runtime_get_custom_context

wasm_native_create_custom_ctx_key
wasm_native_destroy_custom_ctx_key
wasm_native_set_custom_context   (or set_custom_ctx)
wasm_native_set_custom_context_spread
wasm_native_get_custom_context
wasm_native_call_custom_context_dtors
```

wenyongh: Or wasm_runtime_create_module_isnt_ctx_key, wasm_runtime_set_module_inst_ctx?

core/iwasm/interpreter/wasm_runtime.h
yamt: if wasi is enabled, probably.

core/iwasm/common/wasm_native.c
yamt: i can make it a build-time option if desirable.

core/iwasm/common/wasm_native.c
yamt: > Is it because that wasm_runtime_module_instance_{set|get}_context can set/get context with key while wasm_runtime_{set|get}_custom_data can not?

right.
i want something which can be used by multiple subsystems.

besides that, i want it to have a cleanup callback on instance destory.

core/iwasm/common/wasm_runtime_common.h
yamt: i agree it's long.
i'm not sure if it's confusing.

tonibofarull (approved):

core/iwasm/common/wasm_native.c
tonibofarull: I'd like to control wasi-nn context with this API for both libc wasi and libc builtin.
I think that the feature can also be helpful for a generic native library.

core/iwasm/common/wasm_runtime_common.h
wenyongh: yes, but adding prefix `wasm_runtime` really make me confusing, e.g. wasm_runtime_module_instance_set_context, how about:
```C
wasm_module_instance_create_context_key
wasm_module_instance_destroy_context_key
wasm_module_instance_set_context
wasm_module_instance_set_context_spread
wasm_module_instance_get_context
```

And using `cmake -DWAMR_BUILD_MODULE_INST_CONTEXT=1/0` and macro `WASM_ENABLE_MODULE_INST_CONTEXT` to control the feature.

The feature can be enabled no matter whether libc-wasi/libc-builtin is enabled or not.

And when libc-wasi or libc-uvwasi is enabled, the feature will be also enable automatically.

xujuntwt95329: Agree with @wenyongh

yamt: > yes, but adding prefix `wasm_runtime` really makes me confused, e.g. wasm_runtime_module_instance_set_context, how about:

i thought we were prefixing all api with `wasm_runtime_`.

> 
> ```c
> wasm_module_instance_create_context_key
> wasm_module_instance_destroy_context_key
> wasm_module_instance_set_context
> wasm_module_instance_set_context_spread
> wasm_module_instance_get_context
> ```
> 
> And using `cmake -DWAMR_BUILD_MODULE_INST_CONTEXT=1/0` and macro `WASM_ENABLE_MODULE_INST_CONTEXT` to control the feature.

instance vs INST will make me confused.
let's be consistent.
which do you prefer?

> 
> The feature can be enabled no matter whether libc-wasi/libc-builtin is enabled or not.
> 
> And when libc-wasi or libc-uvwasi is enabled, the feature will be also enable automatically.

wenyongh: > > yes, but adding prefix `wasm_runtime` really makes me confused, e.g. wasm_runtime_module_instance_set_context, how about:
> 
> i thought we were prefixing all api with `wasm_runtime_`.

Do you mean somewhat like `wasm_runtime_module_inst_create_context_key` and `wasm_runtime_module_inst_set_context`?
   
> > And using `cmake -DWAMR_BUILD_MODULE_INST_CONTEXT=1/0` and macro `WASM_ENABLE_MODULE_INST_CONTEXT` to control the feature.
> 
> instance vs INST will make me confused. let's be consistent. which do you prefer?
> 

`module_inst` in API names and `MODULE_INST` in cmake variable and compiler macro are better for me, and if you prefer using `wasm_runtime_` prefix for API:
```C
wasm_runtime_module_inst_create_context_key
wasm_runtime_module_inst_destroy_context_key
wasm_runtime_module_inst_set_context
wasm_runtime_module_inst_set_context_spread
wasm_runtime_module_inst_get_context

WAMR_BUILD_MODULE_INST_CONTEXT
WASM_ENABLE_MODULE_INST_CONTEXT
```

Not sure if it is good to others.

yamt: > > > yes, but adding prefix `wasm_runtime` really makes me confused, e.g. wasm_runtime_module_instance_set_context, how about:
> > 
> > 
> > i thought we were prefixing all api with `wasm_runtime_`.
> 
> Do you mean somewhat like `wasm_runtime_module_inst_create_context_key` and `wasm_runtime_module_inst_set_context`?

as most of the existing apis in wasm_export.h seem to have the `wasm_runtime_` prefix,
i thought there was a project-wide policy.
it isn't my own preference.
if you prefer shorter names, it's also ok for me.

also, module_instance/module_inst part is not essential for me either.
after all, many of our api is implicitly about module instances. (eg. wasm_runtime_instantiate, wasm_runtime_set_custom_data, etc)

wenyongh: > > Do you mean somewhat like `wasm_runtime_module_inst_create_context_key` and `wasm_runtime_module_inst_set_context`?
> 
> as most of the existing apis in wasm_export.h seem to have the `wasm_runtime_` prefix, i thought there was a project-wide policy. it isn't my own preference. if you prefer shorter names, it's also ok for me.
> 

Agree.

> also, module_instance/module_inst part is not essential for me either. after all, many of our api is implicitly about module instances. (eg. wasm_runtime_instantiate, wasm_runtime_set_custom_data, etc)

So you mean:
```C
wasm_runtime_create_context_key
wasm_runtime_destroy_context_key
wasm_runtime_set_context
wasm_runtime_set_context_spread
wasm_runtime_get_context
```
It is OK for me, looks concise.

@xujuntwt95329 @tonibofarull What's you opinion? Is it good to you?

ayakoakasaka: There is no problem if there is an implicit rule that `module_instance` is omitted.

xujuntwt95329: It's also OK to me, thanks

core/iwasm/common/wasm_native.h
```diff
@@ -68,6 +68,32 @@ bool
 wasm_native_unregister_natives(const char *module_name,
                                NativeSymbol *native_symbols);
 
+#if WASM_ENABLE_MODULE_INST_CONTEXT != 0
+void *
+wasm_native_create_context_key(void (*dtor)(wasm_module_inst_t inst,
```
wenyongh, line 73: Had better use `WASMModuleInstanceCommon *` instead of `wasm_module_inst_t`

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -5681,3 +5660,36 @@ wasm_runtime_is_import_global_linked(const char *module_name,
     return false;
 #endif
 }
+
+#if WASM_ENABLE_MODULE_INST_CONTEXT != 0
+void *
+wasm_runtime_create_context_key(void (*dtor)(wasm_module_inst_t inst,
```
wenyongh, line 5666: Same as above

core/iwasm/common/wasm_runtime_common.h
```diff
@@ -939,6 +939,25 @@ WASM_RUNTIME_API_EXTERN bool
 wasm_runtime_unregister_natives(const char *module_name,
                                 NativeSymbol *native_symbols);
 
+/* See wasm_export.h for description */
+WASM_RUNTIME_API_EXTERN void *
+wasm_runtime_create_context_key(void (*dtor)(wasm_module_inst_t inst,
```
wenyongh, line 944: Same as above

wenyongh (commented):
LGTM

core/iwasm/common/wasm_native.h
yamt: what are the principles?

wenyongh: Hi, in the original design, the wasm_module_t/wasm_module_inst_t/wasm_exec_env_t/.. are only used only in wasm_export.h, native libraries (libc-builtin, libc-wasi and more in core/iwasm/libraries) and main.c, and the detailed structure name (WASMModule/WASMModuleInstance/WASMExecEnv/..) are used in runtime internal. Though there may be some code inside runtime using the the formers, we had better obey the rules.

yamt: done

PR #2437, opened 2023-08-08
Auto-check wrgsbase in cmake script

wenyongh:
Auto-check whether `WRGSBASE` instruction is supported in linux x86-64 in the
cmake script. And if not, disable writing x86 GS register.


ThomasDevoogdt:
@wenyongh, this creates an error for cross compilation targets, can you make this try_run conditionally?
See https://cmake.org/cmake/help/latest/command/try_run.html#behavior-when-cross-compiling.

PR #2445, opened 2023-08-10
fix some memory leak issue

xujuntwt95329:


core/shared/mem-alloc/ems/ems_kfc.c
```diff
@@ -139,10 +139,14 @@ gc_destroy_with_pool(gc_handle_t handle)
 
     if (heap->extra_info_node_cnt > 0) {
         for (i = 0; i < heap->extra_info_node_cnt; i++) {
+            extra_info_node_t *node = heap->extra_info_nodes[i];
 #if BH_ENABLE_GC_VERIFY != 0
             os_printf("Memory leak detected: gc object [%p] not claimed\n",
-                      heap->extra_info_nodes[i]->obj);
+                      node->obj);
 #endif
+            if (heap->is_reclaim_enabled) {
```
wenyongh, line 147: Seems no need to add check here? If there is extra_info_node, the heap must be a gc heap, and in wasm_exec_env.c, it must have been reclaim enabled by calling mem_allocator_enable_gc_reclaim.
How about changing to bh_assert(heap->is_reclaim_enabled)?

xujuntwt95329: Done, thanks

PR #2447, opened 2023-08-11
wasi-nn: Improve TPU support

tonibofarull:
1. Allow TPU and GPU support at the same time.
2. Add Dockerfile to run example with [Coral USB](https://coral.ai/products/accelerator/).


core/iwasm/libraries/wasi-nn/src/wasi_nn_tensorflowlite.cpp
```diff
@@ -216,7 +216,15 @@ tensorflowlite_init_execution_context(void *tflite_ctx, graph g,
...
-#elif defined(WASI_NN_ENABLE_EXTERNAL_DELEGATE)
+#else
+            NN_WARN_PRINTF("GPU not enabled.");
+            use_default = true;
+#endif
+            break;
+        }
+        case tpu:
+        {
+#if defined(WASI_NN_ENABLE_EXTERNAL_DELEGATE)
```
wenyongh, line 227: We usually define macros in core/config.h, set them to 0, and use `cmake -DWAMR_BUILD_XXX=1` to set the macros to 1 (in build-scripts/config_common.cmake or other files), and then use `#if WASM_ENABLE_XXX != 0` in source code instead of `#ifdef` or `#if defined`. Not sure whether you want to modify `WASI_NN_ENABLE_GPU`, `WASI_NN_ENABLE_EXTERNAL_DELEGATE` in that way?

tonibofarull: done https://github.com/bytecodealliance/wasm-micro-runtime/pull/2447/commits/d7a3595f4794bdd81d28c5fa7400feb0347d9df5

core/iwasm/libraries/wasi-nn/cmake/Findtensorflow_lite.cmake
```diff
@@ -18,12 +18,16 @@ if(NOT EXISTS ${TENSORFLOW_LITE})
...
 
-  if(WASI_NN_ENABLE_GPU EQUAL 1)
+  if(WAMR_BUILD_WASI_NN_ENABLE_GPU EQUAL 1)
     # Tensorflow specific:
     # * https://www.tensorflow.org/lite/guide/build_cmake#available_options_to_build_tensorflow_lite
     set (TFLITE_ENABLE_GPU ON)
   endif()
 
+  if (CMAKE_SIZEOF_VOID_P EQUAL 4)
+    set (TFLITE_ENABLE_XNNPACK OFF)
```
yamt, line 28: @tonibofarull do you remember why?

tonibofarull: I was getting compile errors for 32 bits. I don't remember the exact reason.

yamt: was it i386? arm? something else?

tonibofarull: ARM

yamt: ok. thank you.

PR #2451, opened 2023-08-11
wamrc enhancement: more user friendly logic for printing help info

TianlongLiang:
The printed usage message in print_help() may need to be further polished.

After this update, the wamrc can be used to 
1. either compile wasm file to aot file like before 
2. or be used to print target-machine-related help info.

But not at the same time

Under the current modification, the target-abi and arch-related info needs to be queried independently, can't query --target-abi and --arch at the same time. Usage for how to query target-machine-related help info:

```shell
./wamrc --target-abi=help
./wamrc --target=help
# Like before, if want to query cpu/cpu-features, have to clearly choose what target/cpu are using
./wamrc --target=x86_64 --cpu=help
./wamrc --target=x86_64 --cpu=help --cpu-features=+help
```

Maybe there is a more natural or user-friendly method, any suggestions? Any input would be appreciated

### DIFF
--- a/core/iwasm/compilation/aot_llvm.c
+++ b/core/iwasm/compilation/aot_llvm.c
@@ -2745,7 +2745,8 @@ aot_create_comp_context(const AOTCompData *comp_data, aot_comp_option_t option)
             goto fail;
         }
 
-        /* If only to create target machine for querying information, early stop */
+        /* If only to create target machine for querying information, early stop
+         */
         if ((arch && !strcmp(arch, "help")) || (abi && !strcmp(abi, "help"))
             || (cpu && !strcmp(cpu, "help"))
             || (features && !strcmp(features, "+help"))) {
--- a/core/iwasm/compilation/aot_llvm.c
+++ b/core/iwasm/compilation/aot_llvm.c
@@ -2746,11 +2746,10 @@ aot_create_comp_context(const AOTCompData *comp_data, aot_comp_option_t option)
         }
 
         /* If only create target machine for querying information, early stop */
-        if ((arch != NULL && !strcmp(arch, "help"))
-            || (abi != NULL && !strcmp(abi, "help"))
-            || (cpu != NULL && !strcmp(cpu, "help"))
-            || (features != NULL && !strcmp(features, "+help"))) {
-            aot_set_last_error(
+        if ((arch && !strcmp(arch, "help")) || (abi && !strcmp(abi, "help"))
+            || (cpu && !strcmp(cpu, "help"))
+            || (features && !strcmp(features, "+help"))) {
+            LOG_DEBUG(
                 "create LLVM target machine only for printing help infos.");
             goto fail;
         }
--- a/core/iwasm/compilation/aot_llvm.c
+++ b/core/iwasm/compilation/aot_llvm.c
@@ -2745,12 +2745,12 @@ aot_create_comp_context(const AOTCompData *comp_data, aot_comp_option_t option)
             goto fail;
         }
 
-        /* If only create target machine for querying information, early stop */
+        /* If only to create target machine for querying information, early stop */
         if ((arch && !strcmp(arch, "help")) || (abi && !strcmp(abi, "help"))
             || (cpu && !strcmp(cpu, "help"))
             || (features && !strcmp(features, "+help"))) {
             LOG_DEBUG(
-                "create LLVM target machine only for printing help infos.");
+                "create LLVM target machine only for printing help info.");
             goto fail;
         }
     }
--- a/core/iwasm/compilation/aot_llvm.c
+++ b/core/iwasm/compilation/aot_llvm.c
@@ -2744,6 +2744,16 @@ aot_create_comp_context(const AOTCompData *comp_data, aot_comp_option_t option)
             aot_set_last_error("create LLVM target machine failed.");
             goto fail;
         }
+
+        /* If only create target machine for querying information, early stop */
+        if ((arch != NULL && !strcmp(arch, "help"))
+            || (abi != NULL && !strcmp(abi, "help"))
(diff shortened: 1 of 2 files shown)


core/iwasm/compilation/aot_llvm.c
```diff
@@ -2744,6 +2744,16 @@ aot_create_comp_context(const AOTCompData *comp_data, aot_comp_option_t option)
...
             goto fail;
         }
+
+        /* If only create target machine for querying information, early stop */
+        if ((arch != NULL && !strcmp(arch, "help"))
+            || (abi != NULL && !strcmp(abi, "help"))
+            || (cpu != NULL && !strcmp(cpu, "help"))
+            || (features != NULL && !strcmp(features, "+help"))) {
+            aot_set_last_error(
+                "create LLVM target machine only for printing help infos.");
```
wenyongh, line 2754: No need to set_last_error, or the error will be print out.

core/iwasm/compilation/aot_llvm.c
```diff
@@ -2744,6 +2744,16 @@ aot_create_comp_context(const AOTCompData *comp_data, aot_comp_option_t option)
             aot_set_last_error("create LLVM target machine failed.");
             goto fail;
         }
+
+        /* If only create target machine for querying information, early stop */
+        if ((arch != NULL && !strcmp(arch, "help"))
```
wenyongh, line 2749: How about `(arch && !strcmp(arch, "help"))`, `(abi && !strcmp(abi, "help"))`?

wamr-compiler/main.c
```diff
@@ -193,6 +193,11 @@ print_help()
     printf("Examples: wamrc -o test.aot test.wasm\n");
     printf("          wamrc --target=i386 -o test.aot test.wasm\n");
     printf("          wamrc --target=i386 --format=object -o test.o test.wasm\n");
+    printf("Usage for print architecture-related help infos: wamrc [--target/--cpu/--target-abi]=help\n");
+    printf("                                                 wamrc [--cpu-features]=+help\n");
```
wenyongh, line 197: No need show help again here, `--target=help`/`--target-abi=help`/.. have been mentioned in each option:
```
Use --target=help to list supported targets
Use --target-abi=help to list all the ABI supported
Use --cpu=help to list all the CPU supported
Use --cpu-features=+help to list all the features supported 
```

wamr-compiler/main.c
```diff
@@ -193,6 +193,11 @@ print_help()
...
+    printf("Examples: wamrc --target-abi=help\n");
+    printf("          wamrc --target=x86_64 --cpu=help\n");
+    printf("          wamrc --target=x86_64 --cpu=alderlake --cpu-features=+help\n");
```
wenyongh, line 200: How about just mentioning `wamrc --target-abi=help`, `wamrc --target=x86_64 --cpu=help`?
And no need print string `"Examples:` again.

wamr-compiler/main.c
```diff
@@ -516,7 +539,13 @@ main(int argc, char *argv[])
...
     }
 
-    if (argc == 0 || !out_file_name)
+    /* Normal usage, generate output file */
+    if (!use_dummy_wasm && (argc == 0 || !out_file_name))
+        PRINT_HELP_AND_EXIT();
+
+    /* When use wamrc to print architecture related help info, use dummy_wasm
+     * and don't generate anything */
+    if (use_dummy_wasm && out_file_name)
```
wenyongh, line 548: My understanding is that the below commands are allowed:
```bash
wamrc --target=help
wamrc --target=help -o test.aot test.wasm
wamrc --target=help -o test.aot
wamrc --target=help test.wasm
```
Should remove these lines?

TianlongLiang: My modification will disable passing input/output filename while printing help info:
```shell
# can't do this, will print usage 
wamrc --target=help -o test.aot test.wasm
wamrc --target=help -o test.aot
wamrc --target=help test.wasm
```
Since only want help info, there is no point in passing an input/output file name.

wamr-compiler/main.c
```diff
@@ -296,6 +298,12 @@ resolve_segue_flags(char *str_flags)
     return segue_flags;
 }
 
+/* When print help infos for target/cpu/target-abi/cpu-features, load this dummy
```
wenyongh, line 301: info?

wamr-compiler/main.c
```diff
@@ -296,6 +298,12 @@ resolve_segue_flags(char *str_flags)
...
+ * wasm file content rather than from an input file, the dummy wasm file content
+ * is: (module) */
```
wenyongh, line 303: how about "is: magic header + version number"?

core/iwasm/compilation/aot_llvm.c
```diff
@@ -2744,6 +2744,15 @@ aot_create_comp_context(const AOTCompData *comp_data, aot_comp_option_t option)
             aot_set_last_error("create LLVM target machine failed.");
             goto fail;
         }
+
+        /* If only create target machine for querying information, early stop */
```
wenyongh, line 2748: How about "If only to"

core/iwasm/compilation/aot_llvm.c
```diff
@@ -2744,6 +2744,15 @@ aot_create_comp_context(const AOTCompData *comp_data, aot_comp_option_t option)
...
+        if ((arch && !strcmp(arch, "help")) || (abi && !strcmp(abi, "help"))
+            || (cpu && !strcmp(cpu, "help"))
+            || (features && !strcmp(features, "+help"))) {
+            LOG_DEBUG(
+                "create LLVM target machine only for printing help infos.");
```
wenyongh, line 2753: info?

wenyongh (approved):
LGTM

Issue #2412, opened 2023-07-31
LLVM ERROR: failed to perform tail call elimination with MIPS AOT


eloparco:
Hello, I'm getting this error when trying to compile for MIPS:

```bash
$ wasm-micro-runtime/wamr-compiler/build/wamrc --enable-multi-thread --disable-simd --target=mips --size-level=1 -o pigz.aot pigz.wasm
Create AoT compiler with:
  target:        mips
  target cpu:    
  target triple: mips-pc-linux-gnu
  cpu features:  
  opt level:     3
  size level:    1
  output format: AoT file
LLVM ERROR: failed to perform tail call elimination on a call site marked musttail
78805 Abort trap: 6           wasm-micro-runtime/wamr-compiler/build/wamrc --enable-multi-thread --disable-simd --target=mips --size-level=1 -o "pigz.aot" "pigz.wasm"
```

The wasm file is [pigz.wasm.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/12216660/pigz.wasm.zip) and it's coming from https://github.com/bytecodealliance/wasm-parallel-gzip.
It works fine with other targets (I tried aarch64, armv7, x86_64).

eloparco:
I just checked out version `1.2.2` and there the compilation is fine, so the issue was introduced after that.

loganek:
I just did a quick git bisect check; looks like the PR #2244 caused the regression.

yamt:
i tested https://github.com/bytecodealliance/wasm-micro-runtime/pull/2457 with the given wasm file lightly.
(lightly == just aot compile. i have no mips env.)

PR #2457, opened 2023-08-14
aot: disable musttail for mips

yamt:
While i'm not familiar with mips abi, it's reported that it (at least sometimes) doesn't work.
This fixes https://github.com/bytecodealliance/wasm-micro-runtime/issues/2412

### DIFF
--- a/core/iwasm/compilation/aot_llvm.c
+++ b/core/iwasm/compilation/aot_llvm.c
@@ -146,6 +146,13 @@ aot_target_precheck_can_use_musttail(const AOTCompContext *comp_ctx)
          */
         return false;
     }
+    if (!strcmp(comp_ctx->target_arch, "mips")) {
+        /*
+         * cf.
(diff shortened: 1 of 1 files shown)


Zzzabiyaka:
I'm not familiar with the context so the questions can be silly enough:

- Did we have must tail in 1.2.2 release? 
- If so, are we sure no performance regression can happen because of disabling it 

By the way, we have MIPS env so we can help with testing this PR and I wouldn't merge before testing that it not only compiles but works

wenyongh:
@eloparco @yamt Does this PR fix issue #2412? Shall I merge this PR?

yamt:
> @eloparco @yamt Does this PR fix issue #2412? Shall I merge this PR?

yes. at least this fixes the reported aot compilation error.

wenyongh:
> > @eloparco @yamt Does this PR fix issue #2412? Shall I merge this PR?
> 
> yes. at least this fixes the reported aot compilation error.

OK, let's merge it.

PR #2467, opened 2023-08-15
Fix issues reported by Coverity and clear warnings

wenyongh:


core/iwasm/compilation/aot_emit_aot_file.c
```diff
@@ -2642,10 +2649,6 @@ aot_resolve_stack_sizes(AOTCompContext *comp_ctx, AOTObjectData *obj_data)
                     goto fail;
                 }
             }
-            if (addr > UINT32_MAX) {
```
wenyongh, line 2645: Dead code, addr has been checked in L2624:
```C
            if (strcmp(sec_name, aot_stack_sizes_section_name) || addr != 0) {
                aot_set_last_error(
                    "stack_sizes found at an unexpected location.");
                goto fail;
            }
```
So here addr must be 0.

core/shared/platform/common/posix/posix_socket.c
```diff
@@ -275,7 +275,7 @@ os_socket_recv_from(bh_socket_t socket, void *buf, unsigned int len, int flags,
             return -1;
         }
     }
-    else {
+    else if (src_addr) {
```
wenyongh, line 278: src_addr is checked at L272, meaning it may be NULL here.

PR #2468, opened 2023-08-15
interpreter: fix a wrong optimization

yamt:
this allows the interpreters pass the following test.
```
(module
  (func (export "_start")
    i32.const 0
    i32.load offset=0xffff_fffc
    i32.const 0x1234_5678
    i32.ne
    if
      unreachable
    end
  )
  (memory (export "memory") 65536)
  (data (i32.const 0xffff_fffc) "\78\56\34\12")
)
```


wenyongh:
The issue should have been resolved with #3209, so close this PR.

PR #2470, opened 2023-08-15
Add sample pre-commit hook 

Zzzabiyaka:
Using pre-commit hooks helps us to decrease CI load and improve developers' experience. 

We can discuss adding recommendation to use this hook to CONTRIBUTING.MD if people find it useful 

Also we can consider recommending using `setup.sh` and expand it 

possible ideas how we can expand it 
- installing wasi-sdk 
- Feel free to propose anything that might improve onboarding experience 

All the feedback is much appreciated

### DIFF
--- a/ci/pre_commit_hook_sample
+++ b/ci/pre_commit_hook_sample
@@ -1,0 +1,32 @@
+#!/bin/bash
+
+# Copyright (C) 2023 Amazon.com Inc. or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# This is a sample of pre-commit hook that can be used to make your code fit the WAMR CI code style requirements.
+# You need to have clang-format-12 installed to use this hook.
+# To add this pre-commit hook, copy it to <path_to_wamr>/.git/hooks/pre-commit
+# (you don't need any extensions here)
+
+# Function to check if a file has a C or C++ extension
+
+is_c_or_cpp_file() {
+    file="$1"
+    if [[ "$filename" =~ \.(h|c|cpp)$  ]]; then
+        return 0
+    else
+        return 1
+    fi
+}
+
+# Loop through staged files and apply command "abc" to C and C++ files
+for staged_file in $(git diff --cached --name-only); do
+    if is_c_or_cpp_file "$staged_file"; then
+        clang-format-12 -Werror --style file --dry-run "$staged_file" 2>/dev/null
+        if [ $? -ne 0 ]; then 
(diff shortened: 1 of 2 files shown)


Zzzabiyaka:
> I think that's really helpful, thanks a lot for making the change. I wonder if we could have some sort of `setup` script that developers can execute once after cloning the repo and it will copy the hook script to the right place.

we can but let's think what else can we put to this script (probably in a separate issue and not in this PR but up to you)

loganek:
> > I think that's really helpful, thanks a lot for making the change. I wonder if we could have some sort of `setup` script that developers can execute once after cloning the repo and it will copy the hook script to the right place.
> 
> we can but let's think what else can we put to this script (probably in a separate issue and not in this PR but up to you)

sure, it was just a suggestion. I think though the setup script can start from a simple copying the hook and extended in the future if needed, we don't have to make it perfect at first.

ci/pre_commit_hook_sample
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
```
wenyongh, line 2: Should add license header

ci/setup.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
```
wenyongh, line 2: Should add license header

ci/setup.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
loganek, line 1: Could we also update README.md and/or CONTRIBUTING.md file to mention the script?

Zzzabiyaka: Yeah I've opened this PR to see suggestions where it can be added (readme doesn't fit well as for me, I'd put to contributing.md)

ci/pre_commit_hook_sample
Zzzabiyaka: Thank you!

Actually it's a fair point of what we can try to add to hook (check that all files (c, cpp,h, sh at least) have the license

wenyongh: We normally add the license header in the source files and scripts (shell, cmake, etc.), and don't add it in the document. It should be ok here not to add the license header if we treat it as document and it really hasn't license issue.

Zzzabiyaka: I've already added it

I'm just curious would we like to have this check in pre-commit hook or not (let's say grep source file for `apache-2.0` to see that license header is added and don't allow to commit without it)

wenyongh: Yes, it would be great to have the license header check in pre-commit hook.

PR #2472, opened 2023-08-16
Add mutex stress test 

Zzzabiyaka:
as a part of stress-testing we want to ensure that mutex implementation is working correctly and protecting shared resource to be allocated from other threads when mutex is locked. 

This test covers the most common situations that happen when some program uses mutexes like locks from various threads, locks from the same thread etc

If you have any other cases that left uncovered please mention them in the comments 

cc @loganek @g0djan @eloparco

### DIFF
--- a/.github/workflows/nightly_run.yml
+++ b/.github/workflows/nightly_run.yml
@@ -599,6 +599,11 @@ jobs:
         run: bash build.sh --sysroot "$SYSROOT_PATH"
         working-directory: ./core/iwasm/libraries/lib-wasi-threads/test/
 
+      - name: Build WASI thread stress tests
+        if: matrix.test_option == '$WASI_TEST_OPTIONS'
+        run: bash build.sh --sysroot "$SYSROOT_PATH"
+        working-directory: ./core/iwasm/libraries/lib-wasi-threads/stress-test/
+
       - name: build socket api tests
         if: matrix.test_option == '$WASI_TEST_OPTIONS'
         run: bash build.sh
--- a/tests/wamr-test-suites/wasi-test-script/run_wasi_tests.sh
+++ b/tests/wamr-test-suites/wasi-test-script/run_wasi_tests.sh
@@ -15,25 +15,22 @@ readonly IWASM_CMD="${WORK_DIR}/../../../../product-mini/platforms/${PLATFORM}/b
     --allow-resolve=google-public-dns-a.google.com \
     --addr-pool=::1/128,127.0.0.1/32"
 
-readonly IWASM_CMD_STRESS="${IWASM_CMD} --max-threads=8"
+readonly IWASM_CMD_STRESS="${IWASM_CMD} --max-threads=12"
 readonly WAMRC_CMD="${WORK_DIR}/../../../../wamr-compiler/build/wamrc"
 readonly C_TESTS="tests/c/testsuite/"
 readonly ASSEMBLYSCRIPT_TESTS="tests/assemblyscript/testsuite/"
 readonly THREAD_PROPOSAL_TESTS="tests/proposals/wasi-threads/"
 readonly THREAD_INTERNAL_TESTS="${WAMR_DIR}/core/iwasm/libraries/lib-wasi-threads/test/"
+readonly THREAD_STRESS_TESTS="${WAMR_DIR}/core/iwasm/libraries/lib-wasi-threads/stress-test/"
 readonly LIB_SOCKET_TESTS="${WAMR_DIR}/core/iwasm/libraries/lib-socket/test/"
-readonly STRESS_TESTS=("spawn_stress_test.wasm" "stress_test_threads_creation.wasm")
 
 run_aot_tests () {
     local tests=("$@")
-    local iwasm="${IWASM_CMD}"
     for test_wasm in ${tests[@]}; do
-        local extra_stress_flags=""
-        for stress_test in "${STRESS_TESTS[@]}"; do
-            if [ "$test_wasm" == "$stress_test" ]; then
-                iwasm="${IWASM_CMD_STRESS}"
-            fi  
-        done
+        local iwasm="${IWASM_CMD}"
+        if [[ $test_wasm =~ "stress" ]]; then
+            iwasm="${IWASM_CMD_STRESS}"
+        fi
 
         test_aot="${test_wasm%.wasm}.aot"
         test_json="${test_wasm%.wasm}.json"
@@ -52,7 +49,7 @@ run_aot_tests () {
             expected=$(jq .exit_code ${test_json})
         fi
 
-        ${IWASM_CMD} $extra_stress_flags $test_aot
+        ${iwasm} $test_aot
         ret=${PIPESTATUS[0]}
 
         echo "expected=$expected, actual=$ret"
@@ -66,19 +63,6 @@ if [[ $MODE != "aot" ]];then
     python3 -m venv wasi-env && source wasi-env/bin/activate
     python3 -m pip install -r test-runner/requirements.txt
 
-    # Stress tests require max-threads=8 so they're executed separately
-    for stress_test in "${STRESS_TESTS[@]}"; do
-        if [[ -e "${THREAD_INTERNAL_TESTS}${stress_test}" ]]; then
-            echo "${stress_test}" is a stress test
-            ${IWASM_CMD_STRESS} ${THREAD_INTERNAL_TESTS}${stress_test}
-            ret=${PIPESTATUS[0]}
-            if [ "${ret}" -ne 0 ]; then
-                echo "Stress test ${stress_test} FAILED with code " ${ret}
-                exit_code=${ret}
-            fi
-        fi
-    done
-
     TEST_RUNTIME_EXE="${IWASM_CMD}" python3 test-runner/wasi_test_runner.py \
             -r adapters/wasm-micro-runtime.py \
             -t \
@@ -87,12 +71,20 @@ if [[ $MODE != "aot" ]];then
                 ${THREAD_PROPOSAL_TESTS} \
                 ${THREAD_INTERNAL_TESTS} \
                 ${LIB_SOCKET_TESTS} \
-            --exclude-filter "${THREAD_INTERNAL_TESTS}skip.json"
 
     ret=${PIPESTATUS[0]}
-    if [ "${ret}" -ne 0 ]; then
-        exit_code=${ret}
+
+    TEST_RUNTIME_EXE="${IWASM_CMD_STRESS}" python3 test-runner/wasi_test_runner.py \
+            -r adapters/wasm-micro-runtime.py \
+            -t \
+                ${THREAD_STRESS_TESTS}
+
+    if [ "${ret}" -eq 0 ]; then
+        ret=${PIPESTATUS[0]}
     fi
+    
+    exit_code=${ret}
+    
     deactivate
 else
     target_option=""
@@ -101,7 +93,7 @@ else
     fi
 
     exit_code=0
-    for testsuite in ${THREAD_PROPOSAL_TESTS} ${THREAD_INTERNAL_TESTS}; do
+    for testsuite in ${THREAD_STRESS_TESTS} ${THREAD_PROPOSAL_TESTS} ${THREAD_INTERNAL_TESTS}; do
         tests=$(ls ${testsuite}*.wasm)
         tests_array=($tests)
         run_aot_tests "${tests_array[@]}"
--- a/core/iwasm/libraries/lib-wasi-threads/stress-test/build.sh
+++ b/core/iwasm/libraries/lib-wasi-threads/stress-test/build.sh
@@ -33,20 +33,11 @@ while [[ $# -gt 0 ]]; do
     esac
 done
 
-# Stress tests names
-thread_start_file_exclusions=("linear_memory_size_update.wasm")
-
 rm -rf *.wasm
 rm -rf *.aot
 
 for test_c in *.c; do
     test_wasm="$(basename $test_c .c).wasm"
-
-    if [[ " ${thread_start_file_exclusions[@]} " =~ " ${test_wasm} " ]] ; then
-        thread_start_file=""
-    else
-        thread_start_file=$WAMR_DIR/samples/wasi-threads/wasm-apps/wasi_thread_start.S
-    fi
 
     if [[ -n "$sysroot_path" ]]; then 
         if [ ! -d "$sysroot_path" ]; then 
@@ -60,7 +51,8 @@ for test_c in *.c; do
     $CC \
         -target wasm32-wasi-threads \
         -O2 \
-        -pthread -ftls-model=local-exec \
+        -Wall \
+        -pthread \
         -z stack-size=32768 \
         -Wl,--export=__heap_base \
         -Wl,--export=__data_end \
@@ -68,8 +60,6 @@ for test_c in *.c; do
         -Wl,--export=wasi_thread_start \
         -Wl,--export=malloc \
         -Wl,--export=free \
-        -I $WAMR_DIR/samples/wasi-threads/wasm-apps \
         $sysroot_command \
-        $thread_start_file \
         $test_c -o $test_wasm
 done
--- a/core/iwasm/libraries/lib-wasi-threads/stress-test/errorcheck_mutex_stress_test.c
+++ b/core/iwasm/libraries/lib-wasi-threads/stress-test/errorcheck_mutex_stress_test.c
@@ -1,0 +1,27 @@
+/*
+ * Copyright (C) 2023 Amazon.com Inc. or its affiliates. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+
+#include <pthread.h>
+#include <errno.h>
+#include "mutex_common.h"
+
+int
+main()
+{
+    pthread_mutex_t mutex;
+
+    // Set mutex type to errorcheck. This type provides some additional checks
+    // (for example returns EDEADLK instead of deadlocking in some cases)
+    pthread_mutexattr_t mutex_attr;
+    pthread_mutexattr_init(&mutex_attr);
+    pthread_mutexattr_settype(&mutex_attr, PTHREAD_MUTEX_ERRORCHECK);
+
+    pthread_mutex_init(&mutex, &mutex_attr);
+    pthread_mutexattr_destroy(&mutex_attr);
+
+    run_common_tests(&mutex);
+    fprintf(stderr, "Errorcheck mutex test is completed\n");
+    pthread_mutex_destroy(&mutex);
+}
--- a/core/iwasm/libraries/lib-wasi-threads/stress-test/manifest.json
+++ b/core/iwasm/libraries/lib-wasi-threads/stress-test/manifest.json
@@ -1,0 +1,3 @@
+{
+    "name": "lib-wasi-threads stress tests"
+}
--- a/core/iwasm/libraries/lib-wasi-threads/stress-test/mutex_common.h
+++ b/core/iwasm/libraries/lib-wasi-threads/stress-test/mutex_common.h
@@ -1,0 +1,229 @@
+/*
+ * Copyright (C) 2023 Amazon.com Inc. or its affiliates. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+
+#ifndef MUTEX_COMMON_H
+#define MUTEX_COMMON_H
+
+#include <pthread.h>
+#include <stdio.h>
+#include <assert.h>
+#include <errno.h>
+#include <unistd.h>
+#include <stdbool.h>
+#include <time.h>
+#include <stdlib.h>
+#include <string.h>
+
+enum Constants {
+    NUM_ITER = 250000,
+    NUM_THREADS = 12,
+    NUM_RETRY = 8,
+    RETRY_SLEEP_TIME_US = 1000,
+};
+
+// We're counting how many times each thread was called using this array
+// Main thread is also counted here so we need to make arrays bigger
+typedef struct {
+    int tids[NUM_THREADS + 1];
+    int calls[NUM_THREADS + 1];
+} StatCollector;
+
+typedef struct {
+    pthread_mutex_t *mutex;
+    StatCollector stat;
+    int counter;
+    bool is_sleeping;
+} MutexCounter;
+
+// This enum defines whether thread should sleep to increase contention
+enum SleepState {
+    NON_SLEEP = 0,
+    SLEEP = 1,
+};
+
+void
+mutex_counter_init(MutexCounter *mutex_counter, pthread_mutex_t *mutex,
+                   enum SleepState is_sleeping)
+{
+    memset(mutex_counter, 0, sizeof(*mutex_counter));
+    mutex_counter->mutex = mutex;
+    mutex_counter->is_sleeping = is_sleeping;
+}
+
+// This function spawns the thread using exponential retries if it receives
+// EAGAIN
+static inline void
+spawn_thread(pthread_t *tid, void *func, void *arg)
+{
+    int status_code = -1;
+    int timeout_us = RETRY_SLEEP_TIME_US;
+    for (int tries = 0; status_code != 0 && tries < NUM_RETRY; ++tries) {
+        status_code = pthread_create(tid, NULL, (void *(*)(void *))func, arg);
+        assert(status_code == 0 || status_code == EAGAIN);
+        if (status_code == EAGAIN) {
+            usleep(timeout_us);
+            timeout_us *= 2;
+        }
+    }
+
+    assert(status_code == 0 && "Thread creation should succeed");
+}
+
+// This function adds tid to our stat
+static inline void
+add_to_stat(StatCollector *stat, int tid)
+{
+    int tid_num = 0;
+    for (; tid_num < NUM_THREADS + 1 && stat->tids[tid_num] != 0; ++tid_num) {
+        if (stat->tids[tid_num] == tid) {
+            stat->calls[tid_num]++;
+            return;
+        }
+    }
+
+    assert(tid_num < NUM_THREADS + 1);
+    stat->tids[tid_num] = tid;
+    stat->calls[tid_num] = 1;
+}
+
+// This function prints number of calls by TID
+static inline void
+print_stat(StatCollector *stat)
+{
+    fprintf(stderr, "Thread calls count by TID\n");
+    for (int i = 0; i < NUM_THREADS + 1; ++i) {
+        if (stat->tids[i] != 0) {
+            fprintf(stderr, "TID: %d; Calls: %d\n", stat->tids[i],
+                    stat->calls[i]);
+        }
+    }
+}
+
+// This function is run by the threads, it increases counter in a loop and then
+// sleeps after unlocking the mutex to provide better contention
+static inline void *
+inc_shared_variable(void *arg)
+{
+    MutexCounter *mutex_counter = (MutexCounter *)(arg);
+    int sleep_us = 0;
+    while (!pthread_mutex_lock(mutex_counter->mutex)
+           && mutex_counter->counter < NUM_ITER) {
+        mutex_counter->counter++;
+        add_to_stat(&mutex_counter->stat, (int)(pthread_self()));
+        if (mutex_counter->is_sleeping) {
+            sleep_us = rand() % 1000;
+        }
+
+        assert(pthread_mutex_unlock(mutex_counter->mutex) == 0
+               && "Should be able to unlock a mutex");
+        if (mutex_counter->is_sleeping) {
+            usleep(sleep_us);
+        }
+    }
+
+    assert(mutex_counter->counter == NUM_ITER);
+    assert(pthread_mutex_unlock(mutex_counter->mutex) == 0
+           && "Should be able to unlock the mutex after test execution");
+
+    return NULL;
+}
+
+// Locking and unlocking a mutex in a single thread.
+static inline void *
+same_thread_lock_unlock_test(void *mutex)
+{
+    for (int i = 0; i < NUM_ITER; ++i) {
+        assert(pthread_mutex_lock(mutex) == 0
+               && "Main thread should be able to lock a mutex");
+        assert(pthread_mutex_unlock(mutex) == 0
+               && "Main thread should be able to unlock a mutex");
+    }
+
+    return NULL;
+}
+
+// This function spawns a thread that locks and unlocks a mutex `NUM_ITER` times
+// in a row
+static inline void
+same_non_main_thread_lock_unlock_test(pthread_mutex_t *mutex)
+{
+    pthread_t tid = 0;
+    spawn_thread(&tid, same_thread_lock_unlock_test, mutex);
+
+    assert(tid != 0 && "TID can't be 0 after successful thread creation");
+    assert(pthread_join(tid, NULL) == 0
+           && "Thread should be joined successfully");
+}
+
+// This function checks basic contention between main and non-main thread
+// increasing the shared variable
+static inline void
+two_threads_inc_test(pthread_mutex_t *mutex)
+{
+    MutexCounter mutex_counter;
+    mutex_counter_init(&mutex_counter, mutex, false);
+
+    pthread_t tid = 0;
+    spawn_thread(&tid, inc_shared_variable, &mutex_counter);
+
+    assert(tid != 0 && "TID can't be 0 after successful thread creation");
+    inc_shared_variable(&mutex_counter);
+    assert(pthread_join(tid, NULL) == 0
+           && "Thread should be joined without errors");
+    assert(mutex_counter.counter == NUM_ITER);
+}
+
+// This function creates number of threads specified by NUM_THREADS and run
+// concurrent increasing of shared variable
+static inline void
+max_threads_inc_test(pthread_mutex_t *mutex, int threads_num,
+                     enum SleepState is_sleeping)
+{
+    MutexCounter mutex_counter;
+    mutex_counter_init(&mutex_counter, mutex, is_sleeping);
+
+    pthread_t tids[threads_num];
+    for (int i = 0; i < threads_num; ++i) {
+        spawn_thread(&tids[i], inc_shared_variable, &mutex_counter);
+    }
+
+    inc_shared_variable(&mutex_counter);
+
+    for (int i = 0; i < threads_num; ++i) {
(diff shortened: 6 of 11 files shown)


core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
```diff
@@ -0,0 +1,261 @@
...
+};
+
+// This enum defines whether thread should sleep to increase contention
+enum sleep_state {
+    NON_SLEEP = 0,
+    SLEEP = 1,
+};
+
+void
+spawn_thread(pthread_t *tid, void *func, void *arg)
```
loganek, line 37: is that something we could share with other wasi thread tests, not just mutex tests?

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
```diff
@@ -0,0 +1,261 @@
...
+#include <errno.h>
+#include <unistd.h>
+#include <stdbool.h>
+
+typedef struct {
+    pthread_mutex_t *mutex;
+    int *counter;
+    int limit;
+    bool is_sleeping;
+} MutexCounter;
```
loganek, line 21: I think there's a bit of inconsistency with naming convention (`sneak_case` for enums and `PascalCase` for structs). I'd suggest to keep them in sync.

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
```diff
@@ -0,0 +1,261 @@
...
+    for (int i = 0; i < threads_num; ++i) {
+        assert(pthread_join(tids[i], NULL) == 0
+               && "Thread should be joined without errors");
+        ;
+    }
+}
+
+// This function just runs all the tests described above
+void
+run_common_tests(pthread_mutex_t *mutex)
```
loganek, line 228: we can keep all the tests in one function; splitting them to separate small wasm applications would improve the reporting though.

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
```diff
@@ -0,0 +1,261 @@
...
+    RETRY_SLEEP_TIME_US = 1000,
+};
+
+// This enum defines whether thread should sleep to increase contention
+enum sleep_state {
+    NON_SLEEP = 0,
+    SLEEP = 1,
+};
+
+void
```
loganek, line 36: *This is a comment for all the functions / global variables defined in the file*

I'd strongly advise to move the implementation to .c file or make them static. Otherwise, if the file is included in two compilation units and they're linked together, there'll be a linker error (due to multiple definitions). I know we don't have this scenario now, but with little effort we can make this better for the potential future.

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
```diff
@@ -0,0 +1,261 @@
...
+        status_code = pthread_create(tid, NULL, (void *(*)(void *))func, arg);
+        assert(status_code == 0 || status_code == EAGAIN);
+        if (status_code == EAGAIN) {
+            usleep(timeout_us);
+            timeout_us *= 2;
+        }
+    }
+
+    assert(status_code == 0 && "Thread creation should succeed");
+}
```
loganek, line 51: this function could probably return pthread_t, which I guess would slightly improve the usage of this function from what I can see later in the code.

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
```diff
@@ -0,0 +1,261 @@
...
+
+// We're counting how many times each thread was called using this array
+// Main thread is also counted here so we need to make arrays bigger
+typedef struct {
+    int tids[NUM_THREADS + 1];
+    int calls[NUM_THREADS + 1];
+} StatCollector;
+
+StatCollector call_stat;
```
loganek, line 60: would it be possible to not have this a global, and instead instantiate it in the test and pass a pointer to it to each function that needs that? This statistics are used by many tests, so there's a chance we'll one day forget to call the `clear_stat` function.

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
```diff
@@ -0,0 +1,261 @@
...
+
+void
+clear_stat()
+{
+    for (int i = 0; i < NUM_THREADS + 1; ++i) {
```
loganek, line 65: this can be a bit simpler (and more generic, if we add more fields):  `memset(&call_stat, 0, sizeof(call_stat))`

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
```diff
@@ -0,0 +1,261 @@
...
+#include <pthread.h>
+#include <stdio.h>
+#include <assert.h>
+#include <errno.h>
+#include <unistd.h>
+#include <stdbool.h>
+
+typedef struct {
+    pthread_mutex_t *mutex;
+    int *counter;
```
loganek, line 18: does it have to be a pointer? Based on the usage in the code, this can be just an int value.

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    assert(tid_num < NUM_THREADS + 1);
+    call_stat.tids[tid_num] = tid;
+    call_stat.calls[tid_num] = 1;
+}
+
+// This function ensures that every of NUM_THREADS+1 threads number of
+// operations is in between [NUM_ITER/NUM_THREADS, NUM_ITER/(NUM_THREADS+2)].
+// For 100'000 operations it means that every thread should be called [7142,
```
loganek, line 90: why are we checking that?

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
```diff
@@ -0,0 +1,261 @@
...
+inc_until_limit(void *arg)
+{
+    MutexCounter mutex_counter = *(MutexCounter *)(arg);
+    int sleep_us = 0;
+    while (!pthread_mutex_lock(mutex_counter.mutex)
+           && *mutex_counter.counter < mutex_counter.limit) {
+        (*mutex_counter.counter)++;
+        add_to_stat((int)(pthread_self()));
+        if (mutex_counter.is_sleeping) {
+            sleep_us = hash(*mutex_counter.counter) % 1000;
```
loganek, line 145: why do we hash the counter? Is it to generate a random value? If so, why not to use rand() instead?

loganek: Also, shouldn't this assignment be moved to L151?

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
```diff
@@ -0,0 +1,261 @@
...
+
+    assert(*mutex_counter.counter == mutex_counter.limit);
+    assert(pthread_mutex_unlock(mutex_counter.mutex) == 0
+           && "Should be able to unlock the mutex after test execution");
+
+    return NULL;
+}
+
+void
+non_main_test(pthread_mutex_t *mutex, int num_iter)
```
loganek, line 163: is this test conceptually the same as `same_thread_test` except that the code runs on a spawned thread?

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
```diff
@@ -0,0 +1,261 @@
...
+        call_stat.tids[i] = 0;
+        call_stat.calls[i] = 0;
+    }
+}
+
+void
+add_to_stat(int tid)
+{
+    int tid_num = 0;
+    for (; tid_num < NUM_THREADS + 1 && call_stat.tids[tid_num] != 0;
```
loganek, line 75: won't that algorithm slow down tests that we expect to be executed as fast as possible (to e.g. test for data races)?

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
```diff
@@ -0,0 +1,261 @@
...
+    mutex_counter.mutex = mutex;
+    mutex_counter.counter = &counter;
+    mutex_counter.limit = num_iter;
+    mutex_counter.is_sleeping = false;
+
+    pthread_t tid = 0;
+    spawn_thread(&tid, inc_until_limit, &mutex_counter);
+
+    assert(tid != 0 && "TID can't be 0 after successful thread creation");
+    inc_until_limit(&mutex_counter);
```
loganek, line 194: since `is_sleeping` is disabled, how likely is that the counter will reach the limit here, before another thread is spawned? Is num_iter large enough? Otherwise, we're actually not testing the concurrent updates.

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
```diff
@@ -0,0 +1,261 @@
...
+    mutex_counter.counter = &counter;
+    mutex_counter.limit = num_iter;
+    mutex_counter.is_sleeping = is_sleeping;
+
+    pthread_t tids[threads_num];
+    for (int i = 0; i < threads_num; ++i) {
+        spawn_thread(&tids[i], inc_until_limit, &mutex_counter);
+    }
+
+    inc_until_limit(&mutex_counter);
```
loganek, line 217: same comment as above.

loganek (changes requested):

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
Zzzabiyaka: I think it's fine if you have 12 threads 

it's about 12 operations :)

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
Zzzabiyaka: it's a bit controversial to me 

as I'm going to have other mutex types running the same test it might eventually become 

NUM_OF_MUTEX_TYPES x NUM_OF_TESTS matrix what will be hard to update etc

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
Zzzabiyaka: why not to? 

I mean if you have more or less random sleeps and still spending 50% of operations in one TID probably smth is bad with your system

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
Zzzabiyaka: yeah
it can probably be just a flag (main / non-main)

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
Zzzabiyaka: I don't really know what's considered the best practice for the project but I'm changing to `PascalCase`

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
Zzzabiyaka: I'm trying to stay consistent with pthread (e.g. I want to allow passing the same argument as you'd pass for pthread_create, tid&, ...) 

If you have really strong preference I can change it but from my point of view it's fine

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
Zzzabiyaka: done

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
Zzzabiyaka: well, I'm afraid to be wrong, but this variable is passed to threads as a part of *arg and updated there.

If it'd be a value, every thread would receive a copy and we wouldn't have effective multithreading, because counter here is effectively a shared resource, and as it's not global var I'm passing a pointer. 

Please correct me if I'm wrong

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
Zzzabiyaka: done

Zzzabiyaka: >Also, shouldn't this assignment be moved to L151?

both rand() and hash() that I used before are **not** thread-safe, that's why it's inside the mutex

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
Zzzabiyaka: Done. Thank you

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
Zzzabiyaka: Refactored so it's now local.

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
Zzzabiyaka: well here's the example for fast-interp

I think it's fine 

```
Starting concurrent_inc_non_sleep test
Thread calls count by TID
TID: 71316; Calls: 10044
TID: 104228; Calls: 8523
TID: 137140; Calls: 8725
TID: 170052; Calls: 6955
TID: 202964; Calls: 7502
TID: 235876; Calls: 6953
TID: 268788; Calls: 7106
TID: 301700; Calls: 7237
TID: 334612; Calls: 7666
TID: 367524; Calls: 7547
TID: 400436; Calls: 7195
TID: 433348; Calls: 7170
TID: 1028; Calls: 7377
```

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
Zzzabiyaka: we can but I'm not sure we need a separate `common.h` for one function so I don't know whether it worth it

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
loganek: don't you pass a pointer to the whole `MutexCounter` struct?

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
loganek: I don't see why not, but up to you.

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
loganek: no strong preference  here either, feel free to keep it as it is.

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
loganek: That **might** be true if you run this test a number of times - than **probably** in average that's what you will achieve. However, keeping it like that is asking for flakiness. If you  want to test this behavior, and do it based on statistics, I'd rather move it to separate test suite which is expected to sometimes fail (but then we'd need to figure out how to track this test suite).

We can keep it like that, but as I said - expect flakiness here

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
loganek: so both functions check if mutex can be locked / unlocked many times, it's just that one does it in a main thread, and the other does it in a non-main thread?

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
```diff
@@ -0,0 +1,261 @@
...
+    pthread_t tid = 0;
+    spawn_thread(&tid, inc_until_limit, &mutex_counter);
+
+    assert(tid != 0 && "TID can't be 0 after successful thread creation");
+    pthread_join(tid, NULL);
+    assert(*mutex_counter.counter == mutex_counter.limit);
+}
+
+void
+main_non_main_test(pthread_mutex_t *mutex, int num_iter)
```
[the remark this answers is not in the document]
loganek: I think that's the same for all the other tests in this file.

tests/wamr-test-suites/wasi-test-script/run_wasi_tests.sh
```diff
@@ -15,22 +15,22 @@ readonly IWASM_CMD="${WORK_DIR}/../../../../product-mini/platforms/${PLATFORM}/b
     --allow-resolve=google-public-dns-a.google.com \
     --addr-pool=::1/128,127.0.0.1/32"
 
-readonly IWASM_CMD_STRESS="${IWASM_CMD} --max-threads=8"
```
loganek, line 18: I wonder if there's a way to define this number once, instead of doing it here AND in some of the test files?

Zzzabiyaka: I don't know any except just setting an env var but I can check 

I don't think test file can access any iwasm settings but I might be wrong

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
Zzzabiyaka: You're extremely right, sorry 😂

Zzzabiyaka: well  do but later I put it inside a var 

`    MutexCounter mutex_counter = *(MutexCounter *)(arg); `

It's confusing so I will refactor it probably

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
Zzzabiyaka: I didn't see any flakiness here running the tests locally and in CI  but I trust your experience so I'll remove this check : )

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
Zzzabiyaka: yeah but i think it's a valid case to check. 

tell me if it's not needed

Zzzabiyaka: well I decided to refactor it a bit anyway

Zzzabiyaka: I refactored it to call the same func as in `same_thread_test` but from non-main thread

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
Zzzabiyaka: well, you will if you execute it on iwasm or other way (not via test-runner) because of stderr logs 

you only won't know if you execute via test-runner and I find it annoying that it does not shown stderr output by the way

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
Zzzabiyaka: it's happened once in CI for AOT so I increased to 250k and it didn't slow the test much as it's quite fast itself

.github/workflows/nightly_run.yml
```diff
@@ -599,6 +599,11 @@ jobs:
         run: bash build.sh --sysroot "$SYSROOT_PATH"
         working-directory: ./core/iwasm/libraries/lib-wasi-threads/test/
 
+      - name: Build WASI thread stress tests
```
loganek, line 602: why do we do it on both nightly and on every change? Isn't the test on every change sufficient?

core/iwasm/libraries/lib-wasi-threads/stress_test/build.sh
```diff
@@ -0,0 +1,65 @@
...
+            exit
+            ;;
+        *)
+            echo "Unknown option: $1"
+            exit 1
+            ;;
+    esac
+done
+
+rm -rf *.wasm
```
loganek, line 36: `rm -rf` in an arbitrary directory is quite risky; can you make sure the working directory is set to the right value?

core/iwasm/libraries/lib-wasi-threads/stress_test/build.sh
```diff
@@ -0,0 +1,65 @@
...
+        -O2 \
+        -pthread \
+        -z stack-size=32768 \
+        -Wl,--export=__heap_base \
+        -Wl,--export=__data_end \
+        -Wl,--shared-memory,--max-memory=1966080 \
+        -Wl,--export=wasi_thread_start \
+        -Wl,--export=malloc \
+        -Wl,--export=free \
+        -I $WAMR_DIR/samples/wasi-threads/wasm-apps \
```
loganek, line 62: what do we include from this path?

.github/workflows/nightly_run.yml
Zzzabiyaka: we can run them only in nightly, I don't think it changes much, but we can also test per-pr as it's 10-15 min 

I think both are ok, maybe let's disable them to make per-pr CI faster

core/iwasm/libraries/lib-wasi-threads/stress_test/errorcheck_mutex_stress_test.c
```diff
@@ -0,0 +1,25 @@
...
+
+int
+main()
+{
+    pthread_mutex_t mutex;
+
+    // Set mutex type to errorcheck. This type provides some additional checks
+    // (for example returns EDEADLK instead of deadlocking in some cases)
+    pthread_mutexattr_t mutex_attr;
+    pthread_mutexattr_init(&mutex_attr);
```
loganek, line 18: should you call `pthread_mutexattr_destroy` later on?

core/iwasm/libraries/lib-wasi-threads/stress_test/build.sh
Zzzabiyaka: legacy from copied command I think, removed

core/iwasm/libraries/lib-wasi-threads/stress_test/errorcheck_mutex_stress_test.c
```diff
@@ -0,0 +1,25 @@
...
+#include <errno.h>
+#include "mutex_common.h"
+
+int
+main()
+{
+    pthread_mutex_t mutex;
+
+    // Set mutex type to errorcheck. This type provides some additional checks
+    // (for example returns EDEADLK instead of deadlocking in some cases)
```
loganek, line 16: should we have a test that triggers a deadlock and assert the return flag?

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    assert(mutex_counter->counter == NUM_ITER);
+    assert(pthread_mutex_unlock(mutex_counter->mutex) == 0
+           && "Should be able to unlock the mutex after test execution");
+
+    return NULL;
+}
+
+// Locking and unlocking the same thread
```
loganek, line 124:
```suggestion
// Locking and unlocking a mutex in a single thread.
```

core/iwasm/libraries/lib-wasi-threads/stress_test/errorcheck_mutex_stress_test.c
Zzzabiyaka: I did that test and it worked but then I realised it's not a `stress_test` and more like unit/functional test

I can add but I think this basic functional should be already tested or tested somewhere else

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
```diff
@@ -0,0 +1,224 @@
...
+static inline void *
+same_thread_lock_unlock_test(void *mutex)
+{
+    for (int i = 0; i < NUM_ITER; ++i) {
+        assert(pthread_mutex_lock(mutex) == 0
```
loganek, line 129: don't you need a cast to a `phtread_mutex_t*` to avoid warnings?

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
```diff
@@ -0,0 +1,224 @@
...
+               && "Main thread should be able to lock a mutex");
+        assert(pthread_mutex_unlock(mutex) == 0
+               && "Main thread should be able to unlock a mutex");
+    }
+
+    return NULL;
+}
+
+// This function spawns a thread that locks and unlocks many times in a row
```
loganek, line 138:
```suggestion
// This function spawns a thread that locks and unlocks a mutex `NUM_ITER` times in a row
```

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
Zzzabiyaka: It doesn't give warning even with -Wextra

Zzzabiyaka: but I've actually found one useful warning with -Wall (stat was unused)

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
```diff
@@ -0,0 +1,224 @@
...
+    assert(pthread_join(tid, NULL) == 0
+           && "Thread should be joined successfully");
+}
+
+// This function checks basic contention between main and non-main thread
+// increasing the shared variable
+static inline void
+two_threads_inc_test(pthread_mutex_t *mutex)
+{
+    MutexCounter mutex_counter;
```
loganek, line 155: `MutexCounter::stat` is not being initialized here, but it is used later on in the `inc_shared_variable`. You might consider having a `mutex_counter_init(MutexCounter* counter, <params>)` function to better protect from uninitialized fields (also, if we add one more field to the struct, there'll only be one place to update at least for the default value).

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
```diff
@@ -0,0 +1,224 @@
...
+    assert(mutex_counter.counter == NUM_ITER);
+}
+
+// This function creates number of threads specified by NUM_THREADS and run
+// concurrent increasing of shared variable
+static inline void
+max_threads_inc_test(pthread_mutex_t *mutex, int threads_num, bool is_sleeping)
+{
+    MutexCounter mutex_counter;
+    StatCollector stat;
```
loganek, line 176: unused? Consider enabling extra warnings for compilation to detect this sort of issues.

Zzzabiyaka: yeah, detected with -Wall and removed

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
loganek: ok, let's keep it like that. Btw stderr and stdout is shown when the test fails in the test runner IIRC

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
```diff
@@ -0,0 +1,224 @@
...
+    print_stat(&mutex_counter.stat);
+}
+
+// This function just runs all the tests described above
+static inline void
+run_common_tests(pthread_mutex_t *mutex)
+{
+    srand(time(NULL));
+
+    fprintf(stderr, "Starting main_thread test\n");
```
loganek, line 203: This text is neither descriptive (i.e. I don't know what the test is doing) nor it matches the function name. Even though you can look at the code and find the string and then figure out what's wrong, I think adding a good info text here wouldn't hurt and could make our life slightly better.

Zzzabiyaka: I think just making the text match func name is fine

core/iwasm/libraries/lib-wasi-threads/stress_test/errorcheck_mutex_stress_test.c
loganek: ok, should we also test different types of mutexes? Especially I'd be interested in a stress test for recursive mutex where we lock the mutex without unlocking multiple times in the same thread.

.github/workflows/nightly_run.yml
loganek: I think we can keep them per-pr as you have now (at least for now, until they start taking longer), but if it's per-pr, I don't see much value having it in nightly as well.

core/iwasm/libraries/lib-wasi-threads/stress_test/errorcheck_mutex_stress_test.c
Zzzabiyaka: >ok, should we also test different types of mutexes? Especially I'd be interested in a stress test for recursive mutex where we lock the mutex without unlocking multiple times in the same thread.

I think we should (if we don't have such tests already). but I've done that test locally and it was ok (i just locked 5 times and unlocked 5 times)

Zzzabiyaka: but it's really about wasi-libc testing I think, not even wamr

loganek: stress tests are both about libc and the runtime. This particular one would allow us to test a particular configuration of assembly instructions related to atomic operations. I presume the assembly code of mutex and recursive mutex operation will be similar, but not sure if exactly the same (just like for the `PTHREAD_MUTEX_ERRORCHECK`).

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
```diff
@@ -0,0 +1,229 @@
...
+} MutexCounter;
+
+void
+mutex_counter_init(MutexCounter *mutex_counter, pthread_mutex_t *mutex,
+                   bool is_sleeping)
+{
+    mutex_counter->mutex = mutex;
+    memset(&mutex_counter->stat, 0, sizeof(mutex_counter->stat));
+    mutex_counter->counter = 0;
+    mutex_counter->is_sleeping = is_sleeping;
```
loganek, line 47:
```suggestion
    memset(mutex_counter, 0, sizeof(*mutex_counter));
    mutex_counter->mutex = mutex;
    mutex_counter->is_sleeping = is_sleeping;
```

not critical though, just a suggestion.

core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h
```diff
@@ -0,0 +1,229 @@
...
+}
+
+// This enum defines whether thread should sleep to increase contention
+enum SleepState {
```
loganek, line 51: unused?

Zzzabiyaka:
```

    fprintf(stderr, "Starting max_threads_inc_test_sleep test\n");
    max_threads_inc_test(mutex, NUM_THREADS, SLEEP);
    fprintf(stderr, "Finished concurrent_inc sleep test\n");

    fprintf(stderr, "Starting max_threads_inc_test_non_sleep test\n");
    max_threads_inc_test(mutex, NUM_THREADS, NON_SLEEP);
    fprintf(stderr, "Finished max_threads_inc_test test\n");
    ```

File: core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h (hunk 26 above)
Comment (Zzzabiyaka) [reply core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h]: changed `bool` in function declaration to SleepState

File: core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h (hunk 29 above)
Comment (Zzzabiyaka) [reply core/iwasm/libraries/lib-wasi-threads/stress_test/mutex_common.h]: changed `bool` with `enum SleepState` in func defitions. Thanks for the comment

**Review by loganek — APPROVED** [review]
LGTM, thanks

File: .github/workflows/nightly_run.yml [hunk 30]
```diff
@@ -599,6 +599,11 @@ jobs:
...
+        if: matrix.test_option == '$WASI_TEST_OPTIONS'
+        run: bash build.sh --sysroot "$SYSROOT_PATH"
+        working-directory: ./core/iwasm/libraries/lib-wasi-threads/stress_test/
```
wenyongh, line 605: Could you change folder name to `stress-test`? We normally use `-` for folder name and `_` for file name.

Zzzabiyaka: Done, thank you

lum1n0us:
LGTM

tests/wamr-test-suites/wasi-test-script/run_wasi_tests.sh
```diff
@@ -15,25 +15,22 @@ readonly IWASM_CMD="${WORK_DIR}/../../../../product-mini/platforms/${PLATFORM}/b
...
+readonly IWASM_CMD_STRESS="${IWASM_CMD} --max-threads=12"
 readonly WAMRC_CMD="${WORK_DIR}/../../../../wamr-compiler/build/wamrc"
 readonly C_TESTS="tests/c/testsuite/"
 readonly ASSEMBLYSCRIPT_TESTS="tests/assemblyscript/testsuite/"
 readonly THREAD_PROPOSAL_TESTS="tests/proposals/wasi-threads/"
 readonly THREAD_INTERNAL_TESTS="${WAMR_DIR}/core/iwasm/libraries/lib-wasi-threads/test/"
+readonly THREAD_STRESS_TESTS="${WAMR_DIR}/core/iwasm/libraries/lib-wasi-threads/stress_test/"
```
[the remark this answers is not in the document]
Zzzabiyaka: Thank you, fixed

PR #2473, opened 2023-08-16
Fix spec tests on windows, enable spec tests for windows in CI

loganek:


.github/workflows/compilation_on_windows.yml
```diff
@@ -39,6 +39,12 @@ on:
   # allow to be triggered manually
   workflow_dispatch:
 
+env:
+  # For Spec Test
+  DEFAULT_TEST_OPTIONS: "-s spec -b"
```
wenyongh, line 44: Can we use `-P` option to run the spec test parallelly?

wenyongh (approved):
LGTM

.github/workflows/compilation_on_windows.yml
loganek: There's something in Python script that won't work when enabling parallel execution. I'm yet to investigate it but I thought we can enable those tests in a single-threaded mode.

wenyongh: Got it, thanks.

PR #2484, opened 2023-08-21
Run rust tests from wasi-testsuite

zoraaver:


zoraaver:
Currently can't merge this due to a failing test on android, ubuntu 22.04:

[fd_filestat_get](https://github.com/WebAssembly/wasi-testsuite/blob/main/tests/rust/src/bin/fd_filestat_get.rs)

I can reproduce it locally on OSX but I'm not completely sure I understand if the test is correct. 

Also the following rust tests are failing on nightly (can't reproduce locally):

[interesting_paths](https://github.com/WebAssembly/wasi-testsuite/blob/main/tests/rust/src/bin/interesting_paths.rs)
[close_preopen](https://github.com/WebAssembly/wasi-testsuite/blob/main/tests/rust/src/bin/close_preopen.rs).

Will need to investigate more to figure out why these are failing.

loganek:
I think the test failure is perfectly valid. From the [doc](https://github.com/WebAssembly/WASI/blob/main/legacy/application-abi.md):

> When _start or _initialize is called, environments shall provide file descriptors with indices 0, 1, and 2 representing stream resources for standard input, standard output, and standard error. Environments may provide additional "preopen" file descriptors that can be inspected with fd_prestat_get and fd_prestat_dir_name. These resources may be closed at any time.

loganek:
Please see the related conversation in https://github.com/WebAssembly/wasi-testsuite/pull/59. Depending on the result of that, we'll either delete the test or update the specification (and WAMR implementation accordingly).

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -2021,20 +2057,6 @@ wasmtime_ssp_path_open(
         return error;
     }
 
-    {
```
zoraaver, line 2024: I'm not sure if this block is needed for anything else - it seems like it's giving unnecessary rights to the fd. It was causing this test [path_open_read_without_rights](https://github.com/WebAssembly/wasi-testsuite/blob/main/tests/rust/src/bin/path_open_read_without_rights.rs#L31-L34) to fail.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -1136,7 +1145,45 @@ wasmtime_ssp_fd_renumber(
     fd_object_release(fo);
     --ft->used;
 
+    // Handle renumbering of any preopened resources
```
zoraaver, line 1148: Added logic for renumbering preopened resources so this [test](https://github.com/WebAssembly/wasi-testsuite/blob/main/tests/rust/src/bin/overwrite_preopen.rs) passes. I don't think there's any other reason not to support renumbering preopened resources.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -844,34 +864,35 @@ wasmtime_ssp_fd_close(
...
     if (error != 0) {
+        rwlock_unlock(&prestats->lock);
         rwlock_unlock(&ft->lock);
         return error;
     }
 
     // Remove it from the file descriptor table.
     struct fd_object *fo;
     fd_table_detach(ft, fd, &fo);
+
```
zoraaver, line 883: Added logic to close a preopened resource so [close_preopen](https://github.com/WebAssembly/wasi-testsuite/blob/main/tests/rust/src/bin/close_preopen.rs) passes.

zoraaver:
I've raised a [PR](https://github.com/WebAssembly/wasi-testsuite/pull/85) in wasi-testsuite to fix the tests which had various permissions issues etc.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -1136,7 +1145,45 @@ wasmtime_ssp_fd_renumber(
...
+    }
+    // Renumbering from a non-preopened fd to a preopened fd. In this case, we
+    // can't a keep the destination fd entry in the preopened table so remove
+    // it entirely.
+    else if (prestat_from_error != 0 && prestat_to_error == 0) {
+        error = fd_prestats_remove_entry(prestats, to);
+    }
+    // Renumbering from a preopened fd to a non-preopened fd
+    else if (prestat_from_error == 0 && prestat_to_error != 0) {
+        error = fd_prestats_remove_entry(prestats, from);
```
loganek, line 1177: not sure why would we remove the `from` entry here? I'd assume that both file descriptors should remain valid.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
lum1n0us: Not sure about ⬆️. 
my concern is can users open new files in the closed preopen-directory in their following code? Since the runtime opening the resource, it might not be right to let user code to control preopen resources.

tests/wamr-test-suites/wasi-test-script/run_wasi_tests.sh
```diff
@@ -96,7 +112,13 @@ else
     for testsuite in ${THREAD_STRESS_TESTS} ${THREAD_PROPOSAL_TESTS} ${THREAD_INTERNAL_TESTS}; do
         tests=$(ls ${testsuite}*.wasm)
         tests_array=($tests)
-        run_aot_tests "${tests_array[@]}"
+
+        readarray -t excluded_tests_array < <(jq -c \
+            --slurpfile testsuite_manifest $testsuite/manifest.json \
+            '.[$testsuite_manifest[0].name] // {} | keys[]' \
+            $TEST_FILTER)
```
zoraaver, line 119: It's been raised [here](https://github.com/bytecodealliance/wasm-micro-runtime/issues/2536) already but as a future improvement we could alter the wasi test runner script to add AOT support rather than duplicating the test runner logic (like filtering etc.) here.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
zoraaver: Hmm you're right it, closing the preopened fd won't stop the user from continuing to use any fd's derived from the preopened fd. But I'm not sure if that's an issue?

I don't have all the context but here's the [wasmtime commit](https://github.com/bytecodealliance/wasmtime/commit/67e2e57b02cd244c004379690aa56d3d5e187153) where this test was added. As stated in the commit message, wasi libc doesn't expose preopened fds to user code so the user still won't be able to close them intentionally as far as I understand.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -368,27 +368,39 @@ fd_prestats_grow(struct fd_prestats *pt, size_t min, size_t incr)
         pt->prestats = prestats;
         pt->size = size;
     }
-    return true;
+    return 0;
```
loganek, line 371:
```suggestion
    return __WASI_ESUCCESS;
```

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -368,27 +368,39 @@ fd_prestats_grow(struct fd_prestats *pt, size_t min, size_t incr)
...
+                          __wasi_fd_t fd)
 {
     // Grow the preopened resource table if needed.
-    rwlock_wrlock(&pt->lock);
-    if (!fd_prestats_grow(pt, fd, 1)) {
-        rwlock_unlock(&pt->lock);
-        return false;
+    __wasi_errno_t error = fd_prestats_grow(pt, fd, 1);
+
+    if (error != 0) {
```
loganek, line 381:
```suggestion
    if (error != __WASI_ESUCCESS) {
```

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -368,27 +368,39 @@ fd_prestats_grow(struct fd_prestats *pt, size_t min, size_t incr)
...
 
     pt->prestats[fd].dir = bh_strdup(dir);
-    rwlock_unlock(&pt->lock);
 
     if (pt->prestats[fd].dir == NULL)
-        return false;
+        return __WASI_ENOMEM;
 
-    return true;
+    return 0;
```
loganek, line 390:
```suggestion
    return __WASI_ESUCCESS;
```

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -1100,7 +1121,48 @@ wasmtime_ssp_fd_renumber(struct fd_table *curfds, struct fd_prestats *prestats,
...
+    }
+
+    rwlock_unlock(&prestats->lock);
     rwlock_unlock(&ft->lock);
+
+    if (error != 0) {
+        return error;
+    }
+
     return 0;
```
loganek, line 1166:
```suggestion
    return error;
```

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
loganek: Even though that perfectly matches the standard, I wonder if that won't introduce any backward compatibility within WAMR - @wenyongh / @xwang98 I wonder how should we handle this situations? Have we experienced such things in the past? Would it be enough to highlight this incompatibility in release notes?

yamt (changes requested):
posix.c change seems too big to sneak within a "Run rust tests from wasi-testsuite" PR.
i'd suggest to make it a separate PR.

zoraaver:
> posix.c change seems too big to sneak within a "Run rust tests from wasi-testsuite" PR. i'd suggest to make it a separate PR.

Hey, thanks for the suggestion. I've made the minimum required changes to get the additional rust WASI tests to pass. Without those changes, the PR will not pass in CI so it won't be possible to split the `posix.c` changes into a separate PR. 

The changes mainly concern behaviour around preopens but please let me know if the motivation behind any of the other changes is unclear.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
wenyongh: In my memory we didn't experience such things in libc-wasi, but agree to highlight this incompatibility in release notes. It may be a security issue to give unnecessary rights to the fd, we just fix it to match the standard.

zoraaver: @wenyongh thanks for confirming, I've moved this change to a separate PR, https://github.com/bytecodealliance/wasm-micro-runtime/pull/2579

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
zoraaver: @lum1n0us  raised a separate PR for this, https://github.com/bytecodealliance/wasm-micro-runtime/pull/2578. Please leave a comment there if you still have a concern or suggestion, thanks!

yamt:
failing wasi-threads tests should pass if you rebase. cf. https://github.com/bytecodealliance/wasm-micro-runtime/pull/2568

zoraaver:
@wenyongh would you mind taking another look? Build is green now so should be good to merge, thanks very much.

wenyongh (commented):
LGTM

Issue #2481, opened 2023-08-18
issues wrt clearing exceptions (traps)


yamt:
1. wamr has a few apis to clear exceptions. they are redundant.
    * wasm_runtime_set_exception(inst, NULL)
    * wasm_runtime_clear_exception(inst)

2. these api clear the exceptions in the all instances in the cluster. it doesn't seem to be what's intended in many places. eg. https://github.com/bytecodealliance/wasm-micro-runtime/blob/a2f76cf93c271b02db1456bbb05134b08809a883/core/iwasm/common/wasm_c_api.c#L3336

PR #2509, opened 2023-08-28
Stop abusing shared memory lock to protect exception

yamt:
Fixes: https://github.com/bytecodealliance/wasm-micro-runtime/issues/2407

### DIFF
--- a/core/iwasm/libraries/thread-mgr/thread_manager.c
+++ b/core/iwasm/libraries/thread-mgr/thread_manager.c
@@ -16,10 +16,6 @@
 #include "debug_engine.h"
 #endif
 
-#if WASM_ENABLE_SHARED_MEMORY != 0
-#include "wasm_shared_memory.h"
-#endif
-
 typedef struct {
     bh_list_link l;
     void (*destroy_cb)(WASMCluster *);
@@ -31,6 +27,8 @@ static bh_list *const destroy_callback_list = &destroy_callback_list_head;
 static bh_list cluster_list_head;
 static bh_list *const cluster_list = &cluster_list_head;
 static korp_mutex cluster_list_lock;
+
+static korp_mutex _exception_lock;
 
 typedef void (*list_visitor)(void *, void *);
 
@@ -52,6 +50,10 @@ thread_manager_init()
         return false;
     if (os_mutex_init(&cluster_list_lock) != 0)
         return false;
+    if (os_mutex_init(&_exception_lock) != 0) {
+        os_mutex_destroy(&cluster_list_lock);
+        return false;
+    }
     return true;
 }
 
@@ -66,6 +68,7 @@ thread_manager_destroy()
         cluster = next;
     }
     wasm_cluster_cancel_all_callbacks();
+    os_mutex_destroy(&_exception_lock);
     os_mutex_destroy(&cluster_list_lock);
 }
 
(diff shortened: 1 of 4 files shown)


core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -1240,72 +1243,52 @@ wasm_cluster_resume_all(WASMCluster *cluster)
...
-#endif
-        if (!strstr(wasm_inst->cur_exception, "wasi proc exit")) {
-            bh_memcpy_s(curr_wasm_inst->cur_exception,
-                        sizeof(curr_wasm_inst->cur_exception),
-                        wasm_inst->cur_exception,
-                        sizeof(wasm_inst->cur_exception));
+        exception_lock(wasm_inst);
+        if (data->exception != NULL) {
+            snprintf(wasm_inst->cur_exception, sizeof(wasm_inst->cur_exception),
+                     "Exception: %s", data->exception);
```
wenyongh, line 1264: The check for `Only spread non "wasi proc exit" exception` is ignored, see L1255, L1260 of the original file.

yamt: it's intentional. is it a problem? do you have a test case?

wenyongh: The check was introduced in PR #1988.

Raising "wasi proc exit" exception is an intentional behavior of runtime, it is not an actual exception but somewhat like setting a flag to let current thread stop running opcodes, and after the thread stops and in the end of wasm_runtime_call_wasm, the thread will clear this exception, so it ends normally without exception thrown. For multi-threading, the thread doesn't spread "wasi proc exit" exception, but just set terminate flags of other threads to let them exit also.

It may cause unexpected behavior if thread A spreads this exception to other threads: other thread (let's say thread B) may stop running opcodes first, then handle the "wasi proc exit" exception and clear exceptions of other threads, including thread A. When thread A's exception is cleared, it may continue to run and throw "unreachable" exception (Note that after calling wasi_proc_exit, in most cases the next opcode is unreachable, the bytecodes are generated by emsdk or wasi-sdk). And eventually "unreachable" exception is thrown.

I believe we found the issue when testing the was-thread related test cases and then we fixed it, the issue occurred occasionally. If we want to reproduce it, we may try running the wasi-thread cases many times.

yamt: > The check was introduced in PR #1988.
> 
> Raising "wasi proc exit" exception is an intentional behavior of runtime, it is not an actual exception but somewhat like setting a flag to let current thread stop running opcodes, and after the thread stops and in the end of wasm_runtime_call_wasm, the thread will clear this exception, so it ends normally without exception thrown. For multi-threading, the thread doesn't spread "wasi proc exit" exception, but just set terminate flags of other threads to let them exit also.

while proc exit is not a real trap, what the runtime should do is almost same as real traps.
ie. terminate all threads and return the exit/trap to the api user as the result of the whole "thread group".

> 
> It may cause unexpected behavior if thread A spreads this exception to other threads: other thread (let's say thread B) may stop running opcodes first, then handle the "wasi proc exit" exception and clear exceptions of other threads, including thread A. When thread A's exception is cleared, it may continue to run and throw "unreachable" exception (Note that after calling wasi_proc_exit, in most cases the next opcode is unreachable, the bytecodes are generated by emsdk or wasi-sdk). And eventually "unreachable" exception is thrown.

if it's a problem, real traps have the same problems, don't they?

my impression is that many (all?) of the code clearing other threads' exception are just broken: https://github.com/bytecodealliance/wasm-micro-runtime/issues/2481

> 
> I believe we found the issue when testing the was-thread related test cases and then we fixed it, the issue occurred occasionally. If we want to reproduce it, we may try running the wasi-thread cases many times.

i guess i will restore this (IMO wrong) behavior for now because it isn't the main point of this PR.

yamt: > i guess i will restore this (IMO wrong) behavior for now because it isn't the main point of this PR.

done

yamt: > > if it's a problem, real traps have the same problems, don't they?
> 
> No, real traps are spread to other threads and terminate flags are also set for other threads, but the trap isn't cleared before the thread ends, so thread A's exception won't be cleared by thread B.

a real trap can misbehave in a similar way if the exception is suddenly cleared by the other thread.

> > my impression is that many (all?) of the code clearing other threads' exception are just broken: #2481
> 
> Do you mean to unify `wasm_runtime_set_exception(inst, NULL)` and `wasm_runtime_clear_exception(inst)`, and to remove some unneeded exception clear?

yes.
unifying two api is just cosmetic.
the other one is a bit cumbersome. i guess we need to investigate one-by-one to see if it should clear other threads' exceptions. (i guess most of them need to clear only the local exception.)

wenyongh: OK, thanks, it really takes effort to investigate them one by one.

wenyongh (commented):
LGTM

xujuntwt95329:
LGTM

PR #2516, opened 2023-08-30
Implement async termination of blocking thread

yamt:
send a signal whose handler is no-op to a blocking thread to wake up the blocking syscall
with either EINTR equivalent or partial success.

unlike the approach taken in the `dev/interrupt_block_insn` branch (that is, signal + longjmp similarly to `OS_ENABLE_HW_BOUND_CHECK`),
this PR does not use longjmp because:
* longjmp from signal handler doesn't work on nuttx (https://github.com/apache/nuttx/issues/10326)
* i'm afraid that the singal+longjmp approach is too difficult for average programmers who might implement host functions to deal with. especially when the longjmp can rewind a lot of functions, some of which are potentially user-supplied.

todo/known issues/limitations/open questions:
* ~~make the initialization independent from `OS_ENABLE_HW_BOUND_CHECK`~~
* ~~provide an api (or at least a build-time config) to specifiy a signal to use, instead of hardcoding `SIGUSR1`.~~
* adapt operations other than `fd_read`
  * ~~fd_close~~
  * ~~fd_pread~~
  * ~~fd_pwrite~~
  * ~~fd_read~~
  * ~~fd_write~~
  * ~~path_open~~
    * i guess it can block on eg. fifo. i'm not sure if i want to fix it in this PR.
  * poll_oneoff
    * leave this for now as it's already fixed with a different approach: https://github.com/bytecodealliance/wasm-micro-runtime/pull/1951
  * ~~sock_accept~~
  * ~~sock_close (?)~~ just ENOSYS
  * ~~sock_connect~~
  * ~~sock_recv~~ just a wrapper of sock_recv_from
  * ~~sock_recv_from~~
  * ~~sock_send~~ just a wrapper of sock_send_to
  * ~~sock_send_to~~
  * ~~sock_shutdown~~ usually not block
  * ~~sock_addr_resolve~~
    * this one is not a simple syscall. see the comment in blocking_op.c
  * memory.atomic.wait32/64 
    * leave this for now as it's already fixed with a different approach: https://github.com/bytecodealliance/wasm-micro-runtime/pull/2024
* what to do for platforms other than posix?
* what to do for other wasi implementation? (uvwasi)
* ~~document/comment~~
* test
  * the relevant tests in wasi-threads repo passed on macOS and nuttx/esp32-devkitc
* ~~add an api to terminate an instance~~ https://github.com/bytecodealliance/wasm-micro-runtime/pull/2538
* ~~add a sample~~ https://github.com/bytecodealliance/wasm-micro-runtime/pull/2538
* for simple blocking system call like `readv`, it's easy to wrap it with `wasm_runtime_begin_blocking_op`/`wasm_runtime_end_blocking_op`. but how about something more complex, like a call which possibly blocks deep inside an external library?

relevant issue: https://github.com/bytecodealliance/wasm-micro-runtime/issues/1910


yamt:
the last push passes all test cases in wasi-threads repo on macOS.

core/iwasm/include/wasm_export.h
```diff
@@ -1546,6 +1546,47 @@ wasm_runtime_set_context_spread(wasm_module_inst_t inst, void *key,
...
+/*
+ * wasm_runtime_begin_blocking_op/wasm_runtime_end_blocking_op
+ *
+ * These APIs are intended to be used by the implementations of
+ * host functions. It wraps an operation which possibly blocks for long
+ * to prepare for async termination.
+ *
+ * eg.
+ *
+ *   if (!wasm_runtime_begin_blocking_op) {
```
wenyongh, line 1558: Had better `wasm_runtime_begin_blocking_op(exec_env)`?

core/iwasm/include/wasm_export.h
```diff
@@ -1546,6 +1546,47 @@ wasm_runtime_set_context_spread(wasm_module_inst_t inst, void *key,
...
+ *       return EINTR;
+ *   }
+ *   ret = possibly_blocking_op();
+ *   wasm_runtime_end_blocking_op(env);
```
wenyongh, line 1562: Had better `wasm_runtime_end_blocking_op(exec_env)`?

core/iwasm/include/wasm_export.h
```diff
@@ -1546,6 +1546,47 @@ wasm_runtime_set_context_spread(wasm_module_inst_t inst, void *key,
...
+ * `os_wakeup_blocking_op` call makes the blocking operation
+ * (`possibly_blocking_op` in the above example) return in a timely manner.
+ *
+ * The actual wake up mechanism used by `os_wakeup_blocking_op` is
+ * platform-dependent. It might impose some platform-dependent restrictions
+ * on the implementation of the blocking opearation.
+ *
+ * For example, on POSIX-like platforms, a signal (by default SIGUSR1) is
+ * used. The signal delivery configurations (eg. signal handler, signal mask,
+ * etc) for the signal is set up by the runtime. You can change the signal
```
wenyongh, line 1580: `are set up`?

core/shared/platform/common/posix/posix_blocking_op.c
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright (C) 2023 Midokura Japan KK.  All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+
+#include "platform_api_extension.h"
+
+#ifdef OS_ENABLE_WAKEUP_BLOCKING_OP
```
wenyongh, line 8: Not whether we should make the feature configurable, e.g. `cmake -DWAMR_BUILD_WAKEUP_BLOCKING_OP=1/0`?

core/iwasm/include/wasm_export.h
wenyongh: Should we mention that two APIs take effect only when thread manager is enabled?

wenyongh:
@eloparco Could you help review the PR? It implements feature like `dev/interrupt_block_insn` in different manner.

core/iwasm/include/wasm_export.h
yamt: sure. done.

core/iwasm/include/wasm_export.h
yamt: sure. done.

core/iwasm/include/wasm_export.h
yamt: fixed. thank you.

core/shared/platform/common/posix/posix_blocking_op.c
yamt: are you suggesting to make it a build-time configurable?
i myself do not have a strong opinion either ways.

wenyongh: Yes, build-time configurable. How about adding `cmake -DWAMR_BUILD_WAKEUP_BLOCKING_OP=1/0`, and by default it is enabled? My concern is that some one may want to disable it if not needed, and it can reduce binary (though little).

And if added, had better update doc/build_warm.md accordingly.

yamt: ok.
in that case, i want to make it similar to hw_bound_check.
do you know why WASM_DISABLE_HW_BOUND_CHECK uses a negative notation? ("disable" rather than "enable")

wenyongh: Because the hw bound check for some 64-bit platforms is enabled by default, using `cmake -DWAMR_BUILD_HW_BOUND_CHECK=1/0` makes a little confusing, we used `cmake -DWAMR_DISABLE_HW_BOUND_CHECK=1` instead, and then the macro `WASM_DISABLE_HW_BOUND_CHECK` was used accordingly. Maybe it is not so consistent with other cmake variables and compiler macros :)

yamt: so, how about:

* make it enabled by default where available
* introduce `-DWAMR_DISABLE_WAKEUP_BLOCKING_OP=1/0`
* define `OS_ENABLE_WAKEUP_BLOCKING_OP` only when enabled

wenyongh: It is ok for me, looks well.

eloparco:
> @eloparco Could you help review the PR? It implements feature like dev/interrupt_block_insn in different manner.

Sure, I'll take a look in the next couple of days.

core/shared/platform/common/posix/posix_blocking_op.c
yamt: done

core/iwasm/common/wasm_blocking_op.c
```diff
@@ -0,0 +1,91 @@
...
+    LOCK(env);
+    SET(env, TERMINATE);
+    while (ISSET(env, BLOCKING)) {
+        UNLOCK(env);
+        os_wakeup_blocking_op(env->handle);
+
+        /* relax a bit */
+        os_usleep(50 * 1000);
+        LOCK(env);
+    }
```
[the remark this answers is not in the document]
yamt: fixed. thank you.

wenyongh (commented):
LGTM

eloparco:
Took a look. This doesn't cover user-defined functions that could have been achieved with `longjmp` I think. But, as you said, the `longjmp` approach is probably too complicated.

We should be able to add [these tests](https://github.com/WebAssembly/wasi-threads/pull/41) now, probably in a separate PR.

yamt:
> Took a look. This doesn't cover user-defined functions that could have been achieved with `longjmp` I think. But, as you said, the `longjmp` approach is probably too complicated.

user-defined functions should take care of it by itself, either using the api introduced by this PR or other approaches like https://github.com/bytecodealliance/wasm-micro-runtime/pull/1951. i guess in many cases it's simpler than preparing for signal+longjmp.

> We should be able to add [these tests](https://github.com/WebAssembly/wasi-threads/pull/41) now, probably in a separate PR.

yes. they all pass with this PR: https://github.com/bytecodealliance/wasm-micro-runtime/pull/2516#issuecomment-1701890000

wenyongh:
> Took a look. This doesn't cover user-defined functions that could have been achieved with `longjmp` I think. But, as you said, the `longjmp` approach is probably too complicated.
> 
> We should be able to add [these tests](https://github.com/WebAssembly/wasi-threads/pull/41) now, probably in a separate PR.

@eloparco so the PR is good to you, right? Shall we merge the PR?

eloparco:
> @eloparco so the PR is good to you, right? Shall we merge the PR?

LGTM

PR #2376, opened 2023-07-21
Dev/gc refactor

TianlongLiang:
Modify opcodes:
1.	WASM_OP_CALL_INDIRECT
2.	WASM_OP_SELECT_T 
3.	WASM_OP_TABLE_GET 
4.	WASM_OP_TABLE_SET
5.	WASM_OP_REF_NULL 
6.	WASM_OP_REF_IS_NULL 
7.	WASM_OP_REF_FUNC
8.	WASM_OP_TABLE_INIT
9.	WASM_OP_TABLE_GROW
10.	WASM_OP_TABLE_FILL 
11.	WASM_OP_GET_LOCAL
12.	WASM_OP_SET_LOCAL
13.	WASM_OP_TEE_LOCAL

Currently, put gc APIs wrapper in a new file `aot_gc_object_wrapper.c`

PR #2518, opened 2023-08-30
Enable spec test for AOT GC

no1wudi:


no1wudi:
@wenyongh @TianlongLiang I found that #2376 introduced some issue in AOT GC spectest, such as `as-call_indirect-first` in `block.wast`

And number of failed cases increased from 9 -> 17, please try `./test_wamr.sh -s spec -G -b -t aot -P` with this PR.

TianlongLiang:
I am not 100% sure, but I think this has something to do with the change of table element type. Some opcodes like `call_indirect` are using the table element as a gc reference rather than a function index now. But currently, in the gc branch, we are still using the old loading and instantiating method, the table element will still be a function index at this point in time. That explains why more opcodes will fail after the PR.

PR #2522, opened 2023-08-30
Add callback to handle memory.grow failures

eloparco:
When embedding WAMR, this PR allows to register a callback that is invoked when `memory.grow` fails.

In case of memory allocation failures, some languages allow to handle the error (e.g. by checking the return code of `malloc`/`calloc` in c), some others (e.g. Rust) just panic. For the latter, it is useful to have a callback function that would be called before the wasm execution crashes, allowing the embedder to detect and handle the `memory.grow` error. It comes in handy when using shared memory since shared memory forces the maximum linear memory size to be set at compilation time. Setting a value that is too low can lead to memory allocation failures (i.e. `memory.grow` failures).

It would be nice to capture the crash in Rust directly, but the feature is experimental (https://github.com/rust-lang/rust/issues/51540).

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -1443,6 +1443,9 @@ aot_call_function(WASMExecEnv *exec_env, AOTFunctionInstance *function,
     /* set thread handle and stack boundary */
     wasm_exec_env_set_thread_info(exec_env);
 
+    /* set exec env so it can be later retrieved from instance */
+    ((AOTModuleInstanceExtra *)module_inst->e)->common.cur_exec_env = exec_env;
+
     if (ext_ret_count > 0) {
         uint32 cell_num = 0, i;
         uint8 *ext_ret_types = func_type->types + func_type->param_count + 1;
--- a/core/iwasm/common/wasm_memory.c
+++ b/core/iwasm/common/wasm_memory.c
@@ -8,6 +8,7 @@
 #include "../aot/aot_runtime.h"
 #include "bh_platform.h"
 #include "mem_alloc.h"
+#include "wasm_memory.h"
 
 #if WASM_ENABLE_SHARED_MEMORY != 0
 #include "../common/wasm_shared_memory.h"
@@ -23,6 +24,8 @@ typedef enum Memory_Mode {
 static Memory_Mode memory_mode = MEMORY_MODE_UNKNOWN;
 
 static mem_allocator_t pool_allocator = NULL;
+
+static memory_grow_error_callback memory_grow_error_cb;
 
 #if WASM_MEM_ALLOC_WITH_USER_DATA != 0
 static void *allocator_user_data = NULL;
@@ -763,6 +766,12 @@ wasm_enlarge_memory_internal(WASMModuleInstance *module, uint32 inc_page_count)
 }
 #endif /* end of OS_ENABLE_HW_BOUND_CHECK */
 
+void
+set_memory_grow_error_callback(const memory_grow_error_callback callback)
+{
+    memory_grow_error_cb = callback;
+}
+
 bool
 wasm_enlarge_memory(WASMModuleInstance *module, uint32 inc_page_count)
 {
@@ -777,6 +786,9 @@ wasm_enlarge_memory(WASMModuleInstance *module, uint32 inc_page_count)
     if (module->memory_count > 0)
         shared_memory_unlock(module->memories[0]);
 #endif
+
+    if (!ret && memory_grow_error_cb)
+        memory_grow_error_cb(inc_page_count);
 
     return ret;
 }
--- a/core/iwasm/common/wasm_memory.c
+++ b/core/iwasm/common/wasm_memory.c
@@ -700,11 +700,24 @@ wasm_enlarge_memory_internal(WASMModuleInstance *module, uint32 inc_page_count)
 #endif
 
 return_func:
-    if (!ret && enlarge_memory_error_cb)
+    if (!ret && enlarge_memory_error_cb) {
+        WASMExecEnv *exec_env = NULL;
+
+#if WASM_ENABLE_INTERP != 0
+        if (module->module_type == Wasm_Module_Bytecode)
+            exec_env =
+                ((WASMModuleInstanceExtra *)module->e)->common.cur_exec_env;
+#endif
+#if WASM_ENABLE_AOT != 0
+        if (module->module_type == Wasm_Module_AoT)
+            exec_env =
+                ((AOTModuleInstanceExtra *)module->e)->common.cur_exec_env;
+#endif
+
         enlarge_memory_error_cb(inc_page_count, total_size_old, 0,
-                                failure_reason, (wasm_module_inst_t)module,
-                                wasm_runtime_get_exec_env_singleton(
-                                    (WASMModuleInstanceCommon *)module));
+                                failure_reason,
+                                (WASMModuleInstanceCommon *)module, exec_env);
+    }
 
     return ret;
 }
@@ -793,11 +806,24 @@ wasm_enlarge_memory_internal(WASMModuleInstance *module, uint32 inc_page_count)
 #endif
 
 return_func:
-    if (!ret && enlarge_memory_error_cb)
+    if (!ret && enlarge_memory_error_cb) {
+        WASMExecEnv *exec_env = NULL;
+
+#if WASM_ENABLE_INTERP != 0
+        if (module->module_type == Wasm_Module_Bytecode)
+            exec_env =
+                ((WASMModuleInstanceExtra *)module->e)->common.cur_exec_env;
+#endif
+#if WASM_ENABLE_AOT != 0
+        if (module->module_type == Wasm_Module_AoT)
+            exec_env =
+                ((AOTModuleInstanceExtra *)module->e)->common.cur_exec_env;
+#endif
+
         enlarge_memory_error_cb(inc_page_count, total_size_old, 0,
-                                failure_reason, (wasm_module_inst_t)module,
-                                wasm_runtime_get_exec_env_singleton(
-                                    (WASMModuleInstanceCommon *)module));
+                                failure_reason,
+                                (WASMModuleInstanceCommon *)module, exec_env);
+    }
 
     return ret;
 }
--- a/core/iwasm/common/wasm_memory.c
+++ b/core/iwasm/common/wasm_memory.c
@@ -25,7 +25,7 @@ static Memory_Mode memory_mode = MEMORY_MODE_UNKNOWN;
 
 static mem_allocator_t pool_allocator = NULL;
 
-static memory_grow_error_callback memory_grow_error_cb;
+static enlarge_memory_error_callback_t enlarge_memory_error_cb;
 
 #if WASM_MEM_ALLOC_WITH_USER_DATA != 0
 static void *allocator_user_data = NULL;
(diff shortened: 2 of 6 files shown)


yamt:
>  allowing the embedder to detect and handle the memory.grow error

do you mean just printing a diagnostic message?
or are you thinking about more interesting error handling?

eloparco:
> do you mean just printing a diagnostic message?
or are you thinking about more interesting error handling?

My use case is more about logging, storing some stats and doing some cleanup in the embedder. Without that, the execution can crash at any point (in which there's an allocation) and from the embedder there won't be any way to automatically understand that the cause is the `memory.grow` failure.
I guess there are other possible scenarios and exposing the callback can unlock them.

yamt:
> > do you mean just printing a diagnostic message?
> > or are you thinking about more interesting error handling?
> 
> My use case is more about logging, storing some stats and doing some cleanup in the embedder. Without that, the execution can crash at any point (in which there's an allocation) and from the embedder there won't be any way to automatically understand that the cause is the `memory.grow` failure. I guess there are other possible scenarios and exposing the callback can unlock them.

for logging purpose, i guess it's better to provide more info like:
* context
  * wasm_exec_env_t
* memory
  * module instance
  * memory idx (currently always 0 though)
  * current size of the memory
* the cause of the failure

core/iwasm/common/wasm_memory.c
```diff
@@ -24,6 +25,8 @@ static Memory_Mode memory_mode = MEMORY_MODE_UNKNOWN;
 
 static mem_allocator_t pool_allocator = NULL;
 
+static memory_grow_error_callback memory_grow_error_cb;
```
loganek, line 28:
```suggestion
static enlarge_memory_error_callback_t enlarge_memory_error_cb;
```
for consistency

Zzzabiyaka:
just a general note: is there any specific reason why we need `goto` statement here? 

why cannot we call a separate function and then just `return false`?

maybe I'm missing some context

core/iwasm/include/wasm_export.h
```diff
@@ -1439,6 +1439,23 @@ WASM_RUNTIME_API_EXTERN bool
...
+typedef void (*enlarge_memory_error_callback_t)(
+    uint32_t inc_page_count, uint64_t current_memory_size,
+    uint32_t memory_index, enlarge_memory_error_reason_t failure_reason,
+    wasm_module_inst_t instance, wasm_exec_env_t exec_env);
+
+/**
+ * Setup callback invoked when memory.grow fails
+ */
+WASM_RUNTIME_API_EXTERN void
+set_enlarge_memory_error_callback(
```
wenyongh, line 1456: Had better add prefix `wasm_runtime_`, such as `wasm_runtime_set_enlarge_memory_error_callback`, if the API name is too long, maybe we can change to `wasm_runtime_set_enlarge_mem_error_callback` or something else?

eloparco:
> just a general note: is there any specific reason why we need goto statement here?

It just looked cleaner than calling `enlarge_memory_error_cb()` in all the places where `wasm_enlarge_memory_internal()` returns.

core/iwasm/include/wasm_export.h
eloparco: Right, done

core/iwasm/common/wasm_memory.c
```diff
@@ -685,19 +699,30 @@ wasm_enlarge_memory_internal(WASMModuleInstance *module, uint32 inc_page_count)
     os_writegsbase(memory_data_new);
 #endif
 
+return_func:
+    if (!ret && enlarge_memory_error_cb)
+        enlarge_memory_error_cb(inc_page_count, total_size_old, 0,
+                                failure_reason, (wasm_module_inst_t)module,
+                                wasm_runtime_get_exec_env_singleton(
```
wenyongh, line 706: Just found an issue: the exec_env returned by `wasm_runtime_get_exec_env_singleton` is not always the exec_env that current module instance uses (developer can also call `wasm_runtime_create_exec_env`), using it in the embedder may cause unexpected behavior.

There may be two possible ways: one is to add an extra argument `exec_env` in `wasm_enlarge_memory` and `aot_enlarge_memory`, and pass the correct exec_env from interpreter, fast-jit and AOT code.

The other is to add a field in module instance to store the current exec_env, e.g. WASMModuleInstanceExtra and AOTModuleInstanceExtra, set it every time in `wasm_call_function` and `aot_call_function`.

Maybe the latter is better as the former causes the ABI change of AOT code calling runtime API.

eloparco: > The other is to add a field in module instance to store the current exec_env, e.g. WASMModuleInstanceExtra and AOTModuleInstanceExtra, set it every time in wasm_call_function and aot_call_function.

Tried that, but something got broken for AOT (CI is failing).

core/iwasm/aot/aot_runtime.c
```diff
@@ -1413,6 +1413,8 @@ aot_call_function(WASMExecEnv *exec_env, AOTFunctionInstance *function,
                   unsigned argc, uint32 argv[])
 {
     AOTModuleInstance *module_inst = (AOTModuleInstance *)exec_env->module_inst;
+    module_inst->e->common.exec_env = exec_env;
```
wenyongh, line 1416: Should `((AOTModuleInstanceExtra *)module_inst->e)->common.exec_env`, note that AOTModuleInstanceExtra is different from WASMModuleInstanceExtra

wenyongh: And could you put the line after L1446 `wasm_exec_env_set_thread_info(exec_env)` and add comment?

core/iwasm/common/wasm_memory.c
```diff
@@ -685,19 +699,29 @@ wasm_enlarge_memory_internal(WASMModuleInstance *module, uint32 inc_page_count)
     os_writegsbase(memory_data_new);
 #endif
 
+return_func:
+    if (!ret && enlarge_memory_error_cb)
+        enlarge_memory_error_cb(inc_page_count, total_size_old, 0,
+                                failure_reason, (wasm_module_inst_t)module,
+                                module->e->common.exec_env);
```
wenyongh, line 706: Should distinguish module->e, it may be WASMModuleInstanceExtra or AOTModuleInstanceExtra

core/iwasm/common/wasm_memory.c
```diff
@@ -759,10 +791,23 @@ wasm_enlarge_memory_internal(WASMModuleInstance *module, uint32 inc_page_count)
     memory->mem_bound_check_16bytes.u64 = total_size_new - 16;
 #endif
 
-    return true;
+return_func:
+    if (!ret && enlarge_memory_error_cb)
+        enlarge_memory_error_cb(inc_page_count, total_size_old, 0,
+                                failure_reason, (wasm_module_inst_t)module,
+                                module->e->common.exec_env);
```
wenyongh, line 798: Same as above.

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -2396,6 +2396,7 @@ wasm_call_function(WASMExecEnv *exec_env, WASMFunctionInstance *function,
 {
     WASMModuleInstance *module_inst =
         (WASMModuleInstance *)exec_env->module_inst;
+    module_inst->e->common.exec_env = exec_env;
```
wenyongh, line 2399: Had better add an empty line above or put it after `wasm_exec_env_set_thread_info(exec_env)`

core/iwasm/interpreter/wasm_runtime.h
```diff
@@ -213,6 +213,7 @@ typedef struct CApiFuncImport {
 /* The common part of WASMModuleInstanceExtra and AOTModuleInstanceExtra */
 typedef struct WASMModuleInstanceExtraCommon {
     CApiFuncImport *c_api_func_imports;
+    WASMExecEnv *exec_env;
```
wenyongh, line 216: How about rename the field to `cur_exec_env` and add comment?

core/iwasm/common/wasm_memory.c
wenyongh: It seems to be due to `AOTModuleInstanceExtra` is different from `WASMModuleInstanceExtra`, could you see the comments and try again?

eloparco: Done, thanks

core/iwasm/common/wasm_memory.c
```diff
@@ -685,19 +699,39 @@ wasm_enlarge_memory_internal(WASMModuleInstance *module, uint32 inc_page_count)
...
+            inc_page_count, total_size_old, 0, failure_reason,
+            (WASMModuleInstanceCommon *)module,
+            ((WASMModuleInstanceExtra *)module->e)->common.cur_exec_env);
+#endif
+#if WASM_ENABLE_AOT != 0
+        enlarge_memory_error_cb(
+            inc_page_count, total_size_old, 0, failure_reason,
+            (WASMModuleInstanceCommon *)module,
+            ((AOTModuleInstanceExtra *)module->e)->common.cur_exec_env);
+#endif
```
wenyongh, line 715: The module instance can be AOTModuleInstance when macro WASM_ENABLE_INTERP is 1, suppose WASM_ENABLE_AOT is also enabled. Had better change to be like:
```C
    if (!ret && enlarge_memory_error_cb) {
        WASMExecEnv *exec_env = NULL;

#if WASM_ENABLE_INTERP != 0
        if (module->module_type == Wasm_Module_Bytecode)
            exec_env = ((WASMModuleInstanceExtra *)module->e)->common.cur_exec_env;
#endif
#if WASM_ENABLE_AOT != 0
        if (module->module_type == Wasm_Module_AoT)
            exec_env = ((AOTModuleInstanceExtra *)module->e)->common.cur_exec_env;
#endif

        enlarge_memory_error_cb(
            inc_page_count, total_size_old, 0, failure_reason,
            (WASMModuleInstanceCommon *)module, exec_env);
    }
```

core/iwasm/common/wasm_memory.c
```diff
@@ -759,10 +801,33 @@ wasm_enlarge_memory_internal(WASMModuleInstance *module, uint32 inc_page_count)
...
```
wenyongh, line 817: Same as above

wenyongh (commented):
LGTM

yamt (approved):
lgtm

lum1n0us:
> Without that, the execution can crash at any point (in which there's an allocation) and from the embedder there won't be any way to automatically understand that the cause is the memory.grow failure.

Be curious about the use case. Does it like 
- memory.grow failed in runtime
- rust(user code or std) detects the failure and panic!
- embedded doesn't know anything but WASM app exits.

> I guess there are other possible scenarios and exposing the callback can unlock them.

It seems like a general global exception handler with variants exceptions is better than a set of callbacks for every possible scenarios.

eloparco:
> Be curious about the use case. Does it like
> - memory.grow failed in runtime
> - rust(user code or std) detects the failure and panic!
> - embedded doesn't know anything but WASM app exits.

Yes, that's my use case

> It seems like a general global exception handler with variants exceptions is better than a set of callbacks for every possible scenarios.

Do you mean to implement a general global exception handler in the runtime?
The best option would probably be to override the exception handler for memory allocation in Rust, but that feature is experimental. The problem doesn't appear in C since there we can just check the return code of `malloc`/`realloc` in user code.

lum1n0us:
> The problem doesn't appear in C since there we can just check the return code of malloc/realloc in user code.

Does the *uncheckable exception* only happen during allocation or it is a general behaviour in Rust? Is it safe to say `panic!->abort() (Rust->Wasm)` is the root cause of such scenarios?

eloparco:
> Does the uncheckable exception only happen during allocation or it is a general behaviour in Rust?

Yes, the one that I was trying to address happens in Rust whenever a memory allocation fails. It can be raised by any library included (e.g. standard library) and currently there's no way to catch it. To reproduce it, you can just set the maximum linear memory size to a small value and compile some Rust code that allocates memory.

> Is it safe to say panic!->abort() (Rust->Wasm) is the root cause of such scenarios?

I think so.

loganek:
I think it's worth to mention in this thread that (a cross-runtime) alternative for that would be to add a compiler pass that wraps memory.grow opcode with some additional error checking logic. And that can be extended to handle any other future usecases.

lum1n0us:
Is it safe to say the embedded will meet an un-handled exception when the rust code hit a `panic!`? no matter it is caused by allocation failure, out-of-length index accessing or anything else.

If so, shall we avoid `panic!` especially when plan to be transferred to Wasm? Maybe use `Result<T,E>` as a return instead

eloparco:
>If so, shall we avoid panic! especially when plan to be transferred to Wasm? Maybe use Result<T,E> as a return instead

The `panic!` is not called explicitly (in the user code compiled to wasm) but happens whenever Rust fails memory allocation. Failure to allocate memory is an unrecoverable error in Rust.

lum1n0us:
Is it safe to say the embedded will meet an un-handled exception when the rust code hit a `panic!` no matter if it is called by user code(compiled to wasm) or not.

eloparco:
>Is it safe to say the embedded will meet an un-handled exception when the rust code hit a panic! no matter if it is called by user code(compiled to wasm) or not.

Yes

PR #2523, opened 2023-08-31
Add missing header to win_socket.c

loganek:
This fixes build for Universal Windows Platform configuration

### DIFF
--- a/core/shared/platform/windows/win_socket.c
+++ b/core/shared/platform/windows/win_socket.c
@@ -6,6 +6,8 @@
 #include "platform_api_vmcore.h"
 #include "platform_api_extension.h"
 
+#include <ws2tcpip.h>
+
 /* link with Ws2_32.lib */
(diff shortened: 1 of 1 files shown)


core/shared/platform/windows/win_socket.c
```diff
@@ -6,6 +6,8 @@
 #include "platform_api_vmcore.h"
 #include "platform_api_extension.h"
 
+#include <ws2tcpip.h>
```
wenyongh, line 9: The header file has already been included in "core/shared/platform/windows/platform_internal.h", which is then included in "platform_api_vmcore.h", wondering why there is build issue? What is the error reported?

loganek: oh, you're right. I was getting errors about `inet_pton` not being found, but that must have been some weirdness of my build configuration. After trying again, there's no error even when the header here is not added.

PR #2525, opened 2023-08-31
set_exception_visitor: remove the special case for wasi proc exit

yamt:
while wasi proc exit is not a real trap, what the runtime does on it is mostly same as real traps. that is, kill the siblings threads and represent the exit/trap as the result of the "process" to the user api. there seems no reason to distinuguish it from real traps here.

note that:

* the target thread either doesn't care the specific exception type or ignore wasi proc exit by themselves. (clear_wasi_proc_exit_exception)

* clear_wasi_proc_exit_exception only clears local exception.

### DIFF
--- a/core/iwasm/libraries/thread-mgr/thread_manager.c
+++ b/core/iwasm/libraries/thread-mgr/thread_manager.c
@@ -1260,12 +1260,8 @@ set_exception_visitor(void *node, void *user_data)
 
         exception_lock(wasm_inst);
         if (data->exception != NULL) {
-            /* Only spread non "wasi proc exit" exception */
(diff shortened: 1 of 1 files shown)


wenyongh (commented):
LGTM

@eloparco, could you help review this PR? Seems the issue mentioned in #1988 was already resolved since function `clear_wasi_proc_exit_exception` only clears the exception of current thread, so we can spread "wasi proc exit" exception to other threads to let them exit ASAP (the terminate flag is also set to other threads).

eloparco:
Things have changed since that PR and it seems that the distinction is not needed anymore. We have multiple tests that check exception propagation and the CI is green, so we should be confident to merge.

PR #2529, opened 2023-09-04
handle a return from wasi _start function correctly

yamt:
this fixes a few test cases in wasi-threads testsuite like wasi_threads_return_main_block.

also, move the special handling for wasi proc exit to a more appropriate place.

### DIFF
--- a/core/iwasm/common/wasm_application.c
+++ b/core/iwasm/common/wasm_application.c
@@ -110,12 +110,24 @@ execute_main(WASMModuleInstanceCommon *module_inst, int32 argc, char *argv[])
         const char *wasi_proc_exit_exception = "wasi proc exit";
 
         ret = wasm_runtime_call_wasm(exec_env, func, 0, NULL);
-        /* a successful return from _start is same as proc_exit(0). */
+#if WASM_ENABLE_THREAD_MGR != 0
         if (ret) {
+            /* On a successful return from the `_start` function,
+               we terminate other threads by mimicing wasi:proc_exit(0).
+
+               Note:
+               - A return from the `main` function is an equivalent of
+                 exit(). (C standard)
+               - When exit code is 0, wasi-libc's `_start` function just
+                 returns w/o calling `proc_exit`.
+               - A process termination should terminate threads in
+                 the process. */
+
             wasm_runtime_set_exception(module_inst, wasi_proc_exit_exception);
             /* exit_code is zero-initialized */
             ret = false;
         }
+#endif
         /* report wasm proc exit as a success */
         WASMModuleInstance *inst = (WASMModuleInstance *)module_inst;
         if (!ret && strstr(inst->cur_exception, wasi_proc_exit_exception)) {
--- a/core/iwasm/common/wasm_application.c
+++ b/core/iwasm/common/wasm_application.c
@@ -107,7 +107,22 @@ execute_main(WASMModuleInstanceCommon *module_inst, int32 argc, char *argv[])
        the actual main function. Directly calling main function
        may cause exception thrown. */
     if ((func = wasm_runtime_lookup_wasi_start_function(module_inst))) {
-        return wasm_runtime_call_wasm(exec_env, func, 0, NULL);
+        const char *wasi_proc_exit_exception = "wasi proc exit";
+
+        ret = wasm_runtime_call_wasm(exec_env, func, 0, NULL);
+        /* a successful return from _start is same as proc_exit(0). */
+        if (ret) {
+            wasm_runtime_set_exception(module_inst, wasi_proc_exit_exception);
+            /* exit_code is zero-initialized */
+            ret = false;
+        }
+        /* report wasm proc exit as a success */
+        WASMModuleInstance *inst = (WASMModuleInstance *)module_inst;
+        if (!ret && strstr(inst->cur_exception, wasi_proc_exit_exception)) {
+            inst->cur_exception[0] = 0;
+            ret = true;
+        }
+        return ret;
     }
 #endif /* end of WASM_ENABLE_LIBC_WASI */
 
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -1938,33 +1938,6 @@ wasm_runtime_finalize_call_function(WASMExecEnv *exec_env,
 }
 #endif
 
-static bool
-clear_wasi_proc_exit_exception(WASMModuleInstanceCommon *module_inst_comm)
-{
-#if WASM_ENABLE_LIBC_WASI != 0
-    bool has_exception;
-    char exception[EXCEPTION_BUF_LEN];
-    WASMModuleInstance *module_inst = (WASMModuleInstance *)module_inst_comm;
-
-    bh_assert(module_inst_comm->module_type == Wasm_Module_Bytecode
-              || module_inst_comm->module_type == Wasm_Module_AoT);
-
-    has_exception = wasm_copy_exception(module_inst, exception);
-    if (has_exception && !strcmp(exception, "Exception: wasi proc exit")) {
-        /* The "wasi proc exit" exception is thrown by native lib to
-           let wasm app exit, which is a normal behavior, we clear
-           the exception here. And just clear the exception of current
-           thread, don't call `wasm_set_exception(module_inst, NULL)`
-           which will clear the exception of all threads. */
-        module_inst->cur_exception[0] = '\0';
-        return true;
-    }
-    return false;
-#else
-    return false;
-#endif
-}
-
 bool
 wasm_runtime_call_wasm(WASMExecEnv *exec_env,
                        WASMFunctionInstanceCommon *function, uint32 argc,
@@ -2005,15 +1978,10 @@ wasm_runtime_call_wasm(WASMExecEnv *exec_env,
                                 param_argc, new_argv);
 #endif
     if (!ret) {
-        if (clear_wasi_proc_exit_exception(exec_env->module_inst)) {
-            ret = true;
+        if (new_argv != argv) {
+            wasm_runtime_free(new_argv);
         }
-        else {
-            if (new_argv != argv) {
-                wasm_runtime_free(new_argv);
-            }
-            return false;
-        }
+        return false;
     }
 
 #if WASM_ENABLE_REF_TYPES != 0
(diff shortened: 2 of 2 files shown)


yamt:
extracted from https://github.com/bytecodealliance/wasm-micro-runtime/pull/2516

core/iwasm/common/wasm_application.c
```diff
@@ -107,7 +107,22 @@ execute_main(WASMModuleInstanceCommon *module_inst, int32 argc, char *argv[])
...
     if ((func = wasm_runtime_lookup_wasi_start_function(module_inst))) {
-        return wasm_runtime_call_wasm(exec_env, func, 0, NULL);
+        const char *wasi_proc_exit_exception = "wasi proc exit";
+
+        ret = wasm_runtime_call_wasm(exec_env, func, 0, NULL);
+        /* a successful return from _start is same as proc_exit(0). */
+        if (ret) {
+            wasm_runtime_set_exception(module_inst, wasi_proc_exit_exception);
+            /* exit_code is zero-initialized */
+            ret = false;
```
wenyongh, line 117: Not sure why need to set `wasi_proc_exit_exception` here? If ret is true, there should be no exception thrown and wasi_exit_code was already initialized as 0, seems we can do nothing?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -2005,15 +1978,10 @@ wasm_runtime_call_wasm(WASMExecEnv *exec_env,
...
-        if (clear_wasi_proc_exit_exception(exec_env->module_inst)) {
-            ret = true;
-        }
-        else {
-            if (new_argv != argv) {
-                wasm_runtime_free(new_argv);
-            }
-            return false;
+        if (new_argv != argv) {
+            wasm_runtime_free(new_argv);
```
wenyongh, line 1982: `clear_wasi_proc_exit_exception` is only applied after calling wasi `_start` function now, is it a general handling of wasi? Will developer call other functions in wasi mode and then `wasm_proc_exit` occurs?

core/iwasm/common/wasm_application.c
```diff
@@ -107,7 +107,22 @@ execute_main(WASMModuleInstanceCommon *module_inst, int32 argc, char *argv[])
...
+        }
+        /* report wasm proc exit as a success */
+        WASMModuleInstance *inst = (WASMModuleInstance *)module_inst;
+        if (!ret && strstr(inst->cur_exception, wasi_proc_exit_exception)) {
```
lum1n0us, line 121: It seems both L117 and L115 make `if` in L121 always true.

core/iwasm/common/wasm_runtime_common.c
yamt: i guess it's a gray area.
proc_exit is supposed to terminate the "process".
the concept of "process" only makes sense in the context of a wasi command.

core/iwasm/common/wasm_application.c
wenyongh: But the main thread will wait until all other threads exit in wasm_exec_env_destroy, it doesn't meet the "wasi proc exit" exception, seems no need to propagate "wasi proc exit" for other threads to exit early? Or other threads may not finish their jobs normally, also other threads may fail to throw exception if they continue to run?

wenyongh: That's what I my concern is, suppose the sub thread is doing an important job, e.g. do some calculation, output important data to a file, can we terminate it from the main thread? Is it better to wait until the sub threads exit? In fact, there is API wasm_cluster_wait_for_all_except_self in thread manager.

Not sure whether there is reference about "a return from _start should behave as exit(0)" in wasi-libc?

@loganek What's your opinion?

core/iwasm/common/wasm_runtime_common.c
wenyongh: Sounds reasonable, maybe we can apply the changes first, and modify if needed in the future.

core/iwasm/common/wasm_application.c
yamt: i'm not sure if i understand your concern.
the importance of the job is something apps should consider. not runtime.

> Not sure whether there is reference about "a return from _start should behave as exit(0)" in wasi-libc?

* a return from main() is an equivalent of exit(). (C standard)
* wasi-libc maps a return from main() to a return from `_start` when the exit code is 0. https://github.com/WebAssembly/wasi-libc/blob/ec4566beae84e54952637f0bf61bee4b4cacc087/libc-bottom-half/crt/crt1-command.c#L48-L52

core/iwasm/common/wasm_application.c
lum1n0us: L115 ~ L121 is considered as be interpreted by other threads?  and `cur_exception` is a racing data?

core/iwasm/common/wasm_application.c
wenyongh: I am not very sure whether it is a good behavior that the main thread terminates all others thread instead of waiting for them. But if you want to do that, can we just call the API `wasm_cluster_terminate_all_except_self` of thread manager:
```C
    if (ret) {
        wasm_cluster_terminate_all_except_self(cluster, exec_env);
    }
#endif
```
It also sets the terminate flag for other threads.

core/iwasm/common/wasm_application.c
```diff
@@ -107,7 +107,22 @@ execute_main(WASMModuleInstanceCommon *module_inst, int32 argc, char *argv[])
        the actual main function. Directly calling main function
        may cause exception thrown. */
     if ((func = wasm_runtime_lookup_wasi_start_function(module_inst))) {
-        return wasm_runtime_call_wasm(exec_env, func, 0, NULL);
+        const char *wasi_proc_exit_exception = "wasi proc exit";
```
lum1n0us, line 110: Is it better if we use a global variable or macro to represent the string "wasi proc exit"？ It is a very import constant string now。

core/iwasm/common/wasm_application.c
yamt: i chose to use wasm_runtime_set_exception because:

* what i want to do here is to mimic wasi proc_exit, which uses wasm_runtime_set_exception.
* wasm_cluster_terminate_all_except_self seems unused (thus untested) and broken. (cluster->lock vs cluster_list_lock locking order issue.)
* we don't need to join threads here

core/iwasm/common/wasm_application.c
yamt: ideally we should stop using strings i guess.

core/iwasm/common/wasm_application.c
wenyongh: OK, if to use wasm_runtime_set_exception, could you please add comments to mention somewhat like we want to mimic wasi proc_exit and to terminate other threads? It is not very easy to understand the code :)
And could you wrap the code with `#if WASM_ENABLE_THREAD_MGR != 0 .. #endif`, it is only for multi-threading?

yamt: > OK, if to use wasm_runtime_set_exception, could you please add comments to mention somewhat like we want to mimic wasi proc_exit and to terminate other threads? It is not very easy to understand the code :) And could you wrap the code with `#if WASM_ENABLE_THREAD_MGR != 0 .. #endif`, it is only for multi-threading?

i prefer less #ifdef in general. but ok. done.

lum1n0us:
LGTM

wenyongh (commented):
LGTM

eloparco:
I know we already had this conversation with @yamt, but with these changes,  should we update [trap_after_main_thread_finishes.c](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/libraries/lib-wasi-threads/test/trap_after_main_thread_finishes.c) to have 0 as exit code and to replace the trap [here](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/libraries/lib-wasi-threads/test/trap_after_main_thread_finishes.c#L31) with an assert? And even change the name of the tests and put some comments to describe the expected behavior. I could do it as a separate PR if it makes sense.

Or we drop the test completely since it has some overlapping with https://github.com/WebAssembly/wasi-threads/blob/main/test/testsuite/wasi_threads_exit_main_block.wat.

yamt:
> I know we already had this conversation with @yamt, but with these changes, should we update [trap_after_main_thread_finishes.c](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/libraries/lib-wasi-threads/test/trap_after_main_thread_finishes.c) to have 0 as exit code and to replace the trap [here](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/libraries/lib-wasi-threads/test/trap_after_main_thread_finishes.c#L31) with an assert? And even change the name of the tests and put some comments to describe the expected behavior. I could do it as a separate PR if it makes sense.

i'm not sure if i understand what updates you are suggesting.
it already has 0 as exit code.
the trap and assert are not that different as they are both wasm unreachable.
 
> Or we drop the test completely since it has some overlapping with https://github.com/WebAssembly/wasi-threads/blob/main/test/testsuite/wasi_threads_exit_main_block.wat.

it's also fine for me.

wenyongh:
@eloparco I understand that the main thread already returns 0 and no need to change it. Changing buildin_trap to assert is that buildin_trap will throw exception in spawned thread, while assert (assume we use assert(true statement)) won't throw exception and the spawned thread exits normally.

I think it is a good case and had better not change it. A possible issue is that after main function returns and runtime set_exception("wasi proc exit") to the spawned thread, and then before the spawned thread handles the exception and terminate flag, it runs into the builtin_trap, throws unreachable exception and ends first. That means the main thread returns 0 while sub thread throws exception (and it may spread to the main thread then), not sure whether we should handle it? @yamt

yamt:
> A possible issue is that after main function returns and runtime set_exception("wasi proc exit") to the spawned thread, and then before the spawned thread handles the exception and terminate flag, it runs into the builtin_trap, throws unreachable exception and ends first. That means the main thread returns 0 while sub thread throws exception (and it may spread to the main thread then), not sure whether we should handle it? @yamt

wrt racing multiple traps/exit, i plan to prevent wasm_runtime_set_exception from overwriting existing trap/exit. (not within this PR though)

wenyongh:
> > A possible issue is that after main function returns and runtime set_exception("wasi proc exit") to the spawned thread, and then before the spawned thread handles the exception and terminate flag, it runs into the builtin_trap, throws unreachable exception and ends first. That means the main thread returns 0 while sub thread throws exception (and it may spread to the main thread then), not sure whether we should handle it? @yamt
> 
> wrt racing multiple traps/exit, i plan to prevent wasm_runtime_set_exception from overwriting existing trap/exit. (not within this PR though)

Got it, great, thanks!

PR #2532, opened 2023-09-05
Fix typo for IP address buffer

cngzhnp:
There is a typo for `bh_ip_addr_buffer_t` struct in buffer element. Small cosmetic fix.


core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -3560,7 +3560,8 @@ argv_environ_init(struct argv_environ_values *argv_environ, char *argv_buf,
 
 void
 argv_environ_destroy(struct argv_environ_values *argv_environ)
-{}
+{
```
wenyongh, line 3563: Should not change here, or the CI may report code guideline check error:
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/6088777667/job/16520150912

lum1n0us:
LGTM

PR #2537, opened 2023-09-08
posix os_usleep: fix an integer overflow

yamt:


lum1n0us:
IIUC, `usec % 1000000` is always less than 1,000,000 and `* 1000` will let the max value be 1,000,000,000, which is less than 4,294,967,295.

yamt:
> IIUC, `usec % 1000000` is always less than 1,000,000 and `* 1000` will let the max value be 1,000,000,000, which is less than 4,294,967,295.

hmm. i guess my calculation was wrong. thank you.

Issue #2512, opened 2023-08-29
module instance has too much execution state


yamt:
my impression is that exceptions and stack frames logically belong to wasm_exec_env_t.

it's ok for our current threading model, where module instance and exec env are basically 1:1, though.

PR #2538, opened 2023-09-08
Add an API to terminate instance

yamt:
Note: this doesn't work for some situations like blocking system calls.
Most of them will be fixed (at least for posix-like platforms) by https://github.com/bytecodealliance/wasm-micro-runtime/pull/2516

### DIFF
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -2342,8 +2342,8 @@ wasm_runtime_get_exec_env_singleton(WASMModuleInstanceCommon *module_inst_comm)
     return module_inst->exec_env_singleton;
 }
 
-void
-wasm_set_exception(WASMModuleInstance *module_inst, const char *exception)
+static void
+wasm_set_exception_local(WASMModuleInstance *module_inst, const char *exception)
 {
     exception_lock(module_inst);
     if (exception) {
@@ -2354,13 +2354,22 @@ wasm_set_exception(WASMModuleInstance *module_inst, const char *exception)
         module_inst->cur_exception[0] = '\0';
     }
     exception_unlock(module_inst);
+}
 
+void
+wasm_set_exception(WASMModuleInstance *module_inst, const char *exception)
+{
 #if WASM_ENABLE_THREAD_MGR != 0
     WASMExecEnv *exec_env =
         wasm_clusters_search_exec_env((WASMModuleInstanceCommon *)module_inst);
     if (exec_env) {
-        wasm_cluster_spread_exception(exec_env, exception);
+        wasm_cluster_set_exception(exec_env, exception);
     }
+    else {
+        wasm_set_exception_local(module_inst, exception);
+    }
+#else
+    wasm_set_exception_local(module_inst, exception);
 #endif
 }
 
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -2477,6 +2477,18 @@ wasm_runtime_clear_exception(WASMModuleInstanceCommon *module_inst_comm)
     wasm_runtime_set_exception(module_inst_comm, NULL);
 }
 
+#if WASM_ENABLE_THREAD_MGR != 0
+void
+wasm_runtime_terminate(WASMModuleInstanceCommon *module_inst_comm)
+{
+    WASMModuleInstance *module_inst = (WASMModuleInstance *)module_inst_comm;
+
+    bh_assert(module_inst_comm->module_type == Wasm_Module_Bytecode
+              || module_inst_comm->module_type == Wasm_Module_AoT);
+    wasm_set_exception(module_inst, "terminated by user");
+}
+#endif
+
 void
 wasm_runtime_set_custom_data_internal(
     WASMModuleInstanceCommon *module_inst_comm, void *custom_data)
--- a/core/iwasm/libraries/thread-mgr/thread_manager.c
+++ b/core/iwasm/libraries/thread-mgr/thread_manager.c
@@ -1270,14 +1270,14 @@ set_exception_visitor(void *node, void *user_data)
 }
 
 void
-wasm_cluster_spread_exception(WASMExecEnv *exec_env, const char *exception)
+wasm_cluster_set_exception(WASMExecEnv *exec_env, const char *exception)
 {
     const bool has_exception = exception != NULL;
     WASMCluster *cluster = wasm_exec_env_get_cluster(exec_env);
     bh_assert(cluster);
 
     struct spread_exception_data data;
-    data.skip = exec_env;
+    data.skip = NULL;
     data.exception = exception;
 
     os_mutex_lock(&cluster->lock);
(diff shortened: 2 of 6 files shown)


core/iwasm/common/wasm_runtime_common.c
```diff
@@ -2342,8 +2342,8 @@ wasm_runtime_get_exec_env_singleton(WASMModuleInstanceCommon *module_inst_comm)
     return module_inst->exec_env_singleton;
 }
 
-void
-wasm_set_exception(WASMModuleInstance *module_inst, const char *exception)
+static void
+wasm_set_exception_local(WASMModuleInstance *module_inst, const char *exception)
```
wenyongh, line 2346: Can we remove the lock in this function?

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -1266,18 +1275,21 @@ set_exception_visitor(void *node, void *user_data)
...
-wasm_cluster_spread_exception(WASMExecEnv *exec_env, const char *exception)
+wasm_cluster_set_exception(WASMExecEnv *exec_env, const char *exception)
 {
     const bool has_exception = exception != NULL;
     WASMCluster *cluster = wasm_exec_env_get_cluster(exec_env);
     bh_assert(cluster);
 
     struct spread_exception_data data;
-    data.skip = exec_env;
+    data.skip = NULL;
```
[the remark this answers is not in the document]
yamt: yes, it can be simplified if we want.

product-mini/platforms/posix/main.c
```diff
@@ -97,6 +97,9 @@ print_help()
 #if WASM_ENABLE_LIB_PTHREAD != 0 || WASM_ENABLE_LIB_WASI_THREADS != 0
     printf("  --max-threads=n          Set maximum thread number per cluster, default is 4\n");
 #endif
+#if WASM_ENABLE_THREAD_MGR != 0
+    printf("  --timeout=ms             Set the timeout in ms.\n");
```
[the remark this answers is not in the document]
yamt: ok

core/iwasm/common/wasm_runtime_common.c
yamt: exception_lock() is already no-op for builds w/o threading support.

product-mini/platforms/posix/main.c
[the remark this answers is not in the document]
yamt: done

core/iwasm/common/wasm_runtime_common.c
wenyongh: When thread manager is enabled, wasm_set_exception_local is called only when exec_env isn't in the cluster, seems no lock is needed. But exception_lock() uses lock when thread manager is enabled.

wenyongh (commented):
LGTM

core/iwasm/common/wasm_runtime_common.c
yamt: well, can cluster be NULL here?
even if it's possible, it's exceptional, isn't it?

wenyongh: The cluster may be NULL if `wasm_exec_env_create` hasn't been called with the module_inst, or the exec_env was destroyed by `wasm_exec_env_destroy` and the cluster was destroyed.

Yes, it is exceptional, I just think the `_local` suffix means we can only care about the current module_inst, no need to care about multi-threading. It is also OK if you prefer to keep the exception_lock/unlock.

yamt: > The cluster may be NULL if `wasm_exec_env_create` hasn't been called with the module_inst, or the exec_env was destroyed by `wasm_exec_env_destroy` and the cluster was destroyed.

ok.
i feel it's awkward to use a trap w/o exec env though. cf. https://github.com/bytecodealliance/wasm-micro-runtime/issues/2512

> Yes, it is exceptional, I just think the `_local` suffix means we can only care about the current module_inst, no need to care about multi-threading. It is also OK if you prefer to keep the exception_lock/unlock.

i prefer to keep exception_lock here.
as this is not performance critical at all, consistency outweighs IMO.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -2468,6 +2477,18 @@ wasm_runtime_clear_exception(WASMModuleInstanceCommon *module_inst_comm)
...
 
+#if WASM_ENABLE_THREAD_MGR != 0
+void
+wasm_runtime_terminate(WASMModuleInstanceCommon *module_inst_comm)
+{
+    WASMModuleInstance *module_inst = (WASMModuleInstance *)module_inst_comm;
+
+    bh_assert(module_inst_comm->module_type == Wasm_Module_Bytecode
+              || module_inst_comm->module_type == Wasm_Module_AoT);
+    wasm_set_exception(module_inst, "terminated by user");
```
xujuntwt95329, line 2488: Seems our purpose is to terminate all threads, how about call `wasm_cluster_terminate_all` directly rather than setting an exception?

yamt: wasm_cluster_terminate_all seems unused/untested and broken. (eg. cluster_list_lock vs cluster->lock locking order)
IMO it should be removed.

wenyongh: @yamt We are implementing the thread suspension/resume mechanism in PR #2320, 
in which when a thread wants to suspend/terminate other threads, it will lock the whole thread list firstly (and also get the acknowledge of other threads, other threads will run into waiting status). Shall we change to call wasm_cluster_terminate_all after the PR is implemented and tested?

yamt: > @yamt We are implementing the thread suspension/resume mechanism in PR #2320, in which when a thread wants to suspend/terminate other threads, it will lock the whole thread list firstly (and also get the acknowledge of other threads, other threads will run into waiting status). Shall we change to call wasm_cluster_terminate_all after the PR is implemented and tested?

how it's implemented is not critical for me as far as it works.
if you want to replace it with something else later, it's fine.

however, i want to provide a way for the terminated thread (the embedder app calling one of our execute APIs) to notice the fact it was terminated by `wasm_runtime_terminate`.
a nice property of a trap here is that the existing embedders are likely checking it already.
i'm not sure how `wasm_cluster_terminate_all` can work in that regard.

also, the current `wasm_cluster_terminate_all` seems to join all the threads by itself.
i'm not sure how it's supposed to work for the usage like iwasm, where wasm module is executed by the main thread.

wenyongh: Yes, a trap may be a good option for the embedders to check, maybe we can add a flag to indicate whether to throw the exception or not in wasm_runtime_terminate and wasm_cluster_terminate_all.

The current wasm_cluster_terminate_all joins all the threads which may include the main thread that initializes the runtime (if embedder calls the API in a another thread created by himself), it means that wasm_cluster_terminate_all waits until the main thread exits (and runtime is destroyed). This is not what embedder expects, I think we can change to wait until the cluster is destroyed in the cluster set. A potential issue of this PR is that wasm_cluster_terminate_all doesn't wait until all threads end their jobs, I am not sure whether it will cause unexpected behavior, it should be better if wasm_cluster_terminate_all can figure it out.

PR #2546, opened 2023-09-13
Add context to enlarge memory error callback

eloparco:
Follow-up on https://github.com/bytecodealliance/wasm-micro-runtime/pull/2522.
When trying to use the callback in practice (from an embedder), I realized that not being able to pass some context (i.e. data passed from the embedder that can be used in the callback) is quite limiting.

### DIFF
--- a/core/iwasm/include/wasm_export.h
+++ b/core/iwasm/include/wasm_export.h
@@ -1447,14 +1447,15 @@ typedef enum {
 typedef void (*enlarge_memory_error_callback_t)(
     uint32_t inc_page_count, uint64_t current_memory_size,
     uint32_t memory_index, enlarge_memory_error_reason_t failure_reason,
-    wasm_module_inst_t instance, wasm_exec_env_t exec_env);
+    wasm_module_inst_t instance, wasm_exec_env_t exec_env,
+    void* user_data);
 
 /**
  * Setup callback invoked when memory.grow fails
  */
 WASM_RUNTIME_API_EXTERN void
 wasm_runtime_set_enlarge_mem_error_callback(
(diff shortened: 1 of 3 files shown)


core/iwasm/include/wasm_export.h
```diff
@@ -1447,14 +1447,15 @@ typedef enum {
...
+    wasm_module_inst_t instance, wasm_exec_env_t exec_env,
+    void* context);
 
 /**
  * Setup callback invoked when memory.grow fails
  */
 WASM_RUNTIME_API_EXTERN void
 wasm_runtime_set_enlarge_mem_error_callback(
-    const enlarge_memory_error_callback_t callback);
+    const enlarge_memory_error_callback_t callback, void *context);
```
wenyongh, line 1458: @eloparco How about changing the argument name to `user_data`? I remember we usually called data like this with `user_data`.

core/iwasm/include/wasm_export.h
```diff
@@ -1447,14 +1447,15 @@ typedef enum {
 typedef void (*enlarge_memory_error_callback_t)(
     uint32_t inc_page_count, uint64_t current_memory_size,
     uint32_t memory_index, enlarge_memory_error_reason_t failure_reason,
-    wasm_module_inst_t instance, wasm_exec_env_t exec_env);
+    wasm_module_inst_t instance, wasm_exec_env_t exec_env,
+    void* context);
```
wenyongh, line 1451: Had better `void *user_data`?

eloparco: Sorry I missed that one, fixed

PR #2549, opened 2023-09-14
In summary, show accurate case amount

lum1n0us:
Instead of the total case amount under *spec/test/core*, use a accurate number of how many test cases are going to be executed.


wenyongh (approved):
LGTM

PR #2550, opened 2023-09-14
To support AOT multi-module feature

lum1n0us:


tests/wamr-test-suites/test_wamr.sh
```diff
@@ -423,9 +423,7 @@ function spec_test()
 
     # multi-module only enable in interp mode
     if [[ 1 == ${ENABLE_MULTI_MODULE} ]]; then
-        if [[ $1 == 'classic-interp' || $1 == 'fast-interp' ]]; then
-            ARGS_FOR_SPEC_TEST+="-M "
-        fi
+        ARGS_FOR_SPEC_TEST+="-M "
```
wenyongh, line 426: Had better change the condition, fast-jit doesn't support multi-module yet.

tests/wamr-test-suites/test_wamr.sh
```diff
@@ -328,7 +329,14 @@ function spec_test()
     git checkout -B main
     # [spec] Update note on module initialization trapping (#1493)
     git reset --hard 044d0d2e77bdcbe891f7e0b9dd2ac01d56435f0b
-    git apply ../../spec-test-script/ignore_cases.patch
+
+    # aot multi-module is different with non-aot multi-module
+    if [[ ${ENABLE_MULTI_MODULE} == 1 &&  "${RUNNING_MODE}" == "aot" ]]; then
+      git apply ../../spec-test-script/ignore_cases_aot.patch
```
wenyongh, line 335: How about renaming it to `ignore_cases_aot_multi_module.patch` to make it clear?

tests/wamr-test-suites/test_wamr.sh
```diff
@@ -423,7 +431,7 @@ function spec_test()
 
     # multi-module only enable in interp mode
     if [[ 1 == ${ENABLE_MULTI_MODULE} ]]; then
-        if [[ $1 == 'classic-interp' || $1 == 'fast-interp' ]]; then
+        if [[ $1 != 'fast-jit' ]]; then
```
wenyongh, line 434: Seems we should add `multi-tier-jit` also, which doesn't support multi-module:
`if [[ $1 != 'fast-jit' || $1 != 'multi-tier-jit' ]]; then`

tests/wamr-test-suites/test_wamr.sh
lum1n0us: ✅ changed the file name as you recommended

tests/wamr-test-suites/test_wamr.sh
lum1n0us: ✅ changed the condition based on your advice

PR #2551, opened 2023-09-14
Allow custom signal handler from non-main thread

eloparco:
When hardware bound check is enabled and a custom signal handler is defined, it is stored in a thread-local variable.
If the embedder sets up the WAMR runtime in the main thread, spawns a thread and the signal handler is called from the spawn thread, the custom signal handler is not available.

### DIFF
--- a/core/shared/platform/common/posix/posix_thread.c
+++ b/core/shared/platform/common/posix/posix_thread.c
@@ -509,8 +509,8 @@ mask_signals(int how)
     pthread_sigmask(how, &set, NULL);
 }
 
-static os_thread_local_attribute struct sigaction prev_sig_act_SIGSEGV;
-static os_thread_local_attribute struct sigaction prev_sig_act_SIGBUS;
+static struct sigaction prev_sig_act_SIGSEGV;
+static struct sigaction prev_sig_act_SIGBUS;
 
 /* ASAN is not designed to work with custom stack unwind or other low-level \
  things. > Ignore a function that does some low-level magic. (e.g. walking \
@@ -540,10 +540,15 @@ signal_callback(int sig_num, siginfo_t *sig_info, void *sig_ucontext)
     if (prev_sig_act && (prev_sig_act->sa_flags & SA_SIGINFO)) {
         prev_sig_act->sa_sigaction(sig_num, sig_info, sig_ucontext);
     }
-    else if (prev_sig_act
-             && ((void *)prev_sig_act->sa_sigaction == SIG_DFL
-                 || (void *)prev_sig_act->sa_sigaction == SIG_IGN)) {
+    else if (prev_sig_act && (void *)prev_sig_act->sa_handler == SIG_DFL) {
+        /* Default action */
         sigaction(sig_num, prev_sig_act, NULL);
+    }
+    else if (prev_sig_act && (void *)prev_sig_act->sa_handler == SIG_IGN) {
+        /* Ignore this signal */
+    }
+    else if (prev_sig_act && prev_sig_act->sa_handler) {
+        prev_sig_act->sa_handler(sig_num);
     }
     /* Output signal info and then crash if signal is unhandled */
     else {


eloparco:
@wenyongh Is there any reason for the second condition in the `if`? https://github.com/bytecodealliance/wasm-micro-runtime/blob/83db970953c460738701f0edbf7eb10666e59452/core/shared/platform/common/posix/posix_thread.c#L540

It forces my custom signal handler to be initialized with `SA_SIGINFO` otherwise it goes into https://github.com/bytecodealliance/wasm-micro-runtime/blob/83db970953c460738701f0edbf7eb10666e59452/core/shared/platform/common/posix/posix_thread.c#L550

eloparco:
@wenyongh Is there any reason for the second condition in the `if`? https://github.com/bytecodealliance/wasm-micro-runtime/blob/83db970953c460738701f0edbf7eb10666e59452/core/shared/platform/common/posix/posix_thread.c#L540

It forces my custom signal handler to be initialized with `SA_SIGINFO` otherwise it goes into https://github.com/bytecodealliance/wasm-micro-runtime/blob/83db970953c460738701f0edbf7eb10666e59452/core/shared/platform/common/posix/posix_thread.c#L550

wenyongh:
> When hardware bound check is enabled and a custom signal handler is defined, it is stored in a thread-local variable. If the embedder sets up the WAMR runtime in the main thread, spawns a thread and the signal handler is called from the spawn thread, the custom signal handler is not available.

@eloparco, do you use wasm_runtime_spawn_thread to create the thread? And do you call wasm_runtime_init_thread_env() in at the beginning of the callback of thread spawned? Normally for the thread created by embedder but not runtime, we should call the function to initialize the thread environment, including prev_sig_act_SIGSEGV. Note that eventually os_thread_signal_init is be called.

wenyongh:
> @wenyongh Is there any reason for the second condition in the `if`?
> 
> https://github.com/bytecodealliance/wasm-micro-runtime/blob/83db970953c460738701f0edbf7eb10666e59452/core/shared/platform/common/posix/posix_thread.c#L540
> 

Hi, it was implemented according to the reference document:
```
If  SA_SIGINFO  is  specified in sa_flags, then sa_sigaction (instead of sa_handler) specifies the signal-handling function for signum.
```

But here it ignored the situation that prev_sig_act->sa_handler is not NULL. The document also mentions:
```
sa_handler specifies the action to be associated with signum and is be one of the following:

       * SIG_DFL for the default action.

       * SIG_IGN to ignore this signal.

       * A pointer to a signal handling function.  This function receives the signal number as its only argument.
```

The code may have some issues, per my understanding, the correct handling may be like:
```C
    /* Forward the signal to next handler if found */
    if (prev_sig_act && (prev_sig_act->sa_flags & SA_SIGINFO)) {
        prev_sig_act->sa_sigaction(sig_num, sig_info, sig_ucontext);
    }
    else if (prev_sig_act
             && (void *)prev_sig_act->sa_handler == SIG_DFL) {
        /* Default action */
        sigaction(sig_num, prev_sig_act, NULL);
    }
    else if (prev_sig_act
             && (void *)prev_sig_act->sa_handler== SIG_IGN) {
        /* Ignore this signal */
    }
    else if (prev_sig_act && prev_sig_act->sa_handler) {
        prev_sig_act->sa_handler(sig_num);
    }
    /* Output signal info and then crash if signal is unhandled */
    else {
        switch (sig_num) {
            ...
        }
    }
```

Could you help confirm and have a try? Thanks.

core/shared/platform/common/posix/posix_thread.c
```diff
@@ -509,8 +509,8 @@ mask_signals(int how)
     pthread_sigmask(how, &set, NULL);
 }
 
-static os_thread_local_attribute struct sigaction prev_sig_act_SIGSEGV;
-static os_thread_local_attribute struct sigaction prev_sig_act_SIGBUS;
+static struct sigaction prev_sig_act_SIGSEGV;
```
wenyongh, line 512: Could you try keeping the os_thread_local_attribute flag and call `wasm_runtime_init_thread_env` in the callback of the spawned thread? In that way prev_sig_act_SIGSEGV of that thread will be set.

Per my understanding, each thread can has its sig action, we had better keep the os_thread_local_attribute flag.

eloparco:
> @eloparco, do you use wasm_runtime_spawn_thread to create the thread? And do you call wasm_runtime_init_thread_env() in at the beginning of the callback of thread spawned? Normally for the thread created by embedder but not runtime, we should call the function to initialize the thread environment, including prev_sig_act_SIGSEGV. Note that eventually os_thread_signal_init is be called.

In my use case, the thread creation is done in a separate part of the code that is independent of the runtime so I can't use WAMR functions there.

> Could you help confirm and have a try? Thanks.

Yes, just tried that and it works for me. With your suggestion, if I don't set `SA_SIGINFO` in my custom signal handler, it goes into the 
`if (prev_sig_act && prev_sig_act->sa_handler) {` part and works properly by calling my custom signal handler.

core/shared/platform/common/posix/posix_thread.c
eloparco: My understanding is that the sigaction is per process, not per thread:
> The sigaction() system call is used to change the action taken by a process on receipt of a specific signal. 

(source: https://man7.org/linux/man-pages/man2/sigaction.2.html)

wenyongh: This is the answer I got from ChapGPT:
image

Not very sure, but if keeping os_thread_local_attribute flag works for you, had better keep it.

eloparco: I don't think that's very reliable.
>The signal disposition is a per-process attribute: in a
       multithreaded application, the disposition of a particular signal
       is the same for all threads.

(https://man7.org/linux/man-pages/man7/signal.7.html)
From what I read, signal handers are per process, signal masks are per thread.

If I keep `os_thread_local_attribute`, it doesn't work. And I can't use `wasm_runtime_init_thread_env` in that thread since it doesn't import wamr dependencies. 
I don't think that's any harm in having the prev sigaction global instead of thread-local.

wenyongh: OK, let's merge it.

yamt: @wenyongh do you remember why you added sigaction call here?

yamt: well, it seems that i replied to a wrong thread.
my question is about the sigaction call in signal_callback.

wenyongh: @yamt Someone raised issue that previous signal handlers were supressed: https://github.com/bytecodealliance/wasm-micro-runtime/issues/1262
And I submitted PR https://github.com/bytecodealliance/wasm-micro-runtime/pull/1268 to fix it.

yamt: > @yamt Someone raised issue that previous signal handlers were supressed: #1262 And I submitted PR #1268 to fix it.

i can understand the intention to call prev_sig_act->sa_sigaction.
but i don't understand why to call sigaction.

wenyongh: Do you mean the handling of SIG_DFL? I am a little confused also, I referred to the implementation of wasmtime:
https://github.com/bytecodealliance/wasmtime/blob/main/crates/runtime/src/traphandlers/unix.rs#L169-L170:
```rust
    else if previous.sa_sigaction == libc::SIG_DFL || previous.sa_sigaction == libc::SIG_IGN {
        libc::sigaction(signum, previous, ptr::null_mut());
    }
```
But seems that its implementation is invalid: should use `previous.sa_handler` instead of `previous.sa_sigaction`, and should handle SIG_DFL and SIG_IGN in different ways? Also not sure whether using libc::sigaction is right.

In IBM's document: https://www.ibm.com/docs/en/zos/2.4.0?topic=functions-sigaction-examine-change-signal-action
```C
void (*)(int, siginfo_t *, void *) sa_sigaction
A pointer to the function assigned to handle the signal, or SIG_DFL, or SIG_IGN. 
```
And it mentions:
```C
Notes:
The user must cast SIG_IGN or SIG_DFL to match the sa_sigaction definition. (indicating that the signal is to be ignored).
```

yamt: the intention of the wasmtime code seems like restore the signal handler and then make it re-fault.
i'm not sure it it works reliably.
even if it does, it leaves the signal handler modified, which seems broken for our purpose.
maybe it might not be a problem for wasmtime, where "multi-tenancy" is somehow optional.

wenyongh: So any suggestion to handle SIG_DFL here? Or keep the code unchanged?

yamt: my suggestion is to remove the SIG_DFL (and probably SIG_IGN) cases because apps can hardly rely on them.

eloparco: Not sure about those cases to be honest, personally I only tested the `prev_sig_act->sa_sigaction` and `prev_sig_act->sa_handler` cases.

wenyongh: OK, so let's just handle the two cases, I will submit another PR to fix it.

PR #2567, opened 2023-09-19
core/iwasm/compilation: Support AOT with LLVM 17

fredldotme:
Adapt API usage to new interfaces where applicable, including LLVM function usage, obsoleted llvm::Optional type & removal of unavailable headers.

### DIFF
--- a/core/iwasm/compilation/aot_llvm.c
+++ b/core/iwasm/compilation/aot_llvm.c
@@ -526,12 +526,18 @@ aot_add_precheck_function(AOTCompContext *comp_ctx, LLVMModuleRef module,
     }
     wasm_runtime_free(params);
     params = NULL;
+
+#if LLVM_VERSION_MAJOR < 17
     if (aot_target_precheck_can_use_musttail(comp_ctx)) {
         LLVMSetTailCallKind(retval, LLVMTailCallKindMustTail);
     }
     else {
         LLVMSetTailCallKind(retval, LLVMTailCallKindTail);
     }
+#else
+    LLVMSetTailCall(retval, true);
+#endif
+
     if (ret_type == VOID_TYPE) {
         if (!LLVMBuildRetVoid(b)) {
             goto fail;
@@ -2172,8 +2178,10 @@ bool
 aot_compiler_init(void)
 {
     /* Initialize LLVM environment */
-
+#if LLVM_VERSION_MAJOR < 17
     LLVMInitializeCore(LLVMGetGlobalPassRegistry());
+#endif
+
 #if WASM_ENABLE_WAMR_COMPILER != 0
     /* Init environment of all targets for AOT compiler */
     LLVMInitializeAllTargetInfos();
--- a/core/iwasm/compilation/aot_llvm.h
+++ b/core/iwasm/compilation/aot_llvm.h
@@ -15,15 +15,18 @@
 #include "llvm-c/ExecutionEngine.h"
 #include "llvm-c/Analysis.h"
 #include "llvm-c/BitWriter.h"
+#if LLVM_VERSION_MAJOR < 17
 #include "llvm-c/Transforms/Utils.h"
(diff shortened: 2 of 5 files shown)


lum1n0us (approved):
LGTM

core/iwasm/compilation/aot_llvm.c
```diff
@@ -2172,8 +2178,10 @@ bool
 aot_compiler_init(void)
 {
     /* Initialize LLVM environment */
-
+#if LLVM_VERSION_MAJOR < 17
     LLVMInitializeCore(LLVMGetGlobalPassRegistry());
+#endif
```
wenyongh, line 2183: Hi, I tested the JIT mode with LLVM 17:
```bash
cd product-mini/platforms/linux
mkdir build && cd build
cmake .. -DWAMR_BUILD_JIT=1 -DCMAKE_BUILD_TYPE=Debug
make -j
```
And ran a simple case, segmentation fault occured. Debugged with gdb, it crashed at LLVMJIT.h line 72:
```bash
Program received signal SIGSEGV, Segmentation fault.
0x000055555575ed92 in llvm::orc::LLJIT::getMainJITDylib (this=0x0)
    at core/deps/llvm/build/include/llvm/ExecutionEngine/Orc/LLJIT.h:72
72        JITDylib &getMainJITDylib() { return *Main; }
```
Not sure whether it is due to that this line is disabled. Could you help check?

fredldotme: Might be possible indeed, will check later. The reason everything works here could have to do with everything being statically linked (that includes clang, lldb and wasm-ld, if you want your eyes to fall out check my llvm fork) and LLVM already being initialized properly.

Will check later today.

fredldotme: Okay, so I tried it on macOS now and cannot reproduce.

The CMake command: `cmake .. -DWAMR_BUILD_EXCE_HANDLING=0 -DWAMR_BUILD_TAIL_CALL=1 -DWAMR_BUILD_MULTI_MODULE=1 -DWAMR_BUILD_FAST_INTERP=0 -DWAMR_BUILD_DEBUG_INTERP=1 -DWAMR_BUILD_AOT=0 -DCMAKE_BUILD_TYPE=Debug -DWAMR_DISABLE_HW_BOUND_CHECK=1 -DWAMR_BUILD_JIT=1 -DLLVM_DIR=/Users/alfredneumayer/Projects/open/Tide/3rdparty/llvm/build_osx/lib/cmake/llvm -DWAMR_BUILD_PLATFORM=darwin -DWAMR_BUILD_LIB_WASI_THREADS=1`

Here is the result:

```
alfredneumayer@MacBook-Air-von-Alfred build % ./iwasm --llvm-jit /Users/alfredneumayer/Downloads/TideTestApp\ 3 
Luke Treewalker v0.3!
warning: Seeding randomness
Node depth: 5
Value: 1586703438
Going right
 Value: 1894619784
 Going left
  Value: 1803171417
  Going right
   Value: 756510595
   Going left
    Value: 1677205272
```

Mind you, this is not a static build of LLVM but rather a regular Clang build as one would expect.

wenyongh: Thanks, I built LLVM-17 with gcc-9.4.0 under Ubuntu-20.04. I checked again, it isn't caused by commenting this line, but seems to be caused by the link flags, the error `Symbols not found: [ llvm_orc_registerEHFrameSectionWrapper ]` was reported. The function is implemented in "lib/ExecutionEngine/Orc/TargetProcess/RegisterEHFrames.cpp", and the libLLVMOrcTargetProcess.a is added in the link list of iwasm, but it is strange that symbol isn't linked into iwasm.
However, the example HowToUseLLJIT of LLVM works well. I think they must use different link flags, but it need time to check.

Let's merge this PR first and fix the issue in the future.

wenyongh (commented):
LGTM

PR #2570, opened 2023-09-20
core/iwasm/aot: Default expected target to "arm64" on Mach

fredldotme:
This tricks the AOT loader logic into loading AOT-compiled ELF files on iOS when wamrc is configured to build 'arm64-pc-linux-gnu' targets, the default LLVM detects when wamrc is run directly on the device.

### DIFF
--- a/core/iwasm/aot/arch/aot_reloc_aarch64.c
+++ b/core/iwasm/aot/arch/aot_reloc_aarch64.c
@@ -53,7 +53,11 @@ get_target_symbol_map(uint32 *sym_num)
     return target_sym_map;
 }
 
+#if !__MACH__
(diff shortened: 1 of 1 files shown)


core/iwasm/aot/arch/aot_reloc_aarch64.c
```diff
@@ -53,7 +53,11 @@ get_target_symbol_map(uint32 *sym_num)
     return target_sym_map;
 }
 
+#if !__MACH__
```
wenyongh, line 56: How about `!defined(__MACH__)`? `!__MACH__` may cause compilation error on some platforms since `__MACH__` is not defined.

wenyongh:
Close this PR as similar change has been applied in #2618.

PR #2578, opened 2023-09-21
Add support for closing/renumbering preopen fds

zoraaver:
There doesn't appear to be a clear reason not to support this behaviour. It seems it was disallowed previously as a precaution. See https://github.com/bytecodealliance/wasmtime/commit/67e2e57b02cd244c004379690aa56d3d5e187153 for more context.

### DIFF
--- a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
+++ b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
@@ -339,7 +339,7 @@ fd_prestats_init(struct fd_prestats *pt)
 
 // Grows the preopened resource table to a required lower bound and a
 // minimum number of free preopened resource table entries.
-static bool
+static __wasi_errno_t
 fd_prestats_grow(struct fd_prestats *pt, size_t min, size_t incr)
     REQUIRES_EXCLUSIVE(pt->lock)
 {
@@ -353,7 +353,7 @@ fd_prestats_grow(struct fd_prestats *pt, size_t min, size_t incr)
         struct fd_prestat *prestats =
             wasm_runtime_malloc((uint32)(sizeof(*prestats) * size));
         if (prestats == NULL)
-            return false;
+            return __WASI_ENOMEM;
 
         if (pt->prestats && pt->size > 0) {
             bh_memcpy_s(prestats, (uint32)(sizeof(*prestats) * size),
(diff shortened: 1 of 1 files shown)


core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -1122,8 +1141,45 @@ wasmtime_ssp_fd_renumber(wasm_exec_env_t exec_env, struct fd_table *curfds,
...
+    // Renumbering over two preopened resources.
+    if (prestat_from_error == 0 && prestat_to_error == 0) {
+        error = fd_prestats_remove_entry(prestats, to);
+
+        if (error == 0) {
+            error = fd_prestats_insert_locked(prestats, prestat_from->dir, to);
+
+            if (error == 0) {
+                error = fd_prestats_remove_entry(prestats, from);
+            }
```
wenyongh, line 1162: Same as above, should we restore data changed by `fd_prestats_remove_entry` when `error != 0`?

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -1122,8 +1141,45 @@ wasmtime_ssp_fd_renumber(wasm_exec_env_t exec_env, struct fd_table *curfds,
...
+
+    // Renumbering over two preopened resources.
+    if (prestat_from_error == 0 && prestat_to_error == 0) {
+        error = fd_prestats_remove_entry(prestats, to);
+
+        if (error == 0) {
+            error = fd_prestats_insert_locked(prestats, prestat_from->dir, to);
+
+            if (error == 0) {
+                error = fd_prestats_remove_entry(prestats, from);
```
wenyongh, line 1161: Should we handle it when `error != 0`, or failed to `fd_prestats_remove_entry`? e.g. restore data changed by the above `fd_prestats_remove_entry` and `fd_prestats_insert_locked`?

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -1122,8 +1141,45 @@ wasmtime_ssp_fd_renumber(wasm_exec_env_t exec_env, struct fd_table *curfds,
...
+    // can't a keep the destination fd entry in the preopened table so remove
+    // it entirely.
+    else if (prestat_from_error != 0 && prestat_to_error == 0) {
+        error = fd_prestats_remove_entry(prestats, to);
+    }
+    // Renumbering from a preopened fd to a non-preopened fd
+    else if (prestat_from_error == 0 && prestat_to_error != 0) {
+        error = fd_prestats_insert_locked(prestats, prestat_from->dir, to);
+        if (error == 0) {
+            error = fd_prestats_remove_entry(prestats, from);
```
wenyongh, line 1175: Same as above

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
zoraaver: The only reason `fd_prestats_remove_entry` would fail is if the entry didn't exist in the table but we've already checked that condition above. I'll ignore the return value to make the code a bit easier to read.

`fd_prestats_insert_locked` might fail due to insufficient memory so I've added some code to restore the original state via `fd_prestats_remove_entry`.

wenyongh (commented):
LGTM

PR #2579, opened 2023-09-21
Remove provision of unnecessary fd rights

zoraaver:
The WASI docs allow for fewer rights to be applied to an fd than requested but not more. This behaviour is also asserted in the rust WASI tests, so it's necessary for those to pass as well.


loganek (approved):
LGTM, let's make sure this is highlighted in the release notes.

PR #2580, opened 2023-09-21
Return __WASI_EINVAL from fd_prestat_dir_name

zoraaver:
Return a WASI error code (rather than a host POSIX one). In addition,
there is no need to return an error in the case that the provided buffer
is too large.

### DIFF
--- a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
+++ b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
@@ -819,12 +819,14 @@ wasmtime_ssp_fd_prestat_dir_name(struct fd_prestats *prestats, __wasi_fd_t fd,
         rwlock_unlock(&prestats->lock);
         return error;
     }
-    if (path_len != strlen(prestat->dir)) {
+
+    const size_t prestat_dir_len = strlen(prestat->dir);
+    if (path_len < prestat_dir_len) {
         rwlock_unlock(&prestats->lock);
-        return EINVAL;
+        return __WASI_EINVAL;
     }
 
-    bh_memcpy_s(path, (uint32)path_len, prestat->dir, (uint32)path_len);
+    bh_memcpy_s(path, (uint32)path_len, prestat->dir, (uint32)prestat_dir_len);
 
     rwlock_unlock(&prestats->lock);
 


yamt:
> The error code ENAMETOOLONG is a more accurate description of the error condition if the provided path buffer is too short to fit the prestat directory name. In addition, there is no need to return an error in the case that the provided buffer is too large.

ENAMETOOLONG sounds weird to me for this particular case.

zoraaver:
@yamt I think the error code is fine. From the [posix spec page](https://pubs.opengroup.org/onlinepubs/9699919799.2008edition/basedefs/errno.h.html) `Filename too long` is the description.

I think this matches the error condition here pretty much exactly: the filename is too long to fit in the provided buffer.

yamt:
> @yamt I think the error code is fine. From the [posix spec page](https://pubs.opengroup.org/onlinepubs/9699919799.2008edition/basedefs/errno.h.html) `Filename too long` is the description.
> 
> I think this matches the error condition here pretty much exactly: the filename is too long to fit in the provided buffer.

ENAMETOOLONG usually means that the given filename is too long wrt PATH_MAX/NAME_MAX.
do you have a precedent where it's used to mean "too long for the user-provided buffer"?

loganek:
> ENAMETOOLONG sounds weird to me for this particular case.

Agree, I think EINVAL is fine in that case, as it's invalid value provided by the user that causes this error.

zoraaver:
> ENAMETOOLONG usually means that the given filename is too long wrt PATH_MAX/NAME_MAX.

Ok gotcha, I hadn't actually come across the error code before so just from the name and description it seemed to fit well.

> do you have a precedent where it's used to mean "too long for the user-provided buffer"?

It appears not, I've searched the internet but the error code seems to refer exclusively to that specific case of a path exceeding the OS path max.

I'll still need to change this function to return `__WASI_EINVAL` rather than `__EINVAL`. I'll also raise a PR for the tests to change the error code.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -819,12 +819,14 @@ wasmtime_ssp_fd_prestat_dir_name(struct fd_prestats *prestats, __wasi_fd_t fd,
         rwlock_unlock(&prestats->lock);
         return error;
     }
-    if (path_len != strlen(prestat->dir)) {
+
+    const size_t prestat_dir_len = (size_t)strlen(prestat->dir);
```
wenyongh, line 823: Had better use `strlen(prestat->dir) + 1` to contain `\0` for he below path_len check and bh_memcpy_s?
And no need to add `(size_t)`, the return type of strlen is already size_t.

yamt: iirc, the name here isn't NUL-terminated.

yamt (approved):

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
wenyongh: Yes, if it isn't '\0' terminated, we can merge the PR.
@zoraaver Could you help confirm it?

zoraaver: Hi @wenyongh, I can confirm that the input `path` is not null-terminated. `path_len` should be obtained from [`fd_prestat_get`](https://docs.rs/wasi/latest/wasi/struct.PrestatDir.html) which doesn't not include the null character [in our case]( https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c#L804).

I've removed the cast, thanks for spotting!

wenyongh (approved):
LGTM

PR #2581, opened 2023-09-21
Conditionally support SYNC flags

zoraaver:
To make it clearer to users when synchronization behaviour is not supported, return ENOTSUP when O_RSYNC, O_DSYNC or O_SYNC are respectively not defined. Linux also doesn't support O_RSYNC despite the O_RSYNC flag being defined.

### DIFF
--- a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
+++ b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
@@ -1282,18 +1282,20 @@ wasmtime_ssp_fd_fdstat_get(wasm_exec_env_t exec_env, struct fd_table *curfds,
 
     if ((ret & O_APPEND) != 0)
         buf->fs_flags |= __WASI_FDFLAG_APPEND;
-#ifdef O_DSYNC
+#ifdef CONFIG_HAS_O_DSYNC
     if ((ret & O_DSYNC) != 0)
         buf->fs_flags |= __WASI_FDFLAG_DSYNC;
 #endif
     if ((ret & O_NONBLOCK) != 0)
         buf->fs_flags |= __WASI_FDFLAG_NONBLOCK;
-#ifdef O_RSYNC
+#ifdef CONFIG_HAS_O_RSYNC
     if ((ret & O_RSYNC) != 0)
         buf->fs_flags |= __WASI_FDFLAG_RSYNC;
 #endif
+#ifdef CONFIG_HAS_O_SYNC
     if ((ret & O_SYNC) != 0)
         buf->fs_flags |= __WASI_FDFLAG_SYNC;
+#endif
     return 0;
 }
 
@@ -1306,21 +1308,25 @@ wasmtime_ssp_fd_fdstat_set_flags(wasm_exec_env_t exec_env,
     if ((fs_flags & __WASI_FDFLAG_APPEND) != 0)
         noflags |= O_APPEND;
     if ((fs_flags & __WASI_FDFLAG_DSYNC) != 0)
-#ifdef O_DSYNC
+#ifdef CONFIG_HAS_O_DSYNC
         noflags |= O_DSYNC;
 #else
-        noflags |= O_SYNC;
+        return __WASI_ENOTSUP;
 #endif
     if ((fs_flags & __WASI_FDFLAG_NONBLOCK) != 0)
         noflags |= O_NONBLOCK;
     if ((fs_flags & __WASI_FDFLAG_RSYNC) != 0)
-#ifdef O_RSYNC
+#ifdef CONFIG_HAS_O_RSYNC
         noflags |= O_RSYNC;
 #else
-        noflags |= O_SYNC;
+        return __WASI_ENOTSUP;
 #endif
     if ((fs_flags & __WASI_FDFLAG_SYNC) != 0)
+#ifdef CONFIG_HAS_O_SYNC
         noflags |= O_SYNC;
+#else
+        return __WASI_ENOTSUP;
+#endif
 
     struct fd_object *fo;
     __wasi_errno_t error =
@@ -1971,26 +1977,30 @@ wasmtime_ssp_path_open(wasm_exec_env_t exec_env, struct fd_table *curfds,
     if ((fs_flags & __WASI_FDFLAG_APPEND) != 0)
         noflags |= O_APPEND;
     if ((fs_flags & __WASI_FDFLAG_DSYNC) != 0) {
-#ifdef O_DSYNC
+#ifdef CONFIG_HAS_O_DSYNC
         noflags |= O_DSYNC;
-#else
-        noflags |= O_SYNC;
-#endif
         needed_inheriting |= __WASI_RIGHT_FD_DATASYNC;
+#else
+        return __WASI_ENOTSUP;
+#endif
     }
     if ((fs_flags & __WASI_FDFLAG_NONBLOCK) != 0)
         noflags |= O_NONBLOCK;
     if ((fs_flags & __WASI_FDFLAG_RSYNC) != 0) {
-#ifdef O_RSYNC
+#ifdef CONFIG_HAS_O_RSYNC
         noflags |= O_RSYNC;
-#else
-        noflags |= O_SYNC;
-#endif
         needed_inheriting |= __WASI_RIGHT_FD_SYNC;
+#else
+        return __WASI_ENOTSUP;
+#endif
     }
     if ((fs_flags & __WASI_FDFLAG_SYNC) != 0) {
+#ifdef CONFIG_HAS_O_SYNC
         noflags |= O_SYNC;
         needed_inheriting |= __WASI_RIGHT_FD_SYNC;
+#else
+        return __WASI_ENOTSUP;
+#endif
     }
     if (write && (noflags & (O_APPEND | O_TRUNC)) == 0)
         needed_inheriting |= __WASI_RIGHT_FD_SEEK;
--- a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/ssp_config.h
+++ b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/ssp_config.h
@@ -14,8 +14,8 @@
 #ifndef SSP_CONFIG_H
 #define SSP_CONFIG_H
 
+#include "bh_platform.h"
 #include "gnuc.h"
-#include <stdlib.h>
 
 #if defined(__FreeBSD__) || defined(__APPLE__) \
     || (defined(ANDROID) && __ANDROID_API__ < 28)
@@ -99,6 +99,20 @@
 #define st_atim st_atimespec
 #define st_ctim st_ctimespec
 #define st_mtim st_mtimespec
+#endif
+
+#if defined(O_DSYNC)
+#define CONFIG_HAS_O_DSYNC
+#endif
+
+// POSIX requires O_RSYNC to be defined, but Linux explicitly doesn't support
+// it.
+#if defined(O_RSYNC) && !defined(__linux__)
+#define CONFIG_HAS_O_RSYNC
+#endif
+
+#if defined(O_SYNC)
+#define CONFIG_HAS_O_SYNC
 #endif
 
 #if !defined(BH_PLATFORM_LINUX_SGX)


yamt:
> It's not yet clear whether these flags should be added to the standard so disallow use of them for the time being. See [WebAssembly/wasi-filesystem#98](https://github.com/WebAssembly/wasi-filesystem/issues/98) for more details. These changes are also necessary to ensure the rust WASI tests pass since they assert that path_open returns ENOTSUP when passing O_DSYNC/O_RSYNC/O_SYNC.

it sounds like a bug in tests to me.
preview1 has these flags defined.

lum1n0us:
It also bothers me. They are defined but not used in wasi-libc till ec4566beae84e54952637f0bf61bee4b4cacc087.

yamt:
> It also bothers me. They are defined but not used in wasi-libc till ec4566beae84e54952637f0bf61bee4b4cacc087.

what do you mean by "not used in wasi-libc"?
doesn't open() pass-through them to wasi?

yamt:
https://github.com/WebAssembly/wasi-testsuite/pull/91

zoraaver:
> it sounds like a bug in tests to me.

I don't think it's that clear-cut. I'm not suggesting that this change is definitive or the correct approach but for the time being I would say it's better to explicitly not support these flags rather than offering half-baked support depending on the underlying OS. e.g. Linux doesn't actually support `O_RSYNC`.

The purpose of this change is mainly so we can start running the rust tests in CI, not to spark a discussion on the use of these flags. 

If you want to merge https://github.com/WebAssembly/wasi-testsuite/pull/91, that also works but it made more sense to me personally to remove support for these flags in WAMR.

@wenyongh do you have any further thoughts? This is a breaking change so would require mentioning in the release notes if I'm not mistaken.

wenyongh:
> 
> If you want to merge [WebAssembly/wasi-testsuite#91](https://github.com/WebAssembly/wasi-testsuite/pull/91), that also works but it made more sense to me personally to remove support for these flags in WAMR.
> 
> @wenyongh do you have any further thoughts? This is a breaking change so would require mentioning in the release notes if I'm not mistaken.

How about making these flags optional, just let the developer choose whether to support it or not? By default they are unsupported, and we add a macro to control the code and mention the change in the release notes?

zoraaver:
Thanks, that sounds like a good idea. I'll amend this PR to add a build flag to control whether these flags are supported or not so developers can opt in if they require it for their use cases.

yamt:
do you mean something like CONFIG_HAS_ macros in ssp_config.h?

yamt:
> do you mean something like CONFIG_HAS_ macros in ssp_config.h?

i didn't notice you have already updated the patch when asking this.
anyway, to me, it seems more appropriate to use CONFIG_HAS_ style macros than user-facing config knobs.

build-scripts/config_common.cmake
```diff
@@ -201,6 +201,11 @@ elseif (WAMR_BUILD_LIBC_WASI EQUAL 1)
 else ()
   message ("     Libc WASI disabled")
 endif ()
+if (WAMR_BUILD_ENABLE_LIBC_WASI_SYNC_FLAGS EQUAL 1)
+  message ("     Libc WASI file descriptor sync flags enabled")
+else()
+  message ("     Libc WASI file descriptor sync flags disabled")
```
wenyongh, line 207: Suggest not outputting the message of else since there are a lot messages already, no need to output some not so important messages. How about only outputting it when enabled?

core/iwasm/libraries/libc-wasi/libc_wasi.cmake
```diff
@@ -5,6 +5,12 @@ set (LIBC_WASI_DIR ${CMAKE_CURRENT_LIST_DIR})
 
 add_definitions (-DWASM_ENABLE_LIBC_WASI=1)
 
+if (WAMR_BUILD_ENABLE_LIBC_WASI_SYNC_FLAGS EQUAL 1)
+    add_definitions (-DWASM_ENABLE_LIBC_WASI_SYNC_FLAGS=1)
```
wenyongh, line 9: As suggested by yamt, how about `-DCONFIG_HAS_LIBC_WASI_SYNC_FLAGS=1` instead? And in ssp_config.h, add:
```C
#ifndef CONFIG_HAS_LIBC_WASI_SYNC_FLAGS
#define CONFIG_HAS_LIBC_WASI_SYNC_FLAGS 0
#endif
```

core/iwasm/libraries/libc-wasi/libc_wasi.cmake
```diff
@@ -5,6 +5,12 @@ set (LIBC_WASI_DIR ${CMAKE_CURRENT_LIST_DIR})
...
+else()
+    add_definitions (-DWASM_ENABLE_LIBC_WASI_SYNC_FLAGS=0)
```
wenyongh, line 11: No need to add the else branch if the macro is set to 0 by default in ssp_config.h.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -1970,6 +1970,11 @@ wasmtime_ssp_path_open(wasm_exec_env_t exec_env, struct fd_table *curfds,
     // Convert file descriptor flags.
     if ((fs_flags & __WASI_FDFLAG_APPEND) != 0)
         noflags |= O_APPEND;
+    if ((fs_flags & __WASI_FDFLAG_NONBLOCK) != 0)
+        noflags |= O_NONBLOCK;
+    if (write && (noflags & (O_APPEND | O_TRUNC)) == 0)
+        needed_inheriting |= __WASI_RIGHT_FD_SEEK;
+#if WASM_ENABLE_LIBC_WASI_SYNC_FLAGS == 1
```
wenyongh, line 1977: Had better use `!= 0` instead of `== 1`, we usually use that style.

core/iwasm/libraries/libc-wasi/libc_wasi.cmake
yamt: do we want to have a user-settable option at all?
if O_RSYNC is known broken on Linux, isn't it more straightforward to use `#if defined(__Linux__)`?

build-scripts/config_common.cmake
zoraaver: Sure 👍

core/iwasm/libraries/libc-wasi/libc_wasi.cmake
zoraaver: I'm not sure what guarantees other POSIX-like OS's make about O_RSYNC so `#if defined(__linux__)` probably wouldn't be enough. 

If the user knows the host OS they're going to be running WAMR on supports all these sync flags, they can enable support. So I'd still say it's worth having a user option for that use case.

yamt: do you mean all CONFIG_HAS_ flags should be a user option?

zoraaver: no, just this flag. The other `CONFIG_HAS_` flags can be deduced from the platform which is being compiled for and would cause a compile error if they were incorrectly set. afaik there is no such way to deduce whether a platform supports `O_RSYNC` etc. via macros - if there were, we could get rid of the user option.

N.B. the existence of `O_RSYNC` flag _doesn't imply_ it's supported.

yamt: IMO, it's reasonable to assume the existence of O_RSYNC implies some support of it.
we can add platform ifdefs for known exceptions.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -1970,6 +1970,11 @@ wasmtime_ssp_path_open(wasm_exec_env_t exec_env, struct fd_table *curfds,
     // Convert file descriptor flags.
     if ((fs_flags & __WASI_FDFLAG_APPEND) != 0)
         noflags |= O_APPEND;
+    if ((fs_flags & __WASI_FDFLAG_NONBLOCK) != 0)
+        noflags |= O_NONBLOCK;
+    if (write && (noflags & (O_APPEND | O_TRUNC)) == 0)
+        needed_inheriting |= __WASI_RIGHT_FD_SEEK;
```
yamt, line 1976: is this related to the rest of the PR?

zoraaver: Yes, this is just moving the existing code around (it made sticking all the sync flags behind an ifdef easier). If you look a few lines below here, the same lines are deleted.

core/iwasm/libraries/libc-wasi/libc_wasi.cmake
wenyongh: @yamt do you mean that if macro O_RSYNC is defined, then the fd sync flags should be supported? If that, maybe we can just add simple control in posix.c. But from @zoraaver's description, it's not yet clear whether these flags should be added to the standard or not, see https://github.com/WebAssembly/wasi-filesystem/issues/98, so we add the building option to let developer to choose.

yamt: > @yamt do you mean that if macro O_RSYNC is defined, then the fd sync flags should be supported? If that, maybe we can just add simple control in posix.c.

i meant something like
```
#if defined(O_RSYNC) && !defined(__linux__)
#define HAVE_O_RSYNC
#endif
```

> But from @zoraaver's description, it's not yet clear whether these flags should be added to the standard or not, see [WebAssembly/wasi-filesystem#98](https://github.com/WebAssembly/wasi-filesystem/issues/98), so we add the building option to let developer to choose.

these flags are defined in preview1.
the mentioned discussion is about preview2.

zoraaver: > these flags are defined in preview1.

Yup, sorry I got a bit mixed up. These flags are already defined in preview1 like you mentioned. But I _still_ think we shouldn't support them unconditionally. A user option could be a bit more reliable than deducing support from the OS + existence of the sync flag. But I don't have a strong opinion on it either way tbh, as long as we don't support them by default since that's misleading for users whose programs might depend on the data guarantees offered by `R_SYNC` etc.

> But from @zoraaver's description

That's my bad, I'll edit the description - I didn't realize that issue was referring to preview2 and not preview1 (which appears to be finalized).

yamt: the i/o integrity guarantees are complex and involves many components; OS, filesystem, disk, battery, etc.
cf. https://www.postgresql.org/docs/current/wal-reliability.html

having said that, my impression is that it isn't wasm runtime (or user who configures the runtime) who is responsible to determine what integrity the whole system provides.
IMO, where the underlying platform is supposed to provide the functionality, it's enough for us to pass through the requests to the underlying platform.
how do you think?

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -1992,8 +1995,16 @@ wasmtime_ssp_path_open(wasm_exec_env_t exec_env, struct fd_table *curfds,
         noflags |= O_SYNC;
         needed_inheriting |= __WASI_RIGHT_FD_SYNC;
     }
-    if (write && (noflags & (O_APPEND | O_TRUNC)) == 0)
-        needed_inheriting |= __WASI_RIGHT_FD_SEEK;
+#else
+    // It's not clear whether we want to support these flags in the standard
+    // yet: https://github.com/WebAssembly/wasi-filesystem/issues/98 so disable
```
loganek, line 2000: I don't think this comment is related. It's about preview 2, whereas this function is for preview 1 only. And those flags are already part of the preview 1 standard: https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/docs.md#record-members-4
My suggestion is to update the preview1 doc saying that this feature is not available on some platforms therefore `path_open` might return `ENOTSUP` in that case.

core/iwasm/libraries/libc-wasi/libc_wasi.cmake
wenyongh: @yamt do you mean to define the CONFIG_HAS_O_xxx macros in ssp_config.h, and use them to control the code in posix.c? And don't add WAMR_BUILD_ENABLE_LIBC_WASI_SYNC_FLAGS and CONFIG_HAS_LIBC_WASI_SYNC_FLAGS?

For example, in ssp_config.h, add:
```C
#if defined(O_DSYNC) && !defined(__linux__)
#define CONFIG_HAS_O_DSYNC
#endif

#if defined(O_RSYNC) && !defined(__linux__)
#define CONFIG_HAS_O_RSYNC
#endif

#if defined(O_SYNC) && !defined(__linux__)
#define CONFIG_HAS_O_SYNC
#endif
```
in posix.c:
```C
    if (fs_flags & __WASI_FDFLAG_DSYNC != 0) {
#ifdef CONFIG_HAS_O_DSYNC
        noflags |= O_DSYNC;
        needed_inheriting |= __WASI_RIGHT_FD_DATASYNC;
#else
        return __WASI_ENOTSUP;
#endif
    }

    if (fs_flags & __WASI_FDFLAG_RSYNC != 0) {
#ifdef CONFIG_HAS_O_RSYNC
        noflags |= O_RSYNC;
        needed_inheriting |= __WASI_RIGHT_FD_SYNC;
#else
        return __WASI_ENOTSUP;
#endif
    }

    if (fs_flags & __WASI_FDFLAG_SYNC != 0) {
#ifdef CONFIG_HAS_O_SYNC
        noflags |= O_SYNC;
        needed_inheriting |= __WASI_RIGHT_FD_SYNC;
#else
        return __WASI_ENOTSUP;
#endif
    }
```
It is good to me if it can resolve the issue to run the rust WASI tests.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
wenyongh: Do you agree with the suggestion that yamt proposed above?
https://github.com/bytecodealliance/wasm-micro-runtime/pull/2581#discussion_r1341270520

core/iwasm/libraries/libc-wasi/libc_wasi.cmake
yamt: > @yamt do you mean to define the CONFIG_HAS_O_xxx macros in ssp_config.h, and use them to control the code in posix.c? And don't add WAMR_BUILD_ENABLE_LIBC_WASI_SYNC_FLAGS and CONFIG_HAS_LIBC_WASI_SYNC_FLAGS?

yes.

> It is good to me if it can resolve the issue to run the rust WASI tests.

it doesn't solve the test issue.

my suggestion for the test issue: https://github.com/WebAssembly/wasi-testsuite/pull/91

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
loganek: yes, I think that makes sense.

core/iwasm/libraries/libc-wasi/libc_wasi.cmake
zoraaver: >IMO, where the underlying platform is supposed to provide the functionality, it's enough for us to pass through the requests to the underlying platform.

thanks, that makes sense actually.

I've implemented the suggestion to provide support based on the definition of these sync flags.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
zoraaver: Removed the comment, thanks. I'll raise a PR shortly on the docs to clarify that sync behaviour is platform-dependent.

wenyongh (commented):
LGTM

loganek:
Looks good, thanks

loganek (approved):

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
zoraaver: https://github.com/WebAssembly/WASI/pull/560

PR #2598, opened 2023-09-26
Add Cosmopolitan Libc Platform

G4Vi:
This patch adds the Cosmopolitan Libc platform enabling compatibility with multiple x86_64 operating systems with the same binary. The platform is similar to the Linux platform, but for now only x86_64 with interp modes are supported.

The only major change to the core is `posix.c/convert_errno()` was rewritten to use a switch statement. With Cosmopolitan errno values depend on the currently running operating system, and so they are non-constant and cannot be used in array designators. However, the `cosmocc` compiler allows non-constant case labels in switch statements, enabling the new version.

The spec tests can be ran via `CC=cosmocc ./test_wamr.sh -j cosmopolitan -t classic-interp` or `CC=cosmocc ./test_wamr.sh -j cosmopolitan -t fast-interp` . However, they fail in the same way they do on the Linux platform on my machine. Tests with float64 NaN fail without this patch I am in the process of upstreaming into Cosmopolitan: https://github.com/jart/cosmopolitan/pull/901 .
`tests/wamr-test-suites/test_wamr.sh` lines 669-671 (https://github.com/dylibso/wasm-micro-runtime/blob/339facc7bf80e4a68d81eb56d4ec358c6c257176/tests/wamr-test-suites/test_wamr.sh#L669-L671) contain a hack to enable the test suite to run even though the the platform is incorrectly detected as Linux by other scripts (Linux is the only platform supported to build Cosmopolitan). If this isn't acceptable, I am happy to remove it, but would appreciate some guidance on the best way to propagate the platform to the child scripts, it looks messy how it's handled for linux sgx.

Edit: Updated test instructions after updating `-j` to specify platform in a3449ac.

### DIFF
--- a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
+++ b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
@@ -60,98 +60,101 @@ static_assert(sizeof(struct iovec) == sizeof(__wasi_ciovec_t),
 static __wasi_errno_t
 convert_errno(int error)
 {
-    static const __wasi_errno_t errors[] = {
-#define X(v) [v] = __WASI_##v
-        X(E2BIG),
-        X(EACCES),
-        X(EADDRINUSE),
-        X(EADDRNOTAVAIL),
-        X(EAFNOSUPPORT),
-        X(EAGAIN),
-        X(EALREADY),
-        X(EBADF),
-        X(EBADMSG),
-        X(EBUSY),
-        X(ECANCELED),
-        X(ECHILD),
-        X(ECONNABORTED),
-        X(ECONNREFUSED),
-        X(ECONNRESET),
-        X(EDEADLK),
-        X(EDESTADDRREQ),
-        X(EDOM),
-        X(EDQUOT),
-        X(EEXIST),
-        X(EFAULT),
-        X(EFBIG),
-        X(EHOSTUNREACH),
-        X(EIDRM),
-        X(EILSEQ),
-        X(EINPROGRESS),
-        X(EINTR),
-        X(EINVAL),
-        X(EIO),
-        X(EISCONN),
-        X(EISDIR),
-        X(ELOOP),
-        X(EMFILE),
-        X(EMLINK),
-        X(EMSGSIZE),
-        X(EMULTIHOP),
-        X(ENAMETOOLONG),
-        X(ENETDOWN),
-        X(ENETRESET),
-        X(ENETUNREACH),
-        X(ENFILE),
-        X(ENOBUFS),
-        X(ENODEV),
-        X(ENOENT),
-        X(ENOEXEC),
-        X(ENOLCK),
-        X(ENOLINK),
-        X(ENOMEM),
-        X(ENOMSG),
-        X(ENOPROTOOPT),
-        X(ENOSPC),
-        X(ENOSYS),
+    __wasi_errno_t code = __WASI_ENOSYS;
+#define X(v)               \
+    case v:                \
+        code = __WASI_##v; \
+        break;
+    switch (error) {
+        X(E2BIG)
+        X(EACCES)
+        X(EADDRINUSE)
+        X(EADDRNOTAVAIL)
+        X(EAFNOSUPPORT)
+        X(EAGAIN)
+        X(EALREADY)
+        X(EBADF)
+        X(EBADMSG)
+        X(EBUSY)
+        X(ECANCELED)
+        X(ECHILD)
+        X(ECONNABORTED)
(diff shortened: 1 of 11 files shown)


yamt:
does this mean to build wamr this way: https://justine.lol/ape.html ?

G4Vi:
Yes, exactly! This allows building Actually Portable Executables that run on multiple x86_64 systems with the same binary. `iwasm.com`  can be built by setting up `cosmocc` as described in https://github.com/jart/cosmopolitan#getting-started and running `./build_cosmoscc.sh` from `product-mini/platforms/cosmopolitan`

tests/wamr-test-suites/test_wamr.sh
```diff
@@ -58,7 +59,7 @@ QEMU_FIRMWARE=""
 # prod/testsuite-all branch
 WASI_TESTSUITE_COMMIT="cf64229727f71043d5849e73934e249e12cb9e06"
 
-while getopts ":s:cabgvt:m:MCpSXxwPGQF:" opt
+while getopts ":s:cabgvt:m:MCpSXxwPGQFj:" opt
```
wenyongh, line 62: Should not change "F:" to "F", ':' means there is an extra argument for the the option, e.g. "-F <QEMU_FIRMWARE>" and "-m \<TARGET\>". So here it should be "F:j".

How about using "-j <PLATFORM>" to specify a platform so we could support more platforms in the future? In which platform is set in L55 "PLATFORM=$(uname -s | tr A-Z a-z)" by default, and we can use "./test_wamr.sh -j cosmopolitan" to set it to cosmopolitan.
Here change it to "F:j:", and change below to be like:
```shell
        j)
        echo "test platform " ${OPTARG}
        PLATFORM=${OPTARG}
        ;;
```

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -60,98 +60,101 @@ static_assert(sizeof(struct iovec) == sizeof(__wasi_ciovec_t),
...
-#if EOPNOTSUPP != ENOTSUP
-        [EOPNOTSUPP] = __WASI_ENOTSUP,
-#endif
-#if EWOULDBLOCK != EAGAIN
-        [EWOULDBLOCK] = __WASI_EAGAIN,
-#endif
-    };
-    if (error < 0 || (size_t)error >= sizeof(errors) / sizeof(errors[0])
-        || errors[error] == 0)
+    if (code == 0)
```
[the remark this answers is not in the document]
G4Vi: It could be zero if it is set in a switch case and the `__WASI_` value is zero. If that will never be the case we can remove it. However, I think the old code defended against it with the `errors[error] == 0`  check.

tests/wamr-test-suites/test_wamr.sh
G4Vi: Good catch and suggestion. Fixed`-F` and adapted `-j` to specify platform in a3449ac .

tests/wamr-test-suites/test_wamr.sh
```diff
@@ -162,8 +162,8 @@ do
         QEMU_FIRMWARE=${OPTARG}
         ;;
         j)
-        echo "test cosmopolitan"
-        PLATFORM="cosmopolitan"
+        echo "test platform" ${OPTARG}
```
wenyongh, line 165: Had better add a space, "test platform " ${OPTARG}

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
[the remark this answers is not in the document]
wenyongh: The 0 is __WASI_SUCCESS, see
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/include/wasmtime_ssp.h#L80

which is not in the switch branch here, other __WASI_ values are not 0. It seems that the old code added the unnecessary check.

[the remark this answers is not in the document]
G4Vi: Removed unnecessary check in 27023ee

tests/wamr-test-suites/test_wamr.sh
G4Vi: Adjusted spacing to match in 27023ee

tests/wamr-test-suites/test_wamr.sh
G4Vi: Adjusted spacing to match in 27023ee

wenyongh (approved):
LGTM

Issue #2602, opened 2023-09-27
Different running results while executing the same wasm file with different runtime tools


luxinyi0105:
### Describe the bug
While executing the given testcase with wasm-micro-runtime and other runtime tools, the results are different.

### Test Case
[The given testcase](https://github.com/luxinyi0105/wasm_runtime_test/raw/main/09-27.zip) was simply mutated a wasm file, which was obtained by compiling C program generated with `Csmith` using `Emscripten Compiler`(Emcc). 

The original C program is `c_file.c`, the compilation results with Emscripten is `wasm_file.wasm`, and its wat format is `wat_file.wat`. 

We mutated the wat file as follows:
   change all the `i32/i64.sub` with `i32/i64.add`
   change all the `i32/i64.gt_u` with `i32/i64.le_s`
   change all the `i32/i64.ge_s` with `i32/i64.le_u`
The result after mutation is `mutated_file.wat`, and its wasm format is `mutated_file.wasm`.

### Actual Results
While executing the given testcase with command `iwasm mutated_file.wasm`, wamr will output a very long garbled code and attach `Exception: out of bounds memory access` at the end of it.

Because the output result is too long, I won't show it here.

### Expected Results
I use some other runtime tools. such as wasmer, wasmtime, wasmedge and wasmi, to execute the same testcase. All of their running results are the same: `outputs nothing`. Their running results are shown in the following figure.
![截图 2023-09-27 16-27-06](https://github.com/bytecodealliance/wasm-micro-runtime/assets/92167273/5773094b-980a-4141-a9c7-07e8786265ef)
 
I'm not sure which execution result is correct, but in terms of quantity comparison, wamr doesn't seem to have an advantage. The detail still needs you to confirm.

### Versions and Environment
Tools version: iwasm 1.2.3
Operating system: Ubuntu 22.04.1
Architecture: x86_64

wenyongh:
Hi, thanks for reporting the issue! It is due to the stdin fd is allowed to write by default, I submitted #2606 to fix it, please try again.

PR #2606, opened 2023-09-28
Limit the read/write rights for stdin/stdout/stderr

wenyongh:


yamt:
why?

wenyongh:
> why?

Hi, it is to fix the issue reported in https://github.com/bytecodealliance/wasm-micro-runtime/issues/2602.

The file descriptor of stdin should be read-only, and the file descriptor of stdout/stderr should be write-only. If writing data to stdin, unexpected behavior may occur.

See the handling of wasmtime and wasmedge:
https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasi/src/preview2/preview1.rs#L1052-L1069
https://github.com/WasmEdge/WasmEdge/blob/master/lib/host/wasi/environ.cpp#L58-L66

yamt:
> > why?
> 
> Hi, it is to fix the issue reported in #2602.
> 
> The file descriptor of stdin should be read-only, and the file descriptor of stdout/stderr should be write-only. If writing data to stdin, unexpected behavior may occur.

do you have references to standards/spec?

on traditional posix-like environment, it's common to assign an open tty, which is read/write, for all three descriptors.

> 
> See the handling of wasmtime and wasmedge: https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasi/src/preview2/preview1.rs#L1052-L1069 https://github.com/WasmEdge/WasmEdge/blob/master/lib/host/wasi/environ.cpp#L58-L66

wenyongh:
> > > why?
> > 
> > 
> > Hi, it is to fix the issue reported in #2602.
> > The file descriptor of stdin should be read-only, and the file descriptor of stdout/stderr should be write-only. If writing data to stdin, unexpected behavior may occur.
> 
> do you have references to standards/spec?
> 

I cannot find reference in wasi document, like:
https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/docs.md

But from the linux and libc manual, the stdin is for intput, the stdout is for output:
https://man7.org/linux/man-pages/man3/stdout.3.html
https://www.gnu.org/software/libc/manual/html_node/Standard-Streams.html
And many other runtimes limits their rights. It is strange to write data to stdin and read data from stdout. In issue #2602, writing data to stdin may lead to core dump.

> on traditional posix-like environment, it's common to assign an open tty, which is read/write, for all three descriptors.
> 

Do you have samples? Maybe we can test whether it can run before/after this patch is applied.

yamt:
> > > > why?
> > > 
> > > 
> > > Hi, it is to fix the issue reported in #2602.
> > > The file descriptor of stdin should be read-only, and the file descriptor of stdout/stderr should be write-only. If writing data to stdin, unexpected behavior may occur.
> > 
> > 
> > do you have references to standards/spec?
> 
> I cannot find reference in wasi document, like: https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/docs.md
> 
> But from the linux and libc manual, the stdin is for intput, the stdout is for output: https://man7.org/linux/man-pages/man3/stdout.3.html https://www.gnu.org/software/libc/manual/html_node/Standard-Streams.html And many other runtimes limits their rights. It is strange to write data to stdin and read data from stdout. In issue #2602, writing data to stdin may lead to core dump.
> 
> > on traditional posix-like environment, it's common to assign an open tty, which is read/write, for all three descriptors.
> 
> Do you have samples? Maybe we can test whether it can run before/after this patch is applied.

macOS and ubuntu work like:
```
spacetanuki% cat a.c
#include <errno.h>
#include <stdio.h>
#include <string.h>
#include <unistd.h>

int
main()
{
        ssize_t sz = write(STDIN_FILENO, "hello\n", 6);
        if (sz == -1) {
                printf("write failed: %s\n", strerror(errno));
        }
}
spacetanuki% cc a.c
spacetanuki% ./a.out
hello
spacetanuki%
```

wenyongh:
> > > > > why?
> > > > 
> > > > 
> > > > Hi, it is to fix the issue reported in #2602.
> > > > The file descriptor of stdin should be read-only, and the file descriptor of stdout/stderr should be write-only. If writing data to stdin, unexpected behavior may occur.
> > > 
> > > 
> > > do you have references to standards/spec?
> > 
> > 
> > I cannot find reference in wasi document, like: https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/docs.md
> > But from the linux and libc manual, the stdin is for intput, the stdout is for output: https://man7.org/linux/man-pages/man3/stdout.3.html https://www.gnu.org/software/libc/manual/html_node/Standard-Streams.html And many other runtimes limits their rights. It is strange to write data to stdin and read data from stdout. In issue #2602, writing data to stdin may lead to core dump.
> > > on traditional posix-like environment, it's common to assign an open tty, which is read/write, for all three descriptors.
> > 
> > 
> > Do you have samples? Maybe we can test whether it can run before/after this patch is applied.
> 
> macOS and ubuntu work like:
> 
> ```
> spacetanuki% cat a.c
> #include <errno.h>
> #include <stdio.h>
> #include <string.h>
> #include <unistd.h>
> 
> int
> main()
> {
>         ssize_t sz = write(STDIN_FILENO, "hello\n", 6);
>         if (sz == -1) {
>                 printf("write failed: %s\n", strerror(errno));
>         }
> }
> spacetanuki% cc a.c
> spacetanuki% ./a.out
> hello
> spacetanuki%
> ```

Yes, the case works in my side also. But it may be unsafe to write long buffer to stdin, the below case ran crashed in the MobaXterm when I used it as ssh client for the Ubuntu-22.04 (Error 232: The pipe is being closed.):
```C
#include <errno.h>
#include <stdio.h>
#include <string.h>
#include <unistd.h>

int
main()
{
    unsigned char buf[1024 * 1024];
    int i;

    for (i = 0; i < sizeof(buf) / sizeof(unsigned char); i++) {
        buf[i] = (unsigned char)i;
    }

    for (i = 0; i < 10; i++) {
        ssize_t sz = write(STDIN_FILENO, buf, sizeof(buf) / sizeof(unsigned char));
        if (sz == -1) {
            printf("write failed: %s\n", strerror(errno));
        }
    }

    return 0;
}
```

yamt:
> > > > > > why?
> > > > > 
> > > > > 
> > > > > Hi, it is to fix the issue reported in #2602.
> > > > > The file descriptor of stdin should be read-only, and the file descriptor of stdout/stderr should be write-only. If writing data to stdin, unexpected behavior may occur.
> > > > 
> > > > 
> > > > do you have references to standards/spec?
> > > 
> > > 
> > > I cannot find reference in wasi document, like: https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/docs.md
> > > But from the linux and libc manual, the stdin is for intput, the stdout is for output: https://man7.org/linux/man-pages/man3/stdout.3.html https://www.gnu.org/software/libc/manual/html_node/Standard-Streams.html And many other runtimes limits their rights. It is strange to write data to stdin and read data from stdout. In issue #2602, writing data to stdin may lead to core dump.
> > > > on traditional posix-like environment, it's common to assign an open tty, which is read/write, for all three descriptors.
> > > 
> > > 
> > > Do you have samples? Maybe we can test whether it can run before/after this patch is applied.
> > 
> > 
> > macOS and ubuntu work like:
> > ```
> > spacetanuki% cat a.c
> > #include <errno.h>
> > #include <stdio.h>
> > #include <string.h>
> > #include <unistd.h>
> > 
> > int
> > main()
> > {
> >         ssize_t sz = write(STDIN_FILENO, "hello\n", 6);
> >         if (sz == -1) {
> >                 printf("write failed: %s\n", strerror(errno));
> >         }
> > }
> > spacetanuki% cc a.c
> > spacetanuki% ./a.out
> > hello
> > spacetanuki%
> > ```
> 
> Yes, the case works in my side also. But it may be unsafe to write long buffer to stdin, the below case ran crashed in the MobaXterm when I used it as ssh client for the Ubuntu-22.04 (Error 232: The pipe is being closed.):
> 
> ```c
> #include <errno.h>
> #include <stdio.h>
> #include <string.h>
> #include <unistd.h>
> 
> int
> main()
> {
>     unsigned char buf[1024 * 1024];
>     int i;
> 
>     for (i = 0; i < sizeof(buf) / sizeof(unsigned char); i++) {
>         buf[i] = (unsigned char)i;
>     }
> 
>     for (i = 0; i < 10; i++) {
>         ssize_t sz = write(STDIN_FILENO, buf, sizeof(buf) / sizeof(unsigned char));
>         if (sz == -1) {
>             printf("write failed: %s\n", strerror(errno));
>         }
>     }
> 
>     return 0;
> }
> ```

it sounds like a bug in the terminal software rather than stdout vs stdin.
doesn't it crash with STDOUT_FILENO?

wenyongh:
> > 
> > Yes, the case works in my side also. But it may be unsafe to write long buffer to stdin, the below case ran crashed in the MobaXterm when I used it as ssh client for the Ubuntu-22.04 (Error 232: The pipe is being closed.):
> > ```c
> > #include <errno.h>
> > #include <stdio.h>
> > #include <string.h>
> > #include <unistd.h>
> > 
> > int
> > main()
> > {
> >     unsigned char buf[1024 * 1024];
> >     int i;
> > 
> >     for (i = 0; i < sizeof(buf) / sizeof(unsigned char); i++) {
> >         buf[i] = (unsigned char)i;
> >     }
> > 
> >     for (i = 0; i < 10; i++) {
> >         ssize_t sz = write(STDIN_FILENO, buf, sizeof(buf) / sizeof(unsigned char));
> >         if (sz == -1) {
> >             printf("write failed: %s\n", strerror(errno));
> >         }
> >     }
> > 
> >     return 0;
> > }
> > ```
> 
> it sounds like a bug in the terminal software rather than stdout vs stdin. doesn't it crash with STDOUT_FILENO?

I checked again, it also crashed with STDOUT_FILENO. It should be a bug of the ssh client, both this case (no matter STDIN_FILENO or STDOUT_FILENO) and the case of #2602 got crashed with the ssh client, while they ran well in a Ubuntu of VirtualBox.

It makes me a little confusing, both STDIN_FILENO and STDOUT_FILENO can be used as input and output? What is the difference between them?

yamt:
> It makes me a little confusing, both STDIN_FILENO and STDOUT_FILENO can be used as input and output? What is the difference between them?

a terminal usually allocates a tty and dup() it to file descriptors 0-2.
open mode (O_RDWR etc) belongs to files (the tty in this case), not each file descriptors.
in this case, nothing prevents you from writing to STDIN_FILENO.

when you use a pipe (eg `ls | more`) STDOUT_FILENO and STDIN_FILENO actually reference different files with different open modes.

PR #2607, opened 2023-09-28
Upgrade clang-format version from 12 to 14

eloparco:
`clang-12` has been removed from Ubuntu `22.04`: https://github.com/actions/runner-images/blob/ubuntu22/20230924.1/images/linux/Ubuntu2204-Readme.md and the coding guidelines job started to fail.

[Last job](https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/6324887158/job/17175263160) using 22.04.3 version `20230917.1.0` was still fine, but the [latest](https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/6338189476/job/17214782992?pr=2599) with `20230924.1.0` fails.
Unfortunately, there's no way to pin the version: https://github.com/actions/runner-images/issues/2894.

As an alternative, we could downgrade to use Ubuntu 20.04 and keep `clang-12`.

### DIFF
--- a/.github/workflows/coding_guidelines.yml
+++ b/.github/workflows/coding_guidelines.yml
@@ -16,7 +16,7 @@ concurrency:
 
 jobs:
   compliance_job:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - name: checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


eloparco:
Actually, when I tried applying `clang-14` it was too different from `clang-12`, so better to revert to Ubuntu `20.04` in #2599. I tried to modify `.clang-format` to achieve the same formatting we currently have, but I didn't manage to.

PR #2610, opened 2023-09-29
Enforce compilation errors

eloparco:
Fail compilation of `iwasm` and `wamrc` if warnings.
One of the use cases is to forget removing unused variables, e.g. #1913.


wenyongh:
Close this PR since it hasn't been updated for a long time, and it is not easy to enforce compilation errors since there are many targets and platforms to support.

Issue #1365, opened 2022-08-08
AOT compiled WASM couldn't be loaded on macOS M1 environment


cosmo0920:
## Step to replicates

1. Build ./sample/basic WASM application
2. Move to wamr-compiler and build wamrc
3. Built AOT object from  ./out/wasm-apps/testapp.wasm with wamrc: 
```console
/path/to/wamr-compiler/build/wamrc --cpu=apple-m1 --size-level=3 -o ./out/wasm-apps/testapp.aot ./out/wasm-apps/testapp.wasm
Create AoT compiler with:
  target:        arm64
  target cpu:    apple-m1
  cpu features:  
  opt level:     3
  size level:    3
  output format: AoT file
Compile success, file ./out/wasm-apps/testapp.aot was generated.
```
4. Execute basic sample application with AOT object.
5. Then, :boom:! 
```console
$  ./out/basic -f out/wasm-apps/testapp.aot
Load wasm module failed. error: AOT module load failed: mmap memory failed
```

## Environment

Software versions:

```console
$ sw_vers
ProductName:	macOS
ProductVersion:	12.5
BuildVersion:	21G72
```

Host Compiler version:

```console
$ clang --version
Apple clang version 13.1.6 (clang-1316.0.21.2.5)
Target: arm64-apple-darwin21.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

WASI SDK's compiler version:

```console
$ ~/work/wasi-sdk-16.0/bin/clang --version
clang version 14.0.4 (https://github.com/llvm/llvm-project 29f1039a7285a5c3a9c353d054140bf2556d4c4d)
Target: wasm32-unknown-wasi
Thread model: posix
InstalledDir: ~/work/wasi-sdk-16.0/bin
```

cosmo0920:
I tried to take a look on M1 Mac issue and I found the several issue before calling WASM function part:

1. With macOS arm64 runtime, it is always hardened. We shouldn't use `PROT_EXEC` flags in mmap. Otherwise, mmap is failed.

```diff
diff --git a/wasm-micro-runtime-WAMR-05-18-2022/core/shared/platform/common/posix/posix_memmap.c b/wasm-micro-runtime-WAMR-05-18-2022/core/shared/platform/common/posix/posix_memmap.c
index 2dfbee453..1de331ea3 100644
[...]

PR #2618, opened 2023-10-05
Enable AOT usage on M1 mac

eloparco:
After reading the [docs](https://developer.apple.com/documentation/apple-silicon/porting-just-in-time-compilers-to-apple-silicon), I managed to use AOT on my M1 mac.

Issue: https://github.com/bytecodealliance/wasm-micro-runtime/issues/1365

### DIFF
--- a/core/iwasm/aot/aot_loader.c
+++ b/core/iwasm/aot/aot_loader.c
@@ -3232,10 +3232,13 @@ aot_load_from_aot_file(const uint8 *buf, uint32 size, char *error_buf,
     if (!module)
         return NULL;
 
+    os_thread_jit_write_protect_np(false); /* Make memory writable */
     if (!load(buf, size, module, error_buf, error_buf_size)) {
         aot_unload(module);
         return NULL;
     }
+    os_thread_jit_write_protect_np(true); /* Make memory executable */
+    os_icache_flush(module->code, module->code_size);
 
     LOG_VERBOSE("Load module success.\n");
     return module;
--- a/core/iwasm/aot/arch/aot_reloc_aarch64.c
+++ b/core/iwasm/aot/arch/aot_reloc_aarch64.c
@@ -53,7 +53,12 @@ get_target_symbol_map(uint32 *sym_num)
     return target_sym_map;
 }
 
+#if (defined(__APPLE__) || defined(__MACH__)) && defined(__arm64__)
+#define BUILD_TARGET_AARCH64_DEFAULT "arm64"
+#else
 #define BUILD_TARGET_AARCH64_DEFAULT "aarch64v8"
+#endif
+
 void
 get_current_target(char *target_buf, uint32 target_buf_size)
 {
--- a/core/shared/platform/alios/alios_platform.c
+++ b/core/shared/platform/alios/alios_platform.c
@@ -69,3 +69,7 @@ os_mprotect(void *addr, size_t size, int prot)
 void
 os_dcache_flush()
 {}
+
+void
+os_icache_flush(void *start, size_t len)
+{}
--- a/core/shared/platform/alios/alios_thread.c
+++ b/core/shared/platform/alios/alios_thread.c
@@ -359,3 +359,7 @@ os_thread_get_stack_boundary()
     /* TODO: get alios stack boundary */
     return NULL;
 }
+
+void
+os_thread_jit_write_protect_np(bool enabled)
+{}
--- a/core/shared/platform/common/posix/posix_memmap.c
+++ b/core/shared/platform/common/posix/posix_memmap.c
@@ -5,6 +5,10 @@
 
 #include "platform_api_vmcore.h"
 
+#if (defined(__APPLE__) || defined(__MACH__)) && defined(__arm64__)
+#include <libkern/OSCacheControl.h>
+#endif
+
 #ifndef BH_ENABLE_TRACE_MMAP
 #define BH_ENABLE_TRACE_MMAP 0
 #endif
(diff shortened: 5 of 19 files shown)


cosmo0920 (commented):
Thanks for your work. I confirmed that it works!!

```console
% sw_vers
ProductName:		macOS
ProductVersion:		14.0
BuildVersion:		23A344
```

```console
% git fetch upstream pull/2618/head:pr-2618
% git checkout pr-2618
```

Build wamrc and basic samples with the PR patch.

```console
% ../../wamr-compiler/wamrc --size-level=3 -o ./out/wasm-apps/testapp.aot ./out/wasm-apps/testapp.wasm
```

```console
% file ./out/wasm-apps/testapp.aot
./out/wasm-apps/testapp.aot: data
```

```console
% ./out/basic -f out/wasm-apps/testapp.aot
calling into WASM function: generate_float
Native finished calling wasm function generate_float(), returned a float value: 102009.921875f
calling into WASM function: float_to_string
calling into native function: intToStr
calling into native function: get_pow
calling into native function: intToStr
Native finished calling wasm function: float_to_string, returned a formatted string: 102009.921
calling into WASM function: calculate
calling into native function: calculate_native, n=3, func1=1, func2=2
calling into WASM function: mul5,    mul5 return 15 
call func1 and return n1=15
calling into WASM function: mul7,    mul7 return 105 
call func2 and return n2=105
Native finished calling wasm function: calculate, return: 120
```

yamt:
> After reading the [docs](https://developer.apple.com/documentation/apple-silicon/porting-just-in-time-compilers-to-apple-silicon), I managed to use AOT on my M1 mac.
> 
> Issue: #1365

it seems inefficient to call pthread_jit_write_protect_np before every calls.
isn't it enough to enable write temporary, only for text relocation?

also, the doc you mentioned seems suggesting to invalidate icache. isn't it necessary?

eloparco:
> Thanks for your work.

Thanks for starting the investigation, you input was was very useful!

> it seems inefficient to call pthread_jit_write_protect_np before every calls.

Moved that to call it only before and after module loading

> isn't it enough to enable write temporary, only for text relocation?

Yes, I put it before and after `load` so that we don't need to call it in multiple places inside `load`

> also, the doc you mentioned seems suggesting to invalidate icache. isn't it necessary?

Added that now

core/iwasm/aot/aot_loader.c
```diff
@@ -3232,7 +3236,15 @@ aot_load_from_aot_file(const uint8 *buf, uint32 size, char *error_buf,
...
 
-    if (!load(buf, size, module, error_buf, error_buf_size)) {
+#if defined(__APPLE__) && defined(__arm64__)
+    pthread_jit_write_protect_np(false); // Make memory writable
+#endif
+    bool ret = load(buf, size, module, error_buf, error_buf_size);
+#if defined(__APPLE__) && defined(__arm64__)
+    pthread_jit_write_protect_np(true); // Make memory executable
+    sys_icache_invalidate(module->code, module->code_size);
+#endif
```
wenyongh, line 3246: We usually don't directly call the libc functions in core/iwasm, instead, we encapsulated them in the platform layer and call them. How about adding API in core/shared/platform, e.g. `void os_thread_jit_write_protect_np(bool enabled)`, in:

https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/shared/platform/include/platform_api_vmcore.h#L83

and:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/shared/platform/common/posix/posix_thread.c#L420

and call it here.

and how about wrappring sys_icache_invalidate to os_icache_flush? like os_dcache_flush:

https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/shared/platform/include/platform_api_vmcore.h#L144

https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/shared/platform/common/posix/posix_memmap.c#L253C1-L253C1

And had better set memory executable after ret is true. Reference code:

```C
#if defined(__APPLE__) && defined(__arm64__)
    os_thread_jit_write_protect_np(false); /* Make memory writable */
#endif

    if (!load(buf, size, module, error_buf, error_buf_size)) {
        aot_unload(module);
        return NULL;
    }

#if defined(__APPLE__) && defined(__arm64__)
    if (module->code) {
        os_thread_jit_write_protect_np(true); /* Make memory executable */
        os_icache_flush(module->code, module->code_size);
    }
#endif
```

core/iwasm/aot/arch/aot_reloc_aarch64.c
```diff
@@ -53,7 +53,12 @@ get_target_symbol_map(uint32 *sym_num)
     return target_sym_map;
 }
 
+#if defined(__APPLE__) && defined(__arm64__)
+#define BUILD_TARGET_AARCH64_DEFAULT "arm64"
```
wenyongh, line 57: There was similar change in https://github.com/bytecodealliance/wasm-micro-runtime/pull/2570, but the author hasn't addressed the comment yet, could we also apply its change?
Not sure whether `#if (defined(__APPLE__) || defined(__MACH__)) && defined(__arm64__)` is OK?

core/iwasm/aot/aot_loader.c
eloparco: Thanks, I should have known.
The ifdef should be inside the platform-specific code I think, and no need for `if (module->code)` since `load` would have returned false if that was NULL.
```c
os_thread_jit_write_protect_np(false); /* Make memory writable */
if (!load(buf, size, module, error_buf, error_buf_size)) {
    aot_unload(module);
    return NULL;
}
os_thread_jit_write_protect_np(true); /* Make memory executable */
os_icache_flush(module->code, module->code_size);
```

wenyongh: OK, `module->code` may be NULL if the module hasn't function bytecodes, but I think the pthread_jit_write_protect_np should have handle that case.

wenyongh (commented):
LGTM

PR #2619, opened 2023-10-06
Fix compilation of shift opcodes on x86_64 and i386 architectures

loganek:
This change fixes the case where the right parameter of shift operator is negative, specifically, when both parameters of shift opcode are constants.
According to the WASM spec, negative value of the right parameter is allowed:
> ishr_u𝑁 (𝑖1, 𝑖2)
> • Let 𝑘 be 𝑖2 modulo 𝑁.
> • Return the result of shifting 𝑖1 right by 𝑘 bits, extended with 0 bits.

In general case, for x86 architecture wamrc compiler generates the following IR:
```
%shr_u = lshr i32 %"local0#", %"local1#"
```
Which (I guess, haven't verified it) is then translated to:
```
mov eax, %"local0#"
mov ecx, %"local1#"
shr eax, cl
```
so the assembly code already only uses the low 8 bits of the right parameter of shr_u. That's also why spec tests didn't fail, even though negative values were used.

However, when both parameters of the shift operator are constant, LLVM performs optimization and calculates the value of the shift opcode at compile time here: https://github.com/llvm/llvm-project/blob/release/15.x/llvm/include/llvm/IR/ConstantFolder.h#L59 As part of the compile-time evaluation, LLVM checks if the right parameter is valid (i.e. smaller bit width of the left parameter): https://github.com/llvm/llvm-project/blob/release/15.x/llvm/lib/IR/ConstantFold.cpp#L1163 Because the right parameter is much bigger than bit width of the left parameter in case of negative value, compiler generates poisoned constant.

Please note that the change does not affect the performance, as for constants the expression is still evaluated at compile time, i.e. for the code:

```
  (func (export "_start")
    i32.const -1
    i32.const -5
    i32.shr_u
    i32.const 31
    i32.ne
    if
      unreachable
    end
  )
```
the IR output is:
```
define void @"aot_func#0"(ptr %exec_env) {
func_begin:
  %aot_inst_addr = getelementptr inbounds ptr, ptr %exec_env, i32 2
  %aot_inst = load ptr, ptr %aot_inst_addr, align 8
  %cur_exception = getelementptr inbounds i8, ptr %aot_inst, i32 104
  %func_ptrs_offset = getelementptr inbounds i8, ptr %aot_inst, i32 40
  %func_ptrs_ptr = load ptr, ptr %func_ptrs_offset, align 8
  br label %func_end

func_end:                                         ; preds = %func_begin
  ret void
}
```

so we can see that shr_u was evaluated at compile time


tests/wamr-compiler/test_shift_negative_constants.wat
```diff
@@ -0,0 +1,40 @@
+;; Those tests verify if passing constant negative value
```
wenyongh, line 1: Had better add license header?

tests/wamr-test-suites/wamr-compiler-test-scripts/run_wamr_compiler_tests.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
wenyongh, line 1: Had better add license header?

tests/wamr-test-suites/wamr-compiler-test-scripts/run_wamr_compiler_tests.sh
wenyongh, line 1: Shall we rename the folder to `wamr-compiler-test-script`? I saw other two folders were named as spec-test-script and wasi-test-script.

core/iwasm/compilation/aot_emit_numberic.c
```diff
@@ -171,6 +171,15 @@
         right = shift_count_mask;                                      \
     } while (0)
 
+static bool
+is_shift_count_mask_needed(AOTCompContext *comp_ctx, LLVMValueRef left,
+                           LLVMValueRef right)
+{
+    return (strcmp(comp_ctx->target_arch, "x86_64") != 0
+            && strcmp(comp_ctx->target_arch, "i386") != 0)
+           || (LLVMIsAConstant(left) && LLVMIsAConstant(right));
```
lum1n0us, line 180: my suggestion is using `LLVMIsEfficientConstInt()` here. The new function is able to make sure both `left` and `right` are *constant int* only.

tests/wamr-compiler/test_shift_negative_constants.wat
```diff
@@ -0,0 +1,40 @@
...
+      (i32.const -4)
+    )
+  )
+
+  (func (export "_start")
+    call $i32_shr_u
+    call $i32_shr_s
+    call $i32_shl
+  )
+)
```
lum1n0us, line 40: how about using `assert_return` to check return values?

loganek: hmm, isn't that only available for WAST? Having said that, I can rename `$assert_eq` to `$assert_return` and move assertions to the `_start` function if you think it's clearer.

wenyongh (commented):
LGTM

lum1n0us (approved):
LGTM

Issue #2620, opened 2023-10-06
WAMRC terminator bug


arjunr2:
`wamrc` has a bug in the LLVM compilation flow when a compare instruction precedes a parametrized block. I have attached a sample WASM file below that demonstrates this, where the sign extension instruction is inserted after the terminator.

The compiler is built with default flags: `cmake -B build; make -C build`

[bug.tar.gz](https://github.com/bytecodealliance/wasm-micro-runtime/files/12835365/bug.tar.gz)

TianlongLiang:
Hi, https://github.com/bytecodealliance/wasm-micro-runtime/pull/2622 should fix this issue. Thanks for spotting this issue 😄

wenyongh:
Close this issue since it was fixed.

PR #2622, opened 2023-10-07
fix potential wamrc bug when translating block opcode

TianlongLiang:
Fix a wamrc bug mentioned in https://github.com/bytecodealliance/wasm-micro-runtime/issues/2620 (the sign extension instruction is inserted after the terminator when compare instruction precedes a block)


wenyongh (commented):
LGTM

PR #2624, opened 2023-10-08
Use another default pipeline when opt-level is 0

lum1n0us:
accroding to the description of `buildPerModuleDefaultPipeline()`, it is not allowed using `O0` level.

the issue will be triggered with llvm libraries with assertions by using `wamrc --disable-llvm-lot --opt-level 0 ...`

### DIFF
--- a/core/iwasm/compilation/aot_llvm_extra.cpp
+++ b/core/iwasm/compilation/aot_llvm_extra.cpp
@@ -343,18 +343,23 @@ aot_apply_llvm_new_pass_manager(AOTCompContext *comp_ctx, LLVMModuleRef module)
             ExitOnErr(PB.parsePassPipeline(MPM, comp_ctx->llvm_passes));
         }
 
-        if (!disable_llvm_lto) {
-            /* Apply LTO for AOT mode */
-            if (comp_ctx->comp_data->func_count >= 10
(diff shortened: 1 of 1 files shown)


core/iwasm/compilation/aot_llvm_extra.cpp
```diff
@@ -354,7 +354,9 @@ aot_apply_llvm_new_pass_manager(AOTCompContext *comp_ctx, LLVMModuleRef module)
                 MPM.addPass(PB.buildLTODefaultPipeline(OL, NULL));
         }
         else {
-            MPM.addPass(PB.buildPerModuleDefaultPipeline(OL));
+            MPM.addPass(OL == OptimizationLevel::O0
+                            ? PB.buildO0DefaultPipeline(OL)
+                            : PB.buildPerModuleDefaultPipeline(OL));
```
wenyongh, line 359: The above buildLTOPreLinkDefaultPipeline/buildLTODefaultPipeline also cannot be used for O0, could we also fix it, seems it can be:
```C
            if (OL == OptimizationLevel::O0)
                MPM.addPass(PB.buildO0DefaultPipeline(OL, true));
            else if (comp_ctx->comp_data->func_count >= 10
                     || comp_ctx->enable_llvm_pgo || comp_ctx->use_prof_file)
                ..
            else
```

wenyongh (commented):
LGTM

PR #2627, opened 2023-10-09
Fix AOT shift operations for indirect constants

loganek:
When doing more investigations related to this: https://github.com/bytecodealliance/wasm-micro-runtime/pull/2619 I found that in some scenarios the constant might not be directly available to the LLVM IR builder, e.g.:
```
  (func $const_ret (result i32)
    i32.const -5
  )

  (func $foo
    (i32.shr_u (i32.const -1) (call $const_ret))
    (i32.const 31)
  )
```
In that case, the right parameter to `i32.shr_u` is not constant, therefore the `SHIFT_COUNT_MASK` isn't applied. However, when the optimization is enabled (`--opt-level` is 2 or 3), the optimization passes resolve the call into constant, and that constant is poisoned, causing the compiler to resolve the whole function to an exception:

```
; Function Attrs: mustprogress nofree norecurse nosync nounwind readnone willreturn
define i32 @"aot_func#0"(ptr nocapture readonly %exec_env) local_unnamed_addr #0 {
func_end:
  ret i32 -5
}

define void @"aot_func#1"(ptr nocapture readonly %exec_env) local_unnamed_addr {
got_exception:
  %aot_inst_addr = getelementptr inbounds ptr, ptr %exec_env, i64 2
  %aot_inst1 = load i64, ptr %aot_inst_addr, align 8
  %0 = inttoptr i64 %aot_inst1 to ptr
  tail call void @aot_set_exception_with_id(ptr %0, i32 0)
  ret void
}

```
We'll enable the `SHIFT_COUNT_MASK` to avoid this unexpected behiavor.

I'm not sure if that's something can be fixed in LLVM itself given that this is very WASM specific behavior.

### DIFF
--- a/core/iwasm/compilation/aot_emit_numberic.c
+++ b/core/iwasm/compilation/aot_emit_numberic.c
@@ -171,15 +171,6 @@
         right = shift_count_mask;                                      \
     } while (0)
 
-static bool
-is_shift_count_mask_needed(AOTCompContext *comp_ctx, LLVMValueRef left,
-                           LLVMValueRef right)
-{
-    return (strcmp(comp_ctx->target_arch, "x86_64") != 0
-            && strcmp(comp_ctx->target_arch, "i386") != 0)
-           || (LLVMIsEfficientConstInt(left) && LLVMIsEfficientConstInt(right));
-}
-
 /* Call llvm constrained floating-point intrinsic */
 static LLVMValueRef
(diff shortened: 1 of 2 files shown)


wenyongh:
@loganek thanks a lot, let's merge this PR first to fix the issue, and check whether we can apply further optimization in the future.

PR #2628, opened 2023-10-09
core/iwasm: Support mapped file system access on non-libuv WASI

fredldotme:
Similar to the for-loop above it, this patch just adds another one iterating through map_dir with the ability to set the guest path to a different target. It follows the format "guest-path::host-path".

### DIFF
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -3046,7 +3046,7 @@ wasm_runtime_init_wasi(WASMModuleInstanceCommon *module_inst,
         if (mapping_copy != mapping_copy_buf)
             wasm_runtime_free(mapping_copy);
 
-        if (!map_mapped && !map_host) {
+        if (!map_mapped || !map_host) {
             if (error_buf)
                 snprintf(error_buf, error_buf_size,
                          "error while pre-opening mapped directory %s: "
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -3029,7 +3029,7 @@ wasm_runtime_init_wasi(WASMModuleInstanceCommon *module_inst,
         char *map_mapped = NULL, *map_host = NULL;
         const unsigned long max_len = strlen(map_dir_list[i]) * 2 + 3;
 
-        // Allocation limit for runtime environments with reduced stack size
+        /* Allocation limit for runtime environments with reduced stack size */
         if (max_len > 256) {
             if (!(mapping_copy = wasm_runtime_malloc(max_len))) {
                 snprintf(error_buf, error_buf_size,
@@ -3038,12 +3038,9 @@ wasm_runtime_init_wasi(WASMModuleInstanceCommon *module_inst,
             }
         }
 
-        strncpy(mapping_copy, map_dir_list[i], strlen(map_dir_list[i]));
+        strncpy(mapping_copy, map_dir_list[i], strlen(map_dir_list[i]) + 1);
         map_mapped = strtok(mapping_copy, "::");
         map_host = strtok(NULL, "::");
-
-        if (mapping_copy != mapping_copy_buf)
-            wasm_runtime_free(mapping_copy);
 
         if (!map_mapped || !map_host) {
             if (error_buf)
@@ -3059,6 +3056,8 @@ wasm_runtime_init_wasi(WASMModuleInstanceCommon *module_inst,
                 snprintf(error_buf, error_buf_size,
                          "error while pre-opening mapped directory %s: %d\n",
                          map_host, errno);
+            if (mapping_copy != mapping_copy_buf)
+                wasm_runtime_free(mapping_copy);
             goto fail;
         }
 
@@ -3068,6 +3067,8 @@ wasm_runtime_init_wasi(WASMModuleInstanceCommon *module_inst,
                 snprintf(error_buf, error_buf_size,
                          "error while pre-opening mapped directory %s: %d\n",
                          map_host, errno);
+            if (mapping_copy != mapping_copy_buf)
+                wasm_runtime_free(mapping_copy);
             goto fail;
         }
 
@@ -3078,8 +3079,13 @@ wasm_runtime_init_wasi(WASMModuleInstanceCommon *module_inst,
                          "error while pre-opening mapped directory %s: "
                          "insertion failed\n",
                          dir_list[i]);
+            if (mapping_copy != mapping_copy_buf)
+                wasm_runtime_free(mapping_copy);
             goto fail;
         }
+
+        if (mapping_copy != mapping_copy_buf)
+            wasm_runtime_free(mapping_copy);
     }
 
     /* addr_pool(textual) -> apool */
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -3047,7 +3047,12 @@ wasm_runtime_init_wasi(WASMModuleInstanceCommon *module_inst,
             wasm_runtime_free(mapping_copy);
 
         if (!map_mapped && !map_host) {
-            continue;
+            if (error_buf)
+                snprintf(error_buf, error_buf_size,
+                         "error while pre-opening mapped directory %s: "
+                         "invalid map\n",
+                         map_host);
+            goto fail;
         }
 
         path = realpath(map_host, resolved_path);
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -3047,6 +3047,8 @@ wasm_runtime_init_wasi(WASMModuleInstanceCommon *module_inst,
                 snprintf(error_buf, error_buf_size,
                          "error while pre-opening mapped directory: "
                          "invalid map\n");
+            if (mapping_copy != mapping_copy_buf)
+                wasm_runtime_free(mapping_copy);
             goto fail;
         }
 
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -3016,6 +3016,40 @@ wasm_runtime_init_wasi(WASMModuleInstanceCommon *module_inst,
         fd_prestats_insert(prestats, dir_list[i], wasm_fd);
     }
 
+    for (i = 0; i < map_dir_count; i++, wasm_fd++) {
+        char mapping_copy[(PATH_MAX * 2) + 3];
+        strncpy(mapping_copy, map_dir_list[i], strlen(map_dir_list[i]));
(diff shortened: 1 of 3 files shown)


core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3016,6 +3016,50 @@ wasm_runtime_init_wasi(WASMModuleInstanceCommon *module_inst,
...
+        if (raw_fd == -1) {
+            if (error_buf)
+                snprintf(error_buf, error_buf_size,
+                         "error while pre-opening directory %s: %d\n", map_host,
+                         errno);
+            goto fail;
+        }
+
+        fd_table_insert_existing(curfds, wasm_fd, raw_fd);
+        fd_prestats_insert(prestats, map_mapped, wasm_fd);
```
wenyongh, line 3060: Should check return values for these two functions, could you also help fix the original code in L3015, L3016?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3016,6 +3016,50 @@ wasm_runtime_init_wasi(WASMModuleInstanceCommon *module_inst,
         fd_prestats_insert(prestats, dir_list[i], wasm_fd);
     }
 
+    for (i = 0; i < map_dir_count; i++, wasm_fd++) {
+        char mapping_copy[(PATH_MAX * 2) + 3];
```
wenyongh, line 3020: A concern is that the array size is very large (should be 4096 * 2 + 3), some platforms' stack may be not enough. Could we define a smaller array buffer and use it when it is enough, and allocate memory when needed? Like:
```C
    char mapping_copy_buf[256];
    char mapping_copy = mapping_copy_buf;

    if (strlen(map_dir_list) * 2 + 3 > 256) {
        if (!(mapping_copy = wasm_runtime_malloc(strlen(map_dir_list) * 2 + 3)))) {
            handle error;
        }
    }
    ...
    if (mapping_copy != mapping_copy_buf)
        wasm_runtime_free(mapping_copy);
```

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3016,6 +3016,50 @@ wasm_runtime_init_wasi(WASMModuleInstanceCommon *module_inst,
...
+        strncpy(mapping_copy, map_dir_list[i], strlen(map_dir_list[i]));
+        mapping_copy[(PATH_MAX * 2) + 2] = '\0';
+
+        const char *map_mapped = strtok(mapping_copy, "::");
+        const char *map_host = strtok(NULL, "::");
```
wenyongh, line 3025: This PR only handles `wasm_runtime_init_wasi` of libc wasi version, would you also handle the version of libc uvwasi? And it didn't add the related argument for iwasm, would you enable it in this PR or in the future? For example:
```bash
iwasm --map-dir=<guest_dir::host_dir>
```
And we can add the argument multiple times: `--map-dir=<guest_dir1::host_dir1> --map-dir=<guest_dir2::host_dir2>`.

fredldotme:
@wenyongh thanks for looking into it, I intend to add the necessary changes soon. I'm currently in the process of another WAMR-based IDE release (this time on macOS) hence it might take a few days until I get to it.

I'm embedding WAMR inside the IDE using a plugin mechanism, and it loads specific implementations of the same interface wherever it sees fit (debugger-enabled interpreter build during debugging, "fast interp" or maybe even JIT/AOT otherwise).
I have only tested the current set of changes inside this confined environment, but I don't have a libuv-enabled build planned atm (no idea of differences and benefits). What would you like to see happen with libuv, doesn't it already pass down the map_dir list to libuv for it to do the right thing?

core/iwasm/common/wasm_runtime_common.c
fredldotme: Will do

core/iwasm/common/wasm_runtime_common.c
fredldotme: I intend to do it similarly to the current `--dir` argument which includes multiple arguments for the same list IIRC.

wenyongh:
> @wenyongh thanks for looking into it, I intend to add the necessary changes soon. I'm currently in the process of another WAMR-based IDE release (this time on macOS) hence it might take a few days until I get to it.
> 
> I'm embedding WAMR inside the IDE using a plugin mechanism, and it loads specific implementations of the same interface wherever it sees fit (debugger-enabled interpreter build during debugging, "fast interp" or maybe even JIT/AOT otherwise). I have only tested the current set of changes inside this confined environment, but I don't have a libuv-enabled build planned atm (no idea of differences and benefits). What would you like to see happen with libuv, doesn't it already pass down the map_dir list to libuv for it to do the right thing?

Thanks, I checked the code of init uvwasi, it seems that map_dir_list[i] is used as a mapped dir of dir_list[i], the format of map_dir_list[i] is not `<guest_dir::host_dir>`:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/common/wasm_runtime_common.c#L3182-L3183
If to apply similar changes like this PR, seems the `init_options.preopenc` should be `dir_count + map_dir_count`, and for `init_options.preopens[0 to dir_count-1]`, its `real_path` and `mapped_path` are the same (=dir_list[i]); for `init_options.preopens[dir_count to dir_count+map_dir_count-1]`, its `real_path` is host_dir in map_dir_list[i], and its`mapped_path` is guest_dir in map_dir_list[i].

I think you can just leave it unchanged, we can open another PR to apply similar changes for libc-uvwasi after this PR is merged.

core/iwasm/common/wasm_runtime_common.c
wenyongh: Do you mean somewhat like `iwasm --dir=<dir1,dir2,..>`? No, currently it is `iwasm --dir=<dir1> --dir=<dir2> ...`, like `iwasm --env="key1=value1" --env="key2=value2"`, you can see the help of iwasm. So I think for the map dir, it should be like `iwasm --map-dir=<guest_dir1::host_dir1> --map-dir=<guest_dir2::host_dir2>`?

And could you help change the comment of wasm_runtime_set_wasi_args_ex in wasm_export.h, at least to mention the format of map dir:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/include/wasm_export.h#L425

core/iwasm/common/wasm_runtime_common.c
fredldotme: Done.

core/iwasm/common/wasm_runtime_common.c
fredldotme: Done.

core/iwasm/common/wasm_runtime_common.c
fredldotme: Done.

fredldotme:
In case this doesn't cause any test build failures I consider this completed.

fredldotme:
Fixed up minor issues and tested the functionality with this Python 3 build: https://github.com/vmware-labs/webassembly-language-runtimes/releases/tag/python%2F3.11.4%2B20230714-11be424

Using: `./iwasm --map-dir=/::<path-to-python-root>`

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3012,8 +3012,74 @@ wasm_runtime_init_wasi(WASMModuleInstanceCommon *module_inst,
...
+        // Allocation limit for runtime environments with reduced stack size
+        if (max_len > 256) {
+            if (!(mapping_copy = wasm_runtime_malloc(max_len))) {
+                snprintf(error_buf, error_buf_size,
+                         "error while allocating for directory mapping\n");
+                goto fail;
+            }
+        }
+
+        strncpy(mapping_copy, map_dir_list[i], strlen(map_dir_list[i]));
```
wenyongh, line 3041: The mapping_copy may be not null-terminated after copy, see manual of strncpy:
`Warning: If there is no null byte among the first n  bytes
       of src, the string placed in dest will not be null-terminated.`

Had better strcpy or `strncpy(mapping_copy, map_dir_list[i], strlen(map_dir_list[i]) + 1)`?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3012,8 +3012,74 @@ wasm_runtime_init_wasi(WASMModuleInstanceCommon *module_inst,
...
+        map_mapped = strtok(mapping_copy, "::");
+        map_host = strtok(NULL, "::");
+
+        if (mapping_copy != mapping_copy_buf)
+            wasm_runtime_free(mapping_copy);
```
wenyongh, line 3046: mapping_copy may be referred by map_mapped and map_host since strtok doesn't allocate new memory (if yes, we should free them later?). Should free mapping_copy after L3082 (and before some `goto fail` sentences)?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3012,8 +3012,74 @@ wasm_runtime_init_wasi(WASMModuleInstanceCommon *module_inst,
...
+        }
+    }
+
+    for (i = 0; i < map_dir_count; i++, wasm_fd++) {
+        char mapping_copy_buf[256];
+        char *mapping_copy = mapping_copy_buf;
+        char *map_mapped = NULL, *map_host = NULL;
+        const unsigned long max_len = strlen(map_dir_list[i]) * 2 + 3;
+
+        // Allocation limit for runtime environments with reduced stack size
```
wenyongh, line 3032: Had better use `/* .. */` to comment the code.

core/iwasm/common/wasm_runtime_common.c
fredldotme: Thanks for correcting, done.

core/iwasm/common/wasm_runtime_common.c
fredldotme: Done.

core/iwasm/common/wasm_runtime_common.c
fredldotme: Done.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3012,8 +3012,80 @@ wasm_runtime_init_wasi(WASMModuleInstanceCommon *module_inst,
...
+        strncpy(mapping_copy, map_dir_list[i], strlen(map_dir_list[i]) + 1);
+        map_mapped = strtok(mapping_copy, "::");
+        map_host = strtok(NULL, "::");
+
+        if (!map_mapped || !map_host) {
+            if (error_buf)
+                snprintf(error_buf, error_buf_size,
+                         "error while pre-opening mapped directory: "
+                         "invalid map\n");
+            goto fail;
```
wenyongh, line 3050: Should `if (mapping_copy != mapping_copy_buf) wasm_runtime_free(mapping_copy)`  before `goto fail`.

wenyongh (approved):
LGTM

PR #2634, opened 2023-10-11
Enable MASM automatically in runtime_lib.cmake

zoraaver:
MASM (Microsoft Macro Assembler) needs to be enabled to successfully compile .asm files on Windows. Without enabling this language automatically in runtime_lib.cmake, Windows cmake projects using WAMR get hard-to-diagnose linker errors about missing symbols as the assembly file is not compiled at all.

### DIFF
--- a/build-scripts/runtime_lib.cmake
+++ b/build-scripts/runtime_lib.cmake
@@ -165,7 +165,11 @@ file (GLOB header
 )
 LIST (APPEND RUNTIME_LIB_HEADER_LIST ${header})
 
-enable_language (ASM)
+if (WAMR_BUILD_PLATFORM STREQUAL "windows")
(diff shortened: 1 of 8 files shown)


build-scripts/runtime_lib.cmake
```diff
@@ -165,7 +165,11 @@ file (GLOB header
 )
 LIST (APPEND RUNTIME_LIB_HEADER_LIST ${header})
 
-enable_language (ASM)
+if (MSVC)
```
wenyongh, line 168: Had better use `if (WAMR_BUILD_PLATFORM STREQUAL "windows")` to control the code like the original code in samples?  Not sure whether other windows compiler like `MINGW` also needs to enable ASM_MASM.

PR #2637, opened 2023-10-12
Implementation for the wasi_clock_time/res

fadumina1:


core/shared/platform/windows/win_clock.c
```diff
@@ -0,0 +1,147 @@
...
+        }
+        case BH_CLOCK_ID_REALTIME:
+        case BH_CLOCK_ID_PROCESS_CPUTIME_ID:
+        case BH_CLOCK_ID_THREAD_CPUTIME_ID:
+        {
+            PULONG maximum_time;
+            PULONG minimum_time;
+            PULONG current_time;
+            NTSTATUS
+            status = NtQueryTimerResolution(&maximum_time, &minimum_time,
```
zoraaver, line 61: This is an undocumented, internal API so it might be worth loading it at runtime (so we don't end up calling it if it's unavailable)? Also I'm not sure if it would be available on UWP.

core/shared/platform/windows/win_clock.c
```diff
@@ -0,0 +1,147 @@
...
+    return (uint64)frequency.QuadPart;
+}
+
+// The implementation below dervies from the following source:
+// https://github.com/WasmEdge/WasmEdge/blob/b70f48c42922ce5ee7730054b6ac0b1615176285/lib/host/wasi/win.h#L210
+static uint64
+from_file_time_to_wasi_timestamp(FILETIME filetime)
+{
+    static const uint64 ntto_unix_epoch =
+        134774 * 86400 * NANOSECONDS_PER_SECOND;
```
zoraaver, line 26: I think you'll need to be careful here about integer overflow - probably add `ull` suffix to every number in the calculation?

core/shared/platform/windows/win_clock.c
```diff
@@ -0,0 +1,147 @@
...
+}
+
+int
+os_clock_res_get(bh_clock_id_t clock_id, uint64 *resolution)
+{
+    switch (clock_id) {
+        case BH_CLOCK_ID_MONOTONIC:
+        {
+            const uint64 result = (uint64)NANOSECONDS_PER_SECOND
+                                  / calculate_monotonic_clock_frequency();
```
zoraaver, line 49: I'm assuming the frequency won't change for the lifetime of the app - we could calculate it once and store the result in a static variable so we don't make unnecessary system calls.

core/shared/platform/windows/win_clock.c
```diff
@@ -0,0 +1,147 @@
...
+
+        case BH_CLOCK_ID_THREAD_CPUTIME_ID:
+        {
+            FILETIME creation_time;
+            FILETIME exit_time;
+            FILETIME kernal_time;
+            FILETIME user_time;
+
+            if (!GetProcessTimes(GetCurrentThread(), &creation_time, &exit_time,
+                                 &kernal_time, &user_time)) {
```
zoraaver, line 133: nit: [typo] `kernel_time`

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -172,6 +172,10 @@ convert_errno(int error)
 #undef X
     return code;
 }
+#ifndef BH_PLATFORM_WINDOWS
+bool
+convert_clockid(__wasi_clockid_t in, clockid_t *out);
```
loganek, line 177: worth adding a TODO here and a brief explanation - I presume we only keep it here so poll_oneoff compiles. Because we plan to refactor poll_oneoff, we'll need to remove it as part of that change.

core/shared/platform/include/platform_common.h
```diff
@@ -118,6 +125,16 @@ typedef void *(*thread_start_routine_t)(void *);
 #define bh_socket_t int
 #endif
 
+/* Get current resolution of clock id returns BHT_OK on success and BHT_ERROR on
+ * failure */
```
loganek, line 129:
```suggestion
/**
 * Get a resolution of the clock
 *
 * @param clock_id clock identifier
 * @param resolution output variable to store the clock resolution
 * @return BHT_OK if success; otherwise, BHT_ERROR
 */
```
please do the same for `os_clock_time_get`

core/shared/platform/windows/win_clock.c
```diff
@@ -0,0 +1,147 @@
...
+                            .HighPart = filetime.dwHighDateTime };
+
+    return temp.QuadPart - ntto_unix_epoch;
+}
+
+uint64
+current_value_of_peformance_counter()
+{
+    LARGE_INTEGER counter;
+    QueryPerformanceCounter(&counter);
```
loganek, line 38: I think we should not ignore the return value here; errors should be propagated all the way up to the user.

core/shared/utils/bh_time.c
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright (C) 2023 Amazon Inc.  All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+
+#include "bh_time.h"
+
+uint64
+convert_timespec(const struct timespec *ts)
```
loganek, line 9: it'd be good to mention what we convert it to, e.g.
```suggestion
convert_timespec_to_nanoseconds(const struct timespec *ts)
```

product-mini/platforms/windows/CMakeLists.txt
```diff
@@ -154,3 +154,13 @@ target_link_libraries (libiwasm ${LLVM_AVAILABLE_LIBS} ${UV_A_LIBS})
...
   target_link_libraries (libiwasm ws2_32)
 endif ()
+
+if (WIN32)
+  target_link_libraries(libiwasm ntdll)
+endif()
+
+if (WIN32)
+  target_link_libraries(iwasm ntdll)
+endif()
```
loganek, line 164: Could we merge those two into one?
```suggestion
if (WIN32)
  target_link_libraries(libiwasm ntdll)
  target_link_libraries(iwasm ntdll)
endif()
```

product-mini/platforms/windows/CMakeLists.txt
```diff
@@ -154,3 +154,13 @@ target_link_libraries (libiwasm ${LLVM_AVAILABLE_LIBS} ${UV_A_LIBS})
...
+
+
```
loganek, line 166: Unnecessary second empty line
```suggestion
```

core/shared/platform/windows/win_clock.c
zoraaver: nvm, from what I can tell ntdll is always loaded by the process and even though this is undocumented it seems to be used quite ubiquitously anyway.

core/shared/platform/include/platform_api_vmcore.h
```diff
@@ -66,6 +66,14 @@ os_vprintf(const char *format, va_list ap);
 uint64
 os_time_get_boot_microsecond(void);
 
+/* Get current resolution of clock id */
+int
+os_clock_res_get(bh_clock_id_t clock_id, uint64 *resolution);
```
loganek, line 71: we already have the declaration in platform_common.h so the declaration should probably be removed from this file.

core/shared/platform/include/platform_common.h
```diff
@@ -37,6 +37,13 @@ extern "C" {
 #define BH_TIME_T_MAX LONG_MAX
 #endif
 
+typedef enum {
+    BH_CLOCK_ID_REALTIME,
+    BH_CLOCK_ID_MONOTONIC,
+    BH_CLOCK_ID_PROCESS_CPUTIME_ID,
+    BH_CLOCK_ID_THREAD_CPUTIME_ID
+} bh_clock_id_t;
```
loganek, line 45: Feels like the enum fits better in the `bh_time.h` (or `bh_clock.h`) file.

core/shared/platform/include/platform_common.h
```diff
@@ -118,6 +125,26 @@ typedef void *(*thread_start_routine_t)(void *);
...
 
+/**
+ * Get a resolution of the clock
+ *
+ * @param clock_id clock identifier
+ * @param resolution output variable to store the clock resolution
+ * @return BHT_OK if success; otherwise, BHT_ERROR
+ */
+int
+os_clock_res_get(bh_clock_id_t clock_id, uint64 *resolution);
```
loganek, line 136: I'm actually wondering if we should move those two declarations to `platform_api_extension.h` file.

core/shared/platform/windows/win_clock.c
```diff
@@ -0,0 +1,174 @@
...
+ */
+
+#include "platform_api_vmcore.h"
+#include <winternl.h>
+
+#define NANOSECONDS_PER_SECOND 1000000000ULL
+#define NANOSECONDS_PER_TICK 100
+
+static int
+calculate_monotonic_clock_frequency(uint64 *outfrequency)
```
loganek, line 13:
```suggestion
calculate_monotonic_clock_frequency(uint64 *out_frequency)
```
please use sneak case.

core/shared/platform/windows/win_clock.c
```diff
@@ -0,0 +1,174 @@
...
+        134774ULL * 86400ULL * NANOSECONDS_PER_SECOND;
+
+    ULARGE_INTEGER temp = { .LowPart = filetime.dwLowDateTime,
+                            .HighPart = filetime.dwHighDateTime };
+
+    return temp.QuadPart - ntto_unix_epoch;
+}
+
+uint64
+get_performance_counter_value(uint64 *outcounter)
```
loganek, line 40:
```suggestion
get_performance_counter_value(uint64 *out_counter)
```

core/shared/platform/windows/win_clock.c
```diff
@@ -0,0 +1,174 @@
...
+    static const uint64 ntto_unix_epoch =
+        134774ULL * 86400ULL * NANOSECONDS_PER_SECOND;
+
+    ULARGE_INTEGER temp = { .LowPart = filetime.dwLowDateTime,
+                            .HighPart = filetime.dwHighDateTime };
+
+    return temp.QuadPart - ntto_unix_epoch;
+}
+
+uint64
```
loganek, line 39:
```suggestion
int
```
Should that be `int` instead? We return `BHT_ERROR`/`BHT_OK` here.

core/shared/platform/windows/win_clock.c
```diff
@@ -0,0 +1,174 @@
...
+    }
+}
+
+int
+os_clock_time_get(bh_clock_id_t clock_id, uint64 precision, uint64 *time)
+{
+    switch (clock_id) {
+        case BH_CLOCK_ID_REALTIME:
+        {
+            FILETIME sysNow;
```
loganek, line 97:
```suggestion
            FILETIME sys_now;
```
please follow sneak_case

product-mini/platforms/windows/CMakeLists.txt
```diff
@@ -137,7 +137,7 @@ add_executable (iwasm main.c ${UNCOMMON_SHARED_SOURCE})
 
 install (TARGETS iwasm DESTINATION bin)
 
-target_link_libraries (iwasm vmlib ${LLVM_AVAILABLE_LIBS} ${UV_A_LIBS})
+target_link_libraries (iwasm vmlib  ${LLVM_AVAILABLE_LIBS} ${UV_A_LIBS})
```
loganek, line 140:
```suggestion
target_link_libraries (iwasm vmlib ${LLVM_AVAILABLE_LIBS} ${UV_A_LIBS})
```
whitespace change

loganek (changes requested):

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -314,41 +283,50 @@ wasi_addr_ip_to_bh_ip_addr_buffer(__wasi_addr_ip_t *addr,
...
+    if (os_clock_res_get(clock_id, resolution) != BHT_OK)
         return convert_errno(errno);
-    *resolution = convert_timespec(&ts);
-    return 0;
-#endif
+    return __WASI_ESUCCESS;
 }
 
 __wasi_errno_t
 wasmtime_ssp_clock_time_get(__wasi_clockid_t clock_id,
```
zoraaver, line 320: Sorry for the late comment, did we consider implementing these functions (wasmtime_ssp_clock_time_get and wasmtime_ssp_clock_res_get) directly in each platform? Unlike the filesystem or socket functions, they don't  contain any logic to lookup a fd object from the fd table so it seems they are very small wrapper around the platform implementation anyway.

Also, since `os_clock_time_get` and `os_clock_res_get` are only used in the WASI libc implementation, we could probably define the signature for these functions in terms of WASI types (`__wasi_clockid_t` etc.) to avoid some boilerplate. What do you think?

loganek: I think that's not been done mainly due to the circular dependency as discussed in  https://github.com/bytecodealliance/wasm-micro-runtime/pull/2585#discussion_r1363526897 

I think we can merge this as it is now, but once your change @zoraaver is merged, we'll make an update to this code and simplify it if possible. How does it sound to you?

loganek:
@wenyongh would you be able to have a look at this PR?

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
zoraaver: Agree, I think the PR is in a good state to merge now and we can simplify the code later once #2585  is merged.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -172,6 +172,15 @@ convert_errno(int error)
...
     return code;
 }
+#ifndef BH_PLATFORM_WINDOWS
+/*
+ * convert_clockid is a plceholder in this file so that poll_oneoff complies,
+ * the implementation has been refactored to posix_clock.c. In the near future
+ * poll_oneoff will be refactored and the below placeholder will be removed.
+ */
+bool
+convert_clockid(__wasi_clockid_t in, clockid_t *out);
```
wenyongh, line 182: How about renaming to wasi_clockid_to_clockid?

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -314,41 +283,50 @@ wasi_addr_ip_to_bh_ip_addr_buffer(__wasi_addr_ip_t *addr,
     }
 }
 
+bool
+convert_wasi_clock_id_to_bh_clock_id(__wasi_clockid_t in, bh_clock_id_t *out)
```
wenyongh, line 287: How about renaming to wasi_clockid_to_bh_clockid?

core/shared/platform/common/posix/posix_clock.c
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright (C) 2023 Amazon Inc.  All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+
+#include "platform_api_vmcore.h"
+#include "bh_platform.h"
+
+bool
+convert_clockid(bh_clock_id_t in, clockid_t *out)
```
wenyongh, line 10: How about renaming to bh_clockid_to_clockid? And found that there is a different implementation in `core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c` with different function prototype?

core/shared/utils/bh_time.c
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright (C) 2023 Amazon Inc.  All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+
+#include "bh_time.h"
+
+uint64
+convert_timespec_to_nanoseconds(const struct timespec *ts)
```
wenyongh, line 9: How about renaming to bh_timespec_to_nanoseconds? Adding bh_ prefix is because that APIs in core/shared/utils usually start with it.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -172,6 +172,38 @@ convert_errno(int error)
 #undef X
     return code;
 }
+#ifndef BH_PLATFORM_WINDOWS
+/*
+ * wasi_clockid_to_clockid is a plceholder in this file so that poll_oneoff complies,
```
wenyongh, line 177: Should be `placeholder`

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -172,6 +172,38 @@ convert_errno(int error)
...
+ * the implementation has been refactored to posix_clock.c. In the near future
+ * poll_oneoff will be refactored and the below placeholder will be removed.
+ */
+bool
+wasi_clockid_to_clockid(__wasi_clockid_t in, clockid_t* out) 
+    {
```
wenyongh, line 183: Invalid coding style, could you use clang-format-12 to format the code?
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/6573605221/job/17856981299

core/shared/platform/windows/win_clock.c
```diff
@@ -0,0 +1,174 @@
...
+{
+    LARGE_INTEGER counter;
+    if (!QueryPerformanceCounter(&counter)) {
+        return BHT_ERROR;
+    }
+    else {
+        *out_counter = counter.QuadPart;
+        return BHT_OK;
+    }
+
```
wenyongh, line 50: Had better remove unused line?

core/shared/platform/windows/win_clock.c
```diff
@@ -0,0 +1,174 @@
...
+                *time = counter * NANOSECONDS_PER_SECOND / frequency;
+            }
+            else {
+                uint64 seconds = counter / frequency;
+                uint64 fractions = counter % frequency;
+                *time = seconds * NANOSECONDS_PER_SECOND
+                        + (fractions * NANOSECONDS_PER_SECOND) / frequency;
+            }
+        }
+            return BHT_OK;
```
wenyongh, line 132: Had better put this line before L131

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -172,6 +172,39 @@ convert_errno(int error)
 #undef X
     return code;
 }
+#ifndef BH_PLATFORM_WINDOWS
+/*
+ * wasi_clockid_to_clockid is a plceholder in this file so that poll_oneoff
+ * complies, the implementation has been refactored to posix_clock.c. In the
+ * near future poll_oneoff will be refactored and the below placeholder will be
+ * removed.
```
loganek, line 180: I guess this comment is no longer relevant?

core/shared/platform/windows/win_clock.c
```diff
@@ -0,0 +1,171 @@
...
```
zoraaver, line 129: The braces for each case in the switch statement aren't strictly necessary but it looks a bit weird to have the return statement outside the block - maybe move it a line above so that it's inside the block?

core/shared/platform/windows/win_clock.c
```diff
@@ -0,0 +1,171 @@
...
+
+int
+os_clock_res_get(bh_clock_id_t clock_id, uint64 *resolution)
+{
+    switch (clock_id) {
+        case BH_CLOCK_ID_MONOTONIC:
+        {
+            uint64 frequency;
+            if (calculate_monotonic_clock_frequency(&frequency) == BHT_ERROR) {
+
```
zoraaver, line 60: In this file, there are blank lines after some of the if statements and before some of the 'case' blocks in the switch statement. I think we can remove them.

core/shared/platform/windows/win_clock.c
```diff
@@ -0,0 +1,171 @@
...
+    if (!QueryPerformanceFrequency(&frequency)) {
+        return BHT_ERROR;
+    }
+    else {
+        *out_frequency = (uint64)frequency.QuadPart;
+        return BHT_OK;
+    }
+}
+
+// The implementation below dervies from the following source:
```
zoraaver, line 25: typo 'derives'

core/shared/platform/common/posix/posix_clock.c
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright (C) 2023 Amazon Inc.  All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+
+#include "platform_api_vmcore.h"
+#include "bh_platform.h"
```
zoraaver, line 7: We can probably remove this header include now

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -94,22 +94,9 @@ ns_lookup_list_search(char **list, const char *host)
...
-        return UINT64_MAX;
-    return (__wasi_timestamp_t)ts->tv_sec * 1000000000
-           + (__wasi_timestamp_t)ts->tv_nsec;
-}
-
-// Converts a CloudABI clock identifier to a POSIX clock identifier.
 #ifndef BH_PLATFORM_WINDOWS
-static bool
-convert_clockid(__wasi_clockid_t in, clockid_t *out)
+bool
```
zoraaver, line 98: We can keep the static here I think

core/shared/platform/common/posix/posix_clock.c
```diff
@@ -0,0 +1,71 @@
...
+            return false;
+    }
+}
+
+static uint64
+timespec_to_nanoseconds(const struct timespec *ts)
+{
+    if (ts->tv_sec < 0)
+        return 0;
+    if ((uint64)ts->tv_sec >= UINT64_MAX / 1000000000)
```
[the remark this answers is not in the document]
zoraaver: Or define a macro `#define NANOSECONDS_PER_SECOND 1000000000ULL` like in `win_clock.c`

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -210,41 +197,50 @@ wasi_addr_ip_to_bh_ip_addr_buffer(__wasi_addr_ip_t *addr,
     }
 }
 
+bool
```
zoraaver, line 200: we can make this static

core/shared/platform/common/posix/posix_clock.c
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright (C) 2023 Amazon Inc.  All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+
+#include "platform_api_vmcore.h"
+#include "bh_platform.h"
+
+bool
```
wenyongh, line 9: Had better make this function static?

core/shared/platform/windows/win_clock.c
```diff
@@ -0,0 +1,171 @@
...
+// https://github.com/WasmEdge/WasmEdge/blob/b70f48c42922ce5ee7730054b6ac0b1615176285/lib/host/wasi/win.h#L210
+static uint64
+from_file_time_to_wasi_timestamp(FILETIME filetime)
```
wenyongh, line 28: How about renaming to `file_time_to_wasi_timestamp` like others?

core/shared/platform/windows/win_clock.c
```diff
@@ -0,0 +1,171 @@
...
+    static const uint64 ntto_unix_epoch =
+        134774ULL * 86400ULL * NANOSECONDS_PER_SECOND;
+
+    ULARGE_INTEGER temp = { .LowPart = filetime.dwLowDateTime,
+                            .HighPart = filetime.dwHighDateTime };
+
+    return temp.QuadPart - ntto_unix_epoch;
+}
+
+int
```
wenyongh, line 39: Had better make this function static?

zoraaver (approved):
LGTM

wenyongh (commented):
LGTM

loganek (approved):
Good job!

PR #2651, opened 2023-10-17
Support stringref proposal

yviansu:


core/iwasm/interpreter/wasm.h
```diff
@@ -970,7 +1039,7 @@ wasm_value_type_size_ex(uint8 value_type, bool gc_enabled)
     else if (value_type == VALUE_TYPE_V128)
         return sizeof(int64) * 2;
 #endif
-    else if (value_type >= (uint8)REF_TYPE_NULLREF /* 0x65 */
+    else if (value_type >= (uint8)REF_TYPE_STRINGVIEWITER /* 0x61 */
```
xujuntwt95329, line 1042: Should use macro to check if stringref enabled

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -3699,6 +3707,65 @@ load_start_section(const uint8 *buf, const uint8 *buf_end, WASMModule *module,
...
+                for (j = 0; j < string_length; j++) {
+                    *(str_content + j) = read_uint8(p);
+                }
+            }
+        }
+    }
+
+    if (p != p_end) {
+        set_error_buf(error_buf, error_buf_size, "section size mismatch");
+        return false;
```
xujuntwt95329, line 3757:
```suggestion
        goto fail;
```

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -5951,6 +6052,60 @@ wasm_loader_find_block_addr(WASMExecEnv *exec_env, BlockAddr *block_addr_cache,
                     case WASM_OP_EXTERN_EXTERNALIZE:
                         break;
 
+                    case WASM_OP_STRING_NEW_UTF8:
```
xujuntwt95329, line 6055: Should add macro check

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -11995,6 +12200,215 @@ wasm_loader_prepare_bytecode(WASMModule *module, WASMFunction *func,
...
+                    case WASM_OP_STRING_ENCODE_LOSSY_UTF8_ARRAY:
+                    case WASM_OP_STRING_ENCODE_WTF8_ARRAY:
+                    {
+                        POP_I32();
+                        POP_REF(REF_TYPE_ARRAYREF);
+                        POP_STRINGREF();
+                        PUSH_I32();
+                        break;
+                    }
+#endif
```
xujuntwt95329, line 12411:
```suggestion
#endif /* end of WASM_ENABLE_STRINGREF != 0 */
```

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -3,6 +3,8 @@
  * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
  */
 
+#include "gc_export.h"
+#include "wasm.h"
```
xujuntwt95329, line 7:
```suggestion
```

lum1n0us:
missing *.c files under *core/iwasm/common/stringref* ?

xujuntwt95329:
> missing *.c files under _core/iwasm/common/stringref_ ?

@lum1n0us 

stringref contains two parts:
- `part 1`: add new ref type in WasmGC, add opcode processing in loader and interpreter
- `part 2`: a library for string representation/encoding/decoding ...

We want to reduce the code size introduced to WAMR, so this PR only include `part 1`, `part 2` will be provided by embedder, they can use their own implementation (e.g. [Wasmnizer-ts use JSString from QuickJS to implement stringref](https://github.com/intel/Wasmnizer-ts/blob/main/runtime-library/stringref/stringref_qjs.c))

Later we will submit another PR of `part 2`, but we want to make it as a sample rather than WAMR core part

core/iwasm/interpreter/wasm.h
```diff
@@ -25,6 +25,10 @@ extern "C" {
 #define VALUE_TYPE_V128 0x7B
 #define VALUE_TYPE_FUNCREF 0x70
 #define VALUE_TYPE_EXTERNREF 0x6F
+#define VALUE_TYPE_STRINGREF 0x64
+#define VALUE_TYPE_STRINGVIEWWTF8 0x63
+#define VALUE_TYPE_STRINGVIEWWTF16 0x62
+#define VALUE_TYPE_STRINGVIEWITER 0x61
```
wenyongh, line 31: The four VALUE_TYPE_STRINGREF types are not used internally, and the similar enum values have been defined in wasm_export.h, how about removing them here? Just keep REF_TYPE_STRINGREFxxx below.

core/iwasm/interpreter/wasm.h
```diff
@@ -132,6 +144,11 @@ typedef void *table_elem_type_t;
 #if WASM_ENABLE_BULK_MEMORY != 0
 #define SECTION_TYPE_DATACOUNT 12
 #endif
+#if WASM_ENABLE_GC != 0
```
wenyongh, line 147: Seems no need to add `#if ENABLE_GC != 0` here, it should be auto enabled when stringref is enabled. The below codes don't add the control like this.

core/iwasm/common/stringref/string_object.h
wenyongh, line 1: How about putting stringref folder under `core/iwasm/common/gc/stringref` and changing iwasm_gc.cmake to control whether and how to add the source files of stringref?

core/iwasm/common/gc/gc_object.c
```diff
@@ -698,6 +709,36 @@ wasm_object_get_ref_list(WASMObjectRef obj, bool *p_is_compact_mode,
         *p_ref_list = NULL;
         return true;
     }
+#if WASM_ENABLE_STRINGREF != 0
+    else if (rtt_type->type_flag == WASM_TYPE_STRINGREF) {
```
wenyongh, line 713: Had better change to below to simplify the code:
```C
    else if (rtt_type->type_flag == WASM_TYPE_STRINGREF
             || rtt_type->type_flag == WASM_TYPE_STRINGVIEWWTF8
             || rtt_type->type_flag == WASM_TYPE_STRINGVIEWWTF16
             || rtt_type->type_flag == WASM_TYPE_STRINGVIEWITER) {
```

core/iwasm/common/gc/gc_object.c
```diff
@@ -750,3 +791,296 @@ wasm_obj_unset_gc_finalizer(wasm_exec_env_t exec_env, void *obj)
...
+                                               &module->rtt_type_lock);
+    }
+#endif
+#if WASM_ENABLE_AOT != 0
+    if (module_inst->module_type == Wasm_Module_AoT) {
+        AOTModule *module =
+            (AOTModule *)((AOTModuleInstance *)module_inst)->module;
+        /* TODO: add stringref_rtts in AOTModule */
+        rtt_type = wasm_stringref_rtt_type_new(
+            WASM_TYPE_STRINGREF, module->rtt_types, &module->rtt_type_lock);
```
wenyongh, line 854: Should call `aot_stringref_rtt_type_new` but it is not implemented yet? Had better comment it here, or compiling AOT feature without interpreter enabled may fail?

core/iwasm/common/gc/gc_object.c
```diff
@@ -750,3 +791,296 @@ wasm_obj_unset_gc_finalizer(wasm_exec_env_t exec_env, void *obj)
...
+    }
+#endif
+#if WASM_ENABLE_AOT != 0
+    if (module_inst->module_type == Wasm_Module_AoT) {
+        AOTModule *module =
+            (AOTModule *)((AOTModuleInstance *)module_inst)->module;
+        /* TODO: add stringref_rtts in AOTModule */
+        rtt_type = wasm_stringref_rtt_type_new(WASM_TYPE_STRINGVIEWWTF8,
+                                               module->rtt_types,
+                                               &module->rtt_type_lock);
```
wenyongh, line 901: Same as above.

core/iwasm/common/gc/gc_object.c
```diff
@@ -750,3 +791,296 @@ wasm_obj_unset_gc_finalizer(wasm_exec_env_t exec_env, void *obj)
...
+
+    wasm_obj_set_gc_finalizer(
+        exec_env, (wasm_obj_t)stringview_wtf8_obj,
+        (wasm_obj_finalizer_t)wasm_stringview_wtf8_obj_finalizer, NULL);
+
+    return stringview_wtf8_obj;
+}
+
+WASMStringviewWTF16ObjectRef
+wasm_stringview_wtf16_obj_new(WASMExecEnv *exec_env, const void *str_obj)
```
wenyongh, line 920: The implementation of the four APIs are very similar, maybe we can unify them in an internal function, like:
```C
static WASMObjectRef
stringref_obj_new(WASMExecEnv *exec_env, uint32 type, const void *str_obj,
                             int32 pos)
```
And these APIs call it with individual types?

build-scripts/runtime_lib.cmake
```diff
@@ -159,6 +159,10 @@ if (WAMR_BUILD_WASM_CACHE EQUAL 1)
     include (${WAMR_ROOT_DIR}/build-scripts/involve_boringssl.cmake)
 endif ()
 
+if (WAMR_BUILD_STRINGREF EQUAL 1)
```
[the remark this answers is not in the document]
yviansu: Yes, in our implementation, we depend on GC feature.

Should we check `if(WAMR_BUILD_STRINGREF EQUAL 1 AND WAMR_BUILD_GC EQUAL 1)` or just `set (WAMR_BUILD_GC 1)` here?

core/iwasm/common/gc/gc_object.c
yviansu: Got it~

build-scripts/runtime_lib.cmake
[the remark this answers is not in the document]
wenyongh: My preference is to `set (WAMR_BUILD_GC 1)` here, so we can just `cmake -DWAMR_BUILD_STRINGREF=1` to build the feature?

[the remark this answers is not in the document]
xujuntwt95329: > My preference is to `set (WAMR_BUILD_GC 1)` here, so we can just `cmake -DWAMR_BUILD_STRINGREF=1` to build the feature?

+1

core/iwasm/common/gc/gc_object.c
yviansu: In interpreter mode, we set stringref related types in a new field `stringref_rtts` in `WASMModule`, and in AOT mode, we hope there is also a `stringref_rtt` field in `AOTModule`, **which not implemented yet.**

What's more, `wasm_stringref_rtt_type_new` will also be called in AOT mode.

core/iwasm/common/gc/gc_object.c
yviansu: Great idea, I will add it.

core/iwasm/common/stringref/string_object.h
yviansu: I think move stringref folder to `core/iwasm/common/gc/stringref` is greate, after all, it can not work without gc feature.

> changing iwasm_gc.cmake to control whether and how to add the source files of stringref

I want to confirm that you mean delete `stringref.cmake` and only control by `iwasm_gc.cmake`?

core/iwasm/interpreter/wasm.h
yviansu: Great idea! 
By the way, the similar enum values have been defined in `gc_export.h`, not `wasm_export.h` :)

build-scripts/runtime_lib.cmake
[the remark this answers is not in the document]
yviansu: Got it~

core/iwasm/interpreter/wasm.h
yviansu: Got it~

core/iwasm/common/stringref/string_object.h
wenyongh: Yes, just remove stringref.cmake if we can.

core/iwasm/common/gc/gc_object.c
wenyongh: Just comment the code here, if rtt_type is NULL, this function will return NULL in L859.

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -2672,6 +2682,727 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+                            wasm_module->string_consts[contents]);
+                        if (!str_obj) {
+                            wasm_set_exception(module,
+                                               "create string object failed");
+                            goto got_exception;
+                        }
+
+                        SYNC_ALL_TO_FRAME();
+                        stringref_obj =
+                            wasm_stringref_obj_new(exec_env, str_obj);
```
wenyongh, line 2752: Same as above for SYNC_ALL_TO_FRAME and str_obj.

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -2672,6 +2682,727 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+                        if (!str_obj) {
```
wenyongh, line 2753: Should be stringref_obj

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -2672,6 +2682,727 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+                            (WASMString)wasm_stringref_obj_get_value(
+                                stringref_obj2));
+                        if (!str_obj) {
+                            wasm_set_exception(module,
+                                               "create string object failed");
+                            goto got_exception;
+                        }
+
+                        SYNC_ALL_TO_FRAME();
+                        stringref_obj =
```
wenyongh, line 2863: Same as above for SYNC_ALL_TO_FRAME and str_obj

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -2672,6 +2682,727 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+                                stringref_obj),
+                            STRING_VIEW_WTF8);
+                        if (!str_obj) {
+                            wasm_set_exception(module,
+                                               "create string object failed");
+                            goto got_exception;
+                        }
+
+                        SYNC_ALL_TO_FRAME();
+                        stringview_wtf8_obj =
```
wenyongh, line 2919: Same as above.

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -2672,6 +2682,727 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+                                stringview_wtf8_obj),
+                            start, end, STRING_VIEW_WTF8);
+                        if (!str_obj) {
+                            wasm_set_exception(module,
+                                               "create string object failed");
+                            goto got_exception;
+                        }
+
+                        SYNC_ALL_TO_FRAME();
+                        stringref_obj =
```
wenyongh, line 3016: Same as above.

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -2672,6 +2682,727 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+                                stringref_obj),
+                            STRING_VIEW_WTF16);
+                        if (!str_obj) {
+                            wasm_set_exception(module,
+                                               "create string object failed");
+                            goto got_exception;
+                        }
+
+                        SYNC_ALL_TO_FRAME();
+                        stringview_wtf16_obj =
```
wenyongh, line 3042: Same as above.

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -2672,6 +2682,727 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+                                stringview_wtf16_obj),
+                            start, end, STRING_VIEW_WTF16);
+                        if (!str_obj) {
+                            wasm_set_exception(module,
+                                               "create string object failed");
+                            goto got_exception;
+                        }
+
+                        SYNC_ALL_TO_FRAME();
+                        stringref_obj =
```
wenyongh, line 3135: Same as above.

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -2672,6 +2682,727 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+                            STRING_VIEW_ITER);
+
+                        if (!str_obj) {
+                            wasm_set_exception(module,
+                                               "create string object failed");
+                            goto got_exception;
+                        }
+
+                        SYNC_ALL_TO_FRAME();
+                        stringview_iter_obj =
```
wenyongh, line 3162: Same as above

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -2672,6 +2682,727 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+                                stringview_iter_obj),
+                            cur_pos, code_points_count, STRING_VIEW_ITER);
+                        if (!str_obj) {
+                            wasm_set_exception(module,
+                                               "create string object failed");
+                            goto got_exception;
+                        }
+
+                        SYNC_ALL_TO_FRAME();
+                        stringref_obj =
```
wenyongh, line 3241: Same as above.

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -2672,6 +2682,727 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+                        str_obj = wasm_string_new_with_encoding(
+                            arr_start_addr, (end - start), flag);
+                        if (!str_obj) {
+                            wasm_set_exception(module,
+                                               "create string object failed");
+                            goto got_exception;
+                        }
+
+                        SYNC_ALL_TO_FRAME();
+                        stringref_obj =
```
wenyongh, line 3313: Same as above.

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -925,6 +925,14 @@ resolve_value_type(const uint8 **p_buf, const uint8 *buf_end,
                                    error_buf, error_buf_size))
             return false;
         *p_need_ref_type_map = true;
+#if WASM_ENABLE_STRINGREF != 0
+        /* covert (ref string) to stringref */
```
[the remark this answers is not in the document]
xujuntwt95329: > Does `(ref string)` equal to stringref? For other reference types, xxxref equals to `(ref null xxx)` but not `(ref xxx)`, e.g., structref is same as `(ref null struct)` but not `(ref struct)`, it is nullable. My understanding is that stringref should comply with the rule, could you check whether stringref is nullable?

This is not well defined, but binaryen‘s optimizer sometimes optimize (stringref) to (ref string), and usually assign a (stringref) to (ref string), we've tested with V8 and these are treated as valid code, so we add such workaround here, otherwise we can't run the optimized wasm module

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -3699,6 +3707,65 @@ load_start_section(const uint8 *buf, const uint8 *buf_end, WASMModule *module,
...
+                      string_length + 1, error_buf, error_buf_size))) {
+                goto fail;
+            }
+            str_content = (uint8 *)module->string_consts[i];
+
+            if (string_length > 0) {
+                for (j = 0; j < string_length; j++) {
+                    *(str_content + j) = read_uint8(p);
+                }
+            }
```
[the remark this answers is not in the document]
xujuntwt95329: > How about using `const_str_list_insert` to read the string like other places?
> 
> ```c
> CHECK_BUF(p, p_end, string_length);
> if (!(module->string_consts[i] = const_str_list_insert(p, string_length, ...))) {
>     ...
> }
> p += string_length;
> ```

Good idea! Thanks~

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -2672,6 +2682,727 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+                            flag = UTF8;
+                        }
+                        else if (opcode == WASM_OP_STRING_NEW_LOSSY_UTF8) {
+                            flag = LOSSY_UTF8;
+                        }
+                        else if (opcode == WASM_OP_STRING_NEW_WTF8) {
+                            flag = WTF8;
+                        }
+
+                        str_obj = wasm_string_new_with_encoding(
```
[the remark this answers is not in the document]
xujuntwt95329: No, this line doesn't create a gc object, it just create a `String` object implemented by external library, implementer can use any implementation, and the memory is not managed by WAMR at all.
The L2723 will eventually create a GC object (stringref) with a opaque pointer pointing to the object created by this line.

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -2672,6 +2682,727 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+                            maddr, bytes_length, flag);
+                        if (!str_obj) {
+                            wasm_set_exception(module,
+                                               "create string object failed");
+                            goto got_exception;
+                        }
+
+                        SYNC_ALL_TO_FRAME();
+                        stringref_obj =
+                            wasm_stringref_obj_new(exec_env, str_obj);
```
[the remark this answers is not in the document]
xujuntwt95329: The str_obj is not a GC managed object, it's provided by string implementer. WAMR just treat it as a raw pointer and will not access its content.
There is a graph in `wasm.h` to explain this

core/iwasm/interpreter/wasm_interp_classic.c
xujuntwt95329: Same as above

core/iwasm/interpreter/wasm_interp_classic.c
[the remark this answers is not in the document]
wenyongh: Got it, thanks.

core/iwasm/common/gc/gc_type.c
```diff
@@ -745,13 +750,26 @@ wasm_is_reftype_supers_of_extern(uint8 type)
...
 inline static bool
 wasm_is_reftype_supers_of_none(uint8 type, const WASMRefType *ref_type,
                                const WASMTypePtr *types, uint32 type_count)
 {
     if (type == REF_TYPE_NULLREF || type == REF_TYPE_I31REF
         || type == REF_TYPE_STRUCTREF || type == REF_TYPE_ARRAYREF
-        || wasm_is_reftype_supers_of_eq(type))
+        || wasm_is_reftype_supers_of_eq(type)
+#if WASM_ENABLE_STRINGREF != 0
+        || type == REF_TYPE_STRINGREF
```
wenyongh, line 770: Can type be REF_TYPE_STRINGVIEWWTF8/REF_TYPE_STRINGVIEWWTF16/REF_TYPE_STRINGVIEWITER?

xujuntwt95329: No, these stringview types are not in the subtyping chain, they are created from stringref when really needed.

wenyongh (approved):
LGTM

PR #2654, opened 2023-10-18
Register GC symbols to AOT loader

no1wudi:


core/iwasm/aot/aot_loader.c
```diff
@@ -1338,7 +1339,7 @@ load_types(const uint8 **p_buf, const uint8 *buf_end, AOTModule *module,
             func_type->ret_cell_num = ret_cell_num;
 
             if (ref_type_map_count == 0) {
-                continue;
+                goto resolve_parent_type;
```
wenyongh, line 1342: Using goto here makes code a little ugly, how about:
```C
    if (ref_type_map_count > 0) {
        ...
    }
```

core/iwasm/aot/aot_loader.c
```diff
@@ -1388,7 +1390,7 @@ load_types(const uint8 **p_buf, const uint8 *buf_end, AOTModule *module,
             read_uint16(buf, buf_end, struct_type->ref_type_map_count);
 
             if (struct_type->ref_type_map_count == 0) {
-                continue;
+                goto resolve_parent_type;
```
wenyongh, line 1393: Same as above

PR #2655, opened 2023-10-20
Fix several typo/warning/unused-code issues

wenyongh:
- Fix typo in wamr-test-suites script
- Fix compilation warnings in libc-wasi posix.c
- Remove unused code fast-jit jit_frontend.c
- Remove duplicated exception print in `iwasm -f <function>`
- Fix return value in void function wasm_runtime_set_wasi_ctx


core/iwasm/common/wasm_application.c
```diff
@@ -632,9 +631,7 @@ execute_func(WASMModuleInstanceCommon *module_inst, const char *name,
     if (argv1)
         wasm_runtime_free(argv1);
 
-    exception = wasm_runtime_get_exception(module_inst);
-    bh_assert(exception);
-    os_printf("%s\n", exception);
```
wenyongh, line 637: Duplicated exception dump is added in iwasm main.c after calling app_instance_func.

PR #2656, opened 2023-10-20
Fix some GC relative issues

no1wudi:


core/iwasm/compilation/aot.c
```diff
@@ -511,7 +464,7 @@ aot_create_comp_data(WASMModule *module, bool gc_enabled)
 
     /* Create function types */
     comp_data->type_count = module->type_count;
-    if (comp_data->type_count && !(comp_data->types = aot_create_types(module)))
+    if (comp_data->type_count && !(comp_data->types = module->types))
         goto fail;
```
wenyongh, line 468: Just `comp_data->types = module->types;`, no need to check type_count and types, it has been checked in wasm loader?

wenyongh (commented):
LGTM

PR #2657, opened 2023-10-20
Partial windows filesystem implementation

zoraaver:
Implement the necessary os_ filesystem functions to enable successful WASI initialisation on Windows. Some small changes were also required to the sockets implementation to use the new windows_handle type. The remaining functions will be implemented in a future PR.

### DIFF
--- a/core/shared/platform/windows/platform_internal.h
+++ b/core/shared/platform/windows/platform_internal.h
@@ -25,6 +25,7 @@
 #include <stdint.h>
 #include <malloc.h>
 #include <process.h>
+#include <winapifamily.h>
 #include <winsock2.h>
 #include <ws2tcpip.h>
 #include <windows.h>
@@ -82,8 +83,6 @@ typedef struct korp_cond {
     struct os_thread_wait_node *thread_wait_list_end;
 } korp_cond;
 
-#define bh_socket_t SOCKET
-
 unsigned
 os_getpagesize();
 void *
@@ -137,13 +136,45 @@ bh_atomic_thread_fence(int mem_order);
 
 #define os_atomic_thread_fence bh_atomic_thread_fence
 
-typedef HANDLE os_file_handle;
-typedef void *os_dir_stream;
+typedef enum windows_handle_type {
+    windows_handle_type_socket,
+    windows_handle_type_file
+} windows_handle_type;
+
+typedef enum windows_access_mode {
+    windows_access_mode_read = 1 << 0,
+    windows_access_mode_write = 1 << 1
+} windows_access_mode;
+
+typedef struct windows_handle {
+    windows_handle_type type;
+    windows_access_mode access_mode;
+    union {
+        HANDLE handle;
+        SOCKET socket;
+    } raw;
+} windows_handle;
+
+typedef struct windows_dir_stream {
+    // Enough space for the wide filename and the info struct itself
+    char info_buf[PATH_MAX * sizeof(wchar_t) + sizeof(FILE_ID_BOTH_DIR_INFO)];
+    char current_entry_name[PATH_MAX];
+    // An offset into info_buf to read the next entry from
+    DWORD cursor;
+    int cookie;
+    windows_handle *handle;
+} windows_dir_stream;
+
+typedef windows_handle *os_file_handle;
+typedef windows_dir_stream *os_dir_stream;
+
 #if WASM_ENABLE_UVWASI != 1
 typedef HANDLE os_raw_file_handle;
 #else
 typedef uint32_t os_raw_file_handle;
 #endif
+
+#define bh_socket_t windows_handle *
 
 #ifdef __cplusplus
 }
--- a/core/shared/platform/windows/win_clock.c
+++ b/core/shared/platform/windows/win_clock.c
@@ -5,6 +5,7 @@
 
 #include "platform_api_vmcore.h"
 #include <winternl.h>
+#include "win_util.h"
 
 #define NANOSECONDS_PER_SECOND 1000000000ULL
 #define NANOSECONDS_PER_TICK 100
@@ -20,20 +21,6 @@ calculate_monotonic_clock_frequency(uint64 *out_frequency)
         *out_frequency = (uint64)frequency.QuadPart;
         return BHT_OK;
     }
-}
-
-// The implementation below derives from the following source:
-// https://github.com/WasmEdge/WasmEdge/blob/b70f48c42922ce5ee7730054b6ac0b1615176285/lib/host/wasi/win.h#L210
-static uint64
-filetime_to_wasi_timestamp(FILETIME filetime)
-{
-    static const uint64 ntto_unix_epoch =
-        134774ULL * 86400ULL * NANOSECONDS_PER_SECOND;
-
-    ULARGE_INTEGER temp = { .LowPart = filetime.dwLowDateTime,
-                            .HighPart = filetime.dwHighDateTime };
-
-    return (temp.QuadPart * 100ull) - ntto_unix_epoch;
 }
 
 static int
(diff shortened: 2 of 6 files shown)


core/shared/platform/windows/platform_internal.h
```diff
@@ -137,14 +136,46 @@ bh_atomic_thread_fence(int mem_order);
...
+typedef windows_handle *os_file_handle;
+typedef windows_dir_stream *os_dir_stream;
+
 #if WASM_ENABLE_UVWASI != 1
 typedef HANDLE os_raw_file_handle;
 #else
 typedef uint32_t os_raw_file_handle;
 #endif
 
+#define bh_socket_t os_file_handle
```
loganek, line 177: why not typedef?

zoraaver: Agree it would be better to typedef but it's defined as a macro on other platforms and if I define it via typedef, I get a bunch of errors due to [this](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/shared/platform/include/platform_common.h#L115-L119) which overwrites the type as`int`. It could be fixed but then I'd need to touch all the other `platform_internal.h`'s which don't define `bh_socket_t` and add a definition so we can remove the default macro define in platform_common.h.

Actually, we can completely remove the `bh_socket_t` type since it always has to be the same as `os_file_handle`. I can raise another PR later on to fix that as part of the Windows work if that sounds good?

zoraaver:
@wenyongh would you be able to take a look at this PR? thanks

core/shared/platform/windows/win_file.c
```diff
@@ -4,117 +4,766 @@
...
+        case WSAEPROTOTYPE:
+            return __WASI_EPROTOTYPE;
+        case WSAESOCKTNOSUPPORT:
+            return __WASI_ENOTSUP;
+        case WSAEINVALIDPROCTABLE:
+        case WSAEINVALIDPROVIDER:
+        case WSAEPROVIDERFAILEDINIT:
+        case WSANOTINITIALISED:
+        default:
+            return __WASI_ENOSYS;
```
loganek, line 88: I don't think `ENOSYS` is the correct one here, that suggests the function is not implemented, whereas it is, it just returns an error. Maybe `EINVAL` would be better instead?

core/shared/platform/windows/win_file.c
```diff
@@ -4,117 +4,766 @@
...
+        case ERROR_NO_UNICODE_TRANSLATION:
+        default:
+            return __WASI_ENOSYS;
+    }
+}
+
+static __wasi_filetype_t
+get_disk_filetype(DWORD attribute)
+{
+    if (attribute == INVALID_FILE_ATTRIBUTES)
```
loganek, line 135: can we have consistent formatting here (either use `{}` or not)?

core/shared/platform/windows/win_file.c
```diff
@@ -4,117 +4,766 @@
...
+    int converted_chars =
+        MultiByteToWideChar(CP_UTF8, 0, str, -1, buf, (int)buf_size);
+
+    if (converted_chars == 0)
+        return convert_windows_error_code(GetLastError());
+
+    return __WASI_ESUCCESS;
+}
+
+// Get the filepath for a handle. The size of the buffer should in terms of
```
loganek, line 218:
```suggestion
// Get the filepath for a handle. The size of the buffer should be specified in terms of
```

core/shared/platform/windows/win_file.c
```diff
@@ -245,23 +953,48 @@ os_get_invalid_dir_stream()
...
+    return NULL;
 }
 
 bool
 os_is_handle_valid(os_file_handle *handle)
 {
-    return false;
+    assert(handle != NULL);
+
+    if ((*handle) == NULL
```
loganek, line 977: can this function be implemented using the `CHECK_VALID_HANDLE` macro?

core/shared/platform/windows/win_file.c
zoraaver: have removed braces for single-line if statements

core/shared/platform/windows/win_clock.c
```diff
@@ -96,7 +83,7 @@ os_clock_time_get(bh_clock_id_t clock_id, uint64 precision, uint64 *time)
 #else
             GetSystemTimeAsFileTime(&SysNow);
```
wenyongh, line 84: Should be `sys_now`

core/shared/platform/windows/win_file.c
```diff
@@ -4,117 +4,766 @@
...
+#define CHECK_VALID_WIN_DIR_STREAM(win_dir_stream)         \
+    do {                                                   \
+        if ((win_dir_stream) == NULL)                      \
+            return __WASI_EINVAL;                          \
+        CHECK_VALID_FILE_HANDLE((win_dir_stream)->handle); \
+    } while (0)
+
+#define CHECK_VALID_PATH(path)          \
+    do {                                \
+        if (strstr(path, "\\") != NULL) \
```
wenyongh, line 47: A little confused, windows path cannot contain '\\'?

core/shared/platform/windows/win_file.c
```diff
@@ -4,117 +4,766 @@
...
+}
+
+#if WINAPI_PARTITION_DESKTOP
+// Modifies the given path in place and replaces it with the filename component
+// (including the extension) of the path.
+static __wasi_errno_t
+extract_filename_from_path(wchar_t *path, size_t buf_size)
+{
+    wchar_t extension[256];
+    wchar_t filename[256];
```
wenyongh, line 341: Buffer size of filename may be a little small?

core/shared/platform/windows/win_file.c
```diff
@@ -4,117 +4,766 @@
...
+        buf->st_atim = convert_filetime_to_wasi_timestamp(
+            (LPFILETIME)&file_id_both_dir_info->LastAccessTime.QuadPart);
+        buf->st_mtim = convert_filetime_to_wasi_timestamp(
+            (LPFILETIME)&file_id_both_dir_info->LastWriteTime.QuadPart);
+        buf->st_ctim = convert_filetime_to_wasi_timestamp(
+            (LPFILETIME)&file_id_both_dir_info->ChangeTime.QuadPart);
+        buf->st_size =
+            (__wasi_filesize_t)(file_id_both_dir_info->EndOfFile.QuadPart);
+
+        break;
```
wenyongh, line 466: Coding style, how about changing to:
```C
        if (wcsncmp(file_id_both_dir_info->FileName, path, filename_length)
            == 0) {
            buf->st_ino = (__wasi_inode_t)(file_id_both_dir_info->FileId.QuadPart);
            ...
            break;
        }
```

core/shared/platform/windows/win_file.c
zoraaver: I'll add a comment to make it clearer - the issue is that the wasm application won't know which platform it's running on - on POSIX, backslashes are a valid filename character whereas on Windows, they are a path separator.

e.g. if the wasm app had the following code
```
FILE *file = fopen("foo\\bar", "r");
```
(assuming backslashes are allowed on Windows) this would open the`bar` file under `foo` rather than opening `foo\\bar` on POSIX which may or may not be what the user wanted.

So I thought the easiest option we have on Windows is to completely disallow backslashes in the path since it's probably not what the user intended (and there isn't a way in Windows to allow backslashes to be part of a filename). It's probably very much an edge case since it would be quite unusual to have a backslash in a filename anyway.

core/shared/platform/windows/win_file.c
zoraaver: Going by https://learn.microsoft.com/en-gb/windows/win32/fileio/maximum-file-path-limitation?tabs=registry, the max filename length is 256 characters.

core/shared/platform/windows/win_file.c
loganek: hmm, I thought it's quite common to have a nested file structure and when the user wants to access a file in one of the (sub)directories, they need to provide a path to it (that includes `\`).

zoraaver: Discussed with @loganek offline and we agreed to remove this check. My main concern was the differing behavior on POSIX and Windows re backslashes but I guess that's better than not allowing backslashes on windows at all.

loganek (approved):
LGTM

loganek (approved):
LGTM

core/shared/platform/windows/win_file.c
wenyongh: Got it, thanks.

wenyongh (commented):
LGTM

Issue #2635, opened 2023-10-11
wasi-threads tests not working with fast interpreter


christinaings:
Hello,

I am one of the Amazon devs working on wasi-threads and we are trying to use [wasm parallel gzip](https://github.com/bytecodealliance/wasm-parallel-gzip/tree/main) to test performance of wasi-threads on devices like NVIDIA Shield. I'm encountering an error when running this with the fast interpreter, namely:

```
2023-10-09 17:39:46.607 11748-11832 wasm_runtime::          com...groom.ignite.testrunner.debug  I  [04:09:37:726 - 943811C0]: 
2023-10-09 17:39:46.608 11748-11832 wasm_runtime::          com...groom.ignite.testrunner.debug  I  WASM module load failed: pre-compiled label offset out of range
```

I have run this with the classic interpreter and do not encounter the issue. Does anyone have any suggestions on how this can be resolved? I am unable to attach the wasm file, as wasm file type is not supported.

wenyongh:
Hi, this may be caused by too much code size in fast interpreter, not sure whether you compiled iwasm with release mode, and did you enable some unnecessary features? If release mode and only necessary features are enabled, could you try disabling the labels-as-values feature: in core/config.h, change `#define WASM_ENABLE_LABELS_AS_VALUES 1` to `#define WASM_ENABLE_LABELS_AS_VALUES 0`, or add `add_definitions(-DWASM_ENABLE_LABELS_AS_VALUES=0)` in CMakeLists.txt.

cimacmillan:
Hi @wenyongh, I think the cause is due to being built in debug mode like you mention. Where it's failing the check before emitting the function address for bytecode dispatch [here](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_loader.c#L5396). The opcode is 0x10 and offset becomes 55972 in debug mode. In release build it's 23748. It is ARMv7 with eabi.

For debug mode, I've set `WASM_ENABLE_LABELS_AS_VALUES 0` as I think that will be fine for our use-case, and the test we have is passing now. 

[...]

PR #2659, opened 2023-10-23
Fix fast-interp "pre-compiled label offset out of range" issue

wenyongh:
When labels-as-values is enabled in a target which doesn't support
unaligned address access, 16-bit offset is used to store the relative
offset between two opcode labels. But it is a little small and the loader
may report "pre-compiled label offset out of range" error.

Emitting 32-bit data instead to resolve the issue: emit label address in
32-bit target and emit 32-bit relative offset in 64-bit target.

See also:
https://github.com/bytecodealliance/wasm-micro-runtime/issues/2635

### DIFF
--- a/core/iwasm/interpreter/wasm_interp_fast.c
+++ b/core/iwasm/interpreter/wasm_interp_fast.c
@@ -1128,12 +1128,27 @@ wasm_interp_dump_op_count()
         goto *p_label_addr;                            \
     } while (0)
 #else
-#define FETCH_OPCODE_AND_DISPATCH()                                 \
-    do {                                                            \
-        const void *p_label_addr = label_base + *(int16 *)frame_ip; \
-        frame_ip += sizeof(int16);                                  \
-        goto *p_label_addr;                                         \
+#if UINTPTR_MAX == UINT64_MAX
+#define FETCH_OPCODE_AND_DISPATCH()                                       \
+    do {                                                                  \
+        const void *p_label_addr;                                         \
+        bh_assert(((uintptr_t)frame_ip & 1) == 0);                        \
+        /* int32 relative offset was emitted in 64-bit target */          \
+        p_label_addr = label_base + (int32)LOAD_U32_WITH_2U16S(frame_ip); \
+        frame_ip += sizeof(int32);                                        \
+        goto *p_label_addr;                                               \
     } while (0)
+#else
+#define FETCH_OPCODE_AND_DISPATCH()                                      \
+    do {                                                                 \
+        const void *p_label_addr;                                        \
+        bh_assert(((uintptr_t)frame_ip & 1) == 0);                       \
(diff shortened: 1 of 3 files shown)


core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -1128,12 +1128,27 @@ wasm_interp_dump_op_count()
...
-#define FETCH_OPCODE_AND_DISPATCH()                                 \
-    do {                                                            \
-        const void *p_label_addr = label_base + *(int16 *)frame_ip; \
-        frame_ip += sizeof(int16);                                  \
-        goto *p_label_addr;                                         \
+#if UINTPTR_MAX == UINT64_MAX
+#define FETCH_OPCODE_AND_DISPATCH()                                       \
+    do {                                                                  \
+        const void *p_label_addr;                                         \
+        bh_assert(((uintptr_t)frame_ip & 1) == 0);                        \
```
xujuntwt95329, line 1135: should `& 3`?

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -5389,21 +5389,27 @@ wasm_loader_pop_frame_csp(WASMLoaderContext *ctx, char *error_buf,
...
-        }                                                                      \
-        wasm_loader_emit_int16(loader_ctx, offset);                            \
+        /* emit int32 relative offset in 64-bit target */                      \
+        wasm_loader_emit_uint32(loader_ctx, offset);                           \
         LOG_OP("\nemit_op [%02x]\t", opcode);                                  \
     } while (0)
+#else
+#define emit_label(opcode)                                       \
+    do {                                                         \
+        uint32 offset = (uint32)(uintptr_t)handle_table[opcode]; \
```
xujuntwt95329, line 5404: had better rename offset to `position` or `label_addr`?

core/iwasm/interpreter/wasm_interp_fast.c
wenyongh: No, the frame_ip must be 2-byte aligned, but not necessarily 4-byte aligned. If we force the frame_ip to be 4-byte aligned in wasm_loader, we need to add padding byte if needed, and here, we need to add code like `frame_ip = align_uint(frame_ip, 4)` or `if (frame_ip & 2) frame_ip += 2`, I had a try, it impacts performance a lot.

xujuntwt95329: Got it, Thanks!

I originally thought the CPU requires 4-bytes align, if 2-bytes align is OK, then no problem here.

core/iwasm/interpreter/wasm_loader.c
wenyongh: Done, thanks.

xujuntwt95329 (approved):
LGTM

PR #2664, opened 2023-10-24
Fix GC AOT struct/array/i31/exception issues

wenyongh:


TianlongLiang:
LGTM

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -2461,9 +2451,11 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
                             goto got_exception;
                         }
-                        i31_val = wasm_i31_obj_get_value(
-                            i31_obj,
-                            opcode == WASM_OP_I31_GET_S ? true : false);
+                        i31_val = (uint32)(uintptr_t)i31_obj;
+                        if (opcode == WASM_OP_I31_GET_U)
+                            i31_val = i31_val >> 1;
+                        else
+                            i31_val = (uint32)((int32)i31_val >> 1);
```
xujuntwt95329, line 2458: Does the standard defines that right shifting signed value will be sign extended?

I saw some comments from stack overflow:

https://stackoverflow.com/questions/28681131/right-shift-for-signed-integer

From the C99 standard N1256 (emphasis mine):

> 6.5.7 Bitwise shift operators
> 5 The result of E1 >> E2 is E1 right-shifted E2 bit positions. If E1 has an unsigned type or if E1 has a signed type and a nonnegative value, the value of the result is the integral part of the quotient of E1 X 2E2. If E1 has a signed type and a negative value, the resulting value is implementation-defined.

xujuntwt95329: The statement in standard is here: http://port70.net/%7Ensz/c/c99/n1256.html#6.5.7p5

wenyongh: OK, thanks, I restored the code of interpreter.

xujuntwt95329 (commented):
LGTM

Issue #2665, opened 2023-10-24
Problem building WAMR 1.2.3


lulingar:
### Description

When building WAMR-1.2.3 for the x86-64 target, I get `libvmlib.a` without symbols.

### Procedure

Working on a git clone of this repo:
```
$ git checkout WAMR-1.2.3
$ mkdir ${HOME}/wamr
$ cmake -B ${HOME}/wamr \
      -DWAMR_BUILD_INTERP=1 -DWAMR_BUILD_FAST_INTERP=1 \
      -DWAMR_BUILD_JIT=0 -DWAMR_BUILD_AOT=1 \
      -DWAMR_BUILD_LIBC_WASI=1 -DWAMR_BUILD_LIBC_BUILTIN=0 \
      -DWAMR_BUILD_LIBC_UVWASI=0 \
      -DWAMR_BUILD_MULTI_MODULE=0 \
      -DWAMR_BUILD_MINI_LOADER=0 \
      -DWAMR_BUILD_SHARED_MEMORY=1 -DWAMR_BUILD_BULK_MEMORY=1 \
      -DWAMR_BUILD_THREAD_MGR=1 -DWAMR_BUILD_LIB_PTHREAD=1 \
      -DWAMR_BUILD_TARGET=X86_64 \
      .
-- The C compiler identification is GNU 11.4.0
-- The CXX compiler identification is GNU 11.4.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Build Configurations:
     Build as target X86_64
     CMAKE_BUILD_TYPE Release
     WAMR Interpreter enabled
     WAMR AOT enabled
     WAMR Fast JIT disabled
     WAMR LLVM ORC JIT disabled
     Libc builtin disabled
     Libc WASI enabled
     Fast interpreter enabled
     Multiple modules disabled
     Bulk memory feature enabled
     Shared memory enabled
     Thread manager enabled
     Lib pthread enabled
     SIMD enabled
     Reference types disabled
-- The ASM compiler identification is GNU
-- Found assembler: /usr/bin/cc
[...]

PR #2669, opened 2023-10-24
Revert "Strip static and shared libraries of iwasm to reduce the bin..."

yamt:
This reverts commit 7e3a29d5e4564406c91ada436b1e33d53d1232cf.

* it doesn't work for macOS. ("unrecognized option: --strip-all")

* it doesn't work for ubuntu. https://github.com/bytecodealliance/wasm-micro-runtime/issues/2665

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -107,11 +107,6 @@ endif ()
 
 set (WAMR_ROOT_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 
-# Set the strip command based on the system (GNU or Clang)
-if (CMAKE_STRIP)
-    set (CMAKE_STRIP_FLAGS "--strip-all")
-endif ()
-
 include (${WAMR_ROOT_DIR}/build-scripts/runtime_lib.cmake)
 
 set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra -Wformat -Wformat-security -Wshadow -Wno-unused-parameter")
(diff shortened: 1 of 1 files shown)


lum1n0us:
A little confused here, do we need a stripped release or a symbolized release?

wenyongh (approved):
LGTM

lum1n0us:
@cngzhnp please be aware of it.

cngzhnp:
@lum1n0us  Thanks for the warning. However, I do not understand that why we need to revert it. If does not work with macOS which is not possible by the way, only we need to adapt CMakeFiles.txt for this, not to revert it.

Someone needs to see symbols in the static library or shared library should generate binary in DEBUG mode, not release mode in my opinion. Could you please explain it to me more @yamt ?

yamt:
> @lum1n0us Thanks for the warning. However, I do not understand that why we need to revert it. If does not work with macOS which is not possible by the way, only we need to adapt CMakeFiles.txt for this, not to revert it.
> 
> Someone needs to see symbols in the static library or shared library should generate binary in DEBUG mode, not release mode in my opinion. Could you please explain it to me more @yamt ?

as far as tested, it produced unlinkable library on ubuntu.
i'm not sure for what platform it has been tested.

cngzhnp:
Do you have reproduce steps how to test that it was unlinkable library?

yamt:
> Do you have reproduce steps how to test that it was unlinkable library?

have you looked at https://github.com/bytecodealliance/wasm-micro-runtime/issues/2665 ?

PR #2673, opened 2023-10-26
Fix repeatedly initialize shared memory data and protect the memory's fields

wenyongh:
Add shared memory lock when accessing some fields of the memory instance
when the memory instance is shared.
And avoid repeatedly initializing the shared memory data when creating the
child thread in lib-pthread or lib-wasi-threads.

### DIFF
--- a/core/iwasm/interpreter/wasm_runtime.c
+++ b/core/iwasm/interpreter/wasm_runtime.c
@@ -1774,10 +1774,23 @@ wasm_instantiate(WASMModule *module, WASMModuleInstance *parent,
         uint8 *memory_data = NULL;
         uint64 memory_size = 0;
         WASMDataSeg *data_seg = module->data_segments[i];
+#if WASM_ENABLE_SHARED_MEMORY != 0
+        bool is_shared_memory;
+#endif
 
 #if WASM_ENABLE_BULK_MEMORY != 0
         if (data_seg->is_passive)
             continue;
+#endif
+
+#if WASM_ENABLE_SHARED_MEMORY != 0
+        /* Currently we have only one memory instance */
+        is_shared_memory = module->memories[0].flags & 0x02 ? true : false;
+        if (is_shared_memory && parent != NULL) {
+            /* Ignore setting memory init data if the memory has been
+             * initialized */
+            continue;
+        }
 #endif
 
         /* has check it in loader */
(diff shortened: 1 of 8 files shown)


xujuntwt95329 (approved):
LGTM

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -1774,12 +1774,31 @@ wasm_instantiate(WASMModule *module, WASMModuleInstance *parent,
...
 #if WASM_ENABLE_BULK_MEMORY != 0
         if (data_seg->is_passive)
             continue;
 #endif
 
+#if WASM_ENABLE_SHARED_MEMORY != 0
+        /* Currently we have only one memory instance */
+        if (module->import_memory_count > 0) {
+            is_shared_memory =
+                module->import_memories[0].u.memory.flags ? true : false;
```
loganek, line 1790: Should we check for the 0x2 flag here as well? Or imported memory cannot have any other flag values other than 0x2?

wenyongh: Yes, it is a mistake, thanks a lot for spotting, fixed.

loganek (approved):
LGTM, just a question on the performance - I see that adds extra lock/unlock for all WASI calls. Is there any other significant scenario that will be affected by this change from perf perspective?

wenyongh:
> LGTM, just a question on the performance - I see that adds extra lock/unlock for all WASI calls. Is there any other significant scenario that will be affected by this change from perf perspective?

Yes, performance may be an issue, we had better test more. Maybe we can refine the code, e.g., add lock/unlock only when the memory is actually shared, and load/store memory's memory_data_size/memory_data_end fields with atomic operations, like what we did for suspend_flags.

wenyongh:
> LGTM, just a question on the performance - I see that adds extra lock/unlock for all WASI calls. Is there any other significant scenario that will be affected by this change from perf perspective?

https://github.com/bytecodealliance/wasm-micro-runtime/pull/2682

wenyongh:
@loganek I have merged #2682, if there is any issue found or there is enhancement needed, let's create another PR to fix/enhance it. Thanks.

loganek:
@wenyongh thanks a lot and sorry for a slow response, I just came back from holidays.

PR #2676, opened 2023-10-27
Minor fixes for Go bindings

benozol:


language-bindings/go/wamr/instance.go
```diff
@@ -272,7 +272,9 @@ func (self *Instance) CallFuncV(funcName string,
     for i = 0; i < result_count; i++ {
         switch result_types[i] {
             case C.WASM_I32:
+                fallthrough
             case C.WASM_FUNCREF:
+                fallthrough
```
benozol, line 277: switch cases in Go require an explicit fallthrough, they break by default

language-bindings/go/go.mod
```diff
@@ -1,4 +1,4 @@
-module gitlab.alipay-inc.com/TNT_Runtime/ant-runtime/bindings/go
+module github.com/bytecodealliance/wasm-micro-runtime/language-bindings/go
```
benozol, line 1: when using the Go bindings in a separate project, the module name must match the project URL, and the test below must use the same URL, too

wenyongh: Should we also we put header and libraries files under `github.com/bytecodealliance/wasm-micro-runtime/language-bindings/go/packaged`?

benozol: The header files and the host-native library are installed under `packaged/` by `language-bindings/go/build.sh`. But there's also the script `language-bindings/go/samples/build.sh`, which does practically the same but always puts the host-native library into `packaged/lib/amd64`, which is incorrect, e.g., on arm64. What about removing the latter, and maybe copying the command to build the test from the former to the latter?

wenyongh: How about not removing `go/samples/build.sh`, instead, just letting it call the `go/build.sh` first and then building the sample, for example:
```bash
#!/usr/bin/env bash

# Copyright (C) 2019 Intel Corporation.  All rights reserved.

CUR_DIR=$PWD

pushd ${CUR_DIR}/.. > /dev/null 2>&1
./build.sh
popd > /dev/null 2>& 1

cd ${CUR_DIR}
rm -f test
go build test.go
./test
```

wenyongh: See https://github.com/bytecodealliance/wasm-micro-runtime/pull/2681

wenyongh (commented):
LGTM

PR #2682, opened 2023-10-30
Refine lock/unlock shared memory

wenyongh:
Split memory instance's field `uint32 ref_count` into `bool is_shared_memory`
and `uint16 ref_count`, and lock the memory only when `is_shared_memory`
flag is true, no need to acquire a lock for non-shared memory when shared
memory feature is enabled.

### DIFF
--- a/core/iwasm/interpreter/wasm_runtime.h
+++ b/core/iwasm/interpreter/wasm_runtime.h
@@ -79,8 +79,11 @@ typedef union {
 struct WASMMemoryInstance {
     /* Module type */
     uint32 module_type;
+
+    bool is_shared_memory;
+
     /* Shared memory flag */
-    bh_atomic_32_t ref_count; /* 0: non-shared, > 0: reference count */
+    bh_atomic_16_t ref_count; /* 0: non-shared, > 0: reference count */
 
     /* Number bytes per page */
     uint32 num_bytes_per_page;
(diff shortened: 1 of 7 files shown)


xujuntwt95329 (commented):
LGTM

core/iwasm/interpreter/wasm_runtime.h
```diff
@@ -79,8 +79,11 @@ typedef union {
 struct WASMMemoryInstance {
     /* Module type */
     uint32 module_type;
+
+    bool is_shared_memory;
+
     /* Shared memory flag */
-    bh_atomic_32_t ref_count; /* 0: non-shared, > 0: reference count */
+    bh_atomic_16_t ref_count; /* 0: non-shared, > 0: reference count */
```
yamt, line 86: @wenyongh 
doesn't this break aot abi?
what size and alignment are you assuming for a bool here?

the whole purpose of having ref_count and is_shared_memory combined was to avoid aot abi.

wenyongh: Pur my understanding, the bool size is 1 byte and the total size of WASMMemoryInstance is same as before, and ref_count is at 2-byte aligned address, it doesn't violate the aot abi. And the AOT code doesn't access the two fields.

There are also bh_static_asserts in aot_runtime.c:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/aot/aot_runtime.c#L37-L46

yamt: iirc, standard-wise, a bool can be larger.
for things where the size is important, it's simpler to use uint8 or 16.

wenyongh: OK, let's create another PR to use uint8 instead.

Issue #2680, opened 2023-10-28
TSAN fast interpreter failure


Zzzabiyaka:
During PoC of #2679 I've found another data race in exception handling https://gist.github.com/Zzzabiyaka/dcad23be15bad484966e91b8da3d659a.

`set_exception_visitor` in the stacktrace makes me suspicious about #2509. @yamt, would you be able to look into that issue ?

wenyongh:
This seems to be because that before creating a new thread, runtime will instantiate the module instance for it (and do the instantiation in the parent thread), and at the end of instantiating, runtime executes the `start` of the module, and to make the exec_env consistent with the exec_envs in the parent thread's call stack, runtime temporarily replace `exec_env->module_inst` with the child module instance.

There may be two options: (1) since the `start` function has been called in the main thread in the first time instantiation, we don't call it again in the later instantiation, (2) add `cluster->lock ` for replacing `exec_env->module_inst`, executing start and restoring the `exec_env->module_inst` in the later instantiation.

@yamt how do think?

yamt:
> This seems to be because that before creating a new thread, runtime will instantiate the module instance for it (and do the instantiation in the parent thread), and at the end of instantiating, runtime executes the `start` of the module, and to make the exec_env consistent with the exec_envs in the parent thread's call stack, runtime temporarily replace `exec_env->module_inst` with the child module instance.
> 
> There may be two options: (1) since the `start` function has been called in the main thread in the first time instantiation, we don't call it again in the later instantiation, (2) add `cluster->lock ` for replacing `exec_env->module_inst`, executing start and restoring the `exec_env->module_inst` in the later instantiation.
> 
> @yamt how do think?

(1) doesn't work because the start function is expected be called for each threads.

(2) might work. but i feel it's still a bit dirty and fragile.

how about (3) create a temporary exec_env to run the start function?

wenyongh:
> > This seems to be because that before creating a new thread, runtime will instantiate the module instance for it (and do the instantiation in the parent thread), and at the end of instantiating, runtime executes the `start` of the module, and to make the exec_env consistent with the exec_envs in the parent thread's call stack, runtime temporarily replace `exec_env->module_inst` with the child module instance.
[...]

PR #2685, opened 2023-10-31
grab cluster->lock when modifying exec_env->module_inst

yamt:
fixes: https://github.com/bytecodealliance/wasm-micro-runtime/issues/2680

also, when switching back to the original module_inst, propagate exception if any.

for a longer term, i think the exeception info should be moved from module_inst to exec_env.
cf. https://github.com/bytecodealliance/wasm-micro-runtime/issues/2512

### DIFF
--- a/core/iwasm/common/wasm_exec_env.c
+++ b/core/iwasm/common/wasm_exec_env.c
@@ -201,7 +201,13 @@ void
 wasm_exec_env_set_module_inst(WASMExecEnv *exec_env,
                               WASMModuleInstanceCommon *const module_inst)
 {
+#if WASM_ENABLE_THREAD_MGR != 0
+    wasm_cluster_traverse_lock(exec_env);
+#endif
     exec_env->module_inst = module_inst;
+#if WASM_ENABLE_THREAD_MGR != 0
+    wasm_cluster_traverse_unlock(exec_env);
+#endif
 }
 
 void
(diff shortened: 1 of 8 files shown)


core/iwasm/common/wasm_exec_env.c
```diff
@@ -201,7 +201,51 @@ void
...
+void
+wasm_exec_env_restore_module_inst(
+    WASMExecEnv *exec_env, WASMModuleInstanceCommon *const module_inst_common)
+{
+    WASMModuleInstanceCommon *old_module_inst_common = exec_env->module_inst;
+    WASMModuleInstance *old_module_inst =
+        (WASMModuleInstance *)old_module_inst_common;
+    WASMModuleInstance *module_inst = (WASMModuleInstance *)module_inst_common;
+    char cur_exception[EXCEPTION_BUF_LEN];
+#if WASM_ENABLE_THREAD_MGR != 0
```
wenyongh, line 222: Had better add an empty line between L221 and L222.

wenyongh (commented):
LGTM

core/iwasm/common/wasm_exec_env.c
yamt: done.

xujuntwt95329 (commented):
LGTM

wenyongh (approved):
LGTM

PR #2697, opened 2023-11-01
Fix AOT compiler issues

wenyongh:


core/iwasm/compilation/aot_emit_numberic.c
```diff
@@ -777,17 +777,25 @@ compile_int_rot(AOTCompContext *comp_ctx, LLVMValueRef left, LLVMValueRef right,
     if (IS_CONST_ZERO(right))
         return left;
 
-    /* Calculate (bits - shif_count) */
+    /* Calculate (bits - shift_count) */
     LLVM_BUILD_OP(Sub, is_i32 ? I32_32 : I64_64, right, bits_minus_shift_count,
                   "bits_minus_shift_count", NULL);
+    /* Calculate (bits - shift_count) & mask */
```
lum1n0us, line 783: not sure about that because of `SHIFT_COUNT_MASK` at L774.

wenyongh: After L774 (right = right & 31), right may be 0, and after L781, shift_count = (32 - right) = 32, shift 32 bits below will get unexpected result (should shift 0 bits instead).

PR #2679, opened 2023-10-28
Enable TSAN for aot and fast interpeter

Zzzabiyaka:
Running CI jobs without thread sanitizer can be risky as we don't have any tool to check possible data races and other problems that might affect the runtime.

Adding these checks for aot and fast interpreter requires less effort than for other mode so can be a good first step into 100% sanitizers coverage. 

in this pr, a couple of warnings are suppressed (see tsan_suppressions.txt for more details) 

**Why do you change runtime_common.h?**

- I change that to get an entrypoint for tsan_suppression, as you can scope it only for a function, and parent function is quite big to scope out 

**Why don't you enable tsan for other modes?**

- Enabling TSAN is tracked in a parent ticket #2248, this PR closes it partially

**Why do you add ubuntu22 jobs to nightly CI while it's already executed in per-pr CI?**
- please see https://bugs.launchpad.net/ubuntu/+source/gcc-9/+bug/2029910. That's why we can't run tsan on ubuntu 20.04 jobs

To run locally: 

```
export TSAN_OPTIONS=suppressions=<path_to_tsan_suppressions.txt>
./test_wamr.sh <your flags> -T tsan
```

an example for wasi-threads would look like:

```
export TSAN_OPTIONS=suppressions=<path_to_tsan_suppressions.txt>
./test_wamr.sh -w -s wasi_certification -t fast-interp -T tsan 
```

PR #2699, opened 2023-11-01
Enable WASI tests on Windows CI

zoraaver:
I've split this up change into two commits to make it (hopefully!) easier to review.
-  the first commit adds the ability to filter tests in the WASI test runner. 
- the second commit actually enables the WASI tests in CI and implements some additional WASI functions on Windows. Without implementing these additional APIs, it's not possible to run any of the tests since most of them need to at least create/delete a file/directory.

### DIFF
--- a/core/iwasm/libraries/lib-socket/test/manifest.json
+++ b/core/iwasm/libraries/lib-socket/test/manifest.json
@@ -1,0 +1,3 @@
+{
+    "name": "WAMR lib-socket tests"
+}
--- a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
+++ b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
@@ -1288,6 +1288,11 @@ path_get(wasm_exec_env_t exec_env, struct fd_table *curfds,
     size_t expansions = 0;
     char *symlink = NULL;
     size_t symlink_len;
+#ifdef BH_PLATFORM_WINDOWS
+#define PATH_SEPARATORS "/\\"
+#else
+#define PATH_SEPARATORS "/"
+#endif
 
     for (;;) {
         // Extract the next pathname component from 'paths[curpath]', null
@@ -1295,9 +1300,10 @@ path_get(wasm_exec_env_t exec_env, struct fd_table *curfds,
         // whether the pathname component is followed by one or more
         // trailing slashes, as this requires it to be a directory.
         char *file = paths[curpath];
-        char *file_end = file + strcspn(file, "/");
-        paths[curpath] = file_end + strspn(file_end, "/");
-        bool ends_with_slashes = *file_end == '/';
+        char *file_end = file + strcspn(file, PATH_SEPARATORS);
+        paths[curpath] = file_end + strspn(file_end, PATH_SEPARATORS);
+        bool ends_with_slashes =
+            (*file_end != '\0' && strchr(PATH_SEPARATORS, *file_end));
         *file_end = '\0';
 
         // Test for empty pathname strings and absolute paths.
@@ -2880,8 +2886,7 @@ __wasi_errno_t
 wasmtime_ssp_sched_yield(void)
 {
 #ifdef BH_PLATFORM_WINDOWS
-    if (!SwitchToThread())
-        return __WASI_EAGAIN;
+    SwitchToThread();
 #else
     if (sched_yield() < 0)
         return convert_errno(errno);
--- a/core/shared/platform/windows/platform_internal.h
+++ b/core/shared/platform/windows/platform_internal.h
@@ -146,8 +146,19 @@ typedef enum windows_access_mode {
     windows_access_mode_write = 1 << 1
 } windows_access_mode;
 
+// These enum values are defined to be the same as the corresponding WASI
+// fdflags so they can be used interchangeably.
+typedef enum windows_fdflags {
+    windows_fdflags_append = 1 << 0,
+    windows_fdflags_dsync = 1 << 1,
+    windows_fdflags_nonblock = 1 << 2,
+    windows_fdflags_rsync = 1 << 3,
+    windows_fdflags_sync = 1 << 4
+} windows_fdflags;
+
 typedef struct windows_handle {
     windows_handle_type type;
+    windows_fdflags fdflags;
     windows_access_mode access_mode;
     union {
         HANDLE handle;
--- a/core/shared/platform/windows/win_file.c
+++ b/core/shared/platform/windows/win_file.c
@@ -7,6 +7,10 @@
 #include "libc_errno.h"
 #include "win_util.h"
 
+#include "PathCch.h"
+
+#pragma comment(lib, "Pathcch.lib")
+
 #define CHECK_VALID_HANDLE_WITH_RETURN_VALUE(win_handle, ret)         \
     do {                                                              \
         if ((win_handle) == NULL                                      \
@@ -188,6 +192,79 @@ get_handle_filepath(HANDLE handle, wchar_t *buf, DWORD buf_size)
     return __WASI_ESUCCESS;
 }
 
+static __wasi_errno_t
+convert_hresult_error_code(HRESULT error_code)
+{
+    switch (error_code) {
+        case E_OUTOFMEMORY:
+            return __WASI_ENOMEM;
+        case E_INVALIDARG:
+        default:
+            return __WASI_EINVAL;
+    }
+}
+
+// Returns the absolute filepath from the relative path to the directory
+// associated with the provided handle.
+static __wasi_errno_t
+get_absolute_filepath(HANDLE handle, const char *relative_path,
+                      wchar_t *absolute_path, size_t buf_len)
+{
+    wchar_t handle_path[PATH_MAX];
+
+    __wasi_errno_t error = get_handle_filepath(handle, handle_path, PATH_MAX);
+
+    if (error != __WASI_ESUCCESS)
+        return error;
+
+    wchar_t relative_wpath[PATH_MAX];
+    error = convert_to_wchar(relative_path, relative_wpath, PATH_MAX);
+
+    if (error != __WASI_ESUCCESS)
+        return error;
+
+    HRESULT ret =
+        PathCchCombine(absolute_path, buf_len, handle_path, relative_wpath);
+    if (ret != S_OK)
+        error = convert_hresult_error_code(ret);
+
+    return error;
+}
+
+static bool
+has_directory_attribute(DWORD attributes)
+{
+    if (attributes == INVALID_FILE_ATTRIBUTES)
+        return false;
+
+    return (attributes & FILE_ATTRIBUTE_DIRECTORY) != 0;
+}
+
+static bool
+is_directory(const wchar_t *path)
+{
+    DWORD attributes = GetFileAttributesW(path);
+
+    return has_directory_attribute(attributes);
+}
+
+static bool
+has_symlink_attribute(DWORD attributes)
+{
+    if (attributes == INVALID_FILE_ATTRIBUTES)
+        return false;
+
+    return (attributes & FILE_ATTRIBUTE_REPARSE_POINT) != 0;
+}
+
+static bool
+is_symlink(const wchar_t *path)
+{
+    DWORD attributes = GetFileAttributesW(path);
+
+    return has_symlink_attribute(attributes);
+}
+
 static void
 init_dir_stream(os_dir_stream dir_stream, os_file_handle handle)
 {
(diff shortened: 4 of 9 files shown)


product-mini/platforms/windows/wasi-filtered-tests.json
```diff
@@ -0,0 +1,41 @@
+{
```
zoraaver, line 1: Open to suggestions on the naming/location of this file. Not sure if it fits here

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -1247,6 +1247,16 @@ path_get(wasm_exec_env_t exec_env, struct fd_table *curfds,
     if (path == NULL)
         return convert_errno(errno);
 
+#ifdef BH_PLATFORM_WINDOWS
```
zoraaver, line 1250: Not strictly related to this PR but I realized we'll need it on Windows so paths with backslashes are expanded properly. It would probably be good to verify the behaviour of backslashes with an additional WASI test.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -2880,8 +2890,7 @@ __wasi_errno_t
 wasmtime_ssp_sched_yield(void)
 {
 #ifdef BH_PLATFORM_WINDOWS
-    if (!SwitchToThread())
-        return __WASI_EAGAIN;
+    SwitchToThread();
```
zoraaver, line 2893: For this change, I'm not sure if we should fail if it doesn't succeed - that implies there are no other threads to switch to which isn't really a failure (https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-switchtothread). Also, [`sched_yield`](https://pubs.opengroup.org/onlinepubs/007904975/functions/sched_yield.html) doesn't define any error codes so I'm not sure callees expect this function call to ever fail.

product-mini/platforms/windows/wasi_filtered_tests.json
```diff
@@ -0,0 +1,41 @@
+{
```
zoraaver, line 1: Open to suggestions on the naming/location of this file. Not sure if it fits here

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
loganek: That looks good to me. Regarding posix's `shed_yield`, I'd keep it as it is. The documentation clearly states that `-1` means error, even though no `errno` code is specified.

product-mini/platforms/windows/wasi_filtered_tests.json
loganek: We don't have any socket tests disabled here although IIRC there's a bunch of functions not implemented. Can we make sure socket tests are actually being executed? Or does it mean we have very low test coverage for sockets?

tests/wamr-test-suites/test_wamr.sh
```diff
@@ -65,7 +67,7 @@ QEMU_FIRMWARE=""
 # prod/testsuite-all branch
 WASI_TESTSUITE_COMMIT="ee807fc551978490bf1c277059aabfa1e589a6c2"
 
-while getopts ":s:cabgvt:m:MCpSXxwPGQF:j:" opt
+while getopts ":s:cabgvt:m:MCpSXxwPGQFf:j:" opt
```
wenyongh, line 70: Could we get the `WASI_TEST_FILTER` from the environment variable, like the `TSAN_OPTIONS` in https://github.com/bytecodealliance/wasm-micro-runtime/pull/2679? A concern is that it is a not so popular config, we had better not use one character for the option, we may add more wasm features in the future, and may need to more characters.

core/shared/platform/windows/win_file.c
```diff
@@ -656,13 +863,79 @@ os_close(os_file_handle handle, bool is_stdio)
...
+        read_operations[i].Offset = query_offset.LowPart;
+        read_operations[i].OffsetHigh = query_offset.HighPart;
+        read_operations[i].hEvent = NULL;
+
+        if (!ReadFileEx(handle, current->buf, (DWORD)current->buf_len,
+                        &read_operations[i], NULL)) {
+            DWORD win_error = GetLastError();
+            if (win_error != ERROR_IO_PENDING) {
+                error = convert_windows_error_code(win_error);
+                break;
```
wenyongh, line 895: Is this an error, shall we check whether `i == iovcnt` or `error` after the for loop to check whether there was an error?

core/shared/platform/windows/win_file.c
```diff
@@ -656,13 +863,79 @@ os_close(os_file_handle handle, bool is_stdio)
...
+            else
+                total_bytes_read += (size_t)bytes_transferred;
+
+            CancelIo(handle);
+
+            for (int j = i + 1; j < iovcnt; ++j) {
+                GetOverlappedResult(handle, &read_operations[j],
+                                    &bytes_transferred, true);
+            }
+            break;
```
wenyongh, line 920: Same as above

core/shared/platform/windows/win_file.c
```diff
@@ -671,7 +944,90 @@ os_readv(os_file_handle handle, const struct __wasi_iovec_t *iov, int iovcnt,
...
+        write_operations[i].Offset = query_offset.LowPart;
+        write_operations[i].OffsetHigh = query_offset.HighPart;
+        write_operations[i].hEvent = NULL;
+
+        if (!WriteFileEx(handle, current->buf, (DWORD)current->buf_len,
+                         &write_operations[i], NULL)) {
+            DWORD win_error = GetLastError();
+            if (win_error != ERROR_IO_PENDING) {
+                error = convert_windows_error_code(win_error);
+                break;
```
wenyongh, line 1002: Same as above

tests/wamr-test-suites/test_wamr.sh
zoraaver: Sure, will change to use an environment variable.

core/shared/platform/windows/win_file.c
zoraaver: Not sure if I'm understanding fully but even if there is an error for one of the read operations, we should still wait on the previous successful read operations. `successful_read_count` is used to make sure we only wait on the operations which were started successfully.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -1288,16 +1288,21 @@ path_get(wasm_exec_env_t exec_env, struct fd_table *curfds,
     size_t expansions = 0;
     char *symlink = NULL;
     size_t symlink_len;
+#ifdef BH_PLATFORM_WINDOWS
+#define PATH_SEPARATORS "/\\"
+#else
+#define PATH_SEPARATORS "/"
+#endif
```
zoraaver, line 1295: Not strictly related to this PR but I realized we'll need it on Windows so paths with backslashes are expanded properly. It would probably be good to verify the behaviour of backslashes with an additional WASI test.

product-mini/platforms/windows/wasi_filtered_tests.json
zoraaver: Thanks for spotting - I didn't realize we needed to compile the socket and thread tests to wasm in CI before executing them. Now we're building them via wasi-sdk and executing them. 

As expected, all (2) socket tests are failing so skipping them for now. Some of the thread tests which use `sleep` also rely on `poll_oneoff` being implemented so skipping those as well.

core/iwasm/libraries/lib-socket/test/manifest.json
```diff
@@ -0,0 +1,3 @@
+{
+    "name": "WASI lib-socket tests"
```
loganek, line 2: I'd call them WAMR lib-socket tests, there's no such thing (yet) as WASI sockets (and even if there was, it's [very different](https://github.com/WebAssembly/wasi-sockets) to the WAMR implementation).

core/shared/platform/windows/platform_internal.h
```diff
@@ -146,8 +146,19 @@ typedef enum windows_access_mode {
     windows_access_mode_write = 1 << 1
 } windows_access_mode;
 
+// These enum values are defined to be the same as the corresponding WASI
+// fdflags so they can be used interchangeably.
+typedef enum windows_fdflags {
```
loganek, line 151: Why do we even define the enum instead of using `__wasi_fdflags_t`? IIRC we already moved it to `platform_wasi.h`

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -1288,16 +1288,22 @@ path_get(wasm_exec_env_t exec_env, struct fd_table *curfds,
     size_t expansions = 0;
     char *symlink = NULL;
     size_t symlink_len;
+#ifdef BH_PLATFORM_WINDOWS
+#define PATH_SEPARATORS "/\\"
```
wenyongh, line 1292: Is it a typo, should use `\\`?

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -1288,16 +1288,22 @@ path_get(wasm_exec_env_t exec_env, struct fd_table *curfds,
...
         // whether the pathname component is followed by one or more
         // trailing slashes, as this requires it to be a directory.
         char *file = paths[curpath];
-        char *file_end = file + strcspn(file, "/");
-        paths[curpath] = file_end + strspn(file_end, "/");
-        bool ends_with_slashes = *file_end == '/';
+        char *file_end = file + strcspn(file, PATH_SEPARATORS);
+        paths[curpath] = file_end + strspn(file_end, PATH_SEPARATORS);
+        bool ends_with_slashes =
+            (*file_end != '\0' && strchr(PATH_SEPARATORS, *file_end));
```
wenyongh, line 1306: How about `*file_end == PATH_SEPARATORS[0]`?

core/shared/platform/windows/win_file.c
```diff
@@ -275,17 +352,17 @@ create_handle(wchar_t *path, bool is_dir, bool follow_symlink, bool readonly)
...
         create_params.dwFileAttributes |= FILE_ATTRIBUTE_READONLY;
-    }
 
     return CreateFile2(path, desired_access,
                        FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE,
                        OPEN_EXISTING, &create_params);
 }
 
-#if WINAPI_PARTITION_DESKTOP
+#if !WINAPI_PARTITION_DESKTOP
```
wenyongh, line 365: Suggest changing to `WINAPI_PARTITION_DESKTOP == 0` to make code easier to understand.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
zoraaver: No, on windows both forward and back slashes are valid path separators e.g.
`dir/nested\\file` should be split into `dir`, `nested` and `file`.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
zoraaver: If there's more than one valid path separator (i.e. Windows) I don't think that would work.
e.g. both `dir1/dir2/dir3/` and `dir1/dir2/dir3\\` should count as `ends_with_slashes == true`.

core/shared/platform/windows/platform_internal.h
loganek: I'm not fully convinced that `platform_wasi.h` should be dependent on `platform_internal.h`. Or if it is, we should probably have another file with just with WASI definitions. 
We could also move the os_file* / os_clock* APIs to `platform_api_extension.h` file.

zoraaver: Yes, that's true. We could separate out the WASI types completely from `platform_wasi.h` into a separate header so that `platform_wasi.h` doesn't depend on `platform_internal.h`.This is what I originally suggested - https://github.com/bytecodealliance/wasm-micro-runtime/pull/2585#discussion_r1363623529 but I think we subsequently agreed to keep the WASI types and os_file_*/os_clock_* declarations in the same header. 

I'm happy to change it so we have a header with just WASI types. - I'm just not sure of the benefit (apart from reducing this small code duplication). As mentioned, in general, there is no reason why `platform_internal.h` should use or depend on WASI types.

@wenyongh perhaps you have some further thoughts on this?

wenyongh: Agree to separate out the WASI types to a file, how about `platform_wasi_types.h`? And should we rename platform_wasi.h?

zoraaver: Ok sure - how about we move the filesystem/clock function declarations to `platform_api_extension.h` as @mkolny suggested and we rename `platform_wasi.h` to `platform_wasi_types.h`? 

Also, I'd rather make these changes in a follow-up PR to avoid squishing them in here as it seems it's a separate issue to enabling the WASI tests in Windows. Then I can remove `windows_fdflags` in the same follow-up PR.

loganek: > Also, I'd rather make these changes in a follow-up PR to avoid squishing them in here as it seems it's a separate issue to enabling the WASI tests in Windows. Then I can remove windows_fdflags in the same follow-up PR.

That sounds reasonable to me.

loganek (approved):
LGTM

core/shared/platform/windows/platform_internal.h
wenyongh: +1

wenyongh (commented):
LGTM

core/shared/platform/windows/platform_internal.h
zoraaver: thanks for approving/merging, have opened https://github.com/bytecodealliance/wasm-micro-runtime/pull/2724 for the WASI type header changes

PR #2707, opened 2023-11-02
Part 2 - exception handling support

raguilar755:
This PR includes updates to exception handling support. This includes:

* Inside the classic interpreter only:
  * Bug fixes and naming convention improvements
  * Additional CI /CD changes to validate ENABLE_EXCE_HANDLING switch builds
    OK on all platforms
  * Import and Export of Exceptions and Tags

### DIFF
--- a/core/iwasm/interpreter/wasm_runtime.c
+++ b/core/iwasm/interpreter/wasm_runtime.c
@@ -276,6 +276,12 @@ memory_instantiate(WASMModuleInstance *module_inst, WASMModuleInstance *parent,
         if (max_page_count > DEFAULT_MAX_PAGES)
             max_page_count = DEFAULT_MAX_PAGES;
     }
+    else { /* heap_size == 0 */
+        if (init_page_count == DEFAULT_MAX_PAGES) {
+            num_bytes_per_page = UINT32_MAX;
+            init_page_count = max_page_count = 1;
+        }
+    }
 
     LOG_VERBOSE("Memory instantiate:");
     LOG_VERBOSE("  page bytes: %u, init pages: %u, max pages: %u",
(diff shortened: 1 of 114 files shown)


core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -726,6 +726,94 @@ functions_instantiate(const WASMModule *module, WASMModuleInstance *module_inst,
...
+#if WASM_ENABLE_FAST_INTERP != 0
+        /* tag->const_cell_num = function->u.func->const_cell_num; */
+#endif
+        tag++;
+    }
+    bh_assert((uint32)(tag - tags) == tag_count);
+
+    return tags;
+fail:
+    wasm_runtime_free(tags);
```
wenyongh, line 812: Had better also destroy the import_tag_ptrs?
```C
    if (module_inst->import_tag_ptrs) {
        wasm_runtime_free(module_inst->import_tag_ptrs);
        module_inst->import_tag_ptrs = NULL;
    }
```

core/iwasm/interpreter/wasm_runtime.h
```diff
@@ -276,6 +310,9 @@ struct WASMModuleInstance {
     /* global and table info */
     uint32 global_data_size;
     uint32 table_count;
+#if WASM_ENABLE_TAGS != 0
+    uint32 tag_count;
```
wenyongh, line 314: This may cause the AOT ABI change: the size of WASMModuleInstance and some fields' offsets should not be changed, or the old AOT file may be unable to run in the new runtime.  We will keep it unchanged until there is a big change. Could you move the fields added in this structure into WASMModuleInstanceExtra? We can use module_inst->e to access them.

raguilar755: Agreed; updating as recommended

core/iwasm/interpreter/wasm_runtime.c
raguilar755: Updated as suggested,

PR #2716, opened 2023-11-03
add gc performance profiling

WenLY1:


core/iwasm/common/wasm_runtime_common.c
```diff
@@ -10,6 +10,7 @@
 #include "wasm_native.h"
 #include "wasm_runtime_common.h"
 #include "wasm_memory.h"
+#include "../../shared/mem-alloc/ems/ems_gc_internal.h"
```
brucehuke, line 13: should it be wrapped with " WASM_ENABLE_GC_PERFORMANCE" ?

core/iwasm/common/wasm_runtime_common.h
```diff
@@ -1071,6 +1071,9 @@ wasm_runtime_invoke_native_raw(WASMExecEnv *exec_env, void *func_ptr,
 void
 wasm_runtime_read_v128(const uint8 *bytes, uint64 *ret1, uint64 *ret2);
 
+void
+wasm_runtime_dump_gc_performance(WASMModuleInstanceCommon *module_inst_common);
```
brucehuke, line 1075: wrap it with "WASM_ENABLE_GC_PERFORMANCE" ?

build-scripts/config_common.cmake
```diff
@@ -296,6 +296,10 @@ if (WAMR_BUILD_JIT_STACK_FRAME EQUAL 1)
   add_definitions (-DWASM_ENABLE_JIT_STACK_FRAME=1)
   message ("     JIT stack frame enabled")
 endif ()
+if (WAMR_BUILD_GC_PERFORMANCE_PROFILING EQUAL 1)
+  add_definitions (-DWASM_ENABLE_GC_PERFORMANCE=1)
```
wenyongh, line 300: Suggest using `WAMR_BUILD_GC_PERF_PROFILING` and `WASM_ENABLE_GC_PERF_PROFILING`

core/iwasm/common/wasm_application.c
```diff
@@ -250,6 +250,10 @@ wasm_application_execute_main(WASMModuleInstanceCommon *module_inst, int32 argc,
     }
 #endif
 
+#if WASM_ENABLE_GC_PERFORMANCE != 0
+    wasm_runtime_dump_gc_performance(module_inst);
```
wenyongh, line 254: How about renaming to `wasm_runtime_dump_gc_perf_profiling`?

core/iwasm/common/wasm_runtime_common.c
wenyongh: Could we add API in the mem allocator and call it instead? Note that we had added a similar API before:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/shared/mem-alloc/mem_alloc.h#L48-L49

core/shared/mem-alloc/ems/ems_alloc.c
```diff
@@ -445,12 +445,27 @@ static int
...
+#endif
     if (heap->is_reclaim_enabled) {
         UNLOCK_HEAP(heap);
         ret = gci_gc_heap(heap);
         LOCK_HEAP(heap);
     }
+#if WASM_ENABLE_GC_PERFORMANCE != 0
+    end = clock();
+    time = end - start;
+    time = time / CLOCKS_PER_SEC;
```
wenyongh, line 462: I guess the time unit here is second, right? Could we use ms or us as the unit, and change the code like:
```C
    uint32 time; (or uint64 time)
    ...
    time = time * 1000 / CLOCK_PER_SEC;  (or time = time * 1000000 / CLOCK_PER_SEC; or us)
```
And in L460, we can use `os_time_get_boot_microsecond` instead, the time unit is us.

core/shared/mem-alloc/ems/ems_gc_internal.h
```diff
@@ -310,7 +310,8 @@ typedef struct gc_heap_struct {
     gc_size_t gc_threshold;
     gc_size_t gc_threshold_factor;
     gc_size_t total_gc_count;
-    gc_size_t total_gc_time;
+    double total_gc_time;
+    double max_gc_time;
```
wenyongh, line 314: How about using uint32 or uint64, and using os_time_get_boot_microsecond to get the time?

core/shared/mem-alloc/ems/ems_alloc.c
```diff
@@ -445,12 +445,25 @@ static int
 do_gc_heap(gc_heap_t *heap)
 {
     int ret = GC_SUCCESS;
+#if WASM_ENABLE_GC_PERF_PROFILING != 0
+    uint64_t start = 0, end = 0, time = 0;
```
wenyongh, line 449: Had better use `uint64` in core/shared, core/iwasm

core/iwasm/common/wasm_application.c
```diff
@@ -250,6 +253,11 @@ wasm_application_execute_main(WASMModuleInstanceCommon *module_inst, int32 argc,
     }
 #endif
 
+#if WASM_ENABLE_GC_PERF_PROFILING != 0
+    void *handle = wasm_runtime_get_gc_heap_handle(module_inst);
+    gc_dump_perf_profiling(handle);
```
wenyongh, line 258: Had better add an API in `core/shared/mem-alloc/mem_alloc.h/c`, for example `mem_allocator_dump_perf_profiling`, which calls `gc_dump_perf_profiling`. We may use different GC allocators in the future, not only ems, and normally runtime only calls API in mem_allocator.

core/iwasm/common/wasm_application.c
```diff
@@ -18,6 +18,9 @@
 #if WASM_ENABLE_STRINGREF != 0
 #include "string_object.h"
 #endif
+#if WASM_ENABLE_GC_PERF_PROFILING
+#include "../../shared/mem-alloc/ems/ems_gc.h"
```
wenyongh, line 22: Should include `mem_alloc.h` instead.

core/iwasm/common/wasm_application.c
```diff
@@ -18,6 +18,9 @@
 #if WASM_ENABLE_STRINGREF != 0
 #include "string_object.h"
 #endif
+#if WASM_ENABLE_GC_PERF_PROFILING
```
wenyongh, line 21: Had better use `WASM_ENABLE_GC_PERF_PROFILING != 0`, we usually use that coding style.

core/shared/mem-alloc/mem_alloc.c
```diff
@@ -127,6 +127,15 @@ mem_allocator_unset_gc_finalizer(mem_allocator_t allocator, void *obj)
 {
     gc_unset_finalizer((gc_handle_t)allocator, (gc_object_t)obj);
 }
+
+#if WASM_ENABLE_GC_PERF_PROFILING != 0
+void
+mem_allocator_dump_perf_profiling(mem_allocator_t allocator)
+{
+    gc_dump_perf_profiling(allocator);
```
wenyongh, line 135: Had better `gc_dump_perf_profiling((gc_handle_t)allocator)`, like above

wenyongh (commented):
LGTM except minor issues

core/shared/mem-alloc/mem_alloc.c
WenLY1: Done

wenyongh (approved):
LGTM

PR #2724, opened 2023-11-06
Move WASI types into separate header

zoraaver:
To allow anything to depend on WASI types, including platform-specific data structures, move the WASI libc filesystem/clock interface into `platform_api_extension.h`, which leaves just WASI types in `platform_wasi.h`. `platform_wasi.h` has been renamed to `platform_wasi_types.h` to reflect that it only defines types now and no function declarations. Finally, these changes allow us to remove the `windows_fdflags` type which was essentially a duplicate of `__wasi_fdflags_t`.


loganek (approved):
LGTM

wenyongh (approved):
LGTM

PR #2725, opened 2023-11-06
Provide default vprintf on UWP

zoraaver:
UWP apps do not have a console attached so any output to stdout/stderr is lost. Therefore, provide a default BH_VPRINTF in that case for debug builds which redirects output to the debugger.

### DIFF
--- a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
+++ b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
@@ -1108,19 +1108,17 @@ wasmtime_ssp_fd_write(wasm_exec_env_t exec_env, struct fd_table *curfds,
     error = blocking_op_writev(exec_env, fo->file_handle, iov, (int)iovcnt,
                                nwritten);
 #else
-    ssize_t len = 0;
     /* redirect stdout/stderr output to BH_VPRINTF function */
     if (fo->is_stdio) {
         int i;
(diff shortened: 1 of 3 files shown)


core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -1108,19 +1108,16 @@ wasmtime_ssp_fd_write(wasm_exec_env_t exec_env, struct fd_table *curfds,
...
-    ssize_t len = 0;
     /* redirect stdout/stderr output to BH_VPRINTF function */
     if (fo->is_stdio) {
-        int i;
-        const struct iovec *iov1 = (const struct iovec *)iov;
-
-        for (i = 0; i < (int)iovcnt; i++, iov1++) {
-            if (iov1->iov_len > 0 && iov1->iov_base) {
+        *nwritten = 0;
+        for (int i = 0; i < (int)iovcnt; i++) {
```
wenyongh, line 1114: Had better declare `int i` after `if (fo->is_stdio) {`, `for (int i ...` is C++ style, normally we don't use that coding style.

wenyongh (commented):
LGTM except minor comment.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
zoraaver: got it, thanks 👍

PR #2727, opened 2023-11-07
Fix more GC AOT/JIT issues

wenyongh:


core/iwasm/common/gc/gc_common.c
```diff
@@ -146,12 +146,22 @@ wasm_obj_get_defined_type_idx(WASMModuleCommon *const module,
                 break;
             }
         }
-
         bh_assert(type_idx < type_count);
     }
 #endif
 #if WASM_ENABLE_AOT != 0
-    /* TODO */
+    if (module->module_type == Wasm_Module_Bytecode) {
```
xujuntwt95329, line 153: Should be `Wasm_Module_AoT`?

core/iwasm/compilation/aot_emit_control.c
```diff
@@ -1325,22 +1325,163 @@ aot_handle_next_reachable_block(AOTCompContext *comp_ctx,
...
 {
     LLVMValueRef gc_obj, value_cmp;
 
-    GET_GC_REF_FROM_STACK(gc_obj);
+    if (!commit_gc_and_check_suspend_flags(comp_ctx, func_ctx, br_depth,
+                                           p_frame_ip)) {
+        return false;
+    }
+
+    POP_GC_REF(gc_obj);
```
xujuntwt95329, line 1471: Why we must pop out the gc_obj here?

core/iwasm/compilation/aot_emit_function.c
```diff
@@ -882,11 +884,7 @@ aot_compile_op_call(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
 
     if (func_type->result_count > 0) {
         /* Push the first result to stack */
-        if (comp_ctx->enable_gc
-            && aot_is_type_gc_reftype(func_type->types[func_type->param_count]))
-            PUSH_GC_REF(value_ret);
-        else
-            PUSH(value_ret, func_type->types[func_type->param_count]);
+        PUSH(value_ret, func_type->types[func_type->param_count]);
```
xujuntwt95329, line 887: For GC ref types, if we don't use `PUSH_GC_REF`, can `aot_value_stack_push` process this type correctly?

core/iwasm/common/gc/gc_common.c
wenyongh: Yes, fixed.

core/iwasm/compilation/aot_emit_control.c
wenyongh: It is to meet the requirements of `compile_gc_cond_br` below, which may take the operands from the stack to the target block (popped from stacks, ADD_TO_PARAM_PHIS, and then pushed to stacks again). Since the operands passed to the target block don't include this gc_obj (it is null when br), we pop it and then push back after `compile_gc_cond_br`.
See also:
https://github.com/WebAssembly/function-references/blob/master/proposals/function-references/Overview.md#optional-references

core/iwasm/compilation/aot_emit_function.c
wenyongh: I converted gc ref types into VALUE_TYPE_GC_REF in PUSH macro, see changes in aot_compiler.h.

core/iwasm/compilation/aot_emit_control.c
xujuntwt95329: Got it, thanks!

core/iwasm/compilation/aot_emit_function.c
xujuntwt95329: Got it, thanks~

core/iwasm/compilation/aot.h
```diff
@@ -177,6 +180,19 @@ typedef struct AOTImportGlobal {
...
+    /*
+     * The data size and data offset of a wasm global may vary
+     * in 32-bit target and 64-bit target, e.g., the size of a
+     * GC obj is 4 bytes in the former and 8 bytes in the
+     * latter, the AOT compiler needs to use the correct data
+     * offset according to the target info.
+     */
+    uint32 size_64bit;
+    uint32 size_32bit;
+    uint32 data_offset_64bit;
```
no1wudi, line 193: Can we merge them? We known the target info at compile time

xujuntwt95329: Seems some functions doesn't have access to target info stored in comp_ctx, if we merge these field, seems we need to modify many API signatures

wenyongh: There may be two possible solutions: one is passing `option.target_arch` to `aot_create_comp_data`, and parsing the target's pointer size according to `option.target_arch`, for example, check whether it contains "x86_64"/"aarch64"/"riscv64", but I am not sure whether it is enough; the other is to set global data size/offset to 0 in `aot_create_comp_data`, and update them again in `aot_create_comp_context` after target info is resolved.

no1wudi: OK, I think we can keep this solution first and refine the usage of target_arch.

no1wudi (approved):

PR #2730, opened 2023-11-07
Fix default vprintf on UWP

zoraaver:
`platform_common.h` already has a declaration for BH_VPRINTF so we can get rid of the one in `platform_internal.h`. Also add some explicit casts to avoid MSVC compiler warnings.

### DIFF
--- a/core/shared/platform/windows/platform_internal.h
+++ b/core/shared/platform/windows/platform_internal.h
@@ -183,9 +183,6 @@ typedef uint32_t os_raw_file_handle;
 // implementation of vprintf on debug builds so output from WASI libc is sent to
 // the debugger and not lost completely.
 #if !defined(BH_VPRINTF) && !defined(NDEBUG) && WINAPI_PARTITION_DESKTOP == 0
-int
-uwp_print_to_debugger(const char *format, va_list ap);
-
 #define BH_VPRINTF uwp_print_to_debugger
 #define UWP_DEFAULT_VPRINTF
 #endif
--- a/core/shared/platform/windows/win_util.c
+++ b/core/shared/platform/windows/win_util.c
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
  */
 
+#include "platform_common.h"
 #include "win_util.h"
 
 __wasi_timestamp_t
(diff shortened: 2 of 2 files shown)


core/shared/platform/windows/platform_internal.h
```diff
@@ -183,9 +183,6 @@ typedef uint32_t os_raw_file_handle;
 // implementation of vprintf on debug builds so output from WASI libc is sent to
 // the debugger and not lost completely.
 #if !defined(BH_VPRINTF) && !defined(NDEBUG) && WINAPI_PARTITION_DESKTOP == 0
-int
-uwp_print_to_debugger(const char *format, va_list ap);
```
wenyongh, line 187: Not sure why remove the declaration? Will it cause compilation warning when compiling  posix.c?

core/shared/platform/windows/win_util.c
```diff
@@ -69,17 +70,17 @@ uwp_print_to_debugger(const char *format, va_list ap)
...
 
-    if (ret >= sizeof(stack_buf)) {
+    if ((size_t)ret >= sizeof(stack_buf)) {
         // Allocate an extra byte for the null terminator.
-        char *heap_buf = BH_MALLOC(ret + 1);
+        char *heap_buf = BH_MALLOC((unsigned int)(ret) + 1);
         buf = heap_buf;
 
         if (heap_buf == NULL) {
             errno = ENOMEM;
```
wenyongh, line 79: Should we use OutputDebugStringA to print the stack_buf (which is part of the whole buffer)?

core/shared/platform/windows/platform_internal.h
zoraaver: There's already a declaration in `platform_common.h`, [here](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/shared/platform/include/platform_common.h#L55-L62). So this one is redundant. `posix.c` doesn't use `BH_VPRINTF` itself but it includes `platform_common.h` indirectly anyway.

core/shared/platform/windows/win_util.c
zoraaver: good idea, will change it to output stack_buf

wenyongh (commented):
LGTM

PR #2739, opened 2023-11-09
Refine mmap usage in iwasm on posix platform

no1wudi:
Use os_mmap will not map file from filesystem to RAM actually.

### DIFF
--- a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
+++ b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
@@ -544,8 +544,14 @@ fd_object_release(wasm_exec_env_t env, struct fd_object *fo)
                 mutex_destroy(&fo->directory.lock);
                 if (os_is_dir_stream_valid(&fo->directory.handle)) {
                     error = os_closedir(fo->directory.handle);
-                    break;
                 }
+                else {
+                    error = (env == NULL)
+                                ? os_close(fo->file_handle, fo->is_stdio)
+                                : blocking_op_close(env, fo->file_handle,
+                                                    fo->is_stdio);
+                }
+                break;
             default:
                 // The env == NULL case is for
                 // fd_table_destroy, path_get, path_put,
@@ -2961,7 +2967,8 @@ argv_environ_init(struct argv_environ_values *argv_environ, char *argv_buf,
 
 void
 argv_environ_destroy(struct argv_environ_values *argv_environ)
-{}
+{
+}
 
 void
 fd_table_destroy(struct fd_table *ft)


product-mini/platforms/posix/main.c
```diff
@@ -934,38 +939,52 @@ main(int argc, char *argv[])
...
-#if WASM_ENABLE_AOT != 0
-    if (wasm_runtime_is_xip_file(wasm_file_buf, wasm_file_size)) {
+    if (fstat(fd, &wasm_file_info) < 0) {
+        printf("Get wasm app file info [%s] failed.\n", wasm_file);
+        goto fail1;
+    }
+
+    /* mmap the wasm file */
+    if ((wasm_file_buf = mmap(NULL, wasm_file_info.st_size,
+                              PROT_READ | PROT_EXEC, MAP_PRIVATE, fd, 0))
```
yamt, line 956: doesn't this break platforms like this?
https://github.com/bytecodealliance/wasm-micro-runtime/blob/797e534cece680fae65734560298ac11fc550a0f/core/shared/platform/nuttx/nuttx_platform.c#L94-L112

anyway, i suspect you should not use mmap directly.
instead, you somehow need to extend the platform api to perform file-mmap where supported (and fallback otherwise), i guess.

no1wudi: OK, Seems extended the os_mmap to supoort map file from given fd is better.
What's your idea? @wenyongh

product-mini/platforms/posix/main.c
```diff
@@ -934,38 +939,52 @@ main(int argc, char *argv[])
...
+        == MAP_FAILED) {
+        printf("Map wasm app file [%s] failed.\n", wasm_file);
+        goto fail1;
+    }
+
+    /* only if the file XIP file, we can use the file buffer directly, for
+     * bytecode file, we need to copy the file buffer
+     * TODO: AOT file can also be used directly if new AOT format is used
```
yamt, line 964: which new AOT format?

if we are changing the format, it would be nicer to allow simpler detection of xip.
like having a small header so that wasm_runtime_is_xip_file only needs to look at it.
(rather than mmap/read the entire file)

no1wudi: Good idea.

The GC feature will comes with breaking changes in dev/gc_refactor branch including above change.

product-mini/platforms/posix/main.c
wenyongh: Yes, normally we should not use mmap directly, instead we had better extend os_mmap and add os_open/os_fstat if needed.

Fortunately, the branch dev/wasi-libc-windows has added API os_open_preopendir and os_fstat in platform_api_extension.h and implemented them in posix platforms and windows platform. See:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/dev/wasi-libc-windows/core/shared/platform/include/platform_api_extension.h#L1210-L1234

And now we are merging branch dev/wasi-libc-windows into main, the PR is under review:
https://github.com/bytecodealliance/wasm-micro-runtime/pull/2740

I think we can wait until the PR is merged and then extend the os_mmap.

PR #2745, opened 2023-11-10
Add eabihf ABI support

no1wudi:


core/iwasm/compilation/aot_llvm.c
```diff
@@ -2510,7 +2523,12 @@ aot_create_comp_context(const AOTCompData *comp_data, aot_comp_option_t option)
              * for Windows/MacOS under Linux host, or generating AOT file for
              * Linux/MacOS under Windows host.
              */
-            if (!strcmp(abi, "msvc")) {
+
+            if (is_baremetal_target(comp_ctx->target_arch, comp_ctx->target_cpu,
+                                    abi)) {
+                vendor_sys = "-none-unknown-";
```
wenyongh, line 2529: Why need to set vendor sys of `thumb` to "-none-unknown-"?

no1wudi: No functional change, just let target triple more close to baremetal platform.

wenyongh: I guess the vendor sys is changed from "-pc-linux-" to "-none-unknown-", not sure whether it affects functionality. The ABI may be changed?

no1wudi: I had a test with binary diff with "-pc-linux-"  and  "-none-unknown-" for a thumb based device, nothing changed in the AOT file.

wenyongh: It may affect the below "msvc" check, for abi == "msvc" and arch == "thumb", the original vendor_sys is "-pc-windows-", but now it is changed to "-none-unknown-". How about adding the check in the else branch:
```C
    if (!strcmp(abi, "msvc")) {
        if (!strcmp(arch1, "i386"))
            vendor_sys = "-pc-win32-";
        else
            vendor_sys = "-pc-windows-";
    }
    else {
        if (is_baremetal_target(arch, cpu, abi))
            vendor_sys = "-unknown-none-";
        else
            vendor_sys = "-pc-linux-";
    }
```
BTW, it seems that the vendor_sys for bare metal should be "-unknown-none-" but not "-none-unknown-"? Refer to:
https://clang.llvm.org/docs/CrossCompilation.html#target-triple
```
The vendor needs to be specified only if there’s a relevant change, for instance between PC and Apple. Most of the time it can be omitted (and Unknown) will be assumed, which sets the defaults for the specified architecture. The system name is generally the OS (linux, darwin), but could be special like the bare-metal “none”.
```

no1wudi: I think the order is ok ? "unknown" for vendor and "none" for system type.

And BTW, does the couple of msvc and thumb is really exists?

wenyongh: > I think the order is ok ? "unknown" for vendor and "none" for system type.

I am not sure, but the document mentions:
```
The vendor needs to be specified only if there’s a relevant change, for instance between PC and Apple.
Most of the time it can be omitted (and Unknown) will be assumed, which sets the defaults for the specified architecture.

The system name is generally the OS (linux, darwin), but could be special like the bare-metal “none”.
```
> 
> And BTW, does the couple of msvc and thumb is really exists?

msvc + thumb may not exist, but we may add more bare metal target in the future? Such as msvc + aarch64?

no1wudi: Fixed

wenyongh (commented):
LGTM

PR #2746, opened 2023-11-10
Fix GC struct field offset calculation in AOT compiler

wenyongh:
The struct field size and field offset of a wasm struct may vary
in 32-bit target and 64-bit target, the aot compiler should not
use the offset calculated in wasm loader. It re-calculates them
according to the target info and whether GC is enabled.

And set the alignment of sruct.get/set when field size is 2, 4, or 8.

### DIFF
--- a/core/iwasm/compilation/aot.c
+++ b/core/iwasm/compilation/aot.c
@@ -384,6 +384,7 @@ fail:
     return NULL;
 }
 
+#if WASM_ENABLE_GC != 0
 static void
 calculate_struct_field_sizes_offsets(AOTCompData *comp_data, bool is_target_x86,
                                      bool gc_enabled)
@@ -431,17 +432,21 @@ calculate_struct_field_sizes_offsets(AOTCompData *comp_data, bool is_target_x86,
         }
     }
 }
+#endif
 
 AOTCompData *
 aot_create_comp_data(WASMModule *module, const char *target_arch,
                      bool gc_enabled)
 {
     AOTCompData *comp_data;
-    bool is_target_x86 = false;
     uint32 import_global_data_size_64bit = 0, global_data_size_64bit = 0, i, j;
     uint32 import_global_data_size_32bit = 0, global_data_size_32bit = 0;
     uint64 size;
+#if WASM_ENABLE_GC != 0
+    bool is_target_x86 = false;
+#endif
 
+#if WASM_ENABLE_GC != 0
     if (!target_arch) {
 #if defined(BUILD_TARGET_X86_64) || defined(BUILD_TARGET_AMD_64) \
     || defined(BUILD_TARGET_X86_32)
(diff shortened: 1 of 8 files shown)


core/iwasm/compilation/aot.c
```diff
@@ -384,13 +384,81 @@ aot_create_funcs(const WASMModule *module)
...
     uint64 size;
+#if WASM_ENABLE_GC != 0
+    bool is_target_x86 = false;
+#endif
+
+#if WASM_ENABLE_GC != 0
+    if (!target_arch) {
+#if defined(BUILD_TARGET_X86_64) || defined(BUILD_TARGET_AMD_64) \
+    || defined(BUILD_TARGET_X86_32)
+        is_target_x86 = true;
```
no1wudi, line 453: Dose this only affect x86? It's better to use a platform independent name for it, such as is_target_64bit?

wenyongh: Yes, the calculations of struct field offset for x86 target and non-x86 target are different:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/dev/gc_refactor/core/iwasm/interpreter/wasm_loader.c#L1189-L1195
https://github.com/bytecodealliance/wasm-micro-runtime/blob/dev/gc_refactor/core/iwasm/aot/aot_loader.c#L1464-L1470

For x86 target, since it supports unaligned memory access, the field data is stored un-aligned.
For non-x86 target, the field data with field size 2 is stored at 2-byte aligned addr, field data with field size no smaller than 4 is stored at 4-byte aligned addr.

no1wudi: If that, should we do align on all platform to unify the code ? Even if x86 support mis-aligned memory access, aligned access may have better performance ?

wenyongh: Using mis-aligned memory access for x86 can save some memory, I am not sure whether it improves the performance. We can do more optimization in the future: put 8-byte fields first, then 4-byte fields, then 2-byte fields and 1-bytes, so all the fields are aligned (8-byte fields may be put at at 4-byte aligned address). Maybe you can help do it if you want?

@xujuntwt95329, do you agree to do align on all platform as @no1wudi suggested?

xujuntwt95329: I'm OK to both aligned and unaligned solution. But we already processed the alignment in code, so I think it's better to keep current solution for now.

BTW, seems the `is_target_x86` include both x86_64 and x86_32 target, would the name be confusing?

wenyongh: is_target_x86 is used in several other places:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/dev/gc_refactor/core/iwasm/compilation/aot_emit_numberic.c#L850-L855
https://github.com/bytecodealliance/wasm-micro-runtime/blob/dev/gc_refactor/core/iwasm/compilation/simd/simd_common.h#L11-L16

Maybe we had better unify them to refine the code.

Issue #2735, opened 2023-11-08
`data.drop` etc should not affect other instances sharing a module


yamt:
currently, `data.drop` instruction is implemented by modifying the underlying module.
it breaks use cases where you have multiple instances sharing a single loaded module.
`elem.drop` seems to have the same problem too.

PR #2747, opened 2023-11-10
fix data/elem drop

yamt:
currently, `data.drop` instruction is implemented by directly modifying the underlying module.
it breaks use cases where you have multiple instances sharing a single loaded module.
`elem.drop` has the same problem too.
this fixes the issue by mirroring the implementation approarch from toywasm: https://github.com/yamt/toywasm/blob/fca52a0174eb1644fea9caa3fe1faad598864f6a/lib/type.h#L488-L494.
that is, keep track of which data/elem segments have been dropped by using bitmaps for each module instances separately.
an obvious drawback of the approach is to consume extra memory for tracking. however,
given the way the wasm spec defines these operations, i suspect this is the only way
to support shared modules. it isn't allowed to simply ignore `data.drop` and `elem.drop` as
we should raise a trap when dropped segments are used.

also, add a sample to demonstrate the issue and make the ci run it.

also, add a missing check of dropped elements to the fast-jit version of `table.init`.

fixes: https://github.com/bytecodealliance/wasm-micro-runtime/issues/2735
fixes: https://github.com/bytecodealliance/wasm-micro-runtime/issues/2772

----

todo:
- ~~fast interpreter~~, ~~aot~~, ~~jit~~, ~~fast-jit~~ (i always feel we have too many variations!)
- ~~investigate `error: unknown import or incompatible import type` seen on  the ci.
   eg. https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/6875997752/job/18700854026?pr=2747~~
- ~~remove JitBitmap (https://github.com/bytecodealliance/wasm-micro-runtime/issues/2754)~~ i ~~will submit a separate PR as this PR is already too big for my taste~~ https://github.com/bytecodealliance/wasm-micro-runtime/pull/2775
- ~~fix a fast-jit specific issue: https://github.com/bytecodealliance/wasm-micro-runtime/issues/2772~~


samples/shared-module/src/main.c
```diff
@@ -0,0 +1,206 @@
...
+    static char global_heap_buf[512 * 1024];
+    char *buffer;
+    char error_buf[128];
+    int opt;
+    char *wasm_path = NULL;
+
+    const unsigned int N = 4;
+    wasm_module_t module = NULL;
+    wasm_module_inst_t module_inst[N];
+    wasm_exec_env_t exec_env[N];
```
wenyongh, line 31: Had better add ` = { 0 }` to initialize module_inst and exec_env, or unexpected behavior may occur when destroying them.

core/iwasm/aot/aot_runtime.c
```diff
@@ -1164,6 +1165,15 @@ aot_instantiate(AOTModule *module, AOTModuleInstance *parent,
     }
 #endif
 
+    common = &((AOTModuleInstanceExtra *)module_inst->e)->common;
+    common->data_dropped = bh_bitmap_new(0, module->mem_init_data_count);
+    common->elem_dropped = bh_bitmap_new(0, module->table_init_data_count);
```
wenyongh, line 1170: Could we allocate memory only when mem_init_data_count/mem_init_data_count is larger than 0? Or there may be a warning:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/common/wasm_memory.c#L245

samples/shared-module/src/main.c
yamt: they are initialized below explicitly.

core/iwasm/interpreter/wasm_runtime.h
```diff
@@ -223,6 +224,8 @@ typedef struct WASMModuleInstanceExtraCommon {
     /* Disable bounds checks or not */
     bool disable_bounds_checks;
 #endif
+    bh_bitmap *data_dropped;
+    bh_bitmap *elem_dropped;
```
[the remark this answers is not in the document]
yamt: done

core/iwasm/aot/aot_runtime.c
yamt: done

samples/shared-module/src/main.c
wenyongh: Got it, thanks.

core/shared/utils/bh_bitmap.c
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright (C) 2021 Intel Corporation.  All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+
+#include <limits.h>
+#include <stddef.h>
+#include <string.h>
```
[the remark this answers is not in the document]
yamt: ok. done.

yamt:
the ci failure doesn't seem related to the changes.
```
  Connection failed [IP: 23.7.100.62 443]
```

Issue #2684, opened 2023-10-31
Feature request: official rust SDK


Mossaka:
Hey 👋, 

I am a core maintainer of the CNCF [containerd/runwasi](https://github.com/containerd/runwasi) project, which facilitates running wasm workloads managed by containerd and Kubernetes. It can be used as a rust library to implement containerd shims who are responsible for managing Wasm containers. We've been asked to add the WebAssembly Micro Runtime support in runwasi as a shim to ensure a wide coverage of runtimes. See issues https://github.com/containerd/runwasi/issues/337 and https://github.com/bytecodealliance/wasm-micro-runtime/issues/2127. 

## Problem Statement
While we're eager to add support for WebAssembly Micro Runtime in runwasi, the lack of an official, up-to-date Rust crate for `wamr` is hindering progress. I am writing to propose **a feature request to publish an official rust SDK for wasm micro runtime**

## Prior Work

There is a community-created rust crate [wamr-sys](https://crates.io/crates/wamr-sys) by @rahul-thakoor on crates.io for #230 . However, this crate is not up to date and lacks active maintenance. A fork by @0xE282B0 adds WASI support, which can be found here: https://github.com/0xE282B0/wamr-sys. 

## Preferred Approach
We prefer a safe Rust SDK over just providing unsafe bindings. This would significantly streamline the integration process and ensure better safety guarantees.

Thank you for considering this feature request. Looking forward to your feedback. 🙏

lum1n0us:
Thanks for contacting with us. 

IMU, the key point is about "a safe Rust SDK over just providing unsafe bindings". I am going to do some research to figure out what does it really means beside using `unsafe` marcos. 😶‍🌫️ 

If there are shareable BKMs and recommanded reading lists, feel free 🤝 .

Mossaka:
I would recommend first to generate the ffi bindings rust bindgen https://github.com/rust-lang/rust-bindgen and then build a SDK on top of it. Here is a safe and ergonomic Rust SDK for Sqlite: https://github.com/rusqlite/rusqlite

0xE282B0:
A minimal bindgen project could look like this: https://github.com/0xE282B0/wasm-micro-runtime/tree/rust-sdk/wamr-sdk-rust

@lum1n0us is it ok to place the rust-sdk in the wamr repo?
Where can I find good examples on how to use the C SDK?

lum1n0us:
@0xE282B0 thanks for your contribution.

[...]

PR #2752, opened 2023-11-11
feat: create rust SDK

0xE282B0:
First commit for the rust-sdk:

- Initial bindgen
- first example using unsafe code

based on discussion in #2684


language-bindings/rust/build.rs
```diff
@@ -0,0 +1,30 @@
...
+
+use cmake::Config;
+use std::{env, path::PathBuf};
+
+fn main() {
+    let mut config = Config::new("../..");
+    let dst = config
+        .generator("Unix Makefiles")
+        .define("WAMR_BUILD_TARGET", env::var("TARGET").unwrap().split("-").next().unwrap().to_uppercase())
+        .build();
```
lum1n0us, line 12: please feel free to correct me.

becuase we can't ask crate users to re-compile,
I guess we need to prepare a compilation options combination here in order to define 
- which features we are going to provide in the final publish crate 
- which running mode we are going to provide in the final publish crate.

0xE282B0: The build.rs is run during the crate installation. It is not uncommon to compile dependencies here.

For the execution mode (like interpreted, AOT, JIT) we can hopefully use feature flags on the crate.

For environments where it is not possible to compile wamr, we could try to "vendor" wamr and include it in the crate.

language-bindings/rust/build.rs
```diff
@@ -0,0 +1,30 @@
+extern crate bindgen;
```
wenyongh, line 1: Should add the license header, note that the license must be "Apache-2.0 WITH LLVM-exception", refer to:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/libraries/lib-wasi-threads/lib_wasi_threads_wrapper.c#L2-L3
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/libraries/debug-engine/debug_engine.c#L2-L3

language-bindings/rust/samples/wasi-hello/src/main.rs
```diff
@@ -0,0 +1,123 @@
+use wamr_sdk_rust::*;
```
wenyongh, line 1: Should add license header.

language-bindings/rust/build.rs
0xE282B0: Sure, makes sense. Does the license header look correct now?
I also added the license to the `Cargo.toml`.

wenyongh: Yes, it looks good. BTW, I didn't find license header in Cargo.toml of this folder (it is added the file of another folder), could you add it?

language-bindings/rust/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "wamr-sdk-rust"
+version = "0.1.0"
+edition = "2021"
```
wenyongh, line 4: Is it better to change to 2023?

language-bindings/rust/samples/wasi-hello/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "wasi-hello"
+version = "0.1.0"
+edition = "2021"
```
wenyongh, line 4: Is it better to change to 2023?

language-bindings/rust/build.rs
```diff
@@ -0,0 +1,35 @@
...
+extern crate cmake;
+
+use cmake::Config;
+use std::{env, path::PathBuf};
+
+fn main() {
+    let mut config = Config::new("../..");
+    let dst = config
+        .generator("Unix Makefiles")
+        .define("WAMR_BUILD_TARGET", env::var("TARGET").unwrap().split("-").next().unwrap().to_uppercase())
```
lum1n0us, line 16: I think we'd better let CMakeFiles to decide the value of `WAMR_BUILD_TAEGET`.

Plus, I am not sure `-G 'Unix Makefiles'` is suitable here. at least, as I know, it won't work on Windows(not with mingw)

language-bindings/rust/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+
```
lum1n0us, line 5: so there will be a `[features]` section here?

language-bindings/rust/samples/wasi-hello/src/main.rs
```diff
@@ -0,0 +1,128 @@
...
+            error_buf.len() as u32,
+        )
+    };
+
+    //let err_u8_vec: Vec<u8> = error_buf.iter().map(|&x| x as u8).filter(|x| *x > 31 && *x < 124).collect();
+    //print!("error {:?}", String::from_utf8(err_u8_vec));
+
+    assert!((module_instance as usize) != 0);
+
+    let _exec_env = unsafe { wasm_runtime_create_exec_env(module_instance, DEFAULT_STACK_SIZE) };
```
wenyongh, line 120: No need to create exec_env if you just to want to call wasm_application_execute_main, in which exec_env is created automatically.

language-bindings/rust/samples/wasi-hello/src/main.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    let success =
+        unsafe { wasm_application_execute_main(module_instance, 0, std::ptr::null_mut()) };
+
+    assert!(success);
+
+    let _main_result = unsafe { wasm_runtime_get_wasi_exit_code(module_instance) };
```
wenyongh, line 127: Not sure whether we should wasm_runtime_deinstantiate(module_inst), wasm_runtime_unload(module) and wasm_runtime_destroy() before return in rust?
See
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/product-mini/platforms/posix/main.c#L978-L999

language-bindings/rust/Cargo.toml
0xE282B0: Editions in rust are used to introduce backward incompatible changes without breaking existing projects; a new edition is not created every year. Currently there are three editions, `2015`, `2018`, and `2021`. New crates created by cargo new are always configured to use the latest stable edition.

There is a guide to rust editions at https://doc.rust-lang.org/edition-guide/editions/index.html

language-bindings/rust/samples/wasi-hello/src/main.rs
0xE282B0: Good to know! I'm not very familiar with the API.
I just realize that the return value is never used. 😅

language-bindings/rust/build.rs
0xE282B0: Oh, I meant I added the license information field:
```toml
  license = "Apache-2.0 WITH LLVM-exception"
```

But I can also leave a license header there 😊

language-bindings/rust/samples/wasi-hello/src/main.rs
0xE282B0: You're right, even if this is just an example, it should clean up in the end.

language-bindings/rust/build.rs
0xE282B0: RE build target: this is a leftover from the crate I forked, I had not tried to remove it.
RE generator: I just have no experience with this, I haven't used windows or mingw in years.

The build process seems to work fine without these two configs, so I remove them.

language-bindings/rust/build.rs
0xE282B0: Oh, right, my bad. There are two `Cargo.toml` and now they should  both have a License header and a license field for the crate:
```toml
  license = "Apache-2.0 WITH LLVM-exception"
```

language-bindings/rust/Cargo.toml
wenyongh: Got it, thanks.

language-bindings/rust/samples/wasi-hello/src/main.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    unsafe {
+        wasm_runtime_deinstantiate(module_instance);
+        wasm_runtime_unload(module);
```
wenyongh, line 129: Should also add `wasm_runtime_destroy()` at the end.

wenyongh (commented):
LGTM

lum1n0us:
LGTM

PR #2762, opened 2023-11-14
Implement GC AOT object reclaim process

wenyongh:


core/iwasm/aot/aot_loader.c
```diff
@@ -1430,14 +1430,33 @@ load_types(const uint8 **p_buf, const uint8 *buf_end, AOTModule *module,
...
+                if (wasm_is_type_reftype(field_type))
+                    ref_field_count++;
+
+                (void)field_flags;
+            }
+
+            struct_type = loader_malloc(
+                sizeof(AOTStructType)
+                    + sizeof(WASMStructFieldType) * (uint64)field_count
+                    + sizeof(uint16) * (uint64)(ref_field_count + 1),
```
xujuntwt95329, line 1458: Why ref_field_count + 1, does it mean reserve one slot to store the `ref_field_count`?

wenyongh: Yes, the first slot is to store the ref_field_count, see:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/dev/gc_refactor/core/iwasm/interpreter/wasm_loader.c#L1147-L1152
and
https://github.com/bytecodealliance/wasm-micro-runtime/blob/dev/gc_refactor/core/iwasm/common/gc/gc_object.c#L737-L743

xujuntwt95329: Got it, thanks!

core/iwasm/compilation/aot_compiler.c
```diff
@@ -785,19 +945,31 @@ aot_compile_func(AOTCompContext *comp_ctx, uint32 func_index)
                 break;
 
             case WASM_OP_BR:
+            {
+                frame_ip_org = frame_ip - 1;
```
xujuntwt95329, line 949: Why we need to `-1` here but in some other opcode (WASM_OP_REF_FUNC) don't `-1`?

core/shared/mem-alloc/ems/ems_gc.c
```diff
@@ -378,8 +379,8 @@ reclaim_instance_heap(gc_heap_t *heap)
                     offset = ref_start_offset + j * sizeof(void *);
                     bh_assert(offset + sizeof(void *) < size);
                     ref = *(gc_object_t *)(((gc_uint8 *)obj) + offset);
-                    if (ref == NULL_REF)
-                        continue; /* NULL REF */
+                    if (ref == NULL_REF || ((uintptr_t)ref & 1))
```
xujuntwt95329, line 382: I think i31ref should not be marked as reference? So it's not possible for i31ref to come here?

core/iwasm/compilation/aot_compiler.c
wenyongh: Here I just want to commit the beginning of the bytecode to stack frame. There are errors in handling REF_FUNC/CALL_REF, I will fix them soon.

xujuntwt95329: Got it. Then the GC related opcode seems need to -2 since their opcode are 2 bytes

core/iwasm/compilation/aot_compiler.c
```diff
@@ -217,6 +217,74 @@ store_value(AOTCompContext *comp_ctx, LLVMValueRef value, uint8 value_type,
         return false;
     }
 
+    LLVMSetAlignment(res, 4);
+
+    return true;
+}
+
+static bool
+store_ref(AOTCompContext *comp_ctx, uint32 ref, LLVMValueRef cur_frame,
```
xujuntwt95329, line 226: Who should be responsible for committing parameters during invoking import function?
Seems the callee won't do this, should we commit the values when calling an import function?

core/shared/mem-alloc/ems/ems_gc.c
wenyongh: The i31 ref in GC AOT compilation is pushed to stack as a GC REF and will be committed to stack frame. And the operands in params/locals/stacks whose type are anyref or eqref can also be i31refs, they may be committed to stack frame also, in the compilation time, the AOT compiler is not easy to know that they are i31ref and doesn't commit ref flag to stack frame.

xujuntwt95329: Got it, thanks!

core/iwasm/compilation/aot_compiler.c
wenyongh: This is handled in aot_alloc_frame by setting the ref arguments' ref flags to 1，and then they are unchanged, no need to commit again. See the changes in aot_alloc_frame and llvm_jit_alloc_frame.

wenyongh: The ref arguments are added to rootset when traversing aot stack frame during GC reclaim.

xujuntwt95329: But this only mark the ref flags, the actual value is not committed to the frame?
I think we should call `aot_gen_commit_values` at https://github.com/bytecodealliance/wasm-micro-runtime/blob/943274e3b90bba1fe2ff34281793b41c3f8b4b7c/core/iwasm/compilation/aot_emit_function.c#L654-L659

otherwise the frame for the native function may contain some dirty value from previous frames

wenyongh: OK, that's really an issue.

core/iwasm/compilation/aot_compiler.c
wenyongh: Yes, `frame_ip_org = frame_ip - 1` was added after reading the GC PREFIX and before reading each sub opcode of GC, see L1319.

core/iwasm/aot/aot_loader.c
```diff
@@ -3856,6 +3886,8 @@ aot_unload(AOTModule *module)
         bh_hash_map_destroy(module->ref_type_set);
     }
     os_mutex_destroy(&module->rtt_type_lock);
+    if (module->rtt_types)
+        wasm_runtime_free(module->rtt_types);
```
xujuntwt95329, line 3890: Should also free every element?

```suggestion
    if (module->rtt_types) {
        uint32 i;
        for (i = 0; i < module->type_count; i++) {
            if (module->rtt_types[i])
                wasm_runtime_free(module->rtt_types[i]);
        }
        wasm_runtime_free(module->rtt_types);
    }
```

wenyongh: Thanks, fixed, and fix issue in commit_params_to_frame_of_import_func.

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -361,11 +367,9 @@ init_frame_refs(uint8 *frame_ref, uint32 cell_num, WASMFunctionInstance *func)
             frame_ref[j++] = 1;
 #endif
         }
-        else if (func->param_types[i] == VALUE_TYPE_I32
-                 || func->param_types[i] == VALUE_TYPE_F32)
-            j++;
-        else
-            j += 2;
+        else {
+            j += wasm_value_type_size(func->param_types[i]);
```
xujuntwt95329, line 371: Should use `wasm_value_type_cell_num`

wenyongh: Done, thanks.

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -324,11 +324,9 @@ init_frame_refs(uint8 *frame_ref, uint32 cell_num, WASMFunctionInstance *func)
             frame_ref[j++] = 1;
 #endif
         }
-        else if (func->param_types[i] == VALUE_TYPE_I32
-                 || func->param_types[i] == VALUE_TYPE_F32)
-            j++;
-        else
-            j += 2;
+        else {
+            j += wasm_value_type_size(func->param_types[i]);
```
xujuntwt95329, line 328:
```suggestion
            j += wasm_value_type_cell_num(func->param_types[i]);
```

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -339,11 +337,9 @@ init_frame_refs(uint8 *frame_ref, uint32 cell_num, WASMFunctionInstance *func)
             frame_ref[j++] = 1;
 #endif
         }
-        else if (func->local_types[i] == VALUE_TYPE_I32
-                 || func->local_types[i] == VALUE_TYPE_F32)
-            j++;
-        else
-            j += 2;
+        else {
+            j += wasm_value_type_size(func->local_types[i]);
```
xujuntwt95329, line 341:
```suggestion
            j += wasm_value_type_cell_num(func->local_types[i]);
```

core/iwasm/aot/aot_runtime.c
```diff
@@ -3125,8 +3125,32 @@ aot_alloc_frame(WASMExecEnv *exec_env, uint32 func_index)
...
+                frame_ref[j++] = 1;
+#endif
+            }
+            else {
+                value_type_cell_num =
+                    wasm_value_type_cell_num(func_type->types[i]);
+                for (k = 0; k < value_type_cell_num; k++)
+                    frame_ref[j++] = 0;
+            }
+        }
```
xujuntwt95329, line 3152: if func_type->param_count == 0, then the reserved 2 slots in frame_ref may remain uninitialized. Seems we should clear the frame_ref[0] and frame_ref[1] in this case:

```suggestion
        }
        
        if (func_type->param_count == 0) {
            frame_ref[0] = frame_ref[1] = 0;
        }
```

core/iwasm/interpreter/wasm_interp_fast.c
wenyongh: Done.

core/iwasm/aot/aot_runtime.c
wenyongh: Done.

PR #2763, opened 2023-11-14
Extend os_mmap to support map file from file handle

no1wudi:


core/shared/platform/common/posix/posix_memmap.c
```diff
@@ -37,7 +37,7 @@ round_down(uintptr_t v, uintptr_t b)
 #endif
 
 void *
-os_mmap(void *hint, size_t size, int prot, int flags)
+os_mmap(void *hint, size_t size, int prot, int flags, os_file_handle *file)
```
wenyongh, line 40: Per my understanding, os_file_handle is already a handle and can be used inside the function body. And for the caller, if it wants to pass the invalid file handle, it can pass `os_get_invalid_handle()` to os_mmap: in posix, os_get_invalid_handle returns -1, and in windows, it returns NULL.

core/iwasm/aot/aot_loader.c
```diff
@@ -1575,7 +1575,7 @@ load_object_data_sections(const uint8 **p_buf, const uint8 *buf_end,
         /* Allocate memory for data */
         if (data_sections[i].size > 0
             && !(data_sections[i].data = os_mmap(NULL, data_sections[i].size,
-                                                 map_prot, map_flags))) {
+                                                 map_prot, map_flags, NULL))) {
```
wenyongh, line 1578: Can pass `os_get_invalid_handle()` instead of `NULL` if we declare os_mmap as `os_mmap(..., os_file_handle file)`

core/shared/platform/common/posix/posix_memmap.c
no1wudi: Get it

core/iwasm/aot/aot_loader.c
no1wudi: Fixed

core/shared/platform/common/posix/posix_file.c
```diff
@@ -52,6 +54,17 @@
 #define CONFIG_HAS_O_SYNC
 #endif
 
+// Provides a dummy definition which is not actually used,
+// but is required to compile.
+#if WASM_ENABLE_LIBC_WASI == 0
+__wasi_errno_t
+convert_errno(int error)
```
zoraaver, line 61: Previously we were only compiling this file if wasi libc was enabled - has that changed? I think that would make more sense since all the functions in this file are only used when wasi libc is enabled. It also means we don't need to add the stub which never gets used.

core/shared/platform/windows/shared_platform.cmake
```diff
@@ -13,9 +13,7 @@ include_directories(${PLATFORM_SHARED_DIR}/../include)
 file (GLOB_RECURSE source_all ${PLATFORM_SHARED_DIR}/*.c
 			      ${PLATFORM_SHARED_DIR}/*.cpp)
 
-if (NOT WAMR_BUILD_LIBC_WASI EQUAL 1)
-    list(REMOVE_ITEM source_all ${PLATFORM_SHARED_DIR}/win_file.c)
```
zoraaver, line 17: Similar to the previous comment, I'm not completely sure why we're changing this? This file only has the implementation of functions used in wasi libc so compiling it isn't necessary when it isn't enabled and it means we have to add a stub for `convert_errno`.

core/shared/platform/common/posix/posix_file.c
no1wudi: In this PR, I'm trying to use `os_file_handle ` to let the `os_memmap` support map memory from file, and then use the ` os_get_invalid_handle()` as invalid file handle.

But the `os_get_invalid_handle()` introduced the usage of `convert_errno()`. Add '-ffunction-section` to compile flag is another approach  but we should not depend on specifal compiler flag .

core/shared/platform/windows/shared_platform.cmake
no1wudi: See above.

no1wudi:
@wenyongh What's your opinion? Should we add stub function here or promote `os_get_invalid_handle()` and some other basic file operation to porting layer (bh_xxx), and use them in wasi libc? Seems the actually difference between the native version and os_xx verison is the errno.

BTW, if we promote the file operation to porting layer, maybe we can unify the iwasm implementation for all platform.

core/shared/platform/common/posix/posix_file.c
wenyongh: The `convert_errorno` has been implemented in `core/shared/platform/common/libc-util/libc_errno.c/h`:
  https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/shared/platform/common/libc-util/libc_errno.c#L9-L10
No need to add it again here. I think maybe we should add `libc_error.c` into file list in Nuttx's makefile.

BTW, had better not change `shared_platform.cmake` of windows platform.

core/shared/platform/common/posix/posix_file.c
```diff
@@ -7,6 +7,8 @@
 #include "libc_errno.h"
 #include <unistd.h>
 
+#include "bh_platform.h"
```
wenyongh, line 10: No need to add this line if the added `convert_errno` in this file is removed, or, keep this file unchanged.

core/shared/platform/windows/win_file.c
```diff
@@ -7,10 +7,23 @@
...
+
 #include "PathCch.h"
 
 #pragma comment(lib, "Pathcch.lib")
 
+// Provides a dummy definition which is not actually used,
+// but is required to compile.
+#if WASM_ENABLE_LIBC_WASI == 0
+__wasi_errno_t
+convert_errno(int error)
```
wenyongh, line 20: Same as above, had better not change this file.

core/shared/platform/windows/shared_platform.cmake
wenyongh: Had better not change this file, convert_errno is already implemented in libc_errno.c?

core/shared/platform/common/posix/posix_file.c
no1wudi: This issue is exists on all platform, if [posix | win]_file.c compiled，then we must add libc_errno.c to build system.

core/shared/platform/windows/shared_platform.cmake
no1wudi: Yes, but that means we must compile it unconditionally .

core/shared/platform/riot/riot_platform.c
```diff
@@ -50,8 +50,12 @@ os_dumps_proc_mem_info(char *out, unsigned int size)
 }
 
 void *
-os_mmap(void *hint, size_t size, int prot, int flags)
+os_mmap(void *hint, size_t size, int prot, int flags, os_file_handle *file)
```
wenyongh, line 53: Should be `os_file_handle file`

core/shared/platform/linux-sgx/sgx_platform.c
```diff
@@ -120,13 +120,20 @@ strcpy(char *dest, const char *src)
...
+os_mmap(void *hint, size_t size, int prot, int flags, os_file_handle file)
 {
     int mprot = 0;
     uint64 aligned_size, page_size;
     void *ret = NULL;
     sgx_status_t st = 0;
 
+    if (os_is_handle_valid(&file)) {
+        os_printf("os_mmap(size=%u, prot=0x%x, file=%x) failed: file is not "
+                  "supported.",
```
wenyongh, line 132: Had better add `\n` for os_printf

core/shared/platform/riot/riot_platform.c
```diff
@@ -50,8 +50,12 @@ os_dumps_proc_mem_info(char *out, unsigned int size)
...
 {
+    if (file) {
```
wenyongh, line 55: Should we add `os_is_handle_valid` for riot platform and call it here?

core/shared/platform/esp-idf/espidf_memmap.c
```diff
@@ -19,8 +19,13 @@ static portMUX_TYPE s_spinlock = portMUX_INITIALIZER_UNLOCKED;
 #endif
 
 void *
-os_mmap(void *hint, size_t size, int prot, int flags)
+os_mmap(void *hint, size_t size, int prot, int flags, os_file_handle file)
 {
+    if (os_is_handle_valid(&file)) {
```
wenyongh, line 24: Seems that os_is_handle_valid isn't implemented in espidf platform

no1wudi: Should we implement `os_is_handle_valid()` for these platforms ?

wenyongh: Not necessarily I think, just let someone implement it when it is actually needed?

no1wudi: OK

core/shared/platform/alios/platform_internal.h
```diff
@@ -68,4 +68,10 @@ typedef int os_file_handle;
 typedef DIR *os_dir_stream;
 typedef int os_raw_file_handle;
 
+static inline os_file_handle
+os_get_invalid_handle()
+{
+    return -1;
+};
```
wenyongh, line 75: Had better remove unused `;`

core/shared/platform/android/platform_internal.h
```diff
@@ -150,6 +150,12 @@ typedef int os_file_handle;
 typedef DIR *os_dir_stream;
 typedef int os_raw_file_handle;
 
+static inline os_file_handle
+os_get_invalid_handle()
+{
+    return -1;
+};
```
wenyongh, line 157: ditto

core/shared/platform/darwin/platform_internal.h
```diff
@@ -113,6 +113,12 @@ typedef int os_file_handle;
 typedef DIR *os_dir_stream;
 typedef int os_raw_file_handle;
 
+static inline os_file_handle
+os_get_invalid_handle()
+{
+    return -1;
+};
```
wenyongh, line 120: ditto and same in some other files

wenyongh (commented):
LGTM except a minor issue

no1wudi:
@zoraaver Could you review again?

PR #2766, opened 2023-11-15
Add compilation flag to enable/disable heap corruption check

wenyongh:
Heap corruption check in ems memory allocator is enabled by default
to improve the security, but it may impact the performance a lot, this
PR adds cmake variable and compiler flag to enable/disable it.


xujuntwt95329 (approved):
LGTM

no1wudi (approved):

PR #2768, opened 2023-11-15
Add clangcl compiler option for Windows CI builds

zoraaver:


zoraaver:
I was struggling to run the spec tests on windows locally (still not sure what the issue was, but seemed to be python related) so opened this PR to run the wasi/spec tests on a WAMR build using clang-cl. All the tests seemed to have passed so closing this now.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/random.c
```diff
@@ -49,19 +49,20 @@ random_buf(void *buf, size_t len)
...
-    if (!crypt_initialized) {
-        CryptAcquireContext(&provider, NULL, NULL, PROV_RSA_FULL,
-                            CRYPT_VERIFYCONTEXT);
-        crypt_initialized = 1;
+    NTSTATUS ret = BCryptGenRandom(NULL, buf, (ULONG)len,
+                         BCRYPT_USE_SYSTEM_PREFERRED_RNG);
+    if (BCryptGenRandom(NULL, buf, (ULONG)len,
+                         BCRYPT_USE_SYSTEM_PREFERRED_RNG)) {
+        os_printf("BCryptGenRandom failed");
+        abort();
```
loganek, line 64: What was the purpose for this change?

zoraaver: Opened a separate PR for that, https://github.com/bytecodealliance/wasm-micro-runtime/pull/2769 - I only noticed when compiling WAMR with uwp toolchain since `CryptGenRandom` isn't available on UWP and then when I opened the windows docs, I noticed it was deprecated and there's a newer version which is recommended so thought I might as well fix it.

PR #2769, opened 2023-11-15
Use next generation crypto API on Windows

zoraaver:
CryptGenRandom is deprecated by Microsoft and may be removed in future
releases. They recommend to use the next generation API instead. See
https://learn.microsoft.com/en-us/windows/win32/seccng/cng-portal for
more details. Also, refactor the random functions to return error codes
rather than aborting the program if they fail.

### DIFF
--- a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/random.c
+++ b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/random.c
@@ -13,14 +13,16 @@
 
 #include "ssp_config.h"
 #include "bh_platform.h"
+#include "libc_errno.h"
 #include "random.h"
 
 #if CONFIG_HAS_ARC4RANDOM_BUF
 
-void
+__wasi_errno_t
 random_buf(void *buf, size_t len)
 {
     arc4random_buf(buf, len);
+    return __WASI_ESUCCESS;
 }
 
 #elif CONFIG_HAS_GETRANDOM
@@ -29,7 +31,7 @@ random_buf(void *buf, size_t len)
 #include <sys/random.h>
 #endif
 
-void
+__wasi_errno_t
 random_buf(void *buf, size_t len)
 {
     for (;;) {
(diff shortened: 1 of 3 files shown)


wenyongh:
@loganek could you help review this PR? Shall I merge it? Thanks.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/random.c
```diff
@@ -49,19 +49,19 @@ random_buf(void *buf, size_t len)
...
 random_buf(void *buf, size_t len)
 {
-    static int crypt_initialized = 0;
-    static HCRYPTPROV provider;
-    if (!crypt_initialized) {
-        CryptAcquireContext(&provider, NULL, NULL, PROV_RSA_FULL,
-                            CRYPT_VERIFYCONTEXT);
-        crypt_initialized = 1;
+    NTSTATUS ret =
+        BCryptGenRandom(NULL, buf, (ULONG)len, BCRYPT_USE_SYSTEM_PREFERRED_RNG);
```
loganek, line 59: according to the [doc](https://learn.microsoft.com/en-us/windows/win32/api/bcrypt/nf-bcrypt-bcryptgenrandom#return-value) the function fails on invalid algorithm's handle or invalid parameter. The former shouldn't be an issue because it's `NULL` in our case, but I the latter is vaguely defined, and I'm not 100% what it means (e.g. is there any min/max length of the buffer?). For that reason, I'd suggest changing the signature of the `random_buf` to return an error and instead of crashing through `abort()` return an error code (that can be then propagated to WASM in the `wasmtime_ssp_random_get()`.

zoraaver: good idea, was thinking the same thing since crashing might be overkill if the function fails

zoraaver: changed to return an error code

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/random.h
```diff
@@ -14,8 +14,10 @@
 #ifndef RANDOM_H
 #define RANDOM_H
 
-void
+__wasi_errno_t
```
loganek, line 17: Should there also be an `include` directive for this type? Otherwise you'll depend on the order of including the file, which usually isn't a good practice.

loganek (approved):

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/random.h
zoraaver: thanks for spotting, added an include for `bh_platform.h`

wenyongh (commented):
LGTM

PR #2771, opened 2023-11-16
aot compiler: Disable musttail for thumb

wenyongh:
Error is reported when executing `wamrc --target=thumb -o <aot_file> <wasm_file>`:
```
LLVM ERROR: failed to perform tail call elimination on a call site marked musttail
Aborted (core dumped)
```


yamt (approved):
lgtm

PR #2774, opened 2023-11-16
Fix printing of external reference addresses in   wasm_application.c

no1wudi:


core/iwasm/common/wasm_application.c
```diff
@@ -613,7 +613,7 @@ execute_func(WASMModuleInstanceCommon *module_inst, const char *name,
                 u.parts[1] = argv1[k + 1];
                 k += 2;
                 if (u.val && u.val != (uintptr_t)-1LL)
-                    os_printf("0x%" PRIxPTR ":ref.extern", (void *)u.val);
+                    os_printf("%p:ref.extern", (void *)u.val);
```
yamt, line 616: u.val is uintptr_t. why don't you remove the cast instead?

core/iwasm/common/wasm_application.c
```diff
@@ -600,7 +600,7 @@ execute_func(WASMModuleInstanceCommon *module_inst, const char *name,
             {
 #if UINTPTR_MAX == UINT32_MAX
                 if (argv1[k] != 0 && argv1[k] != (uint32)-1)
-                    os_printf("0x%" PRIxPTR ":ref.extern", (void *)argv1[k]);
+                    os_printf("%p:ref.extern", (void *)argv1[k]);
```
yamt, line 603: iirc, how `%p` converts is implementation-defined.
isn't it a problem for repl?

no1wudi:
@yamt In fact I'm fine to cast them to uintptr_t but it's from stdint.h, but other type such as uint32 is defined internally, mix usage of them looks a little strange.

yamt (approved):
lgtm

PR #2777, opened 2023-11-16
Fix compilation errors on zephyr platform

wenyongh:


core/shared/platform/zephyr/platform_internal.h
```diff
@@ -74,6 +74,10 @@ typedef korp_thread *korp_tid;
 typedef struct k_mutex korp_mutex;
 typedef unsigned int korp_sem;
 
+/* korp_rwlock is used in platform_api_extension.h,
+   we just define the type to make the compiler happy */
+typedef struct k_mutex korp_rwlock;
```
yamt, line 79: why do you use k_mutex?
i suspect it's safer to use a dummy struct.

wenyongh: The platform_api_extension.h declares os_rwlock_xxx APIs, but they aren't implemented and really used in zephyr, here I just make the compiler work.
Using dummy struct is also OK. How about `typedef struct {} korp_rwlock;`?

yamt: probably `typedef struct { int dummy; } korp_rwlock;` as plain C doesn't have empty structure.

or, maybe it's cleaner to introduce BH_HAS_RWLOCK?

wenyongh: OK, I used the dummy struct first. Will check how to introduce BH_HAS_RWLOCK.

wenyongh: Only several platforms have korp_rwlock and os_dir_stream definition issues, prefer to modify them directly and not modify a lot of files (all platform_internal.h and the platform_api_extension.h), and not introduce new macro as there are a lot of macros already.

yamt (approved):
lgtm

PR #2779, opened 2023-11-16
Fix logging verbose output on test failure

loganek:


tests/wamr-test-suites/spec-test-script/all.py
```diff
@@ -245,7 +245,7 @@ def test_case(
 
             if p.returncode:
                 print(f"failed with a non-zero return code {p.returncode}")
-                if not verbose_flag:
+                if verbose_flag:
```
yamt, line 248: this change doesn't seem to make sense to me.

my reading of the original code is:
* if verbose_flag is true, all output is printed above.
* otherwise, it's printed only on failure.

yamt (changes requested):

PR #2780, opened 2023-11-16
Refine atomic operation flags in bh_atomic

no1wudi:
This also fix the build break in spec test of NuttX.

### DIFF
--- a/core/shared/utils/bh_atomic.h
+++ b/core/shared/utils/bh_atomic.h
@@ -47,6 +47,19 @@ extern "C" {
 typedef uint32 bh_atomic_32_t;
 typedef uint16 bh_atomic_16_t;
 
+/* The flag can be defined by the user if the platform
+ * supports atomic 32-bit operations.
+ * If left undefined, it will be automatically defined
+ * according to the platform.
+ */
+#ifdef WASM_UINT32_IS_ATOMIC
+#define BH_ATOMIC_32_IS_ATOMIC WASM_UINT32_IS_ATOMIC
+#endif /* WASM_UINT32_IS_ATOMIC */
+
+#ifdef WASM_UINT16_IS_ATOMIC
+#define BH_ATOMIC_16_IS_ATOMIC WASM_UINT16_IS_ATOMIC
+#endif /* WASM_UINT16_IS_ATOMIC */
+
 #if defined(__GNUC_PREREQ)
 #if __GNUC_PREREQ(4, 7)
 #define CLANG_GCC_HAS_ATOMIC_BUILTIN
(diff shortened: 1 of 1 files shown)


core/shared/utils/bh_atomic.h
```diff
@@ -59,6 +72,31 @@ typedef uint16 bh_atomic_16_t;
...
+#endif
+
+#ifndef BH_ATOMIC_32_IS_ATOMIC
+#define BH_ATOMIC_32_IS_ATOMIC 0
+#endif
+#ifndef BH_ATOMIC_16_IS_ATOMIC
+#define BH_ATOMIC_16_IS_ATOMIC 0
+#endif
+
+#if BH_ATOMIC_32_IS_ATOMIC == 1
```
wenyongh, line 99: Had better use `!= 0`

core/shared/utils/bh_atomic.h
```diff
@@ -69,7 +107,16 @@ typedef uint16 bh_atomic_16_t;
...
+#else /* else of BH_ATOMIC_32_IS_ATOMIC == 1 */
+#define BH_ATOMIC_32_LOAD(v) (v)
+#define BH_ATOMIC_32_STORE(v, val) (v) = val
+#define BH_ATOMIC_32_FETCH_OR(v, val) nonatomic_32_fetch_or(&(v), val)
+#define BH_ATOMIC_32_FETCH_AND(v, val) nonatomic_32_fetch_and(&(v), val)
+#define BH_ATOMIC_32_FETCH_ADD(v, val) nonatomic_32_fetch_add(&(v), val)
+#define BH_ATOMIC_32_FETCH_SUB(v, val) nonatomic_32_fetch_sub(&(v), val)
+#endif
 
+#if BH_ATOMIC_16_IS_ATOMIC == 1
```
wenyongh, line 119: Same as above.

core/shared/utils/bh_atomic.h
```diff
@@ -69,7 +107,16 @@ typedef uint16 bh_atomic_16_t;
     __atomic_fetch_add(&(v), (val), __ATOMIC_SEQ_CST)
 #define BH_ATOMIC_32_FETCH_SUB(v, val) \
     __atomic_fetch_sub(&(v), (val), __ATOMIC_SEQ_CST)
+#else /* else of BH_ATOMIC_32_IS_ATOMIC == 1 */
+#define BH_ATOMIC_32_LOAD(v) (v)
+#define BH_ATOMIC_32_STORE(v, val) (v) = val
+#define BH_ATOMIC_32_FETCH_OR(v, val) nonatomic_32_fetch_or(&(v), val)
+#define BH_ATOMIC_32_FETCH_AND(v, val) nonatomic_32_fetch_and(&(v), val)
+#define BH_ATOMIC_32_FETCH_ADD(v, val) nonatomic_32_fetch_add(&(v), val)
+#define BH_ATOMIC_32_FETCH_SUB(v, val) nonatomic_32_fetch_sub(&(v), val)
```
wenyongh, line 116: Should move the below implementations of nonatomic_32_fetch_or/nonatomic_32_fetch_and/nonatomic_32_fetch_add/nonatomic_32_fetch_sub here, they should not be wrapped by `BH_ATOMIC_16_IS_ATOMIC`?

no1wudi: Sure

wenyongh (commented):
LGTM

yamt:
i have a concern in the original change which introduced atomic 16.
https://github.com/bytecodealliance/wasm-micro-runtime/pull/2682#discussion_r1396499660

core/shared/utils/bh_atomic.h
```diff
@@ -58,7 +71,37 @@ typedef uint16 bh_atomic_16_t;
...
+
+/* Force disable atomic 16-bit operations on bare-metal RISC-V
+ * because the 16-bit atomic operations is emulated by 32-bit
+ * atomic operations, which has linkage problem on current toolchain:
+ * in function `shared_memory_inc_reference':
+ * wasm_shared_memory.c:85:(.text.shared_memory_inc_reference+0x10): undefined
+ * reference to `__atomic_fetch_add_2'
+ */
+#ifndef WASM_UINT16_IS_ATOMIC
+#if !defined(__linux__) && defined(__riscv)
```
yamt, line 97: what do you mean by `!defined(__linux__)` here?
do you mean `defined(__NuttX__)`?

also, why do you override it here, rather than having it in platform-specific headers?

no1wudi: I guess all non-linux/baremetal platforms has the same issue, not only nuttx, since the usage of the baremetal toolchain have nothing difference with other RTOS platform.

And you can also define the behavior in platfrom-specific setting, explicit settings always have the highest priority.

yamt: how about freebsd?

no1wudi: I think the bsd family should be OK as linux.

yamt (approved):
lgtm

PR #2783, opened 2023-11-17
More precise help info of enabled targets for wamrc

TianlongLiang:
Instead of printing all support targets of wamrc, print only the target that is included in the LLVM library with which wamrc was compiled.


wenyongh (commented):
LGTM

Issue #2782, opened 2023-11-16
Make Zephyr port work as a Zephyr module


hasheddan:
There is currently a [Zephyr module file](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/zephyr/module.yml) in this repository, but it cannot be used because it references `cmake-ext` and `kconfig-ext`, despite `CMakeLists.txt` and `Kconfig` not being present for WAMR in [Zephyr upstream modules](https://github.com/zephyrproject-rtos/zephyr/tree/main/modules). There was discussion about including upstream in https://github.com/zephyrproject-rtos/zephyr/issues/21329, but it appears that the decision for now is for it to live separately. With that being the case, we should make it easy for Zephyr users to pull WAMR in as an external module. This primarily just consists of adding an appropriate `CMakeLists.txt` and `Kconfig` in the WAMR repository, then pointing to them via the `zephyr/module.yml` file.

I'd be happy to make these updates -- @wenyongh what do you think?

wenyongh:
@hasheddan Yes, that's great! Shall we add the `CMakeLists.txt` and `Kconfig` under `<wamr_root>/zephry` folder? There is a `CMakeLists.txt` under `<wamr_root>` already.

hasheddan:
@wenyongh either there or https://github.com/bytecodealliance/wasm-micro-runtime/tree/main/core/shared/platform/zephyr is great with me -- do you have a preference? I'll get started adding it!

wenyongh:
@hasheddan had better put them under `<wamr_root>/zephry`, `core/shared/platform/zephyr` may be too deep and not easy to maintain.

hasheddan:
@wenyongh will do, thanks!

nguterresn:
@hasheddan hi!
I stumbled upon the same issue as you — when including the WAMR in the same west.yml as the other modules, the Zephyr looks up for the Kconfig but it doesn't exist.
Were you able to fix this?

FabienCesari:
Hey there! Same question as @nguterresn - I can't find the `Kconfig `and `CMakeLists.txt` @hasheddan Was this implemented? Thanks!

FabienCesari:
I figured out how to get it to build without the Kconfig and CMakeLists.txt - You should not import WAMR as a zephyr module from the yaml file but instead clone the repo in the module folder. You then will have to modify your project CMakeLists.txt to add the path to the WAMR source and relevant config flags.

PR #2805, opened 2023-11-20
Update Zephyr support to v3.5.0 and make instructions generic to boards

hasheddan:
Includes a number of updates to the Zephyr example, including removing a fair amount of stale configuration. This is part of moving towards supporting WAMR as a Zephyr module as described in #2782. Some functionality is removed as part of this diff (e.g. AOT XTENSA support), but it had become stale while not being actively maintained. It is anticipated that it (and AOT support for other platforms) will be added back, along with CI support for ensuring they do not become stale again.

See individual commits for more specific information about changes.

cc @wenyongh


product-mini/platforms/zephyr/simple/README.md
```diff
@@ -1,106 +1,82 @@
...
+[`west`](https://docs.zephyrproject.org/latest/develop/west/index.html). There
+are already [configuaration files](./boards) for a few boards in this sample.
+However, if you are using a new board, you will need to add your own file for
+the board, or define configuration in the [`prj.conf](./prj.conf). After doing
+so, use the following command with your board identifier to build the sample
+application.
 
 ```shell
-sudo apt-get install qemu
+west build . -b <board-identifier> --pristine -- -DWAMR_BUILD_TARGET=<wamr-arch>
```
TianlongLiang, line 59: Maybe it's better to add a step to copy the `boards/<board>.conf` to `prj.conf` so that a newbie like me won't forget? 😄

no1wudi:
@TianlongLiang Maybe we should not use `fence.i` directly since it has been removed from rv32i standard, I'll submit a fix.

TianlongLiang:
> @TianlongLiang Maybe we should not use `fence.i` directly since it has been removed from rv32i standard, I'll submit a fix.

Good advice, let’s choose this path

no1wudi:
@TianlongLiang Could you try this patch?

https://github.com/bytecodealliance/wasm-micro-runtime/pull/2807

TianlongLiang:
> @TianlongLiang Could you try this patch?
> 
> #2807

Hi, @no1wudi I tried the patch, and still have this error, it seems that the target attribute wasn't set successfully, I am not sure whether this is because of the compiler:
```bash
/root/wasm-micro-runtime/core/iwasm/aot/arch/aot_reloc_riscv.c:188:1: warning: target attribute is not supported on this machine [-Wattributes]
  188 | {
      | ^
/root/wasm-micro-runtime/core/iwasm/aot/arch/aot_reloc_riscv.c: Assembler messages:
/root/wasm-micro-runtime/core/iwasm/aot/arch/aot_reloc_riscv.c:192: Error: unrecognized opcode `fence.i'
/root/wasm-micro-runtime/core/iwasm/aot/arch/aot_reloc_riscv.c:192: Error: unrecognized opcode `fence.i'
```
The Zephyr is using the gcc in its sdk:
```bash
-- Found Git: /usr/bin/git (found version "2.34.1") 
-- The C compiler identification is GNU 12.2.0
-- The CXX compiler identification is GNU 12.2.0
-- The ASM compiler identification is GNU
-- Found assembler: /opt/zephyr-sdk-0.16.3/riscv64-zephyr-elf/bin/riscv64-zephyr-elf-gcc
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /opt/zephyr-sdk-0.16.3/riscv64-zephyr-elf/bin/riscv64-zephyr-elf-gcc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /opt/zephyr-sdk-0.16.3/riscv64-zephyr-elf/bin/riscv64-zephyr-elf-g++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Building ESP-IDF components for target esp32c3
```

no1wudi:
> > @TianlongLiang Could you try this patch?
> > #2807
> 
> Hi, @no1wudi I tried the patch, and still have this error, it seems that the target attribute wasn't set successfully, I am not sure whether this is because of the compiler:
> 

Seems the zephyr's toolchain is not support this attribute, please try the update again.

TianlongLiang:
> > > @TianlongLiang Could you try this patch?
> > > #2807
> > 
> > 
> > Hi, @no1wudi I tried the patch, and still have this error, it seems that the target attribute wasn't set successfully, I am not sure whether this is because of the compiler:
> 
> @TianlongLiang Seems the zephyr's toolchain is not support this attribute, please try the update again.

@no1wudi Now it can be built successfully 😄

product-mini/platforms/zephyr/simple/README.md
hasheddan: Ah great catch! I meant to just had an empty `prj.conf`, which should alleviate this issue :) I may also move some of the config that is not specific to a single board into the `prj.conf` 👍🏻

hasheddan:
@TianlongLiang @no1wudi thanks for y'all's reviews! I don't think we need to necessarily remove the `fence.i` instruction. It just happens that `esp32c3_devkitm` does not support the `zifencei` instruction, but some other targets, such as QEMU, do support `zifencei`. This PR mentions that AOT must be disabled on some targets, and provides the necessary config to do so:

> It and other
options can be disabled by modifying the [CMakeLists.txt](./CMakeLists.txt)
file, or by passing additional arguments at build (e.g. `-DWAMR_BUILD_AOT=0`).

Using `west build -b esp32c3_devkitm . --pristine -- -DWAMR_BUILD_TARGET=RISCV32_ILP32 -DWAMR_BUILD_AOT=0` should work correctly 🙂

hasheddan:
@TianlongLiang @no1wudi I have updated to include the `prj.conf` here as well. I don't think we necessarily need to block on https://github.com/bytecodealliance/wasm-micro-runtime/pull/2807 given that supplying `-DWAMR_BUILD_AOT=0` is described in the instructions. I'm not entirely confident that just removing the `fence.i` instruction means that AOT will work correctly on these targets, even if it does compile. I think it makes sense to validate that as a separate effort.

TianlongLiang:
> @TianlongLiang @no1wudi I have updated to include the `prj.conf` here as well. I don't think we necessarily need to block on #2807 given that supplying `-DWAMR_BUILD_AOT=0` is described in the instructions. I'm not entirely confident that just removing the `fence.i` instruction means that AOT will work correctly on these targets, even if it does compile. I think it makes sense to validate that as a separate effort.

You are right about we need to test the actual run on the target to validate that it actually works. Unfortunately, I don't have a hardware test environment to test it. Maybe we can mention the current AOT situation for `esp32c3_devkitm` in README or somewhere for future user reference. They can try the default setting with AOT enabled first, and if it does not work on AOT, they can use other running modes. What's your suggestion?

product-mini/platforms/zephyr/simple/build_and_run.sh
wenyongh, line 1: Could you not delete this file? It is important for us to test whether WAMR runs OK on some qemus.

product-mini/platforms/zephyr/simple/README.md
```diff
@@ -1,106 +1,82 @@
...
 ```
 
-### Run the build script
+The `<board-identifier>` can be found in the Zephyr supported boards
+documentation. It must also be used as the name of the board configuration file.
+You must define the architecture for WAMR to target with `WAMR_BUILD_TARGET`.
+The list of supported architectures can be found in the main project
+[README.md](../../../../README.md#supported-architectures-and-platforms).
```
wenyongh, line 66: Could you add some examples (e.g. different boards) about how to use the above command?

product-mini/platforms/zephyr/simple/prj.conf
```diff
@@ -1,4 +1,4 @@
-# Copyright (C) 2019 Intel Corporation.  All rights reserved.
+# Copyright (C) 2023 Golioth, Inc.
```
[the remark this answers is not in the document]
hasheddan: This is showing as a changed line, but this is actually a net new file that is replacing the same config being used across all the board config files in `boards/`. I am happy to copy over the Intel copyright if preferred though -- let me know 👍🏻

product-mini/platforms/zephyr/simple/README.md
hasheddan: Sure! For context, I would like to follow this PR (and some other updates to the Zephyr support) with CI that tests at least building for a set of target boards that we want to verify. For now, I'll just add some example build commands though 👍🏻

product-mini/platforms/zephyr/simple/build_and_run.sh
hasheddan: @wenyongh the absence of this file shouldn't prohibit testing against QEMU targets. They can be built and run along with any other target -- is adding some examples for using QEMU targets in the `README.md` sufficient here? Also, I don't believe that there are currently any CI tests for Zephyr -- is that right?

hasheddan:
> Maybe we can mention the current AOT situation for esp32c3_devkitm in README or somewhere for future user reference. They can try the default setting with AOT enabled first, and if it does not work on AOT, they can use other running modes. What's your suggestion?

@TianlongLiang added `esp32c3_devkitm` as an example build target with a note about it not currently supporting AOT 👍🏻

product-mini/platforms/zephyr/simple/build_and_run.sh
hasheddan: Updated `README.md` with information about how to use emulated targets 👍🏻

wenyongh: My preference is not to remove this file, it is really convenient and import for us to test all the targets we want to support, please modify it if needed.

product-mini/platforms/zephyr/simple/prj.conf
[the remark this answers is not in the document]
hasheddan: It has the same config of some of the board files previously did, but is new and gets combined with the remaining board files when a target is matched. I have used the Intel copyright header for now though 👍🏻

product-mini/platforms/zephyr/simple/build_and_run.sh
hasheddan: I have dropped the commit that removed it and updated it to use latest Zephyr updates (including dropping external toolchain references, fixing missing config, and removing no longer supported targets).

hasheddan (commented):
@wenyongh thanks for your feedback! I believe I have addressed all comments, but let me know if there is anything missing :)

product-mini/platforms/zephyr/simple/build_and_run.sh
```diff
@@ -17,11 +16,10 @@ QEMU_ARC_TARGET="qemu_arc"
 usage ()
 {
         echo "USAGE:"
-        echo "$0 $X86_TARGET|$STM32_TARGET|$ESP32_TARGET|$ESP32C3_TARGET|$PARTICLE_ARGON_TARGET|$QEMU_CORTEX_A53|$QEMU_XTENSA_TARGET|$QEMU_RISCV64_TARGET|$QEMU_RISCV32_TARGET|$QEMU_ARC_TARGET"
+        echo "$0 $X86_TARGET|$STM32_TARGET|$ESP32C3_TARGET|$PARTICLE_ARGON_TARGET|$QEMU_CORTEX_A53|$QEMU_XTENSA_TARGET|$QEMU_RISCV64_TARGET|$QEMU_RISCV32_TARGET|$QEMU_ARC_TARGET"
```
wenyongh, line 19: Not sure why the esp32 target is removed, is it because that esp32c3 is more popular and no need to support esp32 again, or here esp32c3 target is also suitable for esp32 target?

no1wudi: Maybe this is a typo, esp32 is xtensa based, esp32c3 is risc-v based, they are different.

wenyongh: If so, had better not remove esp32, including changes in src/main.c, boards/esp32.conf and esp32_custom_linker.ld. We had better support both.

hasheddan: @wenyongh @no1wudi there are _many_ different esp32 variants and just `esp32` is not a valid target in Zephyr anymore, so build will fail. I would like to work on ensuring that all esp32 variants build correctly, but that will be a larger effort.

wenyongh: @hasheddan got it, thanks, so we had better merge this PR first. Could you help enable the esp32 support after "Make Zephyr port work as a Zephyr module" as mentioned in #2782?

@no1wudi how do you think?

hasheddan: > Could you help enable the esp32 support after "Make Zephyr port work as a Zephyr module" as mentioned in #2782?

@wenyongh absolutely!

no1wudi: @wenyongh I'm fine with it

wenyongh: OK, merged.

PR #2806, opened 2023-11-21
Fix an export name check exception

lum1n0us:


yamt (changes requested):
this fix doesn't seem right.
if we want to support names including '\00', we should stop using strlen etc.

lum1n0us:
we are thinking about using an extra `length` field to store the length of a string, in our case an unicode string.

PR #2807, opened 2023-11-21
Fix Zifencei extension issue in RISC-V spec

no1wudi:


core/iwasm/aot/arch/aot_reloc_riscv.c
```diff
@@ -170,14 +170,32 @@ rv_get_val(uint16 *addr)
...
 static void
 rv_set_val(uint16 *addr, uint32 val)
 {
     *addr = (val & 0xffff);
     *(addr + 1) = (val >> 16);
 
+    /* REVISIT: Zephyr's GCC toolchain does not support attribute
+     * target("arch=+zifencei") and on MCU fence.i may equal to nop, so we can
+     * disable it here.
+     */
```
hasheddan, line 195: I believe Zephyr's toolchain does support `zifencei`, but some targets do not support the extension. Have you found that not to be the case?

no1wudi: I mean zephyr's toolchain should support zifencei extension but don't support enable it by `__attribute__((target("arch=+zifencei")))`, consider as an RTOS, the fence.i on micro controller may not actually sense, so can be ignored here.

no1wudi:
@yamt Could you take a look?

yamt:
> @yamt Could you take a look?

do you mean someone else?
i'm not familiar with risc-v or zephyr.

core/iwasm/aot/arch/aot_reloc_riscv.c
hasheddan: I believe just some targets don't support it (such as `esp32c3-devkitm`). However, other Zephyr targets do support it (example: https://github.com/zephyrproject-rtos/zephyr/blob/23cf38934c0f68861e403b22bc3dd0ce6efbfa39/soc/riscv/riscv-privileged/sifive-freedom/Kconfig.soc#L18).

no1wudi: Thanks, does `RISCV_ISA_EXT_ZIFENCEI` visible for us? How can we check it?

hasheddan: It is a Kconfig option, so it would be available as `CONFIG_RISCV_ISA_EXT_ZIFENCEI` (e.g. https://github.com/zephyrproject-rtos/zephyr/blob/23cf38934c0f68861e403b22bc3dd0ce6efbfa39/cmake/compiler/gcc/target_riscv.cmake#L48). Let me see if there is a more portable option here though 👍🏻

no1wudi:
> > @yamt Could you take a look?
> 
> do you mean someone else? i'm not familiar with risc-v or zephyr.

OK, sorry for disturbing you.

core/iwasm/aot/arch/aot_reloc_riscv.c
no1wudi: If so, let's check this option first.

core/iwasm/aot/arch/aot_reloc_riscv.c
```diff
@@ -170,14 +170,32 @@ rv_get_val(uint16 *addr)
...
 static void
 rv_set_val(uint16 *addr, uint32 val)
 {
     *addr = (val & 0xffff);
     *(addr + 1) = (val >> 16);
 
+    /* REVISIT: Zephyr's GCC toolchain does not support attribute
+     * target("arch=+zifencei") so we must check if the extension is
+     * enabled at runtime.
+     */
```
hasheddan, line 195: I don't think it is the case that the toolchain does not support the attribute. Rather, if a target does not support `zifencei` then the attribute will not be available. This is true whether using the Zephyr provided toolchain or another toolchain. Also, I don't think we are evaluating at runtime, the check performed below is evaluated at build time.

no1wudi: I belive the toolchain support this attribute but don't support enable it by grammer `__attribute__(target("arch=+zifenci"))`, zephyr's toolchain will report:
```
/root/wasm-micro-runtime/core/iwasm/aot/arch/aot_reloc_riscv.c:188:1: warning: target attribute is not supported on this machine [-Wattributes]
  188 | {
      | ^
/root/wasm-micro-runtime/core/iwasm/aot/arch/aot_reloc_riscv.c: Assembler messages:
/root/wasm-micro-runtime/core/iwasm/aot/arch/aot_reloc_riscv.c:192: Error: unrecognized opcode `fence.i'
/root/wasm-micro-runtime/core/iwasm/aot/arch/aot_reloc_riscv.c:192: Error: unrecognized opcode `fence.i'
```
But it supported by other toolchain such as xpack's or official toolchain since this grammer is supported in spec:
https://github.com/riscv-non-isa/riscv-c-api-doc/blob/73877888d0d0e0a3a938b811d81a212a5edc4e3c/riscv-c-api.md?plain=1#L209

no1wudi: Or the given target is not support the zifenci, could you share the `-march` parameter from zephyr?

hasheddan: > Or the given target is really don't support the zifenci extension, could you share the -march parameter from zephyr?

This is correct -- the `esp32c3_devkitm` target does not support the `zifencei` extension, but others do, such as `qemu_riscv64`.

core/iwasm/aot/arch/aot_reloc_riscv.c
```diff
@@ -170,14 +170,32 @@ rv_get_val(uint16 *addr)
...
+#if defined(__ZEPHYR__) && defined(CONFIG_RISCV_ISA_EXT_ZIFENCEI)
```
hasheddan, line 196: Also note that GCC / Clang define the `__riscv_zifencei` symbol when supported: https://github.com/riscv-non-isa/riscv-c-api-doc/blob/master/riscv-c-api.md#architecture-extension-test-macro, which may be a more portable option here.

hasheddan:
@no1wudi I don’t have full context on the path in which this functionality is being exercised, but if instruction ordering is important here, I imagine we may want to use a [`fence`](https://danielmangum.com/risc-v-tips/2021-12-28-memory-ordering-fence/) when [`fence.I`](https://danielmangum.com/risc-v-tips/2021-12-29-memory-ordering-fencei/) is not available.

no1wudi:
> @no1wudi I don’t have full context on the path in which this functionality is being exercised, but if instruction ordering is important here, I imagine we may want to use a [`fence`](https://danielmangum.com/risc-v-tips/2021-12-28-memory-ordering-fence/) when [`fence.I`](https://danielmangum.com/risc-v-tips/2021-12-29-memory-ordering-fencei/) is not available.

I guess it's OK for most application, `fence` and `fence.i` used to solve different coherent problem in multiple hart, but there are no data race in this case.

wenyongh:
> > @no1wudi I don’t have full context on the path in which this functionality is being exercised, but if instruction ordering is important here, I imagine we may want to use a [`fence`](https://danielmangum.com/risc-v-tips/2021-12-28-memory-ordering-fence/) when [`fence.I`](https://danielmangum.com/risc-v-tips/2021-12-29-memory-ordering-fencei/) is not available.
> 
> I guess it's OK for most application, `fence` and `fence.i` used to solve different coherent problem in multiple hart system, but there are no data race in this case.

Sounds good, so let's merge this PR? @no1wudi, @hasheddan

no1wudi:
> Sounds good, so let's merge this PR? @no1wudi, @hasheddan

It's OK for me, if a risc-v CPU don't implement `zifencei` extension, it may not be a complex multi core system, such as esp32c3, it's a single core MCU.

hasheddan:
@wenyongh @no1wudi I was curious as to whether we should use `fence` (part of integer instruction set) when the Zifencei extension is not supported. I am happy to investigate that as a follow up if preferred though.

wenyongh:
> @wenyongh @no1wudi I was curious as to whether we should use `fence` (part of integer instruction set) when the Zifencei extension is not supported. I am happy to investigate that as a follow up if preferred though.

Is `fence` instruction always available in all riscv architectures? Will it cause compilation error for some architectures?

no1wudi:
> @wenyongh @no1wudi I was curious as to whether we should use `fence` (part of integer instruction set) when the Zifencei extension is not supported. I am happy to investigate that as a follow up if preferred though.

@hasheddan Thanks if you have time to do this!

The most important reason I'm so casual for this issue is that for many risc-v based system, `fence.i` and `fence` is do nothing in their implementation, just equal to `nop`, so we see the `zifencei` moved from rv32i/rv64i as a dedicated optional extension.

no1wudi:
> > @wenyongh @no1wudi I was curious as to whether we should use `fence` (part of integer instruction set) when the Zifencei extension is not supported. I am happy to investigate that as a follow up if preferred though.
> 
> Is `fence` instruction always available in all riscv architectures? Will it cause compilation error for some architectures?

`fence` supported on all risc-v platform.

wenyongh (commented):
LGTM

PR #2814, opened 2023-11-23
fix GC aot on 32bit target

xujuntwt95329:


core/iwasm/aot/aot_runtime.c
```diff
@@ -3151,9 +3151,11 @@ aot_alloc_frame(WASMExecEnv *exec_env, uint32 func_index)
             }
         }
 
-        if (func_type->param_count == 0) {
+        if (func_type->param_cell_num < 2) {
             /* We reserve at least two cells for native frame */
-            frame_ref[0] = frame_ref[1] = 0;
+            for (; j < 2; j++) {
+                frame_ref[j] = 0;
+            }
```
wenyongh, line 3158: In 32-bit, when param_cell_num = 1 and the first param is GC ref, frame_ref[0] should be 1? How about:
```C
    for (j = func_type->param_cell_num; j < 2; j++) {
         frame_ref[j] = 0;
    }
```

core/iwasm/compilation/aot.c
```diff
@@ -466,6 +511,19 @@ aot_create_comp_data(WASMModule *module, const char *target_arch,
     }
 #endif
 
+    if (!target_arch) {
+#if defined(BUILD_TARGET_X86_64) || defined(BUILD_TARGET_AMD_64)            \
+    || defined(BUILD_TARGET_AARCH64) || defined(BUILD_TARGET_RISCV64_LP64D) \
+    || defined(BUILD_TARGET_RISCV64_LP64)
+        is_64bit_target = true;
+#endif
```
wenyongh, line 519: How about:
```C
#if UINTPTR_MAX == UINT64_MAX
    is_64bit_target = true;
#endif
```

core/iwasm/compilation/aot_emit_variable.c
```diff
@@ -83,7 +83,7 @@ aot_compile_op_set_or_tee_local(AOTCompContext *comp_ctx,
 
     if (comp_ctx->aot_frame) {
         /* Get the slot index */
-        n = comp_ctx->aot_frame->cur_wasm_func->local_offsets[local_idx];
```
wenyongh, line 86: Could we also remove cur_wasm_func field in AOTCompFrame? Seems it is only used here.

core/iwasm/compilation/aot.c
```diff
@@ -466,6 +511,19 @@ aot_create_comp_data(WASMModule *module, const char *target_arch,
...
+    }
+    else {
+        if (strstr(target_arch, "64") != NULL) {
```
wenyongh, line 522: Could we add comment about why we check it with "64"?

core/iwasm/aot/aot_runtime.c
xujuntwt95329: Done, thanks!

core/iwasm/compilation/aot.c
xujuntwt95329: Good idea! updated.

core/iwasm/compilation/aot_emit_variable.c
xujuntwt95329: Yes, removed

core/iwasm/compilation/aot.c
xujuntwt95329: Done

core/iwasm/compilation/aot_llvm.h
```diff
@@ -114,8 +114,6 @@ typedef struct AOTValueSlot {
 typedef struct AOTCompFrame {
     /* The current wasm module */
     WASMModule *cur_wasm_module;
-    /* The current wasm function */
-    WASMFunction *cur_wasm_func;
     /* The current wasm function index */
     uint32 cur_wasm_func_idx;
```
wenyongh, line 118: Can also remove cur_wasm_func_idx, it isn't used also.

wenyongh (commented):
LGTM

core/iwasm/compilation/aot_llvm.h
xujuntwt95329: Removed, thanks

PR #2815, opened 2023-11-23
Export API wasm_runtime_shared_mem_lock/unlock

wenyongh:
Fixes: https://github.com/bytecodealliance/wasm-micro-runtime/issues/2804

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -2386,6 +2386,20 @@ aot_data_drop(AOTModuleInstance *module_inst, uint32 seg_index)
 }
 #endif /* WASM_ENABLE_BULK_MEMORY */
 
+#if WASM_ENABLE_SHARED_MEMORY != 0
+void
+aot_shared_mem_lock(AOTMemoryInstance *memory)
+{
+    shared_memory_lock(memory);
+}
+
+void
+aot_shared_mem_unlock(AOTMemoryInstance *memory)
+{
+    shared_memory_unlock(memory);
+}
+#endif
+
 #if WASM_ENABLE_THREAD_MGR != 0
 bool
 aot_set_aux_stack(WASMExecEnv *exec_env, uint32 start_offset, uint32 size)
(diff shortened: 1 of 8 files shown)


core/iwasm/aot/aot_runtime.c
```diff
@@ -2386,6 +2386,20 @@ aot_data_drop(AOTModuleInstance *module_inst, uint32 seg_index)
 }
 #endif /* WASM_ENABLE_BULK_MEMORY */
 
+#if WASM_ENABLE_SHARED_MEMORY != 0
+void
+aot_shared_mem_lock(AOTMemoryInstance *memory)
```
g0djan, line 2391: Is the purpose of the wrapper to pass "AOTMemoryInstance" without a cast?

wenyongh: Not sure which to cast? Here the function will be called from AOT code before calling aot_memmove and aot_memset, so as to lock the shared memory to avoid data racing in the two functions.

g0djan: My question was more like why not use `shared_mem_lock/shared_mem_lock` without wrapping it in `aot_shared_mem_lock/aot_shared_mem_unlock`

wenyongh: That may be also a choice, for example, change aot_memmove/aot_memset to be like below:
```C
    os_mutex_lock(&g_shared_memory_lock);
#endif

    memmove/memset(...);

    os_mutex_unlock(&g_shared_memory_lock);
#endif
```
A concern is that it will always acquire the lock when shared memory feature is enabled, no matter whether the memory is actual shared or not. In current code, we normally assert the memory is not NULL, and acquire the lock only when the memory is shared:
```C
#define shared_memory_lock(memory)                                            \
    do {                                                                      \
        /*                                                                    \
         * Note: exception logic is currently abusing this lock.              \
         * cf.                                                                \
         * https://github.com/bytecodealliance/wasm-micro-runtime/issues/2407 \
         */                                                                   \
        bh_assert(memory != NULL);                                            \
        if (memory->is_shared_memory)                                         \
            os_mutex_lock(&g_shared_memory_lock);                             \
    } while (0)
```
So I add lock before calling the aot_memmove function. Maybe the former solution is already good, the latter introduces function aot_shared_mem_lock/unlock for AOT code, and might be not so friendly for the old AOT file generated.

loganek:
I'm not against the API but just wonder if there's any usecase right now for this?

wenyongh:
> I'm not against the API but just wonder if there's any usecase right now for this?

Me too, how about not merging it until there is actual requirements?

loganek:
Sounds good to me

PR #2816, opened 2023-11-23
enable stringref aot

xujuntwt95329:


core/iwasm/aot/aot_loader.c
```diff
@@ -871,6 +871,59 @@ load_name_section(const uint8 *buf, const uint8 *buf_end, AOTModule *module,
...
+
+    read_uint32(p, p_end, string_count);
+    if (string_count == 0) {
+        set_error_buf(error_buf, error_buf_size,
+                      "invalid string literal count");
+        goto fail;
+    }
+    module->string_literal_count = string_count;
+
+    size = sizeof(char *) * string_count;
```
wenyongh, line 899: Had better `(uint64)sizeof(char *)` to avoid possible integer overflow.

core/iwasm/aot/aot_loader.c
```diff
@@ -871,6 +871,59 @@ load_name_section(const uint8 *buf, const uint8 *buf_end, AOTModule *module,
...
+    if (!(module->string_literal_ptrs =
+              loader_malloc(size, error_buf, error_buf_size))) {
+        goto fail;
+    }
+
+    size = sizeof(uint32) * string_count;
```
wenyongh, line 905: ditto

core/iwasm/compilation/aot_emit_stringref.c
```diff
@@ -0,0 +1,1452 @@
...
+        goto fail;
+    }
+
+    if (!(str_obj_ptr = LLVMBuildBitCast(comp_ctx->builder, str_obj_ptr,
+                                         GC_REF_PTR_TYPE, "str_obj_gcref_p"))) {
+        aot_set_last_error("llvm build bitcast failed.");
+        goto fail;
+    }
+
+    if (!(str_obj = LLVMBuildLoad2(comp_ctx->builder, GC_REF_TYPE, str_obj_ptr,
```
wenyongh, line 180: Not sure whether the data is stored at n-byte aligned address, had better add `LLVMSetAlignment` for LLVMBuildLoad2 and LLVMBuildStore?

core/iwasm/aot/aot_loader.c
xujuntwt95329: Done, thanks

core/iwasm/compilation/aot_emit_stringref.c
xujuntwt95329: The store is finished in runtime APIs, it should be 4bytes aligned, I've add the `LLVMSetAlignment`
Thanks

PR #2819, opened 2023-11-24
Fix wamr-test-suites script issue for macos

wenyongh:
Update the wabt release binary name for macos and fix a syntax error:

```bash
./test_wamr.sh: line 356: syntax error near unexpected token `;'
./test_wamr.sh: line 356: `                    ;&'
```


yamt (approved):
lgtm

Issue #2530, opened 2023-09-05
AOT ABI reminder


yamt:
there are a few things which we can clean up when we break AOT ABI.
this issue is a reminder for them.

* merge WASMModuleInstanceExtraCommon into WASModuleInstance
* https://github.com/bytecodealliance/wasm-micro-runtime/blob/48b71a05fbbee26524e50a5d1df0ca77cb39e580/core/iwasm/common/wasm_shared_memory.c#L16-L17
* https://github.com/bytecodealliance/wasm-micro-runtime/issues/2469

PR #2821, opened 2023-11-24
Update version number to 1.3.0 and update release notes

wenyongh:
And refine the code format for wasm_export.h.


RELEASE_NOTES.md
```diff
@@ -1,3 +1,155 @@
+## WAMR-1.3.0
```
yamt, line 1: just curious; why bumped to 1.3.0, instead of 2.0.0 or 1.2.4?
i feel our numbering is a bit arbitrary.

wenyongh: Hi, when we introduced the semantic version mechanism, there was a decision made for what each part of the version num means and how we should release the new version (see https://github.com/bytecodealliance/wasm-micro-runtime/commit/717e8a48e29be35a768a5d9f3c992d5f2b47fbaa and https://github.com/bytecodealliance/wasm-micro-runtime/issues/1357):
```
There are three parts in the new version string:
- major. Any incompatible modification on ABIs and APIs will lead to an increment
  in the value of major, which mainly includes: AOT calling conventions, AOT file
  format, wasm_export.h, wasm_c_api.h, and so on.
- minor. It represents new features, including MVP/POST-MVP features, libraries,
  WAMR private ones, and so one.
- patch. It represents patches.
```
The next release includes some new features (multi-module AOT, libc-wasi for windows, module instance context APIs and so on), but doesn't break the AOT ABI, so per my understanding (and discussed with @lum1n0us and @xujuntwt95329), we should increase the minor version number. There are several changes in wasm_export.h, but I think they are minor changes. For the versions released before, we also tried to obey the above rules. And we may merge branch dev/gc_refactor into main in the future to upstream the GC feature (which was planned before and was asked many times by some developers), we will upgrade the version number to 2.0.0 since it changes the AOT file format and some other AOT ABIs, and I think at that time we can also resolve the issue https://github.com/bytecodealliance/wasm-micro-runtime/issues/2530.

PR #2822, opened 2023-11-24
Refine the GC frame ref flag commit

wenyongh:


core/iwasm/compilation/aot_emit_control.c
```diff
@@ -173,20 +173,56 @@ clear_frame_locals(AOTCompFrame *aot_frame)
...
+    /* Reset all the value slots from current frame sp for the else
+       branch since they be the same as starting to translate the
+       if branch */
+    for (p = block->frame_sp_begin; p < p_end; p++) {
+        p->dirty = 0;
+        p->value = NULL;
+        p->type = 0;
+        if (aot_frame->comp_ctx->enable_gc) {
+            p->ref = 0;
+            p->committed_ref = 1;
```
xujuntwt95329, line 191: Why not reset to 0?

xujuntwt95329 (commented):
LGTM

core/iwasm/compilation/aot_emit_control.c
wenyongh: `committed_ref == 1` means that the ref flag was already committed and the committed value is `0`, as `p->ref + == p->committed_ref`, no need to commit again next time.

And when translating the else branch, the status of frame refs from current frame_sp is same as translating the if branch. Note that for `if` opcode, AOT code will jump to if branch or else branch, here the `p->committed_ref` should be 1.

xujuntwt95329: Thanks!

PR #2830, opened 2023-11-28
Implement aot_alloc_frame/aot_free_frame with LLVM IRs

wenyongh:


core/iwasm/aot/aot_runtime.c
```diff
@@ -3237,6 +3248,36 @@ aot_free_frame(WASMExecEnv *exec_env)
...
+    }
+    else {
+        cur_frame->func_perf_prof_info->total_exec_time +=
+            (uintptr_t)os_time_get_boot_microsecond() - cur_frame->time_started;
+        cur_frame->func_perf_prof_info->total_exec_cnt++;
+    }
+#endif
+
+#if WASM_ENABLE_MEMORY_PROFILING != 0
+    if (alloc_frame) {
```
xujuntwt95329, line 3273: Seems `alloc_frame` means start of the function, is it correct to measure stack use at the start of the function rather than end of the function?

wenyongh: Here it does the same thing in aot_alloc_frame, which calls wasm_exec_env_alloc_wasm_frame, and wasm_exec_env_alloc_wasm_frame sets `exec_env->max_wasm_stack_used` if needed:

https://github.com/bytecodealliance/wasm-micro-runtime/blob/dev/gc_refactor/core/iwasm/aot/aot_runtime.c#L3173

https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/common/wasm_exec_env.h#L234-L235

Issue #2469, opened 2023-08-15
aot doesn't seem to handle max pages correctly


yamt:
the following code traps. (x86-64)
```
(module
  (func (export "_start")
    i32.const 0
    i32.load offset=0xffff_fffc
    i32.const 0x1234_5678
    i32.ne
    if
      unreachable
    end
  )
  (memory (export "memory") 65536)
  (data (i32.const 0xffff_fffc) "\78\56\34\12")
)
```

wenyongh:
Hi, this is a defect currently, since the type of field memory_data_size in WASMMemoryInstance is uint32, its max value is UINT32_MAX and we don't want to change the AOT ABI currently, we resize the linear memory to 4G-1 if the initial memory count is 65536. It may take time and effort to fully resolve the issue, e.g. change memory_data_size't type to uint64 and change some related code. How about we release 1.2.3 firstly and fix the issue after that? Or do you suggest to fix the issue now? Thanks.

yamt:
i don't think this is a release blocker or anything too important. i filed this issue just for a reminder.

wenyongh:
The issue should have been resolved in PR https://github.com/bytecodealliance/wasm-micro-runtime/pull/3209, after changing the the field `memory_data_size` of `WASMMemoryInstance` from type `uint32` to `uint64`, and removing the related code to resize the memory data size from 4GB to UINT32.

PR #2834, opened 2023-11-28
Access linear memory size atomically

eloparco:
#2804

[ ] Add fix for AOT

### DIFF
--- a/core/iwasm/common/wasm_memory.c
+++ b/core/iwasm/common/wasm_memory.c
@@ -686,7 +686,7 @@ wasm_enlarge_memory_internal(WASMModuleInstance *module, uint32 inc_page_count)
         memory->num_bytes_per_page = num_bytes_per_page;
         memory->cur_page_count = total_page_count;
         memory->max_page_count = max_page_count;
-        memory->memory_data_size = (uint32)total_size_new;
+        SET_LINEAR_MEMORY_SIZE(memory, (uint32)total_size_new);
         memory->memory_data_end = memory->memory_data + (uint32)total_size_new;
 
         wasm_runtime_set_mem_bound_check_bytes(memory, total_size_new);
@@ -844,7 +844,7 @@ wasm_enlarge_memory_internal(WASMModuleInstance *module, uint32 inc_page_count)
     memory->num_bytes_per_page = num_bytes_per_page;
     memory->cur_page_count = total_page_count;
     memory->max_page_count = max_page_count;
-    memory->memory_data_size = (uint32)total_size_new;
+    SET_LINEAR_MEMORY_SIZE(memory, (uint32)total_size_new);
     memory->memory_data_end = memory->memory_data + (uint32)total_size_new;
 
     wasm_runtime_set_mem_bound_check_bytes(memory, total_size_new);
--- a/core/iwasm/common/wasm_memory.h
+++ b/core/iwasm/common/wasm_memory.h
@@ -14,6 +14,16 @@
 extern "C" {
 #endif
 
+#if WASM_ENABLE_SHARED_MEMORY != 0
+#define GET_LINEAR_MEMORY_SIZE(memory) \
+    BH_ATOMIC_32_LOAD(memory->memory_data_size)
+#define SET_LINEAR_MEMORY_SIZE(memory, size) \
+    BH_ATOMIC_32_STORE(memory->memory_data_size, size)
+#else
+#define GET_LINEAR_MEMORY_SIZE(memory) memory->memory_data_size
+#define SET_LINEAR_MEMORY_SIZE(memory, size) memory->memory_data_size = size
+#endif
+
 bool
 wasm_runtime_memory_init(mem_alloc_type_t mem_alloc_type,
                          const MemAllocOption *alloc_option);
--- a/core/iwasm/interpreter/wasm_interp_classic.c
+++ b/core/iwasm/interpreter/wasm_interp_classic.c
@@ -36,7 +36,7 @@ typedef float64 CellType_F64;
  * multi-threading mode since it may be changed by other
  * threads in memory.grow
  */
-#define get_linear_mem_size() memory->memory_data_size
+#define get_linear_mem_size() GET_LINEAR_MEMORY_SIZE(memory)
 #endif
 
 #if !defined(OS_ENABLE_HW_BOUND_CHECK) \
@@ -1155,7 +1155,13 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
 #if !defined(OS_ENABLE_HW_BOUND_CHECK)              \
     || WASM_CPU_SUPPORTS_UNALIGNED_ADDR_ACCESS == 0 \
     || WASM_ENABLE_BULK_MEMORY != 0
-    uint32 linear_mem_size = memory ? memory->memory_data_size : 0;
+    uint32 linear_mem_size = 0;
+    if (memory)
+#if WASM_ENABLE_THREAD_MGR == 0
+        linear_mem_size = memory->memory_data_size;
+#else
+        linear_mem_size = GET_LINEAR_MEMORY_SIZE(memory);
+#endif
 #endif
     WASMType **wasm_types = module->module->types;
     WASMGlobalInstance *globals = module->e->globals, *global;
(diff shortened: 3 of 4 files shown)


core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -29,14 +29,14 @@ typedef float64 CellType_F64;
 #define BR_TABLE_TMP_BUF_LEN 32
 
 #if WASM_ENABLE_THREAD_MGR == 0
-#define get_linear_mem_size() linear_mem_size
+#define get_linear_mem_size() memory->memory_data_size
```
wenyongh, line 32: Seems no need to modify this line, for single-threading, linear_mem_size is gotten at the beginning of function wasm_interp_call_func_bytecode, and is only updated after a function call and memory.grow opcode. Updating this line may lead to CI failure:
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/7022252919/job/19106153525#step:10:2133

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -20,14 +20,14 @@ typedef float32 CellType_F32;
 typedef float64 CellType_F64;
 
 #if WASM_ENABLE_THREAD_MGR == 0
-#define get_linear_mem_size() linear_mem_size
+#define get_linear_mem_size() memory->memory_data_size
```
wenyongh, line 23: Same as above.

core/iwasm/common/wasm_memory.h
```diff
@@ -14,6 +14,11 @@
 extern "C" {
 #endif
 
+#define GET_LINEAR_MEMORY_SIZE(memory) \
```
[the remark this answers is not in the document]
wenyongh: GET_LINEARY_MEMORY_SIZE should be good, the linear memory size can be 64-bit for memory32 also: when memory page count is 65536, the memory size is 4G ( = UINT32_MAX + 1). There is a limitation in current design, we resize 4G to 4G-1 to fit 32-bit memory_data_size so as not to violate the AOT ABI. We may refactor the code to fix it later. See also:
https://github.com/bytecodealliance/wasm-micro-runtime/issues/2530
https://github.com/bytecodealliance/wasm-micro-runtime/issues/2469

core/iwasm/common/wasm_memory.h
```diff
@@ -14,6 +14,11 @@
...
+    BH_ATOMIC_32_LOAD(memory->memory_data_size)
+#define SET_LINEARM_MEMORY_SIZE(memory, size) \
+    BH_ATOMIC_32_STORE(memory->memory_data_size, size)
+
```
wenyongh, line 21: The BH_ATOMIC_32_LOAD/STORE are only needed when shared memory is enabled, and might be not available in some platforms, how about:
```C
#define GET_LINEAR_MEMORY_SIZE(memory) \
    BH_ATOMIC_32_LOAD(memory->memory_data_size)
#define SET_LINEARM_MEMORY_SIZE(memory, size) \
    BH_ATOMIC_32_STORE(memory->memory_data_size, size)
#else
#define GET_LINEAR_MEMORY_SIZE(memory) memory->memory_data_size
#define SET_LINEARM_MEMORY_SIZE(memory, size) memory->memory_data_size = size
#endif

File: core/iwasm/interpreter/wasm_interp_classic.c (hunk 1 above)
Comment (eloparco) [reply core/iwasm/interpreter/wasm_interp_classic.c]: I was getting a warning when doing that, that's why I replaced it:
```bash
wasm-micro-runtime/core/iwasm/interpreter/wasm_interp_fast.c:1183:39: warning: variable 'linear_mem_size' is uninitialized when used within its own initialization [-Wuninitialized]
    uint32 linear_mem_size = memory ? get_linear_mem_size() : 0;
           ~~~~~~~~~~~~~~~            ^~~~~~~~~~~~~~~~~~~~~
/Users/eloparco/dev/forks/wasm-micro-runtime/core/iwasm/interpreter/wasm_interp_fast.c:23:31: note: expanded from macro 'get_linear_mem_size'
#define get_linear_mem_size() linear_mem_size
                              ^~~~~~~~~~~~~~~
```

I can fix it by rewriting it as 
```c
uint32 linear_mem_size = 0;
if (memory)
    linear_mem_size = get_linear_mem_size();
```

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -36,7 +36,7 @@ typedef float64 CellType_F64;
  * multi-threading mode since it may be changed by other
  * threads in memory.grow
  */
-#define get_linear_mem_size() memory->memory_data_size
+#define get_linear_mem_size() GET_LINEAR_MEMORY_SIZE(memory)
```
loganek, line 39: do we still need `get_linear_mem_size` macro given that `GET_LINEAR_MEMORY_SIZE` will work for both shared and non-shared memory (as per @wenyongh 's comment https://github.com/bytecodealliance/wasm-micro-runtime/pull/2834/files#r1408616821)? Could we just use `GET_LINEAR_MEMORY_SIZE` ?

core/iwasm/common/wasm_memory.c
```diff
@@ -686,7 +686,7 @@ wasm_enlarge_memory_internal(WASMModuleInstance *module, uint32 inc_page_count)
         memory->num_bytes_per_page = num_bytes_per_page;
         memory->cur_page_count = total_page_count;
         memory->max_page_count = max_page_count;
-        memory->memory_data_size = (uint32)total_size_new;
+        SET_LINEAR_MEMORY_SIZE(memory, total_size_new);
```
wenyongh, line 689: Had better `SET_LINEAR_MEMORY_SIZE(memory, (uint32)total_size_new)`, not sure whether the CI failures are caused by this.

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -1180,7 +1180,9 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
 #if !defined(OS_ENABLE_HW_BOUND_CHECK)              \
     || WASM_CPU_SUPPORTS_UNALIGNED_ADDR_ACCESS == 0 \
     || WASM_ENABLE_BULK_MEMORY != 0
-    uint32 linear_mem_size = memory ? memory->memory_data_size : 0;
+    uint32 linear_mem_size = 0;
+    if (memory)
+        linear_mem_size = get_linear_mem_size();
```
wenyongh, line 1185: ditto

core/iwasm/interpreter/wasm_interp_classic.c
eloparco: I'll do that. 

What about AOT instead? I started looking into the code but not very familiar with that part, any hints on where to look to apply similar changes to what I did for classic and fast interpreter?

wenyongh: Sorry I checked the code again in classic interpreter, we had better keep the macro `get_linear_mem_size`, its meaning is different from the macro GET_LINEAR_MEMORY_SIZE: it uses local variable linear_mem_size in sing-threading mode, as linear_mem_size is only updated after function calls and memory.grow, we may improve the performance in interpreter mode. If we change to GET_LINEAR_MEMORY_SIZE, it will get the memory size from memory structure each time, which may impacts the performance in single-threading mode.

wenyongh: > Sorry I checked the code again in classic interpreter, we had better keep the macro `get_linear_mem_size`, its meaning is different from the macro GET_LINEAR_MEMORY_SIZE: it uses local variable linear_mem_size in sing-threading mode, as linear_mem_size is only updated after function calls and memory.grow, we may improve the performance in interpreter mode. If we change to GET_LINEAR_MEMORY_SIZE, it will get the memory size from memory structure each time, which may impacts the performance in single-threading mode.

The nuttx CI failure should be caused by the changing.

wenyongh: > I'll do that.
> 
> What about AOT instead? I started looking into the code but not very familiar with that part, any hints on where to look to apply similar changes to what I did for classic and fast interpreter?

The AOT mode doesn't use linear memory size to do boundary check for normal load/store opcodes, instead it uses the `mem_bound_check_nbytes` fields:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/compilation/aot_emit_memory.c#L42-L60

 It only uses linear memory size in checking the size in opcode memory.fill/memory.copy:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/compilation/aot_emit_memory.c#L935-L944

eloparco: Yes, the problem I had was that I can't use `get_linear_mem_size()` for the first initialization in `wasm_interp_call_func_bytecode` because otherwise it uses `#define get_linear_mem_size() linear_mem_size`. Instead we need to get `memory->memory_data_size` at the beginning, afterwards we can rely on `get_linear_mem_size()` to get the `linear_mem_size` cached value.

wenyongh (commented):
LGTM

loganek (approved):
Lgtm

PR #2837, opened 2023-11-29
use wasm_config_t to pass embedded private configuration

lum1n0us:


core/iwasm/common/wasm_c_api.c
```diff
@@ -418,37 +425,32 @@ static korp_mutex engine_lock = OS_THREAD_MUTEX_INITIALIZER;
...
-        singleton_engine = wasm_engine_new_internal(type, opts);
+        singleton_engine = wasm_engine_new_internal(config);
     else
         singleton_engine->ref_count++;
 
 #if defined(OS_THREAD_MUTEX_INITIALIZER)
     os_mutex_unlock(&engine_lock);
 #endif
 
+    wasm_config_delete(config);
```
wenyongh, line 450: The config may be created by developer or caller, should not free it here. Instead, it should be freed by developer or caller.

core/iwasm/common/wasm_c_api.c
```diff
@@ -418,37 +425,32 @@ static korp_mutex engine_lock = OS_THREAD_MUTEX_INITIALIZER;
 #endif
 
 own wasm_engine_t *
-wasm_engine_new_with_args(mem_alloc_type_t type, const MemAllocOption *opts)
+wasm_engine_new()
+{
+    wasm_config_t *config = wasm_config_new();
+    return wasm_engine_new_with_config(config);
```
wenyongh, line 431: Had better be like:
```C
    wasm_engine_t *engine;
    wasm_config_t *config = wasm_config_new();
    engine = wasm_engine_new_with_config(config);
    wasm_config_delete(config);
    return engine;
```

core/iwasm/common/wasm_c_api.c
```diff
@@ -418,37 +425,32 @@ static korp_mutex engine_lock = OS_THREAD_MUTEX_INITIALIZER;
 #endif
 
 own wasm_engine_t *
-wasm_engine_new_with_args(mem_alloc_type_t type, const MemAllocOption *opts)
```
wenyongh, line 421: This is an ABI change, not sure whether to remove it? If yes, we need to add it in the "Breaking changes" of the [release notes](https://github.com/bytecodealliance/wasm-micro-runtime/pull/2821/files).

core/iwasm/include/wasm_c_api.h
```diff
@@ -191,10 +166,50 @@ typedef union MemAllocOption {
...
-#endif
+#endif /* MEM_ALLOC_OPTION_DEFINED */
+
+/* Runtime configration */
+struct wasm_config_t {
+    mem_alloc_type_t mem_alloc_type;
+    MemAllocOption mem_alloc_option;
+    bool linux_perf_support;
+    /*TODO: wasi args*/
+};
```
wenyongh, line 177: Seems no need to expose wasm_config_t in this header file, is it better to define it in wasm_c_api.c?

core/iwasm/common/wasm_c_api.c
lum1n0us: You are right. I am gona keep the old function and use a `deprecated` attribute to highlight.

wenyongh (commented):
LGTM

core/iwasm/include/wasm_c_api.h
```diff
@@ -191,9 +166,51 @@ typedef union MemAllocOption {
...
+ * then they are thread-safe: we use a global lock to lock the
+ * operations of the engine. Otherwise they are not thread-safe:
+ * when there are engine new/delete operations happening
+ * simultaneously in multiple threads, developer must create
+ * the lock by himself, and add the lock when calling these
+ * functions.
+ */
+WASM_API_EXTERN own wasm_engine_t* wasm_engine_new(void);
+WASM_API_EXTERN own wasm_engine_t* wasm_engine_new_with_config(wasm_config_t*);
+WASM_API_EXTERN own __attribute__((deprecated)) wasm_engine_t *
```
wenyongh, line 213: Windows doesn't recognize this attribute and compilation errors were reported:
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/7041418570/job/19163938548#step:4:706

wenyongh (approved):
LGTM

PR #2839, opened 2023-11-29
Get rid of compilation warnings and minor doc fix

eloparco:
Warnings:
```bash
wasm-micro-runtime/core/iwasm/common/wasm_runtime_common.c:5902:34: note: uninitialized use occurs here
    bh_assert(BH_LIST_SUCCESS == ret);
wasm-micro-runtime/core/iwasm/common/wasm_runtime_common.c:5876:5: note: variable 'ret' is declared here
    bh_list_status ret;
    ^
```

```bash
wasm-micro-runtime/core/shared/platform/common/posix/posix_file.c:482:26: warning: comparison of integers of different signs: 'size_t' (aka 'unsigned long') and 'int' [-Wsign-compare]
    for (size_t i = 0; i < iovcnt; ++i)
                       ~ ^ ~~~~~~
```

### DIFF
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -5873,7 +5873,7 @@ wasm_runtime_register_sub_module(const WASMModuleCommon *parent_module,
 {
     /* register sub_module into its parent sub module list */
     WASMRegisteredModule *node = NULL;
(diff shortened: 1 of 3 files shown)


eloparco:
There's an error on Windows that seems unrelated since I see it in other PRs as well.

PR #2841, opened 2023-11-30
Fix bug in memory.init and memory.grow opcodes for fast-jit when multithreading is enabled

TianlongLiang:
Modify the approach to calculate the offset and access the data members in `WASMMemoryInstance` when shared memory(multithreading) is enabled. Instead of calculating it from the module instance base address, now use `module_inst -> memories[0]` as the base address for memory.init, memory.grow, and certain boundary check related APIs.

### DIFF
--- a/core/iwasm/fast-jit/jit_frontend.c
+++ b/core/iwasm/fast-jit/jit_frontend.c
@@ -218,43 +218,152 @@ get_aux_stack_bottom_reg(JitFrame *frame)
     return frame->aux_stack_bottom_reg;
 }
 
+#if WASM_ENABLE_SHARED_MEMORY != 0
+static bool
+is_shared_memory(WASMModule *module, uint32 mem_idx)
+{
+    WASMMemory *memory;
+    WASMMemoryImport *memory_import;
+    bool is_shared;
+
+    /* TODO: for now, only one memory in total is allowed, in the future should
+     * change this logic to support multiple memories */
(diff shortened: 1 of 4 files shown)


core/iwasm/fast-jit/jit_frontend.c
```diff
@@ -218,59 +218,190 @@ get_aux_stack_bottom_reg(JitFrame *frame)
...
+    /* TODO: for now, only one memory in total is allowed, in the future should
+     * change this logic to support multiple memories */
+    if (module->memory_count != 0) {
+        memory = &module->memories[mem_idx];
+        is_shared = memory->flags & 0x02 ? true : false;
+    }
+    else if (module->import_memory_count != 0) {
+        memory_import = &(module->import_memories[mem_idx].u.memory);
+        is_shared = memory_import->flags & 0x02 ? true : false;
+    }
```
wenyongh, line 238: Had better:
```C
    if (mem_idx < module->import_memory_count) {
        memory_import = &(module->import_memories[mem_idx].u.memory);
        is_shared = memory_import->flags & 0x02 ? true : false;
    }
    else {
        memory = &module->memories[mem_idx - module->import_memory_count];
        is_shared = memory->flags & 0x02 ? true : false;
    }
```
We may implement multiple memories feature in the future.

wenyongh (commented):
LGTM

PR #2843, opened 2023-12-01
Enable spectest on riscv64

no1wudi:


core/iwasm/aot/arch/aot_reloc_riscv.c
```diff
@@ -372,60 +384,37 @@ apply_relocation(AOTModule *module, uint8 *target_section_addr,
...
         case R_RISCV_PCREL_LO12_I: /* S - P */
         {
-            if (reloc_type == R_RISCV_PCREL_LO12_I) {
-                /* A = 0 */
-                val = (int32)((intptr_t)symbol_addr - (intptr_t)addr);
-            }
-            else {
-                val = (int32)((intptr_t)symbol_addr + (intptr_t)reloc_addend);
-            }
+            /* Already handled in R_RISCV_PCREL_HI20 */
```
wenyongh, line 405: R_RISCV_PCREL_HI20 doesn't equal to R_RISCV_PCREL_LO12_I, why not need to handle R_RISCV_PCREL_LO12_I here?

no1wudi: This is a little trick, to reuse the code for both R_RISCV_CALL/R_RISCV_CALL_PLT/R_RISCV_PCREL_HI20, R_RISCV_PCREL_LO12_I handled in line 373.

tests/wamr-test-suites/spec-test-script/runtest.py
```diff
@@ -1072,6 +1072,15 @@ def compile_wasm_to_aot(wasm_tempfile, aot_tempfile, runner, opts, r, output = '
     # exception isn't thrown in several cases
     cmd.append("--disable-llvm-lto")
 
+    # Bounds checks disabled by default for 64-bit targets, enable it for qemu,
+    # which may be used for 64-bit targets
```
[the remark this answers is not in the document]
wenyongh: OK, shall we update the comments to make it clear? Somewhat like "Bounds-checks is disabled by default for 64-bit targets, but it may be not available in QEMU with Nuttx, so we enable it here"?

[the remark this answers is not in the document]
no1wudi: OK

core/iwasm/aot/arch/aot_reloc_riscv.c
wenyongh: But this branch only `break`, it does nothing, it is really strange. Should we add `case R_RISCV_PCREL_LO12_I` after `case R_RISCV_PCREL_HI20` to reuse the code, and then update some code in that handling accordingly?

no1wudi:
Latest spec test here: https://github.com/no1wudi/wasm-micro-runtime/actions/runs/7081980946

core/iwasm/aot/arch/aot_reloc_riscv.c
no1wudi: Essential check added here

wenyongh (commented):
LGTM

PR #2848, opened 2023-12-02
ci: add GH action to test compilation on Zephyr

deadprogram:
This PR is to add a GH action to test compilation of the product-mini "simple" example on Zephyr against the STM32 board.


wenyongh:
Close the PR since it is in draft and hasn't been updated for a long time.

PR #2850, opened 2023-12-02
bugfix: three unsigned narrow op should be interpreted as signed as the simd proposal

liutao-liu:
image


core/iwasm/compilation/simd/simd_conversions.c
```diff
@@ -160,9 +160,9 @@ simd_integer_narrow_common(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
 
     /* sat */
```
lum1n0us, line 161: would u mind quoting the spec restriction here , especially "regardless of the whether ..."

lum1n0us (approved):
LGTM

PR #2851, opened 2023-12-04
Fix typos of CIDR in docs and help text

hasheddan:
Fixes typos of CIDR in socket documentation and libc WASI help text.


product-mini/platforms/common/libc_wasi.c
```diff
@@ -47,7 +47,7 @@ libc_wasi_print_help()
            "--map-dir=<guest-path2::host-path2>\n");
     printf("  --addr-pool=<addrs>      Grant wasi access to the given network "
            "addresses in\n");
-    printf("                           CIRD notation to the program, seperated "
+    printf("                           CIDR notation to the program, seperated "
```
wenyongh, line 50: Could you also fix this file:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/product-mini/platforms/linux-sgx/enclave-sample/App/App.cpp#L231

PR #2856, opened 2023-12-04
Using explicit FPU type in spectest on NuttX

no1wudi:


.github/workflows/spec_test_on_nuttx.yml
```diff
@@ -37,32 +37,32 @@ jobs:
...
           #   config: "boards/arm/imx6/sabre-6quad/configs/nsh",
           #   target: "thumbv7",
-          #   use_fpu: false
+          #   fpu_type: "none"
           # },
           {
             config: "boards/arm/imx6/sabre-6quad/configs/nsh",
             target: "thumbv7_vfp",
-            use_fpu: true
+            fpu_type: "dp"
```
wenyongh, line 50: Seems that "dp" isn't handled in below?

no1wudi: Yes, no need to handle it since it's enabled by default if DPFPU is available

PR #2868, opened 2023-12-05
Fix compilation warnings on Windows

wenyongh:


core/shared/utils/bh_bitmap.h
```diff
@@ -58,7 +58,7 @@ bh_bitmap_delete(bh_bitmap *bitmap)
  * @return true if the index is in range, false otherwise
  */
 static inline bool
-bh_bitmap_is_in_range(bh_bitmap *bitmap, unsigned n)
+bh_bitmap_is_in_range(bh_bitmap *bitmap, uintptr_t n)
```
wenyongh, line 61: The begin_index/end_index fields of bh_bitmap are defined as uintptr_t.

PR #2875, opened 2023-12-06
 Portably handle `fd_advise` on directory fd 

yagehu:
This commit adds a check to `fd_advise`.  If the fd is a directory, return `ebadf`.  This brings iwasm in line with Wasmtime's behavior. WASI folks have [stated](https://github.com/bytecodealliance/wasmtime/issues/6505#issuecomment-1574122949) that `fd_advise` should not work on directories as this is a Linux-specific behavior.


wenyongh (commented):
LGTM

yagehu:
> > This commit adds a check to `fd_advise`. If the fd is a directory, return `ebadf`. This brings iwasm in line with Wasmtime's behavior. WASI folks have [stated](https://github.com/bytecodealliance/wasmtime/issues/6505#issuecomment-1574122949) that `fd_advise` should not work on directories as this is a Linux-specific behavior.
> 
> where in WASI or POSIX is it stated?

The dirfd behavior is not specified by POSIX. I submitted this PR observing that using directory fds with `fd_*` WASI calls fail (with different errnos) in other runtimes (Wasmtime and WasmEdge). Wasmtime has a [conformance test](https://github.com/bytecodealliance/wasmtime/commit/d1d381e2984ee3103f164c8045f40b13e2e6db8c) asserting this behavior.

PR #2879, opened 2023-12-07
Revert "Grab `cluster->lock` when modifying `exec_env->module_inst` (…

yamt:
…#2685)"

This reverts commit 24c4d256b32465e2979d7d7fee358d1db6d90c90.

It has a deadlock issue like:
    wasm_runtime_spawn_exec_env
    wasm_cluster_spawn_exec_env (hold the lock)
    wasm_runtime_instantiate_internal
    wasm_instantiate
    execute_post_instantiate_functions
    wasm_exec_env_set_module_inst (grab the lock again)

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -982,8 +982,7 @@ execute_post_instantiate_functions(AOTModuleInstance *module_inst,
            wasm functions, and ensure that the exec_env's module inst
            is the correct one. */
         module_inst_main = exec_env_main->module_inst;
-        wasm_exec_env_set_module_inst(exec_env,
-                                      (WASMModuleInstanceCommon *)module_inst);
+        exec_env->module_inst = (WASMModuleInstanceCommon *)module_inst;
     }
     else {
         /* Try using the existing exec_env */
@@ -1008,8 +1007,7 @@ execute_post_instantiate_functions(AOTModuleInstance *module_inst,
                module inst to ensure that the exec_env's module inst
                is the correct one. */
             module_inst_main = exec_env->module_inst;
-            wasm_exec_env_set_module_inst(
-                exec_env, (WASMModuleInstanceCommon *)module_inst);
+            exec_env->module_inst = (WASMModuleInstanceCommon *)module_inst;
         }
     }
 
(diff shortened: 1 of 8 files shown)


Zzzabiyaka:
Hi 

Great catch!
Is there any small test that we can add which would detect this issue ? (in a separate PR)

yamt:
> Hi
> 
> Great catch! Is there any small test that we can add which would detect this issue ? (in a separate PR)

i don't have test cases. it's reported by a user.

wenyongh:
> > Hi, in wasm_cluster_spawn_exec_env, seems that the `cluster->lock` is grabbed to check whether there is an exception was thrown and to add the new exec_env into exec_env_list, could we move `wasm_runtime_instantiate_internal` before grabbing the lock (instantiate first and then grab the lock) to resolve the issue?
> 
> maybe. i'm proposing a revert for now because it likely takes a bit for me to come up with a nicer fix.

Hi, I checked the calling of `wasm_runtime_instantiate_internal` in other two places (thread_spawn_wrapper in lib-wasi-threads and pthread_create_wrapper in lib-pthread), the `cluster->lock` isn't acquired in the module instantiation and it seems they work well, so I think here we don't need to grab the `cluster->lock` also. I believe you had thought the solution carefully when implementing the PR and it should be already a nice one, it should work for most cases except this. So I tend to not revert the PR, instead, I submitted PR  #2882 to fix it, could you help review and help ask the user to test again?

@Zzzabiyaka how do you think? You raised the issue in #2680.

yamt:
> > > Hi, in wasm_cluster_spawn_exec_env, seems that the `cluster->lock` is grabbed to check whether there is an exception was thrown and to add the new exec_env into exec_env_list, could we move `wasm_runtime_instantiate_internal` before grabbing the lock (instantiate first and then grab the lock) to resolve the issue?
> > 
> > 
> > maybe. i'm proposing a revert for now because it likely takes a bit for me to come up with a nicer fix.
> 
> Hi, I checked the calling of `wasm_runtime_instantiate_internal` in other two places (thread_spawn_wrapper in lib-wasi-threads and pthread_create_wrapper in lib-pthread), the `cluster->lock` isn't acquired in the module instantiation and it seems they work well, so I think here we don't need to grab the `cluster->lock` also. I believe you had thought the solution carefully when implementing the PR and it should be already a nice one, it should work for most cases except this. So I tend to not revert the PR, instead, I submitted PR #2882 to fix it, could you help review and help ask the user to test again?

i often prefer reverting problematic changes immediately and re-applying it later with a fix.
but after, all it's something for the project to decide.

wenyongh:
> > 
> > 
> > Hi, I checked the calling of `wasm_runtime_instantiate_internal` in other two places (thread_spawn_wrapper in lib-wasi-threads and pthread_create_wrapper in lib-pthread), the `cluster->lock` isn't acquired in the module instantiation and it seems they work well, so I think here we don't need to grab the `cluster->lock` also. I believe you had thought the solution carefully when implementing the PR and it should be already a nice one, it should work for most cases except this. So I tend to not revert the PR, instead, I submitted PR #2882 to fix it, could you help review and help ask the user to test again?
> 
> i often prefer reverting problematic changes immediately and re-applying it later with a fix. but after all, it's something for the project to decide.

Reverting and re-applying make bigger code change, I am afraid it is not so friendly to developers, it may be better to fix the issue directly if we can. Let's merge PR #2882 to see whether things run good😊.

Zzzabiyaka:
>@Zzzabiyaka how do you think? You raised the issue in https://github.com/bytecodealliance/wasm-micro-runtime/issues/2680.

Yeah sure, fine with me

Issue #2880, opened 2023-12-07
Zephyr builds fail when using app-manager or other advanced WAMR features


srberard:
When building on the current version of Zephyr there are a number of errors due to incorrect header files being specified (example below).  It seems that the current support for Zephyr is a bit outdated.  This issue is similar to #1726.  However, that issue only fixed the code used for the minimal example.  Code that references the app manager of graphics libraries fails to compile on current Zephyr versions.

Example compile error:
```shell
/src/project/wasm-micro-runtime/core/app-mgr/app-manager/platform/zephyr/app_mgr_zephyr.c:9:10: fatal error: zephyr.h: No such file or directory
    9 | #include <zephyr.h>
      |          ^~~~~~~~~~
compilation terminated.
````
The fix is straightforward, and I'm working on a pull request and will submit it once I've tested it.

srberard:
Changes merged.

PR #2881, opened 2023-12-07
Bugfix:  Issue #2880 -  Corrects Zephyr include files for current versions of Zephyr

srberard:
This fixes bug #2880.  Zephyr 3.2 made changes to how headers are reference (see [release notes](https://docs.zephyrproject.org/latest/releases/release-notes-3.2.html)).  Work item [49578](https://github.com/zephyrproject-rtos/zephyr/issues/49578) deprecated the old headers names.  The current WAMR codebase references these old headers, thus causing compile errors with current versions of Zephyr.

This update adds #ifdefs around the header names.  With this change, compiling with Zephyr 3.2.0 and above will use the new header files.  Prior versions will use the existing code.

I have tested this with Zephyr 3.2.0, 3.30, and the latest Zephyr release.  I have not verified this with the current LTS version (2.7.0).


wenyongh (commented):
LGTM

Issue #2776, opened 2023-11-16
windows spec test on the ci seems flaky


yamt:
eg. https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/6886633084/job/18732680213?pr=2775
```
;; This module is required to validate, regardless of whether it can be
;; linked. Overloading is not possible in wasm itself, but it is possible
;; in modules from which wasm can import.
Writing WAST module to 'C:\Users\RUNNER~1\AppData\Local\Temp\tmpva61dvn2.wast'
Compiling WASM to 'C:\Users\RUNNER~1\AppData\Local\Temp\tmpbfjw868x.wasm'
Running: ./wabt/out/gcc/Release/wat2wasm.exe --enable-thread --no-check C:\Users\RUNNER~1\AppData\Local\Temp\tmpva61dvn2.wast -o C:\Users\RUNNER~1\AppData\Local\Temp\tmpbfjw868x.wasm
Starting interpreter for module 'C:\Users\RUNNER~1\AppData\Local\Temp\tmpbfjw868x.wasm'
Running: ../../../product-mini/platforms/windows/build/RelWithDebInfo/iwasm.exe --module-path=C:\Users\RUNNER~1\AppData\Local\Temp --heap-size=0 -v=0 --repl C:\Users\RUNNER~1\AppData\Local\Temp\tmpbfjw868x.wasm
module is ....'(module
    (import "not wasm" "overloaded" (func))
    (import "not wasm" "overloaded" (func (param i32)))
    (import "not wasm" "overloaded" (func (param i32 i32)))
    (import "not wasm" "overloaded" (func (param i64)))
    (import "not wasm" "overloaded" (func (param f32)))
    (import "not wasm" "overloaded" (func (param f64)))
    (import "not wasm" "overloaded" (func (result i32)))
    (import "not wasm" "overloaded" (func (result i64)))
    (import "not wasm" "overloaded" (func (result f32)))
    (import "not wasm" "overloaded" (func (result f64)))
    (import "not wasm" "overloaded" (global i32))
    (import "not wasm" "overloaded" (global i64))
    (import "not wasm" "overloaded" (global f32))
    (import "not wasm" "overloaded" (global f64))
    (import "not wasm" "overloaded" (table 0 funcref))
    (import "not wasm" "overloaded" (memory 0))
  )'
exception is ....'unknown import'
[...]

PR #2883, opened 2023-12-08
samples/spawn-thread: disable libc and pthread

yamt:
because this sample doesn't really require them.


yamt:
the ci failure looks like https://github.com/bytecodealliance/wasm-micro-runtime/issues/2776

PR #2887, opened 2023-12-08
Using docker image for spectest

no1wudi:


no1wudi:
It will consume too much resource to slow down the other checks if many PR merge, so let's run the spectest daily again.

wenyongh (approved):

wenyongh (commented):
LGTM

PR #2892, opened 2023-12-09
Handle ambiguous fstflags on fd_filestat_set_times

yagehu:
It's possible to set both `atim` and `atim_now` in the [`fstflags`](https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/docs.md#-fstflags-record) parameter.  Same goes for `mtin` and `mtim_now`.  However, it's ambiguous which time should be set in these two cases.  This commit checks this and returns `inval`.  This behavior is consistent across Wasmtime and WasmEdge.


wenyongh (approved):
LGTM

PR #2896, opened 2023-12-11
Enable GC spectest for NuttX

no1wudi:


wenyongh (approved):
LGTM

.github/workflows/spec_test_on_nuttx.yml
```diff
@@ -75,20 +81,25 @@ jobs:
...
+          #   option: "CONFIG_INTERPRETERS_WAMR_AOT=y\\n"
+          # },
+          # {
+          #   mode: "-t classic-interp",
+          #   option: "CONFIG_INTERPRETERS_WAMR_CLASSIC=y\\n"
+          # },
+          # {
+          #   mode: "-t fast-interp",
+          #   option: "CONFIG_INTERPRETERS_WAMR_FAST=y\\n"
+          # },
```
yamt, line 95: @no1wudi 
why did you comment these out?

no1wudi: There may have some potential bug, spectest for interpreter is extremly slow, about 4H.

yamt: do you mean GC-related bugs?

no1wudi: No, I think this bug is from test script, No-gc case also very slow, with GC enabled the test maybe over 6H that exceed the limit of GA runner

PR #2898, opened 2023-12-11
Fix memory size not updating

yagehu:
This commit fixes linear memory size not updating after growing.  This causes `memory.fill` to throw an exception after `memory.grow`.

### DIFF
--- a/core/iwasm/interpreter/wasm_interp_classic.c
+++ b/core/iwasm/interpreter/wasm_interp_classic.c
@@ -2152,7 +2152,7 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
 #if WASM_ENABLE_THREAD_MGR == 0
                     linear_mem_size = memory->memory_data_size;
 #else
-                    linear_mem_size = get_linear_mem_size();
+                    linear_mem_size = GET_LINEAR_MEMORY_SIZE(memory);
 #endif
 #endif
                 }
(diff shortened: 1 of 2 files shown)


core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -2149,7 +2149,11 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
 #if !defined(OS_ENABLE_HW_BOUND_CHECK)              \
     || WASM_CPU_SUPPORTS_UNALIGNED_ADDR_ACCESS == 0 \
     || WASM_ENABLE_BULK_MEMORY != 0
+#if WASM_ENABLE_THREAD_MGR == 0
```
eloparco, line 2152: You may want to be consistent with what we do at the beginning of the function: https://github.com/bytecodealliance/wasm-micro-runtime/blob/4aee3cf14e471eff3f0a1e731423ad35243b9991/core/iwasm/interpreter/wasm_interp_classic.c#L1160-L1164

Also, similar changes need to be made for `wasm_interp_fast.c`. See https://github.com/bytecodealliance/wasm-micro-runtime/pull/2834

yagehu: Thanks for the review. I just pushed another commit that makes the check consistent and applies the same fix to `wasm_interp_fast.c`.

wenyongh (approved):

PR #2908, opened 2023-12-15
Return stack frames of crashed thread when using wasm-c-api

eloparco:
When using the wasm-c-api and there's a trap, `wasm_func_call()` returns a `wasm_trap_t *` object.
No matter which thread crashes, the trap contains the stack frames of the main thread.

With this PR, when there's an exception, the stack frames of the thread where the exception occurs are stored into the thread cluster. `wasm_func_call()` can then return those stack frames.

### DIFF
--- a/core/iwasm/common/wasm_c_api.c
+++ b/core/iwasm/common/wasm_c_api.c
@@ -1905,7 +1905,7 @@ wasm_frame_func_offset(const wasm_frame_t *frame)
 static wasm_trap_t *
 wasm_trap_new_internal(wasm_store_t *store,
                        WASMModuleInstanceCommon *inst_comm_rt,
-                       const char *error_info)
+                       const char *error_info, Vector *cluster_frames)
 {
     wasm_trap_t *trap;
 #if WASM_ENABLE_DUMP_CALL_STACK != 0
@@ -1935,7 +1935,9 @@ wasm_trap_new_internal(wasm_store_t *store,
 
     /* fill in frames */
 #if WASM_ENABLE_DUMP_CALL_STACK != 0
-    trap->frames = ((WASMModuleInstance *)inst_comm_rt)->frames;
+    trap->frames = cluster_frames
+                       ? cluster_frames
+                       : ((WASMModuleInstance *)inst_comm_rt)->frames;
 
     if (trap->frames) {
         /* fill in instances */
@@ -3268,6 +3270,7 @@ wasm_func_call(const wasm_func_t *func, const wasm_val_vec_t *params,
     WASMFunctionInstanceCommon *func_comm_rt = NULL;
     WASMExecEnv *exec_env = NULL;
     size_t param_count, result_count, alloc_count;
+    Vector *cluster_frames = NULL;
 
     if (!func) {
         return NULL;
@@ -3385,9 +3388,13 @@ failed:
     if (argv != argv_buf)
         wasm_runtime_free(argv);
 
+#if WASM_ENABLE_DUMP_CALL_STACK != 0 && WASM_ENABLE_THREAD_MGR != 0
+    WASMCluster *cluster = wasm_exec_env_get_cluster(exec_env);
+    cluster_frames = cluster->exception_frames;
+#endif
     return wasm_trap_new_internal(
         func->store, func->inst_comm_rt,
-        wasm_runtime_get_exception(func->inst_comm_rt));
+        wasm_runtime_get_exception(func->inst_comm_rt), cluster_frames);
 }
 
 size_t
--- a/core/iwasm/common/wasm_c_api.c
+++ b/core/iwasm/common/wasm_c_api.c
@@ -1981,8 +1981,13 @@ wasm_trap_new_internal(wasm_store_t *store,
         }
 
         for (i = 0; i < trap->frames->num_elems; i++) {
+#if WASM_ENABLE_THREAD_MGR != 0
             ((wasm_frame_vec_t *)trap->frames)->data[i]->instance =
                 frame_instance;
+#else
+            (((wasm_frame_t *)trap->frames->data) + i)->instance =
+                frame_instance;
+#endif
         }
     }
 #else
@@ -2077,7 +2082,11 @@ wasm_trap_trace(const wasm_trap_t *trap, own wasm_frame_vec_t *out)
     }
 
     for (i = 0; i < trap->frames->num_elems; i++) {
+#if WASM_ENABLE_THREAD_MGR != 0
         wasm_frame_t *frame = ((wasm_frame_vec_t *)trap->frames)->data[i];
+#else
+        wasm_frame_t *frame = ((wasm_frame_t *)trap->frames->data) + i;
+#endif
         if (!(out->data[i] =
                   wasm_frame_new(frame->instance, frame->module_offset,
                                  frame->func_index, frame->func_offset))) {
--- a/core/iwasm/common/wasm_c_api.c
+++ b/core/iwasm/common/wasm_c_api.c
@@ -1903,32 +1903,14 @@ wasm_frame_func_offset(const wasm_frame_t *frame)
 }
 
 void
-wasm_frame_vec_clone_internal(wasm_frame_vec_t *src, wasm_frame_vec_t *out)
+wasm_frame_vec_clone_internal(Vector *src, Vector *out)
 {
-    uint32 i;
+    bh_assert(src->num_elems != 0 && src->data);
 
-    wasm_frame_vec_new_uninitialized(out, src->num_elems);
-    if (out->size == 0 || !out->data) {
-        return;
-    }
-
-    for (i = 0; i < src->num_elems; i++) {
-        wasm_frame_t *frame = ((wasm_frame_t *)src->data) + i;
-        if (!(out->data[i] = wasm_frame_copy(frame))) {
-            goto failed;
-        }
-        out->num_elems++;
-    }
-
-    return;
-failed:
-    for (i = 0; i < out->num_elems; i++) {
-        if (out->data[i]) {
-            wasm_runtime_free(out->data[i]);
-        }
-    }
-
-    wasm_runtime_free(out->data);
+    bh_vector_init(out, src->num_elems, sizeof(WASMCApiFrame), false);
+    bh_memcpy_s(out->data, src->num_elems * sizeof(WASMCApiFrame), src->data,
+                src->num_elems * sizeof(WASMCApiFrame));
+    out->num_elems = src->num_elems;
 }
 
 static wasm_trap_t *
@@ -1981,13 +1963,8 @@ wasm_trap_new_internal(wasm_store_t *store,
         }
 
         for (i = 0; i < trap->frames->num_elems; i++) {
-#if WASM_ENABLE_THREAD_MGR != 0
-            ((wasm_frame_vec_t *)trap->frames)->data[i]->instance =
-                frame_instance;
-#else
             (((wasm_frame_t *)trap->frames->data) + i)->instance =
                 frame_instance;
-#endif
         }
     }
 #else
@@ -2082,11 +2059,7 @@ wasm_trap_trace(const wasm_trap_t *trap, own wasm_frame_vec_t *out)
     }
 
     for (i = 0; i < trap->frames->num_elems; i++) {
-#if WASM_ENABLE_THREAD_MGR != 0
-        wasm_frame_t *frame = ((wasm_frame_vec_t *)trap->frames)->data[i];
-#else
         wasm_frame_t *frame = ((wasm_frame_t *)trap->frames->data) + i;
-#endif
         if (!(out->data[i] =
                   wasm_frame_new(frame->instance, frame->module_offset,
                                  frame->func_index, frame->func_offset))) {
@@ -3425,7 +3398,7 @@ failed:
 
 #if WASM_ENABLE_DUMP_CALL_STACK != 0 && WASM_ENABLE_THREAD_MGR != 0
     WASMCluster *cluster = wasm_exec_env_get_cluster(exec_env);
-    cluster_frames = cluster->exception_frames;
+    cluster_frames = &cluster->exception_frames;
 #endif
     return wasm_trap_new_internal(
         func->store, func->inst_comm_rt,
--- a/core/iwasm/common/wasm_c_api.c
+++ b/core/iwasm/common/wasm_c_api.c
@@ -1907,7 +1907,12 @@ wasm_frame_vec_clone_internal(Vector *src, Vector *out)
 {
     bh_assert(src->num_elems != 0 && src->data);
 
-    bh_vector_init(out, src->num_elems, sizeof(WASMCApiFrame), false);
+    bh_vector_destroy(out);
+    if (!bh_vector_init(out, src->num_elems, sizeof(WASMCApiFrame), false)) {
+        bh_vector_destroy(out);
+        return;
+    }
+
     bh_memcpy_s(out->data, src->num_elems * sizeof(WASMCApiFrame), src->data,
                 src->num_elems * sizeof(WASMCApiFrame));
     out->num_elems = src->num_elems;
@@ -3399,10 +3404,17 @@ failed:
 #if WASM_ENABLE_DUMP_CALL_STACK != 0 && WASM_ENABLE_THREAD_MGR != 0
     WASMCluster *cluster = wasm_exec_env_get_cluster(exec_env);
     cluster_frames = &cluster->exception_frames;
+    wasm_cluster_traverse_lock(exec_env);
 #endif
-    return wasm_trap_new_internal(
+
+    wasm_trap_t *trap = wasm_trap_new_internal(
         func->store, func->inst_comm_rt,
         wasm_runtime_get_exception(func->inst_comm_rt), cluster_frames);
+
+#if WASM_ENABLE_DUMP_CALL_STACK != 0 && WASM_ENABLE_THREAD_MGR != 0
+    wasm_cluster_traverse_unlock(exec_env);
+#endif
+    return trap;
 }
 
 size_t
--- a/core/iwasm/common/wasm_c_api.c
+++ b/core/iwasm/common/wasm_c_api.c
@@ -1902,6 +1902,35 @@ wasm_frame_func_offset(const wasm_frame_t *frame)
     return frame ? frame->func_offset : 0;
 }
 
+void
+wasm_frame_vec_clone_internal(wasm_frame_vec_t *src, wasm_frame_vec_t *out)
+{
+    uint32 i;
+
+    wasm_frame_vec_new_uninitialized(out, src->num_elems);
+    if (out->size == 0 || !out->data) {
+        return;
+    }
+
+    for (i = 0; i < src->num_elems; i++) {
+        wasm_frame_t *frame = ((wasm_frame_t *)src->data) + i;
+        if (!(out->data[i] = wasm_frame_copy(frame))) {
+            goto failed;
+        }
+        out->num_elems++;
+    }
+
+    return;
+failed:
+    for (i = 0; i < out->num_elems; i++) {
+        if (out->data[i]) {
+            wasm_runtime_free(out->data[i]);
+        }
+    }
+
+    wasm_runtime_free(out->data);
+}
+
 static wasm_trap_t *
 wasm_trap_new_internal(wasm_store_t *store,
                        WASMModuleInstanceCommon *inst_comm_rt,
@@ -1952,7 +1981,7 @@ wasm_trap_new_internal(wasm_store_t *store,
         }
 
         for (i = 0; i < trap->frames->num_elems; i++) {
-            (((wasm_frame_t *)trap->frames->data) + i)->instance =
+            ((wasm_frame_vec_t *)trap->frames)->data[i]->instance =
                 frame_instance;
         }
     }
@@ -2048,10 +2077,7 @@ wasm_trap_trace(const wasm_trap_t *trap, own wasm_frame_vec_t *out)
     }
 
     for (i = 0; i < trap->frames->num_elems; i++) {
-        wasm_frame_t *frame;
-
-        frame = ((wasm_frame_t *)trap->frames->data) + i;
-
+        wasm_frame_t *frame = ((wasm_frame_vec_t *)trap->frames)->data[i];
         if (!(out->data[i] =
                   wasm_frame_new(frame->instance, frame->module_offset,
                                  frame->func_index, frame->func_offset))) {
--- a/core/iwasm/libraries/thread-mgr/thread_manager.c
+++ b/core/iwasm/libraries/thread-mgr/thread_manager.c
@@ -222,14 +222,6 @@ wasm_cluster_create(WASMExecEnv *exec_env)
     }
     memset(cluster, 0, sizeof(WASMCluster));
 
-#if WASM_ENABLE_DUMP_CALL_STACK != 0
-    if (!(cluster->exception_frames = wasm_runtime_malloc(sizeof(Vector)))) {
-        LOG_ERROR("thread manager error: failed to allocate exception frames");
-        return NULL;
-    }
-    memset(cluster->exception_frames, 0, sizeof(Vector));
-#endif
-
     exec_env->cluster = cluster;
 
     bh_list_init(&cluster->exec_env_list);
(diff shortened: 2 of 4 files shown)


eloparco:
@wenyongh Not sure I understand what CI is complaining about
```bash
CP libvmlib.a  <=  /home/runner/work/wasm-micro-runtime/wasm-micro-runtime/product-mini/platforms/linux-sgx/enclave-sample/../build/libvmlib.a
/usr/bin/ld: libvmlib.a(thread_manager.c.o): in function `wasm_cluster_set_exception':
thread_manager.c:(.text.wasm_cluster_set_exception+0x2a): undefined reference to `wasm_interp_create_call_stack'
/usr/bin/ld: thread_manager.c:(.text.wasm_cluster_set_exception+0x175): undefined reference to `wasm_interp_dump_call_stack'
```
I'm already guarding those 2 function calls with `#if WASM_ENABLE_DUMP_CALL_STACK != 0`.

wenyongh:
> @wenyongh Not sure I understand what CI is complaining about
> 
> ```shell
> CP libvmlib.a  <=  /home/runner/work/wasm-micro-runtime/wasm-micro-runtime/product-mini/platforms/linux-sgx/enclave-sample/../build/libvmlib.a
> /usr/bin/ld: libvmlib.a(thread_manager.c.o): in function `wasm_cluster_set_exception':
> thread_manager.c:(.text.wasm_cluster_set_exception+0x2a): undefined reference to `wasm_interp_create_call_stack'
> /usr/bin/ld: thread_manager.c:(.text.wasm_cluster_set_exception+0x175): undefined reference to `wasm_interp_dump_call_stack'
> ```
> 
> I'm already guarding those 2 function calls with `#if WASM_ENABLE_DUMP_CALL_STACK != 0`.

There are interpreter mode and aot mode, for the latter, we should use `aot_create_call_stack`, and if aot mode is compiled and interpreter isn't compiled, the above error may be reported.

eloparco:
@wenyongh In my last commit I pushed a change to make it work with the thread manager both enabled and disabled. I don't understand why that commit is required since in both cases a `Vector *` is used, but a different cast is necessary to access the elements properly.

As a side note, I was testing my changes with:
```c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <inttypes.h>

#include "wasm_c_api.h"

#define own

void print_frame(wasm_frame_t* frame) {
  printf("> %p @ 0x%zx = %"PRIu32".0x%zx\n",
    wasm_frame_instance(frame),
    wasm_frame_module_offset(frame),
    wasm_frame_func_index(frame),
    wasm_frame_func_offset(frame)
  );
}

int main(int argc, const char* argv[]) {
  // Initialize.
  printf("Initializing...\n");
  wasm_engine_t* engine = wasm_engine_new();
  wasm_store_t* store = wasm_store_new(engine);

  // Load binary.
  printf("Loading binary...\n");
#if WASM_ENABLE_AOT != 0 && WASM_ENABLE_INTERP == 0
  FILE* file = fopen("trap.aot", "rb");
#else
  FILE* file = fopen("trap-threads.wasm", "rb");
#endif
  if (!file) {
    printf("> Error loading module!\n");
    return 1;
  }

  int ret = fseek(file, 0L, SEEK_END);
  if (ret == -1) {
    printf("> Error loading module!\n");
    fclose(file);
    return 1;
  }

  long file_size = ftell(file);
  if (file_size == -1) {
    printf("> Error loading module!\n");
    fclose(file);
    return 1;
  }

  ret = fseek(file, 0L, SEEK_SET);
  if (ret == -1) {
    printf("> Error loading module!\n");
    fclose(file);
    return 1;
  }

  wasm_byte_vec_t binary;
  wasm_byte_vec_new_uninitialized(&binary, file_size);
  if (fread(binary.data, file_size, 1, file) != 1) {
    printf("> Error loading module!\n");
    fclose(file);
    return 1;
  }
  fclose(file);

  // Compile.
  printf("Compiling module...\n");
  own wasm_module_t* module = wasm_module_new(store, &binary);
  if (!module) {
    printf("> Error compiling module!\n");
    return 1;
  }

  wasm_byte_vec_delete(&binary);

  // Instantiate.
  printf("Instantiating module...\n");
  wasm_extern_vec_t imports = WASM_EMPTY_VEC;
  own wasm_instance_t* instance =
    wasm_instance_new(store, module, &imports, NULL);
  if (!instance) {
    printf("> Error instantiating module!\n");
    return 1;
  }

  // Extract export.
  printf("Extracting exports...\n");
  own wasm_extern_vec_t exports;
  wasm_instance_exports(instance, &exports);
  if (exports.size < 2) {
    printf("> Error accessing exports!\n");
    return 1;
  }

  wasm_module_delete(module);
  wasm_instance_delete(instance);

  // Call.
  int _start_function_index = 1;
  const wasm_func_t* func = wasm_extern_as_func(exports.data[_start_function_index]);
  if (func == NULL) {
    printf("> Error accessing export!\n");
    return 1;
  }

  printf("Calling export %d...\n", _start_function_index);
  wasm_val_vec_t args = WASM_EMPTY_VEC;
  wasm_val_vec_t results = WASM_EMPTY_VEC;
  own wasm_trap_t* trap = wasm_func_call(func, &args, &results);
  if (!trap) {
    printf("> Error calling function, expected trap!\n");
    return 1;
  }

  printf("Printing message...\n");
  own wasm_name_t message;
  wasm_trap_message(trap, &message);
  printf("> %s\n", message.data);
  assert(message.num_elems > 0);
  assert(strncmp(message.data, "Exception: ", strlen("Exception: ")) == 0);

  printf("Printing origin...\n");
  own wasm_frame_t* frame = wasm_trap_origin(trap);
  if (frame) {
    print_frame(frame);
    wasm_frame_delete(frame);
  } else {
    printf("> Empty origin.\n");
  }

  printf("Printing trace...\n");
  own wasm_frame_vec_t trace;
  wasm_trap_trace(trap, &trace);
  if (trace.size > 0) {
    for (size_t i = 0; i < trace.size; ++i) {
      print_frame(trace.data[i]);
    }
  } else {
    printf("> Empty trace.\n");
  }

  wasm_frame_vec_delete(&trace);
  wasm_trap_delete(trap);
  wasm_name_delete(&message);

  wasm_extern_vec_delete(&exports);

  // Shut down.
  printf("Shutting down...\n");
  wasm_store_delete(store);
  wasm_engine_delete(engine);

  // All done.
  printf("Done.\n");
  return 0;
}
```
(copied from `samples/wasm-c-api/src/trap.c` with small modifications, it requires setting `WAMR_BUILD_LIB_WASI_THREADS`)

`trap-threads.wasm` was generated from
```rust
fn c() { panic!("PANIC!!!"); }
fn b() { c(); }
fn a() { b(); }

fn main() {
    let handler = std::thread::spawn(|| {
        a();
    });
    handler.join().unwrap();
}
```
And it seems to work properly, i.e. it prints the stack trace of the spawned thread instead of the main thread.

core/iwasm/libraries/thread-mgr/thread_manager.h
```diff
@@ -51,6 +51,13 @@ struct WASMCluster {
 #if WASM_ENABLE_DEBUG_INTERP != 0
     WASMDebugInstance *debug_inst;
 #endif
+
+#if WASM_ENABLE_DUMP_CALL_STACK != 0
+    /* When an exception occurs in a thread, the stack frames of that thread are
+     * saved into the claster
+     */
+    Vector *exception_frames;
```
loganek, line 59: Does it have to be a pointer or can it be a value? We could reduce a number of heap allocations (not much the actual size itself though) and simplify the initialization code.

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -221,6 +222,14 @@ wasm_cluster_create(WASMExecEnv *exec_env)
     }
     memset(cluster, 0, sizeof(WASMCluster));
 
+#if WASM_ENABLE_DUMP_CALL_STACK != 0
+    if (!(cluster->exception_frames = wasm_runtime_malloc(sizeof(Vector)))) {
+        LOG_ERROR("thread manager error: failed to allocate exception frames");
+        return NULL;
+    }
+    memset(cluster->exception_frames, 0, sizeof(Vector));
```
loganek, line 230: Should you use `bh_vector_init` instead?

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -1302,6 +1317,34 @@ wasm_cluster_set_exception(WASMExecEnv *exec_env, const char *exception)
...
+#if WASM_ENABLE_DUMP_CALL_STACK != 0
+    if (has_exception) {
+        /* Save the stack trace of the crashed thread into the cluster */
+        WASMModuleInstance *module_inst =
+            (WASMModuleInstance *)get_module_inst(exec_env);
+
+#if WASM_ENABLE_INTERP != 0
+        if (module_inst->module_type == Wasm_Module_Bytecode
+            && wasm_interp_create_call_stack(exec_env)) {
+            wasm_interp_dump_call_stack(exec_env, false, NULL, 0);
```
loganek, line 1329: What's the point of the call? We ignore the return value, we also pass a NULL buffer and don't print the stack.

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -1302,6 +1317,34 @@ wasm_cluster_set_exception(WASMExecEnv *exec_env, const char *exception)
...
+            wasm_frame_vec_clone_internal(
+                (wasm_frame_vec_t *)module_inst->frames,
+                (wasm_frame_vec_t *)cluster->exception_frames);
+        }
+#endif
+
+#if WASM_ENABLE_AOT != 0
+        if (module_inst->module_type == Wasm_Module_AoT
+            && aot_create_call_stack(exec_env)) {
+            aot_dump_call_stack(exec_env, true, NULL, 0);
```
loganek, line 1339: same here

core/iwasm/common/wasm_c_api.c
```diff
@@ -1902,10 +1902,39 @@ wasm_frame_func_offset(const wasm_frame_t *frame)
     return frame ? frame->func_offset : 0;
 }
 
+void
+wasm_frame_vec_clone_internal(wasm_frame_vec_t *src, wasm_frame_vec_t *out)
```
loganek, line 1906: would it make sense to implement a generic `bh_vector_copy` and move it to `bh_vector.{c|h}`?

eloparco: In principle I should be able to use https://github.com/bytecodealliance/wasm-micro-runtime/blob/6dbfeb25dd164c0ffcec21806e1c1cd0dff27c58/core/iwasm/common/wasm_c_api.c#L226 but I was having some problems accessing the data https://github.com/bytecodealliance/wasm-micro-runtime/pull/2908#issuecomment-1860794770

core/iwasm/libraries/thread-mgr/thread_manager.c
eloparco: The `print=true` is a typo, it should be `false` there as well, thanks for noticing. The calls to `wasm_interp_create_call_stack()` and `wasm_interp_dump_call_stack()` are just to dump the stack trace so that it can be copied into the thread manager.

loganek: I understand why `wasm_interp_create_call_stack` is required but I don't understand what's the purpose of calling `wasm_interp_create_call_stack` given that you pass NULL buffer, disable printing and ignoring return value?

eloparco: True, I misread the function body, I thought it was populating some of the frame fields but it does not. So, yeah, that function call can be removed.

core/iwasm/common/wasm_c_api.c
loganek: Ah yes, you should be able to use that function instead. Can you please debug the code step by step then? Quickly looking at the code it seems like it does the same thing, so would be great to re-use it. Please note that the order of parameters in the vec_copy is swapped (i.e. `out` is a first parameter).

yamt:
> When using the wasm-c-api and there's a trap, `wasm_func_call()` returns a `wasm_trap_t *` object. No matter which thread crashes, the trap contains the stack frames of the main thread.

do you mean, you are using wasm_func_call to call a "main" function, which internally creates "sub" threads?
in that case, to me, it seems reasonable to return the stack frames which includes the main function.

core/iwasm/common/wasm_c_api.c
```diff
@@ -3385,9 +3423,13 @@ wasm_func_call(const wasm_func_t *func, const wasm_val_vec_t *params,
...
         wasm_runtime_free(argv);
 
+#if WASM_ENABLE_DUMP_CALL_STACK != 0 && WASM_ENABLE_THREAD_MGR != 0
+    WASMCluster *cluster = wasm_exec_env_get_cluster(exec_env);
+    cluster_frames = cluster->exception_frames;
+#endif
     return wasm_trap_new_internal(
         func->store, func->inst_comm_rt,
-        wasm_runtime_get_exception(func->inst_comm_rt));
+        wasm_runtime_get_exception(func->inst_comm_rt), cluster_frames);
```
wenyongh, line 3432: @eloparco IIUC, here `func->inst_comm_rt` always points to the main module instance, so wasm_trap_new_internal uses the wrong frames, right? If we use the instance of the current thread, can the issue be resolved?
How about we adding an API in thread_mgr.c/h to lookup the exec_env of current thread, and then using `exec_env->module_inst` here instead of `func->inst_comm_rt`?
Somewhat like wasm_clusters_search_exec_env(module_inst), but traversing the clusters and finding the exec_env whose handle equals to os_self_thread().

eloparco:
>do you mean, you are using wasm_func_call to call a "main" function, which internally creates "sub" threads?

Yes

> in that case, to me, it seems reasonable to return the stack frames which includes the main function.

But the stack trace of the main thread doesn't carry information about what caused the trap. If something crashed in the spawned thread we want to know what made it happen.

The alternative would be to return the stack frames of all threads, but I wanted to stick the wasm-c-api (that returns a single `wasm_trap_t *`). And I think that the stack trace of the crashed thread is probably enough in most scenarios (to troubleshoot the problem).

I'll resume the work once I'm back from vacation.

yamt:
> > do you mean, you are using wasm_func_call to call a "main" function, which internally creates "sub" threads?
> 
> Yes

ok.

> 
> > in that case, to me, it seems reasonable to return the stack frames which includes the main function.
> 
> But the stack trace of the main thread doesn't carry information about what caused the trap. If something crashed in the spawned thread we want to know what made it happen.

i can understand.
but as far as i understand, wasm-c-api doesn't have an api appropriate for the purpose.

>
> The alternative would be to return the stack frames of all threads, but I wanted to stick the wasm-c-api (that returns a single `wasm_trap_t *`). And I think that the stack trace of the crashed thread is probably enough in most scenarios (to troubleshoot the problem).

i suspect you need to invent an extra api either ways.

> 
> I'll resume the work once I'm back from vacation.

enjoy your vacation!

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -1302,6 +1317,34 @@ wasm_cluster_set_exception(WASMExecEnv *exec_env, const char *exception)
...
+            wasm_frame_vec_clone_internal(
+                (wasm_frame_vec_t *)module_inst->frames,
+                (wasm_frame_vec_t *)cluster->exception_frames);
```
wenyongh, line 1342: Invalid use of `module_inst->frames`, its type is `Vector *` and each element is `wasm_frame_t` or `WASMCApiFrame`, the element size is `sizeof(WASMCApiFrame)`. While element type of wasm_frame_vect_t is `wasm_frame_t *` (see WASM_DECLARE_VEC(frame, *) in wasm_c_api.h)., the element size is `sizeof(void *)`.

I guess the issue that using WASM_ENABLE_THREAD_MGR to control accessing the frame element is caused by this. The `cluster->exception_frames` should be declared as `Vector *`, and the clone algorithm need to updated accordingly.

core/iwasm/common/wasm_c_api.c
eloparco: `wasm_trap_new_internal()` is called from the main thread and, at the time of the call, the exec_env of the spawned thread has already been freed here https://github.com/bytecodealliance/wasm-micro-runtime/blob/722381fcf2861596cb4c8ed7a6019deaabb14706/core/iwasm/libraries/thread-mgr/thread_manager.c#L668 so even saving in the cluster the thread id or directly the module instance of the crashed thread wouldn't work.

What about leaving it as it is? If a spawned thread crashes, the module instance reported in the frames is NULL. Not sure if there are use case in which the module instance stored in the frames is needed (not in mine at least).

core/iwasm/libraries/thread-mgr/thread_manager.c
eloparco: Yes, thanks, that's exactly what was happening there

eloparco:
> i can understand.
but as far as i understand, wasm-c-api doesn't have an api appropriate for the purpose.

The current API already returns a `wasm_trap_t *` and I think it would be more appropriate to populate it with the stack frames of the crashed thread rather than the main thread. 

Like when you're debugging with gdb and there's a crash, you get stopped on the crashed thread and running `(gdb) backtrace` returns the stack trace of that crashed thread.

Then yeah, it probably makes sense to have an additional API that would return the stack frames for all threads.

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -1302,6 +1307,30 @@ wasm_cluster_set_exception(WASMExecEnv *exec_env, const char *exception)
...
+    if (has_exception) {
+        /* Save the stack trace of the crashed thread into the cluster */
+        WASMModuleInstance *module_inst =
+            (WASMModuleInstance *)get_module_inst(exec_env);
+
+#if WASM_ENABLE_INTERP != 0
+        if (module_inst->module_type == Wasm_Module_Bytecode
+            && wasm_interp_create_call_stack(exec_env)) {
+            wasm_frame_vec_clone_internal(module_inst->frames,
+                                          &cluster->exception_frames);
```
wenyongh, line 1320: Two threads may set exception simultaneously, and access `cluster->exception_frames` at the same time? Had better move these lines to be after `os_mutex_lock(&cluster->lock)` below?

core/iwasm/common/wasm_c_api.c
wenyongh: OK, a concern is that there may be data racing in accessing the `cluster->exception_frames`: suppose another thread is `wasm_cluster_set_exception`. Had better acquire the `cluster->lock` with `wasm_cluster_traverse_lock` before calling `wasm_trap_new_internal`?

core/iwasm/common/wasm_c_api.c
```diff
@@ -1902,10 +1902,21 @@ wasm_frame_func_offset(const wasm_frame_t *frame)
     return frame ? frame->func_offset : 0;
 }
 
+void
+wasm_frame_vec_clone_internal(Vector *src, Vector *out)
+{
+    bh_assert(src->num_elems != 0 && src->data);
+
+    bh_vector_init(out, src->num_elems, sizeof(WASMCApiFrame), false);
```
wenyongh, line 1910: Should check the return value, there is memory allocation in be_vector_init.

And seems we should `bh_vectory_destroy` it before init, it may have been initialized, and re-initializing it may cause memory leak?

yamt:
> > i can understand.
> > but as far as i understand, wasm-c-api doesn't have an api appropriate for the purpose.
> 
> The current API already returns a `wasm_trap_t *` and I think it would be more appropriate to populate it with the stack frames of the crashed thread rather than the main thread.

it might be appropriate for your use cases. it might not for others.
i personally feel it's less surprising to return the frames of the executed function.

after all, wasm-c-api in not a wamr-specific api. maybe you can make a proposal there if you want.

loganek:
I agree with @yamt that before changing the behavior of the existing API, it's worth discussing it in a different forum. Having said that, IMO a new API that returns stack traces from all the functions would be quite useful for the debugging purposes, and I think it'll serve your purpose @eloparco - even though often having a stack of crashed thread is enough, sometimes it's necessary what was happening in other threads to narrow down the problem.

wenyongh:
> I agree with @yamt that before changing the behavior of the existing API, it's worth discussing it in a different forum. Having said that, IMO a new API that returns stack traces from all the functions would be quite useful for the debugging purposes, and I think it'll serve your purpose @eloparco - even though often having a stack of crashed thread is enough, sometimes it's necessary what was happening in other threads to narrow down the problem.

@loganek so should I merge the PR? It seems that all the issues were resolved and it looks good to me now.

loganek:
Given the expectation from this API is not very well defined, I'm ok with the implementation proposed by @eloparco . However, this is a breaking change and must be highlighted in release notes. I don't know how many usecases will be affected though by that breaking change. I presume this is only for debugging / investigation purposes, so likely not many?

wenyongh:
> Given the expectation from this API is not very well defined, I'm ok with the implementation proposed by @eloparco . However, this is a breaking change and must be highlighted in release notes. I don't know how many usecases will be affected though by that breaking change. I presume this is only for debugging / investigation purposes, so likely not many?

Agree, I think there should be few cases affected, one is there should be few developers using wasm-c-api with multi-threading support, another is that dump-call-stack feature is usually disabled and is enabled mainly for debugging/investigation. For the latter, this PR helps to locate the actual issue. So I think it should acceptable, and yes, we can highlight the breaking change in the release notes. How do you think?

wenyongh:
@loganek I merged the PR, and let's highlight it in release notes.

loganek:
Sounds good, thanks

eloparco:
Thanks for merging, I wanted to point out that reporting only the stack trace of the crashed thread is also the default behaviour when running Wasmtime (tried with the latest dev branch). Running the same wasm file I was using with the wasm-c-api:
```bash
target/debug/wasmtime --dir . -W threads -S threads trap-threads.wasm
thread '<unnamed>' panicked at src/main.rs:2:5:
PANIC!!!
Error: error while executing at wasm backtrace:
    0: 0xcc2c - <unknown>!__rust_start_panic
    1: 0xc711 - <unknown>!rust_panic
    2: 0xc698 - <unknown>!std::panicking::rust_panic_with_hook::h6c1623ae744881b6
    3: 0xb625 - <unknown>!std::panicking::begin_panic_handler::{{closure}}::h8474ba3fc2a2f64f
    4: 0xb58a - <unknown>!std::sys_common::backtrace::__rust_end_short_backtrace::h2d8a3004b92235a1
    5: 0xbfeb - <unknown>!rust_begin_unwind
    6: 0x121f3 - <unknown>!core::panicking::panic_fmt::h4d6ec91b4ec3c102
    7: 0x6943 - <unknown>!hello_cargo::c::hde6caa0faf591098
    8: 0x694d - <unknown>!hello_cargo::b::hf5ff17be23138f3f
    9: 0x6957 - <unknown>!hello_cargo::a::hfef53f62412b3f11
   10: 0x3efc - <unknown>!hello_cargo::main::{{closure}}::h2356b4434e2f30cd
   11: 0x6e65 - <unknown>!std::sys_common::backtrace::__rust_begin_short_backtrace::hc386d72b956247a7
   12: 0x314d - <unknown>!std::thread::Builder::spawn_unchecked_::{{closure}}::{{closure}}::h2008147d7c9f7289
   13: 0x74a0 - <unknown>!<core::panic::unwind_safe::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once::h39f99754f0e7004a
   14:  0x668 - <unknown>!std::panicking::try::do_call::h3bec9c3b34ebe264
   15:  0x5b1 - <unknown>!std::panicking::try::ha20e4a80b1643c99
   16: 0x304c - <unknown>!std::thread::Builder::spawn_unchecked_::{{closure}}::h4e8c3a41a9981963
   17: 0x3f5f - <unknown>!core::ops::function::FnOnce::call_once{{vtable.shim}}::h2ffb3ae107c3a7de
   18: 0xcbbe - <unknown>!std::sys::wasi::thread::Thread::new::thread_start::h3c4c4269952012d9
   19: 0x10e77 - <unknown>!__wasi_thread_start_C
   20: 0x119c9 - <unknown>!wasi_thread_start
note: using the `WASMTIME_BACKTRACE_DETAILS=1` environment variable may show more debugging information

Caused by:
    wasm trap: wasm `unreachable` instruction executed
```

If I have some time in the future, I'll investigate how to report stack frames from all threads (I guess we would have to use a similar approach and modify the `wasm_trap_t` struct, plus adding some additional APIs).

yamt:
> If I have some time in the future, I'll investigate how to report stack frames from all threads (I guess we would have to use a similar approach and modify the `wasm_trap_t` struct, plus adding some additional APIs).

have you done so since then?
have you proposed this to wasm-c-api?

otherwise, i'd suggest to revert this because:

* this is a non-standard behavior, surprising users.
* it's reported broken: https://github.com/bytecodealliance/wasm-micro-runtime/discussions/3800#discussion-7202551

PR #2911, opened 2023-12-19
Emit SIMD flags to AOT file if SIMD actually used

no1wudi:


core/iwasm/interpreter/wasm_loader.c
```diff
@@ -13576,6 +13581,11 @@ wasm_loader_prepare_bytecode(WASMModule *module, WASMFunction *func,
             {
                 uint32 opcode1;
 
+#if WASM_ENABLE_WAMR_COMPILER != 0
+                /* Mark the SIMD instruction is used in this module */
+                module->is_simd_used = true;
```
wenyongh, line 13586: Should we add more checks? e.g., whether a function type's param/result type is v128, whether a function's local type is v128, whether a global type is v128?

wamr-compiler/main.c
```diff
@@ -632,6 +634,15 @@ main(int argc, char *argv[])
...
     }
 
+    /* Only if SIMD is actually used, we need to set the compiler
+     * options */
+    if (((WASMModule *)wasm_module)->is_simd_used) {
+        option.enable_simd = true;
+    }
+    else {
+        option.enable_simd = false;
+    }
```
wenyongh, line 644: Is it better to set the flag inside aot_create_comp_data? The `wasm_module` is passed into the function.

no1wudi: If that, handle it in aot_create_comp_context seems better.

wenyongh: Yes, maybe it is better to handle it in aot_create_comp_context.

core/iwasm/compilation/aot_llvm.c
```diff
@@ -3068,6 +3069,17 @@ aot_create_comp_context(const AOTCompData *comp_data, aot_comp_option_t option)
...
     LLVMDisposeMessage(triple);
 
+#if WASM_ENABLE_WAMR_COMPILER != 0
+    /* Only if SIMD is actually used, we need to set the compiler
+     * options */
+    if (((WASMModule *)comp_data->wasm_module)->is_simd_used) {
+        option->enable_simd = true;
+    }
+    else {
+        option->enable_simd = false;
```
wenyongh, line 3079: Can just `option->enable_simd = comp_data->wasm_module->is_simd_used;`

A concern is that then `wamrc --disable-simd` is omitted, should we do more for it?

no1wudi: Yes in this implementation --disable-simd is omitted, so should we raise error if --disable-simd is set but SIMD is used in the module?

wenyongh: Agree, it should be better.

wenyongh (commented):
LGTM

PR #2924, opened 2023-12-25
Enable wasm_runtime_terminate for single-threading

wenyongh:
The host embedder may also want to terminate the wasm instance
for single-threading mode, and it should work by setting exception
to the wasm instance.

### DIFF
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -2548,7 +2548,6 @@ wasm_runtime_clear_exception(WASMModuleInstanceCommon *module_inst_comm)
     wasm_runtime_set_exception(module_inst_comm, NULL);
 }
 
-#if WASM_ENABLE_THREAD_MGR != 0
 void
 wasm_runtime_terminate(WASMModuleInstanceCommon *module_inst_comm)
 {
@@ -2558,7 +2557,6 @@ wasm_runtime_terminate(WASMModuleInstanceCommon *module_inst_comm)
               || module_inst_comm->module_type == Wasm_Module_AoT);
     wasm_set_exception(module_inst, "terminated by user");
 }
-#endif
 
 void
 wasm_runtime_set_custom_data_internal(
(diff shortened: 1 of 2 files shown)


lum1n0us:
looks good to me.  But, maybe we should hear if @yamt is OK with that.

we are hoping to terminate one instance from another instance in all below scenarios:
- multiple instances in one thread (new case)
- multiple instances in multiple host managed threads (new case)
- multiple instances in multiple runtime managed threads (original scenario)

yamt:
how is it expected to be called? by a host function in the instance itself?

yamt:
> looks good to me. But, maybe we should hear if @yamt is OK with that.
> 
> we are hoping to terminate one instance from another instance in all below scenarios:
> 
>     * multiple instances in one thread (new case)
> 
>     * multiple instances in multiple host managed threads (new case)
> 
>     * multiple instances in multiple runtime managed threads (original scenario)

do you mean, w/o enabling THREAD_MGR?

lum1n0us:
- to `wasm_terminate_instance()` from host side in or out a instance.
- because of the case "multiple instances in one thread", we hope to use `wasm_terminate_instance()` w/o enabling THREAD_MGR.

yamt:
> * to `wasm_terminate_instance()` from host side in or out a instance.
> 
>     * because of the case "multiple instances in one thread", we hope to use `wasm_terminate_instance()` w/o enabling THREAD_MGR.

so, "multiple instances in multiple host managed threads (new case)" will still require THREAD_MGR enabled?

lum1n0us:
may not require too.

yamt:
> may not require too.

well, do you mean you may modify it so in future?
the current implementation requires THREAD_MGR in that case, if i remember correctly.

lum1n0us:
We hope `wasm_terminate_instance()` is not related with `THREAD_MGR`. It should always workable in both `THREAD_MGR=0` and `THREAD_MGR=1`

yamt:
> We hope `wasm_terminate_instance()` is not related with `THREAD_MGR`. It should always workable in both `THREAD_MGR=0` and `THREAD_MGR=1`

i tend to agree THREAD_MGR is controlling too many things.

yamt (approved):
it's fine for me to enable this for the use case where an instance terminates itself.
other cases may need more work.

wenyongh:
Yes, enabling thread manager may add more checks in the AOT code and impact the performance. But when thread manager isn't enabled, whether the exception is thrown is also checked in some places of the AOT code and runtime, which makes the module instance exit in most cases. If we don't want to lose performance a lot, we may use this PR.

PR #2925, opened 2023-12-25
Refine JIT function call process

wenyongh:
- Don't allocate the implicit/unused frame when calling the LLVM JIT function.
- Don't set exec_env's thread handle and stack boundary in the recursive
  calling from host, since they have been set in the first time calling.
- Fix frame not freed in llvm_jit_call_func_bytecode


yamt (approved):
lgtm.
it would have been nicer if these two changes are separate PRs though.

wenyongh:
> lgtm. it would have been nicer if these two changes are separate PRs though.

OK, I will separate this to two PRs.

wenyongh:
> > lgtm. it would have been nicer if these two changes are separate PRs though.
> 
> OK, I will separate this to two PRs.

Done with PR #2940 created.

Issue #2736, opened 2023-11-08
it's nicer to have a runtime check of wrgsbase for segue stuff


yamt:
sometimes you can't have a build-time knowledge about the availability of wrgsbase.
it would be nicer to have a runtime check and disable it accordingly.
cf. https://sourceforge.net/p/valgrind/mailman/message/36612725/

PR #2926, opened 2023-12-25
Allow to set segue flags for wasm-c-api JIT

wenyongh:


lum1n0us (approved):
LGTM

core/iwasm/common/wasm_c_api.c
```diff
@@ -299,6 +299,11 @@ wasm_config_new(void)
 
     memset(config, 0, sizeof(wasm_config_t));
     config->mem_alloc_type = Alloc_With_System_Allocator;
+#if defined(os_writegsbase)
+    /* enable segue for all load/store operations */
+    config->segue_flags = 0x1F1F;
```
yamt, line 304: i feel it's safer to disable segue stuff by default.
https://github.com/bytecodealliance/wasm-micro-runtime/issues/2736
how do you think?

core/iwasm/include/wasm_c_api.h
```diff
@@ -200,6 +201,9 @@ wasm_config_set_mem_alloc_opt(wasm_config_t *, mem_alloc_type_t, MemAllocOption
 WASM_API_EXTERN own wasm_config_t*
 wasm_config_set_linux_perf_opt(wasm_config_t *, bool);
 
+WASM_API_EXTERN wasm_config_t*
+wasm_config_set_segue_flags(wasm_config_t *config, uint32_t segue_flags);
```
yamt, line 205: do we have documentation about these flag bits?

core/iwasm/common/wasm_c_api.c
wenyongh: OK, maybe some targets don't support modifying GS register during execution, let's disable it by default.

core/iwasm/include/wasm_c_api.h
wenyongh: Not yet, how about adding comment here? For iwasm and wamrc, we had added comment in the command help info.

yamt: > Not yet, how about adding comment here? For iwasm and wamrc, we had added comment in the command help info.

it sounds reasonable to me.

core/iwasm/common/wasm_c_api.c
wenyongh: Done.

core/iwasm/include/wasm_c_api.h
wenyongh: Done.

PR #2933, opened 2023-12-27
Append .aot to .wasm as a custom section named "aot"

lum1n0us:


yamt:
how do you plan to use the section?

lum1n0us:
a [pre_compiled](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/wasm/v3/wasm.proto) wasm plugin for Envoy

yamt:
> a [pre_compiled](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/wasm/v3/wasm.proto) wasm plugin for Envoy

how does it load and execute the aot section?
i'm asking because i have alignment concerns.

lum1n0us:
thanks for reminding. in a nutshell, envoy will load aot custom section and pass the content(in a string_view) to WAMR.  https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/master/src/wasm.cc#L293

yamt:
> thanks for reminding. in a nutshell, envoy will load aot custom section and pass the content(in a string_view) to WAMR. https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/master/src/wasm.cc#L293

i suppose that it's safer to make some padding so that the start offset of `section_content` within the file will have the max alignment as we can use in our aot format. (4 byte?)

lum1n0us:
the section layout likes:

```
| section id(0) | section length | section name | section content(.aot) |
```

IIUC, please correct me if wrong, the 4 byte padding goes to the beginning of *section content*, and the consumer(WAMR or embedder) needs to ignore the 4 padding.

yamt:
> the section layout likes:
> 
> ```
> | section id(0) | section length | section name | section content(.aot) |
> ```
> 
> IIUC, please correct me if wrong, the 4 byte padding goes to the beginning of _section content_, and the consumer(WAMR or embedder) needs to ignore the 4 padding.

yea, i guess it's simplest to add the padding at the top of the custom section body.
like:
```
 | section id(0) | section length | section name | padding len | padding (zeros) | section content(.aot) |
```
note that the implementation might be a bit tricky as "section length" is variable lengh (leb128) and it affects the padding length.
you might want to use a redundant encoding of leb128. (eg. always use 5 byte)

wenyongh (approved):

Issue #2951, opened 2023-12-31
Unexpect excrption (integer overflow) on fast interpreter mode


erxiaozhou:
# Build commands
I compile the code with commit id 92bf8547aa64dddad2ecd387a352f89bbc475c88.
Platform: Ubuntu 20.04
CPU: amd64 
compile:

```
export CC=/usr/lib/llvm-16/bin/clang
export CXX=/usr/lib/llvm-16/bin/clang++
cd product-mini/platforms/linux/;rm -rf build
cmake  -DWAMR_BUILD_INTERP=1 -DWAMR_BUILD_REF_TYPES=1 -DWAMR_BUILD_SIMD=1 -DWASM_ENABLE_BULK_MEMORY=1 -DWAMR_BUILD_AOT=0 -DCMAKE_BUILD_TYPE=Release -DWAMR_BUILD_FAST_INTERP=1 -DWAMR_BUILD_LIBC_WASI=0 -DWAMR_BUILD_LIBC_BUILTIN=1 -Bbuild
```

execute:

```
iwasm --heap-size=0  <tc_name>
```
# Cases

[iwasm_fast_jit_unexpected_int_overflow.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/13801277/iwasm_fast_jit_unexpected_int_overflow.zip)

# Actual behavior:

```
Exception: integer overflow
```

# Expected behavior:
No exception

wenyongh:
Fast JIT works normal, it should be the issue of fast-interp according to the build command and test result.

erxiaozhou:
> Fast JIT works normal, it should be the issue of fast-interp according to the build command and test result.

Yes, you are right.

xujuntwt95329:
Thanks for reporting this, we've fixed in https://github.com/bytecodealliance/wasm-micro-runtime/pull/2974

wenyongh:
Close this issue since it was fixed.

Issue #2953, opened 2023-12-31
Unexpect exception (allocate memory failed) on fast interpreter mode


erxiaozhou:
# Build commands
I compile the code with commit id 92bf8547aa64dddad2ecd387a352f89bbc475c88.
Platform: Ubuntu 20.04
CPU: amd64 
compile:

```
export CC=/usr/lib/llvm-16/bin/clang
export CXX=/usr/lib/llvm-16/bin/clang++
cd product-mini/platforms/linux/;rm -rf build
cmake  -DWAMR_BUILD_INTERP=1 -DWAMR_BUILD_REF_TYPES=1 -DWAMR_BUILD_SIMD=1 -DWASM_ENABLE_BULK_MEMORY=1 -DWAMR_BUILD_AOT=0 -DCMAKE_BUILD_TYPE=Release -DWAMR_BUILD_FAST_INTERP=1 -DWAMR_BUILD_LIBC_WASI=0 -DWAMR_BUILD_LIBC_BUILTIN=1 -Bbuild
```

execute:

```
iwasm --heap-size=0  <tc_name>
```
# Cases

[iwasm_fast_interp_unexpected_exception2.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/13801304/iwasm_fast_interp_unexpected_exception2.zip)

# Actual behavior:

```
Exception: allocate memory failed
```

# Expected behavior:
No exception
```
0x16006d52ce3ebd1a:i64
```

xujuntwt95329:
Thanks for reporting this, we've fixed in https://github.com/bytecodealliance/wasm-micro-runtime/pull/2974

wenyongh:
Close this issue since it was fixed.

PR #2974, opened 2024-01-03
fix fast-interp polymorphic stack processing

xujuntwt95329:


xujuntwt95329:
fix #2951 #2952 #2953

wenyongh (approved):
LGTM

PR #2975, opened 2024-01-03
Fix test script error and related wamrc and GC issues

wenyongh:


core/iwasm/aot/aot_runtime.c
```diff
@@ -1684,8 +1686,14 @@ aot_instantiate(AOTModule *module, AOTModuleInstance *parent,
     if (stack_size == 0)
         stack_size = DEFAULT_WASM_STACK_SIZE;
 #if WASM_ENABLE_SPEC_TEST != 0
+#if WASM_ENABLE_TAIL_CALL == 0
     if (stack_size < 128 * 1024)
         stack_size = 128 * 1024;
+#else
+    /* Some tail-call cases require large operand stack */
+    if (stack_size < 10 * 1024 * 1024)
```
no1wudi, line 1694: Tail call case already patched in this commit, should we enlarge the stack size here?

.github/workflows/spec_test_on_nuttx.yml
```diff
@@ -159,6 +159,10 @@ jobs:
           echo "config INTERPRETERS_WAMR_AOT_STACK_FRAME" >> apps/interpreters/wamr/Kconfig
           echo "\tbool \"Enable AOT stack frame\"" >> apps/interpreters/wamr/Kconfig
           echo "\tdefault n" >> apps/interpreters/wamr/Kconfig
+          echo "\n" >> apps/interpreters/wamr/Kconfig
+          echo "config INTERPRETERS_WAMR_TAIL_CALL" >> apps/interpreters/wamr/Kconfig
```
no1wudi, line 163: I think remove the nuttx specific config is better

wenyongh: No idea how to do it? If I remove these lines, CI reported somewhat like "aot load fail: tail call is not enabled"

core/iwasm/aot/aot_runtime.c
wenyongh: The count of tail call for linux was patched but it is also large (~100000), it needs large operand stack. Do you suggest to decrease the size for both linux and nuttx?

.github/workflows/spec_test_on_nuttx.yml
no1wudi: Get it, so it's an missing option on NuttX?

core/iwasm/aot/aot_runtime.c
no1wudi: Is [gc_nuttx_tail_call.patch](https://github.com/bytecodealliance/wasm-micro-runtime/pull/2975/files#diff-6098a9fa70e8d4002392d870fe51ae3c952e347ed0235427dd4a56803fa05ea4) only for NuttX or not?

If this file is only for NuttX, maybe we can make it smaller?
If not, this file should be renamed

.github/workflows/spec_test_on_nuttx.yml
wenyongh: Yes, and I added an option in `product-mini/platforms/nuttx/wamr.mk`.

core/iwasm/aot/aot_runtime.c
wenyongh: It is only for nuttx. For GC spec test, we first apply patch `gc_ignore_cases.patch`, in which the count is reduced to about 100000, which is good for Ubuntu. And if we are testing nuttx qemu, then `gc_nuttx_tail_call.patch` is applied again to reduce the count to about 1000.

no1wudi: So, ` if (stack_size < 10 * 1024 * 1024)` can be removed here?

.github/workflows/spec_test_on_nuttx.yml
no1wudi: Thanks, I'll submit this to NuttX side.

core/iwasm/aot/aot_runtime.c
wenyongh: No, gc spec test in Ubuntu requires large operand stack, though the recursive count for tail call has been reduced to 100000.

no1wudi (approved):

core/iwasm/aot/aot_runtime.c
```diff
@@ -1684,8 +1686,14 @@ aot_instantiate(AOTModule *module, AOTModuleInstance *parent,
     if (stack_size == 0)
         stack_size = DEFAULT_WASM_STACK_SIZE;
 #if WASM_ENABLE_SPEC_TEST != 0
+#if WASM_ENABLE_TAIL_CALL == 0
     if (stack_size < 128 * 1024)
         stack_size = 128 * 1024;
+#else
+    /* Some tail-call cases require large operand stack */
```
yamt, line 1693: which test cases?

wenyongh: Some opcode return_call_indirect and return_call related cases which are added in gc spec proposal repo (and are not added in the default [spec](https://github.com/WebAssembly/spec/tree/main/test/core) repo):
https://github.com/WebAssembly/gc/blob/master/test/core/return_call_indirect.wast#L260-L261
https://github.com/WebAssembly/gc/blob/master/test/core/return_call_indirect.wast#L266-L267
https://github.com/WebAssembly/gc/blob/master/test/core/return_call.wast#L105
https://github.com/WebAssembly/gc/blob/master/test/core/return_call.wast#L111-L112
Since in the LLVM IR translated, we recursively call the function but not re-use the native stack (and operand stack if AOT/JIT stack frame is enabled) to call it, it requires lots of native/operand stack. Maybe we can optimize this in the future?

PR #2976, opened 2024-01-03
Refactor aot stack frame commit

wenyongh:
- Commit locals, stacks and stack pointer to aot frame only when gc is enabled
- Commit instruction pointer to aot frame when stack frame is enabled
- Refine alloc/free aot frame when gc isn't enabled: use fixed frame size
- Support dump call stack with bytecode offset


core/iwasm/compilation/aot_emit_exception.c
```diff
@@ -51,6 +80,16 @@ aot_emit_exception(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
             return false;
         }
 
+        if (comp_ctx->aot_frame) {
+            /* Create exection ip phi */
```
xujuntwt95329, line 84:
```suggestion
            /* Create exception ip phi */
```

wenyongh: Done, thanks!

PR #2978, opened 2024-01-04
Register call wrappers to speedup the aot/jit func call process

wenyongh:
In some scenarios there may be lots of callings to AOT/JIT functions from the
host embedder, which expects good performance for the calling process, while
in the current implementation, runtime calls the wasm_runtime_invoke_native
to prepare the array of registers and stacks for the invokeNative assemble code,
and the latter then puts the elements in the array to physical registers and
native stacks and calls the AOT/JIT function, there may be many data copying
and handlings which impacts the performance.

This PR registers some call wrappers for some simple wasm signatures, and let
runtime call the wrapper to directly invoke the AOT/JIT function instead of
calling wasm_runtime_invoke_native, which speedups the call process.

We may extend the mechanism next to allow the developer to register his call
wrappers to speedup the call process of calling the AOT/JIT functions for some
specific signatures.


core/iwasm/common/wasm_native.c
```diff
@@ -695,3 +714,944 @@ wasm_native_destroy()
...
+        else if (types[i] == VALUE_TYPE_I64)
+            signature[j++] = 'I';
+        else
+            return NULL;
+    }
+
+    low = 0;
+    high = (int32)sizeof(invoke_quicks) / sizeof(InvokeQuick) - 1;
+
+    while (low <= high) {
```
lum1n0us, line 1643: maybe `bsearch()` in "stdlib.h"?

wenyongh: Yes, it is better, in fact we can use qsort and bsearch for the above quick_sort_symbols in register_natives and lookup_symbol. A concern is that are they supported in all the platforms?

PR #2979, opened 2024-01-04
freertos: not support clock_nanosleep() so change config

tkernelcn:
for wasi  sleep() like api,  on freertos platform only can use nanosleep() api, so change the configuration file.


wenyongh (commented):
LGTM

PR #2990, opened 2024-01-09
Add ref flags for function params/locals to AOT module instead of committing them in GC AOT code

TianlongLiang:


core/iwasm/aot/aot_runtime.h
```diff
@@ -183,6 +190,11 @@ typedef struct AOTModule {
     uint32 *max_stack_cell_nums;
 #endif
 
+#if WASM_ENABLE_GC != 0
+    /* params + locals ref flags of (both import and AOTed) functions */
+    struct LocalRefFlag *func_local_ref_flags;
```
wenyongh, line 195: Remove `struct`

core/iwasm/aot/aot_loader.c
```diff
@@ -2574,6 +2574,41 @@ load_function_section(const uint8 *buf, const uint8 *buf_end, AOTModule *module,
 #endif
     }
 
+#if WASM_ENABLE_GC != 0
+    size = sizeof(LocalRefFlag)
```
wenyongh, line 2578: Had better add comment for this code piece.

core/iwasm/aot/aot_runtime.h
```diff
@@ -362,7 +374,9 @@ typedef struct AOTFrame {
      *  local area: parameters and local variables
      *  stack area: wasm operand stack
      *  frame ref flags (GC only):
-     *      whether each cell in local and stack area is a GC obj
+     *      whether each cell in local(LLVM JIT) and stack area is a GC obj
+     *                           paddings(AOT)
+     *                           actual ref flags for local are in AOT module
```
wenyongh, line 379: Here AOTFrame is only for AOT, JIT's frame is WASMInterpFrame.
How about below like:
```C
     *  frame ref flags (GC only):
     *      whether each cell in local and stack area is a GC obj
     *      currently local's ref flags are stored in AOTModule,
     *      here we only reserve the padding bytes 
```

core/iwasm/compilation/aot_compiler.c
```diff
@@ -365,6 +366,9 @@ aot_gen_commit_values(AOTCompFrame *frame)
                             (p + 2)->committed_ref = (p + 3)->committed_ref =
                                 p->ref + 1;
                     }
+                    p++;
+                    p++;
+                    p++;
```
wenyongh, line 371: How about `p += 3`

core/iwasm/compilation/aot_compiler.c
```diff
@@ -318,11 +318,11 @@ aot_gen_commit_values(AOTCompFrame *frame)
     LLVMValueRef value;
     uint32 n;
 
-    for (p = frame->lp; p < frame->sp; p++) {
+    /* First, commit reference flags, ignore local(params + locals) ref flags */
+    for (p = frame->lp + frame->max_local_cell_num; p < frame->sp; p++) {
```
wenyongh, line 322: Should not ignore local ref flags for JIT? Not sure why JIT's CI runs good.
For JIT, should be `p = frame->lp`.

xujuntwt95329 (commented):
LGTM

PR #2992, opened 2024-01-09
enable MAP_32BIT for macOS

yamt:


build-scripts/config_common.cmake
```diff
@@ -451,3 +451,7 @@ if (WAMR_BUILD_LINUX_PERF EQUAL 1)
   add_definitions (-DWASM_ENABLE_LINUX_PERF=1)
   message ("     Enable linux perf support")
 endif ()
+
+if (APPLE)
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-pagezero_size,0x4000")
```
wenyongh, line 456: Is this to make mmap be able to return memory with address under 4GB?

yamt: yes. i added a comment.

wenyongh (approved):

PR #2994, opened 2024-01-10
Fixed the link failure when the sequence of native imports and binary imports are different.

JakeOrange:
When the sequence of native imports and binary imports are different, it can't find right native method. Then will have a error tips "out of bounds memory access".

### DIFF
--- a/core/iwasm/common/wasm_c_api.c
+++ b/core/iwasm/common/wasm_c_api.c
@@ -4886,22 +4886,22 @@ wasm_instance_new_with_args(wasm_store_t *store, const wasm_module_t *module,
         func_host = wasm_extern_as_func(in);
         /* it is a placeholder and let's skip it*/
         if (!func_host->type) {
-            func_import++;
             continue;
         }
+        if (func_host->func_idx_rt < 0)
+            continue;
 
-        func_import->with_env_arg = func_host->with_env;
+        CApiFuncImport *linkFunc = func_import + func_host->func_idx_rt;
+        linkFunc->with_env_arg = func_host->with_env;
         if (func_host->with_env) {
-            func_import->func_ptr_linked = func_host->u.cb_env.cb;
-            func_import->env_arg = func_host->u.cb_env.env;
+            linkFunc->func_ptr_linked = func_host->u.cb_env.cb;
+            linkFunc->env_arg = func_host->u.cb_env.env;
         }
         else {
-            func_import->func_ptr_linked = func_host->u.cb;
-            func_import->env_arg = NULL;
+            linkFunc->func_ptr_linked = func_host->u.cb;
+            linkFunc->env_arg = NULL;
         }
-        bh_assert(func_import->func_ptr_linked);
-
-        func_import++;
+        bh_assert(linkFunc->func_ptr_linked);
     }
 
     /* fill with inst */


core/iwasm/common/wasm_c_api.c
```diff
@@ -4886,22 +4886,22 @@ wasm_instance_new_with_args(wasm_store_t *store, const wasm_module_t *module,
...
         /* it is a placeholder and let's skip it*/
         if (!func_host->type) {
-            func_import++;
             continue;
         }
+        if (func_host->func_idx_rt < 0)
+            continue;
 
-        func_import->with_env_arg = func_host->with_env;
+        CApiFuncImport *linkFunc = func_import + func_host->func_idx_rt;
```
lum1n0us, line 4894: the problem is `funx_idx_rt` is not predictable when people work with host functions. This is also why `wasm_func_new()` assigns `-1` to this field.

JakeOrange:
> The point is `func_idx_rt` can't be operated from host side, which means if developers create a `wasm_func_t` with `wasm_func_new` and `func_idx_rt` will be -1 and they can't assign the right value. Plus, developers can't predicate positions of imports.
> 
> for the problem in title, you might want to see https://github.com/bytecodealliance/wasm-micro-runtime/tree/main/samples/wasm-c-api-imports

I understand your replay.

yamt:
i don't understand what this PR is trying to solve.
maybe you can add a small sample to `samples/` demonstrate the issue.

lum1n0us:
IIUC, @JakeOrange is facing a similar problem in *samples/wasi-c-api-imports*.

JakeOrange:
In the case " WebAssembly.instantiate(fetch(wasmBinaryFile), info)", info is a two-dimensional array.  Before calling the "wasm_instance_new_with_arg", the imported elements were exported and ware stored as a map. The cause is that there is a wrong bind in fill in module_inst->e->c_api_func_imports of wasm_instance_new_with_args.
func_ptr_linked and env_arg are binded wrongly.

lum1n0us:
Again, we thought the big problem in this line is

``` c
CApiFuncImport *linkFunc = func_import + func_host->func_idx_rt;
```

a host function should not operate, or predicate, the value of `func_idx_rt`. Maybe in one case, with a fixed .wasm, it will work. But in a generic way, we still suggest follow *samples/wasm-c-api-imports*.

JakeOrange:
> Again, we thought the big problem in this line is
> 
> ```c
> CApiFuncImport *linkFunc = func_import + func_host->func_idx_rt;
> ```
> 
> a host function should not operate, or predicate, the value of `func_idx_rt`. Maybe in one case, with a fixed .wasm, it will work. But in a generic way, we still suggest follow _samples/wasm-c-api-imports_.

I had implemented it according to sample.
If func_idx_rt is -1, do_link will fail. I can't understand which scenario will happen？

yamt:
> In the case " WebAssembly.instantiate(fetch(wasmBinaryFile), info)", info is a two-dimensional array. Before calling the "wasm_instance_new_with_arg", the imported elements were exported and ware stored as a map. The cause is that there is a wrong bind in fill in module_inst->e->c_api_func_imports of wasm_instance_new_with_args. func_ptr_linked and env_arg are binded wrongly. I had change the map to vector for storing the impoted elements. It ensure the corrent order and bind rightly. The result has become right. Thank you!

is there `WebAssembly.instantiate` implementation backed by WAMR?

lum1n0us:
@JakeOrange 
> If func_idx_rt is -1, do_link will fail

Usually, a developer creates a host function with `wasm_func_new()`, and then pass it to `wasm_instance_new()` as a part of `wasm_extern_vec_t *imports`. Of course, he/she has to make sure the right order of host functions in `imports`. 

`wasm_func_new()` will assign -1 to `func_idx_rt` because in that phase, there is no such knowledge about a. will the host function be used to match an import function. b. which import function will be satisfied by this host function.

Later, in `wasm_instance_new()`, if the function type fits, runtime will assign the right value to `func_idx_rt`. So don't be worried about `func_idx_rt` from the host side. 

@yamt 
> is there WebAssembly.instantiate implementation backed by WAMR

Nay. I guess it is a [JS API](https://developer.mozilla.org/en-US/docs/WebAssembly/JavaScript_interface/Instance).

JakeOrange:
> So don't be worried about `func_idx_rt` from the host side.
Yes. It is my answer on the 3th floor.

PR #2995, opened 2024-01-10
Fix compilation error on iOS due to macOS-specific APIs

panicfrog:
This commit addresses a compilation issue on the iOS platform within the wasm-micro-runtime project-mini. The problem was caused by the usage of certain APIs that are exclusive to macOS. The code has been updated to include platform-specific preprocessor directives, ensuring that these macOS-specific APIs are only called when compiling for macOS. This change preserves the functionality on macOS while allowing successful compilation on iOS.
截屏2024-01-10 15 47 54

### DIFF
--- a/core/shared/platform/common/posix/posix_memmap.c
+++ b/core/shared/platform/common/posix/posix_memmap.c
@@ -6,7 +6,7 @@
 #include "platform_api_vmcore.h"
 
 #if (defined(__APPLE__) || defined(__MACH__)) && defined(__arm64__) \
-    && TARGET_OS_OSX
+    && defined(TARGET_OS_OSX) && TARGET_OS_OSX != 0
 #include <libkern/OSCacheControl.h>
 #endif
 
@@ -42,7 +42,7 @@ os_mmap(void *hint, size_t size, int prot, int flags, os_file_handle file)
 {
     int map_prot = PROT_NONE;
 #if (defined(__APPLE__) || defined(__MACH__)) && defined(__arm64__) \
-    && TARGET_OS_OSX
+    && defined(TARGET_OS_OSX) && TARGET_OS_OSX != 0
     int map_flags = MAP_ANONYMOUS | MAP_PRIVATE | MAP_JIT;
 #else
     int map_flags = MAP_ANONYMOUS | MAP_PRIVATE;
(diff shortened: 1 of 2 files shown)


core/shared/platform/common/posix/posix_memmap.c
```diff
@@ -5,7 +5,8 @@
 
 #include "platform_api_vmcore.h"
 
-#if (defined(__APPLE__) || defined(__MACH__)) && defined(__arm64__)
+#if (defined(__APPLE__) || defined(__MACH__)) && defined(__arm64__) \
+    && TARGET_OS_OSX
```
wenyongh, line 9: Using TARGET_OS_OSX may cause compilation error in some platforms, do you mean `defined(TARGET_OS_OSX)`?

panicfrog: iOS and MacOS both defined TARGET_OS_OSX. 
MacOS:
#define TARGET_OS_OSX               1
iOS:
#define TARGET_OS_OSX               0

panicfrog: "Defined (_ APPLE__) | | defined (_ MACH__) " has already indicated that it is the Apple platform, and all Apple platforms have defined TARGET_OS_OSX

wenyongh: So seems it had better be `&& defined(TARGET_OS_OSX) && TARGET_OS_OSX != 0`?

panicfrog: it looks better now！！！

wenyongh: > "defined (__ APPLE__) | | defined (__ MACH__) " has already indicated that it is the Apple platform, and all Apple platforms have defined TARGET_OS_OSX

OK, if that, `TARGET_OS_OSX != 0` should be good.

Another question is should we apply similar change to sys_icache_invalidate? Is this API also only defined in macOS?

panicfrog: Sorry, I made a mistake. This method is defined on both iOS and macOS and has no restrictions on ARM architecture. 
It would be better to change it to the following：
```
#if (defined(__APPLE__) || defined(__MACH__))
#include <libkern/OSCacheControl.h>
#endif
```
```
f (defined(__APPLE__) || defined(__MACH__))
    sys_icache_invalidate(start, len);
#endif
```

wenyongh: Thanks, sounds good.

wenyongh (commented):
LGTM

panicfrog:
> LGTM

Thanks for reviewing

PR #2998, opened 2024-01-11
CI: test wasm-c-api-imports sample where we test wasm-c-api

yamt:


lum1n0us:
@yamt Let's keep going and add the test wasm-c-api-imports sample into the CI by completing this PR ?

PR #3007, opened 2024-01-12
Fix build errors with new release

Zzzabiyaka:
Hi!

When our team started updating to wamr-1.3.1 release we encountered some errors during the build (for example, not initialised non-trivial class member `__paddings`.

This PR addresses those errors. 

Please let me know if default initializer for `__paddings` should look differently

### DIFF
--- a/core/iwasm/include/wasm_c_api.h
+++ b/core/iwasm/include/wasm_c_api.h
@@ -22,12 +22,12 @@
 #endif
 
 #if defined(__GNUC__) || defined(__clang__)
-#define DEPRECATED __attribute__((deprecated))
+#define WASM_API_DEPRECATED __attribute__((deprecated))
 #elif defined(_MSC_VER)
-#define DEPRECATED __declspec(deprecated)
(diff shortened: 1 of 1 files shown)


loganek:
> Please let me know if default initializer for __paddings should look differently

This field is just for padding and it shouldn't be used for anything else, so any value assigned to it should be ok.

LGTM

loganek (approved):

yamt:
what compiler are you using?
what errors did you get?
is the DEPRECATED changes related to __paddings things? how?

Zzzabiyaka:
`_padding` and DEPRECATED are 2 different issues, generally we should avoid using defines with generic names (such as `DEPRECATED`)

Zzzabiyaka:
`_paddings` issue looked smth like 

```
 error: missing initializer for member ‘wasm_val_t::__paddings’ [-Werror=missing-field-initializers]
   64 |     wasm_val_t res = {WASM_INIT_VAL};
```

PR #3008, opened 2024-01-12
Do not clone stack frames if there's no trap

eloparco:
When running the [no_pthread sample](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/samples/wasi-threads/wasm-apps/no_pthread.c), the assert was failing on `src->num_elems != 0`.
That's because, if the exception is `proc_exit`, there's no trap (the execution didn't fail, no stack frames). 

The problem was not detected in CI because there we build in release mode: https://github.com/bytecodealliance/wasm-micro-runtime/blob/08442458b12244f2cacb2f49573bb84e9433071f/.github/workflows/compilation_on_macos.yml#L328 

Should we use debug instead so that we can take advantage of asserts?

### DIFF
--- a/core/iwasm/common/wasm_c_api.c
+++ b/core/iwasm/common/wasm_c_api.c
@@ -1924,11 +1924,13 @@ wasm_frame_func_offset(const wasm_frame_t *frame)
 void
 wasm_frame_vec_clone_internal(Vector *src, Vector *out)
 {
-    bh_assert(src->num_elems != 0 && src->data);
-
-    bh_vector_destroy(out);
-    if (!bh_vector_init(out, src->num_elems, sizeof(WASMCApiFrame), false)) {
+    if (src->num_elems == 0) {
         bh_vector_destroy(out);
+        return;
+    }
+
+    if (!bh_vector_destroy(out)
+        || !bh_vector_init(out, src->num_elems, sizeof(WASMCApiFrame), false)) {
         return;
     }
 


wenyongh:
> When running the [no_pthread sample](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/samples/wasi-threads/wasm-apps/no_pthread.c), the assert was failing on `src->num_elems != 0`. That's because, if the exception is `proc_exit`, there's no trap (the execution didn't fail, no stack frames).
> 
> The problem was not detected in CI because there we build in release mode:
> 
> https://github.com/bytecodealliance/wasm-micro-runtime/blob/08442458b12244f2cacb2f49573bb84e9433071f/.github/workflows/compilation_on_macos.yml#L328
> 
> Should we use debug instead so that we can take advantage of asserts?

Yes, agree, it should be easier to expose issues, how about using `cmake --build . --config Debug --parallel 4` for some other samples?

wenyongh (commented):
LGTM

core/iwasm/common/wasm_c_api.c
```diff
@@ -1924,7 +1924,8 @@ wasm_frame_func_offset(const wasm_frame_t *frame)
 void
 wasm_frame_vec_clone_internal(Vector *src, Vector *out)
 {
-    bh_assert(src->num_elems != 0 && src->data);
+    if (src->num_elems == 0)
+        return;
```
wenyongh, line 1928: Had better to `bh_vector_destory(out)` before return to free its possible frames and zero-initialize out? For example, an instance's frames may be created previously and then `wasi_proc_exit` occurs.

BTW, in L1932, seems no need to destroy out again if bh_vector_init fails?

PR #3010, opened 2024-01-12
Increase default stack size

eloparco:
With the previous stack size, I can't even run the simple [no_pthread example](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/samples/wasi-threads/wasm-apps/no_pthread.c) on MacOS.

```bash
./iwasm wasm-apps/no_pthread.wasm
[3/3] Linking C executable iwasm

#00 main
#01 __main_void
#02 _start

Exception: native stack overflow
```


wenyongh (commented):
LGTM

yamt:
> With the previous stack size, I can't even run the simple [no_pthread example](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/samples/wasi-threads/wasm-apps/no_pthread.c) on MacOS.
> 
> ```shell
> ./iwasm wasm-apps/no_pthread.wasm
> [3/3] Linking C executable iwasm
> 
> #00 main
> #01 __main_void
> #02 _start
> 
> Exception: native stack overflow
> ```

it works for me on macOS 12.7.1 / x86-64.
```
spacetanuki% ./iwasm wasm-apps/no_pthread.wasm
New thread ID: 1, starting parameter: 52
Updated value: 60
Thread completed, new value: 60, thread id: 1
spacetanuki% 
```

yamt:
i guess it's simpler to allow APP_THREAD_STACK_SIZE_DEFAULT=0, which means "leave it to the system default". how do you think?

wenyongh:
> i guess it's simpler to allow APP_THREAD_STACK_SIZE_DEFAULT=0, which means "leave it to the system default". how do you think?

It should be good if we only use 0 for pthread_create, but there are some other places using it and APP_THREAD_STACK_SIZE_MIN, and it may be used in non-posix platforms:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/23d2e0627cefa0732b21ff6722e750125bb6a2aa/core/shared/platform/common/posix/posix_thread.c#L431-L435
https://github.com/bytecodealliance/wasm-micro-runtime/blob/23d2e0627cefa0732b21ff6722e750125bb6a2aa/core/shared/platform/zephyr/zephyr_thread.c#L331-L335
Seems it's better to set a non-zero value for it.

PR #2922, opened 2023-12-25
Refine wasm-c-api wasm_func_call

wenyongh:
Refine params_to_argv, argv_to_results, wasm_func_param_arity,
wasm_func_result_arity to fasten the wasm-c-api call wasm function process.

PR #3018, opened 2024-01-15
Remove deprecated tests of language-bindings python

wenyongh:
After PR #2922, caller of wasm-c-api wasm_func_call should ensure
the params and results are correctly created, runtime doesn't check
their format again.


lum1n0us:
let's just remove the code instead of...

wenyongh:
> let's just remove the code instead of...

Yes, updated.

lum1n0us (approved):
LGTM

PR #3019, opened 2024-01-15
Build samples in debug mode

eloparco:
Follow-up on #3008. Compiling samples in Debug mode allows us to take advantage of asserts and would have prevented the fix in #3008.


wenyongh (approved):
LGTM

PR #3025, opened 2024-01-15
Do not use pagezero size option if osx version >= 13

### LINKED ISSUE #3009: Unable to run iwasm on MacOS after #2992
After #2992, I'm not able to run `iwasm` on MacOS.

When I build, I get
```bash
ld: warning: non-standard -pagezero_size is deprecated when targeting macOS 13.0 or later
```

And if I run `iwasm` I get:
```bash
./iwasm
[1]    2986 killed     ./iwasm
```

If I try running with a debugger, I get
```bash
lldb -- ./iwasm                               
(lldb) target create "./iwasm"
Current executable set to '/Users/eloparco/dev/forks/wasm-micro-runtime/samples/wasi-threads/build/iwasm' (arm64).
(lldb) run
error: Malformed Mach-o file
```

If I comment this line out
https://github.com/bytecodealliance/wasm-micro-runtime/blob/08442458b12244f2cacb2f49573bb84e9433071f/build-scripts/config_common.cmake#L469
everything is fine.

eloparco:
Fixes #3009

### DIFF
--- a/build-scripts/config_common.cmake
+++ b/build-scripts/config_common.cmake
@@ -464,7 +464,13 @@ else ()
 endif ()
 
 if (APPLE)
+  EXEC_PROGRAM(sw_vers ARGS -productVersion OUTPUT_VARIABLE OS_X_VERSION)
+  STRING(REGEX MATCH "([0-9]+)" OS_X_VERSION ${OS_X_VERSION})
+  if (OS_X_VERSION LESS 13)
+    set(PAGE_ZERO_SIZE_OPTION "-Wl,-pagezero_size,0x4000")
+  endif ()
+
   # On recent macOS versions, by default, the size of page zero is 4GB.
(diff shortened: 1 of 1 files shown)


build-scripts/config_common.cmake
```diff
@@ -464,7 +464,13 @@ else ()
 endif ()
 
 if (APPLE)
+  EXEC_PROGRAM(sw_vers ARGS -productVersion OUTPUT_VARIABLE OS_X_VERSION)
```
yamt, line 467: i guess we should check cmake target instead.

eloparco: you mean like here? https://github.com/neovim/neovim/blob/3fef24aa01ab2223d5455e7b22d218ea91495c86/cmake.deps/CMakeLists.txt#L61-L66

PR #3029, opened 2024-01-15
Allow using mmap for shared linear memory if OS_ENABLE_HW_BOUND_CHECK is disabled

loganek:
… is disabled

For shared memory, the max memory size must be defined in advanced. Re-allocation for growing memory can't be used as it might change the base address, therefore for OS_ENABLE_HW_BOUND_CHECK enabled the memory is mmaped, and if the flag is disabled, the memory is allocated. This change introduces a flag that allows users to use mmap for reserving memory address space even if the OS_ENABLE_HW_BOUND_CHECK is disabled.


yamt:
> @loganek we are considering using os_mmap to allocate memory no matter whether OS_ENABLE_HW_BOUND_CHECK is enabled or not, since currently the code of memory instantiation/de-instantiation is very complex and not easy to maintain, and we may implement wasm64 feature later, which will make them more complex. os_mmap should have been implemented in all the platforms, and using it may greatly simplify the code, and for this PR, we don't need to introduce WAMR_ENABLE_SHARED_MEMORY_MMAP. How do you think?

how do you plan to implement the realloc case?

yamt:
> This change introduces a flag that allows users to use mmap for reserving memory address space even if the OS_ENABLE_HW_BOUND_CHECK is disabled.

why do you want to use mmap?

wenyongh:
> > @loganek we are considering using os_mmap to allocate memory no matter whether OS_ENABLE_HW_BOUND_CHECK is enabled or not, since currently the code of memory instantiation/de-instantiation is very complex and not easy to maintain, and we may implement wasm64 feature later, which will make them more complex. os_mmap should have been implemented in all the platforms, and using it may greatly simplify the code, and for this PR, we don't need to introduce WAMR_ENABLE_SHARED_MEMORY_MMAP. How do you think?
> 
> how do you plan to implement the realloc case?

Do you mean realloc memory when memory.grow? Yes, it is a good question, per my understanding, for shared memory, no need to mmap a new memory, just increase the current page count. For non-shared memory, runtime should mmap a new memory, duplicate memory data to it and update the related info, like:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/bb053e3a2de24daaabdbbea7dfa7a3dd1ab8cf87/core/iwasm/common/wasm_memory.c#L705-L716

yamt:
> > > @loganek we are considering using os_mmap to allocate memory no matter whether OS_ENABLE_HW_BOUND_CHECK is enabled or not, since currently the code of memory instantiation/de-instantiation is very complex and not easy to maintain, and we may implement wasm64 feature later, which will make them more complex. os_mmap should have been implemented in all the platforms, and using it may greatly simplify the code, and for this PR, we don't need to introduce WAMR_ENABLE_SHARED_MEMORY_MMAP. How do you think?
> > 
> > 
> > how do you plan to implement the realloc case?
> 
> Do you mean realloc memory when memory.grow? Yes, it is a good question, per my understanding, for shared memory, no need to mmap a new memory, just increase the current page count. For non-shared memory, runtime should mmap a new memory, duplicate memory data to it and update the related info, like:
> 
> https://github.com/bytecodealliance/wasm-micro-runtime/blob/bb053e3a2de24daaabdbbea7dfa7a3dd1ab8cf87/core/iwasm/common/wasm_memory.c#L705-L716

realloc can be far cheaper than malloc+memcpy+free.
i'd suggest to introduce os_mremap, which a platform can use realloc to implement. (and maybe mremap on linux)

wenyongh:
> > > > @loganek we are considering using os_mmap to allocate memory no matter whether OS_ENABLE_HW_BOUND_CHECK is enabled or not, since currently the code of memory instantiation/de-instantiation is very complex and not easy to maintain, and we may implement wasm64 feature later, which will make them more complex. os_mmap should have been implemented in all the platforms, and using it may greatly simplify the code, and for this PR, we don't need to introduce WAMR_ENABLE_SHARED_MEMORY_MMAP. How do you think?
> > > 
> > > 
> > > how do you plan to implement the realloc case?
> > 
> > 
> > Do you mean realloc memory when memory.grow? Yes, it is a good question, per my understanding, for shared memory, no need to mmap a new memory, just increase the current page count. For non-shared memory, runtime should mmap a new memory, duplicate memory data to it and update the related info, like:
> > https://github.com/bytecodealliance/wasm-micro-runtime/blob/bb053e3a2de24daaabdbbea7dfa7a3dd1ab8cf87/core/iwasm/common/wasm_memory.c#L705-L716
> 
> realloc can be far cheaper than malloc+memcpy+free. i'd suggest to introduce os_mremap, which a platform can use realloc to implement. (and maybe mremap on linux)

Thanks a lot, it is good.

loganek:
@wenyongh, I think using mmap in all the cases sounds reasonable, and given that `grow` isn't that frequent operation, and it always allocates big chunk of memory, that shouldn't impact the performance either. Having said that, I'd suggest we merge this PR and continue the work on simplifications on top of that so our usecase is unblocked. What do you think?

wenyongh:
> @wenyongh, I think using mmap in all the cases sounds reasonable, and given that `grow` isn't that frequent operation, and it always allocates big chunk of memory, that shouldn't impact the performance either. Having said that, I'd suggest we merge this PR and continue the work on simplifications on top of that so our usecase is unblocked. What do you think?

@loganek ok, it is good to me, let's review the PR and merge it after it is carefully checked.

loganek:
Thanks, please review it then and let me know if you have any feedback. As a followup, I can work on migrating to mmap completely later this week or next week.

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -142,15 +145,21 @@ memories_deinstantiate(WASMModuleInstance *module_inst,
...
+                        wasm_munmap_linear_memory(memories[i]->memory_data,
+                                                  map_size, map_size);
+                    }
+                    else
 #endif
-                    os_munmap((uint8 *)memories[i]->memory_data,
-                              8 * (uint64)BH_GB);
+                        wasm_runtime_free(memories[i]->memory_data);
+#else
+                    map_size = memories[i]->num_bytes_per_page
```
wenyongh, line 159: Had better be `(uint64)memories[i]->num_bytes_per_page * memories[i]->cur_page_count` to avoid possible integer overflow?

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -398,15 +392,13 @@ memory_instantiate(WASMModuleInstance *module_inst, WASMModuleInstance *parent,
...
-    if (memory->memory_data)
-        wasm_runtime_free(memory->memory_data);
-#else
-#ifdef BH_PLATFORM_WINDOWS
-    os_mem_decommit(mapped_mem, memory_data_size);
-#endif
-    os_munmap(mapped_mem, map_size);
+#ifdef WASM_LINEAR_MEMORY_MMAP
+    if (mapped_mem)
+        wasm_munmap_linear_memory(mapped_mem, memory_data_size, map_size);
```
wenyongh, line 397: Same as above, should we restore UINT32 to 4G for memory_data_size?

core/iwasm/aot/aot_runtime.c
```diff
@@ -623,16 +617,13 @@ memory_instantiate(AOTModuleInstance *module_inst, AOTModuleInstance *parent,
...
-        wasm_runtime_free(memory_inst->memory_data);
-#else
-#ifdef BH_PLATFORM_WINDOWS
-    if (memory_inst->memory_data)
-        os_mem_decommit(p, memory_data_size);
-#endif
-    os_munmap(mapped_mem, map_size);
+#ifdef WASM_LINEAR_MEMORY_MMAP
+    if (mapped_mem)
+        wasm_munmap_linear_memory(mapped_mem, memory_data_size, map_size);
```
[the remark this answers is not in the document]
loganek: I've now moved uint32 truncation in `wasm_mmap_linear_memory()` before the `os_mem_commit` so we always commit uint32 (I don't think we need to commit more as wasm is 32bit now). So we should not need to restore the value now.

core/iwasm/interpreter/wasm_runtime.c
loganek: see my comment above.

loganek:
@wenyongh looks like the build failed for unrelated reason - would you be able to restart the build?

core/iwasm/aot/aot_runtime.c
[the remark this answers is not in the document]
wenyongh: Got it, thanks.

core/iwasm/common/wasm_runtime_common.h
```diff
@@ -1093,6 +1098,14 @@ wasm_runtime_quick_invoke_c_api_native(WASMModuleInstanceCommon *module_inst,
...
 wasm_runtime_show_app_heap_corrupted_prompt();
 
+void
+wasm_munmap_linear_memory(void *mapped_mem, uint64 commit_size,
+                          uint64 map_size);
+
+void *
+wasm_mmap_linear_memory(uint64_t map_size, uint64 *io_memory_data_size,
+                        char *error_buf, uint32 error_buf_size);
+
```
wenyongh, line 1108: Is it better to move `wasm_munmap_linear_memory`/`wasm_mmap_linear_memory` from wasm_runtime_common.h/c to wasm_memory.h/c? We normally implement memory related operations (except shared memory) in the latter.

wenyongh (commented):
LGTM with minor comment.

core/iwasm/common/wasm_runtime_common.h
loganek: Sure, done.

build-scripts/config_common.cmake
```diff
@@ -248,6 +248,12 @@ if (WAMR_BUILD_SHARED_MEMORY EQUAL 1)
 else ()
   add_definitions (-DWASM_ENABLE_SHARED_MEMORY=0)
 endif ()
+if (WAMR_ENABLE_SHARED_MEMORY_MMAP EQUAL 1)
+  add_definitions (-DWASM_ENABLE_SHARED_MEMORY_MMAP=1)
+  message ("     Shared memory allocated using mmap enabled")
+else ()
+  add_definitions (-DWASM_ENABLE_SHARED_MEMORY_MMAP=0)
```
g0djan, line 255: nit: message (" Shared memory allocated using mmap disabled")

easier to read logs

core/iwasm/aot/aot_runtime.c
```diff
@@ -362,14 +365,21 @@ memories_deinstantiate(AOTModuleInstance *module_inst)
...
+                               * memory_inst->max_page_count;
+                    wasm_munmap_linear_memory(memory_inst->memory_data,
+                                              map_size, map_size);
+                }
+                else
 #endif
-                os_munmap(memory_inst->memory_data, 8 * (uint64)BH_GB);
+                    wasm_runtime_free(memory_inst->memory_data);
+#else
+                map_size = (uint64)memory_inst->num_bytes_per_page
```
g0djan, line 379: is this map_size defined? it's out of `#ifdef WASM_LINEAR_MEMORY_MMAP` section and map_size defined at line 344 in such section

build-scripts/config_common.cmake
loganek: Good point, we'll remove this soon anyway so I'll leave it as it is.

PR #3031, opened 2024-01-16
Check passive mode in fast-jit table.init

wenyongh:


yamt (approved):
lgtm. i plan to drop non-passive elements on instantiation (and probably revert this) though.

PR #3033, opened 2024-01-16
Avoid unused thread_id warning and recompile multi-module sample

eloparco:
- Unused var warning (when building in release mode):
```bash
wasm-micro-runtime/samples/wasi-threads/wasm-apps/no_pthread.c:44:9: warning: variable 'thread_id' set but not used [-Wunused-but-set-variable]
    int thread_id;
        ^
1 warning generated.
```

- When using the multi-module example and changing the files in `wasm-apps`, they don't get recompiled when the sample is rebuilt. Adding `BUILD_ALWAYS` as it's done in other samples.


wenyongh (commented):
LGTM

PR #3036, opened 2024-01-16
Don't redefine D_INO if already defined

Zzzabiyaka:
on some closed-source platforms we need to  predefine this option to better reflect platform peculiarities


loganek (approved):
LGTM

PR #3039, opened 2024-01-17
Disable quick aot entry for interp and fast-jit

wenyongh:
Quick aot/jit entry is only for aot and llvm jit now, no need to enable it
for interpreter and fast-jit when aot and llvm jit are disabled.


build-scripts/config_common.cmake
```diff
@@ -457,18 +457,22 @@ if (WAMR_BUILD_LINUX_PERF EQUAL 1)
...
   message ("     Linux perf support enabled")
 endif ()
-if (NOT DEFINED WAMR_BUILD_QUICK_AOT_ENTRY)
-  # Enable quick aot/jit entries by default
-  set (WAMR_BUILD_QUICK_AOT_ENTRY 1)
-endif ()
-if (WAMR_BUILD_QUICK_AOT_ENTRY EQUAL 1)
-  add_definitions (-DWASM_ENABLE_QUICK_AOT_ENTRY=1)
-  message ("     Quick AOT/JIT entries enabled")
+if (WAMR_BUILD_AOT EQUAL 1 OR WAMR_BUILD_AOT EQUAL 1)
```
no1wudi, line 460: Why need to check the condition twice by or?

wenyongh: Should be WAMR_BUILD_JIT, fixed.

no1wudi (approved):

PR #3046, opened 2024-01-18
classic-interp: Handle SIMD opcode when JIT is enabled

wenyongh:
Though SIMD isn't supported by interpreter, when JIT is enabled,
developer may run `iwasm --interp <wasm_file>` to trigger the SIMD
opcode in interpreter, which isn't handled before this PR.


lum1n0us:
LGTM

PR #3048, opened 2024-01-18
Enhance aot feature flags emit and load

wenyongh:


core/iwasm/interpreter/wasm_loader.c
```diff
@@ -3095,13 +3133,15 @@ load_import_section(const uint8 *buf, const uint8 *buf_end, WASMModule *module,
...
 
-#if WASM_ENABLE_REF_TYPES == 0 && WASM_ENABLE_GC == 0
                     if (module->import_table_count > 1) {
+#if WASM_ENABLE_REF_TYPES == 0 && WASM_ENABLE_GC == 0
                         set_error_buf(error_buf, error_buf_size,
                                       "multiple tables");
                         return false;
-                    }
+#elif WASM_ENABLE_WAMR_COMPILER != 0
+                        module->is_ref_types_used = true;
```
xujuntwt95329, line 3142: Is it reasonable to indicate ref type usage according to import table count?

wenyongh: I think it should be reasonable, multiple tables are only allowed when reference types is enabled, which introduces table.init/table.fill/table.copy/table.get/table.set opcodes to operate on the specify table(s). As you can see above, when REF_TYPES and GC are disabled, "multiple tables" error is reported.
https://github.com/WebAssembly/reference-types/blob/master/proposals/reference-types/Overview.md#language-extensions
image

xujuntwt95329: Got it, thanks

Issue #3053, opened 2024-01-19
wasi pthread_mutex_lock issue


tkernelcn:
test code as below, wasi-sdk version : https://github.com/WebAssembly/wasi-libc/releases/tag/wasi-sdk-21

when .wasm running, a lot of warning infos,  is it a bug?  Thanks.
## infos:
```shell
[00:00:12:000 - 5676AB60]: HashMap find elem failed: map or key is NULL.                                                                                                                                                                                                                                                                                                                    
[00:00:12:000 - 5676AB60]: HashMap find elem failed: map or key is NULL.                                                                                                                                                                                                                                                                                                            
[00:00:12:000 - 5676AB60]: HashMap find elem failed: map or key is NULL.                                                                                                                                                                                                                                                                                                            
[00:00:12:000 - 5676AB60]: HashMap find elem failed: map or key is NULL.                                                                                                                                                                                                                                                                                                       
[00:00:12:000 - 5676AB60]: HashMap find elem failed: map or key is NULL.                                                                                                                                                                                                                                                                                                                
[00:00:12:000 - 5676AB60]: HashMap find elem failed: map or key is NULL.                                                                                                                                                                                                                                                                                                                  
[00:00:12:000 - 5676AB60]: HashMap find elem failed: map or key is NULL.                                                                                                                                                                                                                                                                                                                 
[00:00:12:000 - 5676AB60]: HashMap find elem failed: map or key is NULL.                                                                                                                                                                                                                                                                                                                   
[00:00:12:000 - 5676AB60]: HashMap find elem failed: map or key is NULL.                                                                                                                                                                                                                                                                                                                   
[00:00:12:000 - 5676AB60]: HashMap find elem failed: map or key is NULL.                                                                                                                                                                                                                                                                                                                   
[00:00:12:000 - 5676AB60]: HashMap find elem failed: map or key is NULL.                                                                                                                                                                                                                                                                                                                    
[00:00:12:000 - 5676AB60]: HashMap find elem failed: map or key is NULL.                                                                                                                                                                                                                                                                                                                    
[00:00:12:000 - 5676AB60]: HashMap find elem failed: map or key is NULL.                                                                                                                                                                                                                                                                                                                    
[00:00:12:000 - 5676AB60]: HashMap find elem failed: map or key is NULL.                                                                                                                                                                                                        
[00:00:12:000 - 5676AB60]: HashMap find elem failed: map or key is NULL.                                                                                                                      
```

## source
[...]

PR #3055, opened 2024-01-19
Check arguments before calling bh_hash_map_find

wenyongh:
Check whether the arguments are NULL before calling bh_hash_map_find,
or lots of "HashMap find elem failed: map or key is NULL" warnings may
be dumped. Reported in #3053.

### DIFF
--- a/core/iwasm/common/wasm_shared_memory.c
+++ b/core/iwasm/common/wasm_shared_memory.c
@@ -176,7 +176,7 @@ acquire_wait_info(void *address, AtomicWaitNode *wait_node)
     AtomicWaitInfo *wait_info = NULL;
     bh_list_status ret;
 
-    if (address)
+    if (wait_map && address) /* avoid passing NULL to bh_hash_map_find */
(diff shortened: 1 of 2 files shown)


core/iwasm/common/wasm_shared_memory.c
```diff
@@ -176,7 +176,7 @@ acquire_wait_info(void *address, AtomicWaitNode *wait_node)
     AtomicWaitInfo *wait_info = NULL;
     bh_list_status ret;
 
-    if (address)
+    if (wait_map && address) /* avoid passing NULL to bh_hash_map_find */
```
yamt, line 179: as far as i understand, "address" here can't be NULL.

wenyongh: Yes, you are right, here the address comes from opcode atomic.wait32, atomic.wait64 and atomic.notify, all of them operate on an address inside the linear memory, which should not be NULL. I changed to bh_assert instead.

yamt (approved):
lgtm

PR #3060, opened 2024-01-19
Add flag to control Winsocket initialization

loganek:
When WAMR is embedded to other application, the lifecycle of the socket might conflict with other usecases. E.g. if WAMR is deinitialized before any other use of sockets, the application goes into the invalid state. The new flag allows host application to take control over the socket initialization.

### DIFF
--- a/core/config.h
+++ b/core/config.h
@@ -231,6 +231,17 @@
 #define WASM_ENABLE_LOG 1
 #endif
 
+/* When this flag is set, WAMR will not automatically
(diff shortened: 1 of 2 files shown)


wenyongh (approved):

loganek:
@wenyongh does the change look good to you? If so, do you think we can merge it?

wenyongh:
> @wenyongh does the change look good to you? If so, do you think we can merge it?

Sure, I merged it.

Issue #3035, opened 2024-01-16
LLVM 18: R_X86_64_32S relocation is used for medium model (--size-level=1)


yamt:
with the recent version of LLVM, wamrc --size-level=1 often generates R_X86_64_32S relocations.
it's bad because we suggest --size-level=1 to avoid that kind of relocations.
https://github.com/bytecodealliance/wasm-micro-runtime/blob/bb053e3a2de24daaabdbbea7dfa7a3dd1ab8cf87/core/iwasm/aot/arch/aot_reloc_x86_64.c#L196-L201
it seems the relocation is for jump tables.

a workaround: disable jump tables
another workaround: use large model (but it's broken for other reasons: https://github.com/bytecodealliance/wasm-micro-runtime/issues/3034)

PR #3067, opened 2024-01-22
aot_reloc_x86_64: suggest to try --size-level=0 as well

yamt:
cf. https://github.com/bytecodealliance/wasm-micro-runtime/issues/3035


yamt:
the windows job failure looks like https://github.com/bytecodealliance/wasm-micro-runtime/issues/2776

the other job failure also looks unrelated to this PR.
```
CMake Error at googletest-subbuild/googletest-populate-prefix/src/googletest-populate-stamp/download-googletest-populate.cmake:170 (message):
  Each download failed!

    error: downloading 'https://github.com/google/googletest/archive/03597a01ee50ed33e9dfd640b249b4be3799d395.zip' failed
          status_code: 56
          status_string: "Failure when receiving data from the peer"
          log:
          --- LOG BEGIN ---
          timeout on name lookup is not supported
    Trying 192.30.255.112:443...
```

wenyongh (commented):
LGTM

Issue #3069, opened 2024-01-22
allocate_aux_stack/free_aux_stack locking issues


yamt:
currently, allocate_aux_stack and free_aux_stack are called with cluster->lock held.
however, depending on configurations, they might end up with calling functions exported by the wasm module.
if the wasm does something which acquires cluster->lock, it would deadlock.

PR #3073, opened 2024-01-23
thread-mgr: fix locking problems around aux stack allocation

yamt:
Fixes: https://github.com/bytecodealliance/wasm-micro-runtime/issues/3069


yamt:
the ci failure looks like https://github.com/bytecodealliance/wasm-micro-runtime/issues/2776

wenyongh (commented):
LGTM

PR #3075, opened 2024-01-23
Remove a lot of "unused parameter" warnings

loganek:
They might shadow some of the real issues, so better to keep the number of warnings as low as possible.

### DIFF
--- a/core/iwasm/libraries/libc-builtin/libc_builtin_wrapper.c
+++ b/core/iwasm/libraries/libc-builtin/libc_builtin_wrapper.c
@@ -473,12 +473,16 @@ snprintf_wrapper(wasm_exec_env_t exec_env, char *str, uint32 size,
 static int
 puts_wrapper(wasm_exec_env_t exec_env, const char *str)
 {
+    (void)exec_env;
+
     return os_printf("%s\n", str);
 }
 
 static int
 putchar_wrapper(wasm_exec_env_t exec_env, int c)
 {
+    (void)exec_env;
+
     os_printf("%c", c);
     return 1;
 }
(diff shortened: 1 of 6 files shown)


eloparco:
We should report those warnings as errors at some point, I was trying that here #2610 a while ago, but cmake flags where getting overwritten on Android for some reason

wenyongh (commented):
LGTM with minor comments

core/iwasm/libraries/libc-builtin/libc_builtin_wrapper.c
```diff
@@ -473,12 +473,16 @@ snprintf_wrapper(wasm_exec_env_t exec_env, char *str, uint32 size,
 static int
 puts_wrapper(wasm_exec_env_t exec_env, const char *str)
 {
+    (void)exec_env;
```
yamt, line 476: i feel it's cleaner to use attribute((unused)) to suppress the warning. how do you think?

wenyongh: It may be good, but I found that other project like zephyr also uses `(void)xxx`, and it define macro `ARG_UNUSED(x)`:
https://github.com/zephyrproject-rtos/zephyr/blob/b302308e5f666a960a84ff8d76cfbe531a932fdf/include/zephyr/toolchain/gcc.h#L263

Maybe we can also define a similar macro?

wenyongh:
@loganek, @yamt I merged this PR first as there are also some other places using `(void)xxx`. If we want to use `attribute((unused))` or define a new macro, we may upload another PR to achieve it.

loganek:
> @loganek, @yamt I merged this PR first as there are also some other places using `(void)xxx`. If we want to use `attribute((unused))` or define a new macro, we may upload another PR to achieve it.

Thanks, I don't have any strong opinion. If anybody wants to follow-up on that and open a PR to introduce a macro, I'd be happy to review it. I'd like the solution to be working across all supported compilers though, and `(void)xxx` does that now - we should not regress that.

PR #3080, opened 2024-01-24
wasi: apply wasm_runtime_begin_blocking_op to poll as well

yamt:
While we used a different approach for poll_oneoff [1], the implementation works only when the poll list includes an absolute clock event. That is, if we have a thread which is polling on descriptors without a timeout, we fail to terminate the thread.

This commit fixes it by applying wasm_runtime_begin_blocking_op to poll as well.

[1] https://github.com/bytecodealliance/wasm-micro-runtime/pull/1951

### DIFF
--- a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/blocking_op.c
+++ b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/blocking_op.c
@@ -7,6 +7,7 @@
 
 #include "ssp_config.h"
 #include "blocking_op.h"
+#include "libc_errno.h"
 
 __wasi_errno_t
 blocking_op_close(wasm_exec_env_t exec_env, os_file_handle handle,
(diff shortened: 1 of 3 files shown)


yamt:
the ci failure looks unrelated.
```
 Download action repository 'actions/checkout@v3' (SHA:f43a0e5ff2bd294095638e18286ca9a3d1956744)
Download action repository 'actions/cache@v3' (SHA:e12d46a63a90f2fae62d114769bbf2a179198b5c)
Warning: Failed to download action 'https://api.github.com/repos/actions/cache/tarball/e12d46a63a90f2fae62d114769bbf2a179198b5c'. Error: nodename nor servname provided, or not known (api.github.com:443)
Warning: Back off 20.183 seconds before retry.
Warning: Failed to download action 'https://api.github.com/repos/actions/cache/tarball/e12d46a63a90f2fae62d114769bbf2a179198b5c'. Error: nodename nor servname provided, or not known (api.github.com:443)
Warning: Back off 10.301 seconds before retry.
Error: nodename nor servname provided, or not known (api.github.com:443)
```

wenyongh (commented):
LGTM

Issue #2846, opened 2023-12-02
zephyr: unable to build simple example targeting the nucleo_f767zi board with wasi libc


deadprogram:
I am trying to build the `product-mini/platforms/zephyr/simple` example targeting the `nucleo_f767zi` board with `WAMR_BUILD_LIBC_WASI` so that a WASM module compiled for `wasi` can execute on the board itself.

This is the command I am trying to use to build:

```
ZEPHYR_BASE=~/zephyrproject/zephyr west build -b nucleo_f767zi . -p always -- -DWAMR_BUILD_TARGET=THUMBV7 -DWAMR_BUILD_LIBC_WASI=1
```
However, the code is unable to build without errors:

```
-- Zephyr version: 3.5.99 (/home/ron/zephyrproject/zephyr), build: zephyr-v3.5.0-2399-gc94cd30cecbf                                                                          [31/204] Building C object CMakeFiles/app.dir/home/ron/Development/wasm-micro-runtime/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/random.c.obj            
FAILED: CMakeFiles/app.dir/home/ron/Development/wasm-micro-runtime/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/random.c.obj                               
ccache /home/ron/zephyr-sdk-0.16.4/arm-zephyr-eabi/bin/arm-zephyr-eabi-gcc -DBH_FREE=wasm_runtime_free -DBH_MALLOC=wasm_runtime_malloc -DBH_PLATFORM_ZEPHYR -DBUILD_TARGET=\"
THUMBV7\" -DBUILD_TARGET_THUMB -DCORE_CM7 -DHSE_VALUE=8000000 -DKERNEL -DPICOLIBC_LONG_LONG_PRINTF_SCANF -DSTM32F767xx -DUSE_FULL_LL_DRIVER -DUSE_HAL_DRIVER -DWASM_DISABLE_H
W_BOUND_CHECK=0 -DWASM_DISABLE_STACK_HW_BOUND_CHECK=0 -DWASM_DISABLE_WAKEUP_BLOCKING_OP=0 -DWASM_ENABLE_AOT=1 -DWASM_ENABLE_BULK_MEMORY=1 -DWASM_ENABLE_FAST_INTERP=0 -DWASM_
[...]

PR #3086, opened 2024-01-25
zephyr: include libc utils when building with WASI support

hasheddan:
Follows up the change in
https://github.com/bytecodealliance/wasm-micro-runtime/pull/2585 to include libc utils when targeting zephyr, which abstracts posix filesystem functions.

Fixes https://github.com/bytecodealliance/wasm-micro-runtime/issues/2846 (there may be subsequent issues that need to be addressed regarding WASI support on Zephyr)

### DIFF
--- a/core/shared/platform/zephyr/shared_platform.cmake
+++ b/core/shared/platform/zephyr/shared_platform.cmake
@@ -12,5 +12,11 @@ include (${CMAKE_CURRENT_LIST_DIR}/../common/math/platform_api_math.cmake)
 
 file (GLOB_RECURSE source_all ${PLATFORM_SHARED_DIR}/*.c)
 
+if (WAMR_BUILD_LIBC_WASI EQUAL 1)
+  list(APPEND source_all ${PLATFORM_SHARED_DIR}/../common/posix/posix_file.c)
+  include (${CMAKE_CURRENT_LIST_DIR}/../common/libc-util/platform_common_libc_util.cmake)
+  set(source_all ${source_all} ${PLATFORM_COMMON_LIBC_UTIL_SOURCE})
+endif ()
+
 set (PLATFORM_SHARED_SOURCE ${source_all} ${PLATFORM_COMMON_MATH_SOURCE})
 


salva00:
I have tried to fix the issue manually by modifying the `shared_platform.cmake` but it persisted :(
PS: I'm building with ` west build . -b qemu_cortex_a53 -p always -- -DWAMR_BUILD_LIBC_WASI=1 -DWAMR_BUILD_LIB_WASI_THREADS=1 -DWAMR_BUILD_TARGET=AARCH64`

hasheddan:
@salva00 thanks for reporting! Could you share your build logs?

salva00:
Sure! This is my build log:
[build_log.txt](https://github.com/bytecodealliance/wasm-micro-runtime/files/14056075/build_log.txt)

hasheddan:
@salva00 thanks! It appears that this PR does solve the inclusion of `libc_errno.h`, but there are additional errors that follow it.

zoraaver:
Hi, I'm not really familiar with zephyr and whether it is POSIX compliant or not. If it is, it should be a matter of adding the missing headers to [`platform_internal.h`](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/shared/platform/zephyr/platform_internal.h) to fix these build errors. Otherwise, you would need to implement the [filesystem APIs](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/shared/platform/include/platform_api_extension.h#L1115-L1143) for zephyr.

lucasAbadFr:
Hello, any news on this issue ? 

Zephyr [File System API](https://docs.zephyrproject.org/latest/services/file_system/index.html) is not POSIX compliant, so I think it's a bad idea to include `posix_file.c`. Also Zephyr is a RTOS so it doesn't necesarily comes with a FS (I never tried to use one).

But Zephyr implement it's own socket stack, which will requiere `WAMR_BUILD_LIBC_WASI=1`.

If I understand correctly, the first issue lies in the `sandboxed-system-primitives` and because in Zephyr:
* struct `timespec` doesn't exist 
* struct `pollfd` doesn't exist 
* type `nfds_t` doesn't exist

I think that this issue is not limited to Zephyr but apply to most RTOS (freeRTOS, RIOT, ...)

Should we do like with windows and abstract these concepts (at the OS level) ?

@zoraaver What is your opinion on this ?

zoraaver:
Hi @lucasAbadFr,

> Zephyr [File System API](https://docs.zephyrproject.org/latest/services/file_system/index.html) is not POSIX compliant, so I think it's a bad idea to include posix_file.c. Also Zephyr is a RTOS so it doesn't necessarily comes with a FS (I never tried to use one).

That sounds fine to me, although it seems Zephyr does have a filesytem API -https://docs.zephyrproject.org/latest/services/file_system/index.html? If you just want socket support, you could just stub the filesystem functions (which it looks like you've done in https://github.com/bytecodealliance/wasm-micro-runtime/pull/3335). Either way, you probably don't want to include `posix_file.c` if Zephyr does not provide the actual POSIX file APIs.

> - struct timespec doesn't exist 
> - struct pollfd doesn't exist
> - type nfds_t doesn't exist

I think it's fine to abstract these types, similar to how we did for `os_file_handle`. These structures weren't an issue for Windows because we didn't need poll support for our use case so we just ifdeffed them out in `posix.c`. If you want poll support, I guess you'll have to abstract `poll` and implement it for Zephyr. Otherwise, you could alter the ifdef's in `posix.c` so that the code is not compiled for Zephyr (obviously you won't have poll support in Zephyr then).

The best option in either case depends on your needs i.e. whether you need filesystem support or poll support in Zephyr.

I hope that's helpful. Let me know if anything is unclear.

wenyongh:
Close this PR since similar changes had been applied in branch [dev/zephyr_file_socket](https://github.com/bytecodealliance/wasm-micro-runtime/tree/dev/zephyr_file_socket), see PR https://github.com/bytecodealliance/wasm-micro-runtime/pull/3633.

PR #3088, opened 2024-01-26
Add comments to suppress warning from clang-tidy

no1wudi:
Suppress style warnings for macro definition, name of these macros is inconsistent with others (upper case).


wenyongh (commented):
LGTM

PR #3097, opened 2024-01-29
Use logger for runtime prints

eloparco:
When reporting logs from the runtime, it's better to use `BH_LOG` (since it can be overridden with `WAMR_BH_LOG`) instead of `os_printf`. In this way, we can distinguish runtime output from wasm execution output.

From #3070:

> Also, I think there's a few places where os_printf is used in the runtime where actually we should bh_log
https://github.com/bytecodealliance/wasm-micro-runtime/pull/3070#issuecomment-1905856689

Occurrences of `os_printf` have been replaced, apart from a few files (e.g. the platform-specific ones in `core/shared/platforms/`).

### DIFF
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -270,7 +270,7 @@ decode_insn(uint8 *insn)
 
         /* Print current instruction */
         /*
-        os_printf("%012" PRIX64 "  ", runtime_address);
+        LOG_VERBOSE("%012" PRIX64 "  ", runtime_address);
         puts(buffer);
         */
 
@@ -376,7 +376,7 @@ runtime_exception_handler(EXCEPTION_POINTERS *exce_info)
 #endif
     }
 
-    os_printf("Unhandled exception thrown:  exception code: 0x%lx, "
+    LOG_ERROR("Unhandled exception thrown:  exception code: 0x%lx, "
               "exception address: %p, exception information: %p\n",
               ExceptionRecord->ExceptionCode, ExceptionRecord->ExceptionAddress,
               sig_addr);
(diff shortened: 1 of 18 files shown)


wenyongh:
@eloparco thanks for the PR, I am not sure whether it is a good practice to replace os_printf to LOG_VERBOSE, there are some concerns: (1) LOG_VERBOSE only outputs logs when `-v=5` is specified for iwasm, by default there is no output, (2) LOG_VERBOSE dumps extra info, e.g. timestamp and thread id, (3) developer may be confused which function to use, LOG_VERBOSE, LOG_WARNING or LOG_ERROR?

I think we had better collect more input from other developers and check what is the best choice.

core/iwasm/fast-jit/cg/x86-64/jit_codegen_x86_64.cpp
```diff
@@ -556,7 +556,7 @@ dump_native(char *data, uint32 length)
         &decoder, data + offset, length - offset, &instruction, operands,
         ZYDIS_MAX_OPERAND_COUNT_VISIBLE, ZYDIS_DFLAG_VISIBLE_OPERANDS_ONLY))) {
         /* Print current instruction pointer */
-        os_printf("%012" PRIX64 "  ", runtime_address);
+        LOG_VERBOSE("%012" PRIX64 "  ", runtime_address);
```
wenyongh, line 559: using LOG_VERBOSE won't dump the log when log level is smaller than 5.

core/iwasm/fast-jit/jit_dump.c
```diff
@@ -15,45 +15,45 @@ jit_dump_reg(JitCompContext *cc, JitReg reg)
 
     switch (kind) {
         case JIT_REG_KIND_VOID:
-            os_printf("VOID");
+            LOG_VERBOSE("VOID");
```
wenyongh, line 18: using LOG_VERBOSE won't dump the log when log level is smaller than 5. And for fast-jit dump log, the log is has format and is complex and long, I am afraid adding timestamp and thread id may make it a little noisy:
```
; Function: $f87
; Constant table sizes: VOID=0 I32=0 I64=1 F32=0 F64=0 V64=0 V128=0 V256=0
; Label number: 4
; Instruction number: 43
; Register numbers: VOID=0 I32=16 I64=40 F32=16 F64=16 V64=0 V128=0 V256=0
; Label annotations: basic_block begin_bcip end_bcip jitted_addr
; Instruction annotations:
; Register annotations: def_insn

L0:
    ; PREDS()
    ; BEGIN_BCIP=0x0000 END_BCIP=0x0000

7F7E6AA1A9C8  mov r8, [r15+0xCA0]
7F7E6AA1A9CF  mov r10, [r15+0xC98]
7F7E6AA1A9D6  mov r9, r8
7F7E6AA1A9D9  add r9, 0x240
7F7E6AA1A9E0  cmp r9, r10
7F7E6AA1A9E3  jnbe 0x00007F7E6AA1AAB5
7F7E6AA1A9E9  sub r9, 0x120
7F7E6AA1A9F0  mov [r15+0xCA0], r9
7F7E6AA1A9F7  mov r9, r8
7F7E6AA1A9FA  add r9, 0x5C
7F7E6AA1A9FE  mov [r8+0x30], r9
```

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -2442,7 +2442,7 @@ wasm_dump_perf_profiling(const WASMModuleInstance *module_inst)
     char *func_name;
     uint32 i;
 
-    os_printf("Performance profiler data:\n");
+    LOG_VERBOSE("Performance profiler data:\n");
```
wenyongh, line 2445: Same as dump performance profiling info in aot_runtime.c.

wenyongh (commented):
Suggest to only change the dump of error/failure to LOG_ERROR, and leave others unchanged first?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -376,7 +376,7 @@ runtime_exception_handler(EXCEPTION_POINTERS *exce_info)
 #endif
     }
 
-    os_printf("Unhandled exception thrown:  exception code: 0x%lx, "
+    LOG_ERROR("Unhandled exception thrown:  exception code: 0x%lx, "
```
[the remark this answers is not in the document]
eloparco: >Like the output of error/exception in wasm loader/instantiation/interpreter

Which one are you referring to?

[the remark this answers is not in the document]
wenyongh: I meant printing the error/exception in the product-mini main.c:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/product-mini/platforms/posix/main.c#L899
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/product-mini/platforms/posix/main.c#L970

They are printed with printf, I am not sure whether here we should also use printf to dump "Unhandled exception thrown: xxx" like them.

wenyongh (approved):
LGTM

PR #3102, opened 2024-01-31
Fix AOT compilation on MacOS

eloparco:
After #2995, AOT stopped working properly on my arm MacOS:
```bash
[10:31:34:240 - 206309B40]: /Users/eloparco/dev/forks/wasm-micro-runtime/core/iwasm/common/wasm_runtime_common.c, line 1270, WASM module load failed
AOT module load failed: mmap memory failed
```

That's because, without `#include <TargetConditionals.h>`, `TARGET_OS_OSX` is undefined (since it's definition is in that header file).

I wanted to setup a step in CI for MacOS to avoid things to break again in the future, but in CI AOT seems to work even without (I guess because there we use an Intel MacOS node, not ARM).

### DIFF
--- a/core/shared/platform/common/posix/posix_memmap.c
+++ b/core/shared/platform/common/posix/posix_memmap.c
@@ -7,6 +7,7 @@
 
 #if defined(__APPLE__) || defined(__MACH__)
 #include <libkern/OSCacheControl.h>
+#include <TargetConditionals.h>
 #endif
 
 #ifndef BH_ENABLE_TRACE_MMAP
--- a/core/shared/platform/common/posix/posix_thread.c
+++ b/core/shared/platform/common/posix/posix_thread.c
@@ -9,6 +9,10 @@
 #include "platform_api_vmcore.h"
 #include "platform_api_extension.h"
 
+#if defined(__APPLE__) || defined(__MACH__)
+#include <TargetConditionals.h>
+#endif
+
 typedef struct {
     thread_start_routine_t start;
     void *arg;


yamt:
> I wanted to setup a step in CI for MacOS to avoid things to break again in the future, but in CI AOT seems to work even without (I guess because there we use an Intel MacOS node, not ARM).

there seems to be a runner with M1.
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

eloparco:
> there seems to be a runner with M1.
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

`macos-14`? It seems to be in beta

yamt:
> > there seems to be a runner with M1.
> > https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories
> 
> `macos-14`? It seems to be in beta

sure.
i have no idea about "beta" implications because i haven't used it.

eloparco:
>i have no idea about "beta" implications because i haven't used it.

Me neither, it says: `Warning: Beta and Deprecated Images are provided "as-is", "with all faults" and "as available" and are excluded from the service level agreement and warranty.` I'll wait until it's stable and then I'll try to add it to CI.

wenyongh:
@eloparco, @yamt should I merge the PR? Or wait for new update?

yamt:
> @eloparco, @yamt should I merge the PR? Or wait for new update?

no reason to block.

wenyongh (approved):

eloparco:
I'll open a PR to add the M1 node when I have time

Issue #1884, opened 2023-01-11
Adding Support for Exception Handling


woodsmc:
This issue has been created to track the implementation of exception handling support. This is based on the:

1. [Exception proposal](https://github.com/WebAssembly/exception-handling/tree/main/proposals/exception-handling).
2. [Exception WASM Specification](https://webassembly.github.io/exception-handling/core/bikeshed/#control-instructions%E2%91%A0).

## Current Status
Exception handling is currently supported by Web browsers and the Emscripten compiler tool chain, but at the time of writing there isn't support for the exception handling in WAMR. Indeed exception handling in general appears to be missing from many non-browser based runtimes. 

### Current Focus
This work focuses on the addition of support for exception handling in the WAMR runtime. It excludes work on language run times (e.g. C++ RT), and compiler tool chains. It is assumed that this work has already been completed, see current status above and Emscripten's existing support.

### Source Control Guidance - Working Branch
Implementation will be done on a git branch, `dev/exce_handling`. When the code on this branch is stable and the necessary quality bar and tests have passed then, this branch will be peer reviewed and merged into the main branch.

### Feature Switch for Exception Handling

The feature will be developed with a switch, enabling it to be turned on / off. This feature switch is `WAMR_BUILD_EXCE_HANDLING` and can be enabled by specifying the following at the command line:

```bash
    cmake -DWAMR_BUILD_EXCE_HANDLING=1
```

This can be turned on / off within build-scripts or defined in `config_common.cmake`.

This means that the code base will use the `WASM_ENABLE_EXCE_HANDLING` C macro to control the feature in code, e.g.

```c
    #ifdef WASM_ENABLE_EXCE_HANDLING
    // Feature implementatation here
    // ... 
    // ...
    #endif // WASM_ENABLE_EXCE_HANDLING
```

### Approach
The proposal breaks down adding exception handling into three key steps, each step reflecting execution style used by WAMR. 

* **Step 1** : The initial focus will be on the fully interpreted mode, it will deliberatly exclude any JIT implementations. This allows us to ensure that the implementation is correct and behaves as expected.
* **Step 2** : Once Step 1 is completed, Step 2 will start. This is focused on the Fast JIT and LLVM JIT implementation. At the moment this is considered out of scope.
* **Step 3** : The final step focuses on the AOT implementation, ensuring that the behaviour and performance is a expected. At the moment this is considered out of scope.

At the time of writing focus is only on Step 1.

[...]

PR #3107, opened 2024-01-31
EH: fix validation of delegate opcode

yamt:
cf. https://github.com/bytecodealliance/wasm-micro-runtime/issues/1884#issuecomment-1914000294

### DIFF
--- a/core/iwasm/interpreter/wasm_loader.c
+++ b/core/iwasm/interpreter/wasm_loader.c
@@ -7129,6 +7129,40 @@ fail:
     return NULL;
 }
 
+#if WASM_ENABLE_EXCE_HANDLING != 0
+static BranchBlock *
+check_branch_block_for_delegate(WASMLoaderContext *loader_ctx, uint8 **p_buf,
+                                uint8 *buf_end, char *error_buf,
+                                uint32 error_buf_size)
+{
+    uint8 *p = *p_buf, *p_end = buf_end;
+    BranchBlock *frame_csp_tmp;
+    uint32 depth;
+
+    read_leb_uint32(p, p_end, depth);
+    /*
+     * Note: "delegate 0" means the surrounding block, not the
+     * try-delegate block itself.
+     *
+     * Note: the caller hasn't popped the try-delegate frame yet.
+     */
+    bh_assert(loader_ctx->csp_num > 0);
+    if (loader_ctx->csp_num - 1 <= depth) {
+        set_error_buf(error_buf, error_buf_size, "unknown delegate label");
+        goto fail;
+    }
+    frame_csp_tmp = loader_ctx->frame_csp - depth - 2;
+#if WASM_ENABLE_FAST_INTERP != 0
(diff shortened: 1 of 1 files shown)


core/iwasm/interpreter/wasm_loader.c
```diff
@@ -7128,6 +7128,30 @@ check_branch_block(WASMLoaderContext *loader_ctx, uint8 **p_buf, uint8 *buf_end,
...
+check_branch_block_for_delegate(WASMLoaderContext *loader_ctx, uint8 **p_buf,
+                                uint8 *buf_end, char *error_buf,
+                                uint32 error_buf_size)
+{
+    uint8 *p = *p_buf, *p_end = buf_end;
+    BranchBlock *frame_csp_tmp;
+    uint32 depth;
+
+    read_leb_uint32(p, p_end, depth);
+    CHECK_BR(depth);
```
wenyongh, line 7142: Not sure whether it should be `CHECK_BR(depth + 1)`? Seems the target block to check is `loader_ctx->frame_csp - depth - 2`.
In check_branch_block, the code is:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/3111a86cb341cee9eb0b3e7d8d1d43327846a627/core/iwasm/interpreter/wasm_loader.c#L7120-L7121

yamt: good point. i will check.

yamt: fixed.

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -7129,6 +7129,40 @@ check_branch_block(WASMLoaderContext *loader_ctx, uint8 **p_buf, uint8 *buf_end,
...
+
+    read_leb_uint32(p, p_end, depth);
+    /*
+     * Note: "delegate 0" means the surrounding block, not the
+     * try-delegate block itself.
+     *
+     * Note: the caller hasn't popped the try-delegate frame yet.
+     */
+    bh_assert(loader_ctx->csp_num > 0);
+    if (loader_ctx->csp_num - 1 <= depth) {
```
wenyongh, line 7150: Not sure whether the check is enough? The original code calls check_branch_block:
check_branch_block -> CHECK_BR -> wasm_loader_check_br
And in wasm_loader_check_br, there are more checks:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/3111a86cb341cee9eb0b3e7d8d1d43327846a627/core/iwasm/interpreter/wasm_loader.c#L7058-L7088

Is it better to use `CHECK_BR(depth + 1)`?

yamt: the most of checks in wasm_loader_check_br are not appropriate for delegate, where parameters are given via the exception, not stack.

wenyongh: OK, got it, thanks!

wenyongh (approved):

Issue #2921, opened 2023-12-24
Unexpected behavior in JIT mode


erxiaozhou:
# Build commands
I compile the code with commit id 40b430fd240ad45f5c7a7dd0239fa2425e7c294e.
Platform: Ubuntu 20.04
CPU: amd64 
compile:

```
export CC=/usr/lib/llvm-16/bin/clang
export CXX=/usr/lib/llvm-16/bin/clang++
cd product-mini/platforms/linux/;rm -rf build
cmake  -DWASM_ENABLE_AOT=1 -DWAMR_BUILD_JIT=1 -DWAMR_BUILD_FAST_JIT=0   -DCMAKE_INSTALL_PREFIX="xxx" -DWAMR_BUILD_REF_TYPES=1 -DCMAKE_BUILD_TYPE=Release  -DWASM_ENABLE_BULK_MEMORY=1 -DWAMR_BUILD_LIBC_WASI=0 -DWAMR_BUILD_LIBC_BUILTIN=1 -DWAMR_BUILD_SIMD=1 -Bbuild
```

execute:
```shell
iwasm --heap-size=0 --llvm-jit -f to_test <tc_name>
```
# Cases
[v128.load64_lane_2635_10_1703418883268537.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/13762543/v128.load64_lane_2635_10_1703418883268537.zip)

# Actual behavior:
No exception
```
<0x0000000000000000 0xffffffffffffffff>:v128
```

# Expected behavior:
An exception indicates that "0000523: error: unexpected opcode: 0xfd 0xff 0x1"

TianlongLiang:
Hi, the reason you get both the error message `fast jit compilation failed: Error: unsupported opcode` and the successful execution result  `<0x0000000000000000 0xffffffffffffffff>:v128` is you are using the Multi-tier JIT enabled `iwasm`. By enabling the flag `DWAMR_BUILD_FAST_JIT` after your initial CMake cache variable setup, you would make both `DWAMR_BUILD_FAST_JIT=1` and `DWAMR_BUILD_JIT=1`, making the `iwasm` Multi-tier JIT enabled. 

[...]

PR #3115, opened 2024-02-01
Fix read and validation of misc/simd/atomic sub opcodes

wenyongh:
The format of sub opcodes after misc, simd and atomic prefix is leb uint32.

The issue was found in #2921.


TianlongLiang:
LGTM

PR #3117, opened 2024-02-02
add wasm_runtime_get_cur_local_obj_ref API support and change wasm runtime APIs name.

LevelCA:


core/iwasm/common/gc/gc_common.c
```diff
@@ -892,6 +892,15 @@ wasm_runtime_pop_local_object_refs(WASMExecEnv *exec_env, uint32 n)
     } while (--n > 0);
 }
 
+WASMLocalObjectRef *
+wasm_runtime_get_current_local_object_ref(WASMExecEnv *exec_env)
```
xujuntwt95329, line 896: The `current` in function name seems unnecessary. How about rename it as `wasm_runtime_get_local_object_ref`?

xujuntwt95329: And do you want to get non-first ref?
e.g.

``` C
wasm_runtime_get_local_object_ref(exec_env, 0);
wasm_runtime_get_local_object_ref(exec_env, 1);
wasm_runtime_get_local_object_ref(exec_env, 2);
...
```

LevelCA: `wasm_runtime_get_local_object_ref`  the name is ok. I just need to get the current ref, as I continue to push into ref, when I need to pop, I can continue to move the pointer until it is consistent with the current pointer I obtained before, which means that I have popped all the obj that I pushed before.

LevelCA: @xujuntwt95329

wenyongh: How about `wasm_runtime_cur_local_object_ref`?

LevelCA: `wasm_runtime_get_cur_local_object_ref ` is better, what do you think?

wenyongh: It is OK to me. BTW, should we shorten the function names, like:
```C
wasm_runtime_push_local_obj_ref
wasm_runtime_pop_local_obj_ref
wasm_runtime_pop_local_obj_refs
wasm_runtime_get_cur_local_obj_ref
```
@xujuntwt95329

xujuntwt95329: > It is OK to me. BTW, should we shorten the function names, like:
> 
> ```c
> wasm_runtime_push_local_obj_ref
> wasm_runtime_pop_local_obj_ref
> wasm_runtime_pop_local_obj_refs
> wasm_runtime_get_cur_local_obj_ref
> ```
> 
> @xujuntwt95329

These names seems better

LevelCA: OK, thanks, done

wenyongh: @LevelCA should also change wasm_runtime_push_local_object_ref/wasm_runtime_pop_local_object_ref/wasm_runtime_pop_local_object_refs to wasm_runtime_push_local_obj_ref/wasm_runtime_pop_local_obj_ref/wasm_runtime_pop_local_obj_refs, and the related codes in core/iwasm?

LevelCA: @wenyongh e. it's my fault.  All have been modified. thks

core/iwasm/include/gc_export.h
```diff
@@ -910,7 +910,18 @@ wasm_runtime_pop_local_object_ref(wasm_exec_env_t exec_env);
  * @param n number to pop
  */
 WASM_RUNTIME_API_EXTERN void
-wasm_runtime_pop_local_object_refs(wasm_exec_env_t exec_env, uint32_t n);
+wasm_runtime_pop_local_obj_refs(wasm_exec_env_t exec_env, uint32_t n);
+
+/**
+ * get current local object ref from stack
```
wenyongh, line 916: Could you change to `Get` as it is upper case in other places?

wenyongh (commented):
LGTM with minor comment

core/iwasm/include/gc_export.h
LevelCA: aha, OK,  done!

xujuntwt95329 (commented):
LGTM

PR #3120, opened 2024-02-02
Add ARM MacOS to CI

eloparco:
Add ARM MacOS to CI, to prevent regressions like the one fixed in https://github.com/bytecodealliance/wasm-micro-runtime/pull/3102 in the future.

I added the AOT compilation only to the `wasi-threads` example only because it was the easiest one to apply the change to.
I tried it without the change in #3102 (https://github.com/eloparco/wasm-micro-runtime/actions/runs/7755056858/job/21149684750) to make sure the problem would have been detected (and it was indeed).


wenyongh (commented):
LGTM

Issue #2231, opened 2023-05-23
aot/jit: tail-call implementation looks wrong


yamt:
* `return_call_indirect` doesn't seem to use a tail call at all.
* `return_call` uses llvm `tail` attribute, which is merely a hint. the recent versions of llvm has the `musttail` attirubute, which can be used instead.

Note: llvm c api doesn't seem to have an api to use `musttail` yet.

wenyongh:
> * `return_call_indirect` doesn't seem to use a tail call at all.
> * `return_call` uses llvm `tail` attribute, which is merely a hint. the recent versions of llvm has the `musttail` attirubute, which can be used instead.
> 
> Note: llvm c api doesn't seem to have an api to use `musttail` yet.

Thanks for the reminding, we implemented tail-call long time ago, I remembered we tested the spec cases of that proposal for both interpreter and AOT. And yes, we should test again and try to add the testing to the CI.

For `return_call_indirect`, maybe it is enough to `call_indirect` first and then `return` the result:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/compilation/aot_compiler.c#L413-L416

For the `musttail` attribute, we may set it with llvm c++ API, we set the function attribute `disable-tail-call` when needed:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/compilation/aot_llvm_extra.cpp#L319

yamt:
> > * `return_call_indirect` doesn't seem to use a tail call at all.
> > * `return_call` uses llvm `tail` attribute, which is merely a hint. the recent versions of llvm has the `musttail` attirubute, which can be used instead.
> > 
> > Note: llvm c api doesn't seem to have an api to use `musttail` yet.
> 
> Thanks for the reminding, we implemented tail-call long time ago, I remembered we tested the spec cases of that proposal for both interpreter and AOT. And yes, we should test again and try to add the testing to the CI.

FYI, recently llvm fixed a codegen bug in wasm tail-call. that would make tail-call a bit more usable. (thus more chances to expose implementation issues i guess.)

> 
> For `return_call_indirect`, maybe it is enough to `call_indirect` first and then `return` the result: https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/compilation/aot_compiler.c#L413-L416

it isn't enough because it can consume unbounded amount of stack.

> 
[...]

PR #3125, opened 2024-02-04
Implement GC (Garbage Collection) feature for interpreter, AOT and LLVM-JIT

wenyongh:


core/iwasm/aot/aot_runtime.c
```diff
@@ -1188,8 +1626,14 @@ aot_instantiate(AOTModule *module, AOTModuleInstance *parent,
     if (stack_size == 0)
         stack_size = DEFAULT_WASM_STACK_SIZE;
 #if WASM_ENABLE_SPEC_TEST != 0
-    if (stack_size < 48 * 1024)
-        stack_size = 48 * 1024;
+#if WASM_ENABLE_TAIL_CALL == 0
+    if (stack_size < 128 * 1024)
+        stack_size = 128 * 1024;
```
yamt, line 1631: @wenyongh 
why have you bumped stack size here?

core/iwasm/aot/aot_runtime.c
```diff
@@ -1188,8 +1626,14 @@ aot_instantiate(AOTModule *module, AOTModuleInstance *parent,
...
+#else
+    /* Some tail-call cases require large operand stack */
```
yamt, line 1633: @wenyongh
which cases are they?
tail-call tests consuming much stack sounds like a real bug in our implementation, doesn't it?

core/iwasm/aot/aot_runtime.c
wenyongh: I can't remember it well now since it took a long time to develop the GC feature, seems that at first the tail call cases were disabled temporarily or not added to GC spec proposal, and running GC cases requires more native stack. But now the tail call is enabled when testing GC cases, maybe can restore it to 48KB now. I tested the value with some features and didn't find issue: `./test -s spec -t aot -b -P`, `./test -s spec -t aot -p -b -P`, `./test -s spec -t aot -M -b -P`, `./test -s spec -t aot -W -b -P`.

core/iwasm/aot/aot_runtime.c
wenyongh: The cases can be found: https://github.com/WebAssembly/gc/blob/master/test/core/return_call.wast#L105-L118.

Yes, it is really an issue, due to our limited knowledge, we didn't find a better way to improve it when implementing AOT tail call feature. You also mentioned the problem in https://github.com/bytecodealliance/wasm-micro-runtime/issues/2231, do you think we can resolve it with musttail attribute or is there other way to resolve it? Thanks.

yamt: ok. i commented my current thoughts in https://github.com/bytecodealliance/wasm-micro-runtime/issues/2231.

PR #3132, opened 2024-02-05
Zero the memory from os_mmap in NuttX

no1wudi:


core/shared/platform/nuttx/nuttx_platform.c
```diff
@@ -125,6 +125,10 @@ os_mmap(void *hint, size_t size, int prot, int flags, os_file_handle file)
     if (posix_memalign(&p, 32, size)) {
         return NULL;
     }
+
+    /* Zero the memory that required by os_mmap */
```
wenyongh, line 129: How about `Zero the memory which is required by os_mmap`?

core/shared/platform/nuttx/nuttx_platform.c
```diff
@@ -125,6 +125,10 @@ os_mmap(void *hint, size_t size, int prot, int flags, os_file_handle file)
...
+    memset(p, 0, size);
```
wenyongh, line 130: Should also zero the memory returned in L97 and L111, for example:
```C
        p = up_textheap_memalign(sizeof(void *), size);
        if (p) {
            /* Zero the memory which is required by os_mmap */
            memset(p, 0, size);
        }
        return p;
```
```C
        p = in_ibus_ext(i_addr) ? i_addr : d_addr;
        if (p) {
            /* Zero the memory which is required by os_mmap */
            memset(p, 0, size);
        }
        return p;
```

core/shared/platform/nuttx/nuttx_platform.c
```diff
@@ -161,7 +173,8 @@ os_dcache_flush()
 
 void
 os_icache_flush(void *start, size_t len)
-{}
+{
```
wenyongh, line 176: Should not change the code, or CI guideline check will fail.

wenyongh (commented):
LGTM

PR #3136, opened 2024-02-05
Classic and fast interpreter threads opcodes boundary check fix 

TianlongLiang:
Use CHECK_MEMORY_OVERFLOW instead of CHECK_BULK_MEMORY_OVERFLOW to avoid overflow in threads atomic opcodes boundary check.


xujuntwt95329 (commented):
LGTM

wenyongh (commented):
LGTM

PR #3141, opened 2024-02-06
aot: place precheck wrapper before the corresponding wrapped function

yamt:
This increases the chance to use "short" calls.

Assumptions:
- LLVM preserves the order of functions in a module
- The wrapper function are smaller than the wrapped functions
- The target CPU has "short" PC-relative variation of call/jmp instructions and they are preferrable over the "long" ones.

A motivation:
- To avoid some relocations for XIP, I want to use xtensa PC-relative call instructions, which can only reach ~512KB.


wenyongh (commented):
LGTM

PR #3142, opened 2024-02-06
Using indirect call in pre-checker function to avoid the relocation occurrence in XIP mode.

dongsheng28849455:
The stack profiler aot_func#xxx will call wrapped function of aot_func_internal#xxx using symbol reference,  in some platform, like xtensa, it’s translated into a native long call, which need resolve the indirect address by relocation, this will break the XIP feature which requires the eliminating of relocation.
The solution is to change the symbol reference into an indirect call through the lookup table, the code will be like this:

```
call_wrapped_func:                                ; preds = %stack_bound_check_block
  %func_addr1 = getelementptr inbounds ptr, ptr %func_ptrs_ptr, i32 75
  %func_tmp2 = load ptr, ptr %func_addr1, align 4
  tail call void %func_tmp2(ptr %exec_env)
  ret void
```

### DIFF
--- a/core/iwasm/aot/aot_loader.c
+++ b/core/iwasm/aot/aot_loader.c
@@ -2558,9 +2558,19 @@ load_function_section(const uint8 *buf, const uint8 *buf_end, AOTModule *module,
         }
     }
 
+#if defined(BUILD_TARGET_XTENSA)
+    /*
+     * For Xtensa XIP, real func_count is doubled, including aot_func and
+     * aot_func_internal, so need to divide func_count by 2 here.
+     */
+    if (module->is_indirect_mode)
+        size = sizeof(uint32) * (uint64)module->func_count / 2;
+    else
+#else
     size = sizeof(uint32) * (uint64)module->func_count;
+#endif
 
-    if (size > 0) {
+        if (size > 0) {
 #if WASM_ENABLE_AOT_STACK_FRAME != 0
         if (!(module->max_local_cell_nums =
                   loader_malloc(size, error_buf, error_buf_size))) {
--- a/core/iwasm/aot/aot_loader.c
+++ b/core/iwasm/aot/aot_loader.c
@@ -2500,15 +2500,26 @@ load_function_section(const uint8 *buf, const uint8 *buf_end, AOTModule *module,
     const uint8 *p = buf, *p_end = buf_end;
     uint32 i;
     uint64 size, text_offset;
+    uint32 func_count = module->func_count;
 
-    size = sizeof(void *) * (uint64)module->func_count;
+#if defined(BUILD_TARGET_XTENSA)
+    /*
+     * For Xtensa XIP, real func_count is doubled, including aot_func and
+     * aot_func_internal, so need to multipy func_count by 2 here.
+     */
+    if (module->is_indirect_mode) {
+        func_count *= 2;
+    }
+#endif
+
+    size = sizeof(void *) * (uint64)func_count;
     if (size > 0
         && !(module->func_ptrs =
                  loader_malloc(size, error_buf, error_buf_size))) {
         return false;
     }
 
-    for (i = 0; i < module->func_count; i++) {
+    for (i = 0; i < func_count; i++) {
         if (sizeof(void *) == 8) {
             read_uint64(p, p_end, text_offset);
         }
@@ -2543,14 +2554,14 @@ load_function_section(const uint8 *buf, const uint8 *buf_end, AOTModule *module,
         module->start_function = NULL;
     }
 
-    size = sizeof(uint32) * (uint64)module->func_count;
+    size = sizeof(uint32) * (uint64)func_count;
     if (size > 0
         && !(module->func_type_indexes =
                  loader_malloc(size, error_buf, error_buf_size))) {
         return false;
     }
 
-    for (i = 0; i < module->func_count; i++) {
+    for (i = 0; i < func_count; i++) {
         read_uint32(p, p_end, module->func_type_indexes[i]);
         if (module->func_type_indexes[i] >= module->type_count) {
             set_error_buf(error_buf, error_buf_size, "unknown type");
@@ -2558,19 +2569,9 @@ load_function_section(const uint8 *buf, const uint8 *buf_end, AOTModule *module,
         }
     }
 
-#if defined(BUILD_TARGET_XTENSA)
-    /*
-     * For Xtensa XIP, real func_count is doubled, including aot_func and
-     * aot_func_internal, so need to divide func_count by 2 here.
-     */
-    if (module->is_indirect_mode)
-        size = sizeof(uint32) * (uint64)module->func_count / 2;
-    else
-#else
     size = sizeof(uint32) * (uint64)module->func_count;
-#endif
 
-        if (size > 0) {
+    if (size > 0) {
 #if WASM_ENABLE_AOT_STACK_FRAME != 0
         if (!(module->max_local_cell_nums =
                   loader_malloc(size, error_buf, error_buf_size))) {
--- a/core/iwasm/compilation/aot_emit_aot_file.c
+++ b/core/iwasm/compilation/aot_emit_aot_file.c
@@ -2423,10 +2423,7 @@ aot_emit_init_data_section(uint8 *buf, uint8 *buf_end, uint32 *p_offset,
         return false;
 
     offset = align_uint(offset, 4);
-    if (need_call_wrapped_indirect(obj_data))
-        EMIT_U32(comp_data->func_count * 2);
-    else
-        EMIT_U32(comp_data->func_count);
+    EMIT_U32(comp_data->func_count);
     EMIT_U32(comp_data->start_func_index);
 
     EMIT_U32(comp_data->aux_data_end_global_index);
--- a/core/iwasm/compilation/aot_emit_aot_file.c
+++ b/core/iwasm/compilation/aot_emit_aot_file.c
@@ -180,6 +180,16 @@ is_little_endian_binary(const AOTObjectData *obj_data)
 }
 
 static bool
+need_call_wrapped_indirect(const AOTObjectData *obj_data)
+{
+    const bool need_precheck = obj_data->comp_ctx->enable_stack_bound_check
+                               || obj_data->comp_ctx->enable_stack_estimation;
+
+    return obj_data->comp_ctx->is_indirect_mode && need_precheck
+           && !strncmp(obj_data->comp_ctx->target_arch, "xtensa", 6);
+}
+
+static bool
 str_starts_with(const char *str, const char *prefix)
 {
     size_t len_pre = strlen(prefix), len_str = strlen(str);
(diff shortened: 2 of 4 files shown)


yamt:
an alternative for xtensa is to use pc-relative calls. eg. https://github.com/yamt/llvm-project/commit/3f04244a4fd94952a3f208b7d1ee3afef2da1e3a
good: more efficient.
bad: a hack in llvm. i'm not sure how difficult a proper support would be.

wenyongh:
> an alternative for xtensa is to use pc-relative calls. eg. [yamt/llvm-project@3f04244](https://github.com/yamt/llvm-project/commit/3f04244a4fd94952a3f208b7d1ee3afef2da1e3a) good: more efficient. bad: a hack in llvm. i'm not sure how difficult a proper support would be.

Yes, patching llvm seems a good way instead of changing the AOT file format, @dongsheng28849455 could we apply the patch to llvm before we build llvm for aot compiler? We may add the patch file under `build-script` folder (like file `lldb_wasm.patch`), and change the build_llvm.py to apply the patch?

dongsheng28849455:
> > an alternative for xtensa is to use pc-relative calls. eg. [yamt/llvm-project@3f04244](https://github.com/yamt/llvm-project/commit/3f04244a4fd94952a3f208b7d1ee3afef2da1e3a) good: more efficient. bad: a hack in llvm. i'm not sure how difficult a proper support would be.
> 
> Yes, patching llvm seems a good way instead of changing the AOT file format, @dongsheng28849455 could we apply the patch to llvm before we build llvm for aot compiler? We may add the patch file under `build-script` folder (like file `lldb_wasm.patch`), and change the build_llvm.py to apply the patch?

we have tested with coremark, seems not much impact for performance.
This might be a more general solution, I'm not sure if any other target has similar issue? and we have to maintain the patch in future.

wenyongh:
> > 
> > Yes, patching llvm seems a good way instead of changing the AOT file format, @dongsheng28849455 could we apply the patch to llvm before we build llvm for aot compiler? We may add the patch file under `build-script` folder (like file `lldb_wasm.patch`), and change the build_llvm.py to apply the patch?
> 
> we have tested with coremark, seems not much impact for performance. This might be a more general solution, I'm not sure if any other target has similar issue? or we have to maintain the patch in future.

There isn't similar issue reported in other targets yet, seems that we only need to add it for xtensa. The patch is simple and it should be easy to maintain, normally we only need to update the patch when we upgrade the LLVM version. Otherwise, we have to maintain two different AOT file formats (and not sure whether we should change the aot loader), it is also a little tiresome.

dongsheng28849455:
> > > Yes, patching llvm seems a good way instead of changing the AOT file format, @dongsheng28849455 could we apply the patch to llvm before we build llvm for aot compiler? We may add the patch file under `build-script` folder (like file `lldb_wasm.patch`), and change the build_llvm.py to apply the patch?
> > 
> > 
> > we have tested with coremark, seems not much impact for performance. This might be a more general solution, I'm not sure if any other target has similar issue? or we have to maintain the patch in future.
> 
> There isn't similar issue reported in other targets yet, seems that we only need to add it for xtensa. The patch is simple and it should be easy to maintain, normally we only need to update the patch when we upgrade the LLVM version. Otherwise, we have to maintain two different AOT file formats (and not sure whether we should change the aot loader), it is also a little tiresome.

why we should maintain two AOT format? I think it is the same with normal XIP aot, and do not need change aot loader, just enlarge the the lookup table and follow the aot loading process as before.

yamt:
> we have tested with coremark, seems not much impact for performance.

can you share the numbers?

dongsheng28849455:
> > we have tested with coremark, seems not much impact for performance.
> 
> can you share the numbers?

test on esp32 device.
about 81.6, with --enable-multi-thread  option for wamrc 1.3.1 with this patch.
as a comparison, recedingly on wamrc1.1.2, the score is 88, but please note, which is without  --enable-multi-thread, the option will more or less reduce the execution speed.

wenyongh:
> > > > Yes, patching llvm seems a good way instead of changing the AOT file format, @dongsheng28849455 could we apply the patch to llvm before we build llvm for aot compiler? We may add the patch file under `build-script` folder (like file `lldb_wasm.patch`), and change the build_llvm.py to apply the patch?
> > > 
> > > 
> > > we have tested with coremark, seems not much impact for performance. This might be a more general solution, I'm not sure if any other target has similar issue? or we have to maintain the patch in future.
> > 
> > 
> > There isn't similar issue reported in other targets yet, seems that we only need to add it for xtensa. The patch is simple and it should be easy to maintain, normally we only need to update the patch when we upgrade the LLVM version. Otherwise, we have to maintain two different AOT file formats (and not sure whether we should change the aot loader), it is also a little tiresome.
> 
> why we should maintain two AOT format? I think it is the same with normal XIP aot, and do not need change aot loader, just enlarge the the lookup table and follow the aot loading process as before.

It changes the XIP's file format, so there are more differences between the formats of XIP and non-XIP than before. And from the comment in the code, it is only for XIP mode for xtensa, but from the changed code, it is for XIP for all targets, I am not sure whether we should leave the format unchanged for non-xtensa targets since the PR increases the binary size. My opinion is we had better keep the AOT file format as simple as we can to reduce the maintain effort, but it deserves more discussion.

@no1wudi could you help review the PR? Not sure whether the change of AOT file format is good to you since you also use XIP a lot.

no1wudi:
I prefer to enhance LLVM, change the AOT file format will cause extra overhead for ARM/RISC-V.

core/iwasm/compilation/aot_llvm.c
```diff
@@ -732,7 +777,9 @@ aot_add_llvm_func(AOTCompContext *comp_ctx, LLVMModuleRef module,
     }
 
     if (need_precheck) {
-        if (!comp_ctx->is_jit_mode)
+        if (!comp_ctx->is_jit_mode
+            && (!comp_ctx->is_indirect_mode
+                && !strncmp(comp_ctx->target_arch, "xtensa", 6)))
```
wenyongh, line 782: Seems it should be like below?
```C
        if (!comp_ctx->is_jit_mode
            && !(comp_ctx->is_indirect_mode
                 && !strncmp(comp_ctx->target_arch, "xtensa", 6)))
```

core/iwasm/compilation/aot_llvm.c
```diff
@@ -537,8 +539,51 @@ aot_build_precheck_function(AOTCompContext *comp_ctx, LLVMModuleRef module,
...
+
+        LLVMTypeRef func_ptr_type;
+        LLVMValueRef wrapped_func_indirect;
+        uint32 import_func_count = comp_ctx->comp_data->import_func_count;
+        uint32 func_count = comp_ctx->func_ctx_count;
+
+        /* Check function index */
+        if (func_index >= import_func_count + func_count) {
+            aot_set_last_error("Function index out of range.");
+            return false;
```
wenyongh, line 559: Should be `goto fail`?

core/iwasm/compilation/aot_emit_aot_file.c
```diff
@@ -2409,7 +2423,10 @@ aot_emit_init_data_section(uint8 *buf, uint8 *buf_end, uint32 *p_offset,
         return false;
 
     offset = align_uint(offset, 4);
-    EMIT_U32(comp_data->func_count);
+    if (need_call_wrapped_indirect(obj_data))
+        EMIT_U32(comp_data->func_count * 2);
+    else
+        EMIT_U32(comp_data->func_count);
```
wenyongh, line 2429: Could we not change here since the above get_func_section_size and below emitting text_offset_of_aot_func_internal/func_type_index are calculated with the original func_count? We can just add reading extra text_offset_of_aot_func_internal and func_type_index in the aot loader.

core/iwasm/aot/aot_loader.c
```diff
@@ -2558,9 +2558,19 @@ load_function_section(const uint8 *buf, const uint8 *buf_end, AOTModule *module,
...
+    /*
+     * For Xtensa XIP, real func_count is doubled, including aot_func and
+     * aot_func_internal, so need to divide func_count by 2 here.
+     */
+    if (module->is_indirect_mode)
+        size = sizeof(uint32) * (uint64)module->func_count / 2;
+    else
+#else
     size = sizeof(uint32) * (uint64)module->func_count;
+#endif
```
wenyongh, line 2571: If we don't emit double func_count in aot_emit_aot_file.c, then here we don't need to change. We can change the above reading func_ptrs and func_type_index to be like below?
```C
    size = sizeof(void *) * ((uint64)module->func_count * 2);
    if (size > 0
        && !(module->func_ptrs =
                 loader_malloc(size, error_buf, error_buf_size))) {
        return false;
    }
    for (i = 0; i < module->func_count * 2; i++) {
        ...
    }

    size = sizeof(uint32) * ((uint64)module->func_count * 2);
    if (size > 0
        && !(module->func_type_indexes =
                 loader_malloc(size, error_buf, error_buf_size))) {
        return false;
    }
    for (i = 0; i < module->func_count * 2; i++) {
        ...
    }
```

zhanheng1: hi @wenyongh ，module->func_count has been kept unchanged in latest commit 1fe0beacdbf53028fad5c1b271173af3a93e2104. 
Some macros and ifdef are added in aot_loader.c and aot_runtime.c like below:

```
uint32 func_count = module->func_count;
#if defined(BUILD_TARGET_XTENSA)
    /*
     * For Xtensa XIP, real func_count is doubled, including aot_func and
     * aot_func_internal, so need to multipy func_count by 2 here.
     */
    if (module->is_indirect_mode) {
        func_count *= 2;
    }
#endif

size = sizeof(void *) * ((uint64)module->func_count * 2);
    if (size > 0
        && !(module->func_ptrs =
                 loader_malloc(size, error_buf, error_buf_size))) {
        return false;
    }
    for (i = 0; i < module->func_count * 2; i++) {
        ...
    }
......
```

core/iwasm/compilation/aot_llvm.c
```diff
@@ -537,8 +539,51 @@ aot_build_precheck_function(AOTCompContext *comp_ctx, LLVMModuleRef module,
...
         name = "";
     }
-    LLVMValueRef retval =
-        LLVMBuildCall2(b, func_type, wrapped_func, params, param_count, name);
+
+    LLVMValueRef retval;
+    if (comp_ctx->is_indirect_mode
+        && !strncmp(comp_ctx->target_arch, "xtensa", 6)) {
+        /* call wrapped_func indirectly */
+        if (!create_func_ptrs(comp_ctx, func_ctx)) {
```
wenyongh, line 547: create_func_ptrs here may cause func_ptrs duplicated created since it will be created again in:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/169e1648156d008825831fb00d5e3acbe021b480/core/iwasm/compilation/aot_llvm.c#L1775-L1777

Could we remove the code here, and move the code in the above link to be before calling aot_add_llvm_func:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/169e1648156d008825831fb00d5e3acbe021b480/core/iwasm/compilation/aot_llvm.c#L1705-L1709

zhanheng1: These two func_ptrs are pointed to different func_ctx, one is for the original function context like aot_func_internal#0, and the other one is for precheck function like aot_func#0. 

Different func_ctx can not use the same func_ptr, if we try to use the func_ptr of a func_ctx in other func_ctx,  error log like "Referring to an argument in another function" will output during compilation. So it may be better to call create_func_ptrs() twice for two func_ctx.

dongsheng28849455:
LGTM

no1wudi (approved):

PR #3150, opened 2024-02-07
Fix null pointer access in fast interpreter mode  when configurable software bound check is enabled

loganek:
The wasm_interp_call_func_bytecode is called for the first time with the empty module / exec_env to generate a global_handle_table. Before that happens though, the function checks if the module instance has bounds check enabled. Because the module instance is null, the program crashes. I added an extra check to prevent the crashes.

### DIFF
--- a/core/iwasm/common/wasm_memory.c
+++ b/core/iwasm/common/wasm_memory.c
@@ -104,6 +104,10 @@ static inline bool
 is_bounds_checks_enabled(WASMModuleInstanceCommon *module_inst)
 {
 #if WASM_CONFIGURABLE_BOUNDS_CHECKS != 0
+    if (!module_inst) {
+        return true;
+    }
+
     return wasm_runtime_is_bounds_checks_enabled(module_inst);
 #else
     return true;


core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -1483,8 +1483,9 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
 #if !defined(OS_ENABLE_HW_BOUND_CHECK) \
     || WASM_CPU_SUPPORTS_UNALIGNED_ADDR_ACCESS == 0
 #if WASM_CONFIGURABLE_BOUNDS_CHECKS != 0
-    bool disable_bounds_checks = !wasm_runtime_is_bounds_checks_enabled(
-        (WASMModuleInstanceCommon *)module);
+    bool disable_bounds_checks = !module
+                                 && !wasm_runtime_is_bounds_checks_enabled(
```
wenyongh, line 1487: Seems to be `!(module && wasm_runtime_is_bounds_checks_enabled((WASMModuleInstanceCommon *)module))`?

And other places which call `wasm_runtime_is_bounds_checks_enabled` may have the same issue, like wasm_interp_classic.c and wasm_memory.c, so how about changing the implementation of wasm_runtime_is_bounds_checks_enabled instead:
```C
bool
wasm_runtime_is_bounds_checks_enabled(WASMModuleInstanceCommon *module_inst)
{
    if (!module_inst)
        return false;
    ...
}
```

loganek: > Seems to be `!(module && wasm_runtime_is_bounds_checks_enabled((WASMModuleInstanceCommon *)module))`?
> 
It doesn't really matter, as the function should never execute any instruction if the module is not defined (therefore the flag is unused).

> And other places which call `wasm_runtime_is_bounds_checks_enabled` may have the same issue, like wasm_interp_classic.c and wasm_memory.c, so how about changing the implementation of wasm_runtime_is_bounds_checks_enabled instead:
> 
> ```c
> bool
> wasm_runtime_is_bounds_checks_enabled(WASMModuleInstanceCommon *module_inst)
> {
>     if (!module_inst)
>         return false;
>     ...
> }
> ```

I checked classic interpreter and this one doesn't have the problem. About moving the check to the function itself - I'm not 100% sure; it feels like it's caller's responsibility to make sure the correct parameter is being passed, although I don't insist.

wenyongh: > > Seems to be `!(module && wasm_runtime_is_bounds_checks_enabled((WASMModuleInstanceCommon *)module))`?
> 
> It doesn't really matter, as the function should never execute any instruction if the module is not defined (therefore the flag is unused).

But the logic is strange/tricky, `!module && !wasm_runtime_is_bounds_checks_enabled(module)` means only when module is NULL and `module's disable_bounds_checks flag is true`, the expression is true. If module is already NULL, we don't need to add the latter expression, the whole sentence is equal to `!module`? I still think it should be `!(module && wasm_runtime_is_bounds_checks_enabled((WASMModuleInstanceCommon *)module))`, only when module isn't NULL and its disable_bounds_checks flag is false, can bounds checks be actually enabled?

> 
> > And other places which call `wasm_runtime_is_bounds_checks_enabled` may have the same issue, like wasm_interp_classic.c and wasm_memory.c, so how about changing the implementation of wasm_runtime_is_bounds_checks_enabled instead:
> > ```c
> > bool
> > wasm_runtime_is_bounds_checks_enabled(WASMModuleInstanceCommon *module_inst)
> > {
> >     if (!module_inst)
> >         return false;
> >     ...
> > }
> > ```
> 
> I checked classic interpreter and this one doesn't have the problem. About moving the check to the function itself - I'm not 100% sure; it feels like it's caller's responsibility to make sure the correct parameter is being passed, although I don't insist.

OK, though I think it is more general to add the check inside wasm_runtime_is_bounds_checks_enabled.

loganek: ok, I moved the check to the `wasm_runtime_is_bounds_checks_enabled` function.

wenyongh (approved):

PR #2984, opened 2024-01-08
a rough idea about the Rust API

lum1n0us:
The PR is trying to define the rust API layout (physically and logically). Based on those, finish examples (in comments) and documents.

PR #3152, opened 2024-02-07
[WIP] enable llvm jit by default

lum1n0us:
It bases on https://github.com/bytecodealliance/wasm-micro-runtime/pull/2984


language-bindings/rust/src/runtime.rs
```diff
@@ -164,13 +157,22 @@ mod tests {
...
+        unsafe { wasm_runtime_free(small_buf) };
     }
 
     #[test]
     fn test_runtime_builder_default() {
+        // use Mode_Default
         let runtime = Runtime::builder().use_system_allocator().build();
         assert!(runtime.is_ok());
         drop(runtime);
+
```
lum1n0us, line 172: after `drop`, the following `malloc` should be failed. I will fix in #3167

wenyongh (approved):

PR #3162, opened 2024-02-15
zephyr: use zephyr sys_cache instead of CMSIS

mkolchurin:
when running AOT code in Zephyr on STM32H743VIT6 without `CONFIG_CACHE_MANAGEMENT=y`, a hard fault occurs, which leads to `SCB_CleanDCache()`.

I think it’s better to use the functions built into Zephyr

### DIFF
--- a/core/shared/platform/zephyr/zephyr_platform.c
+++ b/core/shared/platform/zephyr/zephyr_platform.c
@@ -202,10 +202,14 @@ void
 os_dcache_flush()
 {
 #if defined(CONFIG_CPU_CORTEX_M7) && defined(CONFIG_ARM_MPU)
+#if KERNEL_VERSION_NUMBER < 0x030300 /* version 3.3.0 */
     uint32 key;
     key = irq_lock();
     SCB_CleanDCache();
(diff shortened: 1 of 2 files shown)


core/shared/platform/zephyr/zephyr_platform.c
```diff
@@ -202,10 +202,7 @@ void
 os_dcache_flush()
 {
 #if defined(CONFIG_CPU_CORTEX_M7) && defined(CONFIG_ARM_MPU)
-    uint32 key;
-    key = irq_lock();
-    SCB_CleanDCache();
-    irq_unlock(key);
+    sys_cache_data_flush_all();
```
wenyongh, line 205: Are the two functions (sys_cache_data_flush_all and sys_cache_instr_flush_range) supported in most zephyr versions, should we add version macro to control the code?

mkolchurin: I think you are right, for the first time this API was added in 3.3.0 [link](https://github.com/zephyrproject-rtos/zephyr/pull/50988).
Added changes

PR #3166, opened 2024-02-18
Implement remaining Windows filesystem functions

zoraaver:
Now that the filesystem implementation is now complete, the previous test filters on Windows can be removed. Some of the tests only pass when certain environment variables have been set on Windows so an extra step has been added in the wasi test runner script to modify the test config files before the tests begin.

### DIFF
--- a/core/shared/platform/windows/win_file.c
+++ b/core/shared/platform/windows/win_file.c
@@ -213,19 +213,18 @@ has_symlink_attribute(DWORD attributes)
     return (attributes & FILE_ATTRIBUTE_REPARSE_POINT) != 0;
 }
 
-static bool
-is_symlink(const wchar_t *path)
-{
-    DWORD attributes = GetFileAttributesW(path);
-
-    return has_symlink_attribute(attributes);
-}
-
 static void
 init_dir_stream(os_dir_stream dir_stream, os_file_handle handle)
 {
     dir_stream->cursor = 0;
     dir_stream->handle = handle;
+    dir_stream->cookie = 0;
+}
+
+static void
+reset_dir_stream(os_dir_stream dir_stream)
+{
+    dir_stream->cursor = 0;
     dir_stream->cookie = 0;
 }
 
@@ -562,7 +561,32 @@ os_fstatat(os_file_handle handle, const char *path,
 {
     CHECK_VALID_FILE_HANDLE(handle);
 
-    return __WASI_ENOSYS;
+    wchar_t absolute_path[PATH_MAX];
+
+    __wasi_errno_t error = get_absolute_filepath(handle->raw.handle, path,
+                                                 absolute_path, PATH_MAX);
+
+    if (error != __WASI_ESUCCESS)
+        return error;
+
+    windows_handle resolved_handle = {
+        .type = windows_handle_type_file,
+        .fdflags = 0,
+        .raw = { .handle = create_handle(
+                     absolute_path, is_directory(absolute_path),
+                     ((lookup_flags & __WASI_LOOKUP_SYMLINK_FOLLOW) != 0),
+                     true) },
+        .access_mode = windows_access_mode_read
+    };
+
+    if (resolved_handle.raw.handle == INVALID_HANDLE_VALUE)
+        return convert_windows_error_code(GetLastError());
+
+    error = get_file_information(&resolved_handle, buf);
+
+    CloseHandle(resolved_handle.raw.handle);
+
+    return error;
 }
 
 __wasi_errno_t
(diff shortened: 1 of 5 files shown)


product-mini/platforms/windows/wasi_filtered_tests.json
```diff
@@ -1,34 +1,7 @@
...
-        "path_rename": "Not implemented",
-        "path_rename_dir_trailing_slashes": "Not implemented",
-        "path_symlink_trailing_slashes": "Not implemented",
-        "poll_oneoff_stdio": "Not implemented",
-        "readlink": "Not implemented (path_symlink)",
-        "symlink_create": "Not implemented",
-        "symlink_filestat": "Not implemented",
-        "symlink_loop": "Not implemented",
-        "truncation_rights": "Not implemented"
+        "path_filestat": "This test attempts to open a file descriptor with fdflags APPEND & SYNC which is not supported on Windows",
```
zoraaver, line 3: I'll raise a separate PR in wasi testsuite to amend the test behaviour - as far as I can tell APPEND isn't important to the test. The test just asserts that the flag is present if it was set so we can change it to another fdflag which doesn't have this issue on Windows.

core/shared/platform/windows/win_file.c
```diff
@@ -1035,15 +1100,71 @@ os_fallocate(os_file_handle handle, __wasi_filesize_t offset,
 {
     CHECK_VALID_FILE_HANDLE(handle);
 
-    return __WASI_ENOSYS;
+    LARGE_INTEGER current_file_size;
+    GetFileSizeEx(handle->raw.handle, &current_file_size);
```
wenyongh, line 1104: Had better check return value of `GetFileSizeEx`?

core/shared/platform/windows/win_file.c
```diff
@@ -1392,15 +1708,38 @@ os_readdir(os_dir_stream dir_stream, __wasi_dirent_t *entry,
...
 __wasi_errno_t
 os_closedir(os_dir_stream dir_stream)
 {
     CHECK_VALID_WIN_DIR_STREAM(dir_stream);
 
-    return __WASI_ENOSYS;
+    bool success = CloseHandle(dir_stream->handle->raw.handle);
+
+    if (!success)
+        return convert_windows_error_code(GetLastError());
```
wenyongh, line 1738: Should we `BH_FREE(dir_stream)` before return error?

wenyongh:
@loganek Could you help review this PR? It should be in ready status, the CI failures are due to other issues and we are looking into them.

core/shared/platform/windows/win_file.c
```diff
@@ -1224,18 +1365,96 @@ os_linkat(os_file_handle from_handle, const char *from_path,
...
+                                  absolute_to_path, PATH_MAX);
+
+    if (error != __WASI_ESUCCESS)
+        return error;
+
+    size_t to_path_len = strlen(to_path);
+
+    // Windows doesn't throw an error in the case that the new path has a
+    // trailing slash but the target to link to is a file.
+    if (to_path[to_path_len - 1] == '/'
```
loganek, line 1392: do we have a test for that in WASI?

tests/wamr-test-suites/wasi-test-script/run_wasi_tests.sh
```diff
@@ -41,6 +41,12 @@ readonly THREAD_INTERNAL_TESTS="${WAMR_DIR}/core/iwasm/libraries/lib-wasi-thread
 readonly THREAD_STRESS_TESTS="${WAMR_DIR}/core/iwasm/libraries/lib-wasi-threads/stress-test/"
 readonly LIB_SOCKET_TESTS="${WAMR_DIR}/core/iwasm/libraries/lib-socket/test/"
 
+add_env_key_to_test_config_file() {
+    filepath="tests/$2/testsuite/$3.json"
```
loganek, line 45: Instead of modifying test files, could we perhaps extend the WAMR adapter and allow passing env variables through it?

core/shared/platform/windows/win_file.c
zoraaver: yes, https://github.com/WebAssembly/wasi-testsuite/blob/main/tests/rust/src/bin/path_link.rs#L139-L144

core/shared/platform/windows/win_file.c
zoraaver: good spot - will check it

core/shared/platform/windows/win_file.c
zoraaver: I don't think so - it's hard to know from the [CloseHandle documentation](https://learn.microsoft.com/en-us/windows/win32/api/handleapi/nf-handleapi-closehandle) but if the error returned is recoverable, we would leak the handle if it isn't closed here and we free the owning `dir_stream` struct. In practice, I think the only reason `CloseHandle` would fail is if the provided handle is invalid - so we should probably call `BH_FREE(dir_stream)` if `GetLastError` returns `ERROR_INVALID_HANDLE`.

tests/wamr-test-suites/wasi-test-script/run_wasi_tests.sh
zoraaver: Opened a PR, https://github.com/WebAssembly/wasi-testsuite/pull/99.

wenyongh (approved):

PR #3167, opened 2024-02-20
[rust sdk] fix a bug in singleton pattern

lum1n0us:
It is based on #3152

### DIFF
--- a/language-bindings/rust/src/instance.rs
+++ b/language-bindings/rust/src/instance.rs
@@ -100,10 +100,14 @@ impl Drop for Instance {
 mod tests {
     use super::*;
     use crate::runtime::Runtime;
+    use wamr_sys::{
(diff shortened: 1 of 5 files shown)


language-bindings/rust/src/instance.rs
```diff
@@ -100,10 +100,14 @@ impl Drop for Instance {
 mod tests {
     use super::*;
     use crate::runtime::Runtime;
+    use wamr_sys::{
+        wasm_runtime_get_running_mode, RunningMode_Mode_Interp, RunningMode_Mode_LLVM_JIT,
```
wenyongh, line 104: Is it better to rename to `Running_Mode_Interp`, `Running_Mode_LLVM_JIT`?

wenyongh (approved):

PR #3169, opened 2024-02-20
Small llvm-jit/aot bug fix

TianlongLiang:
When dealing with non-gc enabled funcref/externref, need to make sure to push them as i32 type


wenyongh (approved):
LGTM

PR #3171, opened 2024-02-21
Fix inconsistent coding convention

no1wudi:


core/iwasm/common/wasm_shared_memory.c
```diff
@@ -232,14 +232,14 @@ destroy_wait_info(void *wait_info)
 }
 
 static void
-map_try_release_wait_info(HashMap *wait_map_, AtomicWaitInfo *wait_info,
+map_try_release_wait_info(HashMap *wait_map, AtomicWaitInfo *wait_info,
```
wenyongh, line 235: This argument may shadow the variable defined in L44 and cause compilation warning, had better use other name here.

wenyongh: Or how about removing this argument, as this function is only called by wasm_runtime_atomic_wait, which always passes the global wait_map to this function.

no1wudi: Fixed, but it's strange, my compiler don't report the shadowing

wenyongh (approved):

core/iwasm/common/wasm_shared_memory.c
wenyongh: Yes, maybe it is only reported by some compiler.

PR #3172, opened 2024-02-21
VSCode IDE enhancement and readme update

TianlongLiang:
- Temporarily disable the CI test for the VS Code extension, as the Rust formatter seems to malfunction after a recent VS Code update
- Add configuration for iwasm heap size, allowing users to flexibly `malloc` memory. This also fix the current bug that when default size is 0, it can't run and debug.
- Apply code style formatting for WAMR IDE source code and add a format check for it in CI.


wenyongh (approved):
LGTM

PR #3177, opened 2024-02-22
zephyr: include math only with minimal libc

mkolchurin:
use math functions, only with `CONFIG_MINIMAL_LIBC=y`.

`CONFIG_PICOLIBC=y` or `CONFIG_NEWLIB_LIBC=y` provide math functions that are used by wasm, and compilation fails when they are selected.


mkolchurin:
It looks like picolibc and newlib_libc have been around since version 2.1 ([link](https://github.com/zephyrproject-rtos/zephyr/commit/c1f865b412683970968780a96fb8a09c8c4000da)) and we don't need any additional versions checks

wenyongh (approved):

PR #3181, opened 2024-02-26
Fix ref.func opcode check when GC is enabled

wenyongh:


wenyongh:
Fixed with https://github.com/bytecodealliance/wasm-micro-runtime/pull/3181.

wenyongh:
The root cause is when GC is enabled, the type of table segment can be (ref func) besides with funcref, while the original code only checks funcref type.

PR #3182, opened 2024-02-28
[WIP] Some preparation for crates publishing 

lum1n0us:
Plan to
- keep wamr-sys-gen in wamr repo to generate bindings.rs, which is also lib.rs in wamr-sys. (There is a question about how to sync those two files)
- store wamr-sys and wamr-rust-sdk in a separate repo
- publish wamr-sys and wamr-rust-sdk. or at least those two can be referred with "git"


lum1n0us:
#3191

Issue #3187, opened 2024-02-29
C++ related issues in dwarf_extractor.cpp


yamt:
when compiling a C++ module, in lldb_function_to_function_dbi,

* `function_args` doesn't seem to include the "this" pointer. on the other hand, `variable_list` contains it. it causes the "function args number dismatch" message and then ParamType overrun, which likely involves a crash.
* `variable.GetName()` seems to return NULL for unnamed arg. it causes a NULL dereference.

yamt:
btw, i wonder why dwarf_extractor.cpp uses the LLDB API, rather than, say, llvm/DebugInfo/DWARF/ api.
@wenyongh do you remember?

wenyongh:
> btw, i wonder why dwarf_extractor.cpp uses the LLDB API, rather than, say, llvm/DebugInfo/DWARF/ api. @wenyongh do you remember?

The source debugger feature was implemented by Ant group long time ago, I just cannot remember the details now. Maybe the LLDB API provides more capability?

yamt:
a quick research on how other projects are reading the dwarf section in wasm:

binaryen: llvm DWARFContext https://github.com/WebAssembly/binaryen/blob/d6c5e4ab15df271521df7b35665c7463b2c490ca/src/wasm/wasm-debug.cpp#L96-L116

cranelift: https://github.com/gimli-rs/gimli (something written in rust)

TianlongLiang:
It seems that AOT debugging now only has limited functionality. It was aimed to support C first It hardly covers any additional C++ features. Thus the current function-to-function debug symbol mapping only supports C styles function. And I think that's why it error when it comes to Cpp compiled wasm or Rust compiled wasm

yamt:
> It seems that AOT debugging now only has limited functionality. It was aimed to support C first It hardly covers any additional C++ features. Thus the current function-to-function debug symbol mapping only supports C styles function. And I think that's why it error when it comes to Cpp compiled wasm or Rust compiled wasm

i agree.

i tend to think a fix involves a major surgery like switching from lldb api to llvm api to read dwarf. how do you think?

TianlongLiang:
I think it's feasible, but as you said it would be non-trivial changes. BTW, besides the extractor, will there also be some major modifications for code emission and file generation?

yamt:
> I think it's feasible, but as you said it would be non-trivial changes. BTW, besides the extractor, will there also be some major modifications for code emission and file generation?

[...]

PR #3189, opened 2024-02-29
lldb_function_to_function_dbi: fix a null dereference

yamt:
C++ allows unnamed arguments.
In the debug info, they are represented as DW_TAG_formal_parameter w/o DW_AT_name.
variable.GetName() here returns NULL for them.

cf. https://github.com/bytecodealliance/wasm-micro-runtime/issues/3187


wenyongh:
LGTM

PR #3190, opened 2024-02-29
lldb_function_to_function_dbi: a hack to avoid crashing on C++ methods

yamt:
Also, print the function name on argument mismatch.

### DIFF
--- a/core/iwasm/compilation/debug/dwarf_extractor.cpp
+++ b/core/iwasm/compilation/debug/dwarf_extractor.cpp
@@ -354,10 +354,27 @@ lldb_function_to_function_dbi(const AOTCompContext *comp_ctx,
         LLVMDIBuilderCreateExpression(DIB, NULL, 0);
     auto variable_list =
         function.GetBlock().GetVariables(extractor->target, true, false, false);
+    unsigned int variable_offset = 0;
     if (num_function_args != variable_list.GetSize()) {
-        LOG_ERROR(
-            "function args number dismatch!:value number=%d, function args=%d",
-            variable_list.GetSize(), num_function_args);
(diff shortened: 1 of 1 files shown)


core/iwasm/compilation/debug/dwarf_extractor.cpp
```diff
@@ -354,10 +354,27 @@ lldb_function_to_function_dbi(const AOTCompContext *comp_ctx,
...
-        LOG_ERROR(
-            "function args number dismatch!:value number=%d, function args=%d",
-            variable_list.GetSize(), num_function_args);
+        // A hack to detect C++ "this" pointer.
+        //
+        // REVISIT: is there a more reliable way?
+        // At the DWARF level, we can probably look at DW_AT_object_pointer
+        // and DW_AT_artificial. I'm not sure how it can be done via the
+        // LLDB API though.
+        if (num_function_args + 1 == variable_list.GetSize()) {
```
xujuntwt95329, line 365: I also didn't find a way to get `DW_AT_object_pointer` from LLDB API.

This workaround seems work, but I have a concern on this condition: would it cause error if this is a vaargs function?
Of course this is not a new problem introduced by this PR, the L358 didn't consider vaargs function either. I'm not sure what will happen for vaargs functions here.

yamt: my understanding is that, neither `num_function_args` or `variable_list.GetSize()` here include the hidden arg for stdarg.

xujuntwt95329: Thanks

xujuntwt95329:
LGTM

PR #3194, opened 2024-03-01
fix a bug when applying `offset` in call stacks information

lum1n0us:
the offset value is based on start of wasm file. it also equals to the value of `wasm-objdump -d`


wenyongh (approved):
LGTM

PR #3198, opened 2024-03-02
Allow overwriting max memory on module instantiation

eloparco:
This PR adds a `max_memory_pages` parameter to module instantiation APIs, to allow overriding the max memory defined in the WASM module.

Sticking to the max memory defined in the module is quite limiting when using shared memory in production. If targeted devices have different memory constraints, many wasm files have to be generated with different max memory values. And device constraints may not be known in advance.

Being able to set the max memory value during module instantiation allows to reuse the same wasm module, e.g. by retrying instantiation with different max memory value.

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -790,7 +790,8 @@ memory_instantiate(AOTModuleInstance *module_inst, AOTModuleInstance *parent,
     uint32 num_bytes_per_page = memory->num_bytes_per_page;
     uint32 init_page_count = memory->mem_init_page_count;
     uint32 max_page_count =
-        max_memory_pages == 0 ? memory->mem_max_page_count : max_memory_pages;
+        wasm_runtime_get_max_mem(max_memory_pages, memory->mem_init_page_count,
+                                 memory->mem_max_page_count);
     uint32 inc_page_count, aux_heap_base, global_idx;
     uint32 bytes_of_last_page, bytes_to_page_end;
     uint32 heap_offset = num_bytes_per_page * init_page_count;
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -783,12 +783,14 @@ static AOTMemoryInstance *
 memory_instantiate(AOTModuleInstance *module_inst, AOTModuleInstance *parent,
                    AOTModule *module, AOTMemoryInstance *memory_inst,
                    AOTMemory *memory, uint32 memory_idx, uint32 heap_size,
-                   char *error_buf, uint32 error_buf_size)
+                   uint32 max_memory_pages, char *error_buf,
+                   uint32 error_buf_size)
 {
     void *heap_handle;
     uint32 num_bytes_per_page = memory->num_bytes_per_page;
     uint32 init_page_count = memory->mem_init_page_count;
-    uint32 max_page_count = memory->mem_max_page_count;
+    uint32 max_page_count =
+        max_memory_pages == 0 ? memory->mem_max_page_count : max_memory_pages;
     uint32 inc_page_count, aux_heap_base, global_idx;
     uint32 bytes_of_last_page, bytes_to_page_end;
     uint32 heap_offset = num_bytes_per_page * init_page_count;
@@ -984,7 +986,8 @@ aot_get_default_memory(AOTModuleInstance *module_inst)
 
 static bool
 memories_instantiate(AOTModuleInstance *module_inst, AOTModuleInstance *parent,
-                     AOTModule *module, uint32 heap_size, char *error_buf,
+                     AOTModule *module, uint32 heap_size,
+                     uint32 max_memory_pages, char *error_buf,
                      uint32 error_buf_size)
 {
     uint32 global_index, global_data_offset, base_offset, length;
@@ -1002,9 +1005,9 @@ memories_instantiate(AOTModuleInstance *module_inst, AOTModuleInstance *parent,
 
     memories = module_inst->global_table_data.memory_instances;
     for (i = 0; i < memory_count; i++, memories++) {
-        memory_inst = memory_instantiate(module_inst, parent, module, memories,
-                                         &module->memories[i], i, heap_size,
-                                         error_buf, error_buf_size);
+        memory_inst = memory_instantiate(
+            module_inst, parent, module, memories, &module->memories[i], i,
+            heap_size, max_memory_pages, error_buf, error_buf_size);
         if (!memory_inst) {
             return false;
         }
@@ -1461,7 +1464,7 @@ check_linked_symbol(AOTModule *module, char *error_buf, uint32 error_buf_size)
 AOTModuleInstance *
 aot_instantiate(AOTModule *module, AOTModuleInstance *parent,
                 WASMExecEnv *exec_env_main, uint32 stack_size, uint32 heap_size,
-                char *error_buf, uint32 error_buf_size)
+                uint32 max_memory_pages, char *error_buf, uint32 error_buf_size)
 {
     AOTModuleInstance *module_inst;
 #if WASM_ENABLE_BULK_MEMORY != 0 || WASM_ENABLE_REF_TYPES != 0
@@ -1551,7 +1554,7 @@ aot_instantiate(AOTModule *module, AOTModuleInstance *parent,
         &((AOTModuleInstanceExtra *)module_inst->e)->sub_module_inst_list_head;
     ret = wasm_runtime_sub_module_instantiate(
         (WASMModuleCommon *)module, (WASMModuleInstanceCommon *)module_inst,
-        stack_size, heap_size, error_buf, error_buf_size);
+        stack_size, heap_size, max_memory_pages, error_buf, error_buf_size);
     if (!ret) {
         LOG_DEBUG("build a sub module list failed");
         goto fail;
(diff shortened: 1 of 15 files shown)


core/iwasm/aot/aot_runtime.c
```diff
@@ -783,12 +783,14 @@ static AOTMemoryInstance *
...
-                   char *error_buf, uint32 error_buf_size)
+                   uint32 max_memory_pages, char *error_buf,
+                   uint32 error_buf_size)
 {
     void *heap_handle;
     uint32 num_bytes_per_page = memory->num_bytes_per_page;
     uint32 init_page_count = memory->mem_init_page_count;
-    uint32 max_page_count = memory->mem_max_page_count;
+    uint32 max_page_count =
+        max_memory_pages == 0 ? memory->mem_max_page_count : max_memory_pages;
```
wenyongh, line 793: max_page_count should be no smaller than init_page_count, how about:
```C
uint32 max_page_count =
        max_memory_pages == 0 ? memory->mem_max_page_count :
        (max_memory_pages > init_page_count ? max_memory_pages : init_page_count);
```

core/iwasm/include/wasm_export.h
```diff
@@ -527,6 +527,32 @@ wasm_runtime_instantiate(const wasm_module_t module,
...
+ * @param max_memory_pages the number of pages (64k each) to use as maximum memory
+ * @param error_buf buffer to output the error info if failed
+ * @param error_buf_size the size of the error buffer
+ *
+ * @return return the instantiated WASM module instance, NULL if failed
+ */
+WASM_RUNTIME_API_EXTERN wasm_module_inst_t
+wasm_runtime_instantiate_ex(const wasm_module_t module,
+                         uint32_t default_stack_size, uint32_t host_managed_heap_size,
+                         uint32_t max_memory_pages,
```
wenyongh, line 553: We may add more arguments in the future and had better not change the API prototype again, how about defining an `InstantiationArgs` struct, and here:
```C
wasm_runtime_instantiate_ex(const wasm_module_t module, const InstantiationArgs *args, char *error_buf, uint32_t error_buf_size);
```

core/iwasm/include/wasm_export.h
```diff
@@ -527,6 +527,32 @@ wasm_runtime_instantiate(const wasm_module_t module,
...
+ *        create the operation stack internally with the stack size specified
+ *        here. And API wasm_runtime_create_exec_env() creates the operation
+ *        stack with stack size specified by its parameter, the stack size
+ *        specified here is ignored.
+ * @param host_managed_heap_size the default heap size of the module instance, a heap will
+ *        be created besides the app memory space. Both wasm app and native
+ *        function can allocate memory from the heap.
+ * @param max_memory_pages the number of pages (64k each) to use as maximum memory
+ * @param error_buf buffer to output the error info if failed
+ * @param error_buf_size the size of the error buffer
```
wenyongh, line 546: Many duplicated comments with API wasm_runtime_instantiate, how about just mentioning same as it with some argument changes? Refer to
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/include/wasm_export.h#L489

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -374,7 +375,9 @@ memories_instantiate(const WASMModule *module, WASMModuleInstance *module_inst,
     for (i = 0; i < module->import_memory_count; i++, import++, memory++) {
         uint32 num_bytes_per_page = import->u.memory.num_bytes_per_page;
         uint32 init_page_count = import->u.memory.init_page_count;
-        uint32 max_page_count = import->u.memory.max_page_count;
+        uint32 max_page_count = max_memory_pages == 0
+                                    ? import->u.memory.max_page_count
+                                    : max_memory_pages;
```
wenyongh, line 380: max_page_count should be no smaller than init_page_count, had better add a check like comment in aot_runtime.c?

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -412,12 +415,15 @@ memories_instantiate(const WASMModule *module, WASMModuleInstance *module_inst,
 
     /* instantiate memories from memory section */
     for (i = 0; i < module->memory_count; i++, memory++) {
+        uint32 max_page_count = max_memory_pages == 0
+                                    ? module->memories[i].max_page_count
+                                    : max_memory_pages;
```
wenyongh, line 420: Same as above.

lum1n0us:
```
    uint32 max_page_count =
        max_memory_pages == 0 ? memory->mem_max_page_count : max_memory_pages;
```

Does it suggest that the `max_memory_pages` (specified by API) can be greater than `mem_max_page_count` from memory limitation?

eloparco:
> Does it suggest that the max_memory_pages (specified by API) can be greater than mem_max_page_count from memory limitation?

Is that not allowed? Actually I remember that was mentioned https://bytecodealliance.zulipchat.com/#narrow/stream/349267-wasi-threads/topic/rust/near/402263578, but couldn't find it in the specs.

wenyongh:
> > Does it suggest that the max_memory_pages (specified by API) can be greater than mem_max_page_count from memory limitation?
> 
> Is that not allowed? Actually I remember that was mentioned https://bytecodealliance.zulipchat.com/#narrow/stream/349267-wasi-threads/topic/rust/near/402263578, but couldn't find it in the specs.

No, from the comment of Alex, the max_memory_pages specified by the API can be smaller than what it is specified in wasm file, but should not be larger than it. So how about we add checks: if max_memory_pages from API is smaller than memory->init_page_count, then set it memory->init_page_count; and if max_memory_pages is larger than memory->max_page_count, then set it to memory->max_page_count?
@lum1n0us what's your opinion?

eloparco:
> So how about we add checks: if max_memory_pages from API is smaller than memory->init_page_count, then set it memory->init_page_count; and if max_memory_pages is larger than memory->max_page_count, then set it to memory->max_page_count?

Yeah, makes sense, I'll change the code to do that

core/iwasm/include/wasm_export.h
eloparco: Good idea

core/iwasm/aot/aot_runtime.c
```diff
@@ -783,12 +783,18 @@ static AOTMemoryInstance *
...
+                   uint32 max_memory_pages, char *error_buf,
+                   uint32 error_buf_size)
 {
     void *heap_handle;
     uint32 num_bytes_per_page = memory->num_bytes_per_page;
     uint32 init_page_count = memory->mem_init_page_count;
-    uint32 max_page_count = memory->mem_max_page_count;
+    uint32 max_page_count =
+        max_memory_pages == 0
+            ? memory->mem_max_page_count
```
wenyongh, line 794: How about handling `max_memory_pages == 0` inside `wasm_runtime_get_overwritten_max_mem`, and here it just calls `wasm_runtime_get_overwritten_max_mem`?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1357,24 +1357,44 @@ wasm_runtime_unload(WASMModuleCommon *module)
 #endif
 }
 
+uint32
+wasm_runtime_get_overwritten_max_mem(uint32 max_memory_pages,
+                                     uint32 module_init_page_count,
+                                     uint32 module_max_page_count)
+{
```
wenyongh, line 1364: How about adding
```C
    if (max_memory_pages == 0) {
        return module_max_page_count;
    }
```

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1357,24 +1357,44 @@ wasm_runtime_unload(WASMModuleCommon *module)
...
+    if (max_memory_pages < module_init_page_count) {
+        LOG_WARNING("Cannot override max memory with value lower than module "
```
wenyongh, line 1366: Not sure is it better to use `LOG_WARNING("warning: ...")`?

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -374,7 +375,12 @@ memories_instantiate(const WASMModule *module, WASMModuleInstance *module_inst,
     for (i = 0; i < module->import_memory_count; i++, import++, memory++) {
         uint32 num_bytes_per_page = import->u.memory.num_bytes_per_page;
         uint32 init_page_count = import->u.memory.init_page_count;
-        uint32 max_page_count = import->u.memory.max_page_count;
+        uint32 max_page_count =
+            max_memory_pages == 0
+                ? import->u.memory.max_page_count
```
wenyongh, line 380: How about handling `max_memory_pages == 0` inside `wasm_runtime_get_overwritten_max_mem`, and here it just calls, `wasm_runtime_get_overwritten_max_mem`?

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -412,12 +418,18 @@ memories_instantiate(const WASMModule *module, WASMModuleInstance *module_inst,
 
     /* instantiate memories from memory section */
     for (i = 0; i < module->memory_count; i++, memory++) {
+        uint32 max_page_count =
+            max_memory_pages == 0
+                ? module->memories[i].max_page_count
```
wenyongh, line 423: Same as above

core/iwasm/common/wasm_runtime_common.c
eloparco: If the user is overriding the logger (i.e. `WAMR_BH_LOG`), they probably already have something to distinguish between different log levels: different colors (red for error, yellow for warning) or different log suffixes (e.g. `W`, `E` suffixes). So adding the log level to the log message is redundant.

wenyongh: OK, got it, thanks.

wenyongh (commented):
LGTM

PR #3204, opened 2024-03-06
Update document to add wamr-rust-sdk introduction

wenyongh:


lum1n0us:
Guess we don't need *.github/workflows/rust_linter.yml* and *language-bindings/rust/.gitignore*

PR #3206, opened 2024-03-06
Get location info from function indexes in addr2line script

eloparco:
Updating the `addr2line` script so that:
- line info is printed in a more convenient format, e.g.
```
0: c
        at wasm-micro-runtime/test-tools/addr2line/trap.c:5:1
1: b
        at wasm-micro-runtime/test-tools/addr2line/trap.c:11:12
2: a
        at wasm-micro-runtime/test-tools/addr2line/trap.c:17:12
```
similar to how Rust prints stack traces when there's a panic. In an IDE, the user can conveniently click on the printed path and be redirected to the file line.
- a new `--no-addr` argument can be provided to the script 

It can be used in fast interpreter mode (that is not supported by the script otherwise) or with older wamr versions (where the stack trace only had the function index info and not the function address). In that case, `wasm-objdump` is used to get the function name from the index and `llvm-dwarfdump` to obtain the location info (where the line refers to the start of the function).


eloparco:
Tested with a few simple examples:
```c
int
c(int n)
{
    __builtin_trap();
}

int
b(int n)
{
    n += 3;
    return c(n);
}

int
a(int n)
{
    return b(n);
}

int
main(int argc, char **argv)
{
    int i = 5;
    a(i);

    return 0;
}
```

```rust
fn c(_: i32) -> i32 {
    panic!("this is a panic!");
}

fn b(mut n: i32) -> i32 {
    n += 3;
    c(n)
}

fn a(n: i32) -> i32 {
    b(n)
}

fn main() {
    let i = 5;
    a(i);
}
```

If I want to run it in CI, should I add a new folder in `samples` with a program that crashes (as above)? Or should I use `test-tools/addr2line`?

test-tools/addr2line/addr2line.py
```diff
@@ -157,33 +242,48 @@ def main():
...
+            assert splitted is not None
 
-            offset = int(offset, 16)
-            offset = offset - code_section_start
-            line_info = get_line_info(llvm_dwarf_dump, args.wasm_file, offset)
-            if not line_info:
-                print(line)
+            _, offset, index = splitted
+            if not index.startswith("$f"):
+                print(f"{i}: {index}")
```
lum1n0us, line 260: wondering what kind of case it is?

test-tools/addr2line/addr2line.py
```diff
@@ -96,29 +95,72 @@ def get_line_info(dwarf_dump: Path, wasm_file: Path, offset: int) -> str:
...
 
-            function_name = m.groups()[0]
+        if "DW_AT_decl_file" in line:
+            function_file = get_dwarf_tag_value("DW_AT_decl_file", line)
 
-        if line.startswith("Line info"):
-            location = line
-            return (function_name, location)
+        if "DW_AT_decl_line" in line:
+            function_line = get_dwarf_tag_value("DW_AT_decl_line", line)
```
lum1n0us, line 158: duplicated with L154 and L155 ?

test-tools/addr2line/addr2line.py
```diff
@@ -145,6 +227,9 @@ def main():
     parser.add_argument("--wabt", type=Path, help="path to wabt")
     parser.add_argument("--wasm-file", type=Path, help="path to wasm file")
     parser.add_argument("call_stack_file", type=Path, help="path to a call stack file")
+    parser.add_argument(
+        "--no-addr", action="store_true", help="use call stack without addresses"
```
lum1n0us, line 231: since wamr outputs *addresses-like info* (actually offset to the pre-compiled function start) in call stack in fast-interpreter mode, how about 
```suggestion
        "--no-addr", action="store_true", help="use call stack without addresses or from fast-interpreter mode"
```

test-tools/addr2line/addr2line.py
```diff
@@ -130,13 +172,53 @@ def parse_line_info(line_info: str) -> ():
...
-    return m.groups() if m else None
+    if m is not None:
+        return m.groups()
+
+    # Old format
+    PATTERN = r"#([0-9]+) (\S+)"
+    m = re.match(PATTERN, line)
+    if m is not None:
+        return (m.groups()[0], None, m.groups()[1])
+
```
lum1n0us, line 194: better have an explicit return value when both `match` failed.

wenyongh:
> Tested with a few simple examples:
> 
> ```c
> int
> c(int n)
> {
>     __builtin_trap();
> }
> 
> int
> b(int n)
> {
>     n += 3;
>     return c(n);
> }
> 
> int
> a(int n)
> {
>     return b(n);
> }
> 
> int
> main(int argc, char **argv)
> {
>     int i = 5;
>     a(i);
> 
>     return 0;
> }
> ```
> 
> ```rust
> fn c(_: i32) -> i32 {
>     panic!("this is a panic!");
> }
> 
> fn b(mut n: i32) -> i32 {
>     n += 3;
>     c(n)
> }
> 
> fn a(n: i32) -> i32 {
>     b(n)
> }
> 
> fn main() {
>     let i = 5;
>     a(i);
> }
> ```
> 
> If I want to run it in CI, should I add a new folder in `samples` with a program that crashes (as above)? Or should I use `test-tools/addr2line`?

@eloparco how about creating a new folder under samples, so it not only tests the tool but also shows the usage for the developers.

test-tools/addr2line/addr2line.py
eloparco: What do you mean? L154-155 are for the file, L157-158 for the line

eloparco:
>@eloparco how about creating a new folder under samples, so it not only tests the tool but also shows the usage for the developers.

That's what I had in mind as well, added it.

When I use AOT, the first stack frame address seems to be incorrect:
```bash
#00: 0x0000 - $f5
#01: 0x01ac - $f6
#02: 0x01fa - $f7
#03: 0x0265 - $f8
#04: 0x2365 - $f15
#05: 0x0119 - _start
```
I'm compiling with `--enable-dump-call-stack`. And added `--size-level=0` as well to avoid problems on mac M1 (that I'm using for testing).

wenyongh (commented):
LGTM

lum1n0us:
LGTM

eloparco:
>When I use AOT, the first stack frame address seems to be incorrect:

@wenyongh @lum1n0us any ideas about this one? Opened an issue https://github.com/bytecodealliance/wasm-micro-runtime/issues/3212

wenyongh:
@eloparco please try #3213.

PR #3209, opened 2024-03-08
Refactor APIs and data structures as preliminary work for Memory64

wenyongh:
# Change the data type representing linear memory address from u32 to u64

## APIs signature changes
- (Export)wasm_runtime_module_malloc
  - wasm_module_malloc
    - wasm_module_malloc_internal
  - aot_module_malloc
    - aot_module_malloc_internal
- wasm_runtime_module_realloc
  - wasm_module_realloc
    - wasm_module_realloc_internal
  - aot_module_realloc
    - aot_module_realloc_internal
- (Export)wasm_runtime_module_free
  - wasm_module_free
    - wasm_module_free_internal
  - aot_module_malloc
    - aot_module_free_internal
- (Export)wasm_runtime_module_dup_data
  - wasm_module_dup_data
  - aot_module_dup_data
- (Export)wasm_runtime_validate_app_addr
- (Export)wasm_runtime_validate_app_str_addr
- (Export)wasm_runtime_validate_native_addr
- (Export)wasm_runtime_addr_app_to_native
- (Export)wasm_runtime_addr_native_to_app
- (Export)wasm_runtime_get_app_addr_range
- aot_set_aux_stack
- aot_get_aux_stack
- wasm_set_aux_stack
- wasm_get_aux_stack
- aot_check_app_addr_and_convert, wasm_check_app_addr_and_convert
  and jit_check_app_addr_and_convert
- wasm_exec_env_set_aux_stack
- wasm_exec_env_get_aux_stack
- wasm_cluster_create_thread
- wasm_cluster_allocate_aux_stack
- wasm_cluster_free_aux_stack

## Data structure changes
- WASMModule and AOTModule
  - field aux_data_end, aux_heap_base and aux_stack_bottom
- WASMExecEnv
  - field aux_stack_boundary and aux_stack_bottom
- AOTCompData
  - field aux_data_end, aux_heap_base and aux_stack_bottom
- WASMMemoryInstance(AOTMemoryInstance)
  - field memory_data_size and change __padding to is_memory64
- WASMModuleInstMemConsumption
  - field total_size and memories_size
- WASMDebugExecutionMemory
  - field start_offset and current_pos
- WASMCluster
  - field stack_tops

## Components that are affected by the APIs and data structure changes
- libc-builtin
- libc-emcc
- libc-uvwasi
- libc-wasi
- Python and Go Language Embedding
- Interpreter Debug engine
- Multi-thread: lib-pthread, wasi-threads and thread manager

PR #3218, opened 2024-03-12
API and data structure refactor 

TianlongLiang:
- remove parameter 'signature' from `wasm_runtime_lookup_function`
- move common data members `c_api_func_imports` and `cur_exec_env` from `WASMModuleInstanceExtraCommon` to `WASMModuleInstance`
- In `WASMModuleInstance` enlarge reserved[3] to reserved[5] in case add more data member in the future

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -47,15 +47,15 @@ bh_static_assert(offsetof(AOTModuleInstance, func_type_indexes)
                  == 6 * sizeof(uint64));
 bh_static_assert(offsetof(AOTModuleInstance, cur_exception)
                  == 13 * sizeof(uint64));
+bh_static_assert(offsetof(AOTModuleInstance, c_api_func_imports)
+                 == 13 * sizeof(uint64) + 128 + 8 * sizeof(uint64));
 bh_static_assert(offsetof(AOTModuleInstance, global_table_data)
-                 == 13 * sizeof(uint64) + 128 + 11 * sizeof(uint64));
+                 == 13 * sizeof(uint64) + 128 + 14 * sizeof(uint64));
 
 bh_static_assert(sizeof(AOTMemoryInstance) == 112);
 bh_static_assert(offsetof(AOTTableInstance, elems) == 24);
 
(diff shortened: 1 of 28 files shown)


TianlongLiang:
Hi, @yamt, I did some further refactoring according to the first bulletin mentioned in this issue https://github.com/bytecodealliance/wasm-micro-runtime/issues/2530. Could you please take a look at this PR at your convenience? By the way, I tested the performance on the 32-bit platform after refactoring. On benchmarks, including CoreMark, Dhrystone, and JetStream, there was basically no performance impact for JetStream, and the performance was roughly down by 1% for CoreMark and 2% for Dhrystone. I think generally speaking, we don't have to worry about the performance impact on the 32-bit platform

wenyongh (commented):
LGTM

wenyongh:
@yamt, I merged this PR (and also PR #3209) since we are to implement the memory64 feature and hope to refactor the related APIs and data structures before that, and also to make the related update for wamr-rust-sdk, Python language binding and Go language binding. @TianlongLiang have done a lot of test on the two PRs and also the code has been carefully reviewed. Please   let us know if there are issues found. Many thanks.

yamt:
this patch looks reasonable at a glance. but i have no time to make careful review this month.
please go ahead w/o waiting for me.

i guess ideally we can stop aot relying on known sizeof(WASMModuleInstance) so that all these
"extra" stuff can be merged into the "base" structure. how do you think?

wenyongh:
Yes, it may be another way that AOT doesn't rely on `sizeof(WASMModuleInstance)`, for example, wamrc puts the field offsets of `WASMModuleInstance` inside the AOT file and lets aot code read these fields according to the offsets. The issue is that it impacts the performance since an extra load instruction is added when accessing these fields, my preference is that performance is more important and we can reserve some spaces in WASMModuleInstance for future use, in fact there is a field `uint32 reserved[5]` in it, and we just found that field `used_to_be_wasi_ctx` is unused and can be merged to the `reserved` filed, so now the reserved space is 28 bytes: `uint32 reserved[7]`. Normally it should be enough, do you think we should enlarge it?

BTW, we found the global shared memory lock issue you mentioned in #3209, and plan to add a field in WASMMemoryInstance, `korp_mutex *memory_lock`. Is it good to you?

Issue #3208, opened 2024-03-07
Issue when running examples for ESP 32 with WAMR


TAOFOR4:
Hi,

I am trying to run the samples from the repo on esp c6. But some issues occur.

![Screenshot from 2024-03-07 20-00-37](https://github.com/bytecodealliance/wasm-micro-runtime/assets/114033626/02a91eed-691f-45c7-b6e2-d7bc8562c6a5)

TAOFOR4:
A little update. After I set some compiler options in cmake file it shows:
![Screenshot from 2024-03-07 23-06-55](https://github.com/bytecodealliance/wasm-micro-runtime/assets/114033626/8c33e490-a9cd-47c0-9472-13ef0ee01257)

wenyongh:
Hi, thanks for reporting the issue, could you try the below patch:
```diff
diff --git a/core/shared/platform/esp-idf/espidf_memmap.c b/core/shared/platform/esp-idf/espidf_memmap.c
index 9f3ec47a..6b1b6f04 100644
--- a/core/shared/platform/esp-idf/espidf_memmap.c
+++ b/core/shared/platform/esp-idf/espidf_memmap.c
@@ -59,6 +59,12 @@ os_mmap(void *hint, size_t size, int prot, int flags, os_file_handle file)
     }
 }

+void *
+os_mremap(void *old_addr, size_t old_size, size_t new_size)
+{
+    return os_mremap_slow(old_addr, old_size, new_size);
+}
+
 void
 os_munmap(void *addr, size_t size)
 {
diff --git a/core/shared/platform/esp-idf/platform_internal.h b/core/shared/platform/esp-idf/platform_internal.h
index 70c4fe7b..91a3f801 100644
--- a/core/shared/platform/esp-idf/platform_internal.h
+++ b/core/shared/platform/esp-idf/platform_internal.h
@@ -109,6 +109,8 @@ typedef unsigned int korp_sem;
 #define DT_LNK DTYPE_LINK
 #define DT_SOCK DTYPE_SOCK

+#define os_getpagesize getpagesize
+
 typedef int os_file_handle;
 typedef DIR *os_dir_stream;
 typedef int os_raw_file_handle;
```

TAOFOR4:
Hi, thx for the quick reply! I tried your patch. 
## For the example from WAMR repo it still complains. 
[...]

PR #3224, opened 2024-03-13
Fix compilation errors on esp-idf platform

wenyongh:
Reported in https://github.com/bytecodealliance/wasm-micro-runtime/issues/3208.


wenyongh:
Merge this PR since it resolves issue #3208.

PR #3234, opened 2024-03-15
Add esp32c6 support

TAOFOR4:
This commit adds support for ESP32 C6, which has been mentioned in #3208

Testing output is shown as 
Screenshot from 2024-03-15 17-55-23

### DIFF
--- a/build-scripts/esp-idf/wamr/CMakeLists.txt
+++ b/build-scripts/esp-idf/wamr/CMakeLists.txt
@@ -2,10 +2,10 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 # Set WAMR's build options
-if("${IDF_TARGET}" STREQUAL "esp32c3")
-    set(WAMR_BUILD_TARGET "RISCV32")
-else()
+if("${IDF_TARGET}" STREQUAL "esp32")
     set(WAMR_BUILD_TARGET "XTENSA")
+else()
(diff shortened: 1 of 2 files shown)


build-scripts/esp-idf/wamr/CMakeLists.txt
```diff
@@ -2,10 +2,10 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 # Set WAMR's build options
-if("${IDF_TARGET}" STREQUAL "esp32c3")
-    set(WAMR_BUILD_TARGET "RISCV32")
-else()
+if("${IDF_TARGET}" STREQUAL "esp32")
```
wenyongh, line 5: Using STREQUAL is invalid: STREQUAL checks whether the input `==esp32`, but not "==esp32c3 or ==esp32c6". And it sets the build target to xtensa then, but not riscv32.
How about:
```cmake
if ("${IDF_TARGET}" STREQUAL "esp32c3" OR "${IDF_TARGET}" STREQUAL "esp32c6")
    set(WAMR_BUILD_TARGET "RISCV32")
else ()
    set(WAMR_BUILD_TARGET "XTENSA")
endif ()
```

TAOFOR4: yeah, I forgot ESP32S3 is also XTENSA structure. I will fix it now. Thx. :)

wenyongh (approved):
LGTM

Issue #3220, opened 2024-03-12
Error building on ./build.sh for go binding


SeaRoll:
Steps to reproduce:

1. Download git folder
2. Navigate to go binding folder
3. run `/build.sh`

The terminal output:

```sh
-- Build Configurations:
     Build as target AARCH64
     CMAKE_BUILD_TYPE Release
     WAMR Interpreter enabled
     WAMR AOT enabled
     WAMR Fast JIT disabled
     WAMR LLVM ORC JIT disabled
     Libc builtin enabled
     Libc WASI enabled
     Fast interpreter enabled
     Multiple modules disabled
     Bulk memory feature enabled
     Shared memory enabled
     Thread manager enabled
     Lib pthread enabled
     Wakeup of blocking operations enabled
     SIMD enabled
     Memory profiling enabled
     Dump call stack enabled
     Reference types disabled
     GC performance profiling disabled
     Module instance context enabled
     Quick AOT/JIT entries enabled
-- Configuring done (0.0s)
-- Generating done (0.0s)
-- Build files have been written to: /Users/juyohan/Downloads/wasm-micro-runtime-main/language-bindings/go/build
[ 97%] Built target vmlib
[ 97%] Built target libiwasm
[100%] Built target iwasm
# github.com/bytecodealliance/wasm-micro-runtime/language-bindings/go/wamr [github.com/bytecodealliance/wasm-micro-runtime/language-bindings/go/wamr.test]
./module.go:119:64: cannot use _Ctype_int(stdinfd) (value of type _Ctype_int) as _Ctype_longlong value in variable declaration
./module.go:120:42: cannot use _Ctype_int(stdoutfd) (value of type _Ctype_int) as _Ctype_longlong value in variable declaration
./module.go:120:58: cannot use _Ctype_int(stderrfd) (value of type _Ctype_int) as _Ctype_longlong value in variable declaration
FAIL	github.com/bytecodealliance/wasm-micro-runtime/language-bindings/go/wamr [build failed]
```

TianlongLiang:
Hi, thanks for letting us know, I will submit a patch to fix it. You can simply change the 120-121 line in `language-bindings/go/wamr/module.go`:
from
```golang
                                    C.int(stdinfd), C.int(stdoutfd),
                                    C.int(stderrfd))
```
to
```golang
                                    C.long(stdinfd), C.long(stdoutfd),
                                    C.long(stderrfd))
```

[...]

PR #3235, opened 2024-03-16
Change from `long` to `longlong`

SeaRoll:
- go binding: type error(https://github.com/bytecodealliance/wasm-micro-runtime/issues/3220)
- change from long to longlong due to error:
```sh
./module.go:119:64: cannot use _Ctype_long(stdinfd) (value of type _Ctype_long) as _Ctype_longlong value in variable declaration
./module.go:120:43: cannot use _Ctype_long(stdoutfd) (value of type _Ctype_long) as _Ctype_longlong value in variable declaration
./module.go:120:60: cannot use _Ctype_long(stderrfd) (value of type _Ctype_long) as _Ctype_longlong value in variable declaration
```
- change from `uint32` to `uint64` due to casting error

### DIFF
--- a/language-bindings/go/samples/test.go
+++ b/language-bindings/go/samples/test.go
@@ -87,7 +87,7 @@ func main() {
     var instance *wamr.Instance
     var argv []uint32
     var results []interface{}
-    var offset uint32
+    var offset uint64
     var native_addr *uint8
     var err error
 
--- a/language-bindings/go/wamr/module.go
+++ b/language-bindings/go/wamr/module.go
@@ -117,8 +117,8 @@ func (self *Module) SetWasiArgsEx(dirList [][]byte, mapDirList [][]byte,
     C.wasm_runtime_set_wasi_args_ex(self.module, dirPtr, dirCount,
                                     mapDirPtr, mapDirCount,
                                     envPtr, envCount, argvPtr, argc,
-                                    C.longlong(stdinfd), C.longlong(stdoutfd),
-                                    C.longlong(stderrfd))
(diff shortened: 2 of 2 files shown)


TianlongLiang:
For this fix:
- change from long to longlong due to error:
It will error in my local environment
```shell
./module.go:119:64: cannot use _Ctype_longlong(stdinfd) (type _Ctype_longlong) as type _Ctype_long in assignment
./module.go:119:64: cannot use _Ctype_longlong(stdoutfd) (type _Ctype_longlong) as type _Ctype_long in assignment
./module.go:119:64: cannot use _Ctype_longlong(stderrfd) (type _Ctype_longlong) as type _Ctype_long in assignment
```
I think it better to make your modification a git patch and conditionally patch them based on `uname -a` output we could obtain in `./build.sh`

PS: Here is my output for `uname -a` output:
```shell
Linux TL-Work-PC 5.15.0-94-generic #104~20.04.1-Ubuntu SMP Tue Jan 16 13:34:09 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
```

SeaRoll:
hmm.. I am using an m1 mac, maybe its different?

TianlongLiang:
I think it's indeed a platform-dependent problem, so we'd better conditionally patch it based on the platform

SeaRoll:
sounds like an idea

wenyongh:
@SeaRoll Thanks for reporting the issue and submitting PR to fix it, as @TianlongLiang mentioned and tested, using `C.int64_t` should be good, I manually changed your commit file and merged this PR. Please let us know if there is issue found.

Issue #3238, opened 2024-03-18
Allow defining the ANDROID_ABI and WAMR_BUILD_TARGET for Android building


daniel7grant:
For our use-case, we need to build WAMR for multiple Android platforms, which means that we have to change the ANDROID_ABI and WAMR_BUILD_TARGET variables. Currently these can only be set by modifying `CMakeLists.txt` in `product-mini/platforms/android`. This is harder to do when it comes to automatic building (CI), parallelization, etc.

I'd make two modification to be able to set these parameters with `-D` defines:

```cmake
if (NOT DEFINED ANDROID_ABI)
    set (ANDROID_ABI "x86")
endif ()
```

and:

```cmake
if (NOT DEFINED WAMR_BUILD_TARGET)
    set (WAMR_BUILD_TARGET    "X86_32")
endif ()
```

With these changes, the cross compile could be called without ever changing the files:

```bash
cmake -DANDROID_ABI=armeabi-v7a -DWAMR_BUILD_TARGET=ARMV7A
```

What do you think? I could create a Pull Request for this and the documentation update.

wenyongh:
Hi, yes, it is not a good way to hardcode the configurations in the cmake file, I remember it was written that way due to the GitHub CI which only supports android 32-bit. I submitted PR #3239 to rewrite the cmake file and CI file, could you help have a view? Thanks.

daniel7grant:
Thanks for your work, I tried and it seems good to me.

Simultaneously I started working on this problem (https://github.com/bytecodealliance/wasm-micro-runtime/compare/main...daniel7grant:wasm-micro-runtime:feature/android-build-options), but your `CMakeLists.txt` seems to be more useful. The only thing I would add from my branch is the documentation: to add a note in the `platform-mini/README.md` that the flags should be used instead of modifying the `CMakeLists.txt` file.

wenyongh:
OK, I updated the document according to you branch and made small changes. Please check whether more need to be added or changed?

daniel7grant:
I don't see anything else. Thank you for your help.

wenyongh:
[...]

PR #3239, opened 2024-03-19
Make android platform's cmake flags configurable

wenyongh:
Don't hardcode the cmake configurations in the Android platform's CMakeLists.txt.


wenyongh:
Merge this PR since it was discussed and tested in https://github.com/bytecodealliance/wasm-micro-runtime/issues/3238.

PR #3240, opened 2024-03-19
Dev/memory64

TianlongLiang:
# Basic Memory64 support for classic-interpreter
Adding a new CMake flag(cache variable) `WAMR_BUILD_MEMORY64` to enable memory64 feature, it can only be enabled on the 64-bit platform/target and can only use software boundary check. And when it is enabled, it can support both i32 and i64 linear memory. The main modifications are:

- wasm loader && mini-loader: loading and bytecode validating process 
- wasm runtime: memory instantiating process
- classic-interpreter: wasm code executing process
- memory64 classic-interpreter spec test in `test_wamr.sh` and in CI

Currently, it will support memory64 memory wasm files that only use core spec(including bulk memory proposal) opcodes. 

Future TODOs when memory64 is enabled: 
1. support threads opcodes in classic-interpreter
2. support memory64 memory in related runtime API


core/iwasm/interpreter/wasm_mini_loader.c
```diff
@@ -3587,8 +3670,8 @@ wasm_loader_find_block_addr(WASMExecEnv *exec_env, BlockAddr *block_addr_cache,
             case WASM_OP_I64_STORE8:
             case WASM_OP_I64_STORE16:
             case WASM_OP_I64_STORE32:
-                skip_leb_uint32(p, p_end); /* align */
-                skip_leb_uint32(p, p_end); /* offset */
+                skip_leb_uint32(p, p_end);     /* align */
+                skip_leb_mem_offset(p, p_end); /* offset */
```
wenyongh, line 3674: Don't forget to apply similar changes to atomic/simd memory related opcodes in the future.

tests/wamr-test-suites/test_wamr.sh
```diff
@@ -488,7 +517,7 @@ function spec_test()
 
     local ARGS_FOR_SPEC_TEST=""
 
-    # multi-module only enable in interp mode
+    # multi-module only enable in interp mode and aot moe
```
wenyongh, line 520: mode

tests/wamr-test-suites/test_wamr.sh
```diff
@@ -537,6 +566,13 @@ function spec_test()
         ARGS_FOR_SPEC_TEST+="--gc "
     fi
 
+    # wasm64(memory64) only enable in interp and aot mode
```
wenyongh, line 569: had better `is only enabled`

core/iwasm/interpreter/wasm.h
```diff
@@ -484,6 +491,12 @@ typedef struct WASMTable {
 #endif
 } WASMTable;
 
+#if WASM_ENABLE_MEMORY64 != 0
+typedef uint64 linear_mem_ptr_t;
+#else
+typedef uint32 linear_mem_ptr_t;
```
wenyongh, line 497: Not sure whether `linear_mem_offset_t` is better? Some code in loader and classic interpreter uses string `mem_offset` and `MEM_OFFSET`.

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -4087,8 +4160,15 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                 bh_assert(global_idx < module->e->global_count);
                 global = globals + global_idx;
                 global_addr = get_global_addr(global_data, global);
-                /* TODO: Memory64 the data type depends on mem idx type */
-                aux_stack_top = (uint64)(*(uint32 *)(frame_sp - 1));
+#if WASM_ENABLE_MEMORY64 != 0
+                if (module->memories[0]->is_memory64) {
```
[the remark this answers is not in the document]
TianlongLiang: I thought it would be easier for future multi-memory access if the index is explicitly indicated

core/iwasm/interpreter/wasm.h
TianlongLiang: I forgot to replace it before PR, now replace it with `mem_offset_t`

core/iwasm/interpreter/wasm_interp_classic.c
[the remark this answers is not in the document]
wenyongh: OK, got it, thanks.

build-scripts/config_common.cmake
```diff
@@ -248,6 +248,17 @@ if (WAMR_BUILD_SHARED_MEMORY EQUAL 1)
 else ()
   add_definitions (-DWASM_ENABLE_SHARED_MEMORY=0)
 endif ()
+if (WAMR_BUILD_MEMORY64 EQUAL 1)
+  # if native is 32-bit or cross-compiled to 32-bit
+  if (NOT (CMAKE_SIZEOF_VOID_P EQUAL 8 OR WAMR_BUILD_TARGET MATCHES ".*64.*"))
```
wenyongh, line 253: This causes error not reported in Ubuntu x86-64 when building iwasm with `cmake .. -DWAMR_BUILD_TARGET=X86_32 -DWAMR_BUILD_MEMORY64=1`, using
`if (NOT WAMR_BUILD_TARGET MATCHES ".*64.*")` should be already good, since WAMR_BUILD_TARGET must have been set here (if it isn't set by developer, it will be set at the beginning of this file) and all 64-bit targets have "64" sub string (X86_64, AMD_64, AARCH64 and RISCV64).

build-scripts/config_common.cmake
```diff
@@ -248,6 +248,17 @@ if (WAMR_BUILD_SHARED_MEMORY EQUAL 1)
...
+    message (FATAL_ERROR "-- Memory64 is only available on the 64-bit platform/target")
+  endif()
+  add_definitions (-DWASM_ENABLE_MEMORY64=1)
+  set (WAMR_DISABLE_HW_BOUND_CHECK 1)
+  message ("     Memory64 memory enabled")
+else ()
+  add_definitions (-DWASM_ENABLE_MEMORY64=0)
```
wenyongh, line 260: No need to define the macro, it is set to 0 by default in core/config.h.

wenyongh (approved):
LGTM

Issue #3222, opened 2024-03-13
Nightly run CI flaky cases


TianlongLiang:
Some cases are Flaky and random failing, need to investigate and find a way to fix them if possible

- test cases with `tasn`: due to ASLR, will sometimes have those errors:
  ```
  'FATAL: ThreadSanitizer: unexpected memory mapping 0x62c84d7f2000-0x62c84d7fb000
  ```
- `nofollow_errors` in wasi test suite, now sure why
  ```shell
  Test nofollow_errors failed
    [exit_code] 0 == 1
  STDOUT:
  Exception: unreachable

  STDERR:
  thread 'main' panicked at 'expected errno NOTDIR; got LOOP', src/bin/nofollow_errors.rs:70:5
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
  ```

- maybe sync issue in Python script?
  ```shell
  Traceback (most recent call last):
    File "/home/runner/work/wasm-micro-runtime/wasm-micro-runtime/tests/wamr-test-suites/workspace/./all.py", 
  line 606, in <module>
      sys.exit(0 if main() else 1)
    File "/home/runner/work/wasm-micro-runtime/wasm-micro-runtime/tests/wamr-test-suites/workspace/./all.py", 
  line 554, in main
      ret = test_suite(
    File "/home/runner/work/wasm-micro-runtime/wasm-micro-runtime/tests/wamr-test-suites/workspace/./all.py", 
  line 375, in test_suite
      if not result.successful():
    File "/usr/lib/python3.10/multiprocessing/pool.py", line 761, in successful
      raise ValueError("{0!r} not ready".format(self))
  ValueError: <multiprocessing.pool.ApplyResult object at 0x7ebe85463e20> not ready
  ```

wenyongh:
The ASLR issue was fixed with PR #3233.

TianlongLiang:
`nofollow_errors`: Before https://github.com/bytecodealliance/wasm-micro-runtime/pull/3244 , the fstatat ret will always be -1(failing), so the file stat is uninitialized and has a random value, which will lead to the random true or false value of `S_ISLNK(sb.st_mode)`, causing this test case failing sometimes

TianlongLiang:
`maybe sync issue in Python script?` this should be a CI resource limit thing, didn't reproduce in my local environment, close this issue for now

PR #3244, opened 2024-03-20
Correct the dirfd argument value that passes to fstatat

TianlongLiang:
```C
// dirfd: This is the file descriptor of the directory relative to which the pathname is interpreted. 
int openat(int dirfd, const char *pathname, int flags, ...);
```
The value should be directory handle instead of file handle(which is always -1 in this context) return from `openat` 

This PR will fix the random failing test case `nofollow_errors` mentioned in https://github.com/bytecodealliance/wasm-micro-runtime/issues/3222


wenyongh:
@zoraaver this PR helps fix an error that is often reported by CI, could you help have a look? Thanks.

wenyongh (commented):
LGTM

PR #3249, opened 2024-03-21
append `\0` to every name string in aot name section

lum1n0us:
Since strings in .name section in .wasm is not c-style, need to append a `\0` to strings in .name section in .aot when emitting.

### DIFF
--- a/core/iwasm/compilation/aot_emit_aot_file.c
+++ b/core/iwasm/compilation/aot_emit_aot_file.c
@@ -1485,9 +1485,16 @@ fail_integer_too_large:
         res = (uint32)res64;                                   \
     } while (0)
 
+/*
+ * - transfer .name section in .wasm (comp_data->name_section_buf) to
+ *   aot buf (comp_data->aot_name_section_buf)
+ * - leb128 to u32
+ * - add `\0` at the end of every name, and adjust length(+1)
+ */
 static uint32
 get_name_section_size(AOTCompData *comp_data)
 {
+    /* original name section content in .wasm */
     const uint8 *p = comp_data->name_section_buf,
                 *p_end = comp_data->name_section_buf_end;
     uint8 *buf, *buf_end;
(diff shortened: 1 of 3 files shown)


core/iwasm/compilation/aot_emit_aot_file.c
```diff
@@ -1581,8 +1588,10 @@ get_name_section_size(AOTCompData *comp_data)
         aot_set_last_error("allocate memory for custom name section failed.");
         return 0;
     }
+    memset(buf, 0, (uint32)max_aot_buf_size);
     buf_end = buf + max_aot_buf_size;
 
+    /* the size of "name". it should be 4*/
```
wenyongh, line 1594: Had better add a space: `be 4 */`

core/iwasm/compilation/aot_emit_aot_file.c
```diff
@@ -1592,16 +1601,12 @@ get_name_section_size(AOTCompData *comp_data)
         return 0;
     }
 
-    if (!check_utf8_str(p, name_len)) {
-        aot_set_last_error("invalid UTF-8 encoding");
-        return 0;
-    }
-
+    /* "name" */
```
wenyongh, line 1604: Should we check whether `name_len == 4`? Suppose name_len == 3, the below memcmp may overread.
And it seems that loader also doesn't check it:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_loader.c#L5058

core/iwasm/compilation/aot_emit_aot_file.c
```diff
@@ -1645,9 +1651,11 @@ get_name_section_size(AOTCompData *comp_data)
                         previous_func_index = func_index;
                         read_leb_uint32(p, p_end, func_name_len);
                         offset = align_uint(offset, 2);
-                        EMIT_U16(func_name_len);
+                        /* extra 1 bytes for \0 */
```
wenyongh, line 1654: 1 byte

core/iwasm/compilation/aot_emit_aot_file.c
```diff
@@ -1645,9 +1651,11 @@ get_name_section_size(AOTCompData *comp_data)
...
+                        EMIT_U16(func_name_len + 1);
```
wenyongh, line 1655: Had better check whether `func_name_len < UINT16_MAX`, and use `EMIT_U16((uint16)func_name_len + 1)`? Here uint32 is truncated into uint16.

core/iwasm/compilation/aot_emit_aot_file.c
```diff
@@ -1633,6 +1638,7 @@ get_name_section_size(AOTCompData *comp_data)
                         read_leb_uint32(p, p_end, func_index);
                         offset = align_uint(offset, 4);
                         EMIT_U32(func_index);
+                        /* TODO: may not necessary to check function indexes */
```
wenyongh, line 1641: I remember the same checks in loader were added by Alibaba developers, I think it should be necessary, at least in loader.

core/iwasm/compilation/aot_emit_aot_file.c
lum1n0us: roll back following `EMIT_BUF()`

core/iwasm/compilation/aot_emit_aot_file.c
lum1n0us: `EMIT_U16()` will truncate the value.

Not sure about the check. Since we are behind loader, and it is an valid .wasm. If `func_name_len >= UINT16_MAX`, it will be a to big wasm file to load.

wenyongh (approved):

PR #3254, opened 2024-03-25
impl apis to set and get module name

lum1n0us:


core/iwasm/aot/aot_runtime.c
```diff
@@ -9,6 +9,7 @@
 #include "../common/wasm_runtime_common.h"
 #include "../common/wasm_memory.h"
 #include "../interpreter/wasm_runtime.h"
+#include "platform_common.h"
```
wenyongh, line 12: Please include `bh_platform.h` instead, in the original design, bh_platform.h includes all necessary header files for `core/shared/utils` and `core/shared/platform`, and other component should include it and don't include "platform_xxx.h".

core/iwasm/aot/aot_runtime.c
```diff
@@ -4663,3 +4664,70 @@ aot_traverse_gc_rootset(WASMExecEnv *exec_env, void *heap)
     return true;
 }
 #endif /* end of WASM_ENABLE_GC != 0 */
+
+char *
+const_str_set_insert(const uint8 *str, int32 len, AOTModule *module,
```
wenyongh, line 4669: How about renaming to `aot_const_str_set_insert`, there is another `const_str_list_insert` function in wasm_runtime.c, it may make confusing.

core/iwasm/aot/aot_runtime.c
```diff
@@ -4663,3 +4664,70 @@ aot_traverse_gc_rootset(WASMExecEnv *exec_env, void *heap)
...
+}
+
+bool
+aot_set_module_name(AOTModule *module, const char *name, char *error_buf,
+                    uint32_t error_buf_size)
+{
+    if (!name)
+        return false;
+
+    module->name = const_str_set_insert((const uint8 *)name, strlen(name),
```
wenyongh, line 4724: Had better use `strlen(name)` + 1, note that in the above `const_str_set_insert`, it only `runtime_malloc((uint32)len, error_buf, error_buf_size)`, and `\0` isn't appended.

core/iwasm/aot/aot_runtime.h
```diff
@@ -761,6 +762,20 @@ bool
 aot_traverse_gc_rootset(WASMExecEnv *exec_env, void *heap);
 #endif /* end of WASM_ENABLE_GC != 0 */
 
+char *
+const_str_set_insert(const uint8 *str, int32 len, AOTModule *module,
```
wenyongh, line 766: Had better `aot_const_str_set_insert`

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -6067,6 +5935,8 @@ create_module(char *error_buf, uint32 error_buf_size)
     /* Set start_function to -1, means no start function */
     module->start_function = (uint32)-1;
 
+    module->name = "";
```
wenyongh, line 5938: Should also apply same change in wasm_mini_loader.c

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -4400,3 +4400,154 @@ wasm_propagate_wasi_args(WASMModule *module)
     }
 }
 #endif
+
+bool
+check_utf8_str(const uint8 *str, uint32 len)
```
wenyongh, line 4405: Had better rename to `wasm_check_utf8_str`

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -4400,3 +4400,154 @@ wasm_propagate_wasi_args(WASMModule *module)
...
+        }
+        else {
+            return false;
+        }
+    }
+    return (p == p_end);
+}
+
+char *
+const_str_list_insert(const uint8 *str, uint32 len, WASMModule *module,
```
wenyongh, line 4476: Had better rename to `wasm_const_str_list_insert`

core/iwasm/aot/aot_runtime.c
lum1n0us: Thanks for reminding. Because there are same concern in the first callers in *aot_loader.c*, I will add protections in `aot_const_str_set_insert()`

wenyongh (commented):
LGTM

Issue #3252, opened 2024-03-22
zephyr:  The `simple` sample doesn't run on `nucleo-h563zi`


lucasAbadFr:
Hello, I'm trying to build the `product-mini/platforms/zephyr/simple` example for the `nucleo_h563zi` board. But I'm unable to make it work. Any help is appreciated. Thanks.

I'm on Zephyr 3.6.0 even if it's currently unsupported by the WAMR, the `simple` example should work.

My build command is:
```bash
ZEPHYR_BASE=~/zephyrproject/zephyr west build .  -b nucleo_h563zi -p always -- -DWAMR_BUILD_TARGET=THUMBV8
```
I created a new board configuration for the `nucleo_h563zi` board. I used the `nucleo_h563zi` zephyr default board configuration as a reference:
```conf
# Default Zephyr config: 
# enable uart driver
CONFIG_SERIAL=y

# console
CONFIG_CONSOLE=y
CONFIG_UART_CONSOLE=y

# Enable clock
CONFIG_CLOCK_CONTROL=y

# Enable MPU
CONFIG_ARM_MPU=y

# Enable HW stack protection
CONFIG_HW_STACK_PROTECTION=y

# enable GPIO
CONFIG_GPIO=y
# enable pin controller
CONFIG_PINCTRL=y
```

1. **MPU error**
    * **Error output is:**
    ```bash
    FAILED: #Skip long message
    /home/user/wasm-micro-runtime/core/shared/platform/zephyr/zephyr_platform.c: In function 'disable_mpu_rasr_xn':
    /home/user/wasm-micro-runtime/core/shared/platform/zephyr/zephyr_platform.c:28:16: error: 'MPU_Type' has no member named 'RASR'
    28 |         if (MPU->RASR & MPU_RASR_XN_Msk) {
        |                ^~
    /home/user/wasm-micro-runtime/core/shared/platform/zephyr/zephyr_platform.c:28:25: error: 'MPU_RASR_XN_Msk' undeclared (first use in this function); did you mean 'MPU_RBAR_XN_Msk'?
    28 |         if (MPU->RASR & MPU_RASR_XN_Msk) {
        |                         ^~~~~~~~~~~~~~~
        |                         MPU_RBAR_XN_Msk
    /home/user/wasm-micro-runtime/core/shared/platform/zephyr/zephyr_platform.c:28:25: note: each undeclared identifier is reported only once for each function it appears in
[...]

PR #3255, opened 2024-03-26
Fix compilation errors on zephyr platform

wenyongh:
Refer to https://github.com/bytecodealliance/wasm-micro-runtime/issues/3252


wenyongh:
Merge this PR since it was discussed in #3252 and it resolves issues mentioned in that issue.

Issue #3210, opened 2024-03-08
Segmentation Fault in wasm_interp_call_func_bytecode when running a valid Wasm module


khagankhan:
## Describe the Bug

Running a valid test case on `iwasm` triggers a segmentation fault in `wasm_interp_call_func_bytecode` due to an invalid memory access.

`iwasm --version`: iwasm 1.3.2

`uname -a`:#62-Ubuntu SMP Tue Nov 22 19:54:14 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

## Steps to reproduce:

1. Save the `crash.wat` (Remove .txt extension)
2. Install and build WABT toolkit (`wat2wasm` from WABT can be used to convert `crash.wat` into `crash.wasm` using: `wat2wasm crash.wat -o crash.wasm`. I have also added `crash.wasm` that obviates the need to use `wat2wasm`)
3. Run `/path/to/iwasm crash.wasm`
4. Observe the `abort`.

## Files that cause the `abort`:
[crash.wat.txt](https://github.com/bytecodealliance/wasm-micro-runtime/files/14533620/crash.wat.txt)
[crash.wasm.txt](https://github.com/bytecodealliance/wasm-micro-runtime/files/14533621/crash.wasm.txt)

## Observed Behavior:
```
iwasm crash.wasm 

unhandled SIGSEGV, si_addr: 0x56473f5fa518
Aborted
```

## Expected behavior:
`iwasm` is supposed to run the module successfully. 

## GDB Backtrace

```
Program received signal SIGSEGV, Segmentation fault.
0x00005555555f0bef in wasm_interp_call_func_bytecode (module=0x5555558a4da0, exec_env=0x0, exec_env@entry=0x5555558a53e0, cur_func=0x5555558a3fb5, prev_frame=0xfffffffe, prev_frame@entry=0x5555558a5478) at /users/khan22/wasmoi/targets/wasm-micro-runtime/core/iwasm/interpreter/wasm_interp_fast.c:1658
1658	                        prev_frame->lp[ret_offset] =
(gdb) bt
#0  0x00005555555f0bef in wasm_interp_call_func_bytecode (module=0x5555558a4da0, exec_env=0x0, exec_env@entry=0x5555558a53e0, cur_func=0x5555558a3fb5, prev_frame=0xfffffffe, prev_frame@entry=0x5555558a5478)
[...]

PR #3259, opened 2024-03-27
loader: Remove updating ctx->dynamic_offset in emit_br_info

wenyongh:
Should not update `ctx->dynamic_offset` in emit_br_info, since the `Part e` only
sets the dst offsets, the operand stack should not be changed, e.g., the stack
operands are to be used by the opcodes followed by `br_if` opcode.

Reported in https://github.com/bytecodealliance/wasm-micro-runtime/issues/3210.

### DIFF
--- a/core/iwasm/interpreter/wasm_loader.c
+++ b/core/iwasm/interpreter/wasm_loader.c
@@ -8912,21 +8912,6 @@ wasm_loader_emit_br_info(WASMLoaderContext *ctx, BranchBlock *frame_csp,
             dynamic_offset -= cell;
             emit_operand(ctx, dynamic_offset);
         }
-        /* Update ctx->max_dynamic_offset */
-        if (dynamic_offset > ctx->max_dynamic_offset)
-            ctx->max_dynamic_offset = dynamic_offset;
-        /* Update ctx->dynamic_offset */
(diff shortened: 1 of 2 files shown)


core/iwasm/interpreter/wasm_loader.c
```diff
@@ -8907,12 +8907,24 @@ wasm_loader_emit_br_info(WASMLoaderContext *ctx, BranchBlock *frame_csp,
         /* Part e */
         dynamic_offset =
             frame_csp->dynamic_offset + wasm_get_cell_num(types, arity);
-        ctx->dynamic_offset = dynamic_offset;
         for (i = (int32)arity - 1; i >= 0; i--) {
             cell = (uint8)wasm_value_type_cell_num(types[i]);
             dynamic_offset -= cell;
             emit_operand(ctx, dynamic_offset);
         }
+        /* Update ctx->dynamic_offset and ctx->frame_offset */
```
xujuntwt95329, line 8915: Should we update `ctx->max_dynamic_offset` before this? Otherwise if `frame_csp->dynamic_offset + wasm_get_cell_num(types, arity)` is larger than current `ctx->max_dynamic_offset`, some value may be dropped during the below for loop?

xujuntwt95329:
LGTM

PR #3263, opened 2024-03-28
Add lock and ref_count for runtime init

wenyongh:
ps. https://github.com/bytecodealliance/wasm-micro-runtime/discussions/3253


lum1n0us (approved):
LGTM

PR #3264, opened 2024-03-28
Add wamr to esp-idf components registry.

donghengqaz:
Espressif has [ESP-IDF component registry](https://components.espressif.com/) to manage external components, and developers can use this directly and easily.

I plan to add `wasm-micro-runtime` to this registry, but it seems that wasm-micro-runtime team will not plan to cost much resource to maintain ESP-IDF platform and its related things, so I discuss with some internal Espressif developers. We hope you approve to let us publish WAMR into this registry with specific version(not WAMR official version). If some patches need to add for ESP-IDF platform, I will raised PR here, and when this PR is merged, I publish new version to registry.

If you want to register account in this registry, please add me into your namespace with necessary access permissions, otherwise we plan to use our own account to publish WAMR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,11 @@
 
 cmake_minimum_required (VERSION 3.0)
 
+if(ESP_PLATFORM)
+  include (${COMPONENT_DIR}/build-scripts/esp-idf/wamr/CMakeLists.txt)
+  return()
+endif()
+
 project (iwasm)
 
 set (CMAKE_VERBOSE_MAKEFILE OFF)
--- a/idf_component.yml
+++ b/idf_component.yml
@@ -1,0 +1,8 @@
+version: "1.3.2"
+description: WebAssembly Micro Runtime - A lightweight standalone WebAssembly (Wasm) runtime with small footprint, high performance and highly configurable features
+url: https://bytecodealliance.org/
+repository: https://github.com/bytecodealliance/wasm-micro-runtime.git
+documentation: https://wamr.gitbook.io/
+issues: https://github.com/bytecodealliance/wasm-micro-runtime/issues
+dependencies:
+  idf: ">=4.4"
--- a/core/shared/platform/esp-idf/espidf_memmap.c
+++ b/core/shared/platform/esp-idf/espidf_memmap.c
@@ -55,7 +55,23 @@ os_mmap(void *hint, size_t size, int prot, int flags, os_file_handle file)
 #else
         uint32_t mem_caps = MALLOC_CAP_8BIT;
 #endif
-        return heap_caps_malloc(size, mem_caps);
+        void *buf_origin = heap_caps_malloc(size, mem_caps);
+        if (!buf_origin) {
+            return NULL;
+        }
+
+        // Memory allocation with MALLOC_CAP_SPIRAM or MALLOC_CAP_8BIT will
+        // return 4-byte aligned Reserve extra 4 byte to fixup alignment and
+        // size for the pointer to the originally allocated address
+        void *buf_fixed = buf_origin + sizeof(void *);
+        if ((uintptr_t)buf_fixed & (uintptr_t)0x7) {
+            buf_fixed = (void *)((uintptr_t)(buf_fixed + 4) & (~(uintptr_t)7));
+        }
+
+        uintptr_t *addr_field = buf_fixed - sizeof(uintptr_t);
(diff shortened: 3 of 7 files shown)


idf_component.yml
```diff
@@ -0,0 +1,8 @@
+version: "0.1.0"
```
wenyongh, line 1: Must this file be put under the root directory? There are too many files under root dir now.

CMakeLists.txt
```diff
@@ -3,6 +3,7 @@
 
 cmake_minimum_required (VERSION 3.0)
 
+if(NOT ESP_PLATFORM)
```
wenyongh, line 6: How about:
```cmake
if (ESP_PLATFORM)
  include (${COMPONENT_DIR}/build-scripts/esp-idf/wamr/CMakeLists.txt)
  return ()
endif ()
```

idf_component.yml
donghengqaz: Yes, this is need to be under the root directory, but version can be WAMR version plus bugfix version, for example now WAMR version is 1.3.2, we can make this to be 1.3.2.x(x is bugfix version is needed).

CMakeLists.txt
donghengqaz: OK.

donghengqaz:
@wenyongh Today we have a meeting about this, we would like to use release branch of WAMR, because we may use this in real product, so master branch is not good choice, so I plan to merge this into release/v1.3.x version.

lum1n0us:
@donghengqaz here are some questions to make sure we are on the same page.

- Please be awarded of that 1.3.x branch will be out of maintenance in the end of this year(2024).
- I noticed this PR will be merged only in 1.3.x but not in both (main and 1.3.x). If we all hope a smooth upgrade from 1.3.x to 2.x in next year(2025) after termination of 1.3.x branch, it should be better to merge both and maintain both
- Since we are not sure you are the only customer depends on 1.3.x branch, there might be following PRs from them too. I guess a 1.3.x release is better than "the latest commit of 1.3.x branch". So, what's your opinion about make a new 1.3.x release recently？

donghengqaz:
Hi @lum1n0us, 

> I noticed this PR will be merged only in 1.3.x but not in both (main and 1.3.x). If we all hope a smooth upgrade from 1.3.x to 2.x in next year(2025) after termination of 1.3.x branch, it should be better to merge both and maintain both

You are right, and I will add another PR for main. At first we plan to publish version based on WAMR release branch, because some internal developers and customers need a stable version to start developing product software.

> Since we are not sure you are the only customer depends on 1.3.x branch, there might be following PRs from them too. I guess a 1.3.x release is better than "the latest commit of 1.3.x branch". So, what's your opinion about make a new 1.3.x release recently？

Yes, this is very well. I will continue to publish new version to registry, you can see WAMR version in registry as follows:
```
1.3.2-1
```
The `1.3.2` is WAMR tag version based on release version, and `-1` is bugfix version which is new patch 
merged into release branch. If recently WAMR v1.3.3 is released, I will publish it into registry with version `1.3.3-0`.

wenyongh:
@donghengqaz the CI runs failed since the wamr-ide CI issue isn't fixed yet (which is fixed in main branch), I submitted PR https://github.com/bytecodealliance/wasm-micro-runtime/pull/3268 to fix the issues that were fixed in main branch, and will merge your PR after that PR is merged.

wenyongh:
> @donghengqaz I have merged PR #3268, could you rebase it? BTW, could you submit another PR to main branch?

wenyongh (approved):

donghengqaz:
> > @donghengqaz I have merged PR #3268, could you rebase it? BTW, could you submit another PR to main branch?

OK. In the main branch, which version in file `idf_component.yml` can I use, is `1.3.3` OK?

wenyongh:
> > > @donghengqaz I have merged PR #3268, could you rebase it? BTW, could you submit another PR to main branch?
> 
> OK. In the main branch, which version in file `idf_component.yml` can I use, is `1.3.3` OK?

The next planned release version in main is 2.0.0, maybe you can use it in idf_component.yml?

donghengqaz:
> > > > @donghengqaz I have merged PR #3268, could you rebase it? BTW, could you submit another PR to main branch?
> > 
> > 
> > OK. In the main branch, which version in file `idf_component.yml` can I use, is `1.3.3` OK?
> 
> The next planned release version in main is 2.0.0, maybe you can use it in idf_component.yml?

OK.

PR #3265, opened 2024-03-28
Add module name into aot's perf map

lum1n0us:
Still missing the python script to filter every wasm's profiling data out


core/iwasm/aot/aot_perf_map.c
```diff
@@ -0,0 +1,119 @@
+/*
+ * Copyright (C) 2019 Intel Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+
+#include "aot_perf_map.h"
+#include "bh_log.h"
+#include <stdio.h>
```
wenyongh, line 8: Had better not include <stdio.h>, include "bh_platform.h" instead?

core/iwasm/aot/aot_perf_map.h
```diff
@@ -0,0 +1,15 @@
...
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+
+#ifndef _AOT_PERF_MAP_H_
+#define _AOT_PERF_MAP_H_
+
+#include "aot_runtime.h"
+
+bool
+create_perf_map(const AOTModule *module, char *error_buf,
```
wenyongh, line 12: How about renaming to `aot_create_perf_map`?

wenyongh (commented):
LGTM

Issue #3170, opened 2024-02-20
Wrong execution result in fast interpreter


erxiaozhou:
# Build commands
I compile the code with commit id 63cd567b3fe0bab8874da89d01f89fccbdf1fe28.
Platform: Ubuntu 20.04
CPU: amd64 
compile:

```
export CC=/usr/lib/llvm-16/bin/clang
export CXX=/usr/lib/llvm-16/bin/clang++
cd product-mini/platforms/linux/;rm -rf build
cmake  -DWAMR_BUILD_INTERP=1 -DWAMR_BUILD_REF_TYPES=1 -DWAMR_BUILD_SIMD=1 -DWASM_ENABLE_BULK_MEMORY=1 -DWAMR_BUILD_AOT=0 -DCMAKE_BUILD_TYPE=Release -DWAMR_BUILD_FAST_INTERP=1 -DWAMR_BUILD_LIBC_WASI=0 -DWAMR_BUILD_LIBC_BUILTIN=1 -Bbuild
```

execute:

```
iwasm --heap-size=0  <tc_name>
```
# Case

[iwasm_faster_interp_no_oom.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/14349614/iwasm_faster_interp_no_oom.zip)

## Code in WAT
```
(module
  (type (;0;) (func))
  (type (;1;) (func (result f64)))
  (type (;2;) (func (result i32)))
  (func (;0;) (type 2) (result i32)
    block (result i64)  ;; label = @1
      i64.const -7384628800742112618
      br 0 (;@1;)
    end
    i64.const 8296663890419202031
    i64.clz
    drop
    i32.wrap_i64
    i32.load8_s offset=32226
    ;; return
    )
  (memory (;0;) 1 5)
  (export "to_test" (func 0)))

```

# Actual output:

```
0
```

# Expected output:

```
Exception: out of bounds memory access
```

wenyongh:
@erxiaozhou thanks for spotting the issue, it was fixed in #3170.

PR #3269, opened 2024-04-01
Fix dynamic offset not updated in op_br for block with ret type

wenyongh:
Found the PR https://github.com/bytecodealliance/wasm-micro-runtime/pull/3259
reverted PR https://github.com/bytecodealliance/wasm-micro-runtime/pull/3192.

The issue is that we should update `ctx->dynamic_offset` only for opcode br and
should not update it for opcode br_if. This PR fixes both issue #3170 and #3210.


xujuntwt95329:
LGTM

PR #3270, opened 2024-04-01
Fix CI error when install packages for macos-14

wenyongh:
MacOS CI ran failed with "error: externally-managed-environment" reported
when installing dependencies. Add argument "--break-system-packages" to
fix it.

ps.
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/8503917001/job/23291537189
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/8502961539/job/23289867170


lum1n0us (approved):
LGTM

PR #3276, opened 2024-04-03
Small enhancement on addr2line.py

lum1n0us:
- if can't parse function name with dwarf information, use "name section" instead
- add introduction about using custom section


lum1n0us:
@eloparco please be aware of.

PR #3282, opened 2024-04-06
thread mgr: Free aux stack only when it was allocated

wenyongh:
When thread manager is enabled, the aux stack of exec_env may be allocated
by wasm_cluster_allocate_aux_stack or disabled by setting aux_stack_bottom
as UINTPTR_MAX directly. For the latter, no need to free it.

And fix an issue when paring `--gc-heap-size=n` argument for iwasm, and
fix a variable shadowed warning in fast-jit.


lum1n0us (approved):
LGTM

Issue #3279, opened 2024-04-03
wasm_interp_call_func_native currupts frame->ip, causes SIGSEGV with native tail-calls


SBKarr:
### Subject of the issue

`wasm_interp_call_func_native` [here](https://github.com/bytecodealliance/wasm-micro-runtime/blob/b4cab84e5f6006185f44db2c66fe389850e5f4b9/core/iwasm/interpreter/wasm_interp_fast.c#L5854) uses `prev_frame` as frame origin. Next allocated frame will share same memory space as `frame`  in interpreter.  Then, it's ip will be set to NULL [here](https://github.com/bytecodealliance/wasm-micro-runtime/blob/b4cab84e5f6006185f44db2c66fe389850e5f4b9/core/iwasm/interpreter/wasm_interp_fast.c#L1176). As a result, [UPDATE_ALL_FROM_FRAME](https://github.com/bytecodealliance/wasm-micro-runtime/blob/b4cab84e5f6006185f44db2c66fe389850e5f4b9/core/iwasm/interpreter/wasm_interp_fast.c#L5860C13-L5860C34) restores `frame_ip = NULL`, that causes crash on SIGSEGV on continue.

### Test case

In call of _initialize. Specifically, in `stappler_wasm_wasm_debug_print` line 43.
[module4.txt](https://github.com/bytecodealliance/wasm-micro-runtime/files/14853332/module4.txt)

### Your environment

* Ubuntu 22.04, x86_64, WASI SDK 21
* WAMR 1.3.2 + WAMR  from /release/1.3.x

### Steps to reproduce

- Build WAMR with tail-call support
- Define native functon like this:
```c
static void StapplerWasmDebugPrint(wasm_exec_env_t exec_env, uint32_t ptr, uint32_t size) {
	auto mod = wasm_runtime_get_module_inst(exec_env);
	auto sptr = (const char *)wasm_runtime_addr_app_to_native(mod, ptr);

	printf(sptr);
	printf("\n");
}

static NativeSymbol stapper_wasm_symbols[] = {
	NativeSymbol{"debug-print", (void *)&StapplerWasmDebugPrint, "(ii)", NULL},
};

static void registerWasmModule() {
	wasm_runtime_register_natives("stappler:wasm/wasm", stapper_wasm_symbols, 1);
}
```
- Run code from test case, call `_initialize`

### Expected behavior

Should print `TEST CTOR` from native code

### Actual behavior

`SIGSEGV`

[...]

PR #3283, opened 2024-04-06
interp: Don't restore data from frame after tail call native func

wenyongh:
Don't restore some variables from frame for tail call to native function,
since they weren't committed into the frame previously and are available
after tail calling to the native function.

Found in https://github.com/bytecodealliance/wasm-micro-runtime/issues/3279.


lum1n0us (approved):
LGTM

PR #3291, opened 2024-04-08
zephyr: pthread library: add missing functions

mkolchurin:
added `os_thread_detach`, `os_thread_exit`, `os_cond_broadcast` for use with `WAMR_BUILD_LIB_PTHREAD=1`

### DIFF
--- a/core/shared/platform/zephyr/zephyr_thread.c
+++ b/core/shared/platform/zephyr/zephyr_thread.c
@@ -578,3 +578,33 @@ os_thread_get_stack_boundary()
 void
 os_thread_jit_write_protect_np(bool enabled)
 {}
+
+int
+os_thread_detach(korp_tid thread)
+{
+    (void)thread;
+    return BHT_OK;
+}
+
+void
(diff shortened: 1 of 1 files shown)


core/shared/platform/zephyr/zephyr_thread.c
```diff
@@ -577,4 +577,35 @@ os_thread_get_stack_boundary()
...
+{
+    os_thread_wait_node *node;
+    k_mutex_lock(&cond->wait_list_lock, K_FOREVER);
+    node = cond->thread_wait_list;
+    while (node) {
+        os_thread_wait_node *next = node->next;
+        k_sem_give(&node->sem);
+        node = next;
+    }
+    cond->thread_wait_list = NULL;
```
wenyongh, line 608: Should not set thread_wait_list to NULL? The waiting thread will remove its wait_node from the list, and after each waiting thread removes the node, the thread_wait_list will be NULL, see os_cond_wait_internal.

wenyongh (approved):

PR #3315, opened 2024-04-12
Allow executing malloc/free from native in memory64 mode

loganek:


core/iwasm/interpreter/wasm_mini_loader.c
```diff
@@ -2829,8 +2849,8 @@ load_from_sections(WASMModule *module, WASMSection *sections,
                                 module->functions[func_index]->func_type;
                             if (func_type->param_count == 1
                                 && func_type->result_count == 1
-                                && func_type->types[0] == VALUE_TYPE_I32
-                                && func_type->types[1] == VALUE_TYPE_I32) {
+                                && func_type->types[0] == malloc_free_io_type
+                                && func_type->types[1] == malloc_free_io_type) {
```
wenyongh, line 2853: Should also change the param type for "free" function below, L2879, like in wasm_loader.c.

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -1422,8 +1431,8 @@ execute_malloc_function(WASMModuleInstance *module_inst, WASMExecEnv *exec_env,
         /* the malloc functino from assemblyscript is:
             function __new(size: usize, id: u32)
             id = 0 means this is an ArrayBuffer object */
-        argv[1] = 0;
-        argc = 2;
+        argv[argc] = 0;
+        argc++;
```
wenyongh, line 1435: In loader, the __new function's arguments for memory64 are treated as int64 type, here we should also add extra control for memory64?

wenyongh: And if that, should defined argv as argv[4]

TianlongLiang: I am not sure if the assemblyscript supports memory64 yet, according to https://github.com/AssemblyScript/assemblyscript/wiki/Status-and-Roadmap it's not supported yet, so maybe consider throwing a warning or error here when memory64 is enabled.

loganek: > In loader, the __new function's arguments for memory64 are treated as int64 type, here we should also add extra control for memory64?

I don't think we need to do any change. As far as I understand, we'll have one of those two scenarios:
* wasi libc - in that case, only `malloc(uint64) -> uint64` is called
* assemblyscript - for this, two functions are called: 1) `__new(uint64, uint32) -> uint64` and then `__pin(uint64) -> uint64`

Since we re-use the same array for __new and __pin (and __new leaves the last return value as first argument(s) in the argv so it can be used for __pin) I don't think we'll need to make any further changes.

> I am not sure if the assemblyscript supports memory64 yet, according to Wiki: Status and Roadmap (AssemblyScript/assemblyscript) it's not supported yet, so maybe consider throwing a warning or error here when memory64 is enabled.

Hmm, I'm personally a bit against it. If we do that, we'll need to keep track on assemblyscript roadmap, and it'd be easy to miss. At the same time, if memory64 is not supported yet by AS, users will unlikely ever have AS binary compiled to wasm64. If they do, they likely did some ugly hacks and they're on their own anyway.

wenyongh: Treating __new and __pin in asc as `1) __new(uint64, uint32) -> uint64 and then __pin(uint64) -> uint64` is good, but I think the type of the second argument of __new in loader/mini_loader is invalid: it uses VALUE_TYPE_I64 now.

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -5909,9 +5929,9 @@ load_from_sections(WASMModule *module, WASMSection *sections,
                 func_index = export->index - module->import_function_count;
                 func_type = module->functions[func_index]->func_type;
                 if (func_type->param_count == 2 && func_type->result_count == 1
-                    && func_type->types[0] == VALUE_TYPE_I32
-                    && func_type->types[1] == VALUE_TYPE_I32
-                    && func_type->types[2] == VALUE_TYPE_I32) {
+                    && func_type->types[0] == malloc_free_io_type
+                    && func_type->types[1] == malloc_free_io_type
```
wenyongh, line 5933: Should be VALUE_TYPE_I32?

core/iwasm/interpreter/wasm_mini_loader.c
```diff
@@ -2803,9 +2823,9 @@ load_from_sections(WASMModule *module, WASMSection *sections,
                 func_index = export->index - module->import_function_count;
                 func_type = module->functions[func_index]->func_type;
                 if (func_type->param_count == 2 && func_type->result_count == 1
-                    && func_type->types[0] == VALUE_TYPE_I32
-                    && func_type->types[1] == VALUE_TYPE_I32
-                    && func_type->types[2] == VALUE_TYPE_I32) {
+                    && func_type->types[0] == malloc_free_io_type
+                    && func_type->types[1] == malloc_free_io_type
```
wenyongh, line 2827: Should be VALUE_TYPE_I32?

core/iwasm/interpreter/wasm_runtime.c
loganek: you're right, updated.

wenyongh (approved):
LGTM

TianlongLiang:
LGTM

PR #3319, opened 2024-04-16
CodeQL CI refactor

TianlongLiang:
- move CodeQL scripts to the scripts directory
- only error in CI if it's a CodeQL reported issue was not dismissed before and likely to be an actual error


.github/scripts/codeql_fail_on_error.py
```diff
@@ -0,0 +1,119 @@
+#!/usr/bin/env python3
```
wenyongh, line 1: Had better add license header

wenyongh (approved):

PR #3323, opened 2024-04-17
Fix posix_fadvise error handling

### LINKED ISSUE #3322: `fd_advise` always succeeds
### Subject of the issue

Calling `fd_advise` always succeeds, even with very large offset, length, or a combination of them.

### Test case

```wat
(module
  (type (;0;) (func))
  (type (;1;) (func (param i32 i32 i32 i32 i32 i64 i64 i32 i32) (result i32)))
  (type (;2;) (func (param i32)))
  (type (;4;) (func (param i32 i64 i64 i32) (result i32)))
  (import "wasi_snapshot_preview1" "path_open" (func $path_open (type 1)))
  (import "wasi_snapshot_preview1" "proc_exit" (func $proc_exit (type 2)))
  (import "wasi_snapshot_preview1" "fd_advise" (func $fd_advise (type 3)))
  (func $_start (type 0)
    (local $err i32)

    (local.set $err
      (call $path_open
        (i32.const 3)
        (i32.const 0)
        (i32.const 1024)
        (i32.const 1)
        (i32.const 0)
        (i64.const 262143)
        (i64.const 262143)
        (i32.const 0)
        (i32.const 512)
      )
    )
    (local.get $err)
    (if
      (then (call $proc_exit (i32.const 1)))
    )

    (local.set $err
      (call $fd_advise
        (i32.load (i32.const 512))
        (i64.const 16942241767661461241)
        (i64.const 9765931924571875679)
        (i32.const 0)
      )
    )
    (local.get $err)
    (if
      (then (call $proc_exit (local.get $err)))
    )

    (call $proc_exit (i32.const 0))
  )
  (memory 2)
  (export "memory" (memory 0))
  (export "_start" (func $_start))
  (data (i32.const 1024) "a\00")
)
```

### Your environment

64-bit x86 Fedora Linux

WAMR is latest HEAD.

### Steps to reproduce

Use `wat2wasm` to encode the above WAT code and run it with `iwasm` mounting some directory:

```sh
$ wat2wasm advise.wat -o advise.wasm
$ iwasm --dir=. advise.wasm
$ echo $?
```

### Expected behavior

It should print some non-zero error code, say 28.

### Actual behavior

It prints 0.

### Extra Info

I will raise a PR that fixes this.

yagehu:
`posix_fadvise()` returns 0 on success and the errno on error.  This commit fixes the handling of the return value such that it does not always succeeds.

fixes #3322

### DIFF
--- a/core/shared/platform/common/posix/posix_file.c
+++ b/core/shared/platform/common/posix/posix_file.c
@@ -823,7 +823,7 @@ os_fadvise(os_file_handle handle, __wasi_filesize_t offset,
 
     int ret = posix_fadvise(handle, (off_t)offset, (off_t)length, nadvice);
 
-    if (ret < 0)
+    if (ret != 0)
         return convert_errno(ret);
 
     return __WASI_ESUCCESS;


yamt (approved):
lgtm

PR #3331, opened 2024-04-18
Link libc++ statically to reduce runtime dependency of wamrc

no1wudi:


wamr-compiler/CMakeLists.txt
```diff
@@ -387,6 +387,12 @@ if (NOT MSVC)
       target_link_libraries (wamrc ssp.a ws2_32)
   else()
       target_link_libraries (wamrc -ldl)
+      # Link libc++ statically to reduce the runtime dependency
+      target_link_libraries (wamrc -static-libstdc++)
+      # If not on macOS, link libgcc statically
+      if (NOT APPLE)
+        target_link_libraries (wamrc -static-libgcc)
+      endif()
```
yamt, line 395: which instance of these libraries are used by shared libraries loaded by `--native-lib`?
is it ok to have multiple copies of these libraries in a process?

wamr-compiler/CMakeLists.txt
```diff
@@ -387,6 +387,12 @@ if (NOT MSVC)
       target_link_libraries (wamrc ssp.a ws2_32)
   else()
       target_link_libraries (wamrc -ldl)
+      # Link libc++ statically to reduce the runtime dependency
+      target_link_libraries (wamrc -static-libstdc++)
+      # If not on macOS, link libgcc statically
+      if (NOT APPLE)
+        target_link_libraries (wamrc -static-libgcc)
```
xujuntwt95329, line 394: Can we make this configurable?
I think static link such libraries is not a very common requirement.

no1wudi:
@xujuntwt95329 @yamt I'm trying to make `wamrc` compatible with different linux distributions like https://github.com/WebAssembly/wasi-sdk/releases.

I thinks it's useful since we also provide prebuilt binary for wamrc.

wenyongh:
> @xujuntwt95329 @yamt I'm trying to make `wamrc` compatible with different linux distributions like https://github.com/WebAssembly/wasi-sdk/releases.
> 
> I thinks it's useful since we also provide prebuilt binary for wamrc.

@no1wudi it sounds good to me, but it increases the wamrc binary size. Could we link the static gcc/g++ lib only for the CI prebuilt binaries and keep it unchanged for normal compilation? Not sure whether it is easy to achieve that?

lum1n0us:
I guess libc++ is a basement for every platform. If there are compatible issues with a dynamic libc++ library, it highly likely suggests there is a real compatible issue with the platform(via libc++). Using a static library may let wamrc run, but also covers some real problems, I am afraid.

no1wudi:
Please refer to widely used wasi-sdk and llvm does if you have any concerns:
https://github.com/llvm/llvm-project/blob/de3e4a9dfe89dfc0a4d6d5e0891c542f6c82ca57/llvm/cmake/modules/HandleLLVMStdlib.cmake#L35

https://github.com/WebAssembly/wasi-sdk/blob/99abb024209dc7a9a8d8f3eb6d03908a18bb0594/Makefile#L84

Static link stdc++ should be except increase the binary size, the real problem is glibc, it's can not be linked statically.

The wasi-sdk build the toolchain on older version system (glibc, Ubuntu-18), so if glibc version in your system higher than it then you can run it.

yamt:
> Please refer to widely used wasi-sdk and llvm does if you have any concerns: https://github.com/llvm/llvm-project/blob/de3e4a9dfe89dfc0a4d6d5e0891c542f6c82ca57/llvm/cmake/modules/HandleLLVMStdlib.cmake#L35
> 
> https://github.com/WebAssembly/wasi-sdk/blob/99abb024209dc7a9a8d8f3eb6d03908a18bb0594/Makefile#L84
> 
> Static link stdc++ should be except increase the binary size, the real problem is glibc, it's can not be linked statically.
> 
> The wasi-sdk build the toolchain on older version system (glibc, Ubuntu-18), so if glibc version in your system higher than it then you can run it.

i can understand about libstdc++.
but do you have any specific concerns about libgcc?

also, can you answer this question? https://github.com/bytecodealliance/wasm-micro-runtime/pull/3331#discussion_r1569987973

no1wudi:
> > Please refer to widely used wasi-sdk and llvm does if you have any concerns: https://github.com/llvm/llvm-project/blob/de3e4a9dfe89dfc0a4d6d5e0891c542f6c82ca57/llvm/cmake/modules/HandleLLVMStdlib.cmake#L35
> > https://github.com/WebAssembly/wasi-sdk/blob/99abb024209dc7a9a8d8f3eb6d03908a18bb0594/Makefile#L84
> > Static link stdc++ should be except increase the binary size, the real problem is glibc, it's can not be linked statically.
> > The wasi-sdk build the toolchain on older version system (glibc, Ubuntu-18), so if glibc version in your system higher than it then you can run it.
> 
> i can understand about libstdc++. but do you have any specific concerns about libgcc?
> 
> also, can you answer this question? [#3331 (comment)](https://github.com/bytecodealliance/wasm-micro-runtime/pull/3331#discussion_r1569987973)

Maybe libgcc is not a problem for this case, but I'm not sure will it lead to compatibility issue in early test so it static linked here, we can remove it after more test.

wamr-compiler/CMakeLists.txt
no1wudi: I'm not sure which copy will be used in the native lib, but it seems OK since I have used a cxx based custom compiler plugin with wasi-sdk, it works fine without special action, please refer to: https://clang.llvm.org/docs/ClangPlugins.html

PR #3333, opened 2024-04-19
Fix windows relocation string parsing issue

wenyongh:
The AOT compiler emits string with '\0' now, some string parsing in aot loader
for Windows platform need to be updated.


TianlongLiang:
LGTM

Issue #3337, opened 2024-04-21
Unexpected Execution ()


erxiaozhou:
### Subject of the issue
The WAMR in JIT has unexpected behavior when executing a test case with illegal control flow.
### Test case

[f32x4.gt_55_571_17107906309149652.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/15051233/f32x4.gt_55_571_17107906309149652.zip)

[i16x8.extend_low_i8x16_s_7_200_17126751631561782.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/15051234/i16x8.extend_low_i8x16_s_7_200_17126751631561782.zip)

### Your environment

* Host OS: Ubuntu 20.04
* Commit Id: e7a8b3e743f8821eddbfc526eec19782c51c8408

### Steps to reproduce
1. build WAMR in JIT mode
```
export CC=/usr/lib/llvm-16/bin/clang
export CXX=/usr/lib/llvm-16/bin/clang++
cd product-mini/platforms/linux/;rm -rf build
cmake -DWAMR_BUILD_AOT=1 -DWAMR_BUILD_FAST_JIT=1   -DWAMR_BUILD_REF_TYPES=1 -DCMAKE_BUILD_TYPE=Release  -DWASM_ENABLE_BULK_MEMORY=1 -DWAMR_BUILD_LIBC_WASI=0 -DWAMR_BUILD_LIBC_BUILTIN=1 -DWAMR_BUILD_SIMD=1  -Bbuild
```
3. execute the command
```
iwasm --heap-size=0 --llvm-jit -f "to_test" <test_case>
```

### Expected behavior
 An exception indicating " else found outside of an `if` block"

### Actual behavior
No exception

wenyongh:
@erxiaozhou thanks for reporting the issue, I fixed it in PR #3340, please try again.

wenyongh:
Close the issue as PR #3340 was merged.

Issue #3336, opened 2024-04-21
An unexpected behavior in multiple modes


erxiaozhou:
### Subject of the issue
The WAMR compiled  in various configurations do not raise an exception on an invalid case.
### Test case
[memory.init_2164_858_1712663578577073.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/15051030/memory.init_2164_858_1712663578577073.zip)

### Your environment

* Host OS: Ubuntu 20.04
* Commit Id: e7a8b3e743f8821eddbfc526eec19782c51c8408

### Steps to reproduce
1. build WAMR in class interpreter mode / fast interpreter mode / JIT mode / fast JIT mode (Here take JIT mode as an example)
```
export CC=/usr/lib/llvm-16/bin/clang
export CXX=/usr/lib/llvm-16/bin/clang++
cd product-mini/platforms/linux/;rm -rf build
cmake -DWAMR_BUILD_AOT=1 -DWAMR_BUILD_FAST_JIT=1   -DWAMR_BUILD_REF_TYPES=1 -DCMAKE_BUILD_TYPE=Release  -DWASM_ENABLE_BULK_MEMORY=1 -DWAMR_BUILD_LIBC_WASI=0 -DWAMR_BUILD_LIBC_BUILTIN=1 -DWAMR_BUILD_SIMD=1  -Bbuild
```
3. execute the command
```
iwasm --heap-size=0 --llvm-jit -f "to_test" <test_case>
```

### Expected behavior
undeclared function reference (at offset 0x292)

### Actual behavior
No exception

wenyongh:
@erxiaozhou thanks for spotting the issue, the check for ref.func opcode in wasm loader is incomplete, I submitted PR #3340 to fix it, please help check and test again.

wenyongh:
Close the issue as PR #3340 was merged.

PR #3340, opened 2024-04-22
wasm loader: Fix checks for opcode ref.func and else

wenyongh:
Fix wasm loader integrity checks for opcode ref.func and opcode else:
for opcode ref.func, the function must be an import, exported, or present in a
table elem segment or global initializer to be used as the operand to ref.func,
for opcode else, there must not be a else previously.

Reported in #3336 and #3337.

And fix mini loader PUSH_MEM_OFFSET/POP_MEM_OFFSET macro
definitions due to the introducing of memory64 feature.


TianlongLiang:
LGTM

Issue #3314, opened 2024-04-12
default WASM_STACK_GUARD_SIZE seems too small for linux


yamt:
on ubuntu 20.04 amd64, PLT resolution seems to consume about 1500 bytes stack.
thus, the default 1024 is not sufficient for most non-trivial host functions / aot helpers.

wenyongh:
Yes, how about increasing it to 2048 for linux platform?

BTW, the size of argv_buf in wasm_runtime_invoke_native seems too large and unnecessary since normally the parameter count of wasm function and host function won't be very large:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/30426be82c806391c9709c5c11c891cdf50b528f/core/iwasm/common/wasm_runtime_common.c#L4883-L4884

How about we also decreasing it, e.g. `uint64 argv_buf[12]`?

yamt:
> Yes, how about increasing it to 2048 for linux platform?

sure, it's the simplest solution.

> BTW, the size of argv_buf in wasm_runtime_invoke_native seems too large and unnecessary since normally the parameter count of wasm function and host function won't be very large:
> 
> https://github.com/bytecodealliance/wasm-micro-runtime/blob/30426be82c806391c9709c5c11c891cdf50b528f/core/iwasm/common/wasm_runtime_common.c#L4883-L4884
> 
> How about we also decreasing it, e.g. `uint64 argv_buf[12]`?

it makes sense to decrease it.
what's your rationale for 12?

yamt:
btw, i want to group these issues with a label, say, "native stack overflow detection".
can you give me necessary permissions?

ditto for exception-handling.

wenyongh:
> btw, i want to group these issues with a label, say, "native stack overflow detection". can you give me necessary permissions?
> 
> ditto for exception-handling.

OK, done, please have a try.

yamt:
> > btw, i want to group these issues with a label, say, "native stack overflow detection". can you give me necessary permissions?
> > ditto for exception-handling.
> 
> OK, done, please have a try.

i can assign existing labels. but i can't create new labels.

wenyongh:
> > > btw, i want to group these issues with a label, say, "native stack overflow detection". can you give me necessary permissions?
> > > ditto for exception-handling.
> > 
> > 
> > OK, done, please have a try.
> 
> i can assign existing labels. but i can't create new labels.

[...]

PR #3344, opened 2024-04-23
core/config.h: Bump the default WASM_STACK_GUARD_SIZE

yamt:
The old value (1KB) doesn't seem sufficient for many cases.

I suspect that the new value is still not sufficient for some cases. But it's far safer than the old value.
Consider if the classic interpreter loop (2600 bytes) calls host snprintf. (2000 bytes)

Fixes: https://github.com/bytecodealliance/wasm-micro-runtime/issues/3314

### DIFF
--- a/core/config.h
+++ b/core/config.h
@@ -486,7 +486,7 @@
 /* UVWASI requires larger native stack */
 #define WASM_STACK_GUARD_SIZE (4096 * 6)
 #else
-#define WASM_STACK_GUARD_SIZE (1024)
+#define WASM_STACK_GUARD_SIZE (1024 * 5)
 #endif
 #endif
 
--- a/core/config.h
+++ b/core/config.h
@@ -486,7 +486,26 @@
 /* UVWASI requires larger native stack */
 #define WASM_STACK_GUARD_SIZE (4096 * 6)
 #else
+/*
+ * Use a larger default for platforms like macOS/Linux.
+ *
+ * For example, wasm_interp_call_func_bytecode + wasm_runtime_set_exception
+ * would consume >4KB stack on x86-64 macOS.
+ *
+ * Although product-mini/platforms/nuttx always overrides
+ * WASM_STACK_GUARD_SIZE, exclude NuttX here just in case.
+ */
+#if defined(__APPLE__) || (defined(__unix__) && !defined(__NuttX__))
 #define WASM_STACK_GUARD_SIZE (1024 * 5)
+#else
+/*
+ * Otherwise, assume very small requirement for now.
+ *
+ * Embedders for very small devices likely fine-tune WASM_STACK_GUARD_SIZE
+ * for their specific applications anyway.
+ */
+#define WASM_STACK_GUARD_SIZE 1024
+#endif
 #endif
 #endif
 


core/config.h
```diff
@@ -486,7 +486,7 @@
 /* UVWASI requires larger native stack */
 #define WASM_STACK_GUARD_SIZE (4096 * 6)
 #else
-#define WASM_STACK_GUARD_SIZE (1024)
+#define WASM_STACK_GUARD_SIZE (1024 * 5)
```
wenyongh, line 489: Is it better to increase the size only for linux/macos or posix like platforms? Seems that not all platforms need so much guard stack space, e.g. zephyr/esp-idf 32-bit, had better keep it unchanged for them now?

yamt: maybe for 32-bit archs it can be a bit smaller. but are you sure 1KB is enough?

wenyongh: BTW, maybe we can also decrease the native stack usage by refining the code which calls snprintf? For example, in wasm_set_exception_local and set_exception_visitor, copy character one by one from source string until '\0' is found.

wenyongh: > maybe for 32-bit archs it can be a bit smaller. but are you sure 1KB is enough?

I am not sure, my suggestion is to keep it unchanged temporarily, and change it after we do more test.

wenyongh: ps. keep it unchanged for non-posix like platforms until we do more test.

yamt: > BTW, maybe we can also decrease the native stack usage by refining the code which calls snprintf? For example, in wasm_set_exception_local and set_exception_visitor, copy character one by one from source string until '\0' is found.

i agree it's probably a good idea to reduce the stack usage of wasm_runtime_set_exception and friends.

but i'm not sure if we can reduce WASM_STACK_GUARD_SIZE with it unless we introduce a rather strong policy about where os_printf, LOG_DEBUG, etc can be used.

yamt: > > maybe for 32-bit archs it can be a bit smaller. but are you sure 1KB is enough?
> 
> I am not sure, my suggestion is to keep it unchanged temporarily, and change it after we do more test.

i can probably investigate the situation on xtensa/nuttx.
do you think it's good enough to make it represent the "low end" side?

wenyongh: > > > maybe for 32-bit archs it can be a bit smaller. but are you sure 1KB is enough?
> > 
> > 
> > I am not sure, my suggestion is to keep it unchanged temporarily, and change it after we do more test.
> 
> i can probably investigate the situation on xtensa/nuttx. do you think it's good enough to make it represent the "low end" side?

OK, thanks. Distinguish it into high end side and low end side is good to me, do yo think we should put the macro in each platform's platform_internal.h?

yamt: > > > > maybe for 32-bit archs it can be a bit smaller. but are you sure 1KB is enough?
> > > 
> > > 
> > > I am not sure, my suggestion is to keep it unchanged temporarily, and change it after we do more test.
> > 
> > 
> > i can probably investigate the situation on xtensa/nuttx. do you think it's good enough to make it represent the "low end" side?
> 
> OK, thanks. Distinguish it into high end side and low end side is good to me, do yo think we should put the macro in each platform's platform_internal.h?

maybe. after all, it's very platform dependent.
otoh, i suspect that embedders for very small devices usually fine-tune WASM_STACK_GUARD_SIZE for their specific applications anyway.

wenyongh: > > BTW, maybe we can also decrease the native stack usage by refining the code which calls snprintf? For example, in wasm_set_exception_local and set_exception_visitor, copy character one by one from source string until '\0' is found.
> 
> i agree it's probably a good idea to reduce the stack usage of wasm_runtime_set_exception and friends.
> 
> but i'm not sure if we can reduce WASM_STACK_GUARD_SIZE with it unless we introduce a rather strong policy about where os_printf, LOG_DEBUG, etc can be used.

Yes, I think we can increase WASM_STACK_GUARD_SIZE to 5KB and also refine the code in some places to enhance the security. And decrease WASM_STACK_GUARD_SIZE only when we do a full test to make sure it is good.

wenyongh: > > > > > maybe for 32-bit archs it can be a bit smaller. but are you sure 1KB is enough?
> > > > 
> > > > 
> > > > I am not sure, my suggestion is to keep it unchanged temporarily, and change it after we do more test.
> > > 
> > > 
> > > i can probably investigate the situation on xtensa/nuttx. do you think it's good enough to make it represent the "low end" side?
> > 
> > 
> > OK, thanks. Distinguish it into high end side and low end side is good to me, do yo think we should put the macro in each platform's platform_internal.h?
> 
> maybe. after all, it's very platform dependent. otoh, i suspect that embedders for very small devices usually fine-tune WASM_STACK_GUARD_SIZE for their specific applications anyway.

OK, so had better keep putting the macro in core/config.h, and maybe we can add a document for detailed description about this macro and the native stack overflow check.

yamt: reverted for non-posix platforms for now.

wenyongh (approved):
LGTM

core/config.h
no1wudi: > maybe. after all, it's very platform dependent. otoh, i suspect that embedders for very small devices usually fine-tune WASM_STACK_GUARD_SIZE for their specific applications anyway.

Agree.

PR #3362, opened 2024-04-25
Implement Memory64 support for AOT

loganek:
All spec tests are passing now; I also ran the memory64 tests from @TianlongLiang 's  wamr-test repository.

### DIFF
--- a/core/iwasm/aot/aot_loader.c
+++ b/core/iwasm/aot/aot_loader.c
@@ -9,6 +9,7 @@
 #include "aot_reloc.h"
 #include "../common/wasm_runtime_common.h"
 #include "../common/wasm_native.h"
+#include "../common/wasm_loader_common.h"
 #include "../compilation/aot.h"
 
 #if WASM_ENABLE_DEBUG_AOT != 0
--- a/core/iwasm/aot/aot_loader.c
+++ b/core/iwasm/aot/aot_loader.c
@@ -1043,6 +1043,12 @@ load_memory_info(const uint8 **p_buf, const uint8 *buf_end, AOTModule *module,
 
     for (i = 0; i < module->memory_count; i++) {
         read_uint32(buf, buf_end, module->memories[i].memory_flags);
+
+        if (!wasm_memory_check_flags(module->memories[i].memory_flags,
+                                     error_buf, error_buf_size, true)) {
+            return false;
+        }
+
         read_uint32(buf, buf_end, module->memories[i].num_bytes_per_page);
         read_uint32(buf, buf_end, module->memories[i].mem_init_page_count);
         read_uint32(buf, buf_end, module->memories[i].mem_max_page_count);
--- a/core/iwasm/aot/aot_loader.c
+++ b/core/iwasm/aot/aot_loader.c
@@ -3632,6 +3632,21 @@ fail:
     return ret;
 }
 
+#if WASM_ENABLE_MEMORY64 != 0
+static bool
+has_module_memory64(AOTModule *module)
+{
+    /* TODO: multi-memories for now assuming the memory idx type is consistent
+     * across multi-memories */
+    if (module->import_memory_count > 0)
+        return !!(module->import_memories[0].memory_flags & MEMORY64_FLAG);
+    else if (module->memory_count > 0)
+        return !!(module->memories[0].memory_flags & MEMORY64_FLAG);
+
+    return false;
+}
+#endif
+
 static bool
 load_from_sections(AOTModule *module, AOTSection *sections,
                    bool is_load_from_file_buf, char *error_buf,
@@ -3643,6 +3658,7 @@ load_from_sections(AOTModule *module, AOTSection *sections,
     uint32 i, func_index, func_type_index;
     AOTFuncType *func_type;
     AOTExport *exports;
+    uint8 malloc_free_io_type = VALUE_TYPE_I32;
 
     while (section) {
         buf = section->section_body;
@@ -3717,7 +3733,10 @@ load_from_sections(AOTModule *module, AOTSection *sections,
     module->malloc_func_index = (uint32)-1;
     module->free_func_index = (uint32)-1;
     module->retain_func_index = (uint32)-1;
-
+#if WASM_ENABLE_MEMORY64 != 0
+    if (has_module_memory64(module))
+        malloc_free_io_type = VALUE_TYPE_I64;
+#endif
     exports = module->exports;
     for (i = 0; i < module->export_count; i++) {
         if (exports[i].kind == EXPORT_KIND_FUNC
@@ -3727,8 +3746,8 @@ load_from_sections(AOTModule *module, AOTSection *sections,
                 func_type_index = module->func_type_indexes[func_index];
                 func_type = (AOTFuncType *)module->types[func_type_index];
                 if (func_type->param_count == 1 && func_type->result_count == 1
-                    && func_type->types[0] == VALUE_TYPE_I32
-                    && func_type->types[1] == VALUE_TYPE_I32) {
+                    && func_type->types[0] == malloc_free_io_type
+                    && func_type->types[1] == malloc_free_io_type) {
                     bh_assert(module->malloc_func_index == (uint32)-1);
                     module->malloc_func_index = func_index;
                     LOG_VERBOSE("Found malloc function, name: %s, index: %u",
@@ -3740,9 +3759,9 @@ load_from_sections(AOTModule *module, AOTSection *sections,
                 func_type_index = module->func_type_indexes[func_index];
                 func_type = (AOTFuncType *)module->types[func_type_index];
                 if (func_type->param_count == 2 && func_type->result_count == 1
-                    && func_type->types[0] == VALUE_TYPE_I32
+                    && func_type->types[0] == malloc_free_io_type
                     && func_type->types[1] == VALUE_TYPE_I32
-                    && func_type->types[2] == VALUE_TYPE_I32) {
+                    && func_type->types[2] == malloc_free_io_type) {
                     uint32 j;
                     WASMExport *export_tmp;
 
@@ -3766,8 +3785,8 @@ load_from_sections(AOTModule *module, AOTSection *sections,
                                 (AOTFuncType *)module->types[func_type_index];
                             if (func_type->param_count == 1
                                 && func_type->result_count == 1
-                                && func_type->types[0] == VALUE_TYPE_I32
-                                && func_type->types[1] == VALUE_TYPE_I32) {
+                                && func_type->types[0] == malloc_free_io_type
+                                && func_type->types[1] == malloc_free_io_type) {
                                 bh_assert(module->retain_func_index
                                           == (uint32)-1);
                                 module->retain_func_index = export_tmp->index;
@@ -3793,7 +3812,7 @@ load_from_sections(AOTModule *module, AOTSection *sections,
                 func_type_index = module->func_type_indexes[func_index];
                 func_type = (AOTFuncType *)module->types[func_type_index];
                 if (func_type->param_count == 1 && func_type->result_count == 0
-                    && func_type->types[0] == VALUE_TYPE_I32) {
+                    && func_type->types[0] == malloc_free_io_type) {
                     bh_assert(module->free_func_index == (uint32)-1);
                     module->free_func_index = func_index;
                     LOG_VERBOSE("Found free function, name: %s, index: %u",
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -72,10 +72,13 @@ bh_static_assert(offsetof(AOTFrame, sp) == sizeof(uintptr_t) * 5);
 bh_static_assert(offsetof(AOTFrame, frame_ref) == sizeof(uintptr_t) * 6);
 bh_static_assert(offsetof(AOTFrame, lp) == sizeof(uintptr_t) * 7);
 
+#if WASM_ENABLE_BULK_MEMORY != 0
 static uint32
-clamp_u64_to_u32(uint64 value) {
+clamp_u64_to_u32(uint64 value)
+{
     return value > UINT32_MAX ? UINT32_MAX : (uint32)value;
 }
+#endif
 
 static void
 set_error_buf(char *error_buf, uint32 error_buf_size, const char *string)
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -2347,14 +2347,32 @@ execute_malloc_function(AOTModuleInstance *module_inst, WASMExecEnv *exec_env,
 #endif
     WASMExecEnv *exec_env_created = NULL;
     WASMModuleInstanceCommon *module_inst_old = NULL;
-    uint32 argv[2], argc;
+    uint32 argv[3], argc;
     bool ret;
 
-    argv[0] = size;
-    argc = 1;
-    if (retain_func) {
-        argv[1] = 0;
+#if WASM_ENABLE_MEMORY64 != 0
+    bool is_memory64 = module_inst->memories[0]->is_memory64;
+    if (is_memory64) {
         argc = 2;
+        PUT_I64_TO_ADDR(&argv[0], size);
+    }
+    else
+#endif
+    {
+        argc = 1;
+        argv[0] = size;
+    }
+
+    /* if __retain is exported, then this module is compiled by
+        assemblyscript, the memory should be managed by as's runtime,
+        in this case we need to call the retain function after malloc
+        the memory */
+    if (retain_func) {
+        /* the malloc function from assemblyscript is:
+            function __new(size: usize, id: u32)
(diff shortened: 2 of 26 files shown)


wenyongh:
@loganek the macos CI failure should have been fixed with PR https://github.com/bytecodealliance/wasm-micro-runtime/pull/3366, please rebase your code to run again. Thanks.

loganek:
@wenyongh thanks for letting me know, I'll do that. The pr is in the draft state yet so I fix the ci failures and do a self review; I'll switch it to non-draft once it's ready for the review

loganek:
@wenyongh / @TianlongLiang I think the code is ready for review, I'd appreciate your feedback.

core/iwasm/aot/aot_loader.c
```diff
@@ -3632,6 +3632,21 @@ load_relocation_section(const uint8 *buf, const uint8 *buf_end,
...
+    /* TODO: multi-memories for now assuming the memory idx type is consistent
+     * across multi-memories */
+    if (module->import_memory_count > 0)
+        return !!(module->import_memories[0].memory_flags & MEMORY64_FLAG);
+    else if (module->memory_count > 0)
+        return !!(module->memories[0].memory_flags & MEMORY64_FLAG);
+
+    return false;
+}
+#endif
```
wenyongh, line 3648: In check_feature_flags function of this file, we check whether the related feature is enabled when the feature flag is set in the AOT file. Had better also emit the memory64 feature flag in the AOT file when there is a 64-bit memory in the wasm file, and add the check in check_feature_flags.

tests/wamr-test-suites/spec-test-script/runtest.py
```diff
@@ -1119,9 +1119,6 @@ def compile_wasm_to_aot(wasm_tempfile, aot_tempfile, runner, opts, r, output = '
         cmd.append("--enable-gc")
         cmd.append("--enable-tail-call")
 
-    if opts.memory64:
-        cmd.append("--enable-memory64")
```
wenyongh, line 1123: Seems we don't need to add `--enable-memory64` flag for wamrc, it can auto generate the correct aot file according to the memory definition in the wasm file?

TianlongLiang:
Hi, @loganek, I think maybe it's worth considering 
1. adding a command line option like `--enable-memory64` in wamrc so that user can opt it out like user can choose to `--enable-multi-thread` and `--enable-gc`, in this way, I think `tests/wamr-test-suites/spec-test-script/thread_proposal_remove_memory64_flag_case.patch` probably won't be needed 
2. emitting the corresponding feature flags(`WASM_FEATURE_MEMORY64`) in `aot_emit_aot_file.c` when there is actually i64 type memory section and correspondingly modify `check_feature_flags` in `aot_loader.c`, this way I think it's also can get rid of the patch

wenyongh:
> @wenyongh / @TianlongLiang I think the code is ready for review, I'd appreciate your feedback.

@loganek thanks for the quick update. I just read the code quickly and added two comments, and hope to read it carefully to understand more. But since Tianlong and I are in holiday in the left of this week, we might response slowly then.

loganek:
> But since Tianlong and I are in holiday in the left of this week, we might response slowly then.
no worries, enjoy your holiday :)

tests/wamr-test-suites/spec-test-script/runtest.py
loganek: yes, that's why I removed this option.

loganek:
> adding a command line option like --enable-memory64 in wamrc so that user can opt it out like user can choose to --enable-multi-thread and --enable-gc, in this way, I think tests/wamr-test-suites/spec-test-script/thread_proposal_remove_memory64_flag_case.patch probably won't be needed

I personally don't think this flag is needed, I'm not sure if there's usecase where the user has a memory64 module and does NOT want to build with memory64 - in all cases that should fail, so I'd rather delete the cli option and let compiler automatically detect the type of module.

The patch is to handle edge case where the module is invalid, and I don't think we actually ever want to support such case. In fact, that test was removed in the most recent version of the testsuite (I just didn't want to upgrade the whole test suite in this PR to keep changes focused to memory64 only).

core/iwasm/compilation/aot_emit_memory.c
```diff
@@ -987,13 +1003,13 @@ aot_compile_op_memory_init(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
...
     POP_I32(offset);
-    POP_I32(dst);
+    POP_MEM_OFFSET(dst);
 
     param_types[0] = INT8_PTR_TYPE;
     param_types[1] = I32_TYPE;
     param_types[2] = I32_TYPE;
     param_types[3] = I32_TYPE;
-    param_types[4] = I32_TYPE;
+    param_types[4] = MEMORY64_COND_VALUE(I64_TYPE, I32_TYPE);
```
wenyongh, line 1012: The `dst` argument of `aot_memory_init` is `mem_offset_t`, which is defined as:
```C
#if WASM_ENABLE_MEMORY64 != 0
typedef uint64 mem_offset_t;
#else
typedef uint32 mem_offset_t;
#endif
```
So it is uint64 when WASM_ENABLE_MEMORY64 is set as 1, no matter the actual memory is 64 bit or 32 bit.
And here the param type of dst is decided according to the memory type, it is I32_TYPE when the memory is 32 bit. The issue reported in #3375 may occur again: aot runtime enables memory64 feature and the argument type is uint64, while memory in aot file is 32bit and aot code passes uint32 to the argument.

How about changing the param type in aot_memory_init to size_t, and here set the param type according to `comp_ctx->pointer_size`, and extend the param to uint64 if needed before passing to aot_memory_init? Similar to #3378.

core/iwasm/compilation/aot_emit_memory.c
```diff
@@ -1165,16 +1181,16 @@ aot_compile_op_memory_fill(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx)
...
-    POP_I32(dst);
+    POP_MEM_OFFSET(dst);
 
     if (!(dst_addr = check_bulk_memory_overflow(comp_ctx, func_ctx, dst, len)))
         return false;
 
     param_types[0] = INT8_PTR_TYPE;
     param_types[1] = I32_TYPE;
-    param_types[2] = I32_TYPE;
+    param_types[2] = MEMORY64_COND_VALUE(I64_TYPE, I32_TYPE);
```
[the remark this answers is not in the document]
loganek: thanks, updated

core/iwasm/compilation/aot_emit_memory.c
```diff
@@ -1097,7 +1113,7 @@ aot_compile_op_memory_copy(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx)
 
         param_types[0] = INT8_PTR_TYPE;
         param_types[1] = INT8_PTR_TYPE;
-        param_types[2] = I32_TYPE;
+        param_types[2] = MEMORY64_COND_VALUE(I64_TYPE, I32_TYPE);
```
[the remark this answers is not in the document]
loganek: right, and in fact this should be based on the target architecture, not based on the memory type.

core/iwasm/aot/aot_loader.c
loganek: hmm, I'm quite sure I've got `wasm_memory_check_flags` call in AOT (that's why I refactored it with the `is_aot` flag parameter) but looks like I must have removed it when doing some refactoring. I'll add it back in the `load_memory_info`, I don't think we need to add feature flags if the information about the memory type is already encoded in the aot bundle anyway.

core/iwasm/compilation/aot_emit_memory.c
```diff
@@ -854,7 +869,7 @@ aot_compile_op_memory_grow(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx)
 
     /* Call function aot_enlarge_memory() */
     param_values[0] = func_ctx->aot_inst;
-    param_values[1] = delta;
+    param_values[1] = LLVMBuildTrunc(comp_ctx->builder, delta, I32_TYPE, "");
```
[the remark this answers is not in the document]
loganek: Good idea, updated.

core/iwasm/compilation/aot_emit_memory.c
```diff
@@ -852,9 +885,27 @@ aot_compile_op_memory_grow(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx)
...
+            return false;
+        }
+        BUILD_ICMP(LLVMIntUGT, delta, u32_max, ret_value, "page_size_cmp");
+        if (!(check_page_size = LLVMAppendBasicBlockInContext(
+                  comp_ctx->context, func_ctx->func, "check_page_size"))) {
+            aot_set_last_error("llvm add basic block failed.");
+            goto fail;
+        }
+        if (!(aot_emit_exception(comp_ctx, func_ctx, EXCE_INTEGER_OVERFLOW,
+                                 true, ret_value, check_page_size)))
```
wenyongh, line 901: When `memory.grow` fails, it should return -1 but not throw exception, see wasm spec and the implementation in interpreter:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_interp_classic.c#L4557

Suggest here:
```C
    BUILD_ICMP(LLVMIntUGT, delta, u32_max, ret_value1, "page_size_cmp");
```

And change below to:
```C
    BUILD_ICMP(LLVMIntUGT, ret_value, I8_ZERO, ret_value, "mem_grow_ret");

    BUILD_OP(Or, ret_value, ret_value1, ret_value, "or");  => ret_value = ret_value | ret_value1;

    /* ret_value = ret_value == true ? delta : pre_page_count */
    if (!(ret_value = LLVMBuildSelect(
              comp_ctx->builder, ret_value, mem_size,
              MEMORY64_COND_VALUE(I64_NEG_ONE, I32_NEG_ONE), "mem_grow_ret"))) {
        aot_set_last_error("llvm build select failed.");
        return false;
    }

    PUSH_PAGE_COUNT(ret_value);
```

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -4176,7 +4176,7 @@ llvm_jit_invoke_native(WASMExecEnv *exec_env, uint32 func_idx, uint32 argc,
 #if WASM_ENABLE_BULK_MEMORY != 0
 bool
 llvm_jit_memory_init(WASMModuleInstance *module_inst, uint32 seg_index,
-                     uint32 offset, uint32 len, uint32 dst)
+                     uint32 offset, uint32 len, size_t dst)
```
wenyongh, line 4179: Should also call `clamp_u64_to_u32(memory_inst->memory_data_size - dst)` in below `bh_memcpy_s`

core/iwasm/compilation/aot_emit_memory.c
```diff
@@ -864,13 +915,14 @@ aot_compile_op_memory_grow(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx)
     BUILD_ICMP(LLVMIntUGT, ret_value, I8_ZERO, ret_value, "mem_grow_ret");
 
     /* ret_value = ret_value == true ? delta : pre_page_count */
```
wenyongh, line 917: ret_value = ret_value == true ? pre_page_count : -1

core/iwasm/compilation/aot_emit_memory.c
wenyongh: Sorry that the logic should be `if (delta <= u32_max && aot_enlarge_memory()) then return pre_page_count, else return -1`, so seems had better:
```C
BUILD_ICMP(LLVMIntULE, delta, u32_max, ret_value1, "page_size_cmp");
...

BUILD_ICMP(LLVMIntUGT, ret_value, I8_ZERO, ret_value, "mem_grow_ret");

#if WASM_ENABLE_MEMORY64 != 0
    BUILD_OP(And, ret_value1, ret_value, ret_value, "and");  => ret_value = ret_value1 & ret_value;
#endif
```

core/iwasm/aot/aot_runtime.c
```diff
@@ -2401,8 +2445,16 @@ execute_malloc_function(AOTModuleInstance *module_inst, WASMExecEnv *exec_env,
     if (exec_env_created)
         wasm_exec_env_destroy(exec_env_created);
 
-    if (ret)
-        *p_result = argv[0];
+    if (ret) {
+#if WASM_ENABLE_MEMORY64 != 0
+        if (is_memory64)
+            *p_result = GET_I64_FROM_ADDR(&argv[0]);
```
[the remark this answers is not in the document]
loganek: true, also the `size` parameter should be `uint64`, updated that too.

core/iwasm/interpreter/wasm_runtime.c
loganek: I haven't looked into JIT and there's no way to test memory64 in JIT right now as memory64 is not supported in this mode, but that's a reasonable change, updated.

loganek:
> it's rather important to fail when it's expected to fail because, in case of aot, it's wamrc's resposibility to validate the module.

I agree it's important, I don't think though that the explicit flag is needed; the module is self-explanatory in terms of the memory type. If we really want to do that explicit validation for some reason, I think we should be consistent across the whole wamr, and not only have the explicit flag in the compiler, but also for runtime (and not only for memory64, but also other features like threads). As I said, I don't see a need for that, and that'd slightly complicate user's life as they'll need to know in advance what type of module they load. As of today they can do the validation beforehand if they need it, but it's optional.

core/iwasm/aot/aot_runtime.c
```diff
@@ -2401,24 +2441,46 @@ execute_malloc_function(AOTModuleInstance *module_inst, WASMExecEnv *exec_env,
...
+            *p_result = argv.u32[0];
+        }
+    }
     return ret;
 }
 
 static bool
 execute_free_function(AOTModuleInstance *module_inst, WASMExecEnv *exec_env,
-                      AOTFunctionInstance *free_func, uint32 offset)
+                      AOTFunctionInstance *free_func, uint64 offset)
```
wenyongh, line 2459: Should also modify:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/ca61184ced7d885395bc2425c87f0d063935dd1e/core/iwasm/aot/aot_runtime.c#L2623-L2624
change `(uint32)ptr` to `ptr`

wenyongh:
> > it's rather important to fail when it's expected to fail because, in case of aot, it's wamrc's resposibility to validate the module.
> 
> I agree it's important, I don't think though that the explicit flag is needed; the module is self-explanatory in terms of the memory type. If we really want to do that explicit validation for some reason, I think we should be consistent across the whole wamr, and not only have the explicit flag in the compiler, but also for interpreter (and not only for memory64, but also other features like threads). As I said, I don't see a need for that, and that'd slightly complicate user's life as they'll need to know in advance what type of module they load. As of today they can do the validation beforehand if they need it, but it's optional.

Per my understanding, wamrc supports both memory32 and memory64 and can emit correct AOT files accordingly. And for memory64, the memory info in the AOT file also has the memory64 flag (bit 2 is set), the aot runtime can report error if memory64 feature isn't enabled, which is consistent with interpreter. So no need to explicitly add `--enable-memory64` flag for wamrc, right? I think it is acceptable since it seems more convenient to developers.

wenyongh:
@loganek the PR LGTM except one issue left (change `(uint32)ptr` to `ptr` when calling execute_free_function), could you help address it and resolve the conflicting file? I think we can merge this PR then if @yamt and @TianlongLiang are also good.

core/iwasm/compilation/aot_compiler.c
```diff
@@ -2743,7 +2760,7 @@ aot_compile_func(AOTCompContext *comp_ctx, uint32 func_index)
                     case SIMD_v128_store:
                     {
                         read_leb_uint32(frame_ip, frame_ip_end, align);
-                        read_leb_uint32(frame_ip, frame_ip_end, offset);
+                        read_leb_mem_offset(frame_ip, frame_ip_end, offset);
```
TianlongLiang, line 2763: Should also replace the `read_leb_uint32(frame_ip, frame_ip_end, offset);` for L3009, L3023, L3035. 

And I think it's best to leave a comment before `case WASM_OP_SIMD_PREFIX` and `case WASM_OP_ATOMIC_PREFIX` with: `TODO: in the future, replace the offset type in aot_compile_op_xxx for threads and SIMD opcodes compilation`

loganek: hmm, I think I'd rather just change this now instead of leaving todos

loganek:
@wenyongh I've addressed all the comments, thanks a lot for the review.

Unless there are any other comments, the PR is ready for merge.

wenyongh (approved):
LGTM

wenyongh:
> @wenyongh I've addressed all the comments, thanks a lot for the review.
> 
> Unless there are any other comments, the PR is ready for merge.

Great work! I think we can merge the PR if it is also good from @TianlongLiang.

TianlongLiang:
LGTM

Issue #3329, opened 2024-04-17
Add ability to get import and export information from wasm_export interface


bnason-nf:
#### Feature

When using the wasm_export API, it currently doesn't seem possible to get information about the imports and exports of a module. When using the wasm_c_api API, that information is exposed. It would be nice to add support for this in wasm_export also.

#### Benefit

Using wasm_export is beneficial in some cases where the added functionality it provides is desired, but if it's not a superset of wasm_c_api then there are drawbacks to using either API.

There is a little more discussion in issue #1740.

#### Implementation

I have a working implementation, and will submit a pull request of that as a starting point for the desired implementation.

#### Alternatives

The only alternative I'm aware of is extremely ugly and unsafe:

```C
    char errorString[128];
    wasm_module_t module = wasm_runtime_load(data, dataSize, errorString, sizeof(errorString));

    char moduleHack[64]; /* sizeof(wasm_module_ex_t) */
    memset(moduleHack, 0xff, sizeof(moduleHack)); /* fill with non-zero to make internal wasm_module_ex_t ref_count not be empty */
    *(wasm_module_t *)moduleHack = module;
    wasm_module_t *moduleEx = (wasm_module_t *)moduleHack;

    wasm_importtype_vec_t imports;
    memset(&imports, 0, sizeof(imports));
    wasm_module_imports(moduleEx, &imports);
```

bnason-nf:
@lum1n0us as we discussed in #1740, I've created this issue and a corresponding pull request. I'm not sure if this matches what you'd like to see stylistically, but it's a pretty simple implementation and fully provides the functionality I'm looking for. I'm happy to make any adjustments you'd like to see.

lum1n0us:
> The only alternative I'm aware of is extremely ugly and unsafe:

👍  Can't agree more.

PR #3330, opened 2024-04-17
Add functions to expose module import/export info

bnason-nf:
Resolves #3329.

PR #3363, opened 2024-04-26
Expose import/export function param/result info

bnason-nf:
This is a follow-up change for #3329, it adds APIs to wasm_export.h to provide information about parameters and results for import and export functions.

There is some overlap with these new APIs and these existing APIs:
```
wasm_runtime_lookup_function()
wasm_func_get_param_count(),
wasm_func_get_param_types()
wasm_func_get_result_count()
wasm_func_get_result_types()
```
However these existing APIs require an instantiated module, and also they don't support import functions.


bnason-nf:
The MacOS compilation failure here seems unrelated to this pull request.

wenyongh:
> The MacOS compilation failure here seems unrelated to this pull request.

@bnason-nf the macos CI failure should have been fixed with PR https://github.com/bytecodealliance/wasm-micro-runtime/pull/3366, please rebase your code to run again. Thanks.

core/iwasm/include/wasm_export.h
```diff
@@ -84,6 +84,9 @@ typedef struct wasm_export_type {
     wasm_import_export_kind_t kind;
 } wasm_export_type;
 
+struct WASMFuncType;
+typedef struct WASMFuncType *wasm_function_t;
```
wenyongh, line 88: The related type has been defined in core/iwasm/include/gc_export.h:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/120b9659dd4f3f500419e05fceab1f47aba54529/core/iwasm/include/gc_export.h#L138
And some similar APIs are also declared in gc_export.h and implemented in core/iwasm/common/gc/gc_common.c when GC feature is enabled:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/include/gc_export.h#L240-L277

How about re-using the definition wasm_func_type_t:
```C
struct WASMFuncType;
typedef struct WASMFuncType *wasm_func_type_t;
```
And define the APIs:
```C
WASM_RUNTIME_API_EXTERN wasm_func_type_t
wasm_runtime_get_export_func_type(wasm_module_t const module, const char *name);

WASM_RUNTIME_API_EXTERN wasm_func_type_t
wasm_runtime_get_import_func_type(wasm_module_t const module, const char *name);

/* Get the param type array and the param type count */
WASM_RUNTIME_API_EXTERN wasm_valkind_t *
wasm_func_type_get_param_types(const wasm_func_type_t func_type, uint32 *p_param_count);

/* Get the result type array and the result type count */
WASM_RUNTIME_API_EXTERN wasm_valkind_t *
wasm_func_type_get_result_types(const wasm_func_type_t func_type, uint32 *p_result_count);
```

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3934,6 +3934,184 @@ wasm_runtime_get_export_type(WASMModuleCommon *const module, int32 export_index,
...
+        return NULL;
+    }
+
+#if WASM_ENABLE_AOT != 0
+    if (module->module_type == Wasm_Module_AoT) {
+        const AOTModule *aot_module = (const AOTModule *)module;
+
+        for (uint32_t idx = 0; idx < aot_module->export_count; ++idx) {
+            const AOTExport *export_func = &aot_module->exports[idx];
+            if (!strcmp(export_func->name, name)) {
```
wenyongh, line 3952: Should also check whether `aot_module->exports[idx].kind == EXPORT_KIND_FUNC`

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3934,6 +3934,184 @@ wasm_runtime_get_export_type(WASMModuleCommon *const module, int32 export_index,
...
+        }
+    }
+#endif
+#if WASM_ENABLE_INTERP != 0
+    if (module->module_type == Wasm_Module_Bytecode) {
+        const WASMModule *wasm_module = (const WASMModule *)module;
+
+        for (uint32_t idx = 0; idx < wasm_module->export_count; ++idx) {
+            const WASMExport *export_func = &wasm_module->exports[idx];
+            if (!strcmp(export_func->name, name)) {
```
wenyongh, line 3970: Similar as above.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3934,6 +3934,184 @@ wasm_runtime_get_export_type(WASMModuleCommon *const module, int32 export_index,
 #endif
 }
 
+WASMFuncType *
+wasm_runtime_get_export_function(struct WASMModuleCommon *const module,
+                                 const char *name)
+{
```
lum1n0us, line 3940: 🤔 maybe there is another way to get `WASMFuncType`:

- build a `WASMExport` or use `wasm_runtime_get_export_type`(may need some modification)
- then use `wasm_runtime_get_export_func_type`

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3934,6 +3934,184 @@ wasm_runtime_get_export_type(WASMModuleCommon *const module, int32 export_index,
...
+    }
+#endif
+
+    return NULL;
+}
+
+WASMFuncType *
+wasm_runtime_get_import_function(struct WASMModuleCommon *const module,
+                                 const char *name)
+{
```
lum1n0us, line 3990: Please give `wasm_runtime_get_import_type()` a trial.

core/iwasm/common/wasm_runtime_common.c
wenyongh: There is already wasm_export_type_t/wasm_import_type_t, and the input of wasm_runtime_get_export_type/wasm_runtime_get_export_type is `uint32 export_index`/`uint32 import_index`, but here it wants to retrieve the func type info according to the name, we cannot pass name to the two APIs. One possible way is to add new APIs, like:
```C
WASM_RUNTIME_API_EXTERN void
wasm_runtime_get_export_type_by_name(const wasm_module_t module, const char *name,
                             wasm_export_type *export_type);
WASM_RUNTIME_API_EXTERN void
wasm_runtime_get_import_type_by_name(const wasm_module_t module, const char *name,
                             wasm_import_type *import_type);
```
And extend the structure wasm_import_type and wasm_export_type:
```C
typedef struct wasm_import_type {
    const char *module_name;
    const char *name;
    wasm_import_export_kind_t kind;
    bool linked;
    union {
        wasm_func_type_t func_type;  // func type when the import is a function
        ...  // other info for other import kind
    } u;
} wasm_import_type;

typedef struct wasm_export_type {
    const char *name;
    wasm_import_export_kind_t kind;
    union {
        wasm_func_type_t func_type;  // func type when the import is a function
        ...  // other info for other export kind
} wasm_export_type;
```
And then we get the param/result info from func_type in the result of wasm_export_type and wasm_import_type by using wasm_func_type_get_param_types and wasm_func_type_get_result_types.

wenyongh: BTW, @bnason-nf, could you change `wasm_export_type` and `wasm_import_type` to `wasm_export_type_t` and `wasm_import_type_t`? We usually added `_t` in the types declared in wasm_export.h.

bnason-nf: Hi @wenyongh, that seems to contradict what @lum1n0us said here:
https://github.com/bytecodealliance/wasm-micro-runtime/pull/3330#discussion_r1569823168

bnason-nf:
Thanks @lum1n0us  and @wenyongh, I like your suggestion to re-use the import/export type I added before to also expose the function type, so I've made that change, along with trying to address your other comments.

I'm still a bit confused about wasm_import_type vs. wasm_import_type_t (same for export), but I switched to the _t style as requested.

wenyongh:
> Thanks @lum1n0us and @wenyongh, I like your suggestion to re-use the import/export type I added before to also expose the function type, so I've made that change, along with trying to address your other comments.
> 
> I'm still a bit confused about wasm_import_type vs. wasm_import_type_t (same for export), but I switched to the _t style as requested.

Yes, maybe we can just use wasm_import_t/wasm_export_t? It looks clear and concise. I just think we had better use `_t` style like others, if you and @lum1n0us prefer to using wasm_import_type/wasm_export_type, it is also good to me.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3929,11 +3937,108 @@ wasm_runtime_get_export_type(WASMModuleCommon *const module, int32 export_index,
...
+                                - wasm_module->import_function_count]
+                    ->func_type;
+        }
         return;
     }
 #endif
 }
 
+int32
+wasm_runtime_function_get_param_count(WASMFuncType *const function)
```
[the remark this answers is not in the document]
wenyongh: @bnason-nf I got a new idea, here we just want to get the value kind of the func type's params/results, so how about changing the APIs to be like:
```C
uint32
wasm_func_type_get_param_count(WASMFuncType *const func_type);

wasm_valkind_t
wasm_func_type_get_param_valkind(WASMFuncType *const func_type,
                                 uint32 param_index);

uint32
wasm_func_type_get_result_count(WASMFuncType *const func_type);

wasm_valkind_t
wasm_func_type_get_result_valkind(WASMFuncType *const func_type,
                                  uint32 param_index)
```
Note that `wasm_func_type_get_param_count` and `wasm_func_type_get_result_count` are already declared in gc_export.h and implemented in gc_common.c. We may move their implementation to wasm_runtime_common.c instead.

bnason-nf:
Hi @wenyongh, I'm fine with the approach of moving the functions from gc_export into wasm_export. I think it would require also moving these types into wasm_export.h, are you ok with that?

```
typedef uint8_t wasm_value_type_t;

typedef enum wasm_value_type_enum {
    VALUE_TYPE_I32 = 0x7F,
    VALUE_TYPE_I64 = 0x7E,
    VALUE_TYPE_F32 = 0x7D,
    VALUE_TYPE_F64 = 0x7C,
    VALUE_TYPE_V128 = 0x7B,
    /* GC Types */
    VALUE_TYPE_I8 = 0x78,
    VALUE_TYPE_I16 = 0x77,
    VALUE_TYPE_NULLFUNCREF = 0x73,
    VALUE_TYPE_NULLEXTERNREF = 0x72,
    VALUE_TYPE_NULLREF = 0x71,
    VALUE_TYPE_FUNCREF = 0x70,
    VALUE_TYPE_EXTERNREF = 0x6F,
    VALUE_TYPE_ANYREF = 0x6E,
    VALUE_TYPE_EQREF = 0x6D,
    VALUE_TYPE_I31REF = 0x6C,
    VALUE_TYPE_STRUCTREF = 0x6B,
    VALUE_TYPE_ARRAYREF = 0x6A,
    VALUE_TYPE_HT_NON_NULLABLE_REF = 0x64,
    VALUE_TYPE_HT_NULLABLE_REF = 0x63,
    /* Stringref Types */
    VALUE_TYPE_STRINGREF = 0X67,
    VALUE_TYPE_STRINGVIEWWTF8 = 0x66,
    VALUE_TYPE_STRINGVIEWWTF16 = 0x62,
    VALUE_TYPE_STRINGVIEWITER = 0x61
} wasm_value_type_enum;

typedef int32_t wasm_heap_type_t;

typedef enum wasm_heap_type_enum {
    HEAP_TYPE_FUNC = -0x10,
    HEAP_TYPE_EXTERN = -0x11,
    HEAP_TYPE_ANY = -0x12,
    HEAP_TYPE_EQ = -0x13,
    HEAP_TYPE_I31 = -0x16,
    HEAP_TYPE_NOFUNC = -0x17,
    HEAP_TYPE_NOEXTERN = -0x18,
    HEAP_TYPE_STRUCT = -0x19,
    HEAP_TYPE_ARRAY = -0x1A,
    HEAP_TYPE_NONE = -0x1B
} wasm_heap_type_enum;
```

bnason-nf:
Oh, unless you meant that we can change the signature of those functions, since nothing else seems to be using them?

wenyongh:
> Oh, unless you meant that we can change the signature of those functions, since nothing else seems to be using them?

@bnason-nf sorry to make you confused, my idea is that:
1. keep `wasm_func_type_get_param_type` and `wasm_func_type_get_result_type` unchanged in both gc_export.h and gc_common.c, they return structure wasm_ref_type_t and have been used by other developers.

2. add two new APIs, declare them in wasm_export.h and implement them in wasm_common.c:
```C
wasm_valkind_t
wasm_func_type_get_param_valkind(WASMFuncType *const func_type,
                                 uint32 param_index);
wasm_valkind_t
wasm_func_type_get_result_valkind(WASMFuncType *const func_type,
                                  uint32 param_index)
```
  Note the return type is wasm_valkind_t or uint8, it is different from structure wasm_ref_type_t.

3. move `wasm_func_type_get_param_count` and `wasm_func_type_get_return_count` from gc_export.h and gc_common.c to wasm_export.h and wasm_common.c. Or redeclare them in wasm_export.h and re-implement them in wasm_common.c, but in wasm_common.c, we should add macro control `#if WASM_ENABLE_GC == 0 ... #endif` for them to avoid duplicated implementation when GC is enabled..

4. change wasm_import_type/wasm_export_type to wasm_import_t/wasm_export_t, and add `union { wasm_func_type_t func_type; } u;`  in them. Just like what you have done.

If it is good to you, could you please update the code again? 😊

bnason-nf:
Thanks for explaining so well @wenyongh, I think it's very clear now. I've made the changes you suggested, with just a couple very minor notes:

- I assume you meant wasm_runtime_common.c when you said wasm_common.c since I don't see that file. If not please let me know what I'm missing.
- I changed the return type of wasm_func_type_get_param_count() and wasm_func_type_get_result_count() to be signed to allow for error handling, but it's not a big benefit so please let me know if you prefer to keep the signature identical to avoid compatibility issues with any existing users of the existing GC APIs.

bnason-nf:
Just a note that I rebased and squashed my commits.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3929,11 +3937,108 @@ wasm_runtime_get_export_type(WASMModuleCommon *const module, int32 export_index,
...
+            return WASM_I64;
+        case VALUE_TYPE_F32:
+            return WASM_F32;
+        case VALUE_TYPE_F64:
+            return WASM_F64;
+        case VALUE_TYPE_FUNCREF:
+            return WASM_FUNCREF;
+
+        case VALUE_TYPE_V128:
+        case VALUE_TYPE_EXTERNREF:
```
wenyongh, line 3985: The latest wasm-c-api wasm.h has changed wasm_valkind_enum to:
```C
enum wasm_valkind_enum {
  WASM_I32,
  WASM_I64,
  WASM_F32,
  WASM_F64,
  WASM_EXTERNREF = 128,
  WASM_FUNCREF,
};
```
https://github.com/WebAssembly/wasm-c-api/blob/2ce1367c9d1271c83fb63bef26d896a2f290cd23/include/wasm.h#L180-L187

We had better align to it and also support SIMD VALUE_TYPE_V128, so I think we had better change it to below in both wasm-c-api.h and wasm_export.h:
```C
enum wasm_valkind_enum {
  WASM_I32,
  WASM_I64,
  WASM_F32,
  WASM_F64,
  WASM_V128,
  WASM_VOID,
  WASM_EXTERNREF = 128,
  WASM_FUNCREF,
};
```
There may be many places to change WASM_ANYREF to WASM_EXTERNREF inside runtime src and samples, it is up to you to decide whether to change it or not. But here had better also handle VALUE_TYPE_V128 and VALUE_TYPE_EXTERNREF:
```C
#if WASM_ENABLE_SIMD != 0
    case VALUE_TYPE_V128:
        return WASM_V128;
#endif
    case VALUE_TYPE_EXTERNREF:
        return WASM_EXTERNREF;  (or WASM_ANYREF if it isn't changed)
#endif
```

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3929,11 +3937,108 @@ wasm_runtime_get_export_type(WASMModuleCommon *const module, int32 export_index,
...
+                    ->functions[wasm_export->index
+                                - wasm_module->import_function_count]
+                    ->func_type;
+        }
         return;
     }
 #endif
 }
 
+int32
```
wenyongh, line 3952: Prefer not to change it from uint32 to int32, the API can require that the developer must input a valid func_type, like some libc APIs. It is the responsibility of the developer to pass a valid func_type to the API.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3929,11 +3937,108 @@ wasm_runtime_get_export_type(WASMModuleCommon *const module, int32 export_index,
...
+wasm_func_type_get_param_count(WASMFuncType *const func_type)
+{
+    if (!func_type) {
+        bh_assert(0);
+        return -1;
+    }
```
wenyongh, line 3958: can just `bh_assert(func_type);`

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3929,11 +3937,108 @@ wasm_runtime_get_export_type(WASMModuleCommon *const module, int32 export_index,
...
+        case VALUE_TYPE_VOID:
+        default:
+        {
+            bh_assert(0);
+            return (wasm_valkind_t)-1;
+        }
+    }
+}
+
+int32
```
wenyongh, line 3995: had better uint32

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3929,11 +3937,108 @@ wasm_runtime_get_export_type(WASMModuleCommon *const module, int32 export_index,
...
+wasm_func_type_get_result_count(WASMFuncType *const func_type)
+{
+    if (!func_type) {
+        bh_assert(0);
+        return -1;
+    }
```
wenyongh, line 4001: bh_assert(func_type);

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3929,11 +3937,108 @@ wasm_runtime_get_export_type(WASMModuleCommon *const module, int32 export_index,
...
+            return WASM_F64;
+        case VALUE_TYPE_FUNCREF:
+            return WASM_FUNCREF;
+
+#if WASM_ENABLE_SIMD != 0
+        case VALUE_TYPE_V128:
+#endif
+#if WASM_ENABLE_REF_TYPES != 0
+        case VALUE_TYPE_EXTERNREF:
+#endif
```
wenyongh, line 4032: same as above, return WASM_V128 and WASM_EXTERNREF for them.

core/iwasm/common/wasm_runtime_common.c
wenyongh: ps. no need to add `WASM_VOID` to `wasm_valkind_enum`, it isn't used

bnason-nf:
Hi @wenyongh, I restored the return types. If you don't mind, could we do the valkind enum normalization as a separate pull request?

core/iwasm/include/wasm_export.h
```diff
@@ -1255,7 +1264,51 @@ wasm_runtime_get_export_count(const wasm_module_t module);
  */
 WASM_RUNTIME_API_EXTERN void
 wasm_runtime_get_export_type(const wasm_module_t module, int32_t export_index,
-                             wasm_export_type *export_type);
+                             wasm_export_t *export_type);
+
+/**
+ * Get the number of parameters for a function
```
wenyongh, line 1270: had better `function` -> `function type`

core/iwasm/include/wasm_export.h
```diff
@@ -1255,7 +1264,51 @@ wasm_runtime_get_export_count(const wasm_module_t module);
...
+ *
+ * @param func_type the function
```
wenyongh, line 1272: ditto, same for many places below

core/iwasm/include/wasm_export.h
```diff
@@ -1255,7 +1264,51 @@ wasm_runtime_get_export_count(const wasm_module_t module);
...
+ *
+ * @return the number of parameters for the function
+ */
+WASM_RUNTIME_API_EXTERN uint32_t
+wasm_func_type_get_param_count(wasm_func_type_t const func_type);
+
+/**
+ * Get the kind of a parameter for a function
+ *
+ * @param func_type the function
```
wenyongh, line 1282: ditto

core/iwasm/include/wasm_export.h
```diff
@@ -1255,7 +1264,51 @@ wasm_runtime_get_export_count(const wasm_module_t module);
...
+
+/**
+ * Get the number of results for a function
+ *
+ * @param func_type the function
+ *
+ * @return the number of results for the function
+ */
+WASM_RUNTIME_API_EXTERN uint32_t
+wasm_func_type_get_result_count(wasm_func_type_t const function);
```
wenyongh, line 1299: had better `function` -> `func_type`

wenyongh (commented):
LGTM except minor comments

bnason-nf:
Thanks @wenyongh, I think I got all the spots you're talking about.

wenyongh (approved):

wenyongh:
> Thanks @wenyongh, I think I got all the spots you're talking about.

Welcome, I will merge it after the CI checks.

PR #3366, opened 2024-04-26
CI: use macos-13 instead of macos-latest

yamt:
our workflows assume macos-latest is intel. but now it's arm64. use macos-13, which is documented to be intel, instead.

reference:
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners

XXX what to do for test-tools/wamr-ide/VSCode-Extension/src/utilities/lldbUtilities.ts?

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -28,7 +28,7 @@ jobs:
     #   - https://gh.io/supported-runners-and-hardware-resources
     #   - https://gh.io/using-larger-runners
     # Consider using larger runners for possible analysis time improvements.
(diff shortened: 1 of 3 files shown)


lum1n0us (approved):
LGTM

wenyongh (approved):

wenyongh:
> our workflows assume macos-latest is intel. but now it's arm64. use macos-13, which is documented to be intel, instead.
> 
> reference: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners
> 
> XXX what to do for test-tools/wamr-ide/VSCode-Extension/src/utilities/lldbUtilities.ts?

Let's merge this PR first to address the CI issue, and check whether we should modify this file later.

Issue #3364, opened 2024-04-26
macos CI runs failed


wenyongh:
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/8841321431
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/8842970854

Seems that we had better add `--break-system-packages` flag for all macos OSs in the below command:
`/usr/bin/env python3 -m pip install -r requirements.txt`

yamt:
it seems a bit more complex than that.
our workflows assume macos-latest is intel. but now it's arm64.

yamt:
https://github.com/bytecodealliance/wasm-micro-runtime/pull/3366

PR #3367, opened 2024-04-26
Fix a pip instatllion failure in CI in macos-latest

lum1n0us:
Run /usr/bin/env python3 -m pip install -r requirements.txt error: externally-managed-environment

× This environment is externally managed
...


yamt:
https://github.com/bytecodealliance/wasm-micro-runtime/issues/3364#issuecomment-2078610806

PR #3375, opened 2024-04-29
Use uint32_t for value passed to memmove and memset

nchamzn:
On a 64 bit host the top bits of `n` here can be garbage (as the value is 32bit in the wasm memory).

I have promised @loganek I'll write some tests for this in `tests/wamr-compiler` but so far I haven't made a minimal reproduction, will update the PR when I've done that.

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -2957,13 +2957,13 @@ aot_check_app_addr_and_convert(AOTModuleInstance *module_inst, bool is_str,
 }
 
 void *
-aot_memmove(void *dest, const void *src, size_t n)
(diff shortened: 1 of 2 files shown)


nchamzn:
Hi, thanks for looking - yes your suggestion looks good to me

wenyongh:
> Hi, thanks for looking - yes your suggestion looks good to me

Welcome, I submitted PR #3378 to fix it with more comments added. Could you help check again? I think maybe we can merge it first if it is good and don't block the AOT memory64 feature (#3362) that @loganek is enabling, and then you can also update the tests in `tests/wamr-compiler`.

wenyongh:
Close this PR since the issue was fixed with https://github.com/bytecodealliance/wasm-micro-runtime/pull/3378.

Issue #3354, opened 2024-04-24
On memory usage when loading a Wasm file with fast interpreter


nchamzn:
I did some memory profiling when loading Wasm files using the fast interpreter and noticed that not only do we process the Wasm binary to an internal representation, but we also keep the original Wasm binary itself in memory (And have references to it in the internal representation). 

Is there any scope to optimise this, e.g. by copying only relevant sections from the Wasm binary so that we can free the rest of the buffer?  Are there any known optimisations that can be made to those internal structures to reduce the memory usage of Wamr?

wenyongh:
Hi, if the whole wasm file is loaded by wasm_runtime_load or wasm_runtime_load_ex, it is not easy to free the buffer of wasm file since some fields in the WASMModule may be referring to it. But if the wasm file is read section by section into a wasm_section_list_t and the host embedder allocates different buffers for each section, and the sections are loaded by [wasm_runtime_load_from_sections](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/include/wasm_export.h#L463-L475), then after loading, some of the sections may be freed. You can also refer to the code in wamr-app-framework rep:
https://github.com/bytecodealliance/wamr-app-framework/blob/main/app-mgr/app-manager/module_wasm_app.c#L787-L788

Note that it only frees SECTION_TYPE_DATA section, but for fast-interp, maybe you can try to free SECTION_TYPE_FUNC also.

nchamzn:
Thanks, I just had a go with this suggestion and it works nicely, however some of the code linked is using wamr internals (E.g. os_mmap, many compiler definitions etc) and my implementation involved a lot of copy/paste.

Would you be OK with a PR which added:

```
destroy_all_wasm_sections
destroy_part_wasm_sections
destroy_all_aot_sections
destroy_part_aot_sections
```

and then another API something like:

```
struct load_section_result {
    wasm_section_list_t sections;
    bool is_aot;
};

read_to_section_result read_to_sections(uint8* data, int size)
```

exported from the wamr library?

wenyongh:
> Thanks, I just had a go with this suggestion and it works nicely, however some of the code linked is using wamr internals (E.g. os_mmap, many compiler definitions etc) and my implementation involved a lot of copy/paste.
> 
> Would you be OK with a PR which added:
> 
> ```
> destroy_all_wasm_sections
> destroy_part_wasm_sections
> destroy_all_aot_sections
[...]

PR #3377, opened 2024-04-30
Add APIs to read and destroy sections

eloparco:
#3354

### DIFF
--- a/core/iwasm/aot/aot_loader.c
+++ b/core/iwasm/aot/aot_loader.c
@@ -3875,11 +3875,14 @@ aot_load_from_sections(AOTSection *section_list, char *error_buf,
     if (!module)
         return NULL;
 
+    os_thread_jit_write_protect_np(false); /* Make memory writable */
     if (!load_from_sections(module, section_list, false, error_buf,
                             error_buf_size)) {
         aot_unload(module);
         return NULL;
     }
+    os_thread_jit_write_protect_np(true); /* Make memory executable */
+    os_icache_flush(module->code, module->code_size);
 
     LOG_VERBOSE("Load module from sections success.\n");
     return module;
@@ -4086,10 +4089,16 @@ bool
 aot_read_to_sections(const uint8 *buf, uint32 size, AOTSection **p_section_list,
                      char *error_buf, uint32 error_buf_size)
 {
+    bool ret = false;
     AOTModule
         fake_module; /* Unused, passed since required by create_sections */
-    return create_sections(&fake_module, buf, size, true, p_section_list,
-                           error_buf, error_buf_size);
+
+    os_thread_jit_write_protect_np(false); /* Make memory writable */
+    ret = create_sections(&fake_module, buf, size, true, p_section_list,
+                          error_buf, error_buf_size);
+    os_thread_jit_write_protect_np(true); /* Make memory executable */
+
+    return ret;
 }
 
 static bool
--- a/core/iwasm/aot/aot_loader.c
+++ b/core/iwasm/aot/aot_loader.c
@@ -3886,15 +3886,21 @@ aot_load_from_sections(AOTSection *section_list, char *error_buf,
 }
 
 static void
-destroy_sections(AOTSection *section_list, bool destroy_aot_text)
+destroy_sections(AOTSection *section_list, bool free_section_body,
+                 bool destroy_aot_text)
 {
     AOTSection *section = section_list, *next;
     while (section) {
         next = section->next;
         if (destroy_aot_text && section->section_type == AOT_SECTION_TYPE_TEXT
-            && section->section_body)
+            && section->section_body) {
             os_munmap((uint8 *)section->section_body,
                       section->section_body_size);
+        }
+        else if (free_section_body && section->section_body) {
+            wasm_runtime_free(section->section_body);
+        }
+
         wasm_runtime_free(section);
         section = next;
     }
@@ -3941,8 +3947,8 @@ fail:
 
 static bool
 create_sections(AOTModule *module, const uint8 *buf, uint32 size,
-                AOTSection **p_section_list, char *error_buf,
-                uint32 error_buf_size)
+                bool alloc_mem_for_section_body, AOTSection **p_section_list,
+                char *error_buf, uint32 error_buf_size)
 {
     AOTSection *section_list = NULL, *section_list_end = NULL, *section;
     const uint8 *p = buf, *p_end = buf + size;
@@ -4035,6 +4041,17 @@ create_sections(AOTModule *module, const uint8 *buf, uint32 size,
                     }
                 }
             }
+            else {
+                if (alloc_mem_for_section_body) {
+                    section->section_body = loader_malloc(
+                        section->section_body_size, error_buf, error_buf_size);
+                    if (!section->section_body)
+                        goto fail;
+
+                    bh_memcpy_s(section->section_body,
+                                section->section_body_size, p, section_size);
+                }
+            }
 
             if (!section_list)
                 section_list = section_list_end = section;
@@ -4060,8 +4077,19 @@ create_sections(AOTModule *module, const uint8 *buf, uint32 size,
     return true;
 fail:
     if (section_list)
-        destroy_sections(section_list, destroy_aot_text);
+        destroy_sections(section_list, alloc_mem_for_section_body,
+                         destroy_aot_text);
     return false;
+}
+
+bool
+aot_read_to_sections(const uint8 *buf, uint32 size, AOTSection **p_section_list,
+                     char *error_buf, uint32 error_buf_size)
+{
+    AOTModule
+        fake_module; /* Unused, passed since required by create_sections */
+    return create_sections(&fake_module, buf, size, true, p_section_list,
+                           error_buf, error_buf_size);
 }
 
 static bool
@@ -4086,7 +4114,7 @@ load(const uint8 *buf, uint32 size, AOTModule *module, char *error_buf,
         return false;
     }
 
-    if (!create_sections(module, buf, size, &section_list, error_buf,
+    if (!create_sections(module, buf, size, false, &section_list, error_buf,
                          error_buf_size))
         return false;
 
@@ -4095,14 +4123,15 @@ load(const uint8 *buf, uint32 size, AOTModule *module, char *error_buf,
     if (!ret) {
         /* If load_from_sections() fails, then aot text is destroyed
            in destroy_sections() */
-        destroy_sections(section_list, module->is_indirect_mode ? false : true);
+        destroy_sections(section_list, false,
+                         module->is_indirect_mode ? false : true);
         /* aot_unload() won't destroy aot text again */
         module->code = NULL;
     }
     else {
         /* If load_from_sections() succeeds, then aot text is set to
            module->code and will be destroyed in aot_unload() */
-        destroy_sections(section_list, false);
+        destroy_sections(section_list, false, false);
     }
 
 #if 0
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -12,6 +12,7 @@
 #include "wasm_memory.h"
 #if WASM_ENABLE_INTERP != 0
 #include "../interpreter/wasm_runtime.h"
+#include "wasm_loader.h"
 #endif
 #if WASM_ENABLE_AOT != 0
 #include "../aot/aot_runtime.h"
@@ -7072,7 +7073,7 @@ wasm_runtime_detect_native_stack_overflow_size(WASMExecEnv *exec_env,
     return true;
 }
 
-#if WASM_ENABLE_INTERP != 0 || WASM_ENABLE_JIT != 0
+#if WASM_ENABLE_INTERP != 0
 static void
 destroy_all_wasm_sections(wasm_section_list_t sections)
 {
@@ -7172,11 +7173,11 @@ destroy_part_aot_sections(aot_section_list_t *p_sections, uint8 *section_types,
 
 void
 wasm_runtime_destroy_sections(load_section_result_t *section_result,
-                              uint8_t *section_types, int section_cnt)
+                              uint8 *section_types, int section_cnt)
 {
     bool destroy_all_sections = (section_types == NULL);
     if (destroy_all_sections) {
-#if WASM_ENABLE_INTERP != 0 || WASM_ENABLE_JIT != 0
+#if WASM_ENABLE_INTERP != 0
         if (!section_result->is_aot)
             destroy_all_wasm_sections(section_result->sections);
 #endif
@@ -7186,7 +7187,7 @@ wasm_runtime_destroy_sections(load_section_result_t *section_result,
 #endif
     }
     else {
-#if WASM_ENABLE_INTERP != 0 || WASM_ENABLE_JIT != 0
+#if WASM_ENABLE_INTERP != 0
         if (!section_result->is_aot)
             destroy_part_wasm_sections(&section_result->sections, section_types,
                                        section_cnt);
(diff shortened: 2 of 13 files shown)


yamt:
is this basically a copy from wamr-app-framework?

eloparco:
> is this basically a copy from wamr-app-framework?

Yes, I should have mentioned it in the PR description. 

`destroy_all_wasm_sections`, `destroy_all_aot_sections`, `destroy_part_wasm_sections` and `destroy_part_aot_sections` are exactly the same. `wasm_runtime_read_to_sections` is adapted from `wasm_app_module_on_install_request_byte_arrive`.

We want to be able to free as many resources that are not necessary anymore after module instantiation or loading (especially for constrained embedded devices).

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -7071,3 +7071,495 @@ wasm_runtime_detect_native_stack_overflow_size(WASMExecEnv *exec_env,
...
+
+void
+wasm_runtime_destroy_sections(load_section_result_t *section_result,
+                              uint8_t *section_types, int section_cnt)
+{
+    bool destroy_all_sections = (section_types == NULL);
+    if (destroy_all_sections) {
+        section_result->is_aot
+            ? destroy_all_aot_sections(section_result->sections)
+            : destroy_all_wasm_sections(section_result->sections);
```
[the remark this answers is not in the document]
eloparco: Doing that already, this comment is on an outdated line of code

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -7071,3 +7071,505 @@ wasm_runtime_detect_native_stack_overflow_size(WASMExecEnv *exec_env,
...
+        .total_received_size = 0,
+        .magic = 0,
+        .version = 0,
+        .sections = NULL,
+        .section_end = NULL,
+        .package_type = Package_Type_Unknown,
+    };
+
+    for (uint64_t i = 0; i < size; i++) {
+        bool ret = wasm_section_parse_byte(buf[i], size, &section_ctx);
```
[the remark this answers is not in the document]
eloparco: Makes sense, I pushed a commit to reuse the logic from the wasm and aot loaders

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -7071,3 +7072,158 @@ wasm_runtime_detect_native_stack_overflow_size(WASMExecEnv *exec_env,
...
+        while (cur) {
+            wasm_section_t *next = cur->next;
+            if (cur->section_type == section_type) {
+                if (prev)
+                    prev->next = next;
+                else
+                    *p_sections = next;
+
+                if (cur->section_body != NULL)
+                    wasm_runtime_free(cur->section_body);
```
wenyongh, line 7108: The section_body is just referring to the wasm file buf, so if we didn't add a new argument for create_sections_wasm and didn't allocate memory for section_body, here we cannot free it.

core/iwasm/aot/aot_loader.c
```diff
@@ -4064,6 +4064,15 @@ create_sections(AOTModule *module, const uint8 *buf, uint32 size,
     return false;
 }
 
+bool
+create_sections_aot(const uint8 *buf, uint32 size, AOTSection **p_section_list,
+                    char *error_buf, uint32 error_buf_size)
+{
+    AOTModule fake_module; // Unused, passed since required by create_sections
+    return create_sections(&fake_module, buf, size, p_section_list, error_buf,
+                           error_buf_size);
```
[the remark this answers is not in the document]
eloparco: Right, I ran into the problem this morning when I was trying the code from an embedded with a sanitizer and I was getting an `use-after-free` error.

I'll adjust the code. Probably worth adding an example showing the usage as well.

core/iwasm/aot/aot_loader.c
```diff
@@ -4035,6 +4035,17 @@ create_sections(AOTModule *module, const uint8 *buf, uint32 size,
                     }
                 }
             }
+            else {
+                if (alloc_mem_for_section_body) {
+                    section->section_body = loader_malloc(
+                        section->section_body_size, error_buf, error_buf_size);
```
wenyongh, line 4041: Should also modify `destroy_sections`, which don't free section_body now (only munmap section_body when it is aot text and destroy_aot_text is true). Had better add a new argument `bool free_section_body`, and pass `alloc_mem_for_section_body` to it in this function:
```C
static void
destroy_sections(AOTSection *section_list, bool free_section_body,
                 bool destroy_aot_text)
```

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -7071,3 +7072,158 @@ wasm_runtime_detect_native_stack_overflow_size(WASMExecEnv *exec_env,
...
+        if (section_result->is_aot)
+            destroy_part_aot_sections(&section_result->sections, section_types,
+                                      section_cnt);
+#endif
+    }
+}
+
+bool
+wasm_runtime_read_to_sections(uint8_t *buf, uint64_t size,
+                              load_section_result_t *p_section_result)
```
wenyongh, line 7205: Had better add argument `char *error_buf, uint32 error_buf_size` to this API like other APIs.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -7071,3 +7072,158 @@ wasm_runtime_detect_native_stack_overflow_size(WASMExecEnv *exec_env,
...
+{
+    bool ret = false;
+    char error_buf[1024];
```
wenyongh, line 7208: Had better move to function argument instead.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -7071,3 +7072,158 @@ wasm_runtime_detect_native_stack_overflow_size(WASMExecEnv *exec_env,
...
+#if WASM_ENABLE_AOT != 0
+        if (section_result->is_aot)
+            destroy_part_aot_sections(&section_result->sections, section_types,
+                                      section_cnt);
+#endif
+    }
+}
+
+bool
+wasm_runtime_read_to_sections(uint8_t *buf, uint64_t size,
```
wenyongh, line 7204: Had better use `uint8` and `uint64` inside core/iwasm

core/iwasm/common/wasm_runtime_common.h
```diff
@@ -1204,6 +1204,14 @@ void
 wasm_runtime_set_linux_perf(bool flag);
 #endif
 
+bool
+wasm_runtime_read_to_sections(uint8_t *buf, uint64_t size,
```
wenyongh, line 1208: Had better `uint8`, `uint64`

core/iwasm/common/wasm_runtime_common.h
```diff
@@ -1204,6 +1204,14 @@ void
...
+                              load_section_result_t *p_section_result);
+
+void
+wasm_runtime_destroy_sections(load_section_result_t *section_result,
+                              uint8_t *section_types, int section_cnt);
```
wenyongh, line 1213: Had better `uint8`

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -7071,3 +7072,158 @@ wasm_runtime_detect_native_stack_overflow_size(WASMExecEnv *exec_env,
...
+                cur = next;
+            }
+        }
+    }
+}
+#endif
+
+void
+wasm_runtime_destroy_sections(load_section_result_t *section_result,
+                              uint8_t *section_types, int section_cnt)
```
wenyongh, line 7176: Had better uint8

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -6333,8 +6334,19 @@ create_sections(const uint8 *buf, uint32 size, WASMSection **p_section_list,
             }
 
             section->section_type = section_type;
-            section->section_body = (uint8 *)p;
             section->section_body_size = section_size;
+            if (alloc_mem_for_section_body) {
+                section->section_body = loader_malloc(
+                    section->section_body_size, error_buf, error_buf_size);
```
wenyongh, line 6340: Same as aot, should change destroy_sections in this file.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -7071,3 +7072,158 @@ wasm_runtime_detect_native_stack_overflow_size(WASMExecEnv *exec_env,
...
+wasm_runtime_read_to_sections(uint8_t *buf, uint64_t size,
+                              load_section_result_t *p_section_result)
+{
+    bool ret = false;
+    char error_buf[1024];
+
+#if WASM_ENABLE_INTERP != 0
+    if (get_package_type(buf, size) == Wasm_Module_Bytecode) {
+        ret = wasm_read_to_sections(buf, size, &p_section_result->sections,
+                                    error_buf, sizeof(error_buf));
```
wenyongh, line 7213: Should use `error_buf, error_buf_size` if the API prototype is changed

core/iwasm/aot/aot_loader.c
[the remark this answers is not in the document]
eloparco: By the way, I was able to release all wasm sections after module instantiation when testing in fast interpreter mode, including the code section that is left out here https://github.com/bytecodealliance/wamr-app-framework/blob/main/app-mgr/app-manager/module_wasm_app.c#L728-L746

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -7071,3 +7072,158 @@ wasm_runtime_detect_native_stack_overflow_size(WASMExecEnv *exec_env,
...
+        p_section_result->is_aot = false;
+    }
+#endif
+#if WASM_ENABLE_AOT != 0
+    if (get_package_type(buf, size) == Wasm_Module_AoT) {
+        ret = aot_read_to_sections(buf, size, &p_section_result->sections,
+                                   error_buf, sizeof(error_buf));
```
wenyongh, line 7220: Same as above

core/iwasm/aot/aot_loader.c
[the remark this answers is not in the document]
wenyongh: Good, and yes, it would be great if a sample is added. And we can also mention the method in the document [Tune the memory usage](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/doc/memory_tune.md#tune-the-memory-usage), it should benefit to some developers of the Embdded/IoT.

core/iwasm/interpreter/wasm_loader.c
eloparco: Is it needed? `destroy_sections` is only called by `load`, not by `create_sections`

core/iwasm/aot/aot_loader.c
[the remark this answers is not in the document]
eloparco: Great, let me do it

core/iwasm/interpreter/wasm_loader.c
wenyongh: It is to avoid memory leak, `destroy_sections` is called when `create_sections` failed: some section_bodies may have been allocated when a section body is allocated failed.

core/iwasm/aot/aot_loader.c
```diff
@@ -3886,15 +3886,21 @@ aot_load_from_sections(AOTSection *section_list, char *error_buf,
...
         next = section->next;
         if (destroy_aot_text && section->section_type == AOT_SECTION_TYPE_TEXT
-            && section->section_body)
+            && section->section_body) {
             os_munmap((uint8 *)section->section_body,
                       section->section_body_size);
+        }
+        else {
+            if (free_section_body)
+                wasm_runtime_free(section->section_body);
```
wenyongh, line 3902: Had better
```C
    else if (free_section_body && section->section_body) {
        wasm_runtime_free(section->section_body);
    }
```

yamt:
> > is this basically a copy from wamr-app-framework?
> 
> Yes, I should have mentioned it in the PR description.

please update the description then.

> `destroy_all_wasm_sections`, `destroy_all_aot_sections`, `destroy_part_wasm_sections` and `destroy_part_aot_sections` are exactly the same. `wasm_runtime_read_to_sections` is adapted from `wasm_app_module_on_install_request_byte_arrive`.
> 
> We want to be able to free as many resources that are not necessary anymore after module instantiation or loading (especially for constrained embedded devices).

if you just want to save some memory, why do you need this complicated api?
i suspect what you need is an option to wasm_runtime_load etc.

yamt:
sometimes it's useful to have an api to examine sections. (especially custom sections)
eg. https://github.com/yamt/toywasm/tree/master/libdyld#dylink0-custom-section
but it isn't the purpose of this PR, is it?

core/iwasm/interpreter/wasm_loader.c
wenyongh: I know why now, in this function, when creating section failed, it doesn't free the section nodes created like AOT, and the below code in function load will call destroy_sections to free them even when `create_sections` failed:
```C
    if (!create_sections(buf, size, false, &section_list, error_buf,
                         error_buf_size)
        || !load_from_sections(module, section_list, true, error_buf,
                               error_buf_size)) {
        destroy_sections(section_list);
        return false;
    }
```
But I think it is not a good coding style, and it will lead to memory leak in the wasm_read_to_sections, which doesn't call destroy_sections when creating sections failed:
```C
    return create_sections(buf, size, true, p_section_list, error_buf,
                           error_buf_size);
```
My suggestion is to do like aot: call `destroy_sections` in the lable `fail` when failing to `create_sections`, and change `return false` in this function to `goto fail`. And also add an argument `free_section_body` like aot.

And change below code to be like:
```C
    if (!create_sections(buf, size, false, &section_list, error_buf,
                         error_buf_size))
        return false;

    if (!load_from_sections(module, section_list, true, error_buf,
                               error_buf_size)) {
        destroy_sections(section_list, false); //free_section_body = 
        return false;
    }

    destroy_sections(section_list, false); //free_section_body = false
    return true;
```

wenyongh:
> > > is this basically a copy from wamr-app-framework?
> > 
> > 
> > Yes, I should have mentioned it in the PR description.
> 
> please update the description then.
> 
> > `destroy_all_wasm_sections`, `destroy_all_aot_sections`, `destroy_part_wasm_sections` and `destroy_part_aot_sections` are exactly the same. `wasm_runtime_read_to_sections` is adapted from `wasm_app_module_on_install_request_byte_arrive`.
> > We want to be able to free as many resources that are not necessary anymore after module instantiation or loading (especially for constrained embedded devices).
> 
> if you just want to save some memory, why do you need this complicated api? i suspect what you need is an option to wasm_runtime_load etc.

wasm_runtime_load has been used for a long time, we had better not change its prototype to add an option? And we don't know what the wasm/aot file buf is allocated, normally it can be freed only after wasm_runtime_load, so if we also create sections in wasm_runtime_load and allocate memory for `section->section_body`, there will be almost two copies of wasm/aot file at the same time together with the memory for module data structures, the peak memory usage may be large (though many of them may be freed later) which may not meet the requirement of some embedded/iot devices. And since we have documented that the buffer for wasm_runtime_load is referred after loading, it is an ABI change if we change the behavior and developer may not know that the buf can freed. So I think it is reasonable to provide another way, the PR just introduces two API `wasm_runtime_read_to_sections` and `wasm_runtime_destroy_sections`, it should be acceptable with document and sample added.

yamt:
> > > > is this basically a copy from wamr-app-framework?
> > > 
> > > 
> > > Yes, I should have mentioned it in the PR description.
> > 
> > 
> > please update the description then.
> > > `destroy_all_wasm_sections`, `destroy_all_aot_sections`, `destroy_part_wasm_sections` and `destroy_part_aot_sections` are exactly the same. `wasm_runtime_read_to_sections` is adapted from `wasm_app_module_on_install_request_byte_arrive`.
> > > We want to be able to free as many resources that are not necessary anymore after module instantiation or loading (especially for constrained embedded devices).
> > 
> > 
> > if you just want to save some memory, why do you need this complicated api? i suspect what you need is an option to wasm_runtime_load etc.
> 
> wasm_runtime_load has been used for a long time, we had better not change its prototype to add an option?

we have introduced wasm_runtime_load_ex very recently.

> And we don't know what the wasm/aot file buf is allocated, normally it can be freed only after wasm_runtime_load, so if we also create sections in wasm_runtime_load and allocate memory for `section->section_body`, there will be almost two copies of wasm/aot file at the same time together with the memory for module data structures, the peak memory usage may be large (though many of them may be freed later) which may not meet the requirement of some embedded/iot devices. And since we have documented that the buffer for wasm_runtime_load is referred after loading, it is an ABI change if we change the behavior and developer may not know that the buf can freed. So I think it is reasonable to provide another way, the PR just introduces two API `wasm_runtime_read_to_sections` and `wasm_runtime_destroy_sections`, it should be acceptable with document and sample added.

iiuc, there are three memory allocations involved:

1. user buffer (as large as the module)
2. sections (almost same size)
3. loaded module/instance

with this PR, the peak memory usage is max(1+2, 2+3), right?

if the peak memory usage is a problem, it's better to introduce a stream-style api
so that a user can use a fixed-sized buffer.
wasm module format is designed to be easily loaded that way.
that is,

1. user buffer (1 byte at minimum)
2. loaded module/instance

and/or, maybe we can simply make a user specify a "free the buffer" callback so that some buffers can be freed a bit earlier.

samples/basic/src/read_sections.c
```diff
@@ -0,0 +1,137 @@
...
+
+    module_inst = wasm_runtime_instantiate(module, stack_size, heap_size,
+                                           error_buf, sizeof(error_buf));
+    if (!module_inst) {
+        printf("Instantiate wasm module failed. error: %s.\n", error_buf);
+        goto fail;
+    }
+
+    /* Destroy all sections after instantiation. For AOT, the text sections must
+     * be kept because referenced during execution */
```
yamt, line 116: IMO, an api should not hardcode this kind of implementation details.

wenyongh:
> > > > > is this basically a copy from wamr-app-framework?
> > > > 
> > > > 
> > > > Yes, I should have mentioned it in the PR description.
> > > 
> > > 
> > > please update the description then.
> > > > `destroy_all_wasm_sections`, `destroy_all_aot_sections`, `destroy_part_wasm_sections` and `destroy_part_aot_sections` are exactly the same. `wasm_runtime_read_to_sections` is adapted from `wasm_app_module_on_install_request_byte_arrive`.
> > > > We want to be able to free as many resources that are not necessary anymore after module instantiation or loading (especially for constrained embedded devices).
> > > 
> > > 
> > > if you just want to save some memory, why do you need this complicated api? i suspect what you need is an option to wasm_runtime_load etc.
> > 
> > 
> > wasm_runtime_load has been used for a long time, we had better not change its prototype to add an option?
> 
> we have introduced wasm_runtime_load_ex very recently.

Yes, it is good to modify wasm_runtime_load_ex only, we may add an option for wasm_runtime_load_ex to choose whether to create sections, load and free sections. And even add another option of freeing buffer callback to allow wasm_runtime_load_ex to free the buffer, so it can create sections, free wasm/aot file buffer, load and free sections to reduce the peak memory usage.

> 
> > And we don't know what the wasm/aot file buf is allocated, normally it can be freed only after wasm_runtime_load, so if we also create sections in wasm_runtime_load and allocate memory for `section->section_body`, there will be almost two copies of wasm/aot file at the same time together with the memory for module data structures, the peak memory usage may be large (though many of them may be freed later) which may not meet the requirement of some embedded/iot devices. And since we have documented that the buffer for wasm_runtime_load is referred after loading, it is an ABI change if we change the behavior and developer may not know that the buf can freed. So I think it is reasonable to provide another way, the PR just introduces two API `wasm_runtime_read_to_sections` and `wasm_runtime_destroy_sections`, it should be acceptable with document and sample added.
> 
> iiuc, there are three memory allocations involved:
> 
> 1. user buffer (as large as the module)
> 2. sections (almost same size)
> 3. loaded module/instance
> 
> with this PR, the peak memory usage is max(1+2, 2+3), right?

Yes, the user buffer can be freed after the sections are created.

> 
> if the peak memory usage is a problem, it's better to introduce a stream-style api so that a user can use a fixed-sized buffer. wasm module format is designed to be easily loaded that way. that is,
> 
> 1. user buffer (1 byte at minimum)
> 2. loaded module/instance
> 
> and/or, maybe we can simply make a user specify a "free the buffer" callback so that some buffers can be freed a bit earlier.

How about letting the developer pass a read stream callback to read n bytes from the wasm/aot file, somewhat like
```C
typedef uint32 (*read_file_callback_t)(char *buf, uint32 count);  //read count bytes to buffer

wasm_module_t
wasm_runtime_load_from_stream(read_file_callback_t, char *error_buf, uint32_t error_buf_size);
```
Or just extend wasm_runtime_load_ex to allow developer to pass the read file callback to LoadArgs structure.

Then the loader can then read and create sections first (read section type and section size, create section body with section size, and then read data to section body), and then load from sections, then destroy sections after loading. In this way, we can reduce both the peak memory usage and memory usage after loading.

eloparco:
The specific problem I'm trying to solve is the one described in #3354.

I'm using the wasm-c-api and module and module instance are saved into the store, that can't be freed until the end of wasm execution. And the module in the store has a copy of the module binary https://github.com/bytecodealliance/wasm-micro-runtime/blob/c0e33f08b04a5e03220ef74c2a9dbea8f1f35996/core/iwasm/common/wasm_c_api.c#L2290 which is kept until the store is destroyed.

By using the APIs described in this PR, I can free the sections after module instantiation. Especially the code and data sections, that can be quite large, saving some memory before starting the wasm execution.
I don't think it's possible to obtain the same result with existing APIs.

eloparco:
Actually I was thinking, what if we only save the pointer here https://github.com/bytecodealliance/wasm-micro-runtime/blob/c0e33f08b04a5e03220ef74c2a9dbea8f1f35996/core/iwasm/common/wasm_c_api.c#L2286-L2292 instead of copying? That would allow more control and fix the problem I have.

wenyongh:
> Actually I was thinking, what if we only save the pointer here
> 
> https://github.com/bytecodealliance/wasm-micro-runtime/blob/c0e33f08b04a5e03220ef74c2a9dbea8f1f35996/core/iwasm/common/wasm_c_api.c#L2286-L2292
> 
> instead of copying? That would allow more control and fix the problem I have when using the wasm c api.
> We could specify that behavior in `LoadArgs` from `wasm_module_new_ex`.

@eloparco it seems to be a common issue so I think we can have a try. And we can also try to enable other methods in the future:
- Enable `LoadArgs->no_wasm_binary_copy` or `LoadArgs->clone_wasm_binary` for wasm_runtime_load_ex, like yamt mentioned. For example, create sections internally and free parts of sections at the end of loading. And if free callback is provided in `LoadArgs`, runtime can call it to free the input buffer after creating sections to reduce peak memory usage.

- Create sections from a stream when a read stream callback is provided, in this way, no need to read the whole wasm file to a buffer, so as to reduce the peak memory usage and total memory usage after loading.

- Use `wasm_runtime_read_to_sections` and `wasm_runtime_destroy_sections` that this PR implements. Could you check the comment for destroy_sections and help address it if it is good to you? My suggestion is to merge this PR after it is well reviewed, but had better get others' feedback.

yamt:
> How about letting the developer pass a read stream callback to read n bytes from the wasm/aot file, somewhat like
> 
> ```c
> typedef uint32 (*read_file_callback_t)(char *buf, uint32 count);  //read count bytes to buffer
> 
> wasm_module_t
> wasm_runtime_load_from_stream(read_file_callback_t, char *error_buf, uint32_t error_buf_size);
> ```

i feel that a callback based api in this case is a bit cumbersome to use for many use cases.
consider a loop with non-blocking recv and poll, which loads a module received via a socket.
in that case, a context-based apl similar to what we have in wamr-app-framework would be much simpler to use.

wenyongh:
> > How about letting the developer pass a read stream callback to read n bytes from the wasm/aot file, somewhat like
> > ```c
> > typedef uint32 (*read_file_callback_t)(char *buf, uint32 count);  //read count bytes to buffer
> > 
> > wasm_module_t
> > wasm_runtime_load_from_stream(read_file_callback_t, char *error_buf, uint32_t error_buf_size);
> > ```
> 
> i feel that a callback based api in this case is a bit cumbersome to use for many use cases. consider a loop with non-blocking recv and poll, which loads a module received via a socket. in that case, a context-based apl similar to what we have in wamr-app-framework would be much simpler to use.

Yes, that is better and it would be great if it supports receiving wasm file buffer from socket, uart or from a disk file. Not sure what the context-based api will be like?

wenyongh:
Since https://github.com/bytecodealliance/wasm-micro-runtime/pull/3389 was merged instead, close this PR.

PR #3378, opened 2024-04-30
aot compiler: Fix the length type passed to aot_memmove/aot_memset

wenyongh:
The current length type of aot_memmove/aot_memset is size_t, and on
a 64 bit host it is uint64, while what the aot code passes to it is uint32,
this might lead to unexpected behavior.

ps. https://github.com/bytecodealliance/wasm-micro-runtime/pull/3376.

### DIFF
--- a/core/iwasm/compilation/aot_emit_memory.c
+++ b/core/iwasm/compilation/aot_emit_memory.c
@@ -1094,7 +1094,7 @@ aot_compile_op_memory_copy(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx)
         /* zero extend to uint64 if the target is 64-bit */
         len = LLVMBuildZExt(comp_ctx->builder, len, I64_TYPE, "len64");
         if (!len) {
-            aot_set_last_error("llvm build bitcast failed.");
+            aot_set_last_error("llvm build zero extend failed.");
             return false;
         }
     }
@@ -1185,7 +1185,7 @@ aot_compile_op_memory_fill(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx)
         /* zero extend to uint64 if the target is 64-bit */
         len = LLVMBuildZExt(comp_ctx->builder, len, I64_TYPE, "len64");
         if (!len) {
-            aot_set_last_error("llvm build bitcast failed.");
+            aot_set_last_error("llvm build zero extend failed.");
             return false;
         }
     }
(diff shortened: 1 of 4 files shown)


core/iwasm/compilation/aot_emit_memory.c
```diff
@@ -1090,14 +1090,23 @@ aot_compile_op_memory_copy(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx)
...
+
     call_aot_memmove = comp_ctx->is_indirect_mode || comp_ctx->is_jit_mode;
     if (call_aot_memmove) {
         LLVMTypeRef param_types[3], ret_type, func_type, func_ptr_type;
         LLVMValueRef func, params[3];
 
         param_types[0] = INT8_PTR_TYPE;
         param_types[1] = INT8_PTR_TYPE;
-        param_types[2] = I32_TYPE;
+        param_types[2] = INTPTR_T_TYPE; /* int32 for 32-bit, int64 for 64-bit */
```
loganek, line 1109: Could we perhaps define SIZE_T_TYPE (or MEM_OFFSET_T_TYPE) instead? Even though it will (almost) always have the same value as INTPTR_T_TYPE, semantically pointer and size are different so the code would probably be slightly easier to read and maintain.

yamt:
> ps. #3376.

wrong link?

wenyongh:
> > ps. #3376.
> 
> wrong link?

Yeah, should be #3375, thanks, updated.

core/iwasm/compilation/aot_emit_memory.c
```diff
@@ -1090,14 +1090,23 @@ aot_compile_op_memory_copy(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx)
     if (!(dst_addr = check_bulk_memory_overflow(comp_ctx, func_ctx, dst, len)))
         return false;
 
+    if (comp_ctx->pointer_size == sizeof(uint64)) {
+        /* zero extend to uint64 if the target is 64-bit */
+        len = LLVMBuildZExt(comp_ctx->builder, len, I64_TYPE, "len64");
+        if (!len) {
+            aot_set_last_error("llvm build bitcast failed.");
```
yamt, line 1097: this is not a bitcast.

core/iwasm/compilation/aot_emit_memory.c
```diff
@@ -1172,9 +1181,18 @@ aot_compile_op_memory_fill(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx)
...
```
yamt, line 1188: ditto

core/iwasm/compilation/aot_emit_memory.c
wenyongh: Fixed, thanks!

yamt (approved):
lgtm

PR #3289, opened 2024-04-07
[WIP] Enhance wasm with checkpoint and restore support (#2333)

victoryang00:
- Add wasm_runtime_checkpoint/wasm_runtime_restore API
- Support AOT and Classic Interpreter mode checkpoint and debug through OS signal, tested on windows/mac/linux aarch64/x64
- Static instrument the AOT to have the checkpoint and restore switches
- Add sub extra library folder for implementing the ckpt-restore
- Include extra dependency of yalantinglib

Create a one-commit PR for adding the functionality for now, which requires more code reviews and testing in the future, like when ther I could better organize the third-party [yalantings library](https://github.com/alibaba/yalantinglibs.git) and test functionality for Classic Interpreter mode and AOT mode.

PR #3389, opened 2024-05-03
Allow not copying the wasm binary into the module when using WASM C API

eloparco:
The `wasm_module_new` function copies the WASM binary passed as an argument into the module.
This PR allows passing an additional flag to `wasm_module_new_ex`, to avoid that copy. In that way, the binary can be manually released after module loading (instead of having to wait for the store to be destroyed).


lum1n0us:
not very sure about it. the binary content(in wasm_module_t) will be used during execution. It's on purpose to release the binary when destroying store.

wenyongh:
> not very sure about it. the binary content(in wasm_module_t) will be used during execution. It's on purpose to release the binary when destroying store.

I think we can have a try, it really increases the memory usage if wasm-c-api clones another copy, and if it doesn't do that, the content of input buffer from developer may be modified and the buffer will be referred by runtime after loading, the difference is that developer can not use the input buffer to new a module again.

core/iwasm/include/wasm_c_api.h
```diff
@@ -527,6 +527,7 @@ typedef struct WASMModuleCommon *wasm_module_t;
 #define LOAD_ARGS_OPTION_DEFINED
 typedef struct LoadArgs {
     char *name;
+    bool no_wasm_binary_copy;
```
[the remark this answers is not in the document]
wenyongh: Or is it better to rename it to `clone_wasm_binary`? If yes, we had better set it true in wasm_module_new to keep compatibility:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/common/wasm_c_api.c#L2344

And we can also keep the compatibility for wasm_runtime_load_ex when we enable `LoadArgs->clone_wasm_binary = true` in the future, since the current behavior is like that `LoadArgs->clone_wasm_binary` is set as false now.

eloparco:
>not very sure about it. the binary content(in wasm_module_t) will be used during execution. It's on purpose to release the binary when destroying store.

Is that the case? When opening this PR, I had it tested only in interpreter mode and I thought it wouldn't work with AOT (that's why I created it as a draft and didn't bother fixing the CI errors).
But now I tried it in AOT mode and it seems to work as well. Is there anything I'm missing?

lum1n0us:
> be manually released after module instantiation 

if this means `wasm_byte_vec_delete(&binary)` after `wasm_instance_new()`, I suggest repeat the AOT example with `memset(binary.data, 0, ...)` instead of `wasm_byte_vec_delete`. 

IIUC, the big concern is function bytecodes. Classic-interp always needs the binary content. fast-interp doesn't because of recompilation. xxx_jit needs the content only during bytecodes to IRs translation. aot is there ⬆️. XIP always depends on the binary content. Then, next problem will be const strings. Luck for us, some function, like `wasm_const_str_list_insert()`, `aot_const_str_set_insert()`,  have a flag to control copying. All we need is to go through.

core/iwasm/include/wasm_c_api.h
```diff
@@ -527,16 +527,20 @@ typedef struct WASMModuleCommon *wasm_module_t;
...
+    module instead of being cloned. Hence, it should be freed exactly before
+    calling wasm_func_call and not before, as other Wasm C APIs use it. */
+    bool clone_wasm_binary;
     /* TODO: more fields? */
 } LoadArgs;
 #endif /* LOAD_ARGS_OPTION_DEFINED */
 
-WASM_API_EXTERN own wasm_module_t* wasm_module_new(
-  wasm_store_t*, const wasm_byte_vec_t* binary);
+WASM_API_EXTERN own wasm_module_t* wasm_module_new(wasm_store_t*, wasm_byte_vec_t* binary);
```
yamt, line 539: please avoid dropping const for wasm_module_new.

yamt:
> Luck for us, some function, like `wasm_const_str_list_insert()`, `aot_const_str_set_insert()`, have a flag to control copying. All we need is to go through.

iirc, it isn't by luck.
it's for app-manager, which performs stream-loading, which is basically same as what this PR seems trying.
i don't know how complete it is these days though. (eg. does it cover fast-jit?)

eloparco:
>which is basically same as what this PR seems trying.

This PR is not about stream-loading. It tries to save memory when using WAMR by allowing to release the wasm binary buffer before starting the wasm execution.

core/iwasm/include/wasm_c_api.h
eloparco: I can do that and then use a cast internally from `wasm_module_new`:
```c
module_ex->binary = (wasm_byte_vec_t *)binary;
```
In that way, the external API still has `const` that shows it doesn't modify the binary, that I think it's your concern.

core/iwasm/include/wasm_c_api.h
```diff
@@ -527,16 +527,20 @@ typedef struct WASMModuleCommon *wasm_module_t;
 #define LOAD_ARGS_OPTION_DEFINED
 typedef struct LoadArgs {
     char *name;
+    /* True by default.
```
wenyongh, line 530: Could you change to be somewhat like `Used by wasm-c-api only, true by default`? Since LoadArgs is also used by wasm_runtime_load_ex, but this option doesn't take effect for it.

core/iwasm/include/wasm_c_api.h
wenyongh: Agree to keep const flag for this function, and in wasm_module_new, convert it non-const:
```C
    return wasm_module_new_ex(store, (wasm_byte_vec_t *)binary, &args);
```

doc/memory_tune.md
```diff
@@ -30,3 +30,4 @@ Normally there are some methods to tune the memory usage:
 - set the app heap size with `wasm_runtime_instantiate`
 - use `nostdlib` mode, add `-Wl,--strip-all`: refer to [How to reduce the footprint](./build_wasm_app.md#2-how-to-reduce-the-footprint) of building wasm app for more details
 - use XIP mode, refer to [WAMR XIP (Execution In Place) feature introduction](./xip.md) for more details
+- when using the Wasm C API, set `clone_wasm_binary` in `LoadArgs` to `false` and free the wasm binary buffer (with `wasm_byte_vec_delete`) before wasm module execution
```
wenyongh, line 33: May be invalid? As the wasm/aot file may be referred by wasm module after loading, like re-using strings in wasm file:

https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_runtime.c#L4596-L4604

https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/aot/aot_loader.c#L324-L330

eloparco: It shouldn't be a problem if the module buffer is freed just before wasm execution (that is, `wasm_func_call`), not immediately after loading. Otherwise other APIs used after loading e.g. https://github.com/bytecodealliance/wasm-micro-runtime/blob/ca61184ced7d885395bc2425c87f0d063935dd1e/core/iwasm/include/wasm_c_api.h#L546
will try to use it and crash:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/ca61184ced7d885395bc2425c87f0d063935dd1e/core/iwasm/common/wasm_c_api.c#L2722

yamt:
> > which is basically same as what this PR seems trying.
> 
> This PR is not about stream-loading. It tries to save memory when using WAMR by allowing to release the wasm binary buffer before starting the wasm execution.

the "make a copy instead of keeping references" aspect is basically same.

core/iwasm/include/wasm_c_api.h
eloparco: The cast there wouldn't work if we want to keep const for `wasm_module_new_ex` too

yamt:
iirc, one of implications of is_load_from_file_buf=false is to make a copy. i guess you can somehow use it.

doc/memory_tune.md
wenyongh: Not very sure, it is a little confusing that execution won't use some fields of module? e.g. dumping call stacks when exception is thrown may get the function name from `func_inst->u.func_import->field_name`, `func_inst->u.func->field_name` or `export_func->name`:

https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_runtime.c#L3265-L3269

core/iwasm/include/wasm_c_api.h
wenyongh: Oh, my understanding is that we want to remove `const` for wasm_module_new_ex and keep it for `wasm_module_new`, since the buffer for the former may be modified, while the later won't. Not sure is it good to you, maybe we can add comments for them?

eloparco:
> iirc, one of implications of is_load_from_file_buf=false is to make a copy. i guess you can somehow use it.

I see that I copy is done anyway regardless of that flag https://github.com/bytecodealliance/wasm-micro-runtime/blob/ca61184ced7d885395bc2425c87f0d063935dd1e/core/iwasm/interpreter/wasm_runtime.c#L4625

That flag seems to be used to shift the string https://github.com/bytecodealliance/wasm-micro-runtime/blob/ca61184ced7d885395bc2425c87f0d063935dd1e/core/iwasm/interpreter/wasm_runtime.c#L4596

But again, if we avoid freeing the module buffer until the wasm execution that shouldn't be needed.

yamt:
> > iirc, one of implications of is_load_from_file_buf=false is to make a copy. i guess you can somehow use it.
> 
> I see that a copy is done anyway regardless of that flag
> 
> https://github.com/bytecodealliance/wasm-micro-runtime/blob/ca61184ced7d885395bc2425c87f0d063935dd1e/core/iwasm/interpreter/wasm_runtime.c#L4625
> 
> That flag seems to be used to shift the string
> 
> https://github.com/bytecodealliance/wasm-micro-runtime/blob/ca61184ced7d885395bc2425c87f0d063935dd1e/core/iwasm/interpreter/wasm_runtime.c#L4596

in the latter case, the user-given buffer is used.

> But again, if we avoid freeing the module buffer until the wasm execution that shouldn't be needed.

i don't understand what you mean. can you explain a bit?

doc/memory_tune.md
eloparco: True, I didn't consider that case. Would it be enough to pass `is_load_from_file_buf` as false to cover that case?

eloparco: >Would it be enough to pass is_load_from_file_buf as false to cover that case?

Tried it now (passing `false` as `is_load_from_file_buf` to `load_from_sections`) and it seems to work for the call stack when an exception is thrown. Is there any other case that I'm not considering?

wenyongh: Seems yes. I am also thinking that we should add a flag in LoadArgs to control the is_load_from_file_buf flag inside wasm_loader.c and aot_loader.c, e.g. add an extra argument for wasm_load, wasm_loader_load and aot_loader_from_aot_file, allowing them to pass `is_load_from_file_buf = false` to wasm_const_str_list_insert.

And then change the behavior of wasm_runtime_load_ex accordingly. And for wasm_runtime_load, set the right extra flag in LoadArgs before calling wasm_runtime_load_ex to allow modifying to the buffer, or, keep the same behavior for wasm_runtime_load before modification.

eloparco: >I am also thinking that we should add a flag in LoadArgs to control the is_load_from_file_buf flag

Can't we use `clone_wasm_binary` for that? Or do you think we will need the distinction for other use cases in the future?

eloparco: Also because `clone_wasm_binary=false` wouldn't work properly without `is_load_from_file_buf=false`

wenyongh: It is a different concept from clone_wasm_binary: wasm loader doesn't keep another binary now no matter clone_wasm_binary is true or false, it just modifies the binary and refers to it after loading. Using the flag for wasm loader makes confusing.

How about adding `bool wasm_binary_freeable` in LoadArgs, and mentioning that it is only for wasm/aot loader? And:
- Change wasm_runtime_load_ex, aot_load_from_aot_file, wasm_loader_load, wasm_load to use the flag
- Change wasm_runtime_load and wasm_module_new_ex, pass correct `wasm_binary_freeable` flag to wasm_runtime_load_ex

Or do you have better name for the flag?

eloparco: I think I applied your suggestions, let me know if I missed or misunderstood anything

core/iwasm/common/wasm_c_api.c
```diff
@@ -2343,7 +2353,9 @@ wasm_module_new(wasm_store_t *store, const wasm_byte_vec_t *binary)
 {
     LoadArgs args = { 0 };
     args.name = "";
-    return wasm_module_new_ex(store, binary, &args);
+    args.clone_wasm_binary = true;
+    args.wasm_binary_freeable = false;
```
wenyongh, line 2357: Should not set `args.wasm_binary_freeable` here, but set it in wasm_module_new_ex instead, before calling wasm_runtime_load_ex (L2302):
```C
    args.wasm_binary_freeable = !args.clone_wasm_binary;
    module_ex->module_comm_rt = wasm_runtime_load_ex(...);
```

core/iwasm/include/wasm_c_api.h
```diff
@@ -527,16 +527,23 @@ typedef struct WASMModuleCommon *wasm_module_t;
 #define LOAD_ARGS_OPTION_DEFINED
 typedef struct LoadArgs {
     char *name;
+    /* True by default, used by wasm-c-api only.
+    If false, the wasm input buffer (wasm_byte_vec_t) is referenced by the
+    module instead of being cloned. Hence, it can be freed after module loading. */
+    bool clone_wasm_binary;
+    /* False by default, used by wasm-c-api only.
```
wenyongh, line 534: Should be `used by AOT/wasm loader only`

core/iwasm/common/wasm_c_api.c
```diff
@@ -2283,13 +2285,19 @@ wasm_module_new_ex(wasm_store_t *store, const wasm_byte_vec_t *binary,
...
-        goto free_binary;
+        wasm_byte_vec_copy(module_ex->binary, binary);
+        if (!module_ex->binary->data)
+            goto free_binary;
+    }
+    else {
+        module_ex->binary = binary;
+    }
 
     module_ex->module_comm_rt = wasm_runtime_load_ex(
```
wenyongh, line 2302: `args = wasm_binary_freeable = !args->clone_wasm_binary;` before calling wasm_runtime_load_ex

core/iwasm/include/wasm_c_api.h
```diff
@@ -527,16 +527,23 @@ typedef struct WASMModuleCommon *wasm_module_t;
...
+    If true, the AOT/wasm loader creates a copy of some module fields (e.g. const
+    strings), making it possible to free the wasm binary buffer after loading. */
+    bool wasm_binary_freeable;
     /* TODO: more fields? */
 } LoadArgs;
 #endif /* LOAD_ARGS_OPTION_DEFINED */
 
-WASM_API_EXTERN own wasm_module_t* wasm_module_new(
-  wasm_store_t*, const wasm_byte_vec_t* binary);
+WASM_API_EXTERN own wasm_module_t* wasm_module_new(wasm_store_t*, const wasm_byte_vec_t* binary);
```
wenyongh, line 542: Had better not change the two lines if they are unchanged.

core/iwasm/include/wasm_export.h
```diff
@@ -222,6 +222,10 @@ typedef struct RuntimeInitArgs {
 #define LOAD_ARGS_OPTION_DEFINED
 typedef struct LoadArgs {
     char *name;
+    /* This option is only used by the Wasm C API (see wasm_c_api.h) */
+    bool clone_wasm_binary;
+    /* This option is only used by the Wasm C API (see wasm_c_api.h) */
```
wenyongh, line 227: Should be used by AOT/wasm loader only. Had better add detailed comment here, and in wasm-c-api.h: `This option is only used by the AOT/wasm loader (see wasm_export.h)`

doc/memory_tune.md
wenyongh: OK, I added some comments, could you help check and resolve them? I will check it again then.

core/iwasm/common/wasm_c_api.c
eloparco: Yeah, I was initially thinking of doing that, but I'll have to drop `const` from `LoadArgs`, I'll do it.

doc/memory_tune.md
eloparco: Thanks for the review, I pushed the suggested changes

doc/memory_tune.md
```diff
@@ -30,3 +30,4 @@ Normally there are some methods to tune the memory usage:
 - set the app heap size with `wasm_runtime_instantiate`
 - use `nostdlib` mode, add `-Wl,--strip-all`: refer to [How to reduce the footprint](./build_wasm_app.md#2-how-to-reduce-the-footprint) of building wasm app for more details
 - use XIP mode, refer to [WAMR XIP (Execution In Place) feature introduction](./xip.md) for more details
+- when using the Wasm C API, set `clone_wasm_binary=false` in `LoadArgs` and free the wasm binary buffer (with `wasm_byte_vec_delete`) after module loading
```
wenyongh, line 33: I discussed with @lum1n0us, per our understanding, currently it is only available for fast interpreter mode and AOT mode. For classic interpreter and all JIT modes, the bytecode is used by the interpreter or by the JIT compiler after loading, we may try to clone the bytecode to fix it, but we can do it in another PR.

So could you mention this here to avoid misunderstanding?

And could you also add another item for wasm/AOT loader, since now we can set `wasm_binary_freeable=true` in `LoadArgs` and free the wasm binary buffer after module loading. And similar, it's only available for fast interpreter mode and AOT mode now.

doc/memory_tune.md
wenyongh: Yes, it seems that all the places have been updated. It LGTM except a comment to update the document. Let's wait for comments from other developers before merging it.

wenyongh: BTW, had better try more cases. And it seems that we can close #3377 after this PR is merged?

eloparco: >And it seems that we can close https://github.com/bytecodealliance/wasm-micro-runtime/pull/3377 after this PR is merged?

Yes, this PR does what we were looking for in the initial issue https://github.com/bytecodealliance/wasm-micro-runtime/issues/3354

doc/memory_tune.md
wenyongh: The bytecodes in code section is used in classic interpreter after loading:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_loader.c#L3700-L3710

https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_loader.c#L5722-L5725

wenyongh (commented):
LGTM

lum1n0us:
> What is supposed to break with classic interpreter? Because I gave it a quick try and seemed to run fine.

Please correct me if I am wrong. Your examples will free/release/pollute binary content after module loading (before instantiation). And it works well in classic-interp, fast-interp, jit and aot.

yamt:
my impression is that this requires users too much implementation knowledge.

how about providing an api to query if it's safe to free the buffer?

eg.
```c
bool has_reference_to_underlying_byte_vec(const wasm_module_t *module);
```

wenyongh:
> my impression is that this requires users too much implementation knowledge.
> 
> how about providing an api to query if it's safe to free the buffer?
> 
> eg.
> 
> ```c
> bool has_reference_to_underlying_byte_vec(const wasm_module_t *module);
> ```

It is good, but the API name is a not so consistent with others, how about `bool wasm_module_refer_to_underlying_binary` or others?

doc/memory_tune.md
eloparco: https://github.com/bytecodealliance/wasm-micro-runtime/pull/3389#issuecomment-2102207812 So, yes, I actually free the wasm binary buffer after module instantiation, not loading. In that way, it works with classic interpreter too.

In my case it doesn't make much of a difference if the binary is freed after loading or after instantiation. The main goal is to free it before wasm execution, because the execution will start using additional memory and we don't want to overhead of the binary buffer.

eloparco:
>how about providing an api to query if it's safe to free the buffer?

What would the API do? Just return `is_binary_cloned` from `wasm_module_ex_t` in case of wasm-c-api? That wouldn't simplify much for the user.

What about adding a new example (as I was doing in https://github.com/bytecodealliance/wasm-micro-runtime/pull/3377)? That would make things easier for the user, as they can just follow what is done there.

wenyongh:
> > how about providing an api to query if it's safe to free the buffer?
> 
> What would the API do? Just return `is_binary_cloned` from `wasm_module_ex_t` in case of wasm-c-api? Or you mean we would need to add a new boolean to `wasm_module_ex_t` and set it to true/false after module loading (or instantiation) is completed, and that boolean is what's returned by the new API. That wouldn't simplify much for the user.
> 
> What about adding a new example (as I was doing in #3377)? That would make things easier for the user, as they can just follow what is done there.

@eloparco just my suggestion: maybe we can add two APIs, one is for WAMR api and is added in wasm_export.h, the other is for wasm-c-api and is added inc wasm_c_api.h, the former:
```C
bool wasm_runtime_is_underlying_binary_freeable(const wasm_module_t module);
```
And we can just add a field like `is_binary_freeable` in WASMModule/AOTModule, set it in when wasm_runtime_load/wasm_runtime_load_ex, and return it in the new API.

the latter:
```C
bool wasm_module_is_underlying_binary_cloned(wasm_module_t* module);
```
In the implementation, it is somewhat like: `return ((wasm_module_ex_t *)module)->is_binary_cloned`
```

**Comment by eloparco** [comment]
>@eloparco just my suggestion: maybe we can add two APIs, one is for WAMR api and is added in wasm_export.h, the other is for wasm-c-api and is added inc wasm_c_api.h

Added those, not sure how useful they are. Why would the developer use `wasm_module_is_underlying_binary_cloned`? They already know if they passed `clone_wasm_binary=true` as `LoadArgs`.

Same for `wasm_runtime_is_underlying_binary_freeable`, it won't lift any implementation knowledge from the developer in my opinion. To do something like that, an API like
```
bool
wasm_module_free_wasm_binary_buffer(wasm_byte_vec_t *binary,
                                    wasm_module_t *module)
{
    if (!((wasm_module_ex_t *)module)->is_binary_cloned
        && !wasm_runtime_is_underlying_binary_freeable(module))
        return false;

    wasm_byte_vec_delete(binary);
    return true;
}
```
would be more useful. And something equivalent in the WAMR API.

But I'm fine with the suggestion since it won't make a huge difference for me.

File: doc/memory_tune.md [hunk 11]
```diff
@@ -30,3 +30,5 @@ Normally there are some methods to tune the memory usage:
 - set the app heap size with `wasm_runtime_instantiate`
 - use `nostdlib` mode, add `-Wl,--strip-all`: refer to [How to reduce the footprint](./build_wasm_app.md#2-how-to-reduce-the-footprint) of building wasm app for more details
 - use XIP mode, refer to [WAMR XIP (Execution In Place) feature introduction](./xip.md) for more details
+- when using the Wasm C API in fast interpreter or AOT mode, set `clone_wasm_binary=false` in `LoadArgs` and free the wasm binary buffer (with `wasm_byte_vec_delete`) after module loading; `wasm_runtime_is_underlying_binary_freeable` can be queried to check if the wasm binary buffer can be safely freed
+- when using the wasm/AOT loader in fast interpreter or AOT mode, set `wasm_binary_freeable=true` in `LoadArgs` and free the wasm binary buffer (with `wasm_byte_vec_delete`) after module loading; `wasm_runtime_is_underlying_binary_freeable` can be queried to check if the wasm binary buffer can be safely freed
```
yamt, line 34: can you avoid the api which requires a user to determine the module type before loading it?

core/iwasm/include/wasm_c_api.h
```diff
@@ -527,6 +527,12 @@ typedef struct WASMModuleCommon *wasm_module_t;
 #define LOAD_ARGS_OPTION_DEFINED
 typedef struct LoadArgs {
     char *name;
+    /* True by default, used by wasm-c-api only.
+    If false, the wasm input buffer (wasm_byte_vec_t) is referenced by the
+    module instead of being cloned. Hence, it can be freed after module loading. */
+    bool clone_wasm_binary;
+    /* This option is only used by the AOT/wasm loader (see wasm_export.h) */
+    bool wasm_binary_freeable;
```
yamt, line 535: i don't understand why this needs this complex.

i guess you just need two modes at most.
a. standard-compatible behavior
b. incompatible, but might save memory

doc/memory_tune.md
```diff
@@ -30,3 +30,5 @@ Normally there are some methods to tune the memory usage:
 - set the app heap size with `wasm_runtime_instantiate`
 - use `nostdlib` mode, add `-Wl,--strip-all`: refer to [How to reduce the footprint](./build_wasm_app.md#2-how-to-reduce-the-footprint) of building wasm app for more details
 - use XIP mode, refer to [WAMR XIP (Execution In Place) feature introduction](./xip.md) for more details
+- when using the Wasm C API in fast interpreter or AOT mode, set `clone_wasm_binary=false` in `LoadArgs` and free the wasm binary buffer (with `wasm_byte_vec_delete`) after module loading; `wasm_runtime_is_underlying_binary_freeable` can be queried to check if the wasm binary buffer can be safely freed
```
yamt, line 33: when you say "module loading", what exacty is it?
wasm_module_new_ex?

yamt (commented):
please add an sample, which works for all module types and modes.

doc/memory_tune.md
eloparco: You mean returning an error during loading if for instance the user set `clone_wasm_binary=false` in classic interpreter mode? As in https://github.com/bytecodealliance/wasm-micro-runtime/pull/3389#issuecomment-2102207812, if we wait until after module instantiation (instead of module loading) we can still use classic interpreter and free the buffer.

core/iwasm/include/wasm_c_api.h
eloparco: We had a discussion here: https://github.com/bytecodealliance/wasm-micro-runtime/pull/3389#discussion_r1592616397

eloparco:
>please add an sample, which works for all module types and modes.

Yes, I think that would be useful (I was suggesting it already in https://github.com/bytecodealliance/wasm-micro-runtime/pull/3389#issuecomment-2102225098)

doc/memory_tune.md
eloparco: `wasm_module_new(_ex)` and `wasm_runtime_load(_ex)`

core/iwasm/include/wasm_export.h
```diff
@@ -222,6 +222,13 @@ typedef struct RuntimeInitArgs {
 #define LOAD_ARGS_OPTION_DEFINED
 typedef struct LoadArgs {
     char *name;
+    /* This option is only used by the Wasm C API (see wasm_c_api.h) */
+    bool clone_wasm_binary;
+    /* False by default, used by AOT/wasm loader only.
```
yamt, line 227: what's AOT/wasm loader here?
aot_loader.c, wasm_loader.c, and wasm_mini_loader.c?
do you mean all module loaders?

doc/memory_tune.md
yamt: in the description https://github.com/bytecodealliance/wasm-micro-runtime/pull/3389#issue-2277885176 you said "after module instantiation", which makes me think `wasm_instance_new`. can you update the description?

core/iwasm/include/wasm_export.h
eloparco: Yes, all module loaders basically

yamt:
> > @eloparco just my suggestion: maybe we can add two APIs, one is for WAMR api and is added in wasm_export.h, the other is for wasm-c-api and is added inc wasm_c_api.h
> 
> Added those (separate commit), not sure how useful they are. Why would the developer use `wasm_module_is_underlying_binary_cloned`? They already know if they passed `clone_wasm_binary=true` as `LoadArgs`.
> 
> Same for `wasm_runtime_is_underlying_binary_freeable`, it won't lift any implementation knowledge from the developer in my opinion. To do something like that, an API like
> 
> ```
> bool
> wasm_module_free_wasm_binary_buffer(wasm_byte_vec_t *binary,
>                                     wasm_module_t *module)
> {
>     if (!((wasm_module_ex_t *)module)->is_binary_cloned
>         && !wasm_runtime_is_underlying_binary_freeable(module))
>         return false;
> 
>     wasm_byte_vec_delete(binary);
>     return true;
> }
> ```
> 
> would be more useful. And something equivalent in the WAMR API.
> 
> But I'm fine with the suggestion since it won't make a huge difference for me.

when i suggested a query api, i meant to make it check all necessary conditions
including is_binary_cloned, module type, etc.
so it would be just:
```c
if (wasm_runtime_is_underlying_binary_freeable(...)) {
    wasm_byte_vec_delete(binary);
    have_binary = false;
}
```

wenyongh:
> > > @eloparco just my suggestion: maybe we can add two APIs, one is for WAMR api and is added in wasm_export.h, the other is for wasm-c-api and is added inc wasm_c_api.h
> > 
> > 
> > Added those (separate commit), not sure how useful they are. Why would the developer use `wasm_module_is_underlying_binary_cloned`? They already know if they passed `clone_wasm_binary=true` as `LoadArgs`.
> > Same for `wasm_runtime_is_underlying_binary_freeable`, it won't lift any implementation knowledge from the developer in my opinion. To do something like that, an API like
> > ```
> > bool
> > wasm_module_free_wasm_binary_buffer(wasm_byte_vec_t *binary,
> >                                     wasm_module_t *module)
> > {
> >     if (!((wasm_module_ex_t *)module)->is_binary_cloned
> >         && !wasm_runtime_is_underlying_binary_freeable(module))
> >         return false;
> > 
> >     wasm_byte_vec_delete(binary);
> >     return true;
> > }
> > ```
> >     
> >   
> > would be more useful. And something equivalent in the WAMR API.
> > But I'm fine with the suggestion since it won't make a huge difference for me.
> 
> when i suggested a query api, i meant to make it check all necessary conditions including is_binary_cloned, module type, etc. so it would be just:
> 
> ```c
> if (wasm_runtime_is_underlying_binary_freeable(...)) {
>     wasm_byte_vec_delete(binary);
>     have_binary = false;
> }
> ```

The wasm-c-api wasm_module_ex_t's `is_binary_cloned` and wasm/aot wasm_module_t's `is_binary_freeable` are two options and the behaviors are different, and they are stored in different structures, I think we cannot just provide one `xxx_is_underly_binary_freeable` like API for both of them, and also the input binary for wasm_module_new/wasm_module_new_ex/wasm_runtime_load/wasm_runtime_load_ex is provided by developer, it should not be freed by runtime or we had better not use a runtime API to free it. And the cloned binary of wasm-c-api (if the input binary is cloned) will be freed by runtime, developer should not free it also.

So, per my understanding, developer can use the `xxx_is_underly_binary_freeable` like API to check whether he can free the input binary and free it if yes, for wasm-c-api, his code is somewhat like:
```C
    if (wasm_module_is_underlying_binary_cloned(module)) {  //or if (wasm_module_is_underlying_binary_freeable(module))
        wasm_byte_vec_delete(binary);
        wasm_runtime_free(binary);
    }
```

for wasm/aot loader, his code is somewhat like:
```
    if (wasm_runtime_is_underlying_binary_freeable(module)) {
        wasm_runtime_free(buf);
    }
```

@eloparco @yamt not sure whether that is ok for you? Or how about providing `wasm_module_is_underlying_binary_freeable` instead of `wasm_module_is_underlying_binary_cloned` for wasm-c-api, and `wasm_runtime_is_underlying_binary_freeable` for wasm/aot loader? Or is there other suggestion from you? Thanks.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1350,12 +1350,14 @@ wasm_runtime_load_ex(uint8 *buf, uint32 size, const LoadArgs *args,
...
 #endif
                                           args, error_buf, error_buf_size);
+        ((WASMModule *)module_common)->is_binary_freeable = true;
 #endif
     }
     else if (get_package_type(buf, size) == Wasm_Module_AoT) {
 #if WASM_ENABLE_AOT != 0
         module_common = (WASMModuleCommon *)aot_load_from_aot_file(
             buf, size, args, error_buf, error_buf_size);
+        ((AOTModule *)module_common)->is_binary_freeable = true;
```
wenyongh, line 1360: Should be `((AOTModule *)module_common)->is_binary_freeable = args.wasm_binary_freeable;`

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1350,12 +1350,14 @@ wasm_runtime_load_ex(uint8 *buf, uint32 size, const LoadArgs *args,
                                           true,
 #endif
                                           args, error_buf, error_buf_size);
+        ((WASMModule *)module_common)->is_binary_freeable = true;
```
wenyongh, line 1353: Should be `((WASMModule *)module_common)->is_binary_freeable = args.wasm_binary_freeable;`

wenyongh: And should also modify `wasm_runtime_load_from_sections` in this file, set `((WASMModule *)module_common)->is_binary_freeable =true` and `((AOTModule *)module_common)->is_binary_freeable = true`

eloparco:
> Or how about providing wasm_module_is_underlying_binary_freeable instead of wasm_module_is_underlying_binary_cloned for wasm-c-api, and wasm_runtime_is_underlying_binary_freeable for wasm/aot loader?

Do we need two separate APIs? One (i.e. `wasm_runtime_is_underlying_binary_freeable`, that checks `wasm_binary_freeable`) should be enough since we set `args->wasm_binary_freeable = !args->clone_wasm_binary;`

And `wasm_runtime_is_underlying_binary_freeable` would look like what I already added to this PR.

>including is_binary_cloned, module type, etc.

It's not module type but rather execution mode. I wouldn't add a check on that but rather write in the documentation that, with fast interpreter and AOT, the module can be freed after loading; while, in classic interpreter mode, after instantiation.

yamt:
> I wouldn't add a check on that

why not?

yamt:
> while, in classic interpreter mode, after instantiation.

i believe you can't free the buffer for classic interpreter.

samples/basic/src/free_buffer_early.c
```diff
@@ -0,0 +1,101 @@
...
+    if (!module_inst) {
+        printf("Instantiate wasm module failed. error: %s.\n", error_buf);
+        goto fail;
+    }
+
+    if (!wasm_runtime_is_underlying_binary_freeable(module)) {
+        printf("Unable to free wasm binary buffer.\n");
+        goto fail;
+    }
+    wasm_runtime_free(buffer);
```
wenyongh, line 85: Should set buffer to NULL after free, so we don't need to free it in `fail` label

samples/basic/src/free_buffer_early.c
```diff
@@ -0,0 +1,101 @@
...
+    if (!success) {
+        printf("Unable to execute function.\n");
+        goto fail;
+    }
+
+fail:
+    if (module_inst)
+        wasm_runtime_deinstantiate(module_inst);
+    if (module)
+        wasm_runtime_unload(module);
```
wenyongh, line 98: Should also free buffer if it isn't NULL:
```C
    if (buffer)
        wasm_runtime_free(buffer);
```

samples/basic/src/free_buffer_early.c
```diff
@@ -0,0 +1,101 @@
...
+{
+    fprintf(stdout, "Options:\r\n");
+    fprintf(stdout, "  -f [path of wasm file] \n");
+}
+
+int
+main(int argc, char *argv_main[])
+{
+    static char global_heap_buf[512 * 1024];
+    char *buffer, error_buf[128];
```
wenyongh, line 20: Had better set buffer to NULL

wenyongh:
> > Or how about providing wasm_module_is_underlying_binary_freeable instead of wasm_module_is_underlying_binary_cloned for wasm-c-api, and wasm_runtime_is_underlying_binary_freeable for wasm/aot loader?
> 
> Do we need two separate APIs? One (i.e. `wasm_runtime_is_underlying_binary_freeable`, that checks `wasm_binary_freeable`) should be enough since we set `args->wasm_binary_freeable = !args->clone_wasm_binary;`
> 
> And `wasm_runtime_is_underlying_binary_freeable` would look like what I already added to this PR.

Yeah, I mean one API for wasm-c-api and one for wamr runtime api, and I found you had done that in the PR, it LGTM.

wenyongh:
> > while, in classic interpreter mode, after instantiation.
> 
> i believe you can't free the buffer for classic interpreter.

Agree, how about we merge this PR first and then submit another PR to clone the wasm file's bytecode for classic interpreter mode to avoid referring to the underlying buffer?

samples/basic/src/free_buffer_early.c
```diff
@@ -0,0 +1,101 @@
...
+    module_inst = wasm_runtime_instantiate(module, stack_size, heap_size,
+                                           error_buf, sizeof(error_buf));
+    if (!module_inst) {
+        printf("Instantiate wasm module failed. error: %s.\n", error_buf);
+        goto fail;
+    }
+
+    if (!wasm_runtime_is_underlying_binary_freeable(module)) {
+        printf("Unable to free wasm binary buffer.\n");
+        goto fail;
```
yamt, line 83: my understanding is that early-freeing is merely optimization.
there is no need to bail out.

core/iwasm/include/wasm_c_api.h
yamt: > We had a discussion here: [#3389 (comment)](https://github.com/bytecodealliance/wasm-micro-runtime/pull/3389#discussion_r1592616397)

where? this link doesn't seem working for me.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -7170,3 +7176,22 @@ wasm_runtime_detect_native_stack_overflow_size(WASMExecEnv *exec_env,
     }
     return true;
 }
+
+WASM_RUNTIME_API_EXTERN bool
+wasm_runtime_is_underlying_binary_freeable(const wasm_module_t module)
+{
+    bool is_freeable = false;
+
+#if WASM_ENABLE_INTERP != 0
```
yamt, line 7185: i'd suggest to make this return false for classic interpreter and lazy jit.

core/iwasm/common/wasm_c_api.c
```diff
@@ -2986,6 +2998,12 @@ wasm_module_get_name(wasm_module_t *module)
     return wasm_runtime_get_module_name(module_ex->module_comm_rt);
 }
 
+bool
+wasm_module_is_underlying_binary_freeable(const wasm_module_t *module)
+{
+    return ((wasm_module_ex_t *)module)->is_binary_cloned;
```
yamt, line 3004: i'd suggest to make this check wasm_runtime_is_underlying_binary_freeable on the module as well.

yamt:
> > > while, in classic interpreter mode, after instantiation.
> > 
> > 
> > i believe you can't free the buffer for classic interpreter.
> 
> Agree, how about we merge this PR first and then submit another PR to clone the wasm file's bytecode for classic interpreter mode to avoid referring to the underlying buffer?

i'm not sure if it's a good idea.
i suspect usually code and data segments occupy the major part of wasm module.
if so, for classic interpreter, it might be simpler to do the opposite to what this PR does.
that is, keep references to the underlying buffer for data segments as well.

i feel it's simpler to make wasm_runtime_is_underlying_binary_freeable return false for now, until someone evaluates pros/cons.

core/iwasm/common/wasm_runtime_common.c
wenyongh: classic interpreter will be enabled when JIT is enabled, maybe the code can be like:
```C
#if WASM_ENABLE_INTERP != 0
    if (module->module_type == Wasm_Module_Bytecode) {
#if (WASM_ENABLE_JIT != 0 || WASM_ENABLE_FAST_JIT != 0) && (WASM_ENABLE_LAZY_JIT !=0)
        is_freeable = false;
#elseif WASM_ENABLE_FAST_INTERP == 0
        is_freeable = false;
#else
        is_freeable = ((WASMModule *)module)->is_binary_freeable;
#endif
    }
#endif
```

samples/basic/src/free_buffer_early.c
wenyongh: yes, seems the code should be like below?
```C
    if (wasm_runtime_is_underlying_binary_freeable(module)) {
        wasm_runtime_free(buffer);
        buffer = NULL;
    }
```

samples/basic/src/free_buffer_early.c
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    module_inst = wasm_runtime_instantiate(module, stack_size, heap_size,
+                                           error_buf, sizeof(error_buf));
+    if (!module_inst) {
+        printf("Instantiate wasm module failed. error: %s.\n", error_buf);
+        goto fail;
+    }
+
+    if (!wasm_runtime_is_underlying_binary_freeable(module)) {
```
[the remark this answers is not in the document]
eloparco: The example uses classic interpreter by default https://github.com/bytecodealliance/wasm-micro-runtime/blob/c6d42db598606a672a9f319d442198fb6584421c/samples/basic/CMakeLists.txt#L49 so it doesn't work if I move the free. If we want to do that we need to change it to fast interpreter.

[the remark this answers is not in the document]
wenyongh: yes, had better use fast interpreter instead. Not sure why we can free the buffer after instantiation for classic interpreter, need time to check.

samples/basic/src/free_buffer_early.c
eloparco: >my understanding is that early-freeing is merely optimization.
there is no need to bail out.

Yes, that's true in general, but the example is specifically to demonstrate how to free memory early, so I was considering a failure in doing that as a failure for the example. But better to do as you say since this is not a test but an example.

samples/basic/src/free_buffer_early.c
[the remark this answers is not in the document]
eloparco: Never mind, we can keep it to classic interpreter and make sure that `wasm_runtime_is_underlying_binary_freeable` returns false.

core/iwasm/common/wasm_c_api.c
```diff
@@ -2283,14 +2284,21 @@ wasm_module_new_ex(wasm_store_t *store, const wasm_byte_vec_t *binary,
...
-        goto free_binary;
+        wasm_byte_vec_copy(module_ex->binary, binary);
+        if (!module_ex->binary->data)
+            goto free_binary;
+    }
+    else {
+        module_ex->binary = binary;
+    }
 
+    args->wasm_binary_freeable = !args->clone_wasm_binary;
```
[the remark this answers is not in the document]
eloparco: Yes, wasm-c-api only uses `clone_wasm_binary`, while wamr api uses `wasm_binary_freeable`

eloparco:
@yamt @wenyongh I applied the changes after the discussion, let me know if I missed anything

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1400,6 +1406,7 @@ wasm_runtime_load_from_sections(WASMSection *section_list, bool is_aot,
             LOG_DEBUG("WASM module load failed from sections");
```
wenyongh, line 1406: Seems we had better also set `args.is_binary_freeable = false` in wasm_runtime_load, see L1385, though `LoadArgs args = { 0 };` already set it to false?

wenyongh (commented):
LGTM except two minor comments.

eloparco:
> Hi, I checked the source code again, there are really several places in wasm/aot module which refer to the input buffer

Thanks for spotting that

>I am not sure how to handle them, for 1, there may be two options:
(1) allocate new memory for each data segment, (2) wasm_runtime_is_underlying_binary_freeable checks whether each data segment is dropped with bh_bitmap_get_bit and returns false if one of the data segments is not dropped.

I'll try to update the code to use (1) and see how it goes. How does (2) fix the problem with memory.init? Memory.init would be called during wasm execution and we won't be able to free the wasm binary buffer before that.

>for 2, maybe we can check whether module->string_literal_ptrs is NULL

And what do we do if it's NULL?

>for 3, maybe we can ignore it, and mention in the document that after the underlying binary buffer is freed, developer cannot call wasm_runtime_get_custom_section again.

Yes, makes sense to me

wenyongh:
> > Hi, I checked the source code again, there are really several places in wasm/aot module which refer to the input buffer
> 
> Thanks for spotting that
> 
> > I am not sure how to handle them, for 1, there may be two options:
> > (1) allocate new memory for each data segment, (2) wasm_runtime_is_underlying_binary_freeable checks whether each data segment is dropped with bh_bitmap_get_bit and returns false if one of the data segments is not dropped.
> 
> I'll try to update the code to use (1) and see how it goes. How does (2) fix the problem with memory.init? Memory.init would be called during wasm execution and we won't be able to free the wasm binary buffer before that.

No better way to fix (2), what we can do is to assume that `memory.init` is executed during instantiation and all the passive data segments are consumed and set to dropped state, so `wasm_runtime_is_underlying_binary_freeable` may return true if all of them are dropped.

Agree to use (1) to fully resolve the issue, and we may also free the dropped data segments after they are dropped - not in this PR, we can submit another PR to refine it.

> 
> > for 2, maybe we can check whether module->string_literal_ptrs is NULL
> 
> And what do we do if it's NULL?

`wasm_runtime_is_underlying_binary_freeable` return false when `module->string_literal_ptrs` is not NULL. It only returns true when both 1 is good and `module->string_literal_ptrs` is NULL.

> 
> > for 3, maybe we can ignore it, and mention in the document that after the underlying binary buffer is freed, developer cannot call wasm_runtime_get_custom_section again.
> 
> Yes, makes sense to me

eloparco:
>No better way to fix (2), what we can do is to assume that memory.init is executed during instantiation and all the passive data segments are consumed and set to dropped state, so wasm_runtime_is_underlying_binary_freeable may return true if all of them are dropped.

Right, done that now, but I needed to change the API to accept the instance instead of the module as an argument.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -7170,3 +7178,49 @@ wasm_runtime_detect_native_stack_overflow_size(WASMExecEnv *exec_env,
...
+        is_freeable = module->is_binary_freeable;
+#if WASM_ENABLE_GC != 0 && WASM_ENABLE_STRINGREF != 0
+        is_freeable = is_freeable && module->string_literal_ptrs == NULL;
+#endif
+    }
+#endif /* WASM_ENABLE_AOT != 0 */
+
+    (void)i;
+    return is_freeable;
+}
```
wenyongh, line 7226: The code is a little complex, how about:
```C
WASM_RUNTIME_API_EXTERN bool
wasm_runtime_is_underlying_binary_freeable(const wasm_module_inst_t module_inst)
{
    uint32 i;

#if WASM_ENABLE_INTERP != 0
    if (module_inst->module_type == Wasm_Module_Bytecode) {
#if (WASM_ENABLE_JIT != 0 || WASM_ENABLE_FAST_JIT != 0) \
    && (WASM_ENABLE_LAZY_JIT != 0)
        return false;
#elif WASM_ENABLE_FAST_INTERP == 0
        return false;
#else
        /* Fast interpreter mode */
        WASMModule *module = ((WASMModuleInstance *)module_inst)->module;
        if (!module->is_binary_freeable)
            return false;
#if WASM_ENABLE_GC != 0 && WASM_ENABLE_STRINGREF != 0
        if (module->string_literal_ptrs)
            return false;
#endif
        for (i = 0; i < module->data_seg_count; i++)
            if (!bh_bitmap_get_bit(
                    ((WASMModuleInstance *)module_inst)->e->common.data_dropped,
                    i))
                return fasle;
#endif
#endif
    }
#endif /* WASM_ENABLE_INTERP != 0 */
#if WASM_ENABLE_AOT != 0
        AOTModule *module =
            (AOTModule *)((AOTModuleInstance *)module_inst)->module;
        if (!module->is_binary_freeable)
            return false;
#if WASM_ENABLE_GC != 0 && WASM_ENABLE_STRINGREF != 0
        if (module->string_literal_ptrs)
            return false;
#endif
    }
#endif /* WASM_ENABLE_AOT != 0 */

    (void)i;
    return true;
}
```

wenyongh: And could you also set `args.is_binary_freeable = false` in wasm_runtime_load, L1385?

core/iwasm/include/wasm_export.h
```diff
@@ -1860,6 +1867,16 @@ WASM_RUNTIME_API_EXTERN bool
...
 
+/**
+ * Query whether the wasm binary buffer used to create the module can be freed
+ *
+ * @param module_inst the target module instance
+ * @return true if the wasm binary buffer can be freed
+ */
+WASM_RUNTIME_API_EXTERN bool
+wasm_runtime_is_underlying_binary_freeable(
+    const wasm_module_inst_t module_inst);
```
wenyongh, line 1878: This makes that the binary can be freed only after instantiation, is it better to clone data segment in wasm module and clone string_literal_ptrs in both wasm and aot modules when wasm_binary_freeable is true? So that we can free the input buffer after loading. If it is a little complex, we can help do it after this PR is merged.

core/iwasm/common/wasm_runtime_common.c
eloparco: Right, that looks cleaner! Pushed the change

core/iwasm/include/wasm_export.h
eloparco: Yeah, we would need to clone the `data_dropped` bitmap into the module to avoid using the instance. But, as you say, we can do that in a separate PR

wenyongh: Should be to clone the data segments in wasm file but not to clone the data_dropped bitmap. Anyway, let's do it in another PR.

wenyongh (approved):
LGTM

yamt:
i still think the load_arg->wasm_binary_freeable api is not a good idea.
with this api, if a user wants to save memory, he first needs to determine aot/interpreter type/etc by examining wamr build options and the module to be loaded to decide wasm_binary_freeable=true/false.
IMO, it should be something like "a user just expresses the intent to save memory. the runtime does it automatically when possible" instead.
how do you think?

btw, can you stop marking comments "resolved" so eagerly?

wenyongh:
@yamt, do you mean that developer doesn't need to use wasm_runtime_is_underlying_binary_freeable/wasm_runtime_is_underlying_binary_freeable API to check again? For example, if he wants to save memory, he can just set option in LoadArgs to express the intent, and runtime does all the things, and then he can just free the input binary after loading. I strongly agree that since it simplifies the operations and makes less confusing, and we may clone data segments for wasm module, and even more, clone bytecode if needed, e.g. for classic interpreter and lazy jit. How do you think?

eloparco:
>btw, can you stop marking comments "resolved" so eagerly?

I'm not deleting them, you can still check them if needed. If I already addressed the comment, I prefer to resolved it so that I can better track what still needs to be discussed or addressed.

> @yamt, do you mean that developer doesn't need to use

I guess it's more about avoiding the extra arguments in `load_arg` and try to save memory automatically if possible. And then the user calls `wasm_runtime_is_underlying_binary_freeable` to see it the runtime was able to.

yamt:
with the current api, a user needs something complex to decide wasm_binary_freeable.
eg.
```c
if (get_package_type(buffer) == aot) {
  loadarg.wasm_binary_freeable = true;
} else {
#if WASM_ENABLE_FAST_INTERP != 0
  loadarg.wasm_binary_freeable = true;
#else
  loadarg.wasm_binary_freeable = false;
#endif
}
```

i meant it can be
eg.
```c
loadarg.want_early_free = true; /* just a hint to the runtime */
```

eloparco:
>loadarg.want_early_free = true; /* just a hint to the runtime */

Would it make sense to assume it to always be true? So that we don't need `loadarg` at all and the runtime always try to avoid the wasm binary buffer copy.

wenyongh:
> > loadarg.want_early_free = true; /* just a hint to the runtime */
> 
> Would it make sense to assume it to always be true? So that we don't need `loadarg` at all and the runtime always try to avoid the wasm binary buffer copy.

I think sometimes developer cannot free the input buffer, seems we had better add an option.

yamt:
> > loadarg.want_early_free = true; /* just a hint to the runtime */
> 
> Would it make sense to assume it to always be true?

it would have a negative effect to increase memory usage unless a user actually tries to free the buffer early with wasm_runtime_is_underlying_binary_freeable.

yamt:
> > with the current api, a user needs something complex to decide wasm_binary_freeable. eg.
> > ```c
> > if (get_package_type(buffer) == aot) {
> >   loadarg.wasm_binary_freeable = true;
> > } else {
> > #if WASM_ENABLE_FAST_INTERP != 0
> >   loadarg.wasm_binary_freeable = true;
> > #else
> >   loadarg.wasm_binary_freeable = false;
> > #endif
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > i meant it can be eg.
> > ```c
> > loadarg.want_early_free = true; /* just a hint to the runtime */
> > ```
> 
> Is that to allow the developer to free the wasm binary after wasm/aot loading?

it tells the runtime to make later wasm_runtime_is_underlying_binary_freeable.

> And do you mean to use the same option for both wasm-c-api and wamr api?

maybe. i haven't thought much about wasm-c-api yet.

> And what should the behavior be for wasm/aot loader and wamr-c-api loader?

i'm not sure if i understand this sentence.
are you talking about the default behavior?

wenyongh:
> > 
> > 
> > Is that to allow the developer to free the wasm binary after wasm/aot loading?
> 
> it tells the runtime to perform something which possibly allows early free (typically making extra copies) if it makes sense. otherwise, the runtime can ignore it and make wasm_runtime_is_underlying_binary_freeable return false.
> 

Yes, how about `loadarg.allow_early_free` instead of `loadarg.want_early_free`? And per my underderstanding, it allows to free the binary after loading, but not after instantiation, and we will change argument `wasm_module_inst_t module_inst` of wasm_runtime_is_underlying_binary_freeable to `wasm_module_t module`, just like we discussed before in this PR.

> > And do you mean to use the same option for both wasm-c-api and wamr api?
> 
> maybe. i haven't thought much about wasm-c-api yet.
> 

By default wasm-c-api will clone the binary and already allows to early free, so using the option doesn't make sense for it. Maybe we can also keep `loadarg.clone_wasm_binary` for wasm-c-api, and:
- If `loadarg.clone_wasm_binary` is true, `loadarg.allow_early_free` doesn't take effect
- If `loadarg.clone_wasm_binary` is false, wasm-c-api doesn't clone the binary, and if `loadarg.allow_early_free` is true, the wasm/aot loader clones data if needed

> > And what should the behavior be for wasm/aot loader and wamr-c-api loader?
> 
> i'm not sure if i understand this sentence. are you talking about the default behavior?

I mean the wamr API wasm_runtime_load, wasm_runtime_load_ex and wasm-c-api wasm_module_new, wasm_module_new_ex. Suggest to keep the behavior of wasm_runtime_load and wasm_module_new unchanged, and change the behavior of wasm_runtime_load_ex and wasm_module_new_ex according to the option.

eloparco:
>This makes that the binary can be freed only after instantiation, is it better to clone data segment in wasm module and clone string_literal_ptrs in both wasm and aot modules when wasm_binary_freeable is true? So that we can free the input buffer after loading. If it is a little complex, we can help do it after this PR is merged.

Been trying to use some c sample with threads (https://github.com/bytecodealliance/wasm-micro-runtime/tree/591a20b91741c24cdb9c031d3c5eec1d80bd72dc/core/iwasm/libraries/lib-wasi-threads/test) compiled with wasi-sdk. I see that in such case passive data segments are used and not all are dropped after `memory.init`, leading to `wasm_runtime_is_underlying_binary_freeable` returning false because of https://github.com/bytecodealliance/wasm-micro-runtime/blob/591a20b91741c24cdb9c031d3c5eec1d80bd72dc/core/iwasm/common/wasm_runtime_common.c#L7283.
Both `memory.init` and `memory.drop` are called during instantiation in the wasm start function.

Cloning the data segments into the data module would be the only way to free the wasm binary buffer in that case.

yamt:
> > This makes that the binary can be freed only after instantiation, is it better to clone data segment in wasm module and clone string_literal_ptrs in both wasm and aot modules when wasm_binary_freeable is true? So that we can free the input buffer after loading. If it is a little complex, we can help do it after this PR is merged.
> 
> Been trying to use some c sample with threads (https://github.com/bytecodealliance/wasm-micro-runtime/tree/591a20b91741c24cdb9c031d3c5eec1d80bd72dc/core/iwasm/libraries/lib-wasi-threads/test) compiled with wasi-sdk. I see that in such case passive data segments are used and not all are dropped after `memory.init`, leading to `wasm_runtime_is_underlying_binary_freeable` returning false because of
> 
> https://github.com/bytecodealliance/wasm-micro-runtime/blob/591a20b91741c24cdb9c031d3c5eec1d80bd72dc/core/iwasm/common/wasm_runtime_common.c#L7283
> .

the test only makes sense after instantiation. i'm not sure what you are trying to test.

> Both `memory.init` and `memory.drop` are called during instantiation in the wasm start function.

it's expected for the data segment for shared memory initialization.

> 
> Cloning the data segments into the data module would be the only way to free the wasm binary buffer in that case.

what's "the data module"?

yamt:
> > > Is that to allow the developer to free the wasm binary after wasm/aot loading?
> > 
> > 
> > it tells the runtime to perform something which possibly allows early free (typically making extra copies) if it makes sense. otherwise, the runtime can ignore it and make wasm_runtime_is_underlying_binary_freeable return false.
> 
> Yes, how about `loadarg.allow_early_free` instead of `loadarg.want_early_free`?

either names are fine for me.

my point is to make it a hint so that the runtime can make the decision by itself.

> And per my underderstanding, it allows to free the binary after loading, but not after instantiation, and we will change argument `wasm_module_inst_t module_inst` of wasm_runtime_is_underlying_binary_freeable to `wasm_module_t module`, just like we discussed before in this PR.

i guess it has pros and cons.
right now i don't have a clear opinion about which would be better.

> > > And do you mean to use the same option for both wasm-c-api and wamr api?
> > 
> > 
> > maybe. i haven't thought much about wasm-c-api yet.
> 
> By default wasm-c-api will clone the binary and already allows to early free, so using the option doesn't make sense for it. Maybe we can also keep `loadarg.clone_wasm_binary` for wasm-c-api, and:
> 
>     * If `loadarg.clone_wasm_binary` is true, `loadarg.allow_early_free` doesn't take effect
> 
>     * If `loadarg.clone_wasm_binary` is false, wasm-c-api doesn't clone the binary, and if `loadarg.allow_early_free` is true, the wasm/aot loader clones data if needed
> 
> 
> > > And what should the behavior be for wasm/aot loader and wamr-c-api loader?
> > 
> > 
> > i'm not sure if i understand this sentence. are you talking about the default behavior?
> 
> I mean the wamr API wasm_runtime_load, wasm_runtime_load_ex and wasm-c-api wasm_module_new, wasm_module_new_ex. Suggest to keep the behavior of wasm_runtime_load and wasm_module_new unchanged, and change the behavior of wasm_runtime_load_ex and wasm_module_new_ex according to the option.

i agree the user-visible semantics of old/standard api should not be changed.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -7170,3 +7179,51 @@ wasm_runtime_detect_native_stack_overflow_size(WASMExecEnv *exec_env,
...
+            return false;
+#if WASM_ENABLE_GC != 0 && WASM_ENABLE_STRINGREF != 0
+        if (module->string_literal_ptrs)
+            return false;
+#endif
+#if WASM_ENABLE_BULK_MEMORY != 0
+        for (i = 0; i < module->data_seg_count; i++)
+            if (!bh_bitmap_get_bit(
+                    ((WASMModuleInstance *)module_inst)->e->common.data_dropped,
+                    i))
```
yamt, line 7208: are you assuming this module instance is the only one which is based on the underlying buffer?
it seems wrong because our instantiation api allows to instantiate multiple instances from a single wasm_module_t.

eloparco:
>the test only makes sense after instantiation. i'm not sure what you are trying to test.

Yes, I was calling `wasm_runtime_is_underlying_binary_freeable` after module instantiation to make sure it was safe to free the wasm binary buffer. But `memory.drop` was called on all but one data segment during module instantiation, making `wasm_runtime_is_underlying_binary_freeable` return false.

> it's expected for the data segment for shared memory initialization.

Yes, I was reading the specs and that's expected when using `passive` data segments (as it happens when compiling using WASI threads).

> what's "the data module"?

Sorry, I just meant the module. Since passive data segments can be referred to after module instantiation, I think we need to clone them into the module if we want to free the wasm binary buffer.

core/iwasm/common/wasm_runtime_common.c
eloparco: Yeah, I think that was a simplification, but as you say it doesn't work if there are multiple instances created from the same module. We should be able to remove that check if we clone the data segments into the module, if I understand correctly.

wenyongh:
@yamt since I want to merge branch main into dev/checkpoint_and_restore for @victoryang00 to have a rebase for PR #3289, I merged this PR. I think it basically implemented the ideas we discussed above, (maybe except change the option name loadarg.wasm_binary_freeable to loadard.allow_early_free?). If you prefer to use `allow_early_free` or have more ideas on this feature, we can go on discussion and submit another PR to refactor the code.

yamt:
> @yamt since I want to merge branch main into dev/checkpoint_and_restore for @victoryang00 to have a rebase for PR #3289, I merged this PR. I think it basically implemented the ideas we discussed above, (maybe except change the option name loadarg.wasm_binary_freeable to loadard.allow_early_free?). If you prefer to use `allow_early_free` or have more ideas on this feature, we can go on discussion and submit another PR to refactor the code.

i still think it should be runtime's decision, not user's.
otherwise, the api is too complex for a user to use.

wenyongh:
> > @yamt since I want to merge branch main into dev/checkpoint_and_restore for @victoryang00 to have a rebase for PR #3289, I merged this PR. I think it basically implemented the ideas we discussed above, (maybe except change the option name loadarg.wasm_binary_freeable to loadard.allow_early_free?). If you prefer to use `allow_early_free` or have more ideas on this feature, we can go on discussion and submit another PR to refactor the code.
> 
> i still think it should be runtime's decision, not user's. otherwise, the api is too complex for a user to use.

Not sure what runtime decision means, do you mean that we don't provide `loadarg.wasm_binary_freeable` for wasm_runtime_load_ex and let runtime just treat it as true by default?

yamt:
> > > @yamt since I want to merge branch main into dev/checkpoint_and_restore for @victoryang00 to have a rebase for PR #3289, I merged this PR. I think it basically implemented the ideas we discussed above, (maybe except change the option name loadarg.wasm_binary_freeable to loadard.allow_early_free?). If you prefer to use `allow_early_free` or have more ideas on this feature, we can go on discussion and submit another PR to refactor the code.
> > 
> > 
> > i still think it should be runtime's decision, not user's. otherwise, the api is too complex for a user to use.
> 
> Not sure what runtime decision means, do you mean that we don't provide `loadarg.wasm_binary_freeable` for wasm_runtime_load_ex and let runtime just treat it as true by default?

with the current api, the decision to use the loagarg option is a bit too complex for users to make.
it might decrease the memory usage, or increase the memory usage, or whatever, depending on module type and runtime build options.

as the runtime itself should have a better knowledge, it's better to make users just express the intention
("i want to save memory, accepting documented tradeoffs") and leave the copy-or-not decision to the runtime.

wenyongh:
OK, IIUC, I think you mean that developer sets a flag in loadarg to express that he wants to reduce the memory usage, and runtime tries to achieve it, there might be two options for runtime, one is to clone the data and allow the developer to free the input wasm buffer, the other is to let wasm module refer to the input buffer and not to allow freeing it. But I wonder it may be very complex for runtime, what the strategy should be?

@eloparco any idea from you? Thanks.

eloparco:
> it's better to make users just express the intention

Isn't that similar to what we do already? We set `load_args.clone_wasm_binary = false;` to express the intention and `wasm_module_is_underlying_binary_freeable` to check if the runtime was able to satisfy it. Can't think of anything simpler.

yamt:
> > it's better to make users just express the intention
> 
> Isn't that similar to what we do already? We set `load_args.clone_wasm_binary = false;` to express the intention and `wasm_module_is_underlying_binary_freeable` to check if the runtime was able to satisfy it. Can't think of anything simpler.

the problem in the current api is that it's difficult for a generic embedder (eg. iwasm) to decide optimal load_args parameters.

Issue #3388, opened 2024-05-02
Unexpected execution result on fast interpreter mode


erxiaozhou:
### Subject of the issue
The WAMR on faster interpreter mode has a wrong execution result on the test case.
### Test case

[tt3.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/15186503/tt3.zip)

```
(module
  (type (;0;) (func))
  (type (;1;) (func (param i64 i64) (result i64 i64)))
  (func (;0;) (result i64)
    i64.const 230
    i64.const 366
    i64.div_u
    i64.const 968
    call 1
    i32.const 0
    if (param i64 i64) (result i64 i64)  ;; label = @1
    end
    i64.div_u
    ;; global.get 0
    ;; i64.add
    ;; global.set 0
    )
  (func (;1;) (type 1) (param i64 i64) (result i64 i64)
    i64.const 0
    i64.const 9223372036854775807)
  (global (;0;) (mut i64) (i64.const 853))
  (global (;1;) (mut f32) (f32.const 0x1.2f312cp+3 (;=9.47475;)))
  (export "_start" (func 0))
  (export "to_test" (func 0)))


```

### Your environment

* Host OS: Ubuntu 20.04
* Commit Id: 835188cc53373159165423debc48242ad15548d6

### Steps to reproduce
1. build WAMR in fast interpreter mode
2. execute the command
```
iwasm --heap-size=0 -f "to_test" <test_case>
```

### Expected behavior
```
0
```
### Actual behavior
An exception "Exception: integer divide by zero"

wenyongh:
@erxiaozhou thanks for reporting the issue, it is similar to #3386 when handling the if block in loader:
```wast
    if (param i64 i64) (result i64 i64)  ;; label = @1
    end
```
Please try #3404.

wenyongh:
Close the issue since it was fixed.

Issue #3387, opened 2024-05-02
Unexpected execution in fast interpreter mode


erxiaozhou:
### Subject of the issue
The WAMR on faster interpreter mode has a wrong execution result on the test case.
### Test case
[tt2.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/15186308/tt2.zip)

```
(module
  (type (;0;) (func))
  (type (;1;) (func (result i32)))
  (type (;2;) (func (param f32) (result f32)))
  (func (;0;)  (result i64)
    i32.const -1518330408
    i64.const 1022
    i64.const 904
    i64.add
    global.get 0
    i64.add
    global.set 0
    f32.const 0x1.01499cp+1 (;=2.01006;)
    call 1
    i32.const 0
    if (param f32) (result f32)  ;; label = @1
    end
    i64.trunc_f32_u
    return
    )
  (func (;1;) (type 2) (param f32) (result f32)
    f32.const inf (;=inf;))
  (global (;0;) (mut i64) (i64.const 858))
  (global (;1;) (mut f64) (f64.const 0x1.0370499c98398p+3 (;=8.10746;)))
  (export "_start" (func 0))
  (export "to_test" (func 0)))


```

### Your environment

* Host OS: Ubuntu 20.04
* Commit Id: 835188cc53373159165423debc48242ad15548d6

### Steps to reproduce
1. build WAMR in fast interpreter mode
2. execute the command
```
iwasm --heap-size=0 -f "to_test" <test_case>
```

### Expected behavior
An exception indicating integer overflow (On the instruction i64.trunc_f32_u)
### Actual behavior

```
0
```

wenyongh:
@erxiaozhou thanks for reporting the issue, it is similar to #3386 when handling the if block in loader:
```wast
    if (param f32) (result f32)  ;; label = @1
    end
```
Please try #3404.

wenyongh:
Close the issue since it was fixed.

Issue #3386, opened 2024-05-02
Unexpected execution result on fast interpreter mode


erxiaozhou:
### Subject of the issue
The WAMR on faster interpreter mode has a wrong execution result on the test case.
### Test case

[tt.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/15186123/tt.zip)
```
(module
  (type (;0;) (func))
  (type (;1;) (func (result i32)))
  (type (;2;) (func (param i32 i32) (result i32 i32)))
  (func (;0;) (type 1) (result i32)
    f32.const -0x1.0bb6d6p+2 (;=-4.18303;)
    f32.const -0x1.2a640ap+2 (;=-4.66236;)
    f32.ge
    i32.const 353
    call 1
    i32.const 0
    if (param i32 i32) (result i32 i32)  ;; label = @1
    end
    i32.gt_u
    return
    )
  (func (;1;) (type 2) (param i32 i32) (result i32 i32)
    i32.const -1
    i32.const 2147483647)
  (export "_start" (func 0))
  (export "to_test" (func 0)))

```

### Your environment

* Host OS: Ubuntu 20.04
* Commit Id: 835188cc53373159165423debc48242ad15548d6

### Steps to reproduce
1. build WAMR in fast interpreter mode
2. execute the command
```
iwasm --heap-size=0 -f "to_test" <test_case>
```

### Expected behavior
1
### Actual behavior
```
0
```

wenyongh:
@erxiaozhou thanks for reporting the issue, it is mainly caused by the handling of an if block without else opcode for fast interpreter:
```wast
    if (param i32 i32) (result i32 i32)  ;; label = @1
    end
```
I submitted PR #3404 to fix it, please try again.

PR #3404, opened 2024-05-08
Fix loader handling if block without op else

wenyongh:
If there is no else branch for the if block, make a virtual else opcode for
easier checking and to copy the correct results to the block return address
in fast-interp mode.

Reported in issue #3386, #3387, #3388.


xujuntwt95329 (commented):
LGTM

PR #3406, opened 2024-05-08
Add module global type information APIs

bnason-nf:


core/iwasm/interpreter/wasm.h
```diff
@@ -584,11 +584,15 @@ typedef struct WASMTagImport {
 } WASMTagImport;
 #endif
 
+typedef struct WASMGlobalCommon {
+    uint8 type;
+    bool is_mutable;
+} WASMGlobalCommon;
+
```
lum1n0us, line 591: just my personal suggestion. 

```c
typedef struct WASMGlobalType {
    uint8 val_type;
    bool  is_mutable;
} WASMGlobalType;
```

core/iwasm/interpreter/wasm.h
```diff
@@ -584,11 +584,15 @@ typedef struct WASMTagImport {
...
 typedef struct WASMGlobalImport {
     char *module_name;
     char *field_name;
-    uint8 type;
-    bool is_mutable;
+    WASMGlobalCommon common;
```
lum1n0us, line 595:
```suggestion
    WASMGlobalCommon type;
```

less confusion.

core/iwasm/interpreter/wasm.h
```diff
@@ -705,8 +709,7 @@ struct WASMTag {
 #endif
 
 struct WASMGlobal {
-    uint8 type;
-    bool is_mutable;
+    WASMGlobalCommon common;
```
lum1n0us, line 712:
```suggestion
    WASMGlobalCommon type;
```

less confusion

lum1n0us:
Plus, I kindly thought we should use `wasm_runtime_get_import_type` and `wasm_runtime_get_export_type` to implement `wasm_module_exports` and `wasm_module_imports` in order to keep the code simple.

core/iwasm/interpreter/wasm.h
bnason-nf: I'll make this change.

core/iwasm/interpreter/wasm.h
bnason-nf: I'll make this change.

core/iwasm/interpreter/wasm.h
bnason-nf: I'll make this change.

bnason-nf:
> Plus, I kindly thought we should use `wasm_runtime_get_import_type` and `wasm_runtime_get_export_type` to implement `wasm_module_exports` and `wasm_module_imports` in order to keep the code simple.

I think we can do that once all the import/export types exposed in wasm_export.h. After this change lands and globals are exposed, then also memory and tables would need to be exposed. Once functions, globals, memory, and tables are all exposed, then some of the internals of `wasm_module_exports` and `wasm_module_exports` can be converted to use those new APIs. Unless maybe I'm missing something, and you see a way to partially convert those function internals?

lum1n0us (approved):
LGTM

wenyongh (approved):

Issue #3409, opened 2024-05-09
Compilation errors when build iwasm with GC and EH support


bashor:
### Steps to reproduce

Try to build `iwasm` with GC and EH support, for example, with the following command:
```shell
cmake .. -DWAMR_BUILD_GC=1 -DWAMR_BUILD_EXCE_HANDLING=1 && make
```

Tell us how to reproduce this bug or defect.

### Expected behavior

It's compiled successfully and features work properly.

### Actual behavior

Got errors like:
```
wasm-micro-runtime/core/iwasm/interpreter/wasm_loader.c:2958:27: error: 'WASMType' has no member named 'result_count'; did you mean 'ref_count'?
....
```

PR #3413, opened 2024-05-10
Correct Exception Handling tag type when GC is enabled

TianlongLiang:
Use `WASMFuncType` to represent tag_type in `WASMTagImport` and `WASMTag` so that the type definition is consistent no matter to GC is enabled or disabled. This patch will fix https://github.com/bytecodealliance/wasm-micro-runtime/issues/3409

### DIFF
--- a/core/iwasm/interpreter/wasm_loader.c
+++ b/core/iwasm/interpreter/wasm_loader.c
@@ -11196,13 +11196,9 @@ re_scan:
 #if WASM_ENABLE_GC != 0
                     local_type = tag_type->types[tti];
                     local_idx = tti;
-                    bool is_type_multi_byte =
-                        wasm_is_type_multi_byte_type(local_type);
-                    WASMRefType *ref_type = NULL;
-                    if (is_type_multi_byte) {
-                        bh_assert(param_reftype_maps);
-                        GET_LOCAL_REFTYPE();
-                    }
+                    /* Get the wasm_ref_type if the local_type is multibyte
+                     * type */
+                    GET_LOCAL_REFTYPE();
 #endif
 
                     if (!check_stack_top_values(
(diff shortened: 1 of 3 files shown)


core/iwasm/interpreter/wasm_loader.c
```diff
@@ -11180,11 +11181,40 @@ wasm_loader_prepare_bytecode(WASMModule *module, WASMFunction *func,
...
+#endif
+
                 for (int tti = (int32)tag_type->param_count - 1; tti >= 0;
                      tti--) {
+#if WASM_ENABLE_GC != 0
+                    local_type = tag_type->types[tti];
+                    local_idx = tti;
+                    bool is_type_multi_byte =
+                        wasm_is_type_multi_byte_type(local_type);
+                    WASMRefType *ref_type = NULL;
```
wenyongh, line 11201: ref_type isn't used?

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -11180,11 +11181,40 @@ wasm_loader_prepare_bytecode(WASMModule *module, WASMFunction *func,
...
+                    if (is_type_multi_byte) {
```
wenyongh, line 11202: Seems no need to check again? GET_LOCAL_REFTYPE macro already checks `if (wasm_is_type_multi_byte_type(local_type))`

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -11180,11 +11181,40 @@ wasm_loader_prepare_bytecode(WASMModule *module, WASMFunction *func,
...
+
                     if (!check_stack_top_values(
                             loader_ctx, frame_ref, available_stack_cell,
-                            tag_type->types[tti], error_buf, error_buf_size)) {
+#if WASM_ENABLE_GC != 0
+                            frame_reftype_map, frame_reftype_map_num,
+#endif
+                            tag_type->types[tti],
+#if WASM_ENABLE_GC != 0
+                            ref_type,
```
wenyongh, line 11215: Seems ref_type isn't set above, do you mean `WASMRefType *ref_type = &wasm_ref_type`?

TianlongLiang: I added a comment and use `&wasm_ref_type` directly

wenyongh (approved):
LGTM

Issue #3402, opened 2024-05-08
Unexpected behavior on fast interpreter mode


erxiaozhou:
### Subject of the issue
WAMR on fast interpreter mode raises an unexpected exception.
### Test case

[tt_finterp_exceptionx.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/15244715/tt_finterp_exceptionx.zip)

### Your environment

* Host OS: Ubuntu 20.04
* Commit Id: 1c2a8fca4e9bf969ae27240a0918ee8164dfd598

### Steps to reproduce
1. build WAMR on fast interpreter mode
2. execute the command
```
iwasm --heap-size=0 -f "to_test" <test_case>
```
### Expected behavior
No exception
```
0x2d0:i64
```
### Actual behavior
An exception
```
Exception: integer overflow
```

wenyongh:
@erxiaozhou thanks for reporting the issue, please try #3434.

wenyongh:
Close the issue since it was fixed.

PR #3434, opened 2024-05-14
Fix fast interp RECOVER_BR_INFO and local set/tee

wenyongh:
When copying two cells from src offsets to dst offsets in RECOVER_BR_INFO,
the offsets may be overlapped and the src data may be overwritten, use
GET_I64_FROM_ADDR and then GET_I64_FROM_ADDR instead to resolve it.

And handling VALUE_TYPE_FUNCREF/VALUE_TYPE_EXTERNREF for opcode
local.set and local.tee when reference types feature is enabled.

This PR fixes issue #3401 and #3402.


xujuntwt95329 (commented):
LGTM

PR #3441, opened 2024-05-16
Fix some typos

TianlongLiang:


core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -2364,12 +2364,9 @@ wasm_instantiate(WASMModule *module, WASMModuleInstance *parent,
...
 
-        bh_assert(
-            data_seg->base_offset.init_expr_type == INIT_EXPR_TYPE_GET_GLOBAL
-            || (data_seg->base_offset.init_expr_type == INIT_EXPR_TYPE_I32_CONST
-                && !memory->is_memory64)
-            || (data_seg->base_offset.init_expr_type == INIT_EXPR_TYPE_I64_CONST
-                && memory->is_memory64));
+        bh_assert(data_seg->base_offset.init_expr_type
+                  == (memory->is_memory64 ? INIT_EXPR_TYPE_I64_CONST
+                                          : INIT_EXPR_TYPE_I32_CONST));
```
wenyongh, line 2369: `INIT_EXPR_TYPE_GET_GLOBAL` is removed?

core/shared/utils/bh_log.c
```diff
@@ -38,7 +38,7 @@ bh_log(LogLevel log_level, const char *file, int line, const char *fmt, ...)
     t = t % (60 * 60);
     m = t / 60;
     s = t % 60;
-    mills = (uint32)(usec % 1000);
+    mills = (uint32)(usec / 1000);
```
wenyongh, line 41: Should not change the code? `s = t % 60` is used to get the second value, here `mills = (uint32)(usec % 1000)` is to get the remainder, it should be smaller than 1 second, or 1000 ms.

TianlongLiang: At the beginning of the function, we got usec(microsecond=10-6 second), in order to get millisecond(10-3 second), we should divide it by 1000 rather than get the remainder of usec(still microsecond), I observed a non-incremental log time like `00:00:00:792` to `00:00:00:101` so I traced back here

core/iwasm/interpreter/wasm_runtime.c
TianlongLiang: deleted by mistake, added it back here

core/shared/utils/bh_log.c
wenyongh: But it will keep increasing and be larger than 1000 if using `/`. I guess I know why, seems we should use `mills = (uint32)((usec % 1000000) / 1000)` which limits the usec to be in 1 second first.

wenyongh (approved):

PR #3443, opened 2024-05-16
Add cmake static/shared library build settings

bnason-nf:
Allow builds to target either static library only, shared library only, or both (default).


lum1n0us:
Why not using `cmake --target iwasm_static` or `cmake --target iwasm_shared` to control?

bnason-nf:
As far as I know, the CMake `--target` option is only applicable for wasm-micro-runtime in the case where you are invoking it directly at the top level. If instead you are using wasm-micro-runtime with `add_subdirectory()` in a parent CMake file, there is currently no way to select the target(s).

lum1n0us (approved):
LGTM

CMakeLists.txt
```diff
@@ -136,30 +143,34 @@ endif ()
...
-if (WAMR_BUILD_WASM_CACHE EQUAL 1)
-  target_link_libraries(iwasm_static INTERFACE boringssl_crypto)
-endif ()
+if (WAMR_BUILD_STATIC)
+    add_library(iwasm_static STATIC ${WAMR_RUNTIME_LIB_SOURCE})
+    set_target_properties (iwasm_static PROPERTIES OUTPUT_NAME vmlib)
+    target_include_directories(iwasm_static INTERFACE ${WAMR_ROOT_DIR}/core/iwasm/include)
+    target_link_libraries (iwasm_static INTERFACE ${LLVM_AVAILABLE_LIBS} ${UV_A_LIBS} -lm -ldl -lpthread)
+    if (WAMR_BUILD_WASM_CACHE EQUAL 1)
+    target_link_libraries(iwasm_static INTERFACE boringssl_crypto)
```
wenyongh, line 152: Had better also change indent

CMakeLists.txt
```diff
@@ -136,30 +143,34 @@ endif ()
...
-if (WAMR_BUILD_WASM_CACHE EQUAL 1)
-  target_link_libraries(iwasm_shared INTERFACE boringssl_crypto)
-endif ()
+if (WAMR_BUILD_SHARED)
+    add_library (iwasm_shared SHARED ${WAMR_RUNTIME_LIB_SOURCE})
+    set_target_properties (iwasm_shared PROPERTIES OUTPUT_NAME iwasm)
+    target_include_directories(iwasm_shared INTERFACE ${WAMR_ROOT_DIR}/core/iwasm/include)
+    target_link_libraries (iwasm_shared INTERFACE ${LLVM_AVAILABLE_LIBS} ${UV_A_LIBS} -lm -ldl -lpthread)
+    if (WAMR_BUILD_WASM_CACHE EQUAL 1)
+    target_link_libraries(iwasm_shared INTERFACE boringssl_crypto)
```
wenyongh, line 165: ditto

CMakeLists.txt
```diff
@@ -136,30 +143,34 @@ endif ()
...
+    endif ()
 
-if (MINGW)
-  target_link_libraries (iwasm_shared INTERFACE -lWs2_32 -lwsock32)
-endif ()
+    if (MINGW)
+    target_link_libraries (iwasm_shared INTERFACE -lWs2_32 -lwsock32)
```
wenyongh, line 169: ditto

wenyongh (approved):

PR #3446, opened 2024-05-17
Fix llvm build error

byeongjee:
Otherwise we get error: externally-managed-environment.

I'm not sure if this would be the best way to handle the issue, but we are already doing this anyway (in Dockerfile, for example).

### DIFF
--- a/product-mini/platforms/android/build_llvm.sh
+++ b/product-mini/platforms/android/build_llvm.sh
@@ -3,5 +3,5 @@
 # Copyright (C) 2020 Intel Corporation. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-/usr/bin/env python3 -m pip install --user -r ../../../build-scripts/requirements.txt
+/usr/bin/env python3 -m pip install --user -r ../../../build-scripts/requirements.txt --break-system-packages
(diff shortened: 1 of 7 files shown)


wenyongh:
> Otherwise we get error: externally-managed-environment.
> 
> I'm not sure if this would be the best way to handle the issue, but we are already doing this anyway (in Dockerfile, for example).

Hi, what is the OS you used? We usually run build_llvm.sh in Ubuntu, MacOS and Windows and didn't find the issue. The `--break-system-packages` is added in Dockerfile and workflow since sometimes the OS in them is a little old, had better not add it to build_llvm.sh if not needed.

yamt:
> Hi, what is the OS you used? We usually run build_llvm.sh in Ubuntu, MacOS and Windows and didn't find the issue.

on macOS i have seen the error.
i don't know the exact cause or the correct fix.
i usually just remove `--user` and use virtualenv.

wenyongh:
> > Hi, what is the OS you used? We usually run build_llvm.sh in Ubuntu, MacOS and Windows and didn't find the issue.
> 
> on macOS i have seen the error. i don't know the exact cause or the correct fix. i usually just remove `--user` and use virtualenv.

Yes, maybe we can just manually update the script and had better not add `--break-system-packages` for many files and commit them, not sure whether it will cause unexpected behavior.

PR #3450, opened 2024-05-20
Enable AOT multi-module spec test

lum1n0us:
- doesn't support empty modules
- aot doesn't support import globals

### DIFF
--- a/tests/wamr-test-suites/spec-test-script/multi_module_ignore_cases.patch
+++ b/tests/wamr-test-suites/spec-test-script/multi_module_ignore_cases.patch
@@ -1,0 +1,124 @@
+diff --git a/test/core/imports.wast b/test/core/imports.wast
+index 0cc07cb..4e8367a 100644
+--- a/test/core/imports.wast
++++ b/test/core/imports.wast
+@@ -86,7 +86,7 @@
+ (assert_return (invoke "print64" (i64.const 24)))
+ 
+ (assert_invalid
+-  (module 
++  (module
+     (type (func (result i32)))
+     (import "test" "func" (func (type 1)))
+   )
+@@ -578,6 +578,7 @@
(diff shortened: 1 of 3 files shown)


tests/wamr-test-suites/test_wamr.sh
```diff
@@ -499,7 +499,7 @@ function spec_test()
             git apply ../../spec-test-script/simd_ignore_cases.patch || exit 1
         fi
         if [[ ${ENABLE_MULTI_MODULE} == 1 ]]; then
-            git apply ../../spec-test-script/multi_module_aot_ignore_cases.patch || exit 1
+            git apply ../../spec-test-script/multi_module_ignore_cases.patch || exit 1
```
wenyongh, line 502: Maybe we can:
```
    git apply ../../spec-test-script/multi_module_ignore_cases.patch || exit 1
    if [[ "${RUNNING_MODE}" == "aot" ]]; then
        git apply <extra patch for aot>|| exit 1
    fi
```

tests/wamr-test-suites/spec-test-script/multi_module_ignore_cases.patch
```diff
@@ -94,7 +100,19 @@ index 994e0f4..8fbcc02 100644
  
  (assert_unlinkable
    (module (import "Mg" "mut_glob" (global i32)))
-@@ -130,7 +130,7 @@
+@@ -93,6 +96,7 @@
+ )
+ 
+ 
++(; aot doesn't support to import globals
```
wenyongh, line 107: Had better make another patch for aot multi module, e.g. create a small patch to disable this case based on the patch for interpreter multi module, so here the case can be tested in interpreter mode.

wenyongh (approved):

PR #3452, opened 2024-05-20
Add global export apis

bnason-nf:
I'm very new to this code, so I'm not confident at all that this is the best way (or even a good way) to expose export globals, so please let me know if there is a better strategy or if I've just done it wrong. As a minor detail, I didn't see a good way in the AOT global_instantiate() to fill in the ref_type.

### DIFF
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -91,6 +91,9 @@ wasm_runtime_destroy_registered_module_list();
 
 #define E_TYPE_XIP 4
 
+static uint8
+val_type_to_val_kind(uint8 value_type);
+
 #if WASM_ENABLE_GC == 0 && WASM_ENABLE_REF_TYPES != 0
 /* Initialize externref hashmap */
 static bool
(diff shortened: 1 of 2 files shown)


core/iwasm/include/wasm_export.h
```diff
@@ -1407,6 +1407,70 @@ WASM_RUNTIME_API_EXTERN bool
...
+ * @param export_global the global instance
+ * @param value pointer to the value to store
+ *
+ * @return true if successful, false otherwise
+ *
+ */
+bool
+wasm_runtime_global_instance_set_value(wasm_global_instance_t export_global,
+                                       const wasm_val_t *value);
+
```
lum1n0us, line 1473: reading is fine. I am not sure there should be a writable API.

bnason-nf:
@lum1n0us can you please explain that a little more? I would think that in the case where the global is declared as mutable and exported from the WASM file, the intention is for it to be writable from outside of WASM code.

core/iwasm/include/wasm_export.h
```diff
@@ -1407,6 +1407,70 @@ WASM_RUNTIME_API_EXTERN bool
...
+ * Get the mutability of a global instance
+ *
+ * @param export_global the global instance
+ *
+ * @return true if mutable, false otherwise
+ *
+ */
+bool
+wasm_runtime_global_instance_get_mutable(wasm_global_instance_t export_global);
+
```
lum1n0us, line 1447: how about adding one new function to get global type from a global instance and reusing `wasm_global_type_get_valkind` and `wasm_global_type_get_mutable`? Then we can save one new adding

bnason-nf: @lum1n0us just to make sure I'm understanding, you're saying to create a new struct type to hold the mutable and valkind members, which we could expose in wasm_export.h? If so, what do you think would be a good name for that?

lum1n0us: I am thinking aobut `wasm_global_type_t`

lum1n0us:
> the intention is for it to be writable from outside of WASM code.

Is it about to operate `wasm export global` in a host function?

core/iwasm/include/wasm_export.h
bnason-nf: @lum1n0us can you please explain that a little more? I would think that in the case where the global is declared as mutable and exported from the WASM file, the intention is for it to be writable from outside of WASM code.

core/iwasm/include/wasm_export.h
bnason-nf: I guess we could just move this definition into wasm_export.h (currently it's in wasm.h):

```
typedef struct WASMGlobalType {
    uint8 val_type;
    bool is_mutable;
} WASMGlobalType;
```

bnason-nf: I think @wenyongh's idea below should work, and also addresses this issue, so I will try that.

bnason-nf:
@wenyongh I implemented something very close to your idea, except that I kept the lookup by name instead of by index, since to me it feels more natural to use it that way from the calling application code.

core/iwasm/include/wasm_export.h
lum1n0us: Yes. It's fine if a wasm module is designed to be host-wasm co-operated.

core/iwasm/include/wasm_export.h
```diff
@@ -275,6 +275,13 @@ typedef struct wasm_val_t {
 } wasm_val_t;
 #endif
 
+/* Global instance*/
+typedef struct wasm_global_inst_t {
+    wasm_valkind_t kind;
+    bool is_mutable;
```
lum1n0us, line 281: personal opinion:

``` c
typedef struct wasm_global_inst_t {
  wasm_global_type_t type;
  void *data;
}
```

bnason-nf: @lum1n0us since wasm_global_type_t is a pointer, I think this would require a memory allocation and the corresponding ownership issues. Also, the underlying non-pointer WASMGlobalType is not currently exposed, it's internal only. Is it worth making the necessary changes to avoid duplicating those two members?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1891,6 +1894,67 @@ wasm_runtime_set_module_inst(WASMExecEnv *exec_env,
     wasm_exec_env_set_module_inst(exec_env, module_inst);
 }
 
+bool
+wasm_runtime_get_export_global_inst(const wasm_module_inst_t module_inst,
```
wenyongh, line 1898: Had better use `WASMModuleInstanceCommon *const module_inst` like other places.

core/iwasm/include/wasm_export.h
wenyongh: Yes, wasm_global_type_t is an internal type, unless we add the declarations in wasm_export.h, but it seems no need to do that.

core/iwasm/include/wasm_export.h
wenyongh: The returned global_inst already has `void *global_data` field, developer can read/write from/to it, seems no need to add extra write API?

wenyongh (commented):
LGTM except minor comment.

core/iwasm/common/wasm_runtime_common.c
bnason-nf: I just committed a fix for this.

lum1n0us (approved):
LGTM

Issue #3430, opened 2024-05-13
printf_wrapper (libc builtin) makes multiple calls to os_printf with individual characters


malytomas:
I am using WAMR_BH_VPRINTF to provide custom function in place of os_printf. My function logs the formatted string into a file. Unfortunately, the https://github.com/bytecodealliance/wasm-micro-runtime/blob/8f098a5905d4b9a741933ffdb8a75f493e4e81d8/core/iwasm/libraries/libc-builtin/libc_builtin_wrapper.c#L407 separates a string into individual characters, calling os_printf multiple times, which breaks the logged messages.

wenyongh:
Hi, could you try defining macro BUILTIN_LIBC_BUFFERED_PRINTF and enlarging the buffer size with macro BUILTIN_LIBC_BUFFERED_PRINT_SIZE? It only outputs the buffer when `\n` is met or the buffer is full:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/libraries/libc-builtin/libc_builtin_wrapper.c#L348-L386

malytomas:
> Hi, could you try defining macro BUILTIN_LIBC_BUFFERED_PRINTF and enlarging the buffer size with macro BUILTIN_LIBC_BUFFERED_PRINT_SIZE? It only outputs the buffer when `\n` is met or the buffer is full: https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/libraries/libc-builtin/libc_builtin_wrapper.c#L348-L386

Hi, thanks for bringing these macros to my attention. I have not seen it before.

I have tried it and it loses some text from last few calls to printf in the app. I think preferred approach would be to actually have 1 to 1 mapping between printf in the app and os_printf in the embedder.

Furthermore, by looking in the code, the buffer is not thread safe, and allows mixing text from multiple wamr instances too.
Perhaps, could the buffer be moved inside str_context?

wenyongh:
Yes, moving buffer inside str_context is good, and the text losing may be caused by some remaining characters without '\n' are not dumped. I tried to refactor the code, could you try the patch below:
```diff
diff --git a/core/iwasm/libraries/libc-builtin/libc_builtin_wrapper.c b/core/iwasm/libraries/libc-builtin/libc_builtin_wrapper.c
index 24edd3c0..f4cf6b8e 100644
--- a/core/iwasm/libraries/libc-builtin/libc_builtin_wrapper.c
[...]

PR #3460, opened 2024-05-22
libc-builtin: Enhance buffered print for printf_wrapper

wenyongh:
Move print_buf and print_buf to str_context struct to support multi-threading
when BUILTIN_LIBC_BUFFERED_PRINTF macro is enabled, and fix the issue of
losing some text.

ps. https://github.com/bytecodealliance/wasm-micro-runtime/issues/3430.


TianlongLiang:
LGTM

Issue #3467, opened 2024-05-23
Unexpected behavior on fast interpreter mode


erxiaozhou:
### Subject of the issue
The test case is expected to lead to an exception "unreachable". However, the warm on fast interpreter mode fails to raise the exception. Is it caused by the optimization in the fast interpreter?
### Test case

[tt_unreachable.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/15418698/tt_unreachable.zip)

### Your environment

* Host OS: Ubuntu 20.04
* Commit Id: f378edc408382b7bcf517b463d4a54948739075b

### Steps to reproduce
1. build WAMR on fast interpreter mode
2. execute the command
```
iwasm --heap-size=0 -f "to_test" <test_case>
```
### Expected behavior
```
Exception: unreachable
```
### Actual behavior

```
-1.861157e+19:f32
```

wenyongh:
@erxiaozhou thanks for spotting the issue! I submitted PR #3482 to fix it, please try again.

wenyongh:
Close the issue as it was fixed.

Issue #3468, opened 2024-05-23
A bug in validation


erxiaozhou:
### Subject of the issue
There is a bug in the validation phase of WAMR, and the bug can be triggered by the following test case. 
The test case is invalid since it contains an illegally encoded block type (at offset 0x32e2). However, the test case can pass the validation of multiple runtimes, including classic interpreter, fast interpreter, JIT mode, fast JIT mode.

### Test case

[i64.add_2277BSBBJJJJJPSSSSSSSSSBBB_331_17164376207829096.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/15421510/i64.add_2277BSBBJJJJJPSSSSSSSSSBBB_331_17164376207829096.zip)

### Your environment

* Host OS: Ubuntu 20.04
* Commit Id: f378edc408382b7bcf517b463d4a54948739075b

### Steps to reproduce
1. build WAMR 
2. execute the command. E.g., 
```
iwasm --heap-size=0 -f "to_test" <test_case>  
```
### Expected behavior
An exception indicating "invalid function type"
### Actual behavior
```
0x0:i64
```

wenyongh:
@erxiaozhou thanks for spotting the issue! I submitted PR #3482 to fix it, please try again.

wenyongh:
Close the issue as it was fixed.

PR #3482, opened 2024-05-30
Parse block type with leb i32 encoding and correct dst dynamic offset for opcode br

wenyongh:
Fix several issues in wasm loader:
- Parse a block's type index with leb int32 instead leb uint32
- Correct dst dynamic offset of loop block arguments for opcode br
  when copying the stack operands to the arguments of loop block
- Free each frame_csp's param_frame_offsets when destroy loader ctx
- Fix compilation error in wasm_mini_loader.c
- Add test cases of failed issues

This PR fixes issue #3467 and #3468.


core/iwasm/interpreter/wasm_loader.c
```diff
@@ -8116,6 +8134,8 @@ wasm_loader_ctx_destroy(WASMLoaderContext *ctx)
         if (ctx->frame_csp_bottom) {
 #if WASM_ENABLE_FAST_INTERP != 0
             free_all_label_patch_lists(ctx->frame_csp_bottom, ctx->csp_num);
+            free_all_label_param_frame_offsets(ctx->frame_csp_bottom,
```
xujuntwt95329, line 8137: wasm_loader_pop_frame_csp will also free the param_frame_offsets, should we set it to NULL in wasm_loader_pop_frame_csp?

wenyongh: The wasm_loader_push_frame_csp will memset the frame csp with 0, seems no need to set param_frame_offsets to NULL again:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/67638e24f40a537b194a233a2d388f06db6d9546/core/iwasm/interpreter/wasm_loader.c#L8672-L8678

xujuntwt95329: Got it, thanks!

xujuntwt95329 (commented):
LGTM

PR #3485, opened 2024-05-30
Fix test-wamr-ide CI failure

wenyongh:
Set `moduleResolution` as `node16` in tsconfig.json.

ps.
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/9296272681/job/25586420457
```
node_modules/@vscode/test-electron/out/util.d.ts(1,23): error TS1452:
'resolution-mode' assertions are only supported when `moduleResolution` is `node16` or `nodenext`.
```


TianlongLiang:
LGTM

PR #3487, opened 2024-05-30
generate more kinds of corpus

lum1n0us:


tests/fuzz/wasm-mutator-fuzz/CMakeLists.txt
```diff
@@ -91,10 +91,7 @@ if (NOT DEFINED WAMR_BUILD_SIMD)
   set (WAMR_BUILD_SIMD 1)
 endif ()
 
-if (NOT DEFINED WAMR_BUILD_REF_TYPES)
-  # Disable reference types by default
-  set (WAMR_BUILD_REF_TYPES 0)
-endif ()
+set (WAMR_BUILD_REF_TYPES 1)
```
wenyongh, line 94: Could we keep `if (NOT DEFINED WAMR_BUILD_REF_TYPES) set (WAMR_BUILD_REF_TYPES 1) endif ()`? Maybe we need to disable reference types some time in the future.

tests/fuzz/wasm-mutator-fuzz/smith_wasm.sh
```diff
@@ -8,51 +8,61 @@
...
+mkdir -p "${corpusPath}"
 
 # 3.change dir
-# cd build && mkdir CORPUS_DIR && cd CORPUS_DIR
-cd build && mkdir CORPUS_DIR && cd CORPUS_DIR
+cd "${corpusPath}"
 
 # 4.generate *.wasm file
-echo "Generate $@ files according to user requirements"
+echo "To generate ${EXPECTED_NUM} Wasm files for each kind as required"
```
wenyongh, line 25: Had better not add `To`, Just `Generate ${EXPECTED_NUM} Wasm files for each kind as required`, it is the thing that is happening but not to happen in the future, adding To really makes confusing.

tests/fuzz/wasm-mutator-fuzz/workspace/CMakeLists.txt
```diff
@@ -91,10 +91,7 @@ if (NOT DEFINED WAMR_BUILD_SIMD)
   set (WAMR_BUILD_SIMD 1)
 endif ()
 
-if (NOT DEFINED WAMR_BUILD_REF_TYPES)
-  # Disable reference types by default
-  set (WAMR_BUILD_REF_TYPES 0)
-endif ()
+set (WAMR_BUILD_REF_TYPES 1)
```
wenyongh, line 94: Same as above

wenyongh (approved):

yzha107 (approved):
LGTM~ Thanks, Yi

TianlongLiang:
LGTM

PR #3492, opened 2024-06-03
Fix wasm loader check data segment count

wenyongh:
When datacount section exists, loader will check whether the data count
read from data segment section is same with the data count read from
datacount section, but the value of latter can be 0, loader should not
skip the check when the latter is 0.


xujuntwt95329 (commented):
LGTM

TianlongLiang:
LGTM

Issue #3381, opened 2024-04-30
[RFC] Basic support for Multi-Memory proposal


TianlongLiang:
# Summary

Plan to implement basic support for the multi-memory proposal, first focusing on the classic interpreter and AOT running modes.

## Overview of Basic Support Plan

With basic support completed, users should experience the following:

- Use `cmake -DWAMR_BUILD_MULTI_MEMORY=1` to compile a multi-memory-enabled iwasm VM core(Disable it by default). When the multi-memory macro is turned on, the wasm/aot runtime will be able to support both non-multi-memory and multi-memory wasm/aot files. 
- For the wamrc AOT compiler, no extra work is required during compile time. To compile a multi-memory aot file, use `wamrc --enable-multi-memory -o test.aot test.wasm` to generate a multi-memory aot file, will set the corresponding feature bit in the aot file if actually has multiple memory sections in the wasm file.
- When you integrate a WAMR VMcore into your host application, you can use certain APIs that WAMR exports. These APIs help manage linear memory and enable the conversion and verification of addresses between the host's native environment and the app's linear memory. 
  But for multi-memory, the behavior of those existing APIs will need some modification and a new set of counterpart (extended) APIs should be added. The existing APIs now use the assumingly main memory, namely the `first non-import memory[0]`(if at least one memory section is present) or `import_memory[0]`(if no memory section is present). Additionally, we should add a new set of APIs that take an extra argument mem_idx to manage the memories in multi-memory explicitly. The new APIs are as follows:
  
    ```C
    WASM_RUNTIME_API_EXTERN bool
    wasm_runtime_validate_app_addr_ex(WASMModuleInstanceCommon *module_inst, uint64 app_offset, uint64 size, uint32 memidx);

    WASM_RUNTIME_API_EXTERN bool
    wasm_runtime_validate_app_str_addr_ex(WASMModuleInstanceCommon *module_inst, uint64 app_str_offset, uint32 memidx);

    WASM_RUNTIME_API_EXTERN bool
    wasm_runtime_validate_native_addr_ex(WASMModuleInstanceCommon *module_inst, void *native_ptr, uint64 size, uint32 memidx);

    WASM_RUNTIME_API_EXTERN void *
    wasm_runtime_addr_app_to_native_ex(WASMModuleInstanceCommon *module_inst, uint64 app_offset, uint32 memidx);

    WASM_RUNTIME_API_EXTERN uint64
[...]

PR #3496, opened 2024-06-04
Add wasm_export.h APIs to expose memory type

bnason-nf:


core/iwasm/include/wasm_export.h
```diff
@@ -1436,6 +1477,21 @@ wasm_runtime_get_export_global_inst(const wasm_module_inst_t module_inst,
...
+ * @param name the export memory name
+ * @param memory_inst location to store the memory instance
+ *
+ * @return true if success, false otherwise
+ *
+ */
+WASM_RUNTIME_API_EXTERN bool
+wasm_runtime_get_export_memory_inst(const wasm_module_inst_t module_inst,
+                                    const char *name,
+                                    wasm_memory_inst_t *memory_inst);
```
wenyongh, line 1493: Don't need to export this API to get memory instance's base address and size, use:
```C
WASM_RUNTIME_API_EXTERN void *
wasm_runtime_addr_app_to_native(wasm_module_inst_t module_inst,
                                uint64_t app_offset);

WASM_RUNTIME_API_EXTERN bool
wasm_runtime_get_native_addr_range(wasm_module_inst_t module_inst,
                                   uint8_t *native_ptr,
                                   uint8_t **p_native_start_addr,
                                   uint8_t **p_native_end_addr);
```
instead, e.g.:
```C
uint8 *mem_begin_addr, *mem_end_addr;

mem_begin_addr = wasm_runtime_addr_app_to_native(module_inst, 0);

wasm_runtime_get_native_addr_range(module_inst, mem_begin_addr, NULL, &mem_end_addr);
```

core/iwasm/interpreter/wasm.h
```diff
@@ -536,10 +536,7 @@ typedef struct WASMTableImport {
 typedef struct WASMMemoryImport {
     char *module_name;
     char *field_name;
-    uint32 flags;
-    uint32 num_bytes_per_page;
-    uint32 init_page_count;
-    uint32 max_page_count;
+    WASMMemory memory;
```
wenyongh, line 539: How about here changing it to `WASMMemoryType mem_type`, and changing WASMMemory definition in the above to:
```C
typedef struct WASMMemory {
    uint32 flags;
    uint32 num_bytes_per_page;
    uint32 init_page_count;
    uint32 max_page_count;
} WASMMemory, WASMMemoryType;
```

core/iwasm/compilation/aot.h
```diff
@@ -81,23 +82,9 @@ typedef enum AOTFloatCond {
 typedef struct AOTImportMemory {
     char *module_name;
     char *memory_name;
-    uint32 memory_flags;
-    uint32 num_bytes_per_page;
-    uint32 mem_init_page_count;
-    uint32 mem_max_page_count;
+    AOTMemory memory;
```
wenyongh, line 85: How about adding `typedef WASMMemoryType AOTMemoryType` in the above, and here: `AOTMemoryType mem_type`?

core/iwasm/include/wasm_export.h
```diff
@@ -73,10 +73,13 @@ typedef enum {
...
 struct WASMFuncType;
-typedef struct WASMFuncType *wasm_func_type_t;
+typedef const struct WASMFuncType *wasm_func_type_t;
 
 struct WASMGlobalType;
-typedef struct WASMGlobalType *wasm_global_type_t;
+typedef const struct WASMGlobalType *wasm_global_type_t;
+
+struct WASMMemory;
+typedef const struct WASMMemory *wasm_memory_type_t;
```
wenyongh, line 82: Had better 
```C
struct WASMMemoryType;
typedef const struct WASMMemoryType *wasm_memory_type_t;
```

core/iwasm/include/wasm_export.h
```diff
@@ -1350,6 +1361,36 @@ wasm_global_type_get_valkind(const wasm_global_type_t global_type);
...
+
+/**
+ * Get the initial page count for a memory type
+ *
+ * @param memory_type the memory type
+ *
+ * @return the initial memory page count
+ */
+WASM_RUNTIME_API_EXTERN uint32_t
+wasm_memory_type_get_initial_page_count(const wasm_memory_type_t memory_type);
```
wenyongh, line 1382: Had better `wasm_memory_type_get_init_page_count`

core/iwasm/include/wasm_export.h
```diff
@@ -1350,6 +1361,36 @@ wasm_global_type_get_valkind(const wasm_global_type_t global_type);
...
+
+/**
+ * Get the maximum page count for a memory type
+ *
+ * @param memory_type the memory type
+ *
+ * @return the maximum memory page count
+ */
+WASM_RUNTIME_API_EXTERN uint32_t
+wasm_memory_type_get_maximum_page_count(const wasm_memory_type_t memory_type);
```
wenyongh, line 1392: Had better `wasm_memory_type_get_max_page_count`

core/iwasm/include/wasm_export.h
bnason-nf: Hi @wenyongh, I think that method is sufficient currently, but potentially future WebAssembly specifications might support  multiple `memory` objects in the wasm file:

> In the current version of WebAssembly, at most one memory may be defined or imported in a single module, and all constructs implicitly reference this memory. This restriction may be lifted in future versions.
(https://webassembly.github.io/spec/core/syntax/modules.html#syntax-mem)

Is that something we want to be able to support without changing the API design again? Or would it be such a comprehensive change that we'd expect to change a number of other APIs anyway?

core/iwasm/interpreter/wasm.h
bnason-nf: Yes this is better, I will change it.

core/iwasm/compilation/aot.h
bnason-nf: Agreed.

bnason-nf:
@wenyongh I keep having build failures because I miss one or more of the various configurations when building locally. Is there a good way to build everything locally that is built by the CI checks?

core/iwasm/include/wasm_export.h
wenyongh: The multi-memory for classic interpreter feature is under development and new APIs will be introduced:
https://github.com/bytecodealliance/wasm-micro-runtime/issues/3381

I think you can still use `wasm_runtime_addr_app_to_native_ex` and `wasm_runtime_get_native_addr_range_ex` by passing mem_idx argument to them. Had better not introduce new API if unnecessary since there are lots of APIs now.

bnason-nf: I understand, I have removed it.

core/iwasm/include/wasm_export.h
```diff
@@ -289,6 +295,12 @@ typedef struct wasm_global_inst_t {
     void *global_data;
 } wasm_global_inst_t;
 
+/* Memory instance*/
+typedef struct wasm_memory_inst_t {
+    void *address;
+    uint32_t size;
+} wasm_memory_inst_t;
```
wenyongh, line 302: Had better remove this struct since it is unused.

wenyongh (commented):
LGTM with a minor comment.

core/iwasm/include/wasm_export.h
bnason-nf: Thanks, good catch.

wenyongh (approved):

PR #3502, opened 2024-06-04
Fix wasm_mini_loader.c jit / multi-module compile

bnason-nf:


core/iwasm/interpreter/wasm_mini_loader.c
```diff
@@ -3233,8 +3233,11 @@ load(const uint8 *buf, uint32 size, WASMModule *module,
 }
 
 WASMModule *
-wasm_loader_load(uint8 *buf, uint32 size, const LoadArgs *args, char *error_buf,
-                 uint32 error_buf_size)
+wasm_loader_load(uint8 *buf, uint32 size,
+#if WASM_ENABLE_MULTI_MODULE != 0
+                 bool /* main_module */,
```
wenyongh, line 3238: Is it better to use `bool main_module`, not sure whether using `bool` without argument name will cause compiler warning/error?

wenyongh (approved):

core/iwasm/interpreter/wasm_mini_loader.c
bnason-nf: I'm not aware of any compilers that will warn for an unnamed parameter. However, if we gave it a name, then all compilers are capable of warning about unused named parameters, so we would have to add another block inside the function body inside the function like this:

```
#if WASM_ENABLE_MULTI_MODULE != 0
                 (void)main_module;
#endif
```

This is fine with me if you prefer it, just let me know.

wenyongh: Yes, maybe that's better, could you help update the code? Thanks.

wenyongh: @bnason-nf Since we planning to create a new release and will do more tests, I updated your code manually and commit it.

PR #3503, opened 2024-06-05
Fix posix no libc wasi debug interp build

bnason-nf:
This change supports building with `WAMR_BUILD_LIBC_WASI=0` and `WAMR_BUILD_DEBUG_INTERP=1`, otherwise the `os_socket_*` functions will be undefined.

### DIFF
--- a/core/shared/platform/common/posix/platform_api_posix.cmake
+++ b/core/shared/platform/common/posix/platform_api_posix.cmake
@@ -9,6 +9,11 @@ if (NOT WAMR_BUILD_LIBC_WASI EQUAL 1)
     list(REMOVE_ITEM source_all
         ${PLATFORM_COMMON_POSIX_DIR}/posix_file.c
         ${PLATFORM_COMMON_POSIX_DIR}/posix_clock.c
+    )
+endif()
+
+if ((NOT WAMR_BUILD_LIBC_WASI EQUAL 1) AND (NOT WAMR_BUILD_DEBUG_INTERP EQUAL 1))
+    list(REMOVE_ITEM source_all
         ${PLATFORM_COMMON_POSIX_DIR}/posix_socket.c
     )
 else()


core/shared/platform/common/posix/platform_api_posix.cmake
```diff
@@ -9,6 +9,11 @@ if (NOT WAMR_BUILD_LIBC_WASI EQUAL 1)
     list(REMOVE_ITEM source_all
         ${PLATFORM_COMMON_POSIX_DIR}/posix_file.c
         ${PLATFORM_COMMON_POSIX_DIR}/posix_clock.c
+    )
+endif()
+
+if ((NOT WAMR_BUILD_LIBC_WASI EQUAL 1) AND (NOT WAMR_BUILD_DEBUG_INTERP EQUAL 1))
```
wenyongh, line 15: This causes the condition for the `else()` branch below differs from the original one, how about:
```cmake
if (NOT WAMR_BUILD_LIBC_WASI EQUAL 1)
    list(REMOVE_ITEM source_all
        ${PLATFORM_COMMON_POSIX_DIR}/posix_file.c
        ${PLATFORM_COMMON_POSIX_DIR}/posix_clock.c
    )
    if (NOT WAMR_BUILD_DEBUG_INTERP EQUAL 1)
        list(REMOVE_ITEM source_all
             ${PLATFORM_COMMON_POSIX_DIR}/posix_socket.c
        )
    endif()
else()
    ...
endif()
```

bnason-nf:
Hi @wenyongh, yes, the intention is for the `else()` clause to be different. The reason is that when we are building with `WAMR_BUILD_DEBUG_INTERP=1`, we want to compile posix_socket.c, and we also want to use `platform_common_libc_util.cmake` regardless of the setting of `WAMR_BUILD_LIBC_WASI`. I tested all for combinations of `WAMR_BUILD_LIBC_WASI` and `WAMR_BUILD_DEBUG_INTERP` before submitting this pull request, so unless I made a mistake I think this is behaving as intended.

wenyongh:
> Hi @wenyongh, yes, the intention is for the `else()` clause to be different. The reason is that when we are building with `WAMR_BUILD_DEBUG_INTERP=1`, we want to compile posix_socket.c, and we also want to use `platform_common_libc_util.cmake` regardless of the setting of `WAMR_BUILD_LIBC_WASI`. I tested all four combinations of `WAMR_BUILD_LIBC_WASI` and `WAMR_BUILD_DEBUG_INTERP` before submitting this pull request, so unless I made a mistake I think this is behaving as intended.

OK, got it, thanks.

wenyongh (approved):
LGTM

Issue #3493, opened 2024-06-03
[wasi-nn] add more backends into `libraries/wasi-nn`


lum1n0us:
## Background

We are planning to add more backends into WAMR to enrich its capability in ML/LLM/AI field. Currently the only backend(in `libraries/wasi-nn`) is tensorflow-lite. 

The code in `core/iwasm/libraries/wasi-nn` can be separated into two parts:
- wasi-nn-general. It provides host implementation of wasi-nn APIs. It is the unchanged part for every backends. It will 
  1. accept parameters from the Wasm world
  2. translate parameters from Wasm world to host world
  3. call appropriate backend to finish requests based on `graph_encoding`
  4. translate results from host world to Wasm world
  5. send results to the Wasm world
- wasi-nn-backends. For now, it is only `wasi_nn_tensorflowlite.cpp`. It will use backends' APIs and libraries to finish requests. Usually, variant backends lead to variant implementation.

Every backend contains huge libraries. For example, libtensorflow-lite.a for the tensorflowlite backend is 651M. It is not possible to pack all backends, or even one backend, into the iwasm binary. A nature way is to create a shared library for each backend. 

## Proposal

There will be *libwasi_nn_backend_tflite.so*, *libwasi_nn_backend_openvino.so*, and so on. Users of iwasm need to pass the right backend w/ `--native-lib`, like `--native-lib=libwasi_nn_backend_tflite.so` or `--naive-lib=libwasi_nn_backend_pytorch.so` or something else,  before running a wasi-nn wasm application.

Each of them contains
- a shared wasi-nn host implementation(wasi-nn-general). Match `wasi_ephemeral_nn` imports requirements.
- a separated backend in variant framework. like w/ tflite, w/ openvino, w/ pytorch, w/ onnx, and so on.

## Choice: use vmlib.a or iwasm.so

Since wasi-nn-general(*wasi_nn.c*) using APIs of WAMR(like `wasm_runtime_malloc()`, `bh_hash_map_xxx()`, ...), the final *libwasi_nn_backend_xx.so* not only depends on ML framework libraries but also WAMR libraries, especially libvmlib.a or libiwasm.so. If realized iwasm also depends on the same library, a problem is raised.

[...]

PR #3509, opened 2024-06-06
make wasi-nn backends as separated shared libraries

lum1n0us:


lum1n0us:
@tonibofarull please make a review.  

BTW, as mentioned in https://github.com/bytecodealliance/wasm-micro-runtime/issues/3493, we're working on more backends.

wenyongh (approved):
LGTM

PR #3515, opened 2024-06-11
Add table type API support

bnason-nf:


core/iwasm/interpreter/wasm.h
```diff
@@ -480,18 +480,22 @@ typedef void *WASMString;
...
-     * 1: hax max size
+     * 1: has max size
      * 2: shared
      */
     uint8 flags;
     bool possible_grow;
     uint32 init_size;
     /* specified if (flags & 1), else it is 0x10000 */
     uint32 max_size;
+} WASMTableType;
```
wenyongh, line 495: Had better move field elem_ref_type from WASMTable/WASMTableImport into WASMTableType, it is also one of the basic type info:
```C
#if WASM_ENABLE_GC != 0
    WASMRefType *elem_ref_type;
#endif
```

core/iwasm/include/wasm_export.h
```diff
@@ -295,6 +300,14 @@ typedef struct wasm_global_inst_t {
     void *global_data;
 } wasm_global_inst_t;
 
+/* Table instance*/
+typedef struct wasm_table_inst_t {
+    wasm_valkind_t elem_kind;
+    uint32_t cur_size;
+    uint32_t max_size;
+    const void **elems;
```
wenyongh, line 308: Is there any requirement to get the element of table? This PR creates function instances for all functions in aot runtime, which increases the footprint, in the original design, we only create aot function instances for export functions, and if to call a function in the table, we can call wasm_runtime_call_indirect. Can we just add API to the function index of a specific table element?

core/iwasm/include/wasm_export.h
```diff
@@ -1386,6 +1399,46 @@ wasm_memory_type_get_init_page_count(const wasm_memory_type_t memory_type);
...
 
+/**
+ * Get the element kind for a table type
+ *
+ * @param table_type the table type
+ *
+ * @return the element kind
+ */
+WASM_RUNTIME_API_EXTERN wasm_valkind_t
+wasm_table_type_get_element_kind(const wasm_table_type_t table_type);
```
wenyongh, line 1410: How about `wasm_table_type_get_elem_kind`?

core/iwasm/include/wasm_export.h
wenyongh: OK, adding elems to wasm_table_inst_t sounds good, my concern is not to create all function instances in aot mode, and the type of elems should not be `const void **elems`, in fact, it is `uintptr_t *` when GC isn't enabled and `void **` when GC is enabled, see table_elem_type_t. So firstly, how about changing it to `void *elems` and add comment in wasm_export.h to it?

And secondly, for wasm_runtime_table_get_func_inst (suggest changing to wasm_table_get_func_inst?), we can get the function index from the element and then get the related function instance from module instance with the func index and the input module inst. Could we just get the function instance for the exported functions? e.g. return NULL for unexposed function so that in AOT we don't need to create function instance again?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1966,6 +1966,122 @@ wasm_runtime_get_export_global_inst(WASMModuleInstanceCommon *const module_inst,
...
+        }
+
+#if WASM_ENABLE_GC == 0
+        uint32 func_idx = (uint32)tbl_elem_val;
+#else
+        uint32 func_idx =
+            wasm_func_obj_get_func_idx_bound((WASMFuncObjectRef)tbl_elem_val);
+#endif
+
+        if (func_idx >= wasm_module_inst->e->function_count) {
```
wenyongh, line 2053: Had better just `bh_assert(func_idx < wasm_module_inst->e->function_count)`, func_idx must be smaller than function_count, or it is runtime error.

core/iwasm/aot/aot_runtime.c
```diff
@@ -1183,6 +1183,73 @@ init_func_ptrs(AOTModuleInstance *module_inst, AOTModule *module,
...
+            function->is_import_func = true;
+            function->func_name = module->import_funcs[func_idx].func_name;
+            function->func_index = func_idx;
+            function->u.func_import = &module->import_funcs[func_idx];
+        }
+        else {
+            /* instantiate non-import function */
+            uint32 ftype_index = module->func_type_indexes[func_idx];
+            function->is_import_func = false;
+            function->func_index = module->import_func_count + func_idx;
```
wenyongh, line 1239: Should be `function->func_index = func_idx`?

core/iwasm/aot/aot_runtime.c
```diff
@@ -1183,6 +1183,73 @@ init_func_ptrs(AOTModuleInstance *module_inst, AOTModule *module,
...
+        if (func_idx < module->import_func_count) {
+            /* instantiate function from import section */
+            function->is_import_func = true;
+            function->func_name = module->import_funcs[func_idx].func_name;
+            function->func_index = func_idx;
+            function->u.func_import = &module->import_funcs[func_idx];
+        }
+        else {
+            /* instantiate non-import function */
+            uint32 ftype_index = module->func_type_indexes[func_idx];
```
wenyongh, line 1237: Per my understanding, the func_idx here should be the index based on all functions (import + non-import), but `module->func_type_indexes` only stores func type indexes of non-import functions. Should use `module->func_type_indexes[func_idx - module->import_func_count]`?

core/iwasm/aot/aot_runtime.c
```diff
@@ -1183,6 +1183,73 @@ init_func_ptrs(AOTModuleInstance *module_inst, AOTModule *module,
...
+            function->is_import_func = false;
+            function->func_index = module->import_func_count + func_idx;
+            function->u.func.func_type =
+                (AOTFuncType *)module->types[ftype_index];
+            function->u.func.func_ptr = module->func_ptrs[func_idx];
```
wenyongh, line 1242: Should be `function->u.func.func_ptr = module->func_ptrs[func_idx - module->import_func_count];`?

core/iwasm/aot/aot_runtime.c
```diff
@@ -1650,6 +1716,9 @@ aot_instantiate(AOTModule *module, AOTModuleInstance *parent,
     if (!create_exports(module_inst, module, error_buf, error_buf_size))
         goto fail;
 
+    extra->functions = NULL;
+    extra->function_count = 0;
```
wenyongh, line 1720: The two fields should have been memset to 0, not setting here is also ok.

core/iwasm/aot/aot_runtime.c
```diff
@@ -2024,6 +2101,7 @@ invoke_native_with_hw_bound_check(WASMExecEnv *exec_env, void *func_ptr,
             void (*invoke_native)(void *func_ptr, void *exec_env, uint32 *argv,
                                   uint32 *argv_ret) =
                 func_type->quick_aot_entry;
+            exec_env->attachment = attachment;
```
wenyongh, line 2104: Have you found issue here? If added here, had better add `exec_env->attachment = NULL` after `invoke_native`, like wasm_runtime_invoke_native and wasm_runtime_invoke_native_raw.

wenyongh (approved):
LGTM

lum1n0us (approved):
LGTM

Issue #3514, opened 2024-06-07
Heap Buffer Overflow inside "wasm_loader_prepare_bytecode" Function


mobsceneZ:
### Subject of the issue

Running the CLI iwasm with the given testcase results in segmentation fault, which is caused by heap buffer overflow.

### Test case

[iwasm-poc-05.zip](https://github.com/user-attachments/files/15740484/iwasm-poc-05.zip)

### Your environment

```shell
OS               : Linux 5.15.146.1-microsoft-standard-WSL2 #1 SMP Thu Jan 11 04:09:03 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
Commit           : 028f43bc18494866c44666e54e9c5a2cd84152f5
Version          : 2.1.0
Clang Verison    : 13.0.0
Affected Tool    : iwasm
Enabled Features : Bulk Memory, Reference Types
```
### Steps to reproduce

```shell
Build            : cd product-mini/platforms/linux/ && mkdir -p build && cd build && export CC="clang" CXX="clang++" CFLAGS="-fsanitize=address -g" CXXFLAGS="-fsanitize=address -g" && cmake -DWAMR_BUILD_BULK_MEMORY=1 -DWAMR_BUILD_REF_TYPES=1 .. && make -j
Command          : iwasm -f main iwasm-poc-05
```

### Expected behavior

The program should exit gracefully with possibly some error information.

### Actual behavior

Here is the stack trace provided by AddressSanitizer:
```shell
=================================================================
==23895==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6060000001fe at pc 0x55ea15ed1c32 bp 0x7fff717e3730 sp 0x7fff717e3728
READ of size 2 at 0x6060000001fe thread T0
    #0 0x55ea15ed1c31 in wasm_loader_prepare_bytecode /home/lain/wamr/core/iwasm/interpreter/wasm_loader.c:12145:30
    #1 0x55ea15ea9c10 in load_from_sections /home/lain/wamr/core/iwasm/interpreter/wasm_loader.c:6130:14
    #2 0x55ea15ead2c5 in load /home/lain/wamr/core/iwasm/interpreter/wasm_loader.c:6536:13
    #3 0x55ea15ead2c5 in wasm_loader_load /home/lain/wamr/core/iwasm/interpreter/wasm_loader.c:6713:10
[...]

PR #3516, opened 2024-06-11
wasm loader: Fix pop invalid offset count when stack top is ANY

wenyongh:
In wasm_loader_pop_frame_offset, when the stack is in polymorphic state
and the stack top operand is VALUE_TYPE_ANY, if we popping I64/F64
operand, we should pop one offset but not two offsets.

The issue was reported in #3513 and #3514.


core/iwasm/interpreter/wasm_loader.c
```diff
@@ -10888,6 +10887,12 @@ wasm_loader_get_custom_section(WASMModule *module, const char *name,
 }
 #endif
 
+#if 0
+#define HANDLE_OPCODE(opcode) #opcode
+DEFINE_GOTO_TABLE(const char *, op_mnemonics);
```
TianlongLiang, line 10892: Is this mnemonics opcode table aimed to facilitate debugging in wasm_loader?

wenyongh: Yes, in wasm_loader_prepare_bytecode, when traversing opcodes, we can print opcode name, e.g. `printf("##%s\n", op_mnemonics[opcode]);`.

TianlongLiang: thanks, got it

TianlongLiang:
LGTM

PR #3518, opened 2024-06-12
Use another variable to check if in oss-fuzz environment

lum1n0us:
In order to pass a oss-fuzz build step *check_build*.

Previous `HELPER` is only for local test.

### DIFF
--- a/tests/fuzz/wasm-mutator-fuzz/CMakeLists.txt
+++ b/tests/fuzz/wasm-mutator-fuzz/CMakeLists.txt
@@ -120,8 +120,11 @@ add_compile_options(-fsanitize=fuzzer)
 add_link_options(-fsanitize=fuzzer)
 
 # if not calling from oss-fuzz helper, enable all support sanitizers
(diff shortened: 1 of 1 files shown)


tests/fuzz/wasm-mutator-fuzz/CMakeLists.txt
```diff
@@ -120,8 +120,11 @@ add_compile_options(-fsanitize=fuzzer)
 add_link_options(-fsanitize=fuzzer)
 
 # if not calling from oss-fuzz helper, enable all support sanitizers
-# oss-fuzz always defines `HELPER=True`
-if (NOT "$ENV{HELPER}" STREQUAL "True")
+# oss-fuzz will defines FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION in CFLAGS and CXXFLAGS
```
wenyongh, line 123: had better `defines` -> `define`

wenyongh (approved):
LGTM

PR #3519, opened 2024-06-12
wasm-mutator-fuzz: fix build errors and warnings for macOS

yamt:
Tested with the homebrew version of LLVM 15.
(xcode doesn't seem to ship the fuzzer runtime.)


wenyongh (approved):
LGTM

PR #3526, opened 2024-06-13
Fix several fuzzing issues

lum1n0us:
- UBsan detected: ``` unsigned integer overflow: 2684354559 * 2 cannot be represented in type 'uint32' (aka 'unsigned int') ```
- adjust compilation options

### DIFF
--- a/core/iwasm/interpreter/wasm_loader.c
+++ b/core/iwasm/interpreter/wasm_loader.c
@@ -380,8 +380,7 @@ loader_malloc(uint64 size, char *error_buf, uint32 error_buf_size)
 {
     void *mem;
 
-    if (size >= WASM_MEM_ALLOC_MAX_SIZE
-        || !(mem = wasm_runtime_malloc((uint32)size))) {
+    if (size >= UINT32_MAX || !(mem = wasm_runtime_malloc((uint32)size))) {
         set_error_buf(error_buf, error_buf_size, "allocate memory failed");
         return NULL;
     }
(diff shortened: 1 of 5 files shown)


lum1n0us:
Shall I upload the wasm binary file as a regression?

tests/fuzz/wasm-mutator-fuzz/CMakeLists.txt
```diff
@@ -131,9 +131,20 @@ string(FIND "${CFLAGS_ENV}" "-DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION" IN_OSS_
 if (IN_OSS_FUZZ EQUAL -1)
   message("[ceith]:Enable ASan and UBSan in non-oss-fuzz environment")
   add_compile_options(
-    -fsanitize=signed-integer-overflow
     -fprofile-instr-generate -fcoverage-mapping
+    -fno-sanitize-recover=all
     -fsanitize=address,undefined
+    # reference: https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html
```
lum1n0us, line 137: Sync this configuration up with config-common.cmake?

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -2255,9 +2255,16 @@ load_type_section(const uint8 *buf, const uint8 *buf_end, WASMModule *module,
...
+    uint32 default_max_size;
+
+    bool overflow = __builtin_umul_overflow(init_size, 2, &default_max_size);
+    if (overflow) {
+        default_max_size = UINT_MAX;
+    }
+    else {
+        if (default_max_size < WASM_TABLE_MAX_SIZE)
+            default_max_size = WASM_TABLE_MAX_SIZE;
+    }
```
wenyongh, line 2267: It is not easy to ensure that __builtin_umul_overflow is supported in all platforms and targets, how about changing to be somewhat like:
```C
    if (init_size * 2 < init_size) {
        default_max_size = UINT32_MAX;
    }
    else {
        default_max_size = init_size * 2 > WASM_TABLE_MAX_SIZE
                                  ? init_size * 2
                                  : WASM_TABLE_MAX_SIZE;
    }
```
And had better use UINT32_MAX instead of UINT_MAX?

wenyongh (approved):
LGTM

PR #3530, opened 2024-06-14
sync up with latest wasi-nn spec

lum1n0us:


lum1n0us:
@tonibofarull please take a look and leave your comments. Thanks ☕ 🍺

core/iwasm/libraries/wasi-nn/include/wasi_nn_types.h
```diff
@@ -48,8 +59,10 @@ typedef struct {
     uint32_t size;
 } tensor_dimensions;
 
+// sync up with
+// https://github.com/WebAssembly/wasi-nn/blob/main/wit/wasi-nn.wit#L27
 // The type of the elements in a tensor.
-typedef enum { fp16 = 0, fp32, up8, ip32 } tensor_type;
+typedef enum { fp16 = 0, fp32, fp64, bf16, u8, i32, i64 } tensor_type;
```
tonibofarull, line 65: I know that it is not a decision from your side, but this is a breaking change. How can we inform the users? Maybe we can add it in the README.

core/iwasm/libraries/wasi-nn/src/wasi_nn.c
```diff
@@ -308,15 +327,17 @@ wasi_nn_compute(wasm_exec_env_t exec_env, graph_execution_context ctx)
...
+
     WASINNContext *wasi_nn_ctx = wasm_runtime_get_wasi_nn_ctx(instance);
 
     wasi_nn_error res;
     if (success != (res = is_model_initialized(wasi_nn_ctx)))
         return res;
 
-    res = lookup[wasi_nn_ctx->current_encoding].compute(
-        wasi_nn_ctx->backend_ctx, ctx);
+    res = lookup.compute(wasi_nn_ctx->backend_ctx, ctx);
```
tonibofarull, line 340: All calls to lookup should be checked, right?

```suggestion
    if (lookup.compute) 
      res = lookup.compute(wasi_nn_ctx->backend_ctx, ctx);
    else
      NN_ERR_PRINTF("Backend not loaded.");
```

core/iwasm/libraries/wasi-nn/include/wasi_nn_types.h
lum1n0us: May I ask which part is *the breaking change*, `up8` -> `u8`? Or the whole PR?

core/iwasm/libraries/wasi-nn/src/wasi_nn.c
lum1n0us: Of course. Totally agree.

core/iwasm/libraries/wasi-nn/include/wasi_nn_types.h
tonibofarull: A WASM built before with tensor type `up8` now it will be interpreted as `fp64`. This is a breaking change from wasi-nn, but we might want to inform user to recompile.

tonibofarull (approved):

core/iwasm/libraries/wasi-nn/include/wasi_nn_types.h
lum1n0us: How about 

```
typedef enum { fp16 = 0, fp32, fp64, bf16, u8, i32, i64 } tensor_type;
#else
typedef enum { fp16 = 0, fp32, up8, ip32 } tensor_type;
#endif 
```

I assume people only tend to chase the latest wasi-nn spec when using `-DWASM_ENABLE_WASI_EPHEMERAL_NN=1`

tonibofarull: Makes sense to me! Can we add this decision in the README? :bow:

lum1n0us: Sure.

wenyongh (commented):
LGTM

PR #3535, opened 2024-06-17
validate func type before calculation

lum1n0us:


core/iwasm/aot/aot_loader.c
```diff
@@ -1684,6 +1684,11 @@ load_types(const uint8 **p_buf, const uint8 *buf_end, AOTModule *module,
 
             func_type->ref_type_map_count = ref_type_map_count;
 
+#if WASM_ENABLE_FUZZ_TEST != 0
```
lum1n0us, line 1687: NOT sure `#if ... #endif` is necessary. But it might be helpful for potential performance concern.

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -1607,6 +1607,12 @@ resolve_func_type(const uint8 **p_buf, const uint8 *buf_end, WASMModule *module,
     wasm_dump_func_type(type);
 #endif
 
+    if (!wasm_func_type_validation(type)) {
```
wenyongh, line 1610: For interpreter mode, the param/result type of func_type has been checked in resolve_value_type (L1517 and L1528), had better not add extra function to check it. Had better enhance resolve_value_type if there is issue found.

core/iwasm/interpreter/wasm.h
```diff
@@ -1181,6 +1181,55 @@ wasm_string_equal(const char *s1, const char *s2)
...
+ * Whether a type is known and valid
+ */
+inline static bool
+is_valid_value_type(uint8 value_type)
+{
+    if (value_type == VALUE_TYPE_VOID)
+        return true;
+
+    if (value_type == VALUE_TYPE_I32 || value_type == VALUE_TYPE_F32
+        || value_type == VALUE_TYPE_ANY)
```
wenyongh, line 1195: VALUE_TYPE_ANY is wamr's internal type for the integrity check in wasm loader, had better not add it here?

core/iwasm/interpreter/wasm.h
```diff
@@ -1181,6 +1181,55 @@ wasm_string_equal(const char *s1, const char *s2)
...
+#if WASM_ENABLE_SIMD != 0
+    if (value_type == VALUE_TYPE_V128)
+        return true;
+#endif
+
+#if WASM_ENABLE_GC == 0 && WASM_ENABLE_REF_TYPES != 0
+    if (value_type == VALUE_TYPE_FUNCREF || value_type == VALUE_TYPE_EXTERNREF)
+        return true;
+#elif WASM_ENABLE_GC != 0
+    else if ((value_type >= (uint8)REF_TYPE_ARRAYREF               /* 0x6A */
```
wenyongh, line 1210: Seems should be `if`? `else if` may hasn't matched `if`?

core/iwasm/interpreter/wasm.h
```diff
@@ -1309,6 +1358,24 @@ wasm_type_equal(const WASMType *type1, const WASMType *type2,
                 const WASMTypePtr *types, uint32 type_count);
 #endif
 
+/* FIXME: move to aot_loader.c if only required by aot */
+inline static bool
+is_valid_func_type(WASMFuncType *func_type)
+{
+    unsigned i;
+    for (i = 0; i < func_type->param_count; i++) {
+        if (!is_valid_value_type(func_type->types[i]))
```
wenyongh, line 1367: Shall we exclude `VALUE_TYPE_VOID`, `PACKED_TYPE_I8` and `PACKED_TYPE_I16`?
param/result type should not be void. And i8/i16 is used only in struct field and array element.

lum1n0us: in that case, how about using `is_value_type()` in wasm_loader.c, and move it here?

wenyongh: Here the function is `inline static bool`, moving is_value_type here may increase footprint, had better move to wasm_runtime_common.h and wasm_runtime_common.c?

wenyongh: Move is_value_type to wasm_runtime_common.h/.c and change it to is_valid_value_type, and also add is_valid_func_type in wasm_runtime_common.h/.c?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -7384,3 +7384,48 @@ wasm_runtime_is_underlying_binary_freeable(WASMModuleCommon *const module)
...
+    if (/* I32/I64/F32/F64, 0x7C to 0x7F */
+        (type >= VALUE_TYPE_F64 && type <= VALUE_TYPE_I32)
+#if WASM_ENABLE_GC != 0
+        /* reference types, 0x65 to 0x70 */
+        || wasm_is_type_reftype(type)
+#elif WASM_ENABLE_REF_TYPES != 0
+        || (type == VALUE_TYPE_FUNCREF || type == VALUE_TYPE_EXTERNREF)
+#endif
+#if WASM_ENABLE_SIMD != 0
+#if (WASM_ENABLE_WAMR_COMPILER != 0) || (WASM_ENABLE_JIT != 0)
```
wenyongh, line 7407: SIMD type can be enabled in AOT mode, had better add `|| (WASM_ENABLE_AOT != 0)`?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -7384,3 +7384,48 @@ wasm_runtime_is_underlying_binary_freeable(WASMModuleCommon *const module)
...
+bool
+is_valid_func_type(const WASMFuncType *func_type)
+{
+    unsigned i;
+    for (i = 0; i < func_type->param_count; i++) {
+        if (!is_valid_value_type(func_type->types[i]))
+            return false;
+    }
+
+    for (i = 0; i < func_type->result_count; i++) {
```
wenyongh, line 7425: Had better `for (i = 0; i < func_type->param_count + func_type->result_count; i++) { ... }`?

core/iwasm/common/wasm_runtime_common.h
```diff
@@ -1210,6 +1210,12 @@ void
 wasm_runtime_set_linux_perf(bool flag);
 #endif
 
+bool
+is_valid_value_type(uint8 value_tpye);
+
+bool
+is_valid_func_type(const WASMFuncType *func_type);
```
wenyongh, line 1217: Just found that there are `wasm_loader_common.h` and `wasm_loader_common.c` under core/iwasm/common, do you think it is better to put these two functions there?

core/iwasm/interpreter/wasm_loader.c
wenyongh, line 1: Should also modify wasm_mini_loader.c

core/iwasm/common/wasm_runtime_common.c
lum1n0us: Sure. no need to separate params and results.

core/iwasm/common/wasm_runtime_common.h
lum1n0us: looks better.

wenyongh (approved):
LGTM

PR #3538, opened 2024-06-17
prevent issues from multiple native symbol sections

lum1n0us:
- 69633.


yamt:
> - 69633.

i suspect it's better to have a bit less cryptic description.

core/iwasm/aot/aot_loader.c
```diff
@@ -651,6 +651,12 @@ load_native_symbol_section(const uint8 *buf, const uint8 *buf_end,
     int32 i;
     const char *symbol;
 
+    if (module->native_symbol_list) {
+        set_error_buf(error_buf, error_buf_size,
+                      "duplicate native symbol section");
+        return false;
+    }
```
wenyongh, line 658: The `module->native_symbol_list` is created below (L663), here it should be NULL, seems the fixing is invalid?

lum1n0us: There is a case with two such native symbol sections(don't know how fuzzer figures it out). So ...

wenyongh: Got it, thanks. Could you change `duplicate native symbol section` to `duplicated native symbol section`?

wenyongh (approved):

PR #3539, opened 2024-06-18
Allow loading of missing imports without errors when multi-module is enabled

XeniaLu:
The INTERP module loader was failing when multi-module support is on
and dependant modules are not found; this enforced AOT compiler
integrations to prepare dependant modules while it wasn't necessary.

~~This new option allows integrated AOT compilers to work as if the
multi-module support wasn't turned on.~~

~~This option is also turned on in the AOT compiler CLI, just in case
if `WASM_ENABLE_MULTI_MODULE` is turned on there one day.~~

* WAMRC can generate AOT files even if the main module has 
dependencies that are not found.
* Expanded `check_linked_symbol()` to include checks for memories,
tables and tags during the instantiation phase to ensure error raises


lum1n0us:
So, you're hoping a multi-aot feature which means linking happens when aot loader loading files. But current status is WAMRC requires all dependencies. I agree that your requirement is more nature. And it should be the default behaviour. Not an option.

XeniaLu:
> So, you're hoping a multi-aot feature which means linking happens when aot loader loading files. But current status is WAMRC requires all dependencies. I agree that your requirement is more nature. And even more it should be the default behaviour. Not an option.

I have removed the option as you suggested and made it the default behavior.

> Plus, Since I am not quite familiar with current implementation, may I ask why WAMRC needs to enable "multi-module" feature? In my imagination, people should use wamrc(w/o multi-module feature) to generate AOTs one by one for main module and its dependencies. And AOT loader(w/ multi-module feature) could load them like a ld.linux.

Yes, we don't need to enable "multi-module" for WAMRC. Actually we have integrated the compiler code directly into the same package as the runtime, our compiler and runtime share the same macro control. So we have to ensure that the compiler does not encounter failures when the "multi-module" feature is enabled.

lum1n0us:
> Yes, we don't need to enable "multi-module" for WAMRC

If so, original wasm_loader will not cause any trouble in your project. The question will be, why change? Because all modification disable original *show error and quit* action when couldn't matching import requirements.

BTW, you might want to change the PR title. Now the PR is about something else.

lum1n0us:
thanks for @wenyongh. I kindly understand the situation here. The project needs to `-DWAMR_BUILD_MULTI_MODULE=1` when compiling WAMRC but actually don't want *show error and quit*, which is in `#if WASM_ENABLE_MULTI_MODULE !=0 ... #endif` block, to bother WAMRC compilation. Please correct me if I am wrong.

I guess it becomes another problem: under other non-aot running mode, like interpreter and jit, when enabling `MULTI_MODULE` feature, what kind of behavior the loader should be if can't match any import requirement?

There is be an instantiation check `check_linked_symbol()` in *wasm_runtime.c*. It could be our insurance. But it only cover function and global. Please add check for memory, table and tag.  `#if WASM_ENABLE_WAMR_COMPILER == 0` will protect WAMRC from *linking failures*.

XeniaLu:
> thanks for @wenyongh. I kindly understand the situation here. The project needs to `-DWAMR_BUILD_MULTI_MODULE=1` when compiling WAMRC but actually don't want _show error and quit_, which is in `#if WASM_ENABLE_MULTI_MODULE !=0 ... #endif` block, to bother WAMRC compilation. Please correct me if I am wrong.

In our project, we don't directly use WAMRC, instead we integrated the compiler code into the runtime. And yes, we need `-DWAMR_BUILD_MULTI_MODULE=1` and we perfer not to show errors and quit when dependent modules are not found.

> I guess it becomes another problem: under other non-aot running mode, like interpreter and jit, when enabling `MULTI_MODULE` feature, what kind of behavior the loader should be if can't match any import requirement?

So I think using the `LoadArgs.allow_missing_imports` option is the most appropriate in this case. Without this option, there might be issues in non-aot mode.

yamt:
> > thanks for @wenyongh. I kindly understand the situation here. The project needs to `-DWAMR_BUILD_MULTI_MODULE=1` when compiling WAMRC but actually don't want _show error and quit_, which is in `#if WASM_ENABLE_MULTI_MODULE !=0 ... #endif` block, to bother WAMRC compilation. Please correct me if I am wrong.
> 
> In our project, we don't directly use WAMRC, instead we integrated the compiler code into the runtime. And yes, we need `-DWAMR_BUILD_MULTI_MODULE=1` and we perfer not to show errors and quit when dependent modules are not found.

i suspect such a usage is simply unsupported.
i'm not sure if it should be supported either.
is this the only change you need for your "integrated compiler" use case?

XeniaLu:
> > > thanks for @wenyongh. I kindly understand the situation here. The project needs to `-DWAMR_BUILD_MULTI_MODULE=1` when compiling WAMRC but actually don't want _show error and quit_, which is in `#if WASM_ENABLE_MULTI_MODULE !=0 ... #endif` block, to bother WAMRC compilation. Please correct me if I am wrong.
> > 
> > 
> > In our project, we don't directly use WAMRC, instead we integrated the compiler code into the runtime. And yes, we need `-DWAMR_BUILD_MULTI_MODULE=1` and we perfer not to show errors and quit when dependent modules are not found.
> 
> i suspect such a usage is simply unsupported. i'm not sure if it should be supported either. is this the only change you need for your "integrated compiler" use case?

Yes, the [commit](https://github.com/bytecodealliance/wasm-micro-runtime/pull/3539/commits/4164d049d0d92c769f8bfc3906068f8dffba1680) with the `LoadArgs.allow_missing_imports` option is the only change addresses our immediate needs.

yamt:
> IIUC, for some reason, in your project, WAMRC and IWASM are compiled in a same set of compilation options. It is different with what we did in wamr repo. (Maybe both executables share one library of WAMR?)

i'm interested in why.

if we want to support such a configuration, i guess we need a sample to run on the ci.

XeniaLu:
> Maybe both executables share one library of WAMR?

We are developing a Postgres extension that statically links the WAMR compiler and IWASM code together. Our extension provides a PG function for AOT compilation and also includes a runtime for execution.

If you are interested, I can add you to our Gitee repository. It is currently a pre-open source project.

lum1n0us:
>  statically links the WAMR compiler and IWASM code together.

Never saw that coming. 🤔. 

Even so, we can still use `check_linked_symbol()` to do instanation phase check to make sure every necessary symbols are all set before execution.

XeniaLu:
@lum1n0us Thanks, I've commit the changes for `check_linked_symbol()`. However, the tests are currently failing. I'm investigating on it, and any help you can provide would be greatly appreciated.

wenyongh:
> @lum1n0us Thanks, I've commit the changes for `check_linked_symbol()`. However, the tests are currently failing. I'm investigating on it, and any help you can provide would be greatly appreciated.

@XeniaLu I did some test on this and created a patch based on your PR:
https://github.com/wenyongh/wasm-micro-runtime/pull/932
It basically works, which reports linking error in wasm instantiation when the import global/table/memory/tag isn't builtin and isn't linked in multi-module mode. I removed unnecessary WASM_ENABLE_WAMRC_COMPILER macro controls in wasm_runtime.c since the aot compiler doesn't run into the file, and also modified several patch files for spec tests since the check is a little stricter than before. Maybe you can check whether it fits your requirement?

XeniaLu:
> @XeniaLu I did some test on this and created a patch based on your PR: [wenyongh#932](https://github.com/wenyongh/wasm-micro-runtime/pull/932) It basically works, which reports linking error in wasm instantiation when the import global/table/memory/tag isn't builtin and isn't linked in multi-module mode. I removed unnecessary WASM_ENABLE_WAMRC_COMPILER macro controls in wasm_runtime.c since the aot compiler doesn't run into the file, and also modified several patch files for spec tests since the check is a little stricter than before. Maybe you can check whether it fits your requirement?

@wenyongh Super thanks for the patch. It works perfectly for our needs. I have cherry-picked your patch into this PR.

wenyongh (approved):
LGTM

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1267,6 +1267,9 @@ wasm_runtime_is_built_in_module(const char *module_name)
             || !strcmp("wasi_snapshot_preview1", module_name)
 #if WASM_ENABLE_SPEC_TEST != 0
             || !strcmp("spectest", module_name)
+#endif
+#if WASM_ENABLE_LIB_WASI_THREADS != 0
+            || !strcmp("foo", module_name)
```
lum1n0us, line 1272: wondering why explicit "foo" here? If it came from a test case, better be surrounded by a `#if...#endif`.

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -2920,6 +2904,29 @@ load_memory_import(const uint8 **p_buf, const uint8 *buf_end,
         declare_init_page_count = spectest_memory_init_page;
         declare_max_page_count = spectest_memory_max_page;
     }
+#if WASM_ENABLE_LIB_WASI_THREADS != 0
+    /* a case in wasi-testsuite which imports ("foo" "bar") */
```
lum1n0us, line 2908: if so(to make test cases happy), we'd better make a new preprocessor conditionals. like `WASM_ENABLE_WASI_TEST`.

@wenyongh we need a new marco here to control if in some wasi-testsuites

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -1710,36 +1710,74 @@ check_linked_symbol(WASMModuleInstance *module_inst, char *error_buf,
...
-#endif /* WASM_ENABLE_WAMR_COMPILER == 0 */
         }
     }
 
     for (i = 0; i < module->import_global_count; i++) {
         WASMGlobalImport *global = &((module->import_globals + i)->u.global);
-        if (!global->is_linked) {
+
+        if (!global->is_linked
+            && !wasm_runtime_is_built_in_module(global->module_name)) {
```
lum1n0us, line 1722: wondering why need to check `wasm_runtime_is_built_in_module()` here. IIUC, if requires a built_in module, it should be satisfied in *wasm_loader.c*.

core/iwasm/common/wasm_runtime_common.c
wenyongh: Yes, maybe we can add a cmake variable `cmake -DWAMR_BUILD_WASI_TEST=1/0` and add gcc macro `WASM_ENABLE_WASI_TEST`, like WAMR_BUILD_SPEC_TEST. And enable it when testing wasi_certification in test_wamr.sh (just a guess, to be confirmed):
```diff
diff --git a/tests/wamr-test-suites/test_wamr.sh b/tests/wamr-test-suites/test_wamr.sh
index 135940e1..bbfd5259 100755
--- a/tests/wamr-test-suites/test_wamr.sh
+++ b/tests/wamr-test-suites/test_wamr.sh
@@ -1003,6 +1003,9 @@ function trigger()
                 EXTRA_COMPILE_FLAGS+=" -DWAMR_BUILD_LIBC_UVWASI=0 -DWAMR_BUILD_LIBC_WASI=1"
                 break
             fi
+            if [[ "$test" == "wasi_certification" ]]
+                EXTRA_COMPILE_FLAGS+=" -DWAMR_BUILD_WASI_TEST=1"
+            fi
```

core/iwasm/interpreter/wasm_runtime.c
wenyongh: I am not sure, I just refactored the code based on @XeniaLu's original modification. Maybe @XeniaLu can help remove this line to have a try?

XeniaLu: Sure, I will remove it.

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -1710,36 +1710,73 @@ check_linked_symbol(WASMModuleInstance *module_inst, char *error_buf,
...
-            /* do nothing to avoid confused message */
-#endif /* WASM_ENABLE_WAMR_COMPILER == 0 */
 #endif /* WASM_ENABLE_SPEC_TEST != 0 */
         }
     }
 
+    for (i = 0; i < module->import_table_count; i++) {
+        WASMTableImport *table = &((module->import_tables + i)->u.table);
+
+        if (!wasm_runtime_is_built_in_module(table->module_name)
```
lum1n0us, line 1738: please reconsider whether check `is_built_in` on table too.

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -1710,36 +1710,73 @@ check_linked_symbol(WASMModuleInstance *module_inst, char *error_buf,
...
+                            "failed to link import table (%s, %s)",
+                            table->module_name, table->field_name);
+            return false;
+        }
+    }
+
+    for (i = 0; i < module->import_memory_count; i++) {
+        WASMMemoryImport *memory = &((module->import_memories + i)->u.memory);
+
+        if (!wasm_runtime_is_built_in_module(memory->module_name)
```
lum1n0us, line 1753: please reconsider whether check `is_built_in` on memory too.

core/iwasm/common/wasm_runtime_common.c
XeniaLu: I have added the cmake variable `WAMR_BUILD_WASI_TEST` and the corresponding GCC macro `WASM_ENABLE_WASI_TEST`. Since I'm not very familiar with this part, please let me know if there's anything missing or if further adjustments are needed.

And I have no idea about the failing tests. Could you please help me take a look?

core/iwasm/interpreter/wasm_runtime.c
XeniaLu: I added the `is_built_in` check because the CI test was [failing](https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/9606653696/job/26496605530) at that time.

build-scripts/config_common.cmake
```diff
@@ -233,6 +233,10 @@ if (WAMR_BUILD_SPEC_TEST EQUAL 1)
   add_definitions (-DWASM_ENABLE_SPEC_TEST=1)
   message ("     spec test compatible mode is on")
 endif ()
+if (WAMR_BUILD_WASI_TEST EQUAL 1)
+  add_definitions (-DWASM_ENABLE_WASI_TEST=1)
```
wenyongh, line 237: Please also add lines in core/config.h, e.g. after WASM_ENABLE_SPEC_TEST:
```C
#ifndef WASM_ENABLE_WASI_TEST
#define WASM_ENABLE_WASI_TEST 0
#endif
```
This should be able to fix the nuttx CI failures.

core/iwasm/common/wasm_runtime_common.c
wenyongh: Add lines in core/config.h:
```C
#ifndef WASM_ENABLE_WASI_TEST
#define WASM_ENABLE_WASI_TEST 0
#endif
```

core/iwasm/interpreter/wasm_runtime.c
lum1n0us: 🤔 it is a good fix. 

Although we've linked (table and memory) in wasm_laoder.c for spectest only, there is no flag to store the status. Then we use `wasm_runtime_is_built_in_module()` to implement a strong assumption that all *built-in modules* requirements has been satisfied. 

we can fix it in other PR

lum1n0us (approved):
LGTM

PR #3540, opened 2024-06-18
dwarf_extractor.cpp: enable limited support for C++

yamt:
While band-aid fixes like this is not plausible IMO, some people prefer to have some debug info even if it's partial/limited/broken. this commit partially (re)enables C++ processing. on the other hand, do not bother to process variables because it's known incompatible with C++.


TianlongLiang:
LGTM

wenyongh (approved):

PR #3543, opened 2024-06-18
implementation of wasm shared heap

WenLY1:
- [X] add WAMR_BUILD_SHARED_HEAP and WASM_ENABLE_SHARED_HEAP
- [ ] add wamrc --enable-shared-heap
- [ ] consider the behavior of import memory in multi-module feature
- [ ] adapt address conversion/validation
      aot code boundary check
      wasm_runtime_invoke_native
      wasm_runtime_invoke_native_raw
     ~~wasm_runtime_validate_app_addr~~
     ~~wasm_runtime_addr_app_to_native~~
     ~~wasm_runtime_addr_native_to_app~~
     ~~wasm_runtime_validate_native_addr~~
- [ ] interpreter/fast-jit do boundary check
     ~~classic interp mode~~
     fast interp mode
     jit

- [X] Support setting shared heap’s size and shared heap’s alloc options in RuntimeInitArgs for wasm_runtime_full_init
- [ ] Add check in load module, default memory’s max_memory_size should be no larger than 4G-shared_heap_size, if not, reduce it to 4G-shared_heap_size
- [X] Add API wasm_runtime_shared_malloc/wasm_runtime_shared_free
- [X] App allocates memory from shared heap with API shared_malloc/shared_free
      example
```
            #include <stdio.h>  
            #include <stdlib.h>  
            extern void *shared_malloc(int size);  
            extern void shared_free(void *ptr);  
            int main()  
            {  
	            int *pa = (int *)malloc(4);  
	            *pa = 4;  
	            printf("pa value is %d\n", *pa);  
	            free(pa);  
	            int *ps = (int *)shared_malloc(4);  
	            *ps = 5;  
	            printf("ps value is %d\n", *ps);  
	            shared_free(ps);  
            } 
```


core/iwasm/libraries/libc-builtin/libc_builtin_wrapper.c
```diff
@@ -1060,6 +1087,10 @@ static NativeSymbol native_symbols_libc_builtin[] = {
     REG_NATIVE_FUNC(malloc, "(i)i"),
     REG_NATIVE_FUNC(realloc, "(ii)i"),
     REG_NATIVE_FUNC(calloc, "(ii)i"),
+#if WASM_ENABLE_SHARED_HEAP != 0
+    REG_NATIVE_FUNC(shared_malloc, "(i)i"),
+    REG_NATIVE_FUNC(shared_free, "(*)"),
+#endif
```
yamt, line 1093: i guess this should be a separate lib, not a part of libc_builtin.

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -2268,6 +2268,37 @@ wasm_instantiate(WASMModule *module, WASMModuleInstance *parent,
...
+        }
+
+        if (!(module_inst->shared_heap->data = runtime_malloc(
+                  (uint64)shared_heap_size, error_buf, error_buf_size))) {
+            goto fail;
+        }
+
+        module_inst->shared_heap->size = shared_heap_size;
+
+        if (!mem_allocator_create_with_struct_and_pool(
```
yamt, line 2293: doesn't this allocator put metadata on the heap itself?
if so, i guess it isn't appropriate for shared heap.

core/iwasm/libraries/libc-builtin/libc_builtin_wrapper.c
WenLY1: Got

core/iwasm/interpreter/wasm_runtime.c
WenLY1: The heap is inited by module_inst->shared_heap->data, and in future implementation, module_inst->shared_heap->data is to be shared between multi modules

wenyongh:
@WenLY1 thanks for uploading the PR, I have created branch [dev/shared_heap](https://github.com/bytecodealliance/wasm-micro-runtime/tree/dev/shared_heap) as discussed, please submit the PR to that branch instead.

core/iwasm/interpreter/wasm_runtime.c
yamt: i meant that, if metadata is stored on the shared heap itself, any modules accessible to the heap can easily corrupt the metadata.
is it ok for your use cases?

WenLY1: Because mem_allocator_create_with_struct_and_pool rather than mem_allocator_create is used to create a allocator, the metadata of allocator is separated from the heap

wenyongh:
Close this PR as another PR is under review instead: #3823.

PR #3549, opened 2024-06-20
Fix compilation errors

wenyongh:
The compilation errors were introduced by #3515 and occur in debug building
when wasm mini loader is compiled or GC is enabled.

And remove two wasm files in standalone test-running-modes case,
which will be generated by run.sh.


yamt:
i guess it's worth to have a debug build on the ci.

yamt (approved):

PR #3550, opened 2024-06-20
aot_emit_aot_file.c: Fix debug build

yamt:
After the "Add table type API support" change.


yamt:
https://github.com/bytecodealliance/wasm-micro-runtime/pull/3549 happens to have the identical changes.

PR #3551, opened 2024-06-20
core/shared/platform: Zero memory returned by os_mmap in some platforms

wenyongh:
The os_mmap should zero the mapped memory like what posix mmap does.


dongsheng28849455 (commented):
LGTM

TianlongLiang:
LGTM

PR #3552, opened 2024-06-20
aot compiler: use larger alignment for load/store when possible

yamt:
consider the folling wasm module:

```
(module
  (func (export "foo")
    i32.const 0x104
    i32.const 0x12345678
    i32.store
  )
  (memory 1 1)
)
```

while the address (0x104) is perfectly aligned for i32.store, as our aot compiler uses 1-byte alignment for load/store LLVM IR instructions, it often produces inefficient machine code, especially for alignment-sensitive targets.

for example, the above "foo" function is compiled into the following xtensa machine code.

```
0000002c <aot_func_internal#0>:
  2c:   004136          entry   a1, 32
  2f:   07a182          movi    a8, 0x107
  32:   828a            add.n   a8, a2, a8
  34:   291c            movi.n  a9, 18
  36:   004892          s8i     a9, a8, 0
  39:   06a182          movi    a8, 0x106
  3c:   828a            add.n   a8, a2, a8
  3e:   ffff91          l32r    a9, 3c <aot_func_internal#0+0x10> (ff91828a <aot_func_internal#0+0xff91825e>)
                        3e: R_XTENSA_SLOT0_OP   .literal+0x8
  41:   004892          s8i     a9, a8, 0
  44:   05a182          movi    a8, 0x105
  47:   828a            add.n   a8, a2, a8
  49:   ffff91          l32r    a9, 48 <aot_func_internal#0+0x1c> (ffff9182 <aot_func_internal#0+0xffff9156>)
                        49: R_XTENSA_SLOT0_OP   .literal+0xc
  4c:   41a890          srli    a10, a9, 8
  4f:   0048a2          s8i     a10, a8, 0
  52:   04a182          movi    a8, 0x104
  55:   828a            add.n   a8, a2, a8
  57:   004892          s8i     a9, a8, 0
  5a:   f01d            retw.n
```

note that the each four bytes are stored separately using one-byte-store instruction, s8i.

this commit tries to use larger alignments for load/store LLVM IR instructions when possible.  with this commit, the above example is compiled into the following machine code, which seems more reasonable to me.

```
0000002c <aot_func_internal#0>:
  2c:   004136          entry   a1, 32
  2f:   ffff81          l32r    a8, 2c <aot_func_internal#0> (81004136 <aot_func_internal#0+0x8100410a>)
                        2f: R_XTENSA_SLOT0_OP   .literal+0x8
  32:   416282          s32i    a8, a2, 0x104
  35:   f01d            retw.n
```

Note: this doesn't work well for --xip because
aot_load_const_from_table() hides the constness of the value. maybe we need our own mechanism to propagate the constness and the value.

### DIFF
--- a/core/iwasm/compilation/aot_emit_memory.c
+++ b/core/iwasm/compilation/aot_emit_memory.c
@@ -188,7 +188,7 @@ aot_check_memory_overflow(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
              * has the natural alignment. for platforms using mmap, it can
              * be even larger. for now, use a conservative value.
              */
-            const int max_align = 4;
+            const int max_align = 8;
             int shift = ffs((int)(unsigned int)mem_offset);
             if (shift == 0) {
                 *alignp = max_align;
--- a/core/iwasm/compilation/aot_emit_memory.c
+++ b/core/iwasm/compilation/aot_emit_memory.c
@@ -96,7 +96,8 @@ get_memory_curr_page_count(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx);
 
 LLVMValueRef
 aot_check_memory_overflow(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
-                          mem_offset_t offset, uint32 bytes, bool enable_segue)
+                          mem_offset_t offset, uint32 bytes, bool enable_segue,
+                          unsigned int *alignp)
 {
     LLVMValueRef offset_const =
         MEMORY64_COND_VALUE(I64_CONST(offset), I32_CONST(offset));
@@ -202,8 +203,31 @@ aot_check_memory_overflow(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
                     goto fail;
                 }
             }
+            if (alignp != NULL) {
+                /*
+                 * A note about max_align below:
+                 * the assumption here is the base address of a linear memory
+                 * has the natural alignment. for platforms using mmap, it can
+                 * be even larger. for now, use a conservative value.
+                 */
+                const int max_align = 4;
+                int shift = ffs((int)(unsigned int)mem_offset);
+                if (shift == 0) {
+                    *alignp = max_align;
+                }
+                else {
+                    unsigned int align = 1 << (shift - 1);
+                    if (align > max_align) {
+                        align = max_align;
+                    }
+                    *alignp = align;
+                }
+            }
             return maddr;
         }
+    }
+    if (alignp != NULL) {
+        *alignp = 1;
     }
 
     if (is_target_64bit) {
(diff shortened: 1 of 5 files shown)


core/iwasm/compilation/aot_emit_memory.c
```diff
@@ -202,9 +203,32 @@ aot_check_memory_overflow(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
                     goto fail;
                 }
             }
+            if (alignp != NULL) {
```
wenyongh, line 206: This code piece is only for the situation that `addr + offset` is pre-known and is inside the initial size of linear memory, but it is also suitable for `addr + offset` is pre-known and is inside the maximum size of linear memory, how about put these code before L184 `if (mem_offset + bytes <= mem_data_size)`?

core/iwasm/compilation/aot_emit_memory.c
```diff
@@ -202,9 +203,32 @@ aot_check_memory_overflow(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
...
+                /*
+                 * A note about max_align below:
+                 * the assumption here is the base address of a linear memory
+                 * has the natural alignment. for platforms using mmap, it can
+                 * be even larger. for now, use a conservative value.
+                 */
+                const int max_align = 4;
```
wenyongh, line 213: Currently the linear memory is allocated with os_mmap and normally the returned address is 8-byte aligned, some platforms have applied special handling to ensure that is 8-byte aligned, e.g. esp-idf and nuttx:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/shared/platform/esp-idf/espidf_memmap.c#L39-L41

https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/shared/platform/nuttx/nuttx_platform.c#L82

So is it better to set max_align as 8 here? In my memory, 8-byte aligned is an implicit requirement for memory allocation in WAMR, but we can make it explicit and add assert for os_mmap in each platform implementation?

core/iwasm/compilation/aot_emit_memory.c
```diff
@@ -202,9 +203,32 @@ aot_check_memory_overflow(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
...
+                    if (align > max_align) {
+                        align = max_align;
+                    }
+                    *alignp = align;
+                }
+            }
             return maddr;
         }
     }
+    if (alignp != NULL) {
```
wenyongh, line 229: Can change to `else if` if we handle all the situations for `if (LLVMIsEfficientConstInt(addr))`, or how about putting it before L175?

core/iwasm/compilation/aot_emit_memory.c
yamt: it makes sense.

core/iwasm/compilation/aot_emit_memory.c
yamt: ok.

core/iwasm/compilation/aot_emit_memory.c
yamt: done

core/iwasm/compilation/aot_emit_memory.c
yamt: done

wenyongh (approved):
LGTM

PR #3559, opened 2024-06-21
Add protection in `load_native_symbol_section()`

lum1n0us:
- #69675

### DIFF
--- a/core/iwasm/aot/aot_loader.c
+++ b/core/iwasm/aot/aot_loader.c
@@ -660,15 +660,18 @@ load_native_symbol_section(const uint8 *buf, const uint8 *buf_end,
     read_uint32(p, p_end, cnt);
 
     if (cnt > 0) {
-        module->native_symbol_list = wasm_runtime_malloc(cnt * sizeof(void *));
+        uint64 list_size = cnt * (uint64)sizeof(void *);
+        module->native_symbol_list =
+            loader_malloc(list_size, error_buf, error_buf_size);
         if (module->native_symbol_list == NULL) {
-            set_error_buf(error_buf, error_buf_size,
-                          "malloc native symbol list failed");
             goto fail;
         }
 
         for (i = cnt - 1; i >= 0; i--) {
             read_string(p, p_end, symbol);
+            if (!strlen(symbol))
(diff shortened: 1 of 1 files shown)


core/iwasm/aot/aot_loader.c
```diff
@@ -660,15 +660,20 @@ load_native_symbol_section(const uint8 *buf, const uint8 *buf_end,
...
     if (cnt > 0) {
-        module->native_symbol_list = wasm_runtime_malloc(cnt * sizeof(void *));
+        uint32 list_size = cnt * sizeof(void *);
+        module->native_symbol_list = wasm_runtime_malloc(list_size);
         if (module->native_symbol_list == NULL) {
             set_error_buf(error_buf, error_buf_size,
                           "malloc native symbol list failed");
             goto fail;
         }
+        memset(module->native_symbol_list, 0, list_size);
```
wenyongh, line 670: Integer overflow may occur in `cnt * sizeof(void *)` and had better use loader_malloc which also memset the memory. Had better change the code to be somewhat like:
```C
        uint64 list_size = cnt * (uint64)sizeof(void *);
        module->native_symbol_list = loader_malloc(list_size, error_buf, error_buf_size);
        if (module->native_symbol_list == NULL)
            goto fail;
```

core/iwasm/aot/aot_loader.c
```diff
@@ -660,7 +660,9 @@ load_native_symbol_section(const uint8 *buf, const uint8 *buf_end,
     read_uint32(p, p_end, cnt);
 
     if (cnt > 0) {
-        module->native_symbol_list = wasm_runtime_malloc(cnt * sizeof(void *));
+        uint64 list_size = cnt * (uint64)sizeof(void *);
+        module->native_symbol_list =
+            loader_malloc(list_size, error_buf, error_buf_size);
         if (module->native_symbol_list == NULL) {
             set_error_buf(error_buf, error_buf_size,
                           "malloc native symbol list failed");
```
wenyongh, line 668: No need to set_error_buf again since loader_malloc has done it?

wenyongh (approved):

PR #3563, opened 2024-06-23
Optimize for multi-module support in AOT mode

XeniaLu:
* split the `aot_loader_resolve_function` into two functions to prevent redundant module lookups and loads
* access pre-associated module instances from `func_module_insts`, avoiding unnecessary instances lookups and improving performance

### DIFF
--- a/core/iwasm/aot/aot_loader.c
+++ b/core/iwasm/aot/aot_loader.c
@@ -591,15 +591,17 @@ str2uint64(const char *buf, uint64 *p_res);
 
 #if WASM_ENABLE_MULTI_MODULE != 0
 static void *
-aot_loader_resolve_function(const char *module_name, const char *function_name,
+aot_loader_resolve_function(const AOTModule *module, const char *function_name,
                             const AOTFuncType *expected_function_type,
-                            char *error_buf, uint32 error_buf_size)
+                            char *error_buf, uint32 error_buf_size);
+
+static void *
+aot_loader_resolve_function_ex(const char *module_name,
+                               const char *function_name,
+                               const AOTFuncType *expected_function_type,
+                               char *error_buf, uint32 error_buf_size)
 {
     WASMModuleCommon *module_reg;
-    void *function = NULL;
-    AOTExport *export = NULL;
-    AOTModule *module = NULL;
-    AOTFuncType *target_function_type = NULL;
 
     module_reg = wasm_runtime_find_module_registered(module_name);
     if (!module_reg || module_reg->module_type != Wasm_Module_AoT) {
@@ -608,10 +610,23 @@ aot_loader_resolve_function(const char *module_name, const char *function_name,
         set_error_buf(error_buf, error_buf_size, "unknown import");
         return NULL;
     }
+    return aot_loader_resolve_function((AOTModule *)module_reg, function_name,
+                                       expected_function_type, error_buf,
+                                       error_buf_size);
+}
 
-    module = (AOTModule *)module_reg;
-    export = loader_find_export(module_reg, module_name, function_name,
-                                EXPORT_KIND_FUNC, error_buf, error_buf_size);
+static void *
+aot_loader_resolve_function(const AOTModule *module, const char *function_name,
+                            const AOTFuncType *expected_function_type,
+                            char *error_buf, uint32 error_buf_size)
+{
+    void *function = NULL;
+    AOTExport *export = NULL;
+    AOTFuncType *target_function_type = NULL;
+
+    export = loader_find_export((WASMModuleCommon *)module, module->name,
+                                function_name, EXPORT_KIND_FUNC, error_buf,
+                                error_buf_size);
     if (!export) {
         return NULL;
     }
@@ -633,7 +648,7 @@ aot_loader_resolve_function(const char *module_name, const char *function_name,
     if (!wasm_type_equal((WASMType *)expected_function_type,
                          (WASMType *)target_function_type, module->types,
                          module->type_count)) {
-        LOG_DEBUG("%s.%s failed the type check", module_name, function_name);
+        LOG_DEBUG("%s.%s failed the type check", module->name, function_name);
         set_error_buf(error_buf, error_buf_size, "incompatible import type");
         return NULL;
     }
@@ -2262,9 +2277,14 @@ load_import_funcs(const uint8 **p_buf, const uint8 *buf_end, AOTModule *module,
                     return false;
                 }
             }
-            linked_func = aot_loader_resolve_function(
-                module_name, field_name, declare_func_type, error_buf,
-                error_buf_size);
+            if (!sub_module)
+                linked_func = aot_loader_resolve_function_ex(
+                    module_name, field_name, declare_func_type, error_buf,
+                    error_buf_size);
+            else
+                linked_func = aot_loader_resolve_function(
+                    sub_module, field_name, declare_func_type, error_buf,
+                    error_buf_size);
         }
         import_funcs[i].func_ptr_linked = linked_func;
         import_funcs[i].func_type = declare_func_type;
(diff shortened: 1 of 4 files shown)


core/iwasm/aot/aot_loader.c
```diff
@@ -2262,9 +2277,14 @@ load_import_funcs(const uint8 **p_buf, const uint8 *buf_end, AOTModule *module,
                     return false;
```
lum1n0us, line 2277: better output something to indicate the error.

core/iwasm/aot/aot_loader.c
```diff
@@ -2262,9 +2277,14 @@ load_import_funcs(const uint8 **p_buf, const uint8 *buf_end, AOTModule *module,
...
                 }
             }
-            linked_func = aot_loader_resolve_function(
-                module_name, field_name, declare_func_type, error_buf,
-                error_buf_size);
+            if (!sub_module)
```
lum1n0us, line 2280: IMU, there is a case:

Let's say there is 2 imports.
- 1st import is not in a built-in. `sub_module` will be assigned in L2273.
- 2st import is a built-in. It will pass `if` block in L2272~L2279. Because `sub_module` has been assigned before, it goes to L2285 to search imported in `sub_module` when it should be found in as a built-in.

may `sub_module=NULL` after using.

core/iwasm/aot/aot_runtime.h
```diff
@@ -109,6 +109,7 @@ typedef struct AOTModuleInstanceExtra {
 #if WASM_ENABLE_MULTI_MODULE != 0
     bh_list sub_module_inst_list_head;
     bh_list *sub_module_inst_list;
+    WASMModuleInstanceCommon **func_module_insts;
```
lum1n0us, line 112: IMU, `sub_module_inst_list` is a list to store all `AOTSubModInstNode`. Why need a new `WASMModuleInstanceCommon[]` to do the same thing?

XeniaLu: You are right, `sub_module_inst_list` and `func_module_insts` store the same data. But `func_module_insts` is indexed by the function index, which allows us to directly access the module instance without having to go through `sub_module_inst_list` and lookup by name each time. `sub_module_inst_list` is kept because there are still some places that require a complete and non-duplicated sub-module instances list.

core/iwasm/aot/aot_loader.c
XeniaLu: Yes, it's a bug, I added `sub_module=NULL`.

core/iwasm/aot/aot_loader.c
XeniaLu: added `LOG_ERROR` here.

core/iwasm/aot/aot_runtime.h
wenyongh, line 112: How about renaming to `import_func_module_insts` since we should only need to create module instance for import functions?

core/iwasm/aot/aot_runtime.c
```diff
@@ -1152,29 +1152,63 @@ init_func_ptrs(AOTModuleInstance *module_inst, AOTModule *module,
...
+    const char *sub_inst_name = NULL;
+#endif
 
     if (module->import_func_count + module->func_count == 0)
         return true;
 
-    /* Allocate memory */
+/* Allocate memory */
+#if WASM_ENABLE_MULTI_MODULE != 0
+    if (!(ptr = runtime_malloc(total_size * 2, error_buf, error_buf_size))) {
```
wenyongh, line 1171: Allocating memory for func_ptrs and func_module_insts together makes code a little confusing, suggest to allocate individual memory for func_module_insts and don't forget to free it when destroying aot instance.

And no need to allocate memory with `import_func_count + func_count`, allocating instances for import functions should be enough.

core/iwasm/aot/aot_runtime.c
```diff
@@ -1152,29 +1152,63 @@ init_func_ptrs(AOTModuleInstance *module_inst, AOTModule *module,
...
+        sub_module_list_node = bh_list_first_elem(sub_module_list_node);
+        while (sub_module_list_node) {
+            sub_inst_name =
+                ((AOTSubModInstNode *)sub_module_list_node)->module_name;
+            if (strcmp(sub_inst_name, import_func->module_name) == 0) {
+                extra->func_module_insts[i] =
+                    (WASMModuleInstanceCommon *)((AOTSubModInstNode *)
+                                                     sub_module_list_node)
+                        ->module_inst;
+                break;
```
wenyongh, line 1207: Maybe we can set some elements of `extra->func_module_insts[]` when initializing sub_module_list_node? So no need to traverse the list here.
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/common/wasm_runtime_common.c#L7324-L7329

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -7322,6 +7322,15 @@ wasm_runtime_sub_module_instantiate(WASMModuleCommon *module,
             (WASMModuleInstance *)sub_module_inst;
         sub_module_inst_list_node->module_name =
             sub_module_list_node->module_name;
+
+#if WASM_ENABLE_AOT != 0
+        if (!init_import_func_module_insts(
```
wenyongh, line 7327: There may be many sub modules, and the function may be called multiple times and import_func_module_insts will be allocated many times?

XeniaLu: Indeed, I think it would be better to revert to the previous implementation which initializes the `import_func_module_insts` in function `init_func_ptrs`. In this way, we wouldn't need to worry about repeated calls and multiple memory allocations.  What do you think?

wenyongh: Maybe you can create array import_func_module_insts before wasm_runtime_sub_module_instantiate:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/867dbd8912f8fd489ea9b24497abb9d3c81f1b40/core/iwasm/aot/aot_runtime.c#L1641-L1643

and then initialize the array elements in wasm_runtime_sub_module_instantiate? If it cannot improve the performance, initializing the array after wasm_runtime_sub_module_instantiate is also good.

core/iwasm/aot/aot_runtime.c
```diff
@@ -1640,6 +1640,17 @@ aot_instantiate(AOTModule *module, AOTModuleInstance *parent,
...
 #if WASM_ENABLE_MULTI_MODULE != 0
     extra->sub_module_inst_list = &extra->sub_module_inst_list_head;
+
+    /* Allocate memory for import_func_module_insts*/
+    if (module->import_func_count > 0
+        && !(extra->import_func_module_insts =
+                 runtime_malloc((uint64)module->import_func_count
+                                    * sizeof(WASMModuleInstanceCommon *),
+                                error_buf, error_buf_size))) {
+        wasm_runtime_free(extra->import_func_module_insts);
```
wenyongh, line 1650: No need to free here

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -7322,6 +7322,29 @@ wasm_runtime_sub_module_instantiate(WASMModuleCommon *module,
             (WASMModuleInstance *)sub_module_inst;
         sub_module_inst_list_node->module_name =
             sub_module_list_node->module_name;
+
+#if WASM_ENABLE_AOT != 0
```
wenyongh, line 7326: Should check whether module_inst is aot instance for below codes, like L7283:
```
#if WASM_ENABLE_AOT != 0
        ...
    }
#endif

core/iwasm/aot/aot_runtime.c
XeniaLu: Ah yes, it was added by mistake.

lum1n0us (approved):
LGTM

PR #3564, opened 2024-06-24
aot_resolve_target_info: avoid in-place modification of e_type

yamt:
* I believe that LLVM MemoryBuffer interface is supposed to be read-only and it's allowed to use eg. read-only mmap of the underlying file. It isn't appropriate to modify the view at all.

* in case of WASM_ENABLE_DEBUG_AOT, the whole buffer is written as the text section of the aot file. the modified e_type would confuse dwarf consumers. note that, even when we are using XIP, the debug info usually contains relocations. for example, llvm-dwarfdump doesn't seem to perform relocations on .debug_info section for ET_CORE (== 4 == our E_TYPE_XIP) objects.

### DIFF
--- a/core/iwasm/compilation/aot_emit_aot_file.c
+++ b/core/iwasm/compilation/aot_emit_aot_file.c
@@ -3038,14 +3038,17 @@ typedef struct elf64_rela {
     elf64_sxword r_addend;
 } elf64_rela;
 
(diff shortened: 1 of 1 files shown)


wenyongh (approved):

TianlongLiang:
LGTM

PR #3567, opened 2024-06-24
aot compiler: propagate const-ness by ourselves

yamt:
aot_load_const_from_table() hides the const-ness of the value and prevents optimizations like
https://github.com/bytecodealliance/wasm-micro-runtime/pull/3552.

this commit makes the aot compiler tracks the const-ness of the value directly in the AOTValue and enables the above mentioned optimization for XIP.


yamt:
i have no idea about the CI failure.

wenyongh:
> i have no idea about the CI failure.

Seems like Github issue, I re-ran it.

wenyongh (approved):
LGTM

TianlongLiang:
LGTM

PR #3569, opened 2024-06-25
Add support for memory growth

bnason-nf:


core/iwasm/common/wasm_memory.c
```diff
@@ -912,7 +912,27 @@ wasm_enlarge_memory_internal(WASMModuleInstance *module, uint32 inc_page_count)
 
     return ret;
 }
+bool
```
wenyongh, line 915: Had better add an empty line before this line

core/iwasm/common/wasm_memory.c
```diff
@@ -912,7 +912,27 @@ wasm_enlarge_memory_internal(WASMModuleInstance *module, uint32 inc_page_count)
...
 
+#if WASM_ENABLE_INTERP != 0
+    if (module_inst->module_type == Wasm_Module_Bytecode) {
+        return wasm_enlarge_memory((WASMModuleInstance *)module_inst,
+                                   inc_page_count);
+    }
+#endif
+
+    return false;
+}
```
wenyongh, line 935: Had better add an empty line after this line

core/iwasm/common/wasm_memory.c
```diff
@@ -912,7 +912,27 @@ wasm_enlarge_memory_internal(WASMModuleInstance *module, uint32 inc_page_count)
...
+bool
+wasm_runtime_enlarge_memory(WASMModuleInstanceCommon *module_inst,
+                            uint32_t inc_page_count)
+{
+#if WASM_ENABLE_AOT != 0
+    if (module_inst->module_type == Wasm_Module_AoT) {
+        return aot_enlarge_memory((AOTModuleInstance *)module_inst,
+                                  inc_page_count);
+    }
+
```
wenyongh, line 924: Had better remove this empty line

lum1n0us (approved):
LGTM

core/iwasm/include/wasm_export.h
```diff
@@ -1842,6 +1842,10 @@ WASM_RUNTIME_API_EXTERN bool
 wasm_runtime_is_import_global_linked(const char *module_name,
                                      const char *global_name);
 
+WASM_RUNTIME_API_EXTERN bool
+wasm_runtime_enlarge_memory(wasm_module_inst_t module_inst,
+                            uint32_t inc_page_count);
```
yamt, line 1847: i guess it's better to use uint64_t for memory64.

wenyongh: In the current design of memory64, the max page count is UINT32_MAX since the max size of linear memory size is 274TB roughly and should be able to most of the requirements:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/65bf04e98643f9cbccd6cdc5b8e6d82029c52c38/core/iwasm/interpreter/wasm.h#L97-L108

And the type of page count in WASMMemoryInstance is kept uint32:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/65bf04e98643f9cbccd6cdc5b8e6d82029c52c38/core/iwasm/interpreter/wasm_runtime.h#L116-L118
So here using uint32 should be also OK.

yamt: whatever limitations the current implementation has, it's better to make the user-facing api align with the spec and future-proof, IMO.

wenyongh: OK, @bnason-nf, could you upload another PR to change `uint32_t` to `uint64_t`? Thanks.

bnason-nf: Submitted #3573.

As mentioned by @wenyongh, there are other existing 32 bit memory limitations. Aside from the many `uint32` types used internally for memory sizes, including by the direct `aot_enlarge_memory()` and `wasm_enlarge_memory()` calls by this new function, there are also many places in the public API that use `uint32_t` for memory sizes. So as I'm sure everyone is aware, PR #3573 is just a tiny piece of a much larger effort that will need to take place to make WAMR truly 64 bit compatible.

Issue #1353, opened 2022-08-03
import/export issue


yamt:
i wrote a test case to experiment memory import. (see below)

my intention is to share a single "mem" instance. i believe it's the spec-wise correct behavior.
it works as i expected for other engines. (i tested with the reference interpreter, wabt, wasmer, wasmtime, toywasm)

unfortunately, it didn't work for wamr. it seems wamr creates two instances for "mem".
is it an intended divergence from the spec? i can understand both behaviors can have valid use cases.

while i can "fix" sub_module_instantiate to use shared instances for dependencies,
i feel it actually needs a bigger surgery to fix properly.
IMO, import/export processing should be done at instance level, not module level.
how do you think?

```
(module (memory (export "m1") 1))

(register "mem")

(module
(memory (import "mem" "m1") 0)
(func (export "inc") (param $i i32) (i32.store (local.get $i) (i32.add (i32.const 1) (i32.load (local.get $i)))))
)

(register "sub")

(module
(memory (import "mem" "m1") 0)
(func (export "inc") (import "sub" "inc") (param $i i32))
(func (export "get") (param $i i32) (result i32) (i32.load (local.get $i)))
)

(assert_return (invoke "get" (i32.const 0)) (i32.const 0))
(assert_return (invoke "inc" (i32.const 0)))
(assert_return (invoke "get" (i32.const 0)) (i32.const 1))
(assert_return (invoke "inc" (i32.const 0)))
(assert_return (invoke "get" (i32.const 0)) (i32.const 2))
```

lum1n0us:
YES, it is a design choice about whehter one sub-module should have multiple instances if it is imported multiple times. 

In above case,  *module sum* and *module main*(I am going to call the third module "main") create their own *module mem* instance and create two linear memories. So both functions `inc()` and `get()` actually operate differnt linear memories.

![image](https://user-images.githubusercontent.com/62137521/182846077-a88faa54-26d3-4283-b8db-2d8f05694873.png)

`inc()` writes the No.2 linear memory via `sub.inc()` while `get()` reads the No.1 linear memory.

The most reason of this choice is we want keep a private memory for every module. I guess there are some stories how an unsafe 3rd library successl peek user data inlight us.

[...]

PR #3570, opened 2024-06-25
Add support for multi-memory proposal in classic interpreter

TianlongLiang:
Implement multi-memory for classic-interpreter. Support core spec (and bulk memory) opcodes now, will support atomic opcodes, and add multi-memory export APIs in the future. 
PS: Multi-memory spec test patched a lot for linking test to adapt for multi-module implementation


yamt:
>  PS: Multi-memory spec test patched a lot for linking test to adapt for multi-module implementation

do you mean this kind of incompatibilities? https://github.com/bytecodealliance/wasm-micro-runtime/issues/1353

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -691,6 +697,37 @@ wasm_interp_get_frame_ref(WASMInterpFrame *frame)
 #define read_leb_mem_offset(p, p_end, res) read_leb_uint32(p, p_end, res)
 #endif
 
+#if WASM_ENABLE_MULTI_MEMORY != 0
+/* First read the alignment, then if it has flag indicating following memidx,
+ * read it and set the memory instance to memories[memidx] for following memory
+ * access and boundary check, if it doesn't have flag reset the memory instance
+ * to the default memories[0] */
```
yamt, line 704: it's a bit surprising api to me.
isn't it simpler to introduce macros like read_memarg/skip_memarg?

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -35,15 +35,21 @@ typedef float64 CellType_F64;
...
  * Load memory data size in each time boundary check in
  * multi-threading mode since it may be changed by other
  * threads in memory.grow
  */
 #define get_linear_mem_size() GET_LINEAR_MEMORY_SIZE(memory)
+#elif WASM_ENABLE_MULTI_MEMORY != 0
+/**
+ * Load memory data size in each time boundary check in
+ * multi-memory mode since the memory instance(memories[i]) may be different
+ */
```
yamt, line 49: alternatively, you can update `linear_mem_size` (and `is_memory64` in future) when updating `memory`.

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -4246,13 +4286,12 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                 uint32 flags;
                 mem_offset_t offset, addr;
 
-                read_leb_uint32(frame_ip, frame_ip_end, flags);
+                read_leb_align(frame_ip, frame_ip_end, flags);
                 read_leb_mem_offset(frame_ip, frame_ip_end, offset);
                 addr = POP_MEM_OFFSET();
                 CHECK_MEMORY_OVERFLOW(4);
                 PUSH_I32(LOAD_I32(maddr));
                 CHECK_READ_WATCHPOINT(addr, offset);
```
yamt, line 4294: maybe you need to adapt watchpoint.
i suspect you might need to adapt lldb and its protocol as well.

in the mean time i guess you can make it perform watchpoint processing only when memidx == 0.

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -370,6 +370,89 @@ memory_instantiate(WASMModuleInstance *module_inst, WASMModuleInstance *parent,
     return NULL;
 }
 
+/* Only default memory have host managed heap */
```
yamt, line 373: it's better to avoid duplicating code.

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -426,12 +509,29 @@ memories_instantiate(const WASMModule *module, WASMModuleInstance *module_inst,
...
         {
-            if (!(memories[mem_index] = memory_instantiate(
-                      module_inst, parent, memory, mem_index,
-                      num_bytes_per_page, init_page_count, max_page_count,
-                      actual_heap_size, flags, error_buf, error_buf_size))) {
-                memories_deinstantiate(module_inst, memories, memory_count);
-                return NULL;
+#if WASM_ENABLE_MULTI_MEMORY != 0
+            /* Current assumption is if there doesn't exist a non-import memory,
+             * the first non-import memory would be default memory */
```
yamt, line 514: i failed to understand this comment.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -181,15 +181,40 @@ static RunningMode runtime_running_mode = Mode_Default;
...
+        if (memory_inst) {
+            mapped_mem_start_addr = memory_inst->memory_data;
+            mapped_mem_end_addr = memory_inst->memory_data + 8 * (uint64)BH_GB;
+        }
+        if (memory_inst
+            && (mapped_mem_start_addr <= (uint8 *)sig_addr
+                && (uint8 *)sig_addr < mapped_mem_end_addr)) {
+            /* The address which causes segmentation fault is inside
+               the memory instance's guard regions */
+            wasm_set_exception(module_inst, "out of bounds memory access");
```
yamt, line 204: it's a bit surprising me that a function named `is_xxx` performs destructive operations like this.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -181,15 +181,40 @@ static RunningMode runtime_running_mode = Mode_Default;
...
+{
+    WASMMemoryInstance *memory_inst;
+    uint8 *mapped_mem_start_addr = NULL;
+    uint8 *mapped_mem_end_addr = NULL;
+    uint32 i;
+
+    for (i = 0; i < module_inst->memory_count; ++i) {
+        /* To be compatible with multi memory, get the ith memory instance */
+        memory_inst = wasm_get_memory_i(module_inst, i);
+        if (memory_inst) {
```
yamt, line 195: when can this be NULL here?

TianlongLiang:
> > PS: Multi-memory spec test patched a lot for linking test to adapt for multi-module implementation
> 
> do you mean this kind of incompatibilities? #1353

Yes

core/iwasm/common/wasm_runtime_common.c
TianlongLiang: How about changing the function name to `check_set_sig_addr_oob`

core/iwasm/interpreter/wasm_interp_classic.c
TianlongLiang: Indeed there are more work needs to be done to support multi-memory interp debugging. So currently in the config_common.cmake I enforce to disable the interp debugging feature when enabling multi-memory.

core/iwasm/interpreter/wasm_interp_classic.c
TianlongLiang: good idea, I will try that

core/iwasm/interpreter/wasm_interp_classic.c
TianlongLiang: `read_memarg` does sound better than `read_leb_align` , but still need `read_leb_memidx` for some bulk memory opcodes

core/iwasm/interpreter/wasm_runtime.c
TianlongLiang: So I am assuming that if there are multiple wasm linear memories, there will be one "main(default)" memory. Some data members like heap_base, data_end, and host-managed heap are only available in default memory. Other memories don't have such data members, that's why I have an extra function that contains some duplicate code to handle different cases.

core/iwasm/common/wasm_runtime_common.c
TianlongLiang: I am not sure, I see that the original code handles that condition so I just leave it there. I will double check whether it's necessary

core/iwasm/common/wasm_runtime_common.c
yamt: it sounds fine to me.

or, move wasm_set_exception back to the caller.

core/iwasm/common/wasm_runtime_common.c
yamt: the original code needed to handle the "no memory" case. this loop doesn't need it.

core/iwasm/interpreter/wasm_runtime.c
yamt: "the default memory" makes me think the memory with index==0, regardless of import/non-import.
and i guess it's where we should put the app heap. (well, at least until we decide how to fix https://github.com/bytecodealliance/wasm-micro-runtime/issues/1353)

> that's why I have an extra function that contains some duplicate code to handle different cases.

i prefer to have `bool try_to_insert_app_heap` or something along the line over having two very similar functions like these.

TianlongLiang: so how about using the term "main memory"

yamt: why do you need such a concept?
the app heap should be in the default memory, shouldn't it?

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -697,6 +697,44 @@ wasm_interp_get_frame_ref(WASMInterpFrame *frame)
...
+ * read and update memory related information, if it differs than the
+ * last(cached) one. If it doesn't have flag reset the
+ * memory instance to the default memories[0] */
+#define read_leb_memarg(p, p_end, res)                         \
+    do {                                                       \
+        read_leb_uint32(p, p_end, res);                        \
+        if (!(res & OPT_MEMIDX_FLAG))                          \
+            memidx = 0;                                        \
+        else                                                   \
+            read_leb_memidx(p, p_end, memidx);                 \
```
wenyongh, line 722: Had better use `read_leb_uint32(p, p_end, memidx)` instead, read_leb_memidx will check and update memory/linear_mem_size, which will be repeatedly done below

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -5620,45 +5647,65 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+                        /* Skip src and dst memidx */
                         frame_ip += 2;
+#endif
                         len = POP_MEM_OFFSET();
                         src = POP_MEM_OFFSET();
                         dst = POP_MEM_OFFSET();
 
+#if WASM_ENABLE_MULTI_MEMORY != 0
+                        /* dst memidx */
+                        read_leb_memidx(frame_ip, frame_ip_end, memidx);
```
wenyongh, line 5662: Suggest to put the three lines above, handling with L5653/5654:
```C
#if WASM_ENABLE_MULTI_MEMORY != 0
                        /* dst memidx */
                        read_leb_memidx(frame_ip, frame_ip_end, memidx);
#else
                       /* Skip src and dst memidx */
                        frame_ip += 2;
#endif
```

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -5620,45 +5647,65 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                     {
                         mem_offset_t dst, src, len;
                         uint8 *mdst, *msrc;
+                        uint32 dlen;
```
wenyongh, line 5650: Had better use `uint64` since in memory64, the copy size may be uint64 type

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -5620,45 +5647,65 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+                        /* dst boundary check */
 #ifndef OS_ENABLE_HW_BOUND_CHECK
-                        CHECK_BULK_MEMORY_OVERFLOW(src, len, msrc);
                         CHECK_BULK_MEMORY_OVERFLOW(dst, len, mdst);
 #else
-                        if ((uint64)(uint32)src + len > linear_mem_size)
-                            goto out_of_bounds;
-                        msrc = memory->memory_data + (uint32)src;
-
                         if ((uint64)(uint32)dst + len > linear_mem_size)
```
wenyongh, line 5671: dst may be uint64 type, seems we should not convert it to uint32 before converting uint64?

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -5620,45 +5647,65 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
                             goto out_of_bounds;
                         mdst = memory->memory_data + (uint32)dst;
 #endif
+                        dlen = (uint32)(linear_mem_size - dst);
```
wenyongh, line 5675: Had better declare dlen as uint64, and here just `dlen = linear_mem_size - dst`?

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -5620,45 +5647,65 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
-                        CHECK_BULK_MEMORY_OVERFLOW(src, len, msrc);
                         CHECK_BULK_MEMORY_OVERFLOW(dst, len, mdst);
 #else
-                        if ((uint64)(uint32)src + len > linear_mem_size)
-                            goto out_of_bounds;
-                        msrc = memory->memory_data + (uint32)src;
-
                         if ((uint64)(uint32)dst + len > linear_mem_size)
                             goto out_of_bounds;
                         mdst = memory->memory_data + (uint32)dst;
```
wenyongh, line 5673: Same as above, no need to convert dst to (uint32) here?

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -5620,45 +5647,65 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+                        linear_mem_size = get_linear_mem_size();
+#endif
+                        /* src boundary check */
+#ifndef OS_ENABLE_HW_BOUND_CHECK
+                        CHECK_BULK_MEMORY_OVERFLOW(src, len, msrc);
 #else
-                        /* use memmove when memory64 is enabled since len
-                           may be larger than UINT32_MAX */
-                        memmove(mdst, msrc, len);
+                        if ((uint64)(uint32)src + len > linear_mem_size)
```
wenyongh, line 5688: Same as above, had better remove `(uint32)`?

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -5620,45 +5647,65 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+                            goto out_of_bounds;
+                        msrc = memory->memory_data + (uint32)src;
```
wenyongh, line 5690: Same as above, had better remove `(uint32)`?

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -5620,45 +5647,65 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
 #endif
+
+                        /* allowing the destination and source to overlap */
+                        bh_memmove_s(mdst, (mem_offset_t)dlen, msrc,
+                                     (mem_offset_t)len);
```
wenyongh, line 5695: bh_memmove_s only supports uint32 len, so I used memmove when MEMORY64 is enabled, had better not change it

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -326,12 +340,19 @@ memory_instantiate(WASMModuleInstance *module_inst, WASMModuleInstance *parent,
...
-    memory->memory_data_end = memory->memory_data + memory_data_size;
+    if (is_main_memory) {
+        memory->heap_data = memory->memory_data + heap_offset;
+        memory->heap_data_end = memory->heap_data + heap_size;
+        memory->memory_data_end = memory->memory_data + memory_data_size;
+    }
+    else {
+        memory->heap_data = NULL;
+        memory->heap_data_end = NULL;
+        memory->memory_data_end = NULL;
```
wenyongh, line 351: By default they are zeroed after memory structure is allocated, can remove the else branch?

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -5620,36 +5647,53 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
                         len = POP_MEM_OFFSET();
                         src = POP_MEM_OFFSET();
                         dst = POP_MEM_OFFSET();
 
+#if WASM_ENABLE_MULTI_MEMORY != 0
+                        /* dst memidx */
+                        read_leb_memidx(frame_ip, frame_ip_end, memidx);
+#else
+                        /* Skip src and dst memidx */
+                        frame_ip += 2;
```
wenyongh, line 5661: Had better `frame_ip += 1`, and change comment to `/* Skip dst memidx */`

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -5620,36 +5647,53 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
                             goto out_of_bounds;
-                        msrc = memory->memory_data + (uint32)src;
+                        mdst = memory->memory_data + dst;
+#endif
+                        dlen = linear_mem_size - dst;
 
-                        if ((uint64)(uint32)dst + len > linear_mem_size)
+#if WASM_ENABLE_MULTI_MEMORY != 0
+                        /* src memidx */
+                        read_leb_memidx(frame_ip, frame_ip_end, memidx);
```
wenyongh, line 5678: Had better add else branch:
```C
#else
                        /* Skip src memidx */
                        frame_ip += 1;
#endif
```

core/iwasm/interpreter/wasm_mini_loader.c
```diff
@@ -199,6 +210,35 @@ read_leb(uint8 **p_buf, const uint8 *buf_end, uint32 maxbits, bool sign,
...
+                  < module->import_memory_count + module->memory_count); \
+        (void)memidx;                                                    \
+    } while (0)
+/* Bit 6 indicating the optional memidx, and reset bit 6 for
+ * alignment check */
+#define read_leb_memarg(p, p_end, res)                      \
+    do {                                                    \
+        read_leb_uint32(p, p_end, res);                     \
+        if (res & OPT_MEMIDX_FLAG) {                        \
+            res &= 0xBF;                                    \
```
wenyongh, line 227: Had better `res &= ~OPT_MEMIDX_FLAG`

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -252,6 +263,40 @@ read_leb(uint8 **p_buf, const uint8 *buf_end, uint32 maxbits, bool sign,
...
+            goto fail;                                                      \
+        }                                                                   \
+    } while (0)
+/* Bit 6(0x40) indicating the optional memidx, and reset bit 6 for
+ * alignment check */
+#define read_leb_memarg(p, p_end, res)                      \
+    do {                                                    \
+        read_leb_uint32(p, p_end, res);                     \
+        if (res & OPT_MEMIDX_FLAG) {                        \
+            res &= 0xBF;                                    \
```
wenyongh, line 282: Had better `res &= ~OPT_MEMIDX_FLAG`

wenyongh (approved):
LGTM

core/iwasm/interpreter/wasm_runtime.c
TianlongLiang: Hi, @yamt I think you have a valid point, the app heap in the default memory(index 0) is good enough. I overthink and complex the situation here. Let me list all the situations here for reference:
1. If memory 0 is non-import memory, instantiate the memory as usual. 
2. If memory 0 is import memory, 
    - either it's imported from wasm module(app heap initialized when that wasm submodule memory is instantiated) 
    - or it's imported from the builtin module(like "env" or "spectest" that runtime provides), in that case, it will be instantiated as usual

So it's enough to use the default memory to insert the app heap. What do you think?

yamt: > So it's enough to use the default memory to insert the app heap. What do you think?

it's more than "enough". it's actually how it should be, i guess.

the app heap in non-default memory is almost useless because it can't be addressed by wasm modules written in ordinal languages like C.

TianlongLiang: I agree

wenyongh: @yamt so is there any other comment from you? I guess we can merge it to dev/multi_memory for more testing and merge to main sometime in the future?

yamt: > @yamt so is there any other comment from you?

no other comments.
(i have no time to read the latest version of this PR today.)

> I guess we can merge it to dev/multi_memory for more testing and merge to main sometime in the future?

i guess it's fine.

wenyongh: OK, thanks.

PR #3573, opened 2024-06-26
Use 64 bit wasm_runtime_enlarge_memory() increment

bnason-nf:
_Originally posted by @yamt in https://github.com/bytecodealliance/wasm-micro-runtime/pull/3569#discussion_r1654398315_

### DIFF
--- a/core/iwasm/common/wasm_memory.c
+++ b/core/iwasm/common/wasm_memory.c
@@ -923,6 +923,10 @@ bool
 wasm_runtime_enlarge_memory(WASMModuleInstanceCommon *module_inst,
                             uint64 inc_page_count)
 {
+    if (inc_page_count > UINT32_MAX) {
+        return false;
+    }
+
 #if WASM_ENABLE_AOT != 0
     if (module_inst->module_type == Wasm_Module_AoT) {
         return aot_enlarge_memory((AOTModuleInstance *)module_inst,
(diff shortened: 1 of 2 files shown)


core/iwasm/common/wasm_memory.c
```diff
@@ -921,7 +921,7 @@ wasm_enlarge_memory_internal(WASMModuleInstance *module, uint32 inc_page_count)
 
 bool
 wasm_runtime_enlarge_memory(WASMModuleInstanceCommon *module_inst,
-                            uint32_t inc_page_count)
+                            uint64 inc_page_count)
 {
```
yamt, line 925: please reject inc_page_count > UINT32_MAX rather than truncating silently.

bnason-nf: @yamt: I have done this, but it doesn't protect against `cur_page_count + inc_page_count > UINT32_MAX`, so is really not protecting much at all, and will mean that the person who works on true 64 bit support will need to remove this check.

yamt: > but it doesn't protect against `cur_page_count + inc_page_count > UINT32_MAX`,

really?

i thought this logic was about it. https://github.com/bytecodealliance/wasm-micro-runtime/blob/65bf04e98643f9cbccd6cdc5b8e6d82029c52c38/core/iwasm/common/wasm_memory.c#L789-L792

bnason-nf: @yamt: oh good, I missed that. So then the overflow isn't an issue, just the need to remove this check in the future.

yamt (approved):
lgtm

PR #3577, opened 2024-06-30
Prevent an integer overflow when checking start_func_index

lum1n0us:
[69920](https://oss-fuzz.com/testcase-detail/5082950006276096)

### DIFF
--- a/core/iwasm/aot/aot_loader.c
+++ b/core/iwasm/aot/aot_loader.c
@@ -2486,7 +2486,7 @@ load_init_data_section(const uint8 *buf, const uint8 *buf_end,
     /* check start function index */
     if (module->start_func_index != (uint32)-1
         && (module->start_func_index
-            >= module->import_func_count + module->func_count)) {
+            >= (uint64)module->import_func_count + module->func_count)) {
         set_error_buf(error_buf, error_buf_size,
                       "invalid start function index");
         return false;


core/iwasm/aot/aot_loader.c
```diff
@@ -2486,7 +2486,7 @@ load_init_data_section(const uint8 *buf, const uint8 *buf_end,
     /* check start function index */
     if (module->start_func_index != (uint32)-1
         && (module->start_func_index
-            >= module->import_func_count + module->func_count)) {
+            >= (uint64)module->import_func_count + module->func_count)) {
```
yamt, line 2489: the overflow check should be done earlier, regardless of start_func_index.

yamt (changes requested):

yamt:
btw, does it make much sense to perform fuzzing on the aot loader? (i guess it's what you are doing. this is just a guess because i have no access to the given oss-fuzz url.)
unlike wasm modules, we somehow need to trust aot modules in some extent anyway.

core/iwasm/aot/aot_loader.c
wenyongh: yes, maybe we can check the integer overflow after reading `module->func_count`

lum1n0us:
Use https://github.com/bytecodealliance/wasm-micro-runtime/pull/3579 to fix

PR #3585, opened 2024-07-01
wasm-mutator-fuzz: set compilers earlier

yamt:
CMAKE_C_COMPILER etc should be set before project(), in which cmake tries to decide which compiler to use.


wenyongh (approved):

lum1n0us (approved):
LGTM 👍

Issue #3411, opened 2024-05-09
iwasm: WASM module load failed: type mismatch: expect (ref null ht) but got other1 unknown type


bashor:
### Steps to reproduce

1. Unpack zip [kotlin-wasm-wasi-example-wasm-wasi-dev.wasm.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/15267851/kotlin-wasm-wasi-example-wasm-wasi-dev.wasm.zip)

2. Build `iwasm` with the following command
```
 cmake .. -DWAMR_BUILD_GC=1 -DWAMR_BUILD_BULK_MEMORY=1 -DWAMR_BUILD_EXCE_HANDLING=1 -DWAMR_BUILD_TAIL_CALL=1 -DWAMR_BUILD_CUSTOM_NAME_SECTION=1 -DWAMR_BUILD_EXCE_HANDLING=0 && make
```
3. Run `./iwasm kotlin-wasm-wasi-example-wasm-wasi-dev.wasm`

### Expected behavior

It should print something like:
```
Hello from Kotlin via WASI
Current 'realtime' timestamp is: 1715281633864118000
Current 'monotonic' timestamp is: 210424654558083
```

### Actual behavior
It prints:
```
WASM module load failed: type mismatch: expect (ref null ht) but got other1 unknown type
```

TianlongLiang:
Hi, would you mind telling me which toolchain you are using? I wasn't aware that Kotlin has standalone support for wasm, I would love to give it a try if it's publicly accessible

bashor:
I'm working on Kotlin/Wasm (with other folks), so feel free to reach out if you have any questions.

TianlongLiang:
Thanks, I will definitely check it out!

wenyongh:
Please try #3447.

bashor:
Hi, @wenyongh @TianlongLiang! Thanks for fixes, it really helped with attached examples. 

Previous binaries was compiled with disabled EH on Kotlin side, after turning it on I got the a similar error. 
Here is a new binary --[kotlin-wasm-wasi-example-wasm-wasi.wasm.zip](https://github.com/user-attachments/files/15570015/kotlin-wasm-wasi-example-wasm-wasi.wasm.zip)

wenyongh:
[...]

PR #3586, opened 2024-07-02
wasm loader: Fix several issues in GC and exception handling

wenyongh:
Fix several issues of GC and exception handling in wasm loader:
- Should restore param_reftype_maps/param_reftype_map_count/param_count
  in the handling of opcode throw
- Should set wasm_ref_type when pushing param types of tag type and block type
  if the type is a multi-byte type
- Should set init_values.data as NULL for opcode struct.new_default in load_init_expr

This PR fixes the issues reported in #3411.


xujuntwt95329 (commented):
LGTM

Issue #3580, opened 2024-07-01
buffer overflow in wasm_loader_emit_br_info


yamt:
a crash file for wasm-mutator-fuzz: [crash-27220c0abb57efbe2e501a90e139b1331a5f71d6.gz](https://github.com/user-attachments/files/16047301/crash-27220c0abb57efbe2e501a90e139b1331a5f71d6.gz)
(from https://github.com/yamt/toywasm-fuzzer-corpus)

```
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 1007482135
INFO: Loaded 1 modules   (31916 inline 8-bit counters): 31916 [0x10795e39f, 0x10796604b), 
INFO: Loaded 1 PC tables (31916 PCs): 31916 [0x107966050,0x1079e2b10), 
./wasm_mutator_fuzz: Running 1 inputs 1 time(s) each.
Running: crash-27220c0abb57efbe2e501a90e139b1331a5f71d6
=================================================================
==45439==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x60600000049c at pc 0x00010776f750 bp 0x7ff7b89e7690 sp 0x7ff7b89e7688
READ of size 2 at 0x60600000049c thread T0
    #0 0x10776f74f in wasm_loader_emit_br_info wasm_loader.c:9267
    #1 0x10776cc17 in wasm_loader_check_br wasm_loader.c:10311
    #2 0x10775d92a in check_branch_block wasm_loader.c:10439
    #3 0x107712464 in wasm_loader_prepare_bytecode wasm_loader.c:11674
    #4 0x1076d8a23 in load_from_sections wasm_loader.c:6126
    #5 0x1076e0556 in load wasm_loader.c:6532
    #6 0x1076df8f1 in wasm_loader_load wasm_loader.c:6709
    #7 0x107772b32 in wasm_load wasm_runtime.c:65
    #8 0x1075ed252 in wasm_runtime_load_ex wasm_runtime_common.c:1356
    #9 0x1075edab7 in wasm_runtime_load wasm_runtime_common.c:1401
    #10 0x107513b0f in LLVMFuzzerTestOneInput wasm_mutator_fuzz.cc:34
    #11 0x10784c4e2 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) FuzzerLoop.cpp:612
[...]

PR #3588, opened 2024-07-02
wasm loader: Fix push_frame_offset when pushing v128 type

wenyongh:
Fixes issue https://github.com/bytecodealliance/wasm-micro-runtime/issues/3580.


yamt:
although i feel this is a band-aid fix, i can understand this avoids the particular crash in question.

yamt (approved):

xujuntwt95329 (commented):
LGTM

Issue #2686, opened 2023-10-31
CI should test the top-level cmake files


yamt:
to avoid breakage like https://github.com/bytecodealliance/wasm-micro-runtime/issues/2665

an alternative: remove the top-level cmake files if no one really cares.

lum1n0us:
If removed, will it be strange to people? I mean questions like:
- don't know how to build it
- isn't a good/typical cmake project
- need to be more organized
- ...

PR #3591, opened 2024-07-02
Fix Android pthread linkage

bnason-nf:


CMakeLists.txt
```diff
@@ -147,7 +147,10 @@ if (WAMR_BUILD_STATIC)
     add_library(iwasm_static STATIC ${WAMR_RUNTIME_LIB_SOURCE})
     set_target_properties (iwasm_static PROPERTIES OUTPUT_NAME vmlib)
     target_include_directories(iwasm_static INTERFACE ${WAMR_ROOT_DIR}/core/iwasm/include)
-    target_link_libraries (iwasm_static INTERFACE ${LLVM_AVAILABLE_LIBS} ${UV_A_LIBS} -lm -ldl -lpthread)
+    target_link_libraries (iwasm_static INTERFACE ${LLVM_AVAILABLE_LIBS} ${UV_A_LIBS} -lm -ldl)
+    if (WAMR_BUILD_LIB_PTHREAD EQUAL 1)
+        target_link_libraries (iwasm_static INTERFACE -lpthread)
+    endif()
```
yamt, line 153: pthread is also used for other things like wasi-threads and jit.

yamt (changes requested):

CMakeLists.txt
bnason-nf: Ok, are there any tests or config options that would demonstrate this? All of the checks passed so I'm not sure how to reproduce the issue you're talking about,

yamt: eg. anything with WAMR_BUILD_LIB_WASI_THREADS.

btw, when you say "All of the checks passed", do you mean the CI?
this file is not well tested on the CI. https://github.com/bytecodealliance/wasm-micro-runtime/issues/2686.

bnason-nf: I see, thanks.

wenyongh: Two individual wasm modules may be run in two different threads which are created by host embedder, and pthread APIs are used in some places no matter WAMR_BUILD_LIB_PTRHEAD is enabled or not:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/common/wasm_runtime_common.c#L603-L605
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/shared/mem-alloc/ems/ems_alloc.c#L583

Had better not change here.

wenyongh (approved):
LGTM

yamt (approved):
lgtm

PR #3593, opened 2024-07-03
wasm-mutator-fuzz: disable fast interpreter by default

yamt:
* our interpreters' SIMD support is not complete.

* we want to keep SIMD enabled for this fuzz target because, llvm-jit and aot, which support SIMD, use wasm_loader.c to load and validate the input module.

  well, we probably should test wasm_loader.c with the configuration actually used by llvm-jit and aot. (WASM_ENABLE_JIT/WASM_ENABLE_WAMR_COMPILER) but we ignore the difference for now.

* fast interpreter is not compatible with llvm-jit/aot.

cf. https://github.com/bytecodealliance/wasm-micro-runtime/issues/3580

an alternative:
https://github.com/bytecodealliance/wasm-micro-runtime/pull/3592

### DIFF
--- a/tests/fuzz/wasm-mutator-fuzz/CMakeLists.txt
+++ b/tests/fuzz/wasm-mutator-fuzz/CMakeLists.txt
@@ -76,8 +76,8 @@ if (NOT DEFINED WAMR_BUILD_LIBC_WASI)
 endif ()
 
 if (NOT DEFINED WAMR_BUILD_FAST_INTERP)
-  # Enable fast interpreter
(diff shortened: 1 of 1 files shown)


lum1n0us (approved):
LGTM

wenyongh (commented):
Since we can also configure the running mode and features in oss-fuzz by passing different cmake flags, how about not to change the current CMakeLists.txt?

yamt:
cf. https://github.com/bytecodealliance/wasm-micro-runtime/pull/3611

PR #3595, opened 2024-07-03
Add AOT running mode

bnason-nf:
This adds an enum to RunningMode to be able to determine if a module instance is using AOT. I'm not sure this is the best solution for that, if there's a better way please let me know.

### DIFF
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -841,6 +841,11 @@ wasm_runtime_is_running_mode_supported(RunningMode running_mode)
         return true;
 #endif
     }
+    else if (running_mode == Mode_AOT) {
+#if WASM_ENABLE_AOT != 0
+        return true;
+#endif
+    }
 
     return false;
 }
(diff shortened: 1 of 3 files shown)


wenyongh:
@bnason-nf The aot file can only be executed with AOT mode, and whether an input buffer is an AOT file can be detected by the API exported in wasm_export.h:
```C
https://github.com/bytecodealliance/wasm-micro-runtime/blob/1f94cd4ee506b063d0698d6b40bcff69d5d6f27f/core/iwasm/include/wasm_export.h#L419-L420
```
so we didn't add the AOT mode in the enum. Not sure whether it is good to you?

bnason-nf:
> so we didn't add the AOT mode in the enum. Not sure whether it is good to you?

Thanks @wenyongh, I hadn't noticed that API, I could use it if needed, but I made a new API in #3600 that is slightly better, since it doesn't require me to have the module data and doesn't do any extra memory comparison to get information it already knows. I hope it's ok with you.

Also, as a side note, the name of get_package_type() seems inconsistent with the rest of the APIs, maybe it would be nice to change in the future (even though that would be a breaking change).

PR #3600, opened 2024-07-05
Add wasm_runtime_module_get_package_type()

bnason-nf:
See discussion in #3595.

### DIFF
--- a/core/iwasm/include/wasm_export.h
+++ b/core/iwasm/include/wasm_export.h
@@ -420,11 +420,23 @@ WASM_RUNTIME_API_EXTERN package_type_t
 get_package_type(const uint8_t *buf, uint32_t size);
 
 /**
+ * Get the package type of a buffer (same as get_package_type).
+ *
+ * @param buf the package buffer
+ * @param size the package buffer size
+ *
+ * @return the package type, return Package_Type_Unknown if the type is unknown
+ */
+WASM_RUNTIME_API_EXTERN package_type_t
+wasm_runtime_get_file_package_type(const uint8_t *buf, uint32_t size);
+
+/**
  * Get the package type of a module.
  *
  * @param module the module
  *
- * @return the package type, return Package_Type_Unknown if the type is unknown
+ * @return the package type, return Package_Type_Unknown if the type is
+ * unknown
  */
 WASM_RUNTIME_API_EXTERN package_type_t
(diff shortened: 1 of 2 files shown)


core/iwasm/include/wasm_export.h
```diff
@@ -419,6 +419,16 @@ wasm_runtime_get_mem_alloc_info(mem_alloc_info_t *mem_alloc_info);
...
 
+/**
+ * Get the package type of a module.
+ *
+ * @param module the module
+ *
+ * @return the package type, return Package_Type_Unknown if the type is unknown
+ */
+WASM_RUNTIME_API_EXTERN package_type_t
+wasm_runtime_module_get_package_type(wasm_module_t module);
```
wenyongh, line 430: How about `get_module_package_type` to be consistent with `get_package_type`?

bnason-nf: Hi @wenyongh, we can use that naming if you like, but it seems inconsistent with every the other API in wasm_export.h. The `get_package_type` function is the only one that doesn't start with `wasm_`. In my opinion it would be better to rename `get_package_type` in the future to be consistent with the `wasm_*` naming (if there might be an opportunity to make a breaking change like this). But in the end, it's your decision, so please confirm either way.

wenyongh: OK, if you prefer to be consistent with most of the other APIs, how about using `wasm_runtime_get_module_package_type`?
And for `get_package_type`, we had better not remove it, how about adding another API `wasm_runtime_get_file_package_type` as a counterpart of it?

core/iwasm/include/wasm_export.h
```diff
@@ -419,6 +419,28 @@ wasm_runtime_get_mem_alloc_info(mem_alloc_info_t *mem_alloc_info);
...
+/**
+ * Get the package type of a module.
+ *
+ * @param module the module
+ *
+ * @return the package type, return Package_Type_Unknown if the type is
+ * unknown
+ */
+WASM_RUNTIME_API_EXTERN package_type_t
+wasm_runtime_get_module_package_type(wasm_module_t module);
```
wenyongh, line 442: had better `const wasm_module_t module`

wenyongh (commented):
LGTM exception minor comment

PR #3601, opened 2024-07-05
Add apis to get package version

bnason-nf:


core/iwasm/include/wasm_export.h
```diff
@@ -1826,6 +1836,12 @@ wasm_runtime_get_custom_section(const wasm_module_t module_comm,
 WASM_RUNTIME_API_EXTERN void
 wasm_runtime_get_version(uint32_t *major, uint32_t *minor, uint32_t *patch);
 
+/**
+ * Get max supported version for a package type
+ */
+WASM_RUNTIME_API_EXTERN unsigned int
+wasm_runtime_get_max_package_version(package_type_t package_type);
```
wenyongh, line 1843: How about `uint32 get_package_current_version(package_type_t package_type)`?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -871,6 +871,25 @@ get_package_type(const uint8 *buf, uint32 size)
     return Package_Type_Unknown;
 }
 
+unsigned int
+wasm_runtime_module_get_package_version(wasm_module_t module)
```
wenyongh, line 875: No need to add this function if #3600 is merged, we can use
```C
    package_type_t package_type = get_module_package_type(module);
    uint32 package_current_version = get_package_current_version(package_type);
```
instead.

bnason-nf: I agree that this would work for existing WAMR implementations, which only support a single version of WASM files, and a single version of AOT files. When I created the API, I was also considering the possibility that some future WAMR implementation might support multiple versions of either one of these types of files. Do you think it's worth planning for that possibility in this API, or do you think if for example WASM version 2 is released that WAMR would stop supporting WASM version 1?

core/iwasm/include/wasm_export.h
bnason-nf: I think this is the same type of naming decision as #3600, so I'll just refer to my comment on that issue:

https://github.com/bytecodealliance/wasm-micro-runtime/pull/3600#discussion_r1668790429

And whatever is decided there, we should probably also do here.

core/iwasm/common/wasm_runtime_common.c
wenyongh: For AOT, normally the AOT runtime only supports the latest AOT file format, or AOT_CURRENT_VERSION, it is difficult to support backward compatibility since the AOT file format may change a lot from one version to another version. For interpreter, I am not very sure but it may be possible to support both two versions in runtime, since normally there is a good backward compatibility in the wasm file format.

Here I think maybe we want to get: (1) the version in the wasm/aot file, (2) the version of a wasm/aot module, (3) the current version that runtime supports.
For (1), how about:
```C
uint32_t wasm_runtime_get_file_package_version(const uint8_t *buf, uint32_t size);
```

For (2), I think we should add a field (e.g. `uint32 package_version`) in WASMModule/AOTModule to save the version loaded and return it in the API, how about adding the api:
```C
uint32_t wasm_runtime_get_module_package_version(const wasm_module_t module);
```

For (3), how about
```C
uint32_t wasm_runtime_get_current_package_version(package_type_t package_type);

or

uint32_t wasm_runtime_get_max_package_version(package_type_t package_type);
```
which returns AOT_CURRENT_VERSION or WASM_CURRENT_VERSION. But I am not sure which API name is better.

How do you think?

bnason-nf:
@wenyongh one of the checks failed, but it seems unrelated to my change.

core/iwasm/interpreter/wasm.h
```diff
@@ -1079,6 +1079,8 @@ struct WASMModule {
 
     /* Whether the underlying wasm binary buffer can be freed */
     bool is_binary_freeable;
+
+    uint32_t package_version;
```
wenyongh, line 1083: Same as above, how about changing to `uint32 package_version` and moving it to be after `uint32 module_type`?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -858,10 +858,6 @@ wasm_runtime_set_default_running_mode(RunningMode running_mode)
 PackageType
 get_package_type(const uint8 *buf, uint32 size)
 {
-#if (WASM_ENABLE_WORD_ALIGN_READ != 0)
-    uint32 buf32 = *(uint32 *)buf;
-    buf = (const uint8 *)&buf32;
-#endif
```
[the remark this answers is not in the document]
bnason-nf: I see that the block is trying to handle a special case, but that code is not correct if `buf` is null or `size` is less than 4, which is checked on the next line after that block. Also, it looks like this block is trying to fix an endian-ness problem, but it seems unnecessary since `buf` is accessed as a byte array, unless I'm misunderstanding. Do you know what's special about this device that would require this block?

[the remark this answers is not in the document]
bnason-nf: If you want me to restore this block just let me know, but I think it should at minimum be moved inside the `if` block. Also, if this block is needed, then I will need to add a similar block in `wasm_runtime_get_file_package_version()`.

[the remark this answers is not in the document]
wenyongh: Thanks for pointing out the issue, yes, had better move them into the if block, and apply similar change to wasm_runtime_get_file_package_version.

If IIUC, in the device, we can only read data in 4-byte/word aligned address, and here `*(uint32 *)buf` is good, but `buf[1] == 'a'` will cause core dump. See also:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/aot/aot_loader.c#L131-L148

core/iwasm/interpreter/wasm.h
```diff
@@ -836,6 +836,9 @@ struct WASMModule {
        AOTModule structure. */
     uint32 module_type;
 
+    /* the package version, should be WASM_CURRENT_VERSION */
```
wenyongh, line 839: The comment may be not so accurate, e.g. in the future, WASM_CURRENT_VERSION is upgraded to 2, but here the package_version can be 1 or 2. How about somewhat like the file version read from the wasm file?

core/iwasm/aot/aot_runtime.h
```diff
@@ -130,6 +130,9 @@ typedef struct LocalRefFlag {
 typedef struct AOTModule {
     uint32 module_type;
 
+    /* the package version, should be AOT_CURRENT_VERSION */
```
wenyongh, line 133: Same as wasm.h, how about somewhat like the file version read from the AOT file?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -887,6 +887,60 @@ wasm_runtime_get_module_package_type(WASMModuleCommon *module)
...
 }
 
+uint32
+wasm_runtime_get_file_package_version(const uint8 *buf, uint32 size)
+{
+    if (buf && size >= 8) {
+        uint32 version;
+#if (WASM_ENABLE_WORD_ALIGN_READ != 0)
+        uint32 buf32 = *(uint32 *)(buf + sizeof(uint32));
+        buf = (const uint8 *)&buf32;
```
wenyongh, line 897: It is invalid to read buf[4] to buf[7] below, should read buf[0] to buf[3] instead. How about:
```C
#if (WASM_ENABLE_WORD_ALIGN_READ != 0)
        uint32 buf32 = *(uint32 *)(buf + sizeof(uint32));
        buf = (const uint8 *)&buf32;
        version = buf[0] | buf[1] << 8 | buf[2] << 16 | buf[3] << 24;
#else
        version = buf[4] | buf[5] << 8 | buf[6] << 16 | buf[7] << 24;
#endif
        return version;
```

core/iwasm/aot/aot_runtime.h
bnason-nf: This seems to contradict your earlier comment:
https://github.com/bytecodealliance/wasm-micro-runtime/pull/3601#discussion_r1669519462

core/iwasm/common/wasm_runtime_common.c
bnason-nf: Good catch, I'll fix.

core/iwasm/aot/aot_runtime.h
bnason-nf: I will change it anyway.

wenyongh (approved):
LGTM

core/iwasm/aot/aot_runtime.h
wenyongh: > This seems to contradict your earlier comment: [#3601 (comment)](https://github.com/bytecodealliance/wasm-micro-runtime/pull/3601#discussion_r1669519462)

Yes, but I am not very sure whether AOT runtime of newer version will support AOT file of some older version in the future. Thanks for the update.

PR #3616, opened 2024-07-11
Fix compile warnings/error reported in Windows

wenyongh:
Clear some compile warnings and fix undefined reference error for symbol ffs
in Windows platform.

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -1139,7 +1139,7 @@ memories_instantiate(AOTModuleInstance *module_inst, AOTModuleInstance *parent,
 
         if (memory_inst->memory_data) {
             bh_memcpy_s((uint8 *)memory_inst->memory_data + base_offset,
-                        (uint32)memory_inst->memory_data_size - base_offset,
+                        (uint32)(memory_inst->memory_data_size - base_offset),
                         data_seg->bytes, length);
         }
     }
@@ -1212,7 +1212,7 @@ aot_get_function_instance(AOTModuleInstance *module_inst, uint32 func_idx)
             return NULL;
         }
 
-        extra->function_count = func_count;
+        extra->function_count = (uint32)func_count;
     }
 
     /* instantiate function if needed */
@@ -1765,7 +1765,7 @@ aot_instantiate(AOTModule *module, AOTModuleInstance *parent,
 
 #if WASM_ENABLE_PERF_PROFILING != 0
     total_size = (uint64)sizeof(AOTFuncPerfProfInfo)
-                 * (module->import_func_count + module->func_count);
+                 * ((uint64)module->import_func_count + module->func_count);
     if (!(module_inst->func_perf_profilings =
               runtime_malloc(total_size, error_buf, error_buf_size))) {
         goto fail;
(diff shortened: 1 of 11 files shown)


core/iwasm/aot/aot_runtime.c
```diff
@@ -1765,7 +1765,7 @@ aot_instantiate(AOTModule *module, AOTModuleInstance *parent,
 
 #if WASM_ENABLE_PERF_PROFILING != 0
     total_size = (uint64)sizeof(AOTFuncPerfProfInfo)
-                 * (module->import_func_count + module->func_count);
+                 * ((uint64)module->import_func_count + module->func_count);
```
yamt, line 1768: what problem does this cast solve?

core/iwasm/common/wasm_loader_common.c
```diff
@@ -104,7 +104,9 @@ bool
 is_valid_func_type(const WASMFuncType *func_type)
 {
     unsigned i;
-    for (i = 0; i < func_type->param_count + func_type->result_count; i++) {
+    for (i = 0;
+         i < (unsigned)(func_type->param_count + func_type->result_count);
```
yamt, line 108: what problem does this solve?

core/iwasm/compilation/aot_llvm.c
```diff
@@ -748,8 +748,9 @@ aot_add_llvm_func(AOTCompContext *comp_ctx, LLVMModuleRef module,
              * should be ok. We prefer short-call because it's smaller
              * and more importantly doesn't involve relocations.
              */
-            LLVMAttributeRef attr_short_call = LLVMCreateStringAttribute(
-                comp_ctx->context, "short-call", strlen("short-call"), "", 0);
+            LLVMAttributeRef attr_short_call =
+                LLVMCreateStringAttribute(comp_ctx->context, "short-call",
+                                          (uint32)strlen("short-call"), "", 0);
```
yamt, line 753: i guess this should be unsigned, not uint32.

core/iwasm/aot/aot_runtime.c
wenyongh: The compiler hints that integer overflow may occur in `module->import_func_count + module->func_count`.

core/iwasm/common/wasm_loader_common.c
wenyongh: the type of `func_type->param_count` and `func_type->result_count` is uint16, when comparing to `unsigned i`, `func_type->param_count + func_type->result_count` is converted from uint16 into signed int, and the compiler reported warning of comparison between signed and unsigned.

core/iwasm/compilation/aot_llvm.c
wenyongh: yes, the related parameter of LLVMCreateStringAttribute is `unsigned KLength` indeed.

core/iwasm/aot/aot_runtime.c
yamt: but we know this doesn't overflow, don't we?
i wonder what made the compiler think this particular addition may be more problematic than others.

wenyongh: Here it should not overflow, the overflow was checked in aot loader:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/aot/aot_loader.c#L2487
I just changed the code to let the compiler happy.

Maybe it is due to the adding result is converted to uint64. The msvc reported:
```
Warning	C26451	Arithmetic overflow: Using operator '+' on a 4 byte value and
then casting the result to a 8 byte value. Cast the value to the wider type before
calling operator '+' to avoid overflow (io.2).
```

TianlongLiang:
LGTM

PR #3617, opened 2024-07-11
Fix compile errors on workload bwa and benchmark jetstream

wenyongh:
Error was reported when building jetstream benchmark with emsdk:
```
clang: error: unsupported option '-msse2' for target 'wasm32-wasi'
```

Error was reported when building bwa workload:
```
fatal: reference is not a tree: 139f68fc4c3747813783a488aef2adc86626b01b
```


TianlongLiang:
LGTM

PR #3632, opened 2024-07-16
spec-test-script/runtest.py: don't assume the tmp dir path

yamt:
The current coding assumes the tmp file is created in /tmp. However,

* It's somewhere under /var/folders/... by default on macOS.

* A user can set TMPDIR.

### DIFF
--- a/tests/wamr-test-suites/spec-test-script/runtest.py
+++ b/tests/wamr-test-suites/spec-test-script/runtest.py
@@ -1158,6 +1158,9 @@ def run_wasm_with_repl(wasm_tempfile, aot_tempfile, opts, r):
     tmpfile = aot_tempfile if test_aot else wasm_tempfile
     log("Starting interpreter for module '%s'" % tmpfile)
 
+    if opts.qemu:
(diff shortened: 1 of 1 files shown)


tests/wamr-test-suites/spec-test-script/runtest.py
```diff
@@ -1158,6 +1158,9 @@ def run_wasm_with_repl(wasm_tempfile, aot_tempfile, opts, r):
     tmpfile = aot_tempfile if test_aot else wasm_tempfile
     log("Starting interpreter for module '%s'" % tmpfile)
 
+    if opts.qemu:
+        tmpfile = f"/tmp/{os.path.basename(tmpfile)}"
```
wenyongh, line 1162: Is better to move L1159 to be after these two lines, or the log may be not so accurate?

wenyongh: OK, thanks.

wenyongh (commented):
LGTM

no1wudi (approved):

Issue #3311, opened 2024-04-12
zephyr: unable to compile with `WAMR_BUILD_LIBC_WASI` flag.


lucasAbadFr:
Hello, I'm currently experimenting with sockets in WAMR on Zephyr, and I'm encountering some challenges.
1. I've adapted a section of the Zephyr socket API.
2. I'm attempting to run an HTTP client on WAMR (largely inspired by [this sample](https://docs.zephyrproject.org/latest/samples/net/sockets/http_get/README.html)). 
3. I'm working with the stm32 `nucleo-h563zi` board, but I'm facing compilation issues.

### Compilation issues

1. **Plain compilation**

      I compilated the `http_client` to a wasm module and like with the `simple` sample, I included it in a header file. But when I ran it on the board I had:
      ```bash
      *** Booting Zephyr OS build v3.6.0-961-gddb147d0a45d ***
      [00:00:00:000 - 20001190]: warning: failed to link import function (env, getaddrinfo)
      [00:00:00:000 - 20001190]: warning: failed to link import function (env, socket)
      [00:00:00:000 - 20001190]: warning: failed to link import function (env, connect)
      [00:00:00:000 - 20001190]: warning: failed to link import function (env, send)
      [00:00:00:000 - 20001190]: warning: failed to link import function (env, recv)
      [00:00:00:000 - 20001190]: warning: failed to link import function (env, close)
      [00:00:00:000 - 20001190]: warning: failed to link import function (env, __errno_location)
      Preparing HTTP GET request for http://192.0.2.10:8000/
      Exception: failed to call unlinked import function (env, getaddrinfo)
      elapsed: 72
      ```

      > 📄 **Notes:** I later adapted the sample to not use `getaddrinfo` because it add a useless dependency to `netdb.h` that I did not want.

2. **WAMR_BUILD_LIBC_WASI flag**

      After looking at the `socket-api` sample, I found that I need to use the `WAMR_BUILD_LIBC_WASI` flag, so I added this flag to the `CMakeLists.txt` file but it didn't work.

      The output of the compilation is the same described in this [issue](https://github.com/bytecodealliance/wasm-micro-runtime/issues/2846)
      My errors concist mainly of:
      1. `libc_errno.h` not found.
            ```bash	
            /home/user/wasm-micro-runtime/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/random.c:16:10: fatal error: libc_errno.h: No such file or directory
[...]

PR #3633, opened 2024-07-16
Zephyr: WASI support for sockets and file system

lucasAbadFr:
To address #3311. 

This work also implement the WASI support on Zephyr.


lucasAbadFr:
Hello @wenyongh !

Reaching out after two months, like we said previously I tried to make CI run. 
I also increased support for file system API. 

I might need some helps with the various platforms to pass every tests.

Thanks you.

wenyongh:
> Hello @wenyongh !
> 
> Reaching out after two months, like we said previously I tried to make CI run. I also increased support for file system API.
> 
> I might need some helps with the various platforms to pass every tests.
> 
> Thanks you.

@lucasAbadFr Welcome. Thanks for the great effort, please feel free to ask question about CI if needed, we will try to help resolve it. And BTW, had better rebase your repo with the main branch since some CI errors were fixed recently.

lucasAbadFr:
@wenyongh I managed to make most tests pass, but there are still 6 tests failing: 
 * **1 Windows latest:** build (-DWAMR_BUILD_LIBC_UVWASI=0 -DWAMR_BUILD_LIBC_WASI=1)
 * **5 android:** test (ubuntu-22.04, multi-tier-jit, $WASI_TEST_OPTIONS,...
     * test poll_oneoff_stdio fail
     * test tcp_udp fail

Not sure why these builds/tests are failing.

wenyongh:
@lucasAbadFr the errors reported by CIs can be found below (you can click Actions or visit https://github.com/bytecodealliance/wasm-micro-runtime/actions):

https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/10090660314/job/27900575529
```bash
D:\a\wasm-micro-runtime\wasm-micro-runtime\core\shared\platform\windows\win_file.c(1799,9): error C2231: '.type': left operand points to 'struct', use '->' [D:\a\wasm-micro-runtime\wasm-micro-runtime\product-mini\platforms\windows\build\libiwasm.vcxproj]
D:\a\wasm-micro-runtime\wasm-micro-runtime\core\shared\platform\windows\win_file.c(1799,9): error C2231: '.type': left operand points to 'struct', use '->' [D:\a\wasm-micro-runtime\wasm-micro-runtime\product-mini\platforms\windows\build\vmlib.vcxproj]
D:\a\wasm-micro-runtime\wasm-micro-runtime\core\shared\platform\windows\win_file.c(1799,25): error C2231: '.type': left operand points to 'struct', use '->' [D:\a\wasm-micro-runtime\wasm-micro-runtime\product-mini\platforms\windows\build\libiwasm.vcxproj]
D:\a\wasm-micro-runtime\wasm-micro-runtime\core\shared\platform\windows\win_file.c(1799,25): error C2231: '.type': left operand points to 'struct', use '->' [D:\a\wasm-micro-runtime\wasm-micro-runtime\product-mini\platforms\windows\build\vmlib.vcxproj]
D:\a\wasm-micro-runtime\wasm-micro-runtime\core\shared\platform\windows\win_file.c(1803,9): error C2231: '.fdflags': left operand points to 'struct', use '->' [D:\a\wasm-micro-runtime\wasm-micro-runtime\product-mini\platforms\windows\build\libiwasm.vcxproj]
```

https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/10090660328/job/27900702537
```bash
Test poll_oneoff_stdio failed
  [exit_code] 0 == 1
STDOUT:
Exception: unreachable

STDERR:
thread 'main' panicked at 'stdin read should return at least 1 event', src/bin/poll_oneoff_stdio.rs:51:5

...

Test tcp_udp failed
  [exit_code] 0 == 1
STDOUT:
Testing address family: 1 protocol: 6
Exception: unreachable

STDERR:
Assertion failed: sizeof(struct sockaddr_in6) <= addrlen (../src/wasi/wasi_socket_ext.c: sockaddr_to_wasi_addr: 74)
```

And there are compilation warnings:
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/10090660328/job/27900704394
```bash
In file included from /home/runner/work/wasm-micro-runtime/wasm-micro-runtime/core/shared/utils/bh_platform.h:13,
                 from /home/runner/work/wasm-micro-runtime/wasm-micro-runtime/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/ssp_config.h:17,
                 from /home/runner/work/wasm-micro-runtime/wasm-micro-runtime/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c:14:
In function ‘wasmtime_ssp_sock_recv_from’,
    inlined from ‘wasmtime_ssp_sock_recv’ at /home/runner/work/wasm-micro-runtime/wasm-micro-runtime/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c:2803:12:
/home/runner/work/wasm-micro-runtime/wasm-micro-runtime/core/shared/utils/bh_common.h:17:20: warning: ‘src_addr’ may be used uninitialized [-Wmaybe-uninitialized]
   17 |         int _ret = b_memcpy_s(dest, dlen, src, slen); \
      |                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/wasm-micro-runtime/wasm-micro-runtime/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c:2819:5: note: in expansion of macro ‘bh_memcpy_s’
 2819 |     bh_memcpy_s(&src_addr_copy, sizeof(__wasi_addr_t), src_addr,
      |     ^~~~~~~~~~~
/home/runner/work/wasm-micro-runtime/wasm-micro-runtime/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c: In function ‘wasmtime_ssp_sock_recv’:
/home/runner/work/wasm-micro-runtime/wasm-micro-runtime/core/shared/utils/bh_common.h:51:1: note: by argument 3 of type ‘const void *’ to ‘b_memcpy_s’ declared here
   51 | b_memcpy_s(void *s1, unsigned int s1max, const void *s2, unsigned int n);
```

Maybe you can try to resolve the compiler error and warnings first, and refer to the CI steps to download/build related tools/cases to test the wasi case failure (Test poll_oneoff_stdio/tcp_udp failed).  Since there are lots of changes in the PR, we need more time to read the code and give comments, the response may be a little late. Thanks.

wenyongh:
BTW, the PR https://github.com/bytecodealliance/wasm-micro-runtime/pull/3650 was merged, had better rebase your code with main branch.

wenyongh:
@lucasAbadFr thanks for the update! It seems that you have basically implemented the feature and all the CIs run succeeded, so is it ready for code review?

lucasAbadFr:
@wenyongh it would be appreciated !

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -2809,12 +2814,19 @@ wasmtime_ssp_sock_recv_from(wasm_exec_env_t exec_env, struct fd_table *curfds,
     __wasi_errno_t error;
     bh_sockaddr_t sockaddr;
     int ret;
+    /* make a copy because the value pointed seem to be modifed */
```
wenyongh, line 2817: modifed -> modified

Do you mean src_addr will be modified below? But it is only used in the end of function: 
`bh_sockaddr_to_wasi_addr(&sockaddr, src_addr)`

And you convert src_addr_copy to sockaddr with `wasi_addr_to_bh_sockaddr(&src_addr_copy, &sockaddr)`, and then pass `sockaddr` to blocking_op_socket_recv_from, but it seems that the input fields of `sockaddr` are not used in posix os_socket_recv_from:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/shared/platform/common/posix/posix_socket.c#L258-L284

I think it is because that in your Zephyr's os_socket_recv_from implementation, you need the input value of sockaddr?
```C
temp_addr->sa_family = src_addr->is_ipv4 == true ? AF_INET : AF_INET6;
```
If yes, then had better change the comment to make it more accurate.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -2864,8 +2876,12 @@ wasmtime_ssp_sock_send_to(wasm_exec_env_t exec_env, struct fd_table *curfds,
     __wasi_errno_t error;
     int ret;
     bh_sockaddr_t sockaddr;
+    /* make a copy because the value pointed seem to be modifed */
```
wenyongh, line 2879: Same as above, seems it is because that Zephyr's os_socket_send_to implementation uses the input of dest_addr?

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -3124,15 +3143,14 @@ compare_address(const struct addr_pool *addr_pool_entry,
     }
 
     init_address_mask(maskbuf, addr_size, addr_pool_entry->mask);
-
```
wenyongh, line 3127: had better not change here?

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -3124,15 +3143,14 @@ compare_address(const struct addr_pool *addr_pool_entry,
...
     for (size_t i = 0; i < addr_size; i++) {
         uint8_t addr_mask = target->data[i] & maskbuf[i];
-        uint8_t range_mask = basebuf[i] & maskbuf[i];
```
wenyongh, line 3130: same as above

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -3124,15 +3143,14 @@ compare_address(const struct addr_pool *addr_pool_entry,
...
+        uint8_t range_mask;
+        range_mask = basebuf[i] & maskbuf[i];
         if (addr_mask != range_mask) {
             return false;
         }
     }
-
```
wenyongh, line 3135: same as above

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/ssp_config.h
```diff
@@ -42,28 +42,30 @@
...
     && !defined(ESP_PLATFORM) && !defined(DISABLE_CLOCK_NANOSLEEP)           \
-    && !defined(BH_PLATFORM_FREERTOS)
+    && !defined(BH_PLATFORM_FREERTOS) && !defined(BH_PLATFORM_ZEPHYR)
 #define CONFIG_HAS_CLOCK_NANOSLEEP 1
 #else
 #define CONFIG_HAS_CLOCK_NANOSLEEP 0
 #endif
 
-#if defined(__APPLE__) || defined(__CloudABI__)
+#if defined(__APPLE__) || defined(__CloudABI__) && !defined(BH_PLATFORM_ZEPHYR)
```
wenyongh, line 60: `defined(__APPLE__) || defined(__CloudABI__)` means `BH_PLATFORM_ZEPHYR` isn't defined, should not change this line? Or do you mean `|| defined(BH_PLATFORM_ZEPHYR)`?

core/shared/platform/common/posix/posix_socket.c
```diff
@@ -260,7 +260,7 @@ os_socket_recv_from(bh_socket_t socket, void *buf, unsigned int len, int flags,
                     bh_sockaddr_t *src_addr)
 {
     struct sockaddr_storage sock_addr = { 0 };
-    socklen_t socklen = sizeof(sock_addr);
+    socklen_t socklen = sizeof(struct sockaddr_storage);
```
wenyongh, line 263: Is there issue found in the original code? Had better not change the original implementation?

core/shared/platform/common/posix/posix_socket.c
```diff
@@ -276,9 +276,10 @@ os_socket_recv_from(bh_socket_t socket, void *buf, unsigned int len, int flags,
             return -1;
         }
     }
-    else if (src_addr) {
-        memset(src_addr, 0, sizeof(*src_addr));
-    }
+    // This else if seem to overwrite src_addr when we don't want it to
```
wenyongh, line 279: Same as above, if there is bug found in posix socket implementation, had better file an issue in the WAMR community, so others can check it and figure out how to resolve it.

core/shared/platform/linux/platform_internal.h
```diff
@@ -126,6 +130,10 @@ typedef int os_file_handle;
 typedef DIR *os_dir_stream;
 typedef int os_raw_file_handle;
 
+typedef struct pollfd os_poll_file_handle;
+typedef nfds_t os_nfds_t;
+typedef struct timespec os_timespec;
```
wenyongh, line 135: duplicated definitions, maybe we can remove L67 to L70?

core/shared/platform/zephyr/zephyr_platform.c
```diff
@@ -199,11 +199,14 @@ os_mmap(void *hint, size_t size, int prot, int flags, os_file_handle file)
     return addr;
 }
 
+#if !defined(WASM_ENABLE_LIBC_WASI)
+/* os_mremap cause linking errors */
```
wenyongh, line 203: Seems no need to change this file if `core/shared/platform/zephyr/shared_platform.cmake` doesn't add the two lines below?
```cmake
include (${CMAKE_CURRENT_LIST_DIR}/../common/memory/platform_api_memory.cmake)
set (source_all ${source_all} ${PLATFORM_COMMON_MEMORY_SOURCE})
```

core/shared/platform/zephyr/zephyr_clock.c
```diff
@@ -0,0 +1,22 @@
...
+
+#include "platform_api_extension.h"
+#include "platform_api_vmcore.h"
+
+#include "libc_errno.h"
+
+__wasi_errno_t
+os_clock_res_get(__wasi_clockid_t clock_id, __wasi_timestamp_t *resolution)
+{
+    return __WASI_ENOSYS;
```
wenyongh, line 14: No clock related zephyr API to implement this function? Or could we implement it with k_uptime_get?

core/shared/platform/zephyr/zephyr_file.c
```diff
@@ -0,0 +1,967 @@
...
+    struct zephyr_fs_desc *ptr = NULL;
+    *index = -1; // give a default value to index in case table is full
+
+    k_mutex_lock(&desc_array_mutex, K_FOREVER);
+    for (i = 0; i < CONFIG_WASI_MAX_OPEN_FILES; i++) {
+        if (desc_array[i].used == false) {
+            ptr = &desc_array[i];
+            ptr->used = true;
+            ptr->is_dir = is_dir;
+            ptr->path = strdup(path);
```
wenyongh, line 77: should check the return value

core/shared/platform/zephyr/zephyr_file.c
```diff
@@ -0,0 +1,967 @@
...
+    struct zephyr_fs_desc *ptr;
+
+    *out = BH_MALLOC(sizeof(struct zephyr_handle));
+    if (*out == NULL) {
+        return __WASI_ENOMEM;
+    }
+
+    ptr = zephyr_fs_alloc_obj(true, path, &index);
+    if (ptr == NULL) {
+        return __WASI_EMFILE;
```
wenyongh, line 277: Should BH_FREE(*out) before return

core/shared/platform/zephyr/zephyr_file.c
```diff
@@ -0,0 +1,967 @@
...
+    }
+
+    fs_dir_t_init(&ptr->dir);
+
+    rc = fs_opendir(&ptr->dir, path);
+    if (rc < 0) {
+        zephyr_fs_free_obj(ptr);
+        return convert_errno(-rc);
```
wenyongh, line 285: same as above

core/shared/platform/zephyr/zephyr_file.c
```diff
@@ -0,0 +1,967 @@
...
+
+    *out = BH_MALLOC(sizeof(struct zephyr_handle));
+
+    snprintf(abs_path, MAX_FILE_NAME, "%s/%s", prestat_dir, path);
+    int zmode =
+        wasi_flags_to_zephyr(oflags, fd_flags, lookup_flags, access_mode);
+
+    ptr = zephyr_fs_alloc_obj(false, abs_path, &index);
+    if (!ptr && (index < 0)) {
+        return __WASI_EMFILE;
```
wenyongh, line 374: should BH_FREE(*out) before return

core/shared/platform/zephyr/zephyr_file.c
```diff
@@ -0,0 +1,967 @@
...
+    }
+
+    fs_file_t_init(&ptr->file);
+    rc = fs_open(&ptr->file, abs_path, zmode);
+    if (rc < 0) {
+        zephyr_fs_free_obj(ptr);
+        return convert_errno(-rc);
```
wenyongh, line 381: same as above

core/shared/platform/zephyr/zephyr_file.c
```diff
@@ -0,0 +1,967 @@
...
+os_writev(os_file_handle handle, const struct __wasi_ciovec_t *iov, int iovcnt,
+          size_t *nwritten)
+{
+    struct zephyr_fs_desc *ptr = NULL;
+    ssize_t total_written = 0;
+
+    GET_FILE_SYSTEM_DESCRIPTOR(handle->fd, ptr);
+
+    if (3 < strlen(ptr->path) && ptr->path[0] == 's' && ptr->path[1] == 't'
+        && ptr->path[2] == 'd') {
```
wenyongh, line 557: can use `if (strncmp(ptr->path, "std", 3) == 0)` instead?

core/shared/platform/zephyr/zephyr_file.c
```diff
@@ -0,0 +1,967 @@
...
+#define STDIN_FILENO 0
+#endif
+    os_file_handle handle = BH_MALLOC(sizeof(struct zephyr_handle));
+    if (handle == NULL) {
+        return NULL;
+    }
+    struct zephyr_fs_desc *ptr = NULL;
+
+    /* We allocate a fake stdin reference */
+    ptr = zephyr_fs_alloc_obj(false, "stdin", &handle->fd);
```
wenyongh, line 801: should check the return value? And when is ptr used?

core/shared/platform/zephyr/zephyr_file.c
```diff
@@ -0,0 +1,967 @@
...
+#define STDOUT_FILENO 1
+#endif
+    os_file_handle handle = BH_MALLOC(sizeof(struct zephyr_handle));
+    if (handle == NULL) {
+        return NULL;
+    }
+    struct zephyr_fs_desc *ptr = NULL;
+
+    /* We allocate a fake stdin reference */
+    ptr = zephyr_fs_alloc_obj(false, "stdout", &handle->fd);
```
wenyongh, line 820: same as above

core/shared/platform/zephyr/zephyr_file.c
```diff
@@ -0,0 +1,967 @@
...
+#define STDERR_FILENO 2
+#endif
+    os_file_handle handle = BH_MALLOC(sizeof(struct zephyr_handle));
+    if (handle == NULL) {
+        return NULL;
+    }
+    struct zephyr_fs_desc *ptr = NULL;
+
+    /* We allocate a fake stdin reference */
+    ptr = zephyr_fs_alloc_obj(false, "stderr", &handle->fd);
```
wenyongh, line 839: same as above

core/shared/platform/zephyr/zephyr_file.c
```diff
@@ -0,0 +1,967 @@
...
+        ptr = &desc_array[(int)fd];         \
+        if (!ptr)                           \
+            return __WASI_EBADF;            \
+    } while (0)
+
+// Array to keep track of file system descriptors.
+static struct zephyr_fs_desc desc_array[CONFIG_WASI_MAX_OPEN_FILES];
+
+// mutex to protect the file descriptor array
+K_MUTEX_DEFINE(desc_array_mutex);
```
wenyongh, line 60: The mutex is only used in zephyr_fs_alloc_obj to protect desc_array, but desc_array is used in many places when GET_FILE_SYSTEM_DESCRIPTOR is called, seems we should also use the mutex to protect them?

core/shared/platform/zephyr/zephyr_socket.c
```diff
@@ -0,0 +1,1056 @@
...
+{
+    *sock = BH_MALLOC(sizeof(zephyr_handle));
+    if (!sock) {
+        return BHT_ERROR;
+    }
+
+    (*sock)->fd = zsock_accept(server_sock->fd, addr, addrlen);
+
+    if ((*sock)->fd < 0) {
+        return BHT_ERROR;
```
wenyongh, line 287: should free `*sock` before return

core/shared/platform/include/platform_wasi_types.h
```diff
@@ -173,7 +173,8 @@ typedef uint8_t __wasi_eventtype_t;
 
 typedef uint32_t __wasi_exitcode_t;
 
-typedef uint32_t __wasi_fd_t;
+// typedef uint32_t __wasi_fd_t;
+typedef int __wasi_fd_t;
```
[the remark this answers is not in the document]
lucasAbadFr: done

core/shared/platform/include/platform_wasi_types.h
```diff
@@ -539,7 +540,10 @@ typedef enum {
 
 typedef uint16_t __wasi_ip_port_t;
 
-typedef enum { IPv4 = 0, IPv6 } __wasi_addr_type_t;
+// Ensure that __wasi_addr_type_t has a size of 4 byte (I32).
+// However, it will not have the type safety of enum.
```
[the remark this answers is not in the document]
lucasAbadFr: done

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
lucasAbadFr: Not sure why I made this copy, restested in both case with and without copy and the original implementation worked.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
lucasAbadFr: same as above

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/ssp_config.h
lucasAbadFr: This was for readability, and you're right. I will go back to the original version.

core/shared/platform/common/posix/posix_socket.c
lucasAbadFr: Restored the original implementation.

core/shared/platform/common/posix/posix_socket.c
lucasAbadFr: I investigated this case and found something, I will open an issue. In any case I restored the original implementation.

core/shared/platform/linux/platform_internal.h
lucasAbadFr: removed L67 to L70

core/shared/platform/zephyr/shared_platform.cmake
```diff
@@ -8,11 +8,23 @@ add_definitions(-DBH_PLATFORM_ZEPHYR)
...
+    list(REMOVE_ITEM source_all ${PLATFORM_SHARED_DIR}/zephyr_file.c)
+    list(REMOVE_ITEM source_all ${PLATFORM_SHARED_DIR}/zephyr_clock.c)
+else()
+  include (${CMAKE_CURRENT_LIST_DIR}/../common/libc-util/platform_common_libc_util.cmake)
+  set(source_all ${source_all} ${PLATFORM_COMMON_LIBC_UTIL_SOURCE})
+endif ()
 
-set (PLATFORM_SHARED_SOURCE ${source_all} ${PLATFORM_COMMON_MATH_SOURCE})
+include (${CMAKE_CURRENT_LIST_DIR}/../common/memory/platform_api_memory.cmake)
+set (source_all ${source_all} ${PLATFORM_COMMON_MEMORY_SOURCE})
```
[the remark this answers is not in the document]
lucasAbadFr: You're right. I removed them, this also fixed the issue bellow

core/shared/platform/zephyr/zephyr_platform.c
lucasAbadFr: see above

core/shared/platform/zephyr/zephyr_file.c
lucasAbadFr: done

core/shared/platform/zephyr/zephyr_file.c
lucasAbadFr: I locked the access to the array in the GET_FILE_SYSTEM_DESCRIPTOR macro.

core/shared/platform/zephyr/zephyr_clock.c
lucasAbadFr: I will look into this.

core/shared/platform/zephyr/zephyr_file.c
lucasAbadFr: done

core/shared/platform/zephyr/zephyr_file.c
lucasAbadFr: done

core/shared/platform/zephyr/zephyr_file.c
lucasAbadFr: done

core/shared/platform/zephyr/zephyr_file.c
lucasAbadFr: I used `strncmp` instead.

core/shared/platform/zephyr/zephyr_file.c
lucasAbadFr: Here we can make two different choices:

1. more readable
```c
struct zephyr_fs_desc *ptr = NULL;

    /* We allocate a fake stdin reference */
    ptr = zephyr_fs_alloc_obj(false, "stdout", &handle->fd);
    if (ptr == NULL) {
        /* return an invalid handle instead of NULL */
        handle->is_sock = false;
        handle->fd = -1; // fd should already be -1
        return handle;
    }

    handle->is_sock = false;
    return handle;
```

2. less explicit
```c
    /* We allocate a fake stdin reference */
    zephyr_fs_alloc_obj(false, "stdout", &handle->fd);

    handle->is_sock = false;
    return handle;
```

Knowing that if `zephyr_fs_alloc_obj` fail `handle->fd` is set to `-1`. 

And the function to check validity is defined as:
```c
bool
os_is_handle_valid(os_file_handle *handle)
{
    return (*handle)->fd > -1;
}
```

I will mark as resolved or modify after hearing your opinion. I went with 2 but may be addind a comment will do the job.

product-mini/platforms/zephyr/simple-file/to_c_header.py
```diff
@@ -0,0 +1,23 @@
+# Python script to convert wasm file to byte array in a .h file
```
wenyongh, line 1: had better add license header

product-mini/platforms/zephyr/simple-file/wasm-apps/file.c
```diff
@@ -0,0 +1,50 @@
+#include <stdio.h>
```
wenyongh, line 1: had better add license header

product-mini/platforms/zephyr/simple-http/src/http_get.h
```diff
@@ -0,0 +1,3032 @@
+unsigned char __aligned(4) wasm_test_file[] = {
```
wenyongh, line 1: had better add license header

product-mini/platforms/zephyr/simple-http/to_c_header.py
```diff
@@ -0,0 +1,23 @@
+# Python script to convert wasm file to byte array in a .h file
```
wenyongh, line 1: had better add license header

product-mini/platforms/zephyr/simple-http/wasm-apps/http_get.c
```diff
@@ -0,0 +1,91 @@
+/*
+ * Copyright (c) 2017 Linaro Limited
+ *
+ * SPDX-License-Identifier: Apache-2.0
```
wenyongh, line 4: Is this file ported from other project? If yes, normally we add it into <wamr_root>/ATTRIBUTIONS.md, and add a license file in the related folder.
see:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/ATTRIBUTIONS.md

core/shared/platform/zephyr/shared_platform.cmake
```diff
@@ -1,5 +1,5 @@
-# Copyright (C) 2019 Intel Corporation.  All rights reserved.
-# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#Copyright(C) 2019 Intel Corporation.All rights reserved.
+#SPDX - License - Identifier : Apache - 2.0 WITH LLVM - exception
```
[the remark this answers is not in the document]
lucasAbadFr: Accidently ran clang-format on this file... I will fix that

product-mini/platforms/zephyr/simple-http/wasm-apps/http_get.c
lucasAbadFr: The sample is highly inspired from : https://github.com/zephyrproject-rtos/zephyr/blob/main/samples/net/sockets/http_get/src/http_get.c

Add I see that zephyr already appear in the [ATTRIBUTIONS.md](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/ATTRIBUTIONS.md)

What should we do ?

core/shared/platform/zephyr/zephyr_file.c
```diff
@@ -0,0 +1,975 @@
...
+        k_mutex_unlock(&desc_array_mutex);          \
+    } while (0)
+
+// Array to keep track of file system descriptors.
+static struct zephyr_fs_desc desc_array[CONFIG_WASI_MAX_OPEN_FILES];
+
+// mutex to protect the file descriptor array
+K_MUTEX_DEFINE(desc_array_mutex);
+
+static char prestat_dir[MAX_FILE_NAME + 1];
```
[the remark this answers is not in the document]
lucasAbadFr: For this value I don't think a mutex is necessary because we assign a value to `prestat_dir` just once in `os_open_preopendir`. The `os_open_preopendir` is called at runtime startup in  [`wasm_runtime_init_wasi`](https://github.com/bytecodealliance/wasm-micro-runtime/blob/67dce482018bc3103f02018d9682dcbb5154661f/core/iwasm/common/wasm_runtime_common.c#L3547) and never after. Then in the later function we just access it's value, never modifying it.

In my opinion, this is minimum support because we don't support: 
1. multiple path at startup (just last one will be taken)
2. mapped dirs

We just support the case where we have something like that: 
```c
    const char *dir_list[1] = {"/lfs"};

    wasm_runtime_set_wasi_args(wasm_module, dir_list, DIR_LIST_SIZE, NULL, 0,
                               NULL, 0, NULL, 0);
```
Or with linux it would be equialent to:
```
./iwasm --dir=/my/single/path module.wasm

File: core/shared/platform/zephyr/zephyr_file.c [hunk 33]
```diff
@@ -0,0 +1,975 @@
...
+    int socktype, rc;
+
+    if (!handle->is_sock) {
+        GET_FILE_SYSTEM_DESCRIPTOR(handle->fd, ptr);
+        // debug_zephyr_fs_desc(ptr);
+
+        /* We treat the case of std[in/out/err] */
+        if (ptr->path != NULL
+            && (!strcmp(ptr->path, "stdin") || !strcmp(ptr->path, "stdout")
+                || !strcmp(ptr->path, "stderr"))) {
```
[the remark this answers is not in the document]
lucasAbadFr: I'm not sure because `ptr->path` is assigned in :
* `zephyr_fs_alloc_obj` 
* `os_renameat` 

In my opinion, it is really unlikely that an `open` (call `zephyr_fs_alloc_obj` ) and `rename` will be called concurently and even if it does happen it should be the user responsabilityto put a synchronisation mechanism at a higher level on a `FILE *` or `DIR*` to avoid concurrent access.

If I'm not clear enough please tell me.

core/shared/platform/zephyr/shared_platform.cmake
```diff
@@ -8,11 +8,21 @@ add_definitions(-DBH_PLATFORM_ZEPHYR)
 include_directories(${PLATFORM_SHARED_DIR})
 include_directories(${PLATFORM_SHARED_DIR}/../include)
 
+file (GLOB_RECURSE source_all ${
+    PLATFORM_SHARED_DIR}/*.c)
```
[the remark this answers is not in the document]
lucasAbadFr: I saw that as well, I unintentionally ran clang-format on this file...

core/shared/platform/zephyr/zephyr_file.c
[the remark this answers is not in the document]
wenyongh: OK, thanks for the explanation, so it is set in the startup and isn't changed after that, we don't need to add lock when accessing it.

core/shared/platform/zephyr/zephyr_file.c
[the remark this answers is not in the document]
wenyongh: OK, so your opinion is to protect the slot of desc_array, but not to protect the accessing of the element in that slot, and protecting the latter is the user's responsibility. So let's leave it unchanged now, and check whether we need to do more in the future if there is issue reported?

core/shared/platform/zephyr/zephyr_file.c
wenyongh: But here os_convert_stdin_handle tries to return a converted handle, it can return NULL (when BH_MALLOC fails, it return NULL). My opinion is either to return a well-formed handle, either to return NULL, returning a mal-formed with `handle->fd == -1` is a little tricky, and it seems the caller won't free the handle when it calls os_is_handle_valid and finds it is an invalid handle? How about:

```C
os_file_handle
os_convert_stdin_handle(os_raw_file_handle raw_stdin)
{
#ifndef STDIN_FILENO
#define STDIN_FILENO 0
#endif
    os_file_handle handle = BH_MALLOC(sizeof(struct zephyr_handle));
    if (handle == NULL) {
        return NULL;
    }

    /* We allocate a fake stdin reference */
    if (zephyr_fs_alloc_obj(false, "stdin", &handle->fd) == NULL) {
        BH_FREE(handle);
        return NULL;
    }

    handle->is_sock = false;
    return handle;
}
```

core/shared/platform/zephyr/zephyr_file.c
```diff
@@ -0,0 +1,975 @@
...
+
+    return ptr;
+}
+
+static inline void
+zephyr_fs_free_obj(struct zephyr_fs_desc *ptr)
+{
+    free(ptr->path);
+    ptr->path = NULL;
+    ptr->used = false;
```
wenyongh, line 102: Seems that ptr is always a pointer pointing to an element of desc_array? Should also add `k_mutex_lock(&desc_array_mutex) ... k_mutex_unlock(&desc_array_mutex)` for these lines?

wenyongh: And another issue is that will all allocated elements of desc_array be freed when application exits (e.g. wasm_runtime_destroy() is called)? For example, wasm app opened some files and some elements of desc_array are used and their path fields are allocated, and wasm app doesn't close them before exit.

product-mini/platforms/zephyr/simple-http/src/main.c
```diff
@@ -0,0 +1,141 @@
...
+                                          error_buf, sizeof(error_buf)))) {
+        printf("Failed to load module: %s\n", error_buf);
+        goto fail1;
+    }
+
+    /* Set the WASI context */
+#if WASM_ENABLE_LIBC_WASI != 0
+#define ADDRESS_POOL_SIZE 1
+    const char *addr_pool[ADDRESS_POOL_SIZE] = {
+        "192.0.2.10/24",
```
jachstet-sea, line 82: The fact that there is an address pool that acts as a white list should be mentioned in the Readme. Like a section `Customize the example to your environment` that lists all places that need to be touched if the board's and the server's IP need to be modified

product-mini/platforms/zephyr/simple-http/wasm-apps/http_get.c
```diff
@@ -0,0 +1,91 @@
...
+    int rc = 0;
+
+    printf("[wasm-mod] Preparing HTTP GET request for http://" HTTP_HOST
+           ":" HTTP_PORT HTTP_PATH "\n");
+
+    memset(&addr, 0, sizeof(addr));
+    addr.sin_family = AF_INET;
+    addr.sin_port = htons(8000);
+    addr.sin_addr.s_addr =
+        htonl(3221225994); // hard coded IP address for 192.0.2.10
```
jachstet-sea, line 49: I propose to use the hex notation here instead: `htonl(0xC000020A)`. It works the same but is easier to read and to modify

product-mini/platforms/zephyr/simple-http/wasm-apps/http_get.c
```diff
@@ -0,0 +1,91 @@
...
+#include <unistd.h>
+#include <errno.h>
+#include <string.h>
+
+#ifdef __wasi__
+#include <wasi_socket_ext.h>
+#endif
+
+/* HTTP server to connect to */
+#define HTTP_HOST "192.0.2.10"
```
jachstet-sea, line 21: This might be confusing since this constant is used in the LOG statement (line 42) and the actual request (line 31) but the actual IP to connect to is not derived from this string (see line 49).
In order to modify this example and to change from which HTTP server to read, it is not clear which lines need adaptation.
Same comment for PORT since the actual port used to connect is found in line 47

core/shared/platform/zephyr/zephyr_file.c
```diff
@@ -0,0 +1,975 @@
...
+}
+
+__wasi_errno_t
+os_file_get_fdflags(os_file_handle handle, __wasi_fdflags_t *flags)
+{
+    struct zephyr_fs_desc *ptr = NULL;
+
+    GET_FILE_SYSTEM_DESCRIPTOR(handle->fd, ptr);
+
+    if ((ptr->file.flags & FS_O_APPEND) != 0) {
```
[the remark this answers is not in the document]
lucasAbadFr: IMO, it is the same case as `ptr->path`

core/shared/platform/zephyr/zephyr_file.c
```diff
@@ -0,0 +1,975 @@
...
+    int rc;
+    struct zephyr_fs_desc *ptr = NULL;
+
+    if (is_stdio)
+        return __WASI_ESUCCESS;
+
+    GET_FILE_SYSTEM_DESCRIPTOR(handle->fd, ptr);
+
+    rc = ptr->is_dir ? fs_closedir(&ptr->dir) : fs_close(&ptr->file);
+    zephyr_fs_free_obj(ptr); // free in any case.
```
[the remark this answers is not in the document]
lucasAbadFr: Sure, which approach would you take ? 

```c
static inline void
zephyr_fs_free_obj(struct zephyr_fs_desc *ptr)
{
    if (ptr != NULL) {
        free(ptr->path);
        ptr->path = NULL;
        ptr->is_dir = false;
        ptr->used = false;
    }
}
```

Or 

```c
static inline void
zephyr_fs_free_obj(struct zephyr_fs_desc *ptr)
{
    if (ptr != NULL) {
        free(ptr->path);
        memset(ptr, 0, sizeof(struct zephyr_fs_desc));
    }
}
```

Not sure if there is a bh_ variant for memset , in any case zephyr implement it.

core/shared/platform/zephyr/zephyr_file.c
lucasAbadFr: True, I didn't think about this case.

If we create a function to cleanup the `desc_array`, then we will have to call it in  either `wasm_runtime_destroy` or `wasm_runtime_destroy_wasi` .

But where should we declare such a function that is platform specific ? 

The simple approach would be to put it in the `platform_api_extension.h` and have it locked by `#ifdef BH_PLATFORM_ZEPHYR`. 

I don't think that we want platform  specific code in the plateform API. We could create a macro like `BH_PLATFORM_NEED_WASI_CLEANUP` and then use it to lock with `#ifdef` an `os_cleanup_wasi` or something along those lines.

product-mini/platforms/zephyr/simple-http/src/main.c
lucasAbadFr: added a configuration section in readme.

product-mini/platforms/zephyr/simple-http/wasm-apps/http_get.c
lucasAbadFr: done

product-mini/platforms/zephyr/simple-http/wasm-apps/http_get.c
lucasAbadFr: added a configuration section in readme.

product-mini/platforms/zephyr/simple-http/CMakeLists.txt
```diff
@@ -0,0 +1,90 @@
...
+
+# WAMR Configuration:
+set (WAMR_BUILD_TARGET "THUMB")
+set (WAMR_BUILD_INTERP 1)
+set (WAMR_BUILD_AOT 0)
+set (WAMR_BUILD_LIBC_BUILTIN 1) # printf
+set (WAMR_BUILD_LIBC_WASI 1)
+set (WAMR_BUILD_LIB_PTHREAD 0)
+set (WAMR_BUILD_GLOBAL_HEAP_POOL 1)
+set (WAMR_BUILD_GLOBAL_HEAP_SIZE 131072) # 128 KB
```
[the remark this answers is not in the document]
lucasAbadFr: done

core/shared/platform/zephyr/zephyr_clock.c
lucasAbadFr: I proposed some really basic implementation.

core/shared/platform/zephyr/zephyr_file.c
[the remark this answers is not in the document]
wenyongh: No bh_memset like API, please just use memset.

core/shared/platform/zephyr/zephyr_file.c
wenyongh: BTW, could you add `k_mutex_lock(&desc_array_mutex) ... k_mutex_unlock(&desc_array_mutex)` for these lines since `ptr->used` is set to false without protection?

samples/socket-api/file.wasm
wenyongh, line 1: Same as above, could we remove this wasm file?

core/shared/platform/zephyr/zephyr_file.c
```diff
@@ -0,0 +1,990 @@
...
+    snprintf(abs_path, MAX_FILE_NAME, "%s/%s", prestat_dir, path);
+
+    rc = fs_mkdir(abs_path);
+    if (rc < 0) {
+        return convert_errno(-rc);
+    }
+
+    ptr = zephyr_fs_alloc_obj(true, abs_path, &index);
+    if (!ptr) {
+        return __WASI_EMFILE;
```
wenyongh, line 689: should we remove the dir (e.g., fs_unlink) before return error?

core/shared/platform/zephyr/zephyr_file.c
```diff
@@ -0,0 +1,990 @@
...
+    snprintf(abs_new_path, MAX_FILE_NAME, "%s/%s", prestat_dir, new_path);
+
+    int rc = fs_rename(abs_old_path, abs_new_path);
+    if (rc < 0) {
+        return convert_errno(-rc);
+    }
+
+    GET_FILE_SYSTEM_DESCRIPTOR(old_handle->fd, ptr);
+
+    ptr->path = strdup(new_path);
```
wenyongh, line 723: Should check the return value, and should rename the file to old path if strdup failed?

core/shared/platform/zephyr/zephyr_file.c
```diff
@@ -0,0 +1,990 @@
...
+os_isatty(os_file_handle handle)
+{
+    return __WASI_ENOSYS;
+}
+
+os_file_handle
+os_convert_stdin_handle(os_raw_file_handle raw_stdin)
+{
+#ifndef STDIN_FILENO
+#define STDIN_FILENO 0
```
wenyongh, line 806: Seems that STDIN_FILENO/STDOUT_FILENO/STDERR_FILENO are not used? Is it better to remove them?

core/shared/platform/zephyr/zephyr_socket.c
```diff
@@ -0,0 +1,1049 @@
...
+        (*sock)->fd = zsock_socket(af, SOCK_STREAM, IPPROTO_TCP);
+    }
+    else {
+        (*sock)->fd =
+            zsock_socket(af, SOCK_DGRAM, IPPROTO_UDP); // IPPROTO_UDP or 0 ?
+    }
+
+    (*sock)->is_sock = true;
+
+    return ((*sock)->fd == -1) ? BHT_ERROR : BHT_OK;
```
wenyongh, line 207: Should BH_FREE(*sock) if return BH_ERROR since the caller may not free it?
```C
    if ((*sock)->fd == -1) {
        BH_FREE(*sock);
        return BHT_ERROR;
    }
    return BHT_OK;
}
```

core/shared/platform/zephyr/zephyr_socket.c
```diff
@@ -0,0 +1,1049 @@
...
+    struct sockaddr_storage addr = { 0 };
+    socklen_t socklen;
+    int ret;
+
+    assert(host);
+    assert(port);
+
+    if (!textual_addr_to_sockaddr(host, *port, (struct sockaddr *)&addr,
+                                  &socklen)) {
+        goto fail;
```
wenyongh, line 222: os_socket_close is called in the fail label, but zsock_bind hasn't been called, is better to return BHT_ERROR here?

core/shared/platform/zephyr/zephyr_socket.c
```diff
@@ -0,0 +1,1049 @@
...
+    }
+
+    // F_SETF_SETFD and FD_CLOEXEC are not defined in zephyr.
+    // SO_LINGER: Socket lingers on close (ignored, for compatibility)
+
+    ret = zsock_bind(socket->fd, (struct sockaddr *)&addr, socklen);
+    if (ret < 0) {
+        goto fail;
```
wenyongh, line 230: Same as above, is better to return BHT_ERROR here?

core/shared/platform/zephyr/zephyr_socket.c
```diff
@@ -0,0 +1,1049 @@
...
+#else
+        goto fail;
+#endif
+    }
+
+    return BHT_OK;
+
+fail:
+    // Close the fd because FD_CLOEXEC isn't implemented.
+    os_socket_close(socket);
```
wenyongh, line 253: os_socket_close will close the `socket->fd` and BH_FREE(socket), but they are not opened and allocated in this function, seems here we had better just call `os_socket_shutdown` and return BHT_ERROR?

core/shared/platform/zephyr/zephyr_socket.c
```diff
@@ -0,0 +1,1049 @@
...
+}
+
+__wasi_errno_t
+os_socket_shutdown(bh_socket_t socket)
+{
+    if (zsock_shutdown(socket->fd, ZSOCK_SHUT_RDWR) == -1) {
+        return convert_errno(errno);
+    }
+    return __WASI_ESUCCESS;
+    ;
```
wenyongh, line 388: had better remove the unused line

product-mini/platforms/zephyr/simple-http/wasm-apps/http_get.wasm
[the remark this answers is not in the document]
lucasAbadFr: totally !

core/shared/platform/zephyr/zephyr_file.c
lucasAbadFr: Yes we should. Fixed it.

core/shared/platform/zephyr/zephyr_file.c
lucasAbadFr: yes it is. Done

core/shared/platform/zephyr/zephyr_socket.c
lucasAbadFr: Yes. Done

core/shared/platform/zephyr/zephyr_socket.c
lucasAbadFr: Leaving `os_socket_close` here is an error. 

IMO, even placing `os_socket_shutdown` is not necessary because when `bind` fail there is no such thing as shutdown. It should be the user responsability in higher level code to test `bind` return value and handle it accordingly (either `close` or `shutdown`).

eg:
```c
int result = bind(socket, server_addr, addr_len);
if (result == -1)
{
    // handle failure
    close(socket); 
    // or shutdown(socket); 
}
```

core/shared/platform/zephyr/zephyr_file.c
lucasAbadFr: Thanks for the feedbacks, I will look into this.

samples/socket-api/file.wasm
wenyongh, line 1: could you also remove this file?

core/shared/platform/zephyr/zephyr_file.c
```diff
@@ -0,0 +1,986 @@
...
+    int rc = fs_rename(abs_old_path, abs_new_path);
+    if (rc < 0) {
+        return convert_errno(-rc);
+    }
+
+    GET_FILE_SYSTEM_DESCRIPTOR(old_handle->fd, ptr);
+
+    ptr->path = strdup(new_path);
+    if (ptr->path == NULL) {
+        ptr->path = old_path;
```
wenyongh, line 726: had better also `fs_rename(abs_new_path, abs_old_path)`?

core/shared/platform/zephyr/zephyr_socket.c
wenyongh: OK, thanks for the correction.

core/shared/platform/zephyr/zephyr_file.c
lucasAbadFr: Yes, we should. 

Then here, if the (new) `fs_rename` fail what should we do ? This will put the app in a weird state.

Should we retry: 
```c
do {
    rc = fs_rename(abs_new_path, abs_old_path);
    retry_count++;
} while (rc < 0 && retry_count < 5);

if (rc < 0) {
    // Error: failed to rename back
    return __WASI_EIO;
}
```

The do while is just to illustrate we could use a for then if and break.

wenyongh: Is it better to strdup first and then fs_rename? When fs_rename failed, we can just free the string allocated by strdup.

wenyongh: And also here had better GET_FILE_SYSTEM_DESCRIPTOR first, and then strdup, then fs_rename? Since the function may return failure inside the macro, and fs_rename and free are not called before return.

wenyongh: Another issue is that we should also free the original `ptr->path` but not set it directly. Do you think the below like code is better?

```C
    GET_FILE_SYSTEM_DESCRIPTOR(old_handle->fd, ptr);

    char *path = strdup(new_path);
    if (path == NULL) {
        return __WASI_ENOMEM;
    }

    int rc = fs_rename(abs_old_path, abs_new_path);
    if (rc < 0) {
        free(path);
        return convert_errno(-rc);
    }

    free(ptr->path);
    ptr->path = path;
    return __WASI_ESUCCESS;
```

wenyongh:
> I am working on getting WASI support working on Zephyr and came across this thread. Are there any plans to complete and merge this PR? I'm happy to contribute, just wanted to see if I should start form here or the current trunk.

I am not sure whether @lucasAbadFr has bandwidth to finish it, in fact most of the comments were resolved exception several ones. Maybe we can merge it into another branch (e.g. dev/zephyr_file_socket) and let @srberard continue the work? And when the feature is done, we can merge the commits into the main branch. @lucasAbadFr how do you think?

srberard:
@wenyongh that works for me.

wenyongh:
> @wenyongh that works for me.

Great! I think @lucasAbadFr is busy and may not be able to response quickly, so I merged this PR to branch `dev/zephyr_file_socket`, please help address the comments left.

@lucasAbadFr Many thanks for your contribution! Please let us know if you have any concerns.

srberard:
Excellent.  I will start on this later this week and update this thread.

lucasAbadFr:
Hi @wenyongh and @srberard
As you probably guessed I don't have the bandwith to work on this PR.
If there is any questions or comments I will do my best to respond in time.

lucasAbadFr:
>@lucasAbadFr Many thanks for your contribution! Please let us know if you have any concerns.

You're welcome, and many thanks for all the time taken during reviews.

PR #3638, opened 2024-07-17
WIP xtensa CI bits

yamt:


HongxiaWangSSSS:
Hi, @yamt-san.
I tested the xtensa spec test on your test framework. (only XIP mode). The results are basically consistent with my previous local test results. Some cases related to memory.grow need to be skipped.
I will submit a PR later~
Very thanks for your contribute, its very useful and saves a lot of effort on xtensa test~(*^▽^*)

yamt:
> Hi, @yamt-san. I tested the xtensa spec test on your test framework. (only XIP mode). The results are basically consistent with my previous local test results. Some cases related to memory.grow need to be skipped. I will submit a PR later~ Very thanks for your contribute, its very useful and saves a lot of effort on xtensa test~(_^▽^_)

thank you for testing.
i plan to enable xip once the non-xip case gets stable.

yamt:
this PR is now mostly empty

PR #3639, opened 2024-07-17
Add APIs into wasm_c_api.h to summary wasm function execution duration

lum1n0us:
- `wasm_instance_sum_wasm_exec_time()` -> `wasm_runtime_sum_wasm_exec_time()`
- `wasm_instance_get_wasm_func_exec_time()` -> `wasm_runtime_get_wasm_func_exec_time()`

### DIFF
--- a/core/iwasm/include/wasm_c_api.h
+++ b/core/iwasm/include/wasm_c_api.h
@@ -691,6 +691,11 @@ WASM_API_EXTERN own wasm_instance_t* wasm_instance_new_with_args_ex(
 
 WASM_API_EXTERN void wasm_instance_exports(const wasm_instance_t*, own wasm_extern_vec_t* out);
 
+// Return total wasm functions' execution time in ms
+WASM_API_EXTERN double wasm_instance_sum_wasm_exec_time(const wasm_instance_t*);
+// Return execution time in ms of a given wasm function with
+// func_name. If the function is not found, return 0.
+WASM_API_EXTERN double wasm_instance_get_wasm_func_exec_time(const wasm_instance_t*, const char *);
 
 ///////////////////////////////////////////////////////////////////////////////
 // Convenience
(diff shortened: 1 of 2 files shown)


core/iwasm/include/wasm_c_api.h
```diff
@@ -691,6 +691,11 @@ WASM_API_EXTERN own wasm_instance_t* wasm_instance_new_with_args_ex(
 
 WASM_API_EXTERN void wasm_instance_exports(const wasm_instance_t*, own wasm_extern_vec_t* out);
 
+// Return total wasm functions' execution time in ms
+WASM_API_EXTERN double wasm_instance_sum_wasm_exec_time(const wasm_instance_t*);
+// Return execution time in ms of a given wasm function with
+// func_name. If the function is not found, return 0.
+WASM_API_EXTERN double wasm_instance_get_wasm_func_exec_time(const wasm_instance_t*, const char *);
```
yamt, line 698: it's a bit strange to expose these function names in the api.
what kind of program uses these api?

lum1n0us: They are used in a cloud platform as a part of observation features for Wasm plugins. 

WAMR used to provide both APIs in [*wasm_export.h*](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/include/wasm_export.h#L1662-L1680) only. But this customer's host program only uses *wasm_c_api.h*. So, here we are.

yamt: i didn't know it's already exported by wasm_export.h. thank you.

how is the embedder supposed to know the function name to specify?
it's sometimes import name, sometimes export name, and sometimes the name from name section.
as they are not unique, i suspect a user can't always specify a function reliably.

lum1n0us: we are using *import name* and *name from the name section*. usually, users will pass names in the guest language. Since for a wasm32-wasi, all *imports* are from *wasi-libc*, if users can make sure their code are valid in the guest language, there will be no duplicated name.

wenyongh (commented):
LGTM

PR #3640, opened 2024-07-17
Fix nightly run CI

TianlongLiang:
Manually git clone and install dependencies instead of using git action on Ubuntu 14.04 runner

### DIFF
--- a/.github/workflows/nightly_run.yml
+++ b/.github/workflows/nightly_run.yml
@@ -323,16 +323,13 @@ jobs:
           - make_options_run_mode: $FAST_JIT_BUILD_OPTIONS
             make_options_feature: "-DWAMR_BUILD_MEMORY64=1"
     steps:
-      - name: checkout
-        uses: actions/checkout@v3
-
       - name: Install dependencies
-        uses: nick-fields/retry@v2
-        with:
-          timeout_minutes: 10
-          max_attempts: 3
-          command: apt update && apt install -y make g++-4.8 gcc-4.8 wget git
-          on_retry_command: sudo rm -r /var/lib/apt/lists/*
+        run: |
+          apt update && apt install -y make g++-4.8 gcc-4.8 wget git
+
+      - name: checkout
+        run: |
+          git clone https://github.com/${{ github.repository }} wamr
 
       - name: Install cmake
         run: |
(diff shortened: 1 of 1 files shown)


TianlongLiang:
As mentioned in [this blog](https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/), GitHub actions will default use node20, and we are using GH provided runner, we have no control over the external node version, temporarily replacing the GitHub action in `build_iwasm_linux_gcc4_8` with normal commands

TianlongLiang:
@loganek @yamt what do you guys think?

.github/workflows/nightly_run.yml
```diff
@@ -323,16 +323,13 @@ jobs:
           - make_options_run_mode: $FAST_JIT_BUILD_OPTIONS
             make_options_feature: "-DWAMR_BUILD_MEMORY64=1"
     steps:
-      - name: checkout
-        uses: actions/checkout@v3
```
yamt, line 327: can't you just bump to actions/checkout@v4?

.github/workflows/nightly_run.yml
```diff
@@ -323,16 +323,13 @@ jobs:
...
-
       - name: Install dependencies
-        uses: nick-fields/retry@v2
```
yamt, line 330: the latest one seems to claim node 20 compat.
https://github.com/nick-fields/retry/releases/tag/v3.0.0

btw, is this retry necessary at all these days?

.github/workflows/nightly_run.yml
TianlongLiang: The v4 will face the same issue, node 20 will require a more recent version of glibc and glibcxx than the Ubuntu 14.04 docker image provides

.github/workflows/nightly_run.yml
TianlongLiang: I think it should be fine without it

.github/workflows/nightly_run.yml
yamt: ah, i totally misread the PR description. sorry for noise!

yamt (approved):
looks reasonable to me

.github/workflows/nightly_run.yml
TianlongLiang: No worries, I also made some modifications to my previous comments to make it clearer

wenyongh (approved):

wenyongh:
I merged the PR since we had better let nightly-run CI run every day. @loganek please upload another PR to enhance it if needed, thanks.

PR #3643, opened 2024-07-18
spec-test-script: add xtensa case

yamt:


tests/wamr-test-suites/spec-test-script/runtest.py
```diff
@@ -1143,6 +1143,9 @@ def compile_wasm_to_aot(wasm_tempfile, aot_tempfile, runner, opts, r, output = '
     if test_target.startswith("riscv64"):
         cmd.append("--size-level=1")
 
+    if test_target.startswith("xtensa"):
```
no1wudi, line 1146: Why not using common target to option mapping here?
https://github.com/bytecodealliance/wasm-micro-runtime/blob/7c9686df5fa751f7614b4efacdcd0f3e7c22b825/tests/wamr-test-suites/spec-test-script/runtest.py#L45

yamt: because i didn't noticed it!

yamt: fixed. thank you.

no1wudi (approved):

PR #3645, opened 2024-07-19
CMake flag improvements

bnason-nf:


lum1n0us:
Just my curious, since they are global compile options(for all targets), why not use `add_compile_options()`(once) instead of `target_compile_options()`(multiple times)

bnason-nf:
Good question. Before these changes, the CMake file was modifying the options globally for all C and C++ code. I'm not completely sure whether that was intentional and/or necessary for C, but from what I have seen it does seem to be necessary for C++ due to some of the sub-libraries that use C++. If we think it's also necessary to globally change the C options then I can make those options global as well. If not, I think using target_compile_options() is cleaner (even with the duplication) and follows the existing style more.

bnason-nf:
To be clear, I don't have strong opinions on this, so just let me know if you feel I should abandon this change.

wenyongh:
@bnason-nf Not very sure what is the improvement of this PR? My first impression is that it is a little complex and makes the script not easy to understand - it introduces WAMR_COMPILE_FLAGS, WAMR_C_COMPILE_FLAGS, WAMR_COMPILE_FLAGS_STRING, etc. which makes me very confusing😊

bnason-nf:
Thanks for the feedback. Since this PR seems unpopular, I am going to close it. I will re-submit the commit for hidden visibility in a new PR.

PR #3647, opened 2024-07-19
spec-test-script/runtest.py: use a shorter timeout when expected to fail

yamt:
this is a band-aid fix; ideally we should wait for the expected failure message directly with a timeout, not the successful prompt as we currently do.


wenyongh (approved):

TianlongLiang:
LGTM

PR #3650, opened 2024-07-22
Zephyr User Mode Support

mxsdlr:
Add support for the [Zephyr Usermode](https://docs.zephyrproject.org/latest/kernel/usermode/index.html) to the Zephyr port.

The following changes are applied, if this PR gets accepcted:

- Fix `signbit`, check if it is defined already and only implement it, if not
- Introduce `sys_mutex` and `sys_sem` in favour of `k_mutex` and `k_sem`, when `CONFIG_USERMODE` is enabled
- Remove the installation of the `_stdout_hook_iwasm()` when `CONFIG_USERMODE` is enabled, otherwise this causes MPU errors since the std hook is in the kernel space
- Add a thread name for debugging


core/shared/platform/zephyr/platform_internal.h
```diff
@@ -1,6 +1,8 @@
 /*
  * Copyright (C) 2019 Intel Corporation.  All rights reserved.
  * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ *
+ * SPDX-FileCopyrightText: 2024 Siemens AG (For Zephyr usermode changes)
```
wenyongh, line 5: If using a new license, had better add the license file under current folder, and then add the related info in `<wamr_root>/ATTRIBUTIONS.md` (refer to other components and licenses), could you help do that?

core/shared/platform/zephyr/zephyr_platform.c
```diff
@@ -1,10 +1,13 @@
 /*
  * Copyright (C) 2019 Intel Corporation.  All rights reserved.
  * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ *
+ * SPDX-FileCopyrightText: 2024 Siemens AG (For Zephyr usermode changes)
  */
 
 #include "platform_api_vmcore.h"
 #include "platform_api_extension.h"
+#include "bh_log.h"
```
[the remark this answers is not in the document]
mxsdlr: Thank you for paying attention. I removed this line, you are right, it's not needed.

core/shared/platform/zephyr/platform_internal.h
mxsdlr: I'm currently working on this as I have to check for internal regulations. I will post any updates as soon as I know something. Thank you for pointing out the `ATTRIBUTIONS.md`

wenyongh, line 5: @mxsdlr could you move this line to be after L2, like the files below?

wenyongh (approved):
LGTM

PR #3654, opened 2024-07-22
execute_func: Newline only for non-empty returns

wallento:
For void functions, the return value array is empty. Don't print the newline then.


yamt:
doesn't this confuse programs parsing the output, like our test scripts?

wallento:
The failing checks all seem timeouts, so I think no test checks for this.

yamt:
> The failing checks all seem timeouts, so I think no test checks for this.

i guess a test timeout is actually a typical symptom of unexpected output.
i can't say if they are related to your change or not w/o investigating the test logs.
(sorry, right now i have no time to investigate them by myself.)

PR #3657, opened 2024-07-22
Set posix thread name

bnason-nf:


yamt:
if you want to name internal threads, i guess it would be more useful and flexible to make the thread functions set their names by themselves.

bnason-nf:
What I'm seeing right now is that all the WAMR threads inherit the name of the thread they were spawned from. This is confusing when debugging, so my goal with this change is to avoid that specific confusion. Do we want to do more than that? I see WAMR using threads for the public `wasm_runtime_spawn_thread()` API, for JIT, for debugging, and for `thread-mgr`. Are you suggesting that we add a public API to set the thread name (or modify the existing API), and to give the JIT/debug/thread-mgr distinctive names?

wenyongh:
@bnason-nf seems that you are setting the thread name for debug purpose, how about wrapping the code with BH_DEBUG macro?
```C
#if BH_DEBUG != 0
#if defined __APPLE__
    pthread_setname_np("wamr");
#else
    pthread_setname_np(pthread_self(), "wamr");
#endif
#endif
```
@yamt how do you think?

bnason-nf:
> @bnason-nf seems that you are setting the thread name for debug purpose, how about wrapping the code with BH_DEBUG macro?

Done.

wenyongh (commented):
LGTM

TianlongLiang:
LGTM

yamt:
> Are you suggesting that we add a public API to set the thread name (or modify the existing API), and to give the JIT/debug/thread-mgr distinctive names?

yes.

yamt:
> @bnason-nf seems that you are setting the thread name for debug purpose, how about wrapping the code with BH_DEBUG macro?
> 
> ```c
> #if BH_DEBUG != 0
> #if defined __APPLE__
>     pthread_setname_np("wamr");
> #else
>     pthread_setname_np(pthread_self(), "wamr");
> #endif
> #endif
> ```
> 
> @yamt how do you think?

i have no opinion on whether it should be conditional on BH_DEBUG or not.

wenyongh:
> > @bnason-nf seems that you are setting the thread name for debug purpose, how about wrapping the code with BH_DEBUG macro?
> > ```c
> > #if BH_DEBUG != 0
> > #if defined __APPLE__
> >     pthread_setname_np("wamr");
> > #else
> >     pthread_setname_np(pthread_self(), "wamr");
> > #endif
> > #endif
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @yamt how do you think?
> 
> i have no opinion on whether it should be conditional on BH_DEBUG or not.

OK, maybe we can merge this PR to meet the debug requirement first and then add a public API to set the thread name in the future? @bnason-nf what's your opinion?

bnason-nf:
That sounds fine to me.

PR #3658, opened 2024-07-23
wamrc: Add --mllvm= option

yamt:
this allows users to specify llvm command line options similarly to clang's -mllvm option.

my motivations:
* -debug and friends
* -mtext-section-literals for xtensa

### DIFF
--- a/wamr-compiler/main.c
+++ b/wamr-compiler/main.c
@@ -9,6 +9,8 @@
 #include "wasm_export.h"
 #include "aot_export.h"
 
+#include <llvm-c/Support.h>
+
 #if BH_HAS_DLFCN
 #include <dlfcn.h>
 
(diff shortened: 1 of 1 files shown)


wenyongh (commented):
LGTM

wamr-compiler/main.c
```diff
@@ -738,6 +765,7 @@ main(int argc, char *argv[])
     if (option.custom_sections) {
         free(option.custom_sections);
     }
+    free(llvm_options);
```
wenyongh, line 768: Is it better to check llvm_options and free it only it is not NULL? like L765

yamt: why?

wenyongh: When `--mllvm=xxx` isn't specified, llvm_options is NULL, no need to free it in fact. Though the linux manual of free says no operation is performed if ptr is NULL, I am not sure whether it is a good coding style, we usually free a buffer only when it is NULL. It is good if the same behavior is ensured in other platforms.

yamt: free(NULL) is guaranteed to be a no-op by C language.
i'm not aware of any problematic platforms wrt that.

wenyongh: OK, thanks.

Issue #3608, opened 2024-07-09
CI doesn't seem to cover xtensa at all


yamt:
while we have a considerable amount of xtensa-specific code, the CI doesn't seem to cover it at all.
it would be nice to have basic tests at least.

an obstacle is that we use a different version of llvm for xtensa.

HongxiaWangSSSS:
Hi, @yamt -san

I'd like to know do you have any plan to support xtensa platform based on the current CI.
Currently QEMU supports [esp32s3](https://github.com/espressif/qemu/releases/tag/esp-develop-9.0.0-20240606), and may can help support aot and wasm testing (I haven't tried it yet), But it seems that XIP aot cannot be run either.
We are also very interested in the standard test of this platform, especially XIP~

BR,
Hongxia

yamt:
> Hi, @yamt -san
> 
> I'd like to know do you have any plan to support xtensa platform based on the current CI. Currently QEMU supports [esp32s3](https://github.com/espressif/qemu/releases/tag/esp-develop-9.0.0-20240606), and may can help support aot and wasm testing (I haven't tried it yet), But it seems that XIP aot cannot be run either. We are also very interested in the standard test of this platform, especially XIP~
> 
> BR, Hongxia

i was able to run (at least some parts of) the spec test with test_wamr.sh with esp32s3 qemu locally while ago.
with manually compiled wamrc and nuttx image.
i guess i will submit a few patches sooner or later.

i haven't done anything wrt the "build llvm+wamrc for xtena on the CI" part.

Issue #3646, opened 2024-07-19
spec_test_on_nuttx.yml should not edit kconfig files with sed


yamt:
because it's fragile.
there are more appropriate tools to modify configs.
(eg. kconfig-tweak)

PR #3665, opened 2024-07-25
spec_test_on_nuttx.yml: Add xtensa

yamt:
Fixes: https://github.com/bytecodealliance/wasm-micro-runtime/issues/3608


.github/workflows/spec_test_on_nuttx.yml
```diff
@@ -199,22 +236,91 @@ jobs:
...
-          cd nuttx
           tools/configure.sh ${{ matrix.target_config.config }}
+        working-directory: nuttx
+
+      # depending on configurations, the iwasm command line generated
+      # by spec-test-script can be longer than the default NSH_LINELEN,
+      # which is 64 or 80.
+      - name: Tweak NuttX config
+        run: |
+          kconfig-tweak --set-val CONFIG_NSH_LINELEN 255
```
yamt, line 275: i plan to replace the other sed usage in this file with kconfig-tweak.
but not in this PR.

.github/workflows/spec_test_on_nuttx.yml
```diff
@@ -86,6 +88,11 @@ jobs:
...
             fpu_type: "none"
           },
+          {
+            config: "boards/xtensa/esp32s3/esp32s3-devkit/configs/qemu_debug",
+            target: "xtensa",
+            fpu_type: "none"
+          },
         ]
 
         wamr_test_option: [
```
yamt, line 98: i plan to enable xip later.
but not in this PR.

.github/workflows/spec_test_on_nuttx.yml
yamt: a reminder: https://github.com/bytecodealliance/wasm-micro-runtime/issues/3646

.github/workflows/spec_test_on_nuttx.yml
```diff
@@ -179,6 +197,25 @@ jobs:
...
+        if: matrix.target_config.target != 'xtensa'
+        run: |
+          find nuttx/boards -name defconfig | xargs sed -i '$aCONFIG_INTERPRETERS_WAMR_STACKSIZE=327680\n'
+
+      # because qemu doesn't have a proper emulation of esp32s3 psram,
+      # we are limited to the internal ram, which is about 400KB.
+      - name: Set WAMR stack size for NuttX (xtensa)
+        if: matrix.target_config.target == 'xtensa'
+        run: |
+          find nuttx/boards -name defconfig | xargs sed -i '$aCONFIG_INTERPRETERS_WAMR_STACKSIZE=25600\n'
```
no1wudi, line 211: How about use it for all arch? It's more close to real hardware if it can pass all test

yamt: i don't know if this size is ok for other archs.
i guess it isn't, because some of gc/tali-call tests, which is disabled for xtensa, tend to consume more memory.

no1wudi (approved):

PR #3669, opened 2024-07-25
Add emscripten_sleep() wrapper to libc-emcc

bnason-nf:


bnason-nf:
There was a failure in one of the checks that seems unrelated to the change.

core/iwasm/libraries/libc-emcc/libc_emcc_wrapper.c
```diff
@@ -495,6 +495,12 @@ emscripten_notify_memory_growth_wrapper(wasm_exec_env_t exec_env, int i)
     (void)i;
 }
 
+static void
+emscripten_sleep_wrapper(wasm_exec_env_t exec_env, int timeout_ms)
```
wenyongh, line 499: Is the time unit ms? `man 3 sleep` in linux shows:
```C
NAME
       sleep - sleep for a specified number of seconds

SYNOPSIS
       #include <unistd.h>

       unsigned int sleep(unsigned int seconds);
```

core/iwasm/libraries/libc-emcc/libc_emcc_wrapper.c
```diff
@@ -495,6 +495,12 @@ emscripten_notify_memory_growth_wrapper(wasm_exec_env_t exec_env, int i)
...
+{
+    usleep(timeout_ms * 1000);
```
wenyongh, line 501: integer overflow may occur in `timeout_ms * 1000`, had better check it?
```C
    if (timeout > UINT32_MAX /1000)
        usleep(UINT32_MAX);
    else
        usleep(timeout_ms * 1000);
```

core/iwasm/libraries/libc-emcc/libc_emcc_wrapper.c
bnason-nf: Yes this is confusing, but I believe the implementation is correct. Please see the emscripten documention here:

https://emscripten.org/docs/api_reference/emscripten.h.html#sleeping

core/iwasm/libraries/libc-emcc/libc_emcc_wrapper.c
bnason-nf: Thanks, your comments made me realize that the implementation needs to be more robust in several ways, so I've rewritten it.

bnason-nf: Sorry about that, it's pushed now.

wenyongh (approved):
LGTM

TianlongLiang:
LGTM

Issue #2274, opened 2023-06-07
aarch64 AOT module fails to load on android


zoraaver:
When trying to load an aarch64 AOT file on android, the module fails to load due to an [overflow check](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/aot/arch/aot_reloc_aarch64.c#L298-L299) failing, `AOT module load failed: target address out of range.`.

## Expected behaviour
aarch64 AOT module loads without error
## Actual behaviour
aarch64 AOT module fails to load
## Steps to reproduce
I've attached the example android studio project which reproduces the issue consistently on an arm64 emulator:
[WasmAOT.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/11680810/WasmAOT.zip)

C++ code to load  the AOT module:
```
#include <android/asset_manager_jni.h>
#include <android/log.h>
#include <jni.h>

#include <cstring>
#include <string>
#include <vector>

#include "wasm_export.h"

extern "C" JNIEXPORT jstring JNICALL
Java_com_example_wasmaot_MainActivity_stringFromJNI(
        JNIEnv* env,
        jobject /* this */) {
    std::string hello = "Hello from wasm aot example";
    return env->NewStringUTF(hello.c_str());
}

extern "C" JNIEXPORT jint JNICALL
Java_com_example_wasmaot_MainActivity_loadWasmAOTModule(
        JNIEnv* env,
        jobject /* this */,
        jobject jasset_manager) {
    auto asset_manager = AAssetManager_fromJava(env, jasset_manager);

    auto aot_file_name = std::string("wamr_aot_test_") + AOT_FILE_SUFFIX + std::string(".aot");

    __android_log_print(ANDROID_LOG_INFO,"example.wasmaot", "%s", ("Loading AOT file " + aot_file_name).c_str());

    auto asset = AAssetManager_open(asset_manager, aot_file_name.c_str(), AASSET_MODE_BUFFER);

    if (!asset) {
        __android_log_print(ANDROID_LOG_ERROR,"example.wasmaot", "Failed to open AOT asset");
        return 1;
    }

    auto size = AAsset_getLength(asset);
    auto asset_buffer = AAsset_getBuffer(asset);

    std::vector<uint8_t> buf;
    buf.reserve(size);
    buf.assign(static_cast<const uint8_t*>(asset_buffer), static_cast<const uint8_t*>(asset_buffer) + size);

    AAsset_close(asset);

[...]

PR #3681, opened 2024-08-01
aarch64 AOT module fails to load on android

bianchui:
Fixes: https://github.com/bytecodealliance/wasm-micro-runtime/issues/2274


bianchui:
nuttx build always fail with error `/usr/bin/bash: line 1: /opt/wasi-sdk/bin/clang: No such file or directory`

yamt:
> nuttx build always fail with error `/usr/bin/bash: line 1: /opt/wasi-sdk/bin/clang: No such file or directory`

this error is not the reason of failure.

bianchui:
I add a function merge_data_and_text only for BUILD_TARGET_AARCH64 now, but it may fix other platform need data an text sections merged together.

core/iwasm/aot/aot_loader.c
```diff
@@ -2378,6 +2378,9 @@ destroy_object_data_sections(AOTObjectDataSection *data_sections,
                 }
             }
 #endif
+            if (merged_data) {
+                continue;
```
wenyongh, line 2382: How about
```C
    if (!merged_data)
        os_munmap(data_section->data, data_section->size);
```

core/iwasm/aot/aot_loader.c
```diff
@@ -2532,6 +2535,62 @@ load_init_data_section(const uint8 *buf, const uint8 *buf_end,
     return false;
 }
 
+#if defined(BUILD_TARGET_AARCH64)
```
wenyongh, line 2538: Merging data and text may also benefit other targets, e.g., it may reduce the tlb cache miss and improve the performance since two data may use a same page. How about we make it as a configurable feature, and disable it by default? For example:

add WAMR_BUILD_AOT_MERGED_MMAP in build-scripts/config_common.make,
add WASM_ENABLE_AOT_MERGED_MMAP macro, and set it to 0 by default in core/config.h,
and use WASM_ENABLE_AOT_MERGED_MMAP to control the code in this file?

And here when `merge_data_and_text` fails, just output an warning (e.g., LOG_WARNING), but not return false.

@yamt, how do you think?

yamt: it makes sense.
i'm not sure if it's worth a config though.

i guess we can unconditionally merge many of data sections at least.
llvm tends to produce multiple rodata sections for objects with different characteristics.
the following is taken from an 1MB x86-64 aot file i happened to have.
as you see, these sections are usually not so large.
using os_mmap for each of these small sections is a bit wasteful.
ideally we can merge them during aot-compilation.
```
data_section_count 5
data_section_name .rodata.cst16
data_section_size 736
data section [0] offset 44924 size 736
data_section_name .rodata
data_section_size 4988
data section [1] offset 45676 size 4988
data_section_name .rodata.cst8
data_section_size 680
data section [2] offset 50684 size 680
data_section_name .rodata.cst4
data_section_size 244
data section [3] offset 51384 size 244
data_section_name .rodata.cst32
data_section_size 448
data section [4] offset 51648 size 448
```

if we always merge (or at least place together) some of sections, i guess we can even resolve some relocations during aot-compilation.

wenyongh: Yes, @lum1n0us also suggested to unconditional try to merge data and text, and not to add cmake variable and gcc macro to control the code. It is good to me, my concern is that we should review the code carefully and do more test, to avoid potential logic error and memory leak.

For merging data sections, I think we don't need to change the behavior in the aot compiler, instead in load_object_data_sections of this file, we can traverse the data sections twice, the first time is to calculate the total size of the data sections and then mmap the merged data sections, then setting each data section in the second time traversing.
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/aot/aot_loader.c#L2404-L2417

@bianchui How about adding below fields in AOTModule:
```C
    uint8 *merged_data_sections;
    uint32 merged_data_sections_size;
    uint8 *merged_data_text_sections;
    uint32 merged_data_text_sections_size;
```

And after that, we can mmap the `merged_data_text_sections` for merged `merged_data_sections` and aot text. And normally we had better not do it in two scenarios: (1) the XIP mode, e.g., the aot text is burned in rom device and developer doesn't hope to mmap aot text again in aot loader, (2) the module is loaded by aot_load_from_sections, in which the aot text is pre-mmapped. Note that aot_load_from_sections calls load_from_sections with `is_load_from_file_buf = false`.

yamt: > For merging data sections, I think we don't need to change the behavior in the aot compiler,

we need to change the compiler to merge sections for xip.
and it would make the non xip case simpler as well.

yamt: > my concern is that we should review the code carefully and do more test, to avoid potential logic error and memory leak.

i suppose having conditionals actually increases review and testing effort.

wenyongh: > we need to change the compiler to merge sections for xip. and it would make the non xip case simpler as well.

But this may require changing the AOT format?

> i suppose having conditionals actually increases review and testing effort.

Agree, I just thought that adding a flag and disabling it by default won't impact the current solution. But it is good not to add the flag if we are confident that the new solution will be well-implemented.

core/iwasm/aot/aot_loader.c
bianchui: That's better. Change in new commit.

core/iwasm/aot/aot_loader.c
```diff
@@ -2532,6 +2535,68 @@ load_init_data_section(const uint8 *buf, const uint8 *buf_end,
...
+        int map_flags = MMAP_MAP_32BIT;
+#else
+        int map_flags = MMAP_MAP_NONE;
+#endif
+        sections = module->merged_sections =
+            os_mmap(NULL, total, map_prot, map_flags, os_get_invalid_handle());
+        if (!sections) {
+            return false;
+        }
+        /* data first */
```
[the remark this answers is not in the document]
bianchui: No.
I just put data first because it is parsed in order `AOT_SECTION_TYPE_INIT_DATA` then `AOT_SECTION_TYPE_TEXT`.
Compiler normally put .data section after .text section.

core/iwasm/aot/aot_loader.c
```diff
@@ -3749,6 +3808,11 @@ load_from_sections(AOTModule *module, AOTSection *sections,
                     return false;
                 break;
             case AOT_SECTION_TYPE_TEXT:
+#if defined(BUILD_TARGET_AARCH64)
+                if (!merge_data_and_text(&buf, &buf_end, module, error_buf,
+                                         error_buf_size))
+                    return false;
```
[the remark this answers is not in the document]
bianchui: When merge failed, aot module will refuse to load because [R_AARCH64_ADR_PREL_PG_HI21 need < 4G distance](https://github.com/bytecodealliance/wasm-micro-runtime/issues/2274#issuecomment-2262102200)

core/iwasm/aot/aot_loader.c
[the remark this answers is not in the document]
bianchui: new commit change memory order as compilers do.

core/iwasm/aot/aot_loader.c
```diff
@@ -2532,6 +2535,62 @@ load_init_data_section(const uint8 *buf, const uint8 *buf_end,
...
 }
 
+#if defined(BUILD_TARGET_AARCH64)
+static bool
+merge_data_and_text(const uint8 **buf, const uint8 **buf_end, AOTModule *module,
+                    char *error_buf, uint32 error_buf_size)
+{
+    uint8 *old_buf = (uint8 *)*buf;
+    uint8 *old_end = (uint8 *)*buf_end;
+    uint32 code_size = (uint32)(uint64)(old_end - old_buf);
```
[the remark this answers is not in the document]
bianchui: Thanks, i check all size related code and change it

core/iwasm/aot/aot_loader.c
```diff
@@ -2532,6 +2535,62 @@ load_init_data_section(const uint8 *buf, const uint8 *buf_end,
...
+        uint64 diff =
+            old_end < data_section->data
+                ? (uint64)(data_section->data + data_section->size - old_end)
+                : (uint64)(old_buf - data_section->data);
+        if (diff > ((int64)4 * BH_GB)) {
+            need_merge = true;
+        }
+    }
+    if (need_merge) {
+        int map_prot = MMAP_PROT_READ | MMAP_PROT_WRITE | MMAP_PROT_EXEC;
```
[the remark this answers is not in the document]
bianchui: Sorry for that. Add a commit to change protection.

core/iwasm/aot/aot_loader.c
```diff
@@ -2532,6 +2535,62 @@ load_init_data_section(const uint8 *buf, const uint8 *buf_end,
...
+            return false;
+        }
+        /* data first */
+        for (i = 0; i < module->data_section_count; ++i) {
+            AOTObjectDataSection *data_section = module->data_sections + i;
+            uint8 *old_data = data_section->data;
+            data_section->data = sections;
+            sections +=
+                (data_section->size + k_page_size - 1) & ~(k_page_size - 1);
+            memcpy(data_section->data, old_data, data_section->size);
```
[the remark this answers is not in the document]
bianchui: I change to bh_memcpy_s but size is all calc from same size, is that good enough

core/iwasm/aot/aot_loader.c
```diff
@@ -3749,6 +3814,11 @@ load_from_sections(AOTModule *module, AOTSection *sections,
                     return false;
                 break;
             case AOT_SECTION_TYPE_TEXT:
+#if defined(BUILD_TARGET_AARCH64)
```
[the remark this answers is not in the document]
bianchui: comment for why need this #ifdef is added
if that [unconditional try to merge](https://github.com/bytecodealliance/wasm-micro-runtime/pull/3681#discussion_r1704794923) will apply, then this will be changed.

core/iwasm/aot/aot_loader.c
```diff
@@ -2532,6 +2535,68 @@ load_init_data_section(const uint8 *buf, const uint8 *buf_end,
...
+    /* calc total memory needed */
+    total += (code_size + k_page_size - 1) & ~(k_page_size - 1);
+    for (i = 0; i < module->data_section_count; ++i) {
+        AOTObjectDataSection *data_section = module->data_sections + i;
+        total += (data_section->size + k_page_size - 1) & ~(k_page_size - 1);
+        uint64 diff =
+            old_end < data_section->data
+                ? (uint64)(data_section->data + data_section->size - old_end)
+                : (uint64)(old_buf - data_section->data);
+        if (diff > ((int64)4 * BH_GB)) {
```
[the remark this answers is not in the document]
bianchui: comment added

core/iwasm/aot/aot_loader.c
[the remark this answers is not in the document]
wenyongh: But when merge failed, it also works on non-aarch64 targets, we just hope this failure doesn't block working on them. It is good for aarch64 since the relocation failure will be reported later.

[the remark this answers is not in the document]
bianchui: That's right, i will change it

core/iwasm/aot/aot_loader.c
bianchui: should i merge data first into merged_data_sections
and then merge all into merged_data_text_sections add free merged_data_sections ?

> 
> @bianchui How about adding below fields in AOTModule:
> 
> ```c
>     uint8 *merged_data_sections;
>     uint32 merged_data_sections_size;
>     uint8 *merged_data_text_sections;
>     uint32 merged_data_text_sections_size;
> ```
>

wenyongh: Yes, that would be better, the data sections are merged even when the later merging of them with aot text fails.

BTW, could you merge data and text unconditionally per suggested by @yamt?
And since the patch may impact all the targets and platforms, I think we had better do more tests, how about I create a development branch e.g. `dev/merge_aot_data_text` and you change this PR to submit to that branch instead? We may test more and merge it into main branch when it is stable.

bianchui: > And after that, we can mmap the `merged_data_text_sections` for merged `merged_data_sections` and aot text. And normally we had better not do it in two scenarios: (1) the XIP mode, e.g., the aot text is burned in rom device and developer doesn't hope to mmap aot text again in aot loader, (2) the module is loaded by aot_load_from_sections, in which the aot text is pre-mmapped. Note that aot_load_from_sections calls load_from_sections with `is_load_from_file_buf = false`.

but load_object_data_sections_info() do the mmap in load_object_data_sections()
but that function do not consider is_load_from_file_buf

bianchui: > Yes, that would be better, the data sections are merged even when the later merging of them with aot text fails.
> 
> BTW, could you merge data and text unconditionally per suggested by @yamt? And since the patch may impact all the targets and platforms, I think we had better do more tests, how about I create a development branch e.g. `dev/merge_aot_data_text` and you change this PR to submit to that branch instead? We may test more and merge it into main branch when it is stable.

ok, i can do it. when the branch created, i will change PR merge target

wenyongh: > > And after that, we can mmap the `merged_data_text_sections` for merged `merged_data_sections` and aot text. And normally we had better not do it in two scenarios: (1) the XIP mode, e.g., the aot text is burned in rom device and developer doesn't hope to mmap aot text again in aot loader, (2) the module is loaded by aot_load_from_sections, in which the aot text is pre-mmapped. Note that aot_load_from_sections calls load_from_sections with `is_load_from_file_buf = false`.
> 
> i'm trying to match the `two scenarios` but mmap is being done in load_object_data_sections() and do not consider is_load_from_file_buf

Yes, in load_object_data_sections(), we can just traverse two times, one is to get total size, the other is to mmap and set each data section, no need to consider is_load_from_file_buf. For the two scenarios, I meant we should add checks before calling merge_data_and_text: (1) module is XIP, don't call merge_data_and_text if `module->is_indirect_mode` is true; (2) module is not load from file, since is_load_from_file_buf is provided in load_from_sections, you can extend load_text_section to take that argument and then: load_from_sections(.., is_load_from_file_buf) -> load_text_section(.., is_load_from_file_buf) -> don't call merge_data_and_text if is_load_from_file_buf is false.

bianchui: ok, clear.

wenyongh: > > Yes, that would be better, the data sections are merged even when the later merging of them with aot text fails.
> > BTW, could you merge data and text unconditionally per suggested by @yamt? And since the patch may impact all the targets and platforms, I think we had better do more tests, how about I create a development branch e.g. `dev/merge_aot_data_text` and you change this PR to submit to that branch instead? We may test more and merge it into main branch when it is stable.
> 
> ok, i can do it. when the branch created, i will change PR merge target

The branch `dev/merge_aot_data_text` was created: https://github.com/bytecodealliance/wasm-micro-runtime/tree/dev/merge_aot_data_text.

bianchui: > Merging data and text may also benefit other targets, e.g., it may reduce the tlb cache miss and improve the performance since two data may use a same page. 

merged sections align to page_size and do not share page because .text and .data have protection difference for now.

core/iwasm/aot/aot_loader.c
```diff
@@ -2412,29 +2429,33 @@ load_object_data_sections(const uint8 **p_buf, const uint8 *buf_end,
...
-        read_byte_array(buf, buf_end, data_sections[i].data,
-                        data_sections[i].size);
+    /* Second iteration: Create each data section */
+    for (i = 0; i < module->data_section_count; i++) {
+        if (data_sections[i].size > 0) {
+            bh_memcpy_s(merged_sections, data_sections[i].size,
+                        data_sections[i].data, data_sections[i].size);
+            data_sections[i].data = merged_sections;
+            merged_sections += ((size_t)data_sections[i].size + page_size - 1)
+                               & ~(page_size - 1);
```
wenyongh, line 2457: How about `merged_sections += align_uint(data_sections[i].size, page_size)`?

core/iwasm/aot/aot_loader.c
```diff
@@ -2532,6 +2553,76 @@ load_init_data_section(const uint8 *buf, const uint8 *buf_end,
...
+merge_data_and_text(const uint8 **buf, const uint8 **buf_end, AOTModule *module,
+                    char *error_buf, uint32 error_buf_size)
+{
+    uint8 *old_buf = (uint8 *)*buf;
+    uint8 *old_end = (uint8 *)*buf_end;
+    size_t code_size = (size_t)(old_end - old_buf);
+    uint32 page_size = os_getpagesize();
+    uint64 total_size = 0;
+    uint32 i;
+    uint8 *sections;
```
wenyongh, line 2566: Had better add an empty line after this line

core/iwasm/aot/aot_loader.c
```diff
@@ -2532,6 +2553,76 @@ load_init_data_section(const uint8 *buf, const uint8 *buf_end,
...
+    if (code_size == 0) {
+        return true;
+    }
+    /* calc total memory needed */
+    total_size += ((uint64)code_size + page_size - 1) & ~(page_size - 1);
```
wenyongh, line 2571: how about `total_size += align_uint64((uint64)code_size, page_size)`?

bianchui:
@wenyongh needs help about test error in `spec test on nuttx`, same code working well on android
did i miss something or how can i test locally

core/iwasm/aot/aot_loader.c
```diff
@@ -2532,6 +2553,76 @@ load_init_data_section(const uint8 *buf, const uint8 *buf_end,
...
+    for (i = 0; i < module->data_section_count; ++i) {
+        total_size += ((uint64)module->data_sections[i].size + page_size - 1)
+                      & ~(page_size - 1);
+    }
+    if (total_size > (uint64)code_size) {
```
wenyongh, line 2576: Should be `total_size > 0`? os_mmap should be called no matter that there is data section or not.

core/iwasm/aot/aot_loader.c
```diff
@@ -2532,6 +2553,76 @@ load_init_data_section(const uint8 *buf, const uint8 *buf_end,
...
+#else
+        int map_flags = MMAP_MAP_NONE;
+#endif
+        /* distance between .data and .text should not greater than 4GB for some
+         * targets (eg. arm64 reloc need < 4G distance) */
+        if (total_size > UINT32_MAX) {
+            return false;
+        }
+        sections = module->merged_data_text_sections = os_mmap(
+            NULL, total_size, map_prot, map_flags, os_get_invalid_handle());
```
wenyongh, line 2593: had better `(uint32)total_size`

core/iwasm/aot/aot_loader.c
```diff
@@ -2532,6 +2553,76 @@ load_init_data_section(const uint8 *buf, const uint8 *buf_end,
...
+        if (!sections) {
+            /* merge failed but maybe not critical for some targes */
+            return false;
+        }
+        module->merged_data_text_sections_size = total_size;
```
wenyongh, line 2598: had better `(uint32)total_size`

core/iwasm/aot/aot_loader.c
```diff
@@ -2532,6 +2553,76 @@ load_init_data_section(const uint8 *buf, const uint8 *buf_end,
...
+        /* order not essential just as compilers do: .text section first */
+        *buf = sections;
+        *buf_end = sections + code_size;
+        bh_memcpy_s(sections, code_size, old_buf, code_size);
+        os_mprotect(sections, code_size, map_prot | MMAP_PROT_EXEC);
+        os_munmap(old_buf, code_size);
+        sections += ((uint64)code_size + page_size - 1) & ~(page_size - 1);
```
wenyongh, line 2605: how about `sections += align_uint(code_size, page_size)`

core/iwasm/aot/aot_loader.c
```diff
@@ -2532,6 +2553,76 @@ load_init_data_section(const uint8 *buf, const uint8 *buf_end,
...
+        /* then .data sections */
+        for (i = 0; i < module->data_section_count; ++i) {
+            AOTObjectDataSection *data_section = module->data_sections + i;
+            uint8 *old_data = data_section->data;
+            data_section->data = sections;
+            bh_memcpy_s(data_section->data, data_section->size, old_data,
+                        data_section->size);
+            sections +=
+                ((uint64)data_section->size + page_size - 1) & ~(page_size - 1);
```
wenyongh, line 2614: how about `sections += align_uint(data_section->size, page_size)`

core/iwasm/aot/aot_loader.c
```diff
@@ -2532,6 +2553,76 @@ load_init_data_section(const uint8 *buf, const uint8 *buf_end,
...
+    uint32 i;
+    uint8 *sections;
+    if (code_size == 0) {
+        return true;
+    }
+    /* calc total memory needed */
+    total_size += ((uint64)code_size + page_size - 1) & ~(page_size - 1);
+    for (i = 0; i < module->data_section_count; ++i) {
+        total_size += ((uint64)module->data_sections[i].size + page_size - 1)
+                      & ~(page_size - 1);
```
wenyongh, line 2574: same as above

core/iwasm/aot/aot_loader.c
```diff
@@ -2412,29 +2429,33 @@ load_object_data_sections(const uint8 **p_buf, const uint8 *buf_end,
...
-            && !(data_sections[i].data =
-                     os_mmap(NULL, data_sections[i].size, map_prot, map_flags,
-                             os_get_invalid_handle()))) {
+        merged_sections = module->merged_data_sections = os_mmap(
+            NULL, total_size, map_prot, map_flags, os_get_invalid_handle());
+        if (!merged_sections) {
             set_error_buf(error_buf, error_buf_size, "allocate memory failed");
             return false;
         }
+        module->merged_data_sections_size = total_size;
```
wenyongh, line 2439: had better `= (uint32)total_size`

wenyongh:
> @wenyongh needs help about test error in `spec test on nuttx`, same code working well on android did i miss something or how can i test locally

I didn't have nuttx environment to test the spec cases either, not sure whether it is caused by this sentence `if (total_size > (uint64)code_size) {`. I replied some comments, could you help have look and try again?

wenyongh:
@bianchui Thanks for addressing the the comments. I tested the PR in my local repo, it seems that the nuttx CI failures were caused by partly os_mmap the merged data and aot with extra MMAP_PROT_EXEC flag: runtime os_mmap `module->merged_data_text_sections` with `MMAP_PROT_READ | MMAP_PROT_WRITE` and then os_mprotect the first part (aot text) of it with  `MMAP_PROT_READ | MMAP_PROT_WRITE | MMAP_PROT_EXEC`. Some device may not support that. If I changed os_mmap the whole region with the latter, it works:
https://github.com/wenyongh/wasm-micro-runtime/actions
The PR is:
https://github.com/wenyongh/wasm-micro-runtime/commit/ef1ec7b84f6bdf60648a05ba3b8914f5360395f0

I think we need to discuss how to figure out it, maybe os_mmap with extra MMAP_PROT_EXEC flag for some devices. Or check the return value of os_mprotect.

bianchui:
> I think we need to discuss how to figure out it, maybe os_mmap with extra MMAP_PROT_EXEC flag for some devices. Or check the return value of os_mprotect.
so merge .data is safe because they do not need flag `MMAP_PROT_EXEC`.
merge .data and .text is safe on android and windows at least, because they support page align protect change.
other platforms maybe not support part protect change.
macos with MAP_JIT maybe affected too [Hardened Runtime](https://developer.apple.com/documentation/security/hardened_runtime?language=objc) that`s needs test.
i will try check the mprotect return value first.

bianchui:
another question is they dose not support change protection of part mmaped memory?
or they only dose not support mixed `MMAP_PROT_EXEC` and without that flag memories in one mmaped?

bianchui:
according to man of macos mprotect 
`The mprotect() system call changes the specified pages to have protection prot.  Not all implementations will guarantee protection on a page basis but Mac OS X's current implementation does.`

wenyongh:
> according to man of macos mprotect `The mprotect() system call changes the specified pages to have protection prot. Not all implementations will guarantee protection on a page basis but Mac OS X's current implementation does.`

Yes, and some platforms' os_mprotect just returns 0, like nuttx and esp-idf, they may assume their os_mmap has correctly set the protection flags and won't change again. My suggestion is (1) `os_mmap` with read/write/exec flags for these platforms and not call `os_mprotect` again, (2) `os_mmap` with read/write flags other platforms, and `os_mprotect` the aot text part with extra exec flag, if fails, `os_munmap` merged data and text, and return false. The patch can be found:
https://github.com/wenyongh/wasm-micro-runtime/commit/1ae1f240062bb501a2ab52a9c602bd991e8fb791

Could you help check it? If it is good, could you please update your PR and maybe we can merge the PR for more test?

bianchui:
> Could you help check it? If it is good, could you please update your PR and maybe we can merge the PR for more test?

done with cherry-pick that commit.
is that safe to leave the MMAP_PROT_EXEC bit on for nuttx and esp-idf?

wenyongh:
> > Could you help check it? If it is good, could you please update your PR and maybe we can merge the PR for more test?
> 
> done with cherry-pick that commit. is that safe to leave the MMAP_PROT_EXEC bit on for nuttx and esp-idf?

I am not sure, this means that the data sections in the `merged data and text` are also marked as executable. I think we need the help of nuttx and esp-idf guys.

@yamt, @no1wudi how do you think? Thanks.

yamt:
> > > Could you help check it? If it is good, could you please update your PR and maybe we can merge the PR for more test?
> > 
> > 
> > done with cherry-pick that commit. is that safe to leave the MMAP_PROT_EXEC bit on for nuttx and esp-idf?
> 
> I am not sure, this means that the data sections in the `merged data and text` are also marked as executable. I think we need the help of nuttx and esp-idf guys.
> 
> @yamt, @no1wudi how do you think? Thanks.

for nuttx, depending on the configuration, os_mmap with MMAP_PROT_EXEC allocates memory from a separate region, which might be scarce and even might involve special access restrictions.
on the other hand, the region allocated by os_mmap w/o MMAP_PROT_EXEC might not be executed.
(note: esp32 qemu used by our CI doesn't emulate these bits well. you need to test on a real hardware.)

i guess it's safer to avoid merging text and data for nuttx.

bianchui:
> i guess it's safer to avoid merging text and data for nuttx.
> 
> although i have no idea about esp-idf, i guess it's similar. (as it runs on the same esp32 chips)

so just disable merge .text and .data for nuttx and esp32 is better?

no1wudi:
On resource-constrained embedded platforms, we indeed cannot guarantee that functions like os_mmap, typically implemented with an MMU, will behave consistently with complex Unix-based platforms. Therefore, it's best to assume that such functionalities are disabled by default.

wenyongh:
@bianchui I refactored the code of aot loader in another PR #3711, could you help have a look? Thanks.

bianchui:
> @bianchui I refactored the code of aot loader in another PR #3711, could you help have a look? Thanks.

I read the PR and everything is fine

PR #3682, opened 2024-08-02
Fix a compilation warning

no1wudi:
Fix:
```
wamr/core/iwasm/compilation/aot_llvm.c: In function ‘insert_native_symbol’:
wamr/core/iwasm/compilation/aot_llvm.c:3290:28: warning: comparison of integer expressions of different signedness: ‘int’ and ‘long unsigned int’ [-Wsign-compare]
 3290 |     if (ret < 0 || ret + 1 > sizeof(sym->symbol)) {
      |                            ^
```

### DIFF
--- a/core/iwasm/compilation/aot_llvm.c
+++ b/core/iwasm/compilation/aot_llvm.c
@@ -3287,7 +3287,7 @@ insert_native_symbol(AOTCompContext *comp_ctx, const char *symbol, int32 idx)
     memset(sym, 0, sizeof(AOTNativeSymbol));
     bh_assert(strlen(symbol) <= sizeof(sym->symbol));
     ret = snprintf(sym->symbol, sizeof(sym->symbol), "%s", symbol);
-    if (ret < 0 || ret + 1 > sizeof(sym->symbol)) {
+    if (ret < 0 || ret + 1 > (int)sizeof(sym->symbol)) {
         aot_set_last_error_v("symbol name too long: %s", symbol);
         return false;
     }


core/iwasm/compilation/aot_llvm.c
```diff
@@ -2356,7 +2356,7 @@ jit_stack_size_callback(void *user_data, const char *name, size_t namelen,
     stack_consumption_to_call_wrapped_func =
         musttail ? 0
                  : aot_estimate_stack_usage_for_function_call(
-                     comp_ctx, func_ctx->aot_func->func_type);
+                       comp_ctx, func_ctx->aot_func->func_type);
```
no1wudi, line 2359: This change is from clang-format

core/iwasm/compilation/aot_llvm.c
```diff
@@ -3287,7 +3287,7 @@ insert_native_symbol(AOTCompContext *comp_ctx, const char *symbol, int32 idx)
     memset(sym, 0, sizeof(AOTNativeSymbol));
     bh_assert(strlen(symbol) <= sizeof(sym->symbol));
     ret = snprintf(sym->symbol, sizeof(sym->symbol), "%s", symbol);
-    if (ret < 0 || ret + 1 > sizeof(sym->symbol)) {
+    if (ret < 0 || ret + 1 > (int)sizeof(sym->symbol)) {
```
yamt, line 3290: in general i prefer to cast to the larger type. (size_t)
but i guess it doesn't matter for this specific code.

yamt (approved):

core/iwasm/compilation/aot_llvm.c
no1wudi: Yes, the size of sym->symbol is is only 48 byte for this case

PR #3693, opened 2024-08-07
posix_thread.c: Restore old signal alternate stack before thread exits

wenyongh:
Some host environment may also create an signal alternate stack and access
it after the wasm runtime exits, the runtime should backup the stack info and
restore it before thread exits.

The issue was found in golang:
```
signal 23 received on thread with on signal stack
fatal error: non-Go code disabled signaltstack
```


TianlongLiang:
LGTM

PR #3701, opened 2024-08-12
[refactoring] extract reading leb numbers to a separate file, share the code between loader and mini loader

loganek:
There's probably a number of other places where the `bh_leb_read` could be used (e.g. aot loader) but I'm making the change as small as possible. Further refactoring can be done later.

### DIFF
--- a/core/shared/utils/bh_leb128.c
+++ b/core/shared/utils/bh_leb128.c
@@ -1,0 +1,77 @@
+/*
+ * Copyright (C) 2019 Intel Corporation.  All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+
+#include "bh_leb128.h"
+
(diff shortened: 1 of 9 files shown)


core/shared/utils/bh_leb128.c
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright (C) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```
yamt, line 3: this file seems mostly copy-and-paste from wasm_loader.c.
please retain the original copyright notice.

loganek: good point, updated.

wenyongh (commented):
LGTM

yamt (approved):
lgtm

Issue #3500, opened 2024-06-04
Enabling WAMR_BUILD_DEBUG_INTERP causes the program to run slower


sosotan:
I used the Windows environment to compile iwasm.exe for testing, and when the debugging feature is enabled during compilation, compared to no debugging  feature the program runs very slowly.

After testing, it was found that it is because each opcode executes the macro HANDLE_OP_END, which performs a lock operation os_mutex_lock(&exec_env->wait_lock). Internally, it calls WaitForSingleObject, which executes very frequently, causing the program to run slower.

My build iwasm.exe method (using VS2022):
```
cd wasm-micro-runtime\product-mini\platforms\windows\build
cmake .. -DWAMR_BUILD_FAST_INTERP=0 -DWAMR_BUILD_DEBUG_INTERP=1
```

My test code wasm.c (never mind the div 0 exception)
```
#include <stdio.h>

static void foo() {
  printf("foo=");
}

int main() {
  int i = 1;
  int j = (++i) + (++i);

  for (int count = 0; ; count++) {
    foo();
    printf("%d----------%d\n", j, count);

    if (count > 10) {
      j = i / (j - j);
      printf("%d----------%d\n", j, count);
    }

    for (int w = 0; w < 10; w++) {
      printf("wait%d\n", w);
    }
  }
  return 0;
}
```

My test method:
```
wasi-sdk-20.0\bin\clang.exe -g wasm.c -o wasm.out
iwasm.exe wasm.out
```

TianlongLiang:
Unfortunately, I think this is necessary in order to support (wasm)instruction level step into/over.

iKlask:
It'd be nice if the WAMR engine could skip the extra mutex calls when the WASM binary lacks debug information.

I understand that debug builds are generally slower than release builds. However, I've noticed a significant slowdown in the classic interpreter, even without active debugging or debug symbols in my WASM binary. My current solution is to always use AOT and reserve the interpreter for debugging purposes only, ensuring there is no slowdown in my "release" WASM code.

I must note that I haven't actually looked at the source code for how WAMR executes the opcodes. This observation is based on OP's claim that with `WAMR_BUILD_DEBUG_INTERP `enabled, every opcode triggers `os_mutex_lock(&exec_env->wait_lock)`. If this is causing the slowdowns I'm seeing, it would be nice if it were avoidable when debugging is not required.

wenyongh:
[...]

PR #3704, opened 2024-08-13
debug-interp: Only add lock when signal_flag is SIG_SINGSTEP

wenyongh:
As reported in #3500, when debug interpreter is enabled, the classic interpreter
performs a lock operation to read `exec_env->current_status->signal_flag` and
do further handling before fetching next opcode, which makes the interpreter
run slower.

This PR atomic loads the `exec_env->current_status->signal_flag` without mutex
lock, and only adding lock for further handling when the signal_flag is
WAMR_SIG_SINGSTEP, which improves the performance.

### DIFF
--- a/core/iwasm/interpreter/wasm_interp_classic.c
+++ b/core/iwasm/interpreter/wasm_interp_classic.c
@@ -1399,6 +1399,23 @@ wasm_interp_call_func_import(WASMModuleInstance *module_inst,
 #endif /* WASM_ENABLE_DEBUG_INTERP */
 #endif /* WASM_ENABLE_THREAD_MGR */
 
+#if WASM_ENABLE_THREAD_MGR != 0 && WASM_ENABLE_DEBUG_INTERP != 0
+#if BH_ATOMIC_32_IS_ATOMIC != 0
+#define GET_SIGNAL_FLAG()                                             \
+    do {                                                              \
+        signal_flag =                                                 \
+            BH_ATOMIC_32_LOAD(exec_env->current_status->signal_flag); \
+    } while (0)
+#else
+#define GET_SIGNAL_FLAG()                                    \
+    do {                                                     \
+        os_mutex_lock(&exec_env->wait_lock);                 \
+        signal_flag = exec_env->current_status->signal_flag; \
+        os_mutex_unlock(&exec_env->wait_lock);               \
+    } while (0)
+#endif
+#endif
+
 #if WASM_ENABLE_LABELS_AS_VALUES != 0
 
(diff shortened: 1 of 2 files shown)


core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -1404,12 +1404,42 @@ wasm_interp_call_func_import(WASMModuleInstance *module_inst,
...
 #define HANDLE_OP_END()                                                   \
-    do {                                                                  \
-        /* Record the current frame_ip, so when exception occurs,         \
-           debugger can know the exact opcode who caused the exception */ \
-        frame_ip_orig = frame_ip;                                         \
+    /* Record the current frame_ip, so when exception occurs,             \
+       debugger can know the exact opcode who caused the exception */     \
+    frame_ip_orig = frame_ip;                                             \
+    /* atomic load the signal_flag first, and handle more with lock       \
+       if it is WAMR_SIG_SINGSTEP */                                      \
```
yamt, line 1440: can you add a comment why you don't need to check BH_ATOMIC_32_IS_ATOMIC?

wenyongh: IIUC, in most cpu architectures, the 32-bit load operation is atomic when the address to load is 4-byte aligned. But it should be safer to atomic load the signal_flag or load with lock according to BH_ATOMIC_32_IS_ATOMIC, I updated the code to handle BH_ATOMIC_32_IS_ATOMIC, please help review again. Since most cpu supports 32-bit atomic load, they should be able to benefit from the PR. And for cpu that doesn't support 32-bit atomic load, the performance should be almost the same since the second lock occurs only when the signal_flag gotten is WAMR_SIG_SINGSTEP.

yamt: i'm not quite sure about the locking protocol here.
to me, it seems that the signal_flag is modified w/o any lock or atomic operations. (eg. wasm_cluster_send_signal_all)

wenyongh: Yes, I didn't check that before, it is really chaotic, `wasm_cluster_clear_thread_signal` and `wasm_cluster_thread_send_signal` sets `signal_flag` without lock, and some of its callers call them with lock, e.g., `wasm_cluster_thread_continue` and `set_thread_cancel_flags`, some call them without lock, e.g., `wasm_cluster_send_signal_all`. I think we can merge this PR first and fix them later, I need to discuss with others for how to fix them.

yamt (approved):

PR #3711, opened 2024-08-14
aot loader: Refine os_mmap related code

wenyongh:


core/iwasm/aot/aot_loader.c
```diff
@@ -3616,19 +3623,12 @@ load_relocation_section(const uint8 *buf, const uint8 *buf_end,
         GOTItem *got_item = module->got_item_list;
         uint32 got_item_idx = 0;
 
-        map_prot = MMAP_PROT_READ | MMAP_PROT_WRITE;
-        /* aot code and data in x86_64 must be in range 0 to 2G due to
-           relocation for R_X86_64_32/32S/PC32 */
-        map_flags = MMAP_MAP_32BIT;
-
```
yamt, line 3623: do you know why MMAP_MAP_32BIT is sometimes conditional and sometimes not?

wenyongh: The llvm codegen may generate relocations which require the target address (or the distance calculated) is in the range of 32-bit, e.g., R_X86_64_32, R_X86_64_32S, R_X86_64_PC32, R_RISCV_32. It was found in x86-64 and riscv64, and hasn't been found in aarch64 yet, so I only added the macro control for the former.

yamt: given the current usage, it seems that is_mmap_32bit_needed can be a part of loader_mmap.
how do you think?

wenyongh: Agree, how about in 64-bit target, try to map with MMAP_MAP_32BIT flag and return the mapped mem if success; otherwise, try again without MMAP_MAP_32BIT flag: if the mapped mem is larger than 4GB, the loader can also run into failure when failing to apply relocations, it just returns failure later than what loader_mmap does, but it doesn't block the scenarios that doesn't have these relocations.

And in 32-bit target, map without this flag.

yamt: sounds fine.

wenyongh: done.

PR #3715, opened 2024-08-14
add missing typedefs and headers

sjamesr:


core/shared/utils/bh_atomic.h
```diff
@@ -6,6 +6,8 @@
 #ifndef _BH_ATOMIC_H
 #define _BH_ATOMIC_H
 
+#include <stdint.h>
```
wenyongh, line 9: Could you include `bh_platform.h` instead? like other files of this folder.

core/shared/utils/bh_atomic.h
```diff
@@ -44,6 +46,9 @@ extern "C" {
  * #endif
  */
 
+typedef uint64_t uint64;
+typedef uint32_t uint32;
+typedef uint16_t uint16;
```
wenyongh, line 51: If including `bh_platform.h`, we don't need to add the three lines, since these types are already defined in the platform's header files.

core/iwasm/compilation/aot_llvm_extra.cpp
```diff
@@ -41,6 +41,7 @@
 #include <llvm/Target/CodeGenCWrappers.h>
 #include <llvm/Target/TargetMachine.h>
 #include <llvm/Target/TargetOptions.h>
+#include <llvm/TargetParser/Triple.h>
```
wenyongh, line 44: The `llvm/TargetParser/Triple.h` was introduced since llvm-17, and `llvm/ADT/Triple.h` was used before llvm-17, which is included in this file, see L11. How about adding
```C
#if LLVM_VERSION_MAJOR >= 17
#include <llvm/TargetParser/Triple.h>
#endif
```

core/shared/utils/bh_atomic.h
sjamesr: Done, thank you!

core/shared/utils/bh_atomic.h
sjamesr: Removed

core/iwasm/compilation/aot_llvm_extra.cpp
sjamesr: Done, thanks again

wenyongh (approved):

PR #3721, opened 2024-08-15
AOT: Fix NaN handling for f32/f64 const op

no1wudi:
If the value of a float constant is NaN, it will not be handed correctly by f32.const or f64.const, then it may cause a relocation in the AOT file.

### DIFF
--- a/core/iwasm/compilation/aot_emit_const.c
+++ b/core/iwasm/compilation/aot_emit_const.c
@@ -68,23 +68,21 @@ aot_compile_op_f32_const(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
 {
     LLVMValueRef alloca, value;
 
-    if (!isnan(f32_const)) {
-        if (comp_ctx->is_indirect_mode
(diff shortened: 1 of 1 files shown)


core/iwasm/compilation/aot_emit_const.c
```diff
@@ -123,23 +121,22 @@ aot_compile_op_f64_const(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
...
+        memcpy(&wasm_value.f64, &f64_const, sizeof(float64));
+        value = aot_load_const_from_table(comp_ctx, func_ctx->native_symbol,
+                                          &wasm_value, VALUE_TYPE_F64);
+        if (!value) {
+            return false;
         }
+        PUSH_F64(value);
+    }
+    else if (!isnan(f64_const)) {
+
```
wenyongh, line 136: Had better not add this empty line

wenyongh (commented):
LGTM except a minor comment

PR #3726, opened 2024-08-16
Fix table idx resolving in op call_indirect/return_call_indirect

wenyongh:
The table index in the call_indirect/return_call_indirect opcode should be
one byte 0x00 when ref-types/GC isn't enabled, and should be treated as
leb u32 when ref-types/GC is enabled.

And make aot compiler bail out if ref-types/GC is disabled by command line
argument while ref-types instructions are used.


lum1n0us (approved):
LGTM

PR #3728, opened 2024-08-17
Add file and line number to LOG_ macros

sjamesr:


yamt (commented):
i guess it has been omitted intentionally. probably to save the size.
how much does this change increase the size?

sjamesr:
Built at 740f499e9, libiwasm.so ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, stripped goes from 459608 bytes to 463704 bytes, so around a 1% increase.

Is libiwasm.so size a representative check? Should I be looking at something else?

wenyongh:
I also think it was omitted intentionally, per my understanding, this PR makes the log look ugly and not easy to read, and now only LOG_FATAL and LOG_DEBUG dump the file info and line number. I am not sure why you need to dump so much detailed info? Since normally you can get the file and line number by searching the log string in the wamr folder.

PR #3734, opened 2024-08-20
Fix load error not reported when magic header is invalid

wenyongh:
When AOT isn't enabled and the input is a wasm file, wasm_runtime_load doesn't
report error. Same when interpreter isn't enabled and the input is AOT file.

This PR makes wasm_runtime_load report error "magic header not detected" for
such situations.


TianlongLiang:
LGTM

PR #3742, opened 2024-08-21
Add support for multi-memory proposal in classic interpreter

wenyongh:
Implement multi-memory for classic-interpreter. Support core spec (and bulk memory) opcodes now,
and will support atomic opcodes, and add multi-memory export APIs in the future. 

PS: Multi-memory spec test patched a lot for linking test to adapt for multi-module implementation.


wenyongh:
Merged this PR into branch main from branch dev/multi_memory since the related PR #3570 had been reviewed.

PR #3755, opened 2024-08-28
Mac doesn't support MMAP_MAP_32BIT so this condition ends up asserting.

Andersbakken:
Mac on aarch64 uses posix_memmap.c which doesn't do anything with the flag MMAP_MAP_32BIT for that build so this condition ends up asserting unless the mapping ends up in the first 4 gigs worth of addressable space. Not entirely what problems not guaranteeing the mapping to be in that range will cause but as of now it asserts for me every time. It appears the usage of MAP_32BIT isn't functional after MacOS 13 (if I understand this thread correctly https://github.com/bytecodealliance/wasm-micro-runtime/issues/3009)

### DIFF
--- a/core/iwasm/aot/aot_loader.c
+++ b/core/iwasm/aot/aot_loader.c
@@ -302,7 +302,10 @@ loader_mmap(uint32 size, bool prot_exec, char *error_buf, uint32 error_buf_size)
     int map_flags;
     void *mem;
 
-#if UINTPTR_MAX == UINT64_MAX
+#if defined(BUILD_TARGET_X86_64) || defined(BUILD_TARGET_AMD_64) \
+    || defined(BUILD_TARGET_RISCV64_LP64D)                       \
+    || defined(BUILD_TARGET_RISCV64_LP64)
+#ifndef __APPLE__
     /* The mmapped AOT data and code in 64-bit targets had better be in
        range 0 to 2G, or aot loader may fail to apply some relocations,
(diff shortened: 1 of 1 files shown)


core/iwasm/aot/aot_loader.c
```diff
@@ -302,7 +302,7 @@ loader_mmap(uint32 size, bool prot_exec, char *error_buf, uint32 error_buf_size)
     int map_flags;
     void *mem;
 
-#if UINTPTR_MAX == UINT64_MAX
+#if UINTPTR_MAX == UINT64_MAX && (!defined(__APPLE__)  || !defined(__arm64__))
```
wenyongh, line 305: The `MMAP_MAP_32BIT` is only handled when the target is x86-64 or riscv64, and `__APPLE__` isn't defined. See posix os_mmap:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/shared/platform/common/posix/posix_memmap.c#L83-L98

How about changing to:
```C
#if defined(BUILD_TARGET_X86_64) || defined(BUILD_TARGET_AMD_64) \
    || defined(BUILD_TARGET_RISCV64_LP64D)                       \
    || defined(BUILD_TARGET_RISCV64_LP64)
#ifndef __APPLE__
...
#endif
#endif
```

Andersbakken: Makes sense to me. Updated now.

wenyongh (approved):

yamt (approved):
lgtm

Issue #3758, opened 2024-08-28
[RFC] Option to reduce code size of the generated AOT file when stack trace is enabled


loganek:
### Feature

We'd like to propose an option in the WAMR AOT compiler that will reduce the size of the generated code while still keeping track of some information about the existing callstack.

Currently, when compiling our file (~6 MB, over 100k call instructions and ~15k functions) to AOT with call stack tracking enabled (`--enable-dump-call-stack` option), the resulting file is 25MB (and is even more for ARM). When the call stack tracking is disabled, the size is ~18 MB, but we can't turn it off in production so we can analyze the crashes. Our experiments show that with our settings (no GC, no EH which is not supported yet but will be in the future) we could generate ~20MB bundle while keeping the basic information (just function call stacks).

We're not targeting GC, JIT and EH (once it's ready) modes although the solution potentially can be extended to those features too (didn't dig deep into the implementation).

### Benefit

The main benefit of the feature is a reduced AOT code size (perhaps at the cost of some of the additional features) therefore a lower memory usage and disk usage (or network usage if the bundle is downloaded). 

### Implementation
We considered two main approaches to implement this functionality. We're leaning towards option 1 given the concerns around the other option, but we'd like to open the discussion for the community and also hear some of the ideas.

#### Option 1: keep wrapping `call` instructions

Currently stack traces are implemented by adding extra code around `call` instruction for saving the frame on the stack before the instruction, and popping it after the instruction is executed. This approach assumes that we'll re-use the existing mechanism, but the code for pushing the stack trace will be simplified and configurable.

Right now the --enable-dump-call-stack generates a code for:

* keeping track of the callstack, including:
* checking if there's enough space to allocate a new frame
* previous frame
* function index
* instruction offset
* imported function's parameters
* operand stack (GC only)
* frame ref flags (GC only)
* performance info (only when profiling is enabled)

We'd like to introduce a new flag (the name is yet to be decided, perhaps something like `--dump-call-stack-detail-level` but it's yet to be decided) which will allows to define what's being tracked as part of each call and the level of validation. Each frame will include the following information:
* function number (always present)
* instruction offset (optional, configured through wamrc flag)

For now we'll completely disable validation for stack overflows, but if required, this can be added as an option too in the future.

This feature will not allow to include any other information - the `--enable-dump-call-stack` flag should be used instead if additional data is required. 

 To avoid additional instructions, we'll do the following:
[...]

PR #3763, opened 2024-09-02
Add wamrc parameter to configure stack frame features

loganek:
Those parameters can be used to reduce the size of the AOT code.

There's going to be more changes related to AOT code size reduction, this is just the initial step. I split the whole optimization to smaller chunks to make reviews a bit easier.

### DIFF
--- a/core/iwasm/include/aot_comp_option.h
+++ b/core/iwasm/include/aot_comp_option.h
@@ -6,6 +6,23 @@
 #ifndef __AOT_COMP_OPTION_H__
 #define __AOT_COMP_OPTION_H__
 
+typedef struct {
+    /* Enables or disables bounds checks for stack frames. When enabled, the AOT
+     * compiler generates code to check if the stack pointer is within the
+     * bounds of the current stack frame (and if not, traps). */
+    bool bounds_checks;
+
+    /*  Enables or disables instruction pointer (IP) tracking.*/
+    bool ip;
+
+    /* Enables or disables tracking instruction pointer of a trap. Only takes
+     * effect when `ip` is enabled.*/
+    bool trap_ip;
+
+    /* Enables or disables parameters, locals and stack operands. */
+    bool values;
+} AOTCallStackFeatures;
+
 typedef struct AOTCompOption {
     bool is_jit_mode;
     bool is_indirect_mode;
@@ -22,6 +39,7 @@ typedef struct AOTCompOption {
     bool enable_gc;
     bool enable_aux_stack_check;
     bool enable_aux_stack_frame;
+    AOTCallStackFeatures call_stack_features;
     bool enable_perf_profiling;
     bool enable_memory_profiling;
     bool disable_llvm_intrinsics;
--- a/core/iwasm/compilation/aot_compiler.c
+++ b/core/iwasm/compilation/aot_compiler.c
@@ -337,6 +337,10 @@ aot_gen_commit_values(AOTCompFrame *frame)
     LLVMValueRef value;
     uint32 n;
 
+    if (!frame->comp_ctx->call_stack_features.values) {
+        return true;
+    }
+
     /* First, commit reference flags
      * For LLVM JIT, iterate all local and stack ref flags
      * For AOT, ignore local(params + locals) ref flags */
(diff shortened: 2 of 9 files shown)


loganek:
Similarly to https://github.com/bytecodealliance/wasm-micro-runtime/pull/3764, the build fails due to unstable nuttx tests

core/iwasm/include/aot_comp_option.h
```diff
@@ -6,6 +6,26 @@
...
+
+    /* Enables or disables tracking instruction pointer of a trap. Only takes
+     * effect when `ip` is enabled.*/
+    bool trap_ip;
+
+    /* Enables or disables stack pointer (SP) tracking. */
+    bool sp;
+
+    /* Enables or disables imported functions' parameters tracking. */
+    bool params;
```
wenyongh, line 26: Several questions about the options:
(1) the func_idx will be always committed so no need to add an option for it, right?

(2) in fact, not only function params can be committed, but also function's locals and stack operands. Now they are now committed only when GC feature is enabled: `aot_gen_commit_values()` is called with `if (comp_ctx->enable_gc
                && !aot_gen_commit_values(comp_ctx->aot_frame))` like code.
do you think that we should enable committing them and change `params` to other word like `values`?

(3) by default, GC should enable sp and params (and locals, stack operands), or AOT runtime may run failed. So had better set them true when GC is enabled, no matter what user enables them or not in this flag?

loganek: > Several questions about the options: (1) the func_idx will be always committed so no need to add an option for it, right?
> 

Yes, I don't know if there's any usecase to disable `func_idx`; if needed, that option can be added in the future though.

> (2) in fact, not only function params can be committed, but also function's locals and stack operands. Now they are now committed only when GC feature is enabled: `aot_gen_commit_values()` is called with `if (comp_ctx->enable_gc && !aot_gen_commit_values(comp_ctx->aot_frame))` like code. do you think that we should enable committing them and change `params` to other word like `values`?
> 

I think we could rename it to values and cover more data; if needed this can be more granular in the future

> (3) by default, GC should enable sp and params (and locals, stack operands), or AOT runtime may run failed. So had better set them true when GC is enabled, no matter what user enables them or not in this flag?

isn't stack pointer and stack operands only actually used in the GC mode? If so, should the SP option be deleted at all?

wenyongh: > > Several questions about the options: (1) the func_idx will be always committed so no need to add an option for it, right?
> 
> Yes, I don't know if there's any usecase to disable `func_idx`; if needed, that option can be added in the future though.
> 

OK, sounds good.

> > (2) in fact, not only function params can be committed, but also function's locals and stack operands. Now they are now committed only when GC feature is enabled: `aot_gen_commit_values()` is called with `if (comp_ctx->enable_gc && !aot_gen_commit_values(comp_ctx->aot_frame))` like code. do you think that we should enable committing them and change `params` to other word like `values`?
> 
> I think we could rename it to values and cover more data; if needed this can be more granular in the future
> 

Yes, agree to rename to values.

> > (3) by default, GC should enable sp and params (and locals, stack operands), or AOT runtime may run failed. So had better set them true when GC is enabled, no matter what user enables them or not in this flag?
> 
> Good to know. But isn't stack pointer and stack operands only actually used in the GC mode? If so, should the SP option be deleted at all?

Yes, normally sp and stack operands should be committed at the same time, or it is meaningless for the runtime, and now they are really only used in the GC mode. Agree to remove sp flag and commit sp when committing `values` is enabled, and set `values` to true when GC is enabled. I am not sure whether in the future, the stack operands will be used in other features, like a break point in AOT debugger to view the locals and stacks operands, or dumping more details when an exception is thrown, so had better also keep values flag in `--call-stack-features=`?

loganek: Agree, I'll make the changes once those two PRs (https://github.com/bytecodealliance/wasm-micro-runtime/pull/3768, https://github.com/bytecodealliance/wasm-micro-runtime/pull/3767) to avoid some potential merge conflicts.

loganek: @wenyongh I just made an update, please let me know if you have any feedback.

wenyongh (approved):
LGTM

core/iwasm/include/aot_comp_option.h
wenyongh: It is good to me. I am not sure whether others have comments, let's merge it late today or early tomorrow if there is no new comment?

no1wudi (approved):

core/iwasm/include/aot_comp_option.h
loganek: sounds good. Once this is merged, I'll open another CR with a tiny stack frame then.

PR #3764, opened 2024-09-02
Update ref to the multi-memory tests

loganek:
The specific commit has been deleted, I am pointing to the same commit in the main branch though. This should fix the build e.g. for this PR: https://github.com/bytecodealliance/wasm-micro-runtime/pull/3763

### DIFF
--- a/tests/wamr-test-suites/test_wamr.sh
+++ b/tests/wamr-test-suites/test_wamr.sh
@@ -511,7 +511,7 @@ function spec_test()
         pushd spec
 
         # Reset to commit: "Merge pull request #48 from backes/specify-memcpy-immediate-order"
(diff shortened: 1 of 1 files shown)


loganek:
This seem to be failing for unrelated reason. @wenyongh is there any known issue with running tests on nuttx?

wenyongh:
> This seem to be failing for unrelated reason. @wenyongh is there any known issue with running tests on nuttx?

The CI spec_test_on_nuttx.yml really often ran failed these days, I had to retry several times. I guessed it is caused by the test script, e.g., the timeout to wait for the compilation of AOT compiler is too short and the script reports error before the compilation finishes. But we need to confirm, I will check whether I can fix it.

wenyongh (approved):

PR #3768, opened 2024-09-04
Refactoring: re-use commit IP functionality between exception handling and other cases

loganek:


core/iwasm/compilation/aot_compiler.c
```diff
@@ -630,10 +649,7 @@ aot_gen_commit_sp_ip(AOTCompFrame *frame, bool commit_sp, bool commit_ip)
             return false;
         }
 
-        if (!LLVMBuildStore(comp_ctx->builder, value, value_ptr)) {
-            aot_set_last_error("llvm build store failed");
-            return false;
-        }
+        return aot_gen_commit_ip(comp_ctx, func_ctx, value, is_64bit);
```
wenyongh, line 652: Should check return value, like `if (!aot_gen_commit_ip(...)) return false;`, but not return directly. The below code will commit sp.

loganek: thanks, updated

wenyongh (approved):
LGTM

PR #3773, opened 2024-09-05
AOT call stack optimizations

loganek:
- implemented TINY / STANDARD frame modes - tiny mode is only able to keep track on the IP and func idx, STANDARD mode provides more capabilities (parameters, stack pointer etc)
 - implemented FRAME_PER_FUNCTION / FRAME_PER_CALL modes - frame per function adds code at the beginning and at the end of each function for allocating / deallocating stack frame, whereas in per-call mode the frame is allocated before each call. The exception is call to imported function, where frame-per-function mode also allocates the stack before the `call` instruction (as it can't instrument the imported function)

At the moment TINY + FRAME_PER_FUNCTION is automatically enabled in case GC or perf profiling are disabled and `values` call stack feature is not requested. In all the other cases STANDARD + FRAME_PER_CALL is used.

STANDARD + FRAME_PER_FUNCTION and TINY + FRAME_PER_CALL are currently not implemented but possible, and might be enabled in the future.

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -3864,33 +3864,6 @@ aot_free_frame(WASMExecEnv *exec_env)
     }
 }
 
-/**
- * Releases a frame from the top of the stack only if the top frame
- * describes imported function.
- *
- * This function is used for a frame-per-function in the call_indirect
- * implementation; it's not possible to know at compile time if the the
- * opcode calls to imported function (where the frame must be released,
- * because frame allocation is not part of the function) or the function
- * within the module (in that case the function itself is responsible for
- * allocating / freeing the frame), so we need to check it at runtime.
- */
-void
-aot_free_import_frame(WASMExecEnv *exec_env)
-{
-    void *frame = get_top_frame(exec_env);
-    if (frame) {
-        AOTModule *module =
-            (AOTModule *)((AOTModuleInstance *)exec_env->module_inst)->module;
-        uint32 func_idx = is_tiny_frame(exec_env)
-                              ? ((AOTTinyFrame *)frame)->func_index
-                              : ((AOTFrame *)frame)->func_index;
-        if (func_idx < module->import_func_count) {
-            aot_free_frame(exec_env);
-        }
-    }
-}
-
 void
 aot_frame_update_profile_info(WASMExecEnv *exec_env, bool alloc_frame)
 {
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -4,6 +4,7 @@
  */
 
 #include "aot_runtime.h"
+#include "../compilation/aot_stack_frame.h"
 #include "bh_log.h"
 #include "mem_alloc.h"
 #include "../common/wasm_runtime_common.h"
@@ -72,6 +73,10 @@ bh_static_assert(offsetof(AOTFrame, sp) == sizeof(uintptr_t) * 5);
 bh_static_assert(offsetof(AOTFrame, frame_ref) == sizeof(uintptr_t) * 6);
 bh_static_assert(offsetof(AOTFrame, lp) == sizeof(uintptr_t) * 7);
 
+bh_static_assert(offsetof(AOTTinyFrame, func_index) == sizeof(uint32) * 0);
+bh_static_assert(offsetof(AOTTinyFrame, ip_offset) == sizeof(uint32) * 1);
+bh_static_assert(sizeof(AOTTinyFrame) == sizeof(uint32) * 2);
+
 static void
 set_error_buf(char *error_buf, uint32 error_buf_size, const char *string)
 {
@@ -109,6 +114,55 @@ runtime_malloc(uint64 size, char *error_buf, uint32 error_buf_size)
     memset(mem, 0, (uint32)size);
     return mem;
 }
+
+#if WASM_ENABLE_AOT_STACK_FRAME != 0
+static bool
+is_tiny_frame(WASMExecEnv *exec_env)
+{
+    AOTModule *module =
+        (AOTModule *)((AOTModuleInstance *)exec_env->module_inst)->module;
+
+    return module->feature_flags & WASM_FEATURE_TINY_STACK_FRAME;
+}
+
+static bool
+is_frame_per_function(WASMExecEnv *exec_env)
+{
+    AOTModule *module =
+        (AOTModule *)((AOTModuleInstance *)exec_env->module_inst)->module;
+
+    return module->feature_flags & WASM_FEATURE_FRAME_PER_FUNCTION;
+}
+
+static void *
+get_top_frame(WASMExecEnv *exec_env)
+{
+    if (is_tiny_frame(exec_env)) {
+        return exec_env->wasm_stack.top > exec_env->wasm_stack.bottom
+                   ? exec_env->wasm_stack.top - sizeof(AOTTinyFrame)
+                   : NULL;
+    }
+    else {
+        return exec_env->cur_frame;
+    }
+}
+
+static void *
+get_prev_frame(WASMExecEnv *exec_env, void *cur_frame)
+{
+    bh_assert(cur_frame);
+
+    if (is_tiny_frame(exec_env)) {
+        if ((uint8 *)cur_frame == exec_env->wasm_stack.bottom) {
+            return NULL;
+        }
+        return ((AOTTinyFrame *)cur_frame) - 1;
+    }
+    else {
+        return ((AOTFrame *)cur_frame)->prev_frame;
+    }
+}
+#endif
 
 static bool
 check_global_init_expr(const AOTModule *module, uint32 global_index,
@@ -2265,7 +2319,7 @@ aot_call_function(WASMExecEnv *exec_env, AOTFunctionInstance *function,
         uint32 ext_ret_cell = wasm_get_cell_num(ext_ret_types, ext_ret_count);
         uint64 size;
 #if WASM_ENABLE_AOT_STACK_FRAME != 0
-        struct WASMInterpFrame *prev_frame = exec_env->cur_frame;
+        void *prev_frame = get_top_frame(exec_env);
 #endif
 
         /* Allocate memory all arguments */
@@ -2296,7 +2350,8 @@ aot_call_function(WASMExecEnv *exec_env, AOTFunctionInstance *function,
         }
 
 #if WASM_ENABLE_AOT_STACK_FRAME != 0
-        if (!aot_alloc_frame(exec_env, function->func_index)) {
+        if (!is_frame_per_function(exec_env)
+            && !aot_alloc_frame(exec_env, function->func_index)) {
             if (argv1 != argv1_buf)
                 wasm_runtime_free(argv1);
             return false;
@@ -2324,7 +2379,7 @@ aot_call_function(WASMExecEnv *exec_env, AOTFunctionInstance *function,
         /* Free all frames allocated, note that some frames
            may be allocated in AOT code and haven't been
            freed if exception occurred */
-        while (exec_env->cur_frame != prev_frame)
+        while (get_top_frame(exec_env) != prev_frame)
             aot_free_frame(exec_env);
 #endif
         if (!ret) {
@@ -2367,9 +2422,12 @@ aot_call_function(WASMExecEnv *exec_env, AOTFunctionInstance *function,
     }
     else {
 #if WASM_ENABLE_AOT_STACK_FRAME != 0
-        struct WASMInterpFrame *prev_frame = exec_env->cur_frame;
-
-        if (!aot_alloc_frame(exec_env, function->func_index)) {
+        void *prev_frame = get_top_frame(exec_env);
+        /* Only allocate frame for frame-per-call mode; in the
+           frame-per-function mode the frame is allocated at the
+           beginning of the function. */
+        if (!is_frame_per_function(exec_env)
+            && !aot_alloc_frame(exec_env, function->func_index)) {
             return false;
         }
 #endif
@@ -2394,7 +2452,7 @@ aot_call_function(WASMExecEnv *exec_env, AOTFunctionInstance *function,
         /* Free all frames allocated, note that some frames
            may be allocated in AOT code and haven't been
            freed if exception occurred */
-        while (exec_env->cur_frame != prev_frame)
+        while (get_top_frame(exec_env) != prev_frame)
             aot_free_frame(exec_env);
 #endif
 
@@ -2880,7 +2938,7 @@ aot_invoke_native(WASMExecEnv *exec_env, uint32 func_idx, uint32 argc,
             goto fail;
         }
 #if WASM_ENABLE_AOT_STACK_FRAME != 0
-        struct WASMInterpFrame *prev_frame = exec_env->cur_frame;
+        void *prev_frame = get_top_frame(exec_env);
 
         if (!aot_alloc_frame(exec_env, func_idx)) {
             goto fail;
@@ -2894,7 +2952,7 @@ aot_invoke_native(WASMExecEnv *exec_env, uint32 func_idx, uint32 argc,
         /* Free all frames allocated, note that some frames
            may be allocated in AOT code and haven't been
            freed if exception occurred */
-        while (exec_env->cur_frame != prev_frame)
+        while (get_top_frame(exec_env) != prev_frame)
             aot_free_frame(exec_env);
 #endif
     }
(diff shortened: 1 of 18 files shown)


loganek:
@wenyongh I'm not 100% sure if I haven't missed any places where I should release the frame so I'd appreciate if you feedback especially in that area. Thanks

loganek:
spec tests are failing on nuttx, likely unrelated to this issue and I hope this can fix it: https://github.com/bytecodealliance/wasm-micro-runtime/pull/3771

core/iwasm/aot/aot_runtime.c
```diff
@@ -2880,7 +2937,7 @@ aot_invoke_native(WASMExecEnv *exec_env, uint32 func_idx, uint32 argc,
             goto fail;
         }
 #if WASM_ENABLE_AOT_STACK_FRAME != 0
-        struct WASMInterpFrame *prev_frame = exec_env->cur_frame;
+        void *prev_frame = get_top_frame(exec_env);
 
         if (!aot_alloc_frame(exec_env, func_idx)) {
```
wenyongh, line 2942: Should check `!is_frame_per_function(exec_env)`?

core/iwasm/aot/aot_runtime.c
```diff
@@ -3751,12 +3845,50 @@ aot_free_frame_internal(WASMExecEnv *exec_env)
...
+ * Releases a frame from the top of the stack only if the top frame
+ * describes imported function.
+ *
+ * This function is used for a frame-per-function in the call_indirect
+ * implementation; it's not possible to know at compile time if the the
+ * opcode calls to imported function (where the frame must be released,
+ * because frame allocation is not part of the function) or the function
+ * within the module (in that case the function itself is responsible for
+ * allocating / freeing the frame), so we need to check it at runtime.
+ */
```
wenyongh, line 3876: If the frame of the import function isn't allocated, there may be several issues: (1) the dumped call stacks may be not complete, (2) for GC, the native function's parameters are not committed frame and AOT runtime can not traversing them and add them to GC's root set, GC reclaim process may be invalid, (3) for performance profiling, the execution time of the import function isn't found.
Can we allocate the import function's frame before calling the function and freeing after calling the function? Like what current code does for both import and non-import functions.

core/iwasm/compilation/aot_stack_frame.h
```diff
@@ -0,0 +1,27 @@
...
+extern "C" {
+#endif
+
+typedef struct {
+    /* The non-imported function index of current function */
+    uintptr_t func_index;
+
+    /* Instruction pointer: offset to the bytecode array */
+    uintptr_t ip_offset;
+} AOTTinyFrame;
```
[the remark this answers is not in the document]
loganek: sure, updated

core/iwasm/aot/aot_runtime.c
loganek: > Can we allocate the import function's frame before calling the function and freeing after calling the function? Like what current code does for both import and non-import functions.

@wenyongh that's exactly what happens now; for `call` opcode we have:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/eb4d5bbc8aedef23bf2864c7923ddbc8300004de/core/iwasm/compilation/aot_emit_function.c#L1485-L1503

so the frame will be allocated using `aot_alloc_frame_per_function_frame_for_aot_func` if this is imported function. For `call_indirect` there's *always* a call to `aot_alloc_frame` (there's no condition that checks if it's `frame-per-function` mode), and in the `aot_alloc_frame` the frame allocation is only skipped for non-imported function; otherwise it's allocated:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/eb4d5bbc8aedef23bf2864c7923ddbc8300004de/core/iwasm/aot/aot_runtime.c#L3803-L3822

core/iwasm/aot/aot_runtime.c
```diff
@@ -2367,9 +2421,12 @@ aot_call_function(WASMExecEnv *exec_env, AOTFunctionInstance *function,
...
     else {
 #if WASM_ENABLE_AOT_STACK_FRAME != 0
-        struct WASMInterpFrame *prev_frame = exec_env->cur_frame;
-
-        if (!aot_alloc_frame(exec_env, function->func_index)) {
+        void *prev_frame = get_top_frame(exec_env);
+        /* Only allocate frame for frame-per-call mode; in the
+           frame-per-function mode the frame is allocated at the
+           beginning of the function. */
+        if (!is_frame_per_function(exec_env)
```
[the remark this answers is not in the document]
loganek: good point, updated

core/iwasm/aot/aot_runtime.c
loganek: no, we do want to allocate frame in the `call` instruction for the imported and/or native functions even in the `frame-per-function` mode, otherwise this function will not be included in the callstack (see my comment https://github.com/bytecodealliance/wasm-micro-runtime/pull/3773#discussion_r1746724221)

core/iwasm/aot/aot_runtime.c
wenyongh: Yes, but I think we can add check in the compiled LLVM IR of opcode call_indirect: compare the function index with import_func_count, and call aot_alloc_frame if func_idx < import_func_count. It doesn't necessary always calling aot_alloc_frame.

And seems no need to add the flag `comp_ctx->call_stack_features.frame_per_function`, my idea is that: (1) in the new code of aot compiler, for both full frame and tiny frame, (a) always allocate frame in the beginning of a non-imported function, (b) always allocate frame before calling the imported function (in both op call and op call_indirect), (c) always emit the WASM_FEATURE_FRAME_PER_FUNCTION flag in the aot file. This simplify the aot compilation and seems no need to aot_alloc_import_frame.

(2) in aot runtime, (a) always aot_alloc_frame before calling import function, (b) aot_alloc_frame before calling non-import function if WASM_FEATURE_FRAME_PER_FUNCTION flag isn't detected, so as to keep backward compatibility.

wenyongh: I think it should be simpler, please correct me if I am wrong or there is something I missed.

loganek: > Yes, but I think we can add check in the compiled LLVM IR of opcode call_indirect: compare the function index with import_func_count, and call aot_alloc_frame if func_idx < import_func_count. It doesn't necessary always calling aot_alloc_frame.

That's I think possible, but perhaps will add slightly more code. I don't mind changing it though as in our usecase `call_indirect` are quite rare.

> And seems no need to add the flag comp_ctx->call_stack_features.frame_per_function

Yes, it could potentially be removed; however, I didn't do it (just yet) due to multimodule / dynamic linking - if we always use frame-per-function mode, the callstack will contain duplicates. So I think we should keep both modes in the compiler for now until we figure out how to do it correctly; there are few options that come to my mind:
* explicitly expose frame-per-function as parameter to wamrc, so users can choose the mode
* allow users to define which functions are provided by another module and which functions are imported from native or modules without instrumentation

Either is fine for me, but also I don't think any of the option needs to be implemented right now and can be added in the future without breaking the ABI. What do you think?

wenyongh: Yes, for me the first option (expose frame-per-function as parameter to wamrc) is better and can be added in the future. The latter may be a little complex for developer to choose which functions?

loganek: agree, I'll make it a follow-up task for the https://github.com/bytecodealliance/wasm-micro-runtime/issues/3758

loganek: @wenyongh I also removed `aot_free_import_frame` function and moved the allocation of the frame for imported function to the AOT code. Let me know if you have any questions.

core/iwasm/compilation/aot_emit_function.c
```diff
@@ -2534,6 +2565,12 @@ aot_compile_op_call_indirect(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
         goto fail;
     }
 
+    if (comp_ctx->aot_frame
+        && comp_ctx->call_stack_features.frame_per_function) {
+        aot_alloc_frame_per_function_frame_for_aot_func(comp_ctx, func_ctx,
+                                                        func_idx);
```
wenyongh, line 2571: commit_params_to_frame_of_import_func was called in L2542, which assumes that the frame has been allocated, so here we should move these lines to be fore L2542?

And had better check the return value of aot_alloc_frame_per_function_frame_for_aot_func, if failed, goto fail

core/iwasm/compilation/aot_emit_function.c
```diff
@@ -2545,6 +2582,11 @@ aot_compile_op_call_indirect(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
         && !check_call_return(comp_ctx, func_ctx, res))
         goto fail;
 
+    if (comp_ctx->aot_frame
+        && comp_ctx->call_stack_features.frame_per_function) {
+        aot_free_frame_per_function_frame_for_aot_func(comp_ctx, func_ctx);
```
wenyongh, line 2587: had better check return value

core/iwasm/compilation/aot_emit_function.c
```diff
@@ -3133,7 +3177,7 @@ aot_compile_op_call_ref(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
         PUSH(result_phis[i], func_type->types[func_param_count + i]);
     }
 
-    if (comp_ctx->enable_aux_stack_frame) {
+    if (comp_ctx->aux_stack_frame_type == AOT_STACK_FRAME_TYPE_STANDARD) {
```
wenyongh, line 3180: Should be `if (comp_ctx->aux_stack_frame_type
        && !comp_ctx->call_stack_features.frame_per_function) {` like L2982?

core/iwasm/compilation/aot_stack_frame_comp.c
```diff
@@ -0,0 +1,146 @@
...
+aot_free_tiny_frame_for_aot_func(AOTCompContext *comp_ctx,
+                                 AOTFuncContext *func_ctx)
+{
+    LLVMValueRef wasm_stack_top_ptr = func_ctx->wasm_stack_top_ptr,
+                 wasm_stack_top;
+    LLVMValueRef offset;
+
+    ADD_LOAD(wasm_stack_top, INT8_PTR_TYPE, wasm_stack_top_ptr);
+
+    offset = I32_CONST(-sizeof(AOTTinyFrame));
```
wenyongh, line 93: not sure whether `-sizeof` is an issue for 64-bit, maybe it will be convert to positive u64 for LLVMBuildInBoundsGEP? Is better `offset = comp_ctx->pointer_size == 8 ? I64_CONST(-sizeof(AOTTinyFrame) : I32_CONST(-sizeof(AOTTinyFrame))`?

And had better check the return value

core/iwasm/compilation/aot_stack_frame_comp.c
```diff
@@ -0,0 +1,146 @@
...
+aot_alloc_frame_per_function_frame_for_aot_func(AOTCompContext *comp_ctx,
+                                                AOTFuncContext *func_ctx,
+                                                LLVMValueRef func_index)
+{
+    switch (comp_ctx->aux_stack_frame_type) {
+        case AOT_STACK_FRAME_TYPE_TINY:
+            return aot_alloc_tiny_frame_for_aot_func(comp_ctx, func_ctx,
+                                                     func_index);
+        default:
+            aot_set_last_error("unsupported mode");
```
wenyongh, line 130: Seems we can export `alloc_frame_for_aot_func` in aot_emit_function.c and call it here for AOT_STACK_FRAME_TYPE_STANDARD, but need to confirm.

core/iwasm/compilation/aot_stack_frame_comp.c
```diff
@@ -0,0 +1,146 @@
...
+
+bool
+aot_free_frame_per_function_frame_for_aot_func(AOTCompContext *comp_ctx,
+                                               AOTFuncContext *func_ctx)
+{
+    switch (comp_ctx->aux_stack_frame_type) {
+        case AOT_STACK_FRAME_TYPE_TINY:
+            return aot_free_tiny_frame_for_aot_func(comp_ctx, func_ctx);
+        default:
+            aot_set_last_error("unsupported mode");
```
wenyongh, line 143: Same as above, maybe we can call free_frame_for_aot_func for AOT_STACK_FRAME_TYPE_STANDARD.

core/iwasm/compilation/aot_stack_frame_comp.c
loganek: yes, that'd likely be the case, however, for now the STANDARD frame is not supported for frame-per-function, and I just left it as 'unsupported'. It's already a follow-up task in https://github.com/bytecodealliance/wasm-micro-runtime/issues/3758

core/iwasm/compilation/aot_stack_frame_comp.c
loganek: see my comment above.

core/iwasm/compilation/aot_emit_control.c
```diff
@@ -814,6 +815,13 @@ aot_compile_op_end(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
         if ((next_llvm_end_block = find_next_llvm_end_block(block)))
             MOVE_BLOCK_BEFORE(block->llvm_end_block, next_llvm_end_block);
     }
+    if (block->label_type == LABEL_TYPE_FUNCTION
+        && comp_ctx->aux_stack_frame_type
+        && comp_ctx->call_stack_features.frame_per_function
+        && !aot_free_frame_per_function_frame_for_aot_func(comp_ctx,
+                                                           func_ctx)) {
```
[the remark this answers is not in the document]
loganek: That's a good point. However, that will make some of the `func_end` blocks that contain `phi` instructions invalid, as `PHI` must be always placed at the beginning of the block (see https://llvm.org/docs/LangRef.html#phi-instruction). In addition to what you've posted above, I'll also have to modify `CREATE_RESULT_VALUE_PHIS` macro so it inserts PHI instruction before any other instructions.

core/iwasm/compilation/aot_emit_control.c
```diff
@@ -1361,6 +1369,13 @@ aot_compile_op_return(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
         (*p_frame_ip - 1) - comp_ctx->comp_data->wasm_module->buf_code);
 #endif
 
+    if (comp_ctx->aux_stack_frame_type
+        && comp_ctx->call_stack_features.frame_per_function
+        && !aot_free_frame_per_function_frame_for_aot_func(comp_ctx,
+                                                           func_ctx)) {
```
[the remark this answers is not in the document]
loganek: I think it's a nice optimization idea, but if you don't mind I'd do it as follow-up to keep the PR as small as possible. I'll make a task in the https://github.com/bytecodealliance/wasm-micro-runtime/issues/3758 so I don't forget about it.

core/iwasm/compilation/aot_emit_control.c
```diff
@@ -38,13 +39,24 @@ format_block_name(char *name, uint32 name_size, uint32 block_index,
...
+            goto fail;                                                       \
+        }                                                                    \
+        if (!strcmp(name, "func_end") && comp_ctx->aux_stack_frame_type      \
+            && comp_ctx->call_stack_features.frame_per_function) {           \
+            LLVMBasicBlockRef cur_block =                                    \
+                LLVMGetInsertBlock(comp_ctx->builder);                       \
+            SET_BUILDER_POS(new_llvm_block);                                 \
+            if (!aot_free_frame_per_function_frame_for_aot_func(comp_ctx,    \
+                                                                func_ctx)) { \
+                return false;                                                \
```
wenyongh, line 56: had better `got fail`, some resources may be destroyed in label `fail`.

core/iwasm/compilation/aot_emit_control.c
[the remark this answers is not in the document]
wenyongh: OK, got it, thanks.

core/iwasm/compilation/aot_emit_control.c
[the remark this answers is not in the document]
wenyongh: OK, that's good.

core/iwasm/compilation/aot_emit_function.c
```diff
@@ -2508,7 +2541,13 @@ aot_compile_op_call_indirect(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
     /* Translate call import block */
     LLVMPositionBuilderAtEnd(comp_ctx->builder, block_call_import);
 
-    if (comp_ctx->enable_aux_stack_frame
+    if (comp_ctx->aot_frame && comp_ctx->call_stack_features.frame_per_function
+        && !aot_alloc_frame_per_function_frame_for_aot_func(comp_ctx, func_ctx,
+                                                            func_idx)) {
+        return false;
```
wenyongh, line 2547: had better `goto fail`

wenyongh (commented):
LGTM with minor issues.

PR #3778, opened 2024-09-09
support WAMR AOT module debug.

LevelCA:
1. WAMR aot module debug workflow like this:
aot


yamt:
an interesting idea.
do you have some automation for the gdb side?

LevelCA:
> an interesting idea. do you have some automation for the gdb side?

Thanks. yeap.
When an AOT module was loaded, it triggers a breakpoint set by the Python script at __loadable_wamr_module_register_code_ptr on the PC, this breakpoint initializes the module when GDB starts in debugging mode on the PC, implement gdb custom commands, automatically read local files with debugging information, and compare them with loaded files to achieve the purpose of debugging aot, the entire workflow is shown in the figure above.

yamt:
> > an interesting idea. do you have some automation for the gdb side?
> 
> Thanks. yeap. When an AOT module was loaded, it triggers a breakpoint set by the Python script at __loadable_wamr_module_register_code_ptr on the PC, this breakpoint initializes the module when GDB starts in debugging mode on the PC, implement gdb custom commands, automatically read local files with debugging information, and compare them with loaded files to achieve the purpose of debugging aot, the entire workflow is shown in the figure above.

is your python script something which can be included in this PR?

PR #3783, opened 2024-09-10
Appease GCC -Wformat

yamt:
I'm not sure we want to use C99 %tu here.
While C99 %zu is more widely used in WAMR, %tu is rare (if any) and I'm not sure if it's ubiquitously implemented in platforms we support.


wenyongh (commented):
LGTM

PR #3785, opened 2024-09-10
Implement option for skipping function index in the callstack

loganek:
This helps reducing the size of the generated source code even further.

Also add a script that converts instruction pointers to function indexes (or function names)


wenyongh (approved):
LGTM

PR #3786, opened 2024-09-11
Add memory instance support apis

bnason-nf:
Now that WAMR supports multiple memory instances, this adds some APIs to access them in a standard way.

This involved moving some existing utility functions out from the `WASM_ENABLE_MULTI_MODULE` blocks they were nested in, but multi-memory and multi-module seem independent as far as I can tell so I assume that's okay.

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -1060,6 +1060,8 @@ aot_lookup_memory(AOTModuleInstance *module_inst, char const *name)
     return NULL;
 #else
     (void)module_inst->export_memories;
+    if (!module_inst->memories)
+        return NULL;
     return module_inst->memories[0];
 #endif
 }
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -1049,11 +1049,36 @@ fail1:
     return NULL;
 }
 
-static AOTMemoryInstance *
+AOTMemoryInstance *
+aot_lookup_memory(AOTModuleInstance *module_inst, char const *name)
+{
+#if WASM_ENABLE_MULTI_MEMORY != 0
+    uint32 i;
+    for (i = 0; i < module_inst->export_memory_count; i++)
+        if (!strcmp(module_inst->export_memories[i].name, name))
+            return module_inst->export_memories[i].memory;
+    return NULL;
+#else
+    (void)module_inst->export_memories;
+    return module_inst->memories[0];
+#endif
+}
+
+AOTMemoryInstance *
 aot_get_default_memory(AOTModuleInstance *module_inst)
 {
     if (module_inst->memories)
         return module_inst->memories[0];
+    else
+        return NULL;
+}
+
+AOTMemoryInstance *
+aot_get_memory_with_index(AOTModuleInstance *module_inst, uint32 index)
+{
+    bh_assert(index < module_inst->memory_count);
+    if (module_inst->memories)
+        return module_inst->memories[index];
     else
         return NULL;
 }
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -1078,11 +1078,9 @@ aot_get_default_memory(AOTModuleInstance *module_inst)
 AOTMemoryInstance *
 aot_get_memory_with_index(AOTModuleInstance *module_inst, uint32 index)
 {
-    bh_assert(index < module_inst->memory_count);
-    if (module_inst->memories)
-        return module_inst->memories[index];
-    else
+    if ((index >= module_inst->memory_count) || !module_inst->memories)
         return NULL;
+    return module_inst->memories[index];
 }
 
 static bool
--- a/core/iwasm/common/wasm_memory.c
+++ b/core/iwasm/common/wasm_memory.c
@@ -963,16 +963,16 @@ wasm_enlarge_memory(WASMModuleInstance *module, uint32 inc_page_count)
 {
     bool ret = false;
 
+    if (module->memory_count > 0) {
 #if WASM_ENABLE_SHARED_MEMORY != 0
-    if (module->memory_count > 0)
         shared_memory_lock(module->memories[0]);
 #endif
-    ret = wasm_enlarge_memory_internal((WASMModuleInstanceCommon *)module,
-                                       module->memories[0], inc_page_count);
+        ret = wasm_enlarge_memory_internal((WASMModuleInstanceCommon *)module,
+                                           module->memories[0], inc_page_count);
 #if WASM_ENABLE_SHARED_MEMORY != 0
-    if (module->memory_count > 0)
         shared_memory_unlock(module->memories[0]);
 #endif
+    }
 
     return ret;
 }
@@ -983,17 +983,17 @@ wasm_enlarge_memory_with_idx(WASMModuleInstance *module, uint32 inc_page_count,
 {
     bool ret = false;
 
+    if (memidx < module->memory_count) {
 #if WASM_ENABLE_SHARED_MEMORY != 0
-    if (memidx < module->memory_count)
         shared_memory_lock(module->memories[memidx]);
 #endif
-    ret =
-        wasm_enlarge_memory_internal((WASMModuleInstanceCommon *)module,
-                                     module->memories[memidx], inc_page_count);
+        ret = wasm_enlarge_memory_internal((WASMModuleInstanceCommon *)module,
+                                           module->memories[memidx],
+                                           inc_page_count);
 #if WASM_ENABLE_SHARED_MEMORY != 0
-    if (memidx < module->memory_count)
         shared_memory_unlock(module->memories[memidx]);
 #endif
+    }
 
     return ret;
 }
--- a/core/iwasm/common/wasm_memory.c
+++ b/core/iwasm/common/wasm_memory.c
@@ -756,7 +756,7 @@ wasm_mmap_linear_memory(uint64_t map_size, uint64 commit_size)
     return wasm_mremap_linear_memory(NULL, 0, map_size, commit_size);
 }
 
-bool
+static bool
 wasm_enlarge_memory_internal(WASMModuleInstanceCommon *module,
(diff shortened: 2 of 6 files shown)


lum1n0us:
It seems that functions(and code) related to exporting should not be limited by the `WASM_ENABLE_MULTI_MODULE`.

core/iwasm/include/wasm_export.h
```diff
@@ -939,6 +943,102 @@ WASM_RUNTIME_API_EXTERN void
...
+ * @brief Lookup a memory instance by name
+ *
+ * @param module_inst The module instance
+ * @param name The name of the memory instance
+ *
+ * @return The memory instance if found, NULL otherwise
+ */
+WASM_RUNTIME_API_EXTERN wasm_memory_inst_t
+wasm_runtime_memory_inst_lookup(const wasm_module_inst_t module_inst,
+                                const char *name);
```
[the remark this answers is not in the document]
wenyongh: And there is `wasm_runtime_enlarge_memory`, maybe we can just skip `_inst`?
```C
wasm_runtime_lookup_memory
wasm_runtime_get_default_memory
wasm_runtime_get_memory
wasm_memory_get_cur_page_count
wasm_memory_get_max_page_count
wasm_memory_get_num_bytes_per_page
...
```

core/iwasm/include/wasm_export.h
```diff
@@ -939,6 +943,102 @@ WASM_RUNTIME_API_EXTERN void
...
+ * @brief Enlarge a memory instance by a number of pages
+ *
+ * @param memory_inst The memory instance
+ * @param inc_page_count The number of pages to add
+ *
+ * @return True if successful, false otherwise
+ */
+WASM_RUNTIME_API_EXTERN bool
+wasm_memory_inst_enlarge(wasm_memory_inst_t memory_inst,
+                         uint64_t inc_page_count);
```
wenyongh, line 1040: There is already a function `bool wasm_runtime_enlarge_memory(module_inst, inc_page_count)` but it enlarge the default memory. Do you think it is better to use `wasm_runtime_enlarge_memory_with_idx`?

core/iwasm/aot/aot_runtime.c
```diff
@@ -1049,7 +1049,22 @@ memory_instantiate(AOTModuleInstance *module_inst, AOTModuleInstance *parent,
...
+{
+#if WASM_ENABLE_MULTI_MEMORY != 0
+    uint32 i;
+    for (i = 0; i < module_inst->export_memory_count; i++)
+        if (!strcmp(module_inst->export_memories[i].name, name))
+            return module_inst->export_memories[i].memory;
+    return NULL;
+#else
+    (void)module_inst->export_memories;
+    return module_inst->memories[0];
```
TianlongLiang, line 1063: It's better to check whether `module_inst->memories` is NULL before accessing it, like in `aot_get_default_memory`

The same rationale applies to `wasm_get_default_memory`

core/iwasm/aot/aot_runtime.c
```diff
@@ -1058,6 +1073,16 @@ aot_get_default_memory(AOTModuleInstance *module_inst)
         return NULL;
 }
 
+AOTMemoryInstance *
+aot_get_memory_with_index(AOTModuleInstance *module_inst, uint32 index)
+{
+    bh_assert(index < module_inst->memory_count);
```
TianlongLiang, line 1079: I think if we want to make the `aot_get_memory_with_index` export API, we need to check the index in the below if statement rather than simply assert. In the past, the loader will basically check and make sure the index is not out of bounds,  if now the user will provide the index, we better to further check it. 

The same rationale applies to `wasm_get_memory_with_index`

core/iwasm/common/wasm_memory.c
```diff
@@ -975,7 +969,8 @@ wasm_enlarge_memory(WASMModuleInstance *module, uint32 inc_page_count)
     if (module->memory_count > 0)
         shared_memory_lock(module->memories[0]);
 #endif
-    ret = wasm_enlarge_memory_internal(module, inc_page_count, 0);
+    ret = wasm_enlarge_memory_internal((WASMModuleInstanceCommon *)module,
+                                       module->memories[0], inc_page_count);
```
TianlongLiang, line 973: Since it's an export API, we also need to check whether `module->memories` is valid.

core/iwasm/common/wasm_memory.c
```diff
@@ -994,7 +989,9 @@ wasm_enlarge_memory_with_idx(WASMModuleInstance *module, uint32 inc_page_count,
     if (memidx < module->memory_count)
         shared_memory_lock(module->memories[memidx]);
 #endif
-    ret = wasm_enlarge_memory_internal(module, inc_page_count, memidx);
+    ret =
+        wasm_enlarge_memory_internal((WASMModuleInstanceCommon *)module,
+                                     module->memories[memidx], inc_page_count);
```
TianlongLiang, line 994: If it will be an export API, we also need to check whether `module->memories[memidx` is valid. Ignore me if this won't be called in export API :)

core/iwasm/common/wasm_memory.c
```diff
@@ -1003,6 +1000,104 @@ wasm_enlarge_memory_with_idx(WASMModuleInstance *module, uint32 inc_page_count,
...
+
+bool
+wasm_memory_inst_enlarge(WASMMemoryInstance *memory, uint64 inc_page_count)
+{
+    bool ret;
+
+#if WASM_ENABLE_SHARED_MEMORY != 0
+    shared_memory_lock(memory);
+#endif
+    ret = wasm_enlarge_memory_internal(NULL, memory, (uint32)inc_page_count);
```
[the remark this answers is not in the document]
bnason-nf: @wenyongh in order to support externally created (imported) memory objects, you have to be able to create them independent of a module or module instance, so I think we should leave this API without a required module argument. However, I'm OK with adding a module argument and making it optional.

core/iwasm/common/wasm_memory.c
```diff
@@ -913,21 +908,20 @@ wasm_enlarge_memory_internal(WASMModuleInstance *module, uint32 inc_page_count,
     wasm_runtime_set_mem_bound_check_bytes(memory, total_size_new);
 
 return_func:
-    if (!ret && enlarge_memory_error_cb) {
+    if (!ret && module && enlarge_memory_error_cb) {
```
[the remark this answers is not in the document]
bnason-nf: This relates to the previous comments, if we added a module argument to `wasm_memory_inst_enlarge()`, I think it should be optional. If so, we would still need to check it here.

core/iwasm/include/wasm_export.h
bnason-nf: This relates to previous comments:

https://github.com/bytecodealliance/wasm-micro-runtime/pull/3786#discussion_r1759226339
https://github.com/bytecodealliance/wasm-micro-runtime/pull/3786#discussion_r1759226373

For memories created independently of a module or module instance, providing an index has no meaning, so I think this API should remain like this.

core/iwasm/common/wasm_memory.c
bnason-nf: I think after the fixes I made for your previous comments, there are no longer any cases that can call `wasm_enlarge_memory_internal()` without first checking that the memory instance arg is valid. Please let me know if you think I missed something though. However, the first block of code in `wasm_enlarge_memory_internal()` checks that memory is valid so I think it should be safe regardless.

```
    if (!memory) {
        ret = false;
        goto return_func;
    }
```

core/iwasm/aot/aot_runtime.c
```diff
@@ -1049,7 +1049,24 @@ memory_instantiate(AOTModuleInstance *module_inst, AOTModuleInstance *parent,
...
+    uint32 i;
+    for (i = 0; i < module_inst->export_memory_count; i++)
+        if (!strcmp(module_inst->export_memories[i].name, name))
+            return module_inst->export_memories[i].memory;
+    return NULL;
+#else
+    (void)module_inst->export_memories;
+    if (!module_inst->memories)
+        return NULL;
+    return module_inst->memories[0];
```
lum1n0us, line 1065: I believe that functions `lookup_xx()` are intended to focus on exported functions, global variables, tables, and memories, as Spec. required. However, lines 1062 to 1065 seem to return the default memory, which is causing confusion.

bnason-nf: Yes it is a bit confusing, but for this aot_lookup_memory() implementation, I directly mimicked wasm_lookup_memory() which @wenyongh originally implemented. Do you think that implementation needs to be fixed in some way also?

lum1n0us: Oh, I understand the issue now. WAMR currently doesn't support the export and import of memory and tables. That's why we implemented a temporary solution. We're also considering updating our approach to meet the specification requirements. It might be the right moment to correct this. 

However, it's still up to you to choose whether to include this implementation in the current pull request or not.

@wenyongh, what's your opinion?

wenyongh: Agree, had better also modify wasm_lookup_memory (and wasm_lookup_table if needed). But since module_inst->export_memories isn't created, seems we should lookup the module's export table to check whether the memory is exported.

bnason-nf: I implemented exporting of memory and tables a while back. Once this PR lands, I also plan to do another PR to add support for memory imports, and hopefully tables after that.

core/iwasm/common/wasm_memory.c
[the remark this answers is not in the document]
wenyongh: OK, got it.

core/iwasm/aot/aot_runtime.c
lum1n0us: Excellent 🎆. I'm planning to work on implementing instantiation linking, which necessitates complete support for export and import functionality. Your assistance will help expedite this task.

wenyongh: @lum1n0us So let's merge this PR first and wait for @bnason-nf's next PR?

wenyongh (approved):
LGTM

PR #3787, opened 2024-09-11
Ignore temporary file from aider

no1wudi:
Aider is AI pair programming in your terminal: https://aider.chat

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@
 .clangd
 .DS_Store
 *.o
+.aider*
 
 core/deps/**
 core/shared/mem-alloc/tlsf


lum1n0us:
How does aider compare to others, like github copilot, claude-dev, in terms of performance?

no1wudi:
It should be more close to claude-dev with ability to edit multiple file and create new file.

Works fine on my side to execute detemined tasks, such as function refactor, add new member to the struct and update the code that referenced to it, and even to simple debug itself.

But, the accuracy is highly correlated with the model used.

wenyongh (approved):

PR #3790, opened 2024-09-12
Add no_resolve to LoadArgs and wasm_runtime_resolve_symbols so one can

Andersbakken:
delay resolving of symbols.

This is useful for inspecting the module between loading and instantiating.

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -5111,3 +5111,125 @@ aot_get_module_name(AOTModule *module)
 {
     return module->name;
 }
+
+bool
+aot_resolve_symbols(AOTModule *module)
+{
+    bool ret = true;
+    uint32 idx;
+    for (idx = 0; idx < module->import_func_count; ++idx) {
+        AOTImportFunc *aot_import_func = &module->import_funcs[idx];
+        if (!aot_import_func->func_ptr_linked) {
+            if (!aot_resolve_import_func(module, aot_import_func)) {
+                LOG_WARNING("Failed to link function (%s, %s)",
+                            aot_import_func->module_name,
+                            aot_import_func->func_name);
+                ret = false;
+            }
+        }
+    }
+    return ret;
+}
+
+#if WASM_ENABLE_MULTI_MODULE != 0
+static void *
+aot_resolve_function(const AOTModule *module, const char *function_name,
+                     const AOTFuncType *expected_function_type, char *error_buf,
+                     uint32 error_buf_size);
+
+static void *
+aot_resolve_function_ex(const char *module_name, const char *function_name,
+                        const AOTFuncType *expected_function_type,
+                        char *error_buf, uint32 error_buf_size)
+{
+    WASMModuleCommon *module_reg;
+
+    module_reg = wasm_runtime_find_module_registered(module_name);
+    if (!module_reg || module_reg->module_type != Wasm_Module_AoT) {
+        LOG_DEBUG("can not find a module named %s for function %s", module_name,
+                  function_name);
+        set_error_buf(error_buf, error_buf_size, "unknown import");
+        return NULL;
+    }
+    return aot_resolve_function((AOTModule *)module_reg, function_name,
+                                expected_function_type, error_buf,
+                                error_buf_size);
+}
+
+static void *
+aot_resolve_function(const AOTModule *module, const char *function_name,
+                     const AOTFuncType *expected_function_type, char *error_buf,
+                     uint32 error_buf_size)
+{
+    void *function = NULL;
+    AOTExport *export = NULL;
+    AOTFuncType *target_function_type = NULL;
+
+    export = loader_find_export((WASMModuleCommon *)module, module->name,
+                                function_name, EXPORT_KIND_FUNC, error_buf,
+                                error_buf_size);
+    if (!export) {
+        return NULL;
+    }
+
+    /* resolve function type and function */
+    if (export->index < module->import_func_count) {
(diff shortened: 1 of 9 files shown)


core/iwasm/aot/aot_runtime.c
```diff
@@ -5020,3 +5020,121 @@ aot_get_module_name(AOTModule *module)
 {
     return module->name;
 }
+
+void
+aot_resolve_symbols(AOTModule *module)
```
wenyongh, line 5025: Is it better to change the return type to bool for aot_resolve_symbols, aot_resolve_import_func, wasm_resolve_symbols, wasm_resolve_import_func and wasm_runtime_resolve_symbols? So that developer can know whether wasm_runtime_resolve_symbols is successful or not?

And this function we try linking all the functions (don't stop when failed to link a function), and eventually return false if a function is linked unsuccessfully?

core/iwasm/aot/aot_runtime.c
```diff
@@ -5020,3 +5020,121 @@ aot_get_module_name(AOTModule *module)
...
+{
+    for (uint32 idx = 0; idx < module->import_func_count; ++idx) {
```
wenyongh, line 5027: had better declare `uint32 idx` first, it is a C source file, we usually do that way

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -83,6 +83,56 @@ wasm_unload(WASMModule *module)
     wasm_loader_unload(module);
 }
 
+void
+wasm_resolve_symbols(WASMModule *module)
+{
+    for (uint32 idx = 0; idx < module->import_function_count; ++idx) {
```
wenyongh, line 89: had better declare `uint32 idx` first, it is a C source file, we usually do that way

core/iwasm/aot/aot_runtime.c
Andersbakken: Agree on it returning bool. That seems like good information. Failure to link a function is treated as a possibly benign error when not using no_resolve and is indeed sometimes not a problem so I think it's best to keep linking as many functions as possible.

core/iwasm/interpreter/wasm_runtime.c
Andersbakken: Fixed.

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -83,6 +83,125 @@ wasm_unload(WASMModule *module)
...
+        if (!linked && !wasm_resolve_import_func(module, import)) {
+            ret = false;
+        }
+    }
+    return ret;
+}
+
+#if WASM_ENABLE_MULTI_MODULE != 0
+static WASMFunction *
+wasm_runtime_resolve_function(const char *module_name,
```
wenyongh, line 108: Had better change to `wasm_resolve_function`, function `wasm_runtime_xxx` is normally a `core/iwasm/common` layer function for both interpreter and aot. `wasm_xxx` is normally for interpreter.

lum1n0us:
Just out of personal curiosity, may I inquire about the purpose of this pull request? Is there a particular use case that our current loading mechanisms are unable to accommodate?

core/iwasm/aot/aot_runtime.c
Andersbakken: That seems better. Will fix.

Andersbakken:
We're 

> Just out of personal curiosity, may I inquire about the purpose of this pull request? Is there a particular use case that our current loading mechanisms are unable to accommodate?

We're using WAMR along with a scripting runtime and we need to generate dynamic imports. To do this we need to know the function signatures before instantiating the module. So far we've gotten around it by just loading the module first, then inspecting it. Then loading it again. This feels hacky and inefficient. With this functionality we are able to avoid the second load.

core/iwasm/interpreter/wasm_runtime.c
Andersbakken: Makes sense.

wenyongh (commented):
LGTM

core/iwasm/include/wasm_export.h
```diff
@@ -252,6 +252,11 @@ typedef struct LoadArgs {
     const strings), making it possible to free the wasm binary buffer after
     loading. */
     bool wasm_binary_freeable;
+
+    /* false by default, if true, don't resolve the symbols yet. The
+       wasm_runtime_load_ex has to be followed by a wasm_runtime_resolve_symbols
+       call */
+    bool no_resolve;
```
wenyongh, line 259: Should also add the field in the struct LoadArgs in wasm_c_api.h, note that LoadArgs is defined in both wasm_export.h and wasm_c_api.h and is controlled by macro LOAD_ARGS_OPTION_DEFINED.

lum1n0us:
> We're using WAMR along with a scripting runtime and we need to generate dynamic imports. To do this we need to know the function signatures before instantiating the module. So far we've gotten around it by just loading the module first, then inspecting it. Then loading it again. This feels hacky and inefficient. With this functionality we are able to avoid the second load.

It's good to be informed. There will be some improvements and code refactoring in the area of linking soon. Please stay updated on the progress.

wenyongh:
@Andersbakken I found another issue and added a comment, could you help address it so that I can merge the PR?

core/iwasm/include/wasm_export.h
Andersbakken: Ah. I'd missed that. Thanks. Fixed now.

PR #3795, opened 2024-09-16
Add memory import support

bnason-nf:
This adds support to the public export API and internal implementation for import memory objects. The exception to this is that the AOT compiler and loader are currently hardcoded not to support import memories, so that will need to be fixed in a future change.

Please let me know if there's anything about the approach used here that should be improved or re-done.

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -819,7 +819,7 @@ memories_deinstantiate(AOTModuleInstance *module_inst)
     AOTMemoryInstance *memory_inst;
 
     for (i = 0; i < module_inst->memory_count; i++) {
-        memory_inst = module_inst->memories[i];
+        memory_inst = module_inst->memories[i].memory;
         if (memory_inst) {
 #if WASM_ENABLE_SHARED_MEMORY != 0
             if (shared_memory_is_shared(memory_inst)) {
@@ -872,7 +872,7 @@ memory_instantiate(AOTModuleInstance *module_inst, AOTModuleInstance *parent,
         AOTMemoryInstance *shared_memory_instance;
         bh_assert(memory_idx == 0);
         bh_assert(parent->memory_count > memory_idx);
-        shared_memory_instance = parent->memories[memory_idx];
+        shared_memory_instance = parent->memories[memory_idx].memory;
         shared_memory_inc_reference(shared_memory_instance);
         return shared_memory_instance;
     }
@@ -1071,7 +1071,7 @@ aot_lookup_memory(AOTModuleInstance *module_inst, char const *name)
     (void)module_inst->export_memories;
     if (!module_inst->memories)
         return NULL;
-    return module_inst->memories[0];
+    return module_inst->memories[0].memory;
 #endif
 }
 
@@ -1079,7 +1079,7 @@ AOTMemoryInstance *
 aot_get_default_memory(AOTModuleInstance *module_inst)
 {
     if (module_inst->memories)
-        return module_inst->memories[0];
+        return module_inst->memories[0].memory;
     else
         return NULL;
 }
@@ -1089,7 +1089,7 @@ aot_get_memory_with_index(AOTModuleInstance *module_inst, uint32 index)
 {
     if ((index >= module_inst->memory_count) || !module_inst->memories)
         return NULL;
-    return module_inst->memories[index];
+    return module_inst->memories[index].memory;
 }
 
 static bool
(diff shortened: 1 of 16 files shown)


lum1n0us:
The `WASM_ENABLE_MULTI_MODULE` setting is currently restricted to importing only globals and functions. We prefer to keep it this way unless there are explicit requirements for change.

We're in the process of developing instantiation linking (as opposed to `MULTI_MODULE`, which is loader linking). This new approach will encompass all four elements: global variables, functions, memory, and tables. Therefore, it seems more appropriate to incorporate "memory import support" into the new linking framework rather than considering it a separate feature.

In this series of PRs, I believe we can complete APIs to obtain import types and export items.

bnason-nf:
Hi @lum1n0us,

This change is not intended to have any special support for multiple modules, it's intended to add support for import memories in general. Would you mind clarifying what you saw in the change that might not fit with that?

Thanks,
Benbuck

bnason-nf:
To be clear, I fully agree with you adding support for importing globals, functions, memories, and tables, and linking them at instantiation time. This change is intended to be the first piece of that, with others to come later, please see our previous brief discussion here:

https://github.com/bytecodealliance/wasm-micro-runtime/pull/3786#discussion_r1759637244

lum1n0us:
Apologies for the confusion. Yes, this pull request is about how to import memory, which is also a part of the linking process. Multi-module is a type of linking technology, referred to as loading linking. This is different from the instantiation linking defined by the Spec., which WAMR has not yet supported. I have opened [a new issue to track the implementation of the linking as defined by the Spec](https://github.com/bytecodealliance/wasm-micro-runtime/issues/3807)

bnason-nf:
Hi @lum1n0us,

I believe that this PR does implement instantiation linking for memory objects. For example in `wasm_export.h`, the new `WASMImportInst` imports member was added to `InstantiationArgs`. Would you mind explaining please what you see that doesn't meet that criteria, or what I'm misunderstanding?

Thanks,
Benbuck

lum1n0us:
> I believe that this PR does implement instantiation linking for memory objects. For example in wasm_export.h, the new WASMImportInst imports member was added to InstantiationArgs. 

You are absolutely right. I am just hoping that we can design the entire instantiation linking process as much as we can, which includes not only memory linking but also other aspects, to ensure it fully aligns with the Spec. definition.

Some of this work has already been completed in your previous pull requests. Thank you, by the way.

bnason-nf:
Yes we're in full agreement on wanting to fully support the specification for importing all types. I tried to plan for that by building scaffolding in the `WASMImportInst` structure:

```
    wasm_import_export_kind_t kind;
    union {
        wasm_memory_inst_t memory_inst;
    } u;
```

Which should allow us to easily add other types. But if you don't like this approach in general, that's fine, I'm just trying to understand what design choices you think would be better.

lum1n0us:
I've created [an issue to gather ideas about instantiation linking](https://github.com/bytecodealliance/wasm-micro-runtime/issues/3807). Please leave your comments and thoughts there.

lum1n0us:
About `WASMImportInst`:

- Include other types of instances in the union, such as function, global, table, and tag.
- The name `WASMImportInst` might need to be changed. My initial thought is `WASMExternal`, similar to wasm_c_api.

wenyongh:
> About `WASMImportInst`:
> 
> ~- Include other types of instances in the union, such as function, global, table, and tag.~
> 
> * The name `WASMImportInst` might need to be changed. My initial thought is `WASMExternal`, similar to wasm_c_api.
> 

Using extern seems good but WASMExternal makes me a little confused with wasm-c-api's structures. Since now there are type wasm_memory_inst_t, wasm_function_inst_t, wasm_global_inst_t and wasm_table_inst_t, how about using **WASMExternInst** or **WASMExternInstance**? And then `typedef WASMExternInst *wasm_extern_inst_t`.

bnason-nf:
Hi  @wenyongh. Yeah, I also wasn't sure the memory wrapper approach would be the best choice, so I'm not surprised to hear you don't love it either. I could try to re-do this PR with a new approach if we want, but I'm not sure if that's worth doing?

wenyongh:
> Hi @wenyongh. Yeah, I also wasn't sure the memory wrapper approach would be the best choice, so I'm not surprised to hear you don't love it either. I could try to re-do this PR with a new approach if we want, but I'm not sure if that's worth doing?

It is up to you, but I think it helps, at least it is a good reference for the future instantiation linking implementation. And if you want to do it, could you modify WASMImportInst to WASMExternInst or WASMExternInstance?

lum1n0us:
Plus, IIUC, we don't need `WASMMemoryWrapper` to package host created `WASMMemoryInstance` and runtime created `WASMMemoryInstance`. Both should be same thing.

That's why I also thought we should not have another implementation in `wasm_runtime_memory_create()`. It should use `memory_instantiation()` in some way.

lum1n0us:
@bnason-nf  please take a review of #3845, #3887, #3893. Or just #3893. Let me know your comments.

wenyongh:
Close this PR since #3845 and #3887 have been merged.

PR #3798, opened 2024-09-17
Fix and suppress some ASAN problems.

Andersbakken:
This addresses some of the ASAN problems we have run into. Some of them we had to suppress because we couldn't find a way to fix them and likely not all code paths have been exercised.


matthargett:
this appears to be superceded by #4900 , and the number of code changes since the last rebase looks like they have moved around some of the places that were previously suppressed. can this be closed and specific fixes (if any are needed?) for latest sources be opened separately?

PR #3806, opened 2024-09-20
Add shared heap sample

wenyongh:


samples/shared-heap/src/main.c
```diff
@@ -0,0 +1,309 @@
...
+            break;
+        }
+
+        snprintf(buf, 1024, "Hello, this is buf %u allocated from shared heap",
+                 i + 1);
+
+        printf("wasm app1 send buf: %s\n\n", buf);
+        if (!bh_post_msg(queue, 1, buf, 1024 * i)) {
+            printf("Failed to post message to queue\n");
+            wasm_runtime_shared_heap_free(module_inst, argv[0]);
```
WenLY1, line 60: seems that argv[0] is not initialized

wenyongh: yes, should be offset. fixed.

WenLY1 (approved):

PR #3811, opened 2024-09-23
Support table64 extension in classic-interp and AOT running mode

TianlongLiang:
Add table64 extension(in Memory64 proposal) support in classic-interp and AOT running mode, currently still use uint32 to represent table's init and max size to keep AOT ABI unchanged.


core/iwasm/compilation/aot_emit_table.c
```diff
@@ -484,7 +581,14 @@ aot_compile_op_table_size(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
         goto fail;
     }
 
-    PUSH_I32(tbl_sz);
+#if WASM_ENABLE_MEMORY64 != 0
+    if (IS_TABLE64(tbl_idx)) {
+        if (!zero_extend_u64(comp_ctx, &tbl_sz, "length64")) {
+            return false;
```
wenyongh, line 587: had better `goto fail` like above

core/iwasm/compilation/aot_emit_table.c
```diff
@@ -545,7 +653,14 @@ aot_compile_op_table_grow(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
         goto fail;
     }
 
-    PUSH_I32(ret);
+#if WASM_ENABLE_MEMORY64 != 0
+    if (IS_TABLE64(tbl_idx)) {
+        if (!zero_extend_u64(comp_ctx, &ret, "table_size64")) {
+            return false;
```
wenyongh, line 659: had better `goto fail` like above

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -4616,13 +4633,16 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+                mem_offset_t delta;
 
                 read_leb_memidx(frame_ip, frame_ip_end, mem_idx);
                 prev_page_count = memory->cur_page_count;
-                delta = (uint32)POP_PAGE_COUNT();
+                delta = POP_PAGE_COUNT();
 
-                if (!wasm_enlarge_memory_with_idx(module, delta, mem_idx)) {
+                if (delta > UINT32_MAX
+                    || !wasm_enlarge_memory_with_idx(module, (uint32)delta,
```
wenyongh, line 4644: Is it better to change to:
```C
    if (
#if WASM_ENABLE_MEMORY64 != 0
        delta > UINT32_MAX ||
#endif
        !wasm_enlarge_memory_with_idx(module, ...)
```

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -5809,8 +5833,12 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                                     .value_count;
                         }
 
-                        if (offset_len_out_of_bounds(s, n, tbl_seg_len)
-                            || offset_len_out_of_bounds(d, n,
+                        /* TODO: memory64 current implementation of table64
+                         * still assume the max table size UINT32_MAX
+                         */
+                        if (d > UINT32_MAX
```
wenyongh, line 5839: Same as above

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -5877,14 +5905,25 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+#endif
+                        s = (tbl_elem_idx_t)POP_TBL_ELEM_IDX();
+#if WASM_ENABLE_MEMORY64 != 0
+                        is_table64 = dst_tbl_inst->is_table64;
+#endif
+                        d = (tbl_elem_idx_t)POP_TBL_ELEM_IDX();
 
-                        if (offset_len_out_of_bounds(d, n,
-                                                     dst_tbl_inst->cur_size)
+                        if (n > UINT32_MAX || s > UINT32_MAX || d > UINT32_MAX
```
wenyongh, line 5922: same as above

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -51,6 +51,20 @@ has_module_memory64(WASMModule *module)
...
+static bool
+is_table_64bit(WASMModule *module, uint32 table_idx)
+{
+    if (module->import_table_count > 0
+        && table_idx < module->import_table_count)
+        return (module->import_tables[table_idx].u.table.table_type.flags
+                & TABLE64_FLAG)
+               != 0;
+    else if (module->table_count > 0)
+        return (module->tables[table_idx].table_type.flags & TABLE64_FLAG) != 0;
```
wenyongh, line 64: Seems the table_idx has been checked before? Or we should set error buf before returning false. How about just:
```
    if (table_idx < module->import_table_count)
        return (module->import_tables[table_idx].u.table.table_type.flags
                & TABLE64_FLAG)
               != 0;
    else
        return (module->tables[table_idx].table_type.flags & TABLE64_FLAG) != 0;
```

wenyongh (approved):
LGTM

core/iwasm/common/wasm_loader_common.c
```diff
@@ -60,6 +91,36 @@ wasm_memory_check_flags(const uint8 mem_flag, char *error_buf,
...
+            wasm_loader_set_error_buf(error_buf, error_buf_size,
+                                      "invalid limits flags", is_aot);
+            return false;
+        }
+#endif
+    }
+
+    if (table_flag > MAX_TABLE_SIZE_FLAG + TABLE64_FLAG) {
+        wasm_loader_set_error_buf(error_buf, error_buf_size,
+                                  "invalid limits flags", is_aot);
```
lum1n0us, line 117: The error message is the same as the one on L109. It would be better if it were different.

core/iwasm/compilation/aot_compiler.c
```diff
@@ -146,9 +146,20 @@ aot_validate_wasm(AOTCompContext *comp_ctx)
...
+        if (IS_MEMORY64) {
+            aot_set_last_error(
+                "Compiling wasm64 to 32bit platform is not allowed");
+            return false;
+        }
+
+        for (uint32 i = 0; i < comp_ctx->comp_data->table_count; ++i) {
+            if (IS_TABLE64(i)) {
+                aot_set_last_error(
+                    "Compiling wasm64 to 32bit platform is not allowed");
```
lum1n0us, line 159: The error message is the same as the one on L152. It would be better if it were different.

core/iwasm/compilation/aot_compiler.c
```diff
@@ -146,9 +146,20 @@ aot_validate_wasm(AOTCompContext *comp_ctx)
     }
 
 #if WASM_ENABLE_MEMORY64 != 0
-    if (comp_ctx->pointer_size < sizeof(uint64) && IS_MEMORY64) {
-        aot_set_last_error("Compiling wasm64 to 32bit platform is not allowed");
-        return false;
+    if (comp_ctx->pointer_size < sizeof(uint64)) {
+        if (IS_MEMORY64) {
```
lum1n0us, line 150: I'm just wondering, why does L150 only check `memories[0]` while L157 needs to go through the entire table, considering there could be multiple memories?

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -2315,7 +2321,7 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
 
                 /**
                  * type check. compiler will make sure all like
-                 * (call_indirect (type $x) (i32.const 1))
+                 * (call_indirect (type $x) (it.const 1))
```
lum1n0us, line 2324: `i32.const` ?

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -2548,23 +2565,29 @@ load_table_import(const uint8 **p_buf, const uint8 *buf_end,
...
 #endif /* end of WASM_ENABLE_GC == 0 */
 
-    read_leb_uint32(p, p_end, declare_max_size_flag);
-    if (declare_max_size_flag > 1) {
-        set_error_buf(error_buf, error_buf_size, "integer too large");
+    p_org = p;
+    read_leb_uint32(p, p_end, table_flag);
+    is_table64 = table_flag & TABLE64_FLAG;
+    if (p - p_org > 1) {
+        LOG_VERBOSE("integer representation too long");
```
lum1n0us, line 2572: If the special case doesn't need the same error message as in L2572, we might change it to something more specific about the `table_flag` itself.

core/iwasm/compilation/aot_compiler.c
TianlongLiang: Currently, we assume that the memory is all 64bit or all 32bit, so it's okay only to check the first memory section

core/iwasm/interpreter/wasm_interp_classic.c
TianlongLiang: `it` is short for index type, depends on the table flag, it can be i32 or i64

lum1n0us (approved):
LGTM

Issue #3731, opened 2024-08-19
AOT module load failed: resolve symbol __atomic_compare_exchange_4 failed


Kvencc:
The following are the details when wamrc is compiled：
wamrc --target=riscv32 --target-abi=ilp32d --size-level=1 --enable-multi-thread --format=aot -o foo.aot foo.asm
Create AoT compiler with:
  target:        riscv32
  target cpu:
  target triple: riscv32-pc-linux-ilp32d
  cpu features:  +d
  opt level:     3
  size level:    1
  output format: AoT file
Compile success, file foo.aot was generated.

Then when I tried to load and run this aot module, this error occurred：

AOT module load failed: resolve symbol __atomic_compare_exchange_4 failed.

I tried to use the same compilation option on other platforms, but only modified the -target option, which seems to work normally.

Do I need to provide implementation for functions like __atomic_compare_exchange_4 and __atomic_store_4 in riscv?

Thank you very much for your reply.

Kvencc:
If I don't specify the target, won't wamrc be compiled into the host architecture by default? I have x86_64 here. But I need to run this aot module on the riscv platform.

When I run on the riscv platform if dont specify the target, an error like this will be returned：
AOT module load failed: invalid target bit width, expected 32-bit but got 64-bit

TianlongLiang:
So you want to do cross-compiling.

AOT file generated on the x86_64 platform
iwasm ran on the riscv32 platform(is it also compiled on the riscv32 platform or use cross-compile toolchain generated on the x86_64 platform?)

Am I right?

And the error message:
```bash
AOT module load failed: resolve symbol __atomic_compare_exchange_4 failed.
```

refer to the lack of support for atomic intrinsic on the riscv32 platform, I am not very familiar with the risc32 platform, I suspect that it might be some extra compile flag needed to be added when compiling iwasm on the risv32 platform to support atomic operations

Kvencc:
Yes, I ported iwasm to my application and used cross-compilation tools to build it on the x86_64 platform.
And the aot file is also generated on the x86 platform.

Can I choose not to use these platform-related features in iwasm? If possible, which options should I specify in the process of compiling iwasm and aot?

Kvencc:
I broke the debugging point in the resolve_target_sym function of aot_loader.c, and the following functions returned null:
- __atomic_compare_exchange_4

- __atomic_store_4

- __divsi3

- __mulsi3

- __udivsi3

[...]

PR #3812, opened 2024-09-24
Fix missing symbols when using aot mode on riscv platforms. (#3731)

Kvencc:
Fix missing symbols when using AOT mode on RISCV platforms.

If the target platform does not support the “Atomic Instructions” or “Integer Multiplication and Division” features, or does not turn on support for these CPU features when using wamrc to compile the aot file.

Just add these two macros when compiling iwasm:

- ARCH_RISCV_COMPILE_AOT_WITHOUT_FEATURE_A=1
- ARCH_RISCV_COMPILE_AOT_WITHOUT_FEATURE_M=1

### DIFF
--- a/core/iwasm/aot/arch/aot_reloc_riscv.c
+++ b/core/iwasm/aot/arch/aot_reloc_riscv.c
@@ -106,8 +106,9 @@ void __umoddi3(void);
 void __umodsi3(void);
 void __unorddf2(void);
 void __unordsf2(void);
-void __atomic_compare_exchange_4(void);
-void __atomic_store_4(void);
+bool __atomic_compare_exchange_4(volatile void *, void *, unsigned int,
+                                 bool, int, int);
+void __atomic_store_4(volatile void *, unsigned int, int);
 /* clang-format on */
 
 static SymbolMap target_sym_map[] = {
--- a/core/iwasm/aot/arch/aot_reloc_riscv.c
+++ b/core/iwasm/aot/arch/aot_reloc_riscv.c
@@ -49,21 +49,8 @@
 #define NEED_SOFT_I64_DIV
 #endif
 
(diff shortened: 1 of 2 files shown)


no1wudi:
Could you please describe the specific use case?

In my view, if the purpose is to detect whether atomic or multiplication/division instructions are supported, one can use the built-in macros in GCC to identify this without the need for additional configurations.

Kvencc:
Yes, you have a point. The current processing really isn't necessary.
In that case, it seems like just adding the missing atomic operation symbols and `__floatdisf`.

Is this a good way to handle it?

no1wudi:
> Is this a good way to handle it?

Yes, I think something like `__riscv_atomic` is enough to handle it

no1wudi (approved):

core/iwasm/aot/arch/aot_reloc_riscv.c
```diff
@@ -101,6 +106,8 @@ void __umoddi3(void);
 void __umodsi3(void);
 void __unorddf2(void);
 void __unordsf2(void);
+void __atomic_compare_exchange_4(void);
+void __atomic_store_4(void);
```
wenyongh, line 110: The build_iwasm_on_nuttx CI reported errors:
image
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/11052682397/job/30739828374

How about changing to below lines to have a try:
```C
bool __atomic_compare_exchange_4(volatile void *, void *, unsigned int,
                                 bool, int, int);
void __atomic_store_4(volatile void *, unsigned int, int);
```

Kvencc: Sure, I'll try it.

wenyongh (approved):

PR #3815, opened 2024-09-25
Implement shared heap for AOT

wenyongh:


core/iwasm/compilation/aot_emit_memory.c
```diff
@@ -270,6 +275,116 @@ aot_check_memory_overflow(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
...
+        if (!(maddr_phi = LLVMBuildPhi(comp_ctx->builder, INT8_PTR_TYPE,
+                                       "maddr_phi"))) {
+            aot_set_last_error("llvm build phi failed");
+            goto fail;
+        }
+
+        LLVMPositionBuilderAtEnd(comp_ctx->builder, block_curr);
+
+        if (!is_target_64bit) {
+            /* Check whether interger overflow occurs in addr + offset */
```
WenLY1, line 302: how about add the checked addr by aot_checked_addr_list_add as line 448. By doing this, the interger overflow does not need to check twice

samples/shared-heap/wasm-apps/test1.c
```diff
@@ -14,12 +15,41 @@ shared_free(void *ptr);
 void *
 my_shared_malloc(uint32_t size, uint32_t index)
 {
-    char *buf = shared_malloc(size);
+    char *buf1 = NULL, *buf2 = NULL, *buf;
```
WenLY1, line 18: redefinition of 'buf'

core/iwasm/compilation/aot_emit_memory.c
wenyongh: The integer overflow isn't check twice, when shared heap is enabled and the target is 32-bit, it runs into L425.

core/iwasm/aot/aot_runtime.h
```diff
@@ -111,6 +111,14 @@ typedef struct AOTFunctionInstance {
 
 typedef struct AOTModuleInstanceExtra {
     DefPointer(const uint32 *, stack_sizes);
+    /*
+     * Ajusted shared heap based addr to simple the calculation
```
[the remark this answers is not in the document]
wenyongh: done

samples/shared-heap/wasm-apps/test1.c
wenyongh: here it defines buf1, buf2 and buf, should not be an issue.

WenLY1: Ok, it's my mistake

WenLY1 (approved):

PR #3818, opened 2024-09-26
Implement the first few SIMD opcodes for fast interpreter (v128.const, v128.any_true)

loganek:
Tested on the following code:
```
(module
  (import "wasi_snapshot_preview1" "proc_exit" (func $proc_exit (param i32)))
  
  (memory (export "memory") 1)

  ;; WASI entry point
  (func $main (export "_start")
    v128.const i8x16 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
    v128.any_true
    if
      unreachable
    end
    
    v128.const i8x16 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15     
    v128.any_true
    i32.const 0
    i32.eq
    if
      unreachable
    end

    i32.const 0
    call $proc_exit
  )
)
```

### DIFF
--- a/core/iwasm/common/wasm_runtime_common.h
+++ b/core/iwasm/common/wasm_runtime_common.h
@@ -37,6 +37,10 @@ extern "C" {
     do {                                   \
         *(int64 *)(addr) = (int64)(value); \
     } while (0)
+#define PUT_V128_TO_ADDR(addr, value) \
+    do {                              \
+        *(V128 *)(addr) = (value);    \
+    } while (0)
 #define PUT_F64_TO_ADDR(addr, value)           \
     do {                                       \
         *(float64 *)(addr) = (float64)(value); \
@@ -49,6 +53,7 @@ extern "C" {
 #define GET_I64_FROM_ADDR(addr) (*(int64 *)(addr))
 #define GET_F64_FROM_ADDR(addr) (*(float64 *)(addr))
 #define GET_REF_FROM_ADDR(addr) (*(void **)(addr))
+#define GET_V128_FROM_ADDR(addr) (*(V128 *)(addr))
 
 /* For STORE opcodes */
 #define STORE_I64 PUT_I64_TO_ADDR
@@ -82,6 +87,15 @@ STORE_U8(void *addr, uint8_t value)
     } while (0)
 
 #else /* WASM_CPU_SUPPORTS_UNALIGNED_ADDR_ACCESS != 0 */
+
+#define PUT_V128_TO_ADDR(addr, value)        \
+    do {                                     \
+        uint32 *addr_u32 = (uint32 *)(addr); \
+        addr_u32[0] = (value).i32x4[0];      \
+        addr_u32[1] = (value).i32x4[1];      \
+        addr_u32[2] = (value).i32x4[2];      \
+        addr_u32[3] = (value).i32x4[3];      \
+    } while (0)
 
 #define PUT_I64_TO_ADDR(addr, value)         \
     do {                                     \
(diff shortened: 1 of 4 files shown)


core/iwasm/common/wasm_runtime_common.h
```diff
@@ -49,6 +49,7 @@ extern "C" {
 #define GET_I64_FROM_ADDR(addr) (*(int64 *)(addr))
 #define GET_F64_FROM_ADDR(addr) (*(float64 *)(addr))
 #define GET_REF_FROM_ADDR(addr) (*(void **)(addr))
+#define GET_V128_FROM_ADDR(addr) (*(V128 *)(addr))
```
wenyongh, line 52: Do you think we should define the macro below when WASM_CPU_SUPPORTS_UNALIGNED_ADDR_ACCESS macro is 0?

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -5642,7 +5644,38 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+                switch (opcode) {
+                    case SIMD_v128_const:
+                    {
+                        uint8 *orig_ip = frame_ip;
+
+                        frame_ip += sizeof(V128);
+                        addr_ret = GET_OFFSET();
 
+                        bh_memcpy_s(frame_lp + addr_ret, sizeof(V128), orig_ip,
+                                    sizeof(V128));
```
wenyongh, line 5661: Not sure whether using memcpy may impact the performance, maybe it is better if we define the macro like GET_V128_FROM_ADDR and PUT_V128_TO_ADDR in wasm_runtime_common.h, and it should be needed when more opcodes are implemented.

core/iwasm/common/wasm_runtime_common.h
loganek: right, updated.

core/iwasm/interpreter/wasm_interp_fast.c
loganek: Hmm, the other I*_CONST opcodes were implemented using bh_memcpy_s, but I don't mind doing it differently. Updated the code.

core/iwasm/common/wasm_runtime_common.h
```diff
@@ -124,6 +138,17 @@ STORE_U8(void *addr, uint8_t value)
...
+static inline V128
+GET_V128_FROM_ADDR(uint32 *addr)
+{
+    V128 ret;
+    ret.i32x4[0] = addr[0];
+    ret.i32x4[1] = addr[1];
+    ret.i32x4[2] = addr[2];
+    ret.i32x4[3] = addr[3];
+    return ret;
+}
```
wenyongh, line 150: Not very sure why the CIs can run successfully, but I think that V128 may be undeclared in many targets/platforms when WASM_CPU_SUPPORTS_UNALIGNED_ADDR_ACCESS is 0, and compilation error may be reported, so here how about wrapping this function with `#if WASM_ENABLE_SIMD != 0 .. #endif`?

wenyongh (commented):
LGTM with minor comment

core/iwasm/common/wasm_runtime_common.h
loganek: `V128` is unconditionally declared in the `wasm.h` file so I don't think there's a need for that extra ifdef.

wenyongh: Hmm, got it, I wrongly thought that it is the type `v128` which is defined in `wasm_runtime_common.c` to support `wasm_runtime_invoke_native`, and is defined only when SIMD is enabled.

PR #3834, opened 2024-10-02
add missing stdint.h header

sjamesr:
We use bazel to build wamrc internally at Google, it complains if we don't have this header included (missing types such as `uint32_t`)


wenyongh (approved):

lum1n0us (changes requested):
LGTM

lum1n0us (approved):
LGTM

PR #3835, opened 2024-10-02
Emit load_addr and load_size if WAMR_ENABLE_COMPILER is set

sjamesr:
Currently, the open-source builds set WASM_ENABLE_DUMP_CALL_STACK, which causes these two fields to be emitted. They are required by aot_emit_exception.cc.

Internally at Google, we don't enable call stack dumps, so we've been using the attached patch to make sure the fields are emitted anyway.

Does it make sense to build the compiler without ENABLE_DUMP_CALL_STACK? This could just be an oversight on our part when we import WAMR.

### DIFF
--- a/core/iwasm/interpreter/wasm.h
+++ b/core/iwasm/interpreter/wasm.h
@@ -976,8 +976,9 @@ struct WASMModule {
     uint64 buf_code_size;
 #endif
 
-#if WASM_ENABLE_DEBUG_INTERP != 0 || WASM_ENABLE_FAST_JIT != 0 \
-    || WASM_ENABLE_DUMP_CALL_STACK != 0 || WASM_ENABLE_JIT != 0
+#if WASM_ENABLE_DEBUG_INTERP != 0 || WASM_ENABLE_FAST_JIT != 0  \
+    || WASM_ENABLE_DUMP_CALL_STACK != 0 || WASM_ENABLE_JIT != 0 \
+    || WASM_ENABLE_WAMR_COMPILER != 0
     uint8 *load_addr;
     uint64 load_size;
 #endif


loganek:
@sjamesr could you tell a bit about your usecase? WAMR compiler usually is used as a developer tools and rarely in production, so size micro optimizations like this one is usually not a problem. You can still enable dump callstacks feature for the runtime itself though. So my recommended approach would be to:
* compile the aot **without** specifying `--enable-dump-call-stack`
* compile the runtime with `-DWAMR_BUILD_DUMP_CALL_STACK=0`  CMake flag (which is the default one IIRC).
Would that work for you?

sjamesr:
Currently we build the compiler as you suggest, without WASM_ENABLE_DUMP_CALL_STACK defined. The change I sent is required to get the compiler to build in this case. If you apply the following patch to disable it, I think you'll find wamrc doesn't build.

I was thinking about enabling WASM_ENABLE_DUMP_CALL_STACK for the compiler, because that's what is done upstream, but it's good to know that leaving it off should be OK.

```
diff --git a/wamr-compiler/CMakeLists.txt b/wamr-compiler/CMakeLists.txt
index 2ab0462b..210ff481 100644
--- a/wamr-compiler/CMakeLists.txt
+++ b/wamr-compiler/CMakeLists.txt
@@ -42,7 +42,7 @@ add_definitions(-DWASM_ENABLE_SIMD=1)
 add_definitions(-DWASM_ENABLE_REF_TYPES=1)
 add_definitions(-DWASM_ENABLE_CUSTOM_NAME_SECTION=1)
 add_definitions(-DWASM_ENABLE_AOT_STACK_FRAME=1)
-add_definitions(-DWASM_ENABLE_DUMP_CALL_STACK=1)
+add_definitions(-DWASM_ENABLE_DUMP_CALL_STACK=0)
 add_definitions(-DWASM_ENABLE_MODULE_INST_CONTEXT=1)
```

```
/home/sjr/wasm-micro-runtime/core/iwasm/compilation/aot_compiler.c: In function ‘aot_gen_commit_sp_ip’:
/home/sjr/wasm-micro-runtime/core/iwasm/compilation/aot_compiler.c:666:66: error: ‘WASMModule’ has no member named ‘load_addr’
  666 |                 value = I64_CONST((uint64)(uintptr_t)(ip - module->load_addr));
      |                                                                  ^~

```

loganek:
Got it. Yeah personally I don't see why the `WAMR_BUILD_DUMP_CALL_STACK` flag would be disabled for the wamrc itself, ideally this should be untouched, and only define/undefined it when building the runtime.

wenyongh (approved):

Issue #3840, opened 2024-10-05
wasm_application_execute_func does not return the return value


greenaddress:
Thanks for filing a bug or defect report! Please fill out the TODOs below.

### Subject of the issue

Describe the bug or defect here.

When calling wasm_application_execute_func the returned value is not passed back to the caller.

### Test case

Upload the related wasm file, wast file or the source files if you can.

int32_t run_me(void) {
    return 10;
}

You can't get 10 by calling  wasm_application_execute_func and looking in argv[0] for it.

### Your environment

* Linux
* WAMR version 2.1.2, platform, cpu architecture, running mode, etc.

### Steps to reproduce

Tell us how to reproduce this bug or defect.

### Expected behavior

Tell us what should happen
argv[0] should have the returned value of the function, at least for simple types (i32 for example)

### Actual behavior

Tell us what happens instead

### Extra Info

Anything else you'd like to add?

a patch that works at least for i32 is
```
diff --git i/core/iwasm/common/wasm_application.c w/core/iwasm/common/wasm_application.c
index 3b3be16c..42a21a99 100644
--- i/core/iwasm/common/wasm_application.c
+++ w/core/iwasm/common/wasm_application.c
@@ -685,6 +685,10 @@ execute_func(WASMModuleInstanceCommon *module_inst, const char *name,
         goto fail;
     }
 
+    if (argv && type->result_count == 1 && type->types[type->param_count] == VALUE_TYPE_I32)
+        /* copy the return value */
+        *(int *)argv = (int)argv1[0];
+
 #if WASM_ENABLE_GC != 0
     ref_type_map = type->result_ref_type_maps;
 #endif
```

PR #3841, opened 2024-10-05
add ability for wasm_application_execute_func to return return values

greenaddress:
This only works for i32 values.

If you think this should be extended to other kind of values please let me know

It should address https://github.com/bytecodealliance/wasm-micro-runtime/issues/3840 at least for me

### DIFF
--- a/core/iwasm/common/wasm_application.c
+++ b/core/iwasm/common/wasm_application.c
@@ -685,6 +685,10 @@ execute_func(WASMModuleInstanceCommon *module_inst, const char *name,
         goto fail;
     }
 
+    if (argv && type->result_count == 1 && type->types[type->param_count] == VALUE_TYPE_I32)
+        /* copy the return value */
+        *(int *)argv = (int)argv1[0];
+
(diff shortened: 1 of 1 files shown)


core/iwasm/common/wasm_application.c
```diff
@@ -685,6 +685,10 @@ execute_func(WASMModuleInstanceCommon *module_inst, const char *name,
         goto fail;
     }
 
+    if (argv && type->result_count == 1 && type->types[type->param_count] == VALUE_TYPE_I32)
+        /* copy the return value */
+        *(int *)argv = (int)argv1[0];
```
wenyongh, line 690: This only works for i32 type, if to support writing return values back to the array, we had better support all the types and support multiple return values, but it will lead to a lot of changes (e.g. main.c under product-mini) and API incompatibility, e.g., in 32-bit platform, there is only one argument and the size of argv is 4 bytes, but the return type is i64 and is 8 bytes, then developer had to pass larger argv array to this function. Could you use wasm_runtime_call_wasm instead?

PR #3848, opened 2024-10-09
Use alternate method to check getrandom support

bnason-nf:
I was still getting compile failures on an older Linux system (3.10), so hopefully this is a more direct way to check for getrandom support.


lum1n0us (approved):
LGTM

wenyongh (approved):

PR #3852, opened 2024-10-12
Enable libc-wasi for windows msvc build

wenyongh:
The default iwasm building in Windows MSVC enables libc-uvwasi because
libc-wasi isn't supported at the beginning. Since libc-wasi had been refactored
and is supported in Windows msys2 building, and libc-wasi supports more
functionalities(e.g. sockets) than libc-uvwasi, this PR fixes some issues to
enable libc-wasi in windows MSVC buidlings.


lum1n0us (approved):
LGTM

xujuntwt95329 (approved):
LGTM

PR #3859, opened 2024-10-15
Add first NEON SIMD opcode implementations to fast interp

jammar1:
Add some implementations of SIMD opcodes using NEON instructions.
Tested using:
```
(module
  (import "wasi_snapshot_preview1" "proc_exit" (func $proc_exit (param i32)))
  
  (memory (export "memory") 1)

  (func $assert_true (param v128)
    local.get 0
    v128.any_true 
    i32.eqz
    if
      unreachable
    end
  )

  (func $main (export "_start")

    i32.const 0
    i32.const 32
    memory.grow
    drop

    i32.const 0
    v128.const i8x16 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15
    v128.store

    i32.const 0
    v128.load

    v128.const i8x16 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15
    i8x16.eq
    call $assert_true

    i32.const 16
    v128.const i8x16 16 17 18 19 20 21 22 23 24 25 26 27 28 29 30 31
    v128.store

    i32.const 16
    v128.load
    v128.const i8x16 16 17 18 19 20 21 22 23 24 25 26 27 28 29 30 31
    i8x16.eq
    call $assert_true

    i32.const 0
    v128.load
    v128.const i8x16 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15
    i8x16.eq
    call $assert_true
    drop

    ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
    
    i32.const 0
    i32.const 1
    memory.grow
    drop

    i32.const 0
    i64.const 0x7F80FF017E02FE80
    i64.store

    i32.const 0
    v128.load8x8_s

    v128.const i16x8 127 -128 -1 1 126 2 -2 -128

    i16x8.eq
    call $assert_true

    ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;

    i32.const 0
    i64.const 0x80FE027E01FF807F
    i64.store

    i32.const 0
    v128.load8x8_u

    v128.const i16x8 128 254 2 126 1 255 128 127

    i16x8.eq
    call $assert_true

    ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;

    i32.const 0
    i64.const 0x8000FFFE7FFF0001
    i64.store

    i32.const 0
    v128.load16x4_s

    v128.const i32x4 -32768 -2 32767 1

    i32x4.eq
    call $assert_true

    ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;

    i32.const 0
    i64.const 0x8000FFFE7FFF0001 
    i64.store

    i32.const 0
    v128.load16x4_u

    v128.const i32x4 32768 65534 32767 1   

    i32x4.eq
    call $assert_true

    ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;

    i32.const 0
    i64.const 0x8000000000000001
    i64.store

    i32.const 0
    v128.load32x2_s

    v128.const i64x2 -2147483648 1 

    i64x2.eq
    call $assert_true

    ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;

    i32.const 0
    i64.const 0x8000000000000001
    i64.store

    i32.const 0
    v128.load32x2_u

    v128.const i64x2 2147483648 1

    i64x2.eq
    call $assert_true

    ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;

    call $proc_exit

  )
)
```


build-scripts/config_common.cmake
```diff
@@ -597,3 +597,6 @@ endif()
 if (NOT WAMR_BUILD_SANITIZER STREQUAL "")
   message ("     Sanitizer ${WAMR_BUILD_SANITIZER} enabled")
 endif ()
+if (ENABLE_NEON EQUAL 1)
+  add_definitions(-DENABLE_NEON=1)
+endif()
```
jammar1, line 602: Most of the devices we target are ARMv7 rather than ARMv8 (and even the v8 devices don't usually run in AArch64 mode. Most, but not all of the v7 devices have NEON support.

So I want to add a new macro for this, but not sure if this is the right thing to do?

I was thinking I could do something like a `WAMR_BUILD_ARM_NEON` style option and then `#ifdef WAMR_BUILD_ARM_NEON != 0 || BUILD_TARGET_AARCH64` or something like that?

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -5654,21 +5657,165 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+                            read_uint32(frame_ip); // TODO Check with an offset!
+                        addr = GET_OPERAND(uint32, I32, 0);
+                        frame_ip += 2;
+                        addr_ret = GET_OFFSET();
+                        CHECK_MEMORY_OVERFLOW(16);
+                        PUT_V128_TO_ADDR(frame_lp + addr_ret, LOAD_V128(maddr));
+                        break;
+                    }
                     case SIMD_v128_load8x8_s:
+#if ENABLE_NEON == 1
```
jammar1, line 5672: Should I do something similar to the invokeNative files where I define some function/symbol that is defined for each architecture in a separate file and put the opcode implementation there, or is it better inline here?

jammar1:
Als regarding testing, I think I will need to add some qemu tests for ARMv7 + NEON in this PR. Is that okay?

build-scripts/config_common.cmake
loganek: I think the new build flag makes sense if really NEON support is non-deterministic only based on the architecture. I'd however suggest a few changes here:
* use `WAMR_BUILD_*` prefix for consistency (e.g. `WAMR_BUILD_ARM_NEON`)
* consider enabling it by default for some common combinations
* for the `#ifdef WAMR_BUILD_ARM_NEON != 0 || BUILD_TARGET_AARCH64` condition, I'm not sure if the build target is needed; we should just use the NEON ifdef, otherwise it should not be defined if not desired.

build-scripts/config_common.cmake
```diff
@@ -597,3 +597,6 @@ endif()
 if (NOT WAMR_BUILD_SANITIZER STREQUAL "")
   message ("     Sanitizer ${WAMR_BUILD_SANITIZER} enabled")
 endif ()
+if (ENABLE_NEON EQUAL 1)
+  add_definitions(-DENABLE_NEON=1)
```
loganek, line 601:
```suggestion
  add_definitions(-DWASM_ENABLE_NEON=1)
```

loganek: also, you should add this to the `config.h` file and set it to 0 by default

core/iwasm/interpreter/wasm_interp_fast.c
loganek: my preference would be to not keep it here as it might grow a lot if we start supporting more architectures.

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -5654,21 +5657,165 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+                        frame_ip += 2;
+                        addr_ret = GET_OFFSET();
+                        CHECK_MEMORY_OVERFLOW(16);
+
+                        int8x8_t v8 = vld1_s8((const int8_t *)maddr);
+                        v8 = vrev64_s8(v8);
+                        int16x8_t v16 = vmovl_s8(v8);
+
+                        V128 result;
+                        vst1q_s16((int16_t *)&result, v16);
```
loganek, line 5686: I guess only this part is architecture specific so the other code can go outside of the `ifdef` (and possibly have an inefficient, plain C implementation in the `else` block?)

loganek: The same for all the other opcodes below.

core/iwasm/common/wasm_runtime_common.h
```diff
@@ -80,6 +85,7 @@ STORE_U8(void *addr, uint8_t value)
 #define LOAD_U32(addr) (*(uint32 *)(addr))
 #define LOAD_I16(addr) (*(int16 *)(addr))
 #define LOAD_U16(addr) (*(uint16 *)(addr))
+#define LOAD_V128(addr) (*(V128 *)(addr))
```
wenyongh, line 88: Had better also define STORE_V128 and LOAD_V128 when WASM_CPU_SUPPORTS_UNALIGNED_ADDR_ACCESS isn't 1.
see L95 to L386.

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -5654,21 +5657,165 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+                        uint32 offset, addr;
+                        offset = read_uint32(frame_ip);
+                        addr = GET_OPERAND(uint32, I32, 0);
+                        frame_ip += 2;
+                        addr_ret = GET_OFFSET();
+                        CHECK_MEMORY_OVERFLOW(16);
+
+                        int8x8_t v8 = vld1_s8((const int8_t *)maddr);
+                        v8 = vrev64_s8(v8);
+                        int16x8_t v16 = vmovl_s8(v8);
```
wenyongh, line 5683: Had better not use arch/os specified API here, could we wrap them to common APIs? And are there common implementations which doesn't depend on specified arch/os?

yamt: +1

using arch specific things directly like this PR should be the last resort after considering other options.

have you already considered libraries like simde? or compiler intrinsics? or even pure C implementation?
cf. https://github.com/yamt/toywasm/blob/e9226fae3e610d37742aa7afc8b0b7a7cf04f941/lib/insn_impl_simd.h#L8-L12

jammar1: Happy to use simde - I'll update this PR adding that and implementing a couple of opcodes.

jammar1: Done, added the library and implemented the swizzle opcode using it. Should be quite quick to add the rest of them now.

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -5647,13 +5651,34 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
 #if WASM_ENABLE_SIMD != 0
+/* TODO: Add x86 with additional #if */
+#define SIMD_V128_TO_SIMDE_V128(v) \
+    vreinterpretq_s32_u8(vld1q_u8((uint8_t *)&(v)))
+
+#define SIMDE_V128_TO_SIMD_V128(sv, v)              \
+    do {                                            \
+        uint8x16_t temp = vreinterpretq_u8_s32(sv); \
+        vst1q_u8((uint8_t *)&(v), temp);            \
+    } while (0)
```
jammar1, line 5662: I think this is fine for now - I think they're implemented in an agnostic way in the simde header but it's probably better to add other implementations in here too.

core/iwasm/common/wasm_runtime_common.h
```diff
@@ -264,7 +271,92 @@ STORE_U16(void *addr, uint16_t value)
...
+        } u;                                        \
+        if ((addr_ & (uintptr_t)15) == 0)           \
+            *(V128 *)(addr) = (V128)(value);        \
+        else {                                      \
+            u.val = (V128)(value);                  \
+            if ((addr_ & (uintptr_t)7) == 0) {      \
+                ((uint64 *)(addr))[0] = u.u64[0];   \
+                ((uint64 *)(addr))[1] = u.u64[1];   \
+            }                                       \
+            else if ((addr_ & (uintptr_t)3) == 0) { \
```
wenyongh, line 293: If we always only access the v128 operands in the interpreter's stack, since the stack operand must be stored in 4-byte aligned address (only i32/i64/f32/f64/v128 operands), here we can simplify the code to be like:
```C
    else {
        bh_assert((addr_ & (uintptr_t)3) == 0);
        ((uint32 *)(addr))[0] = u.u32[0];
        ((uint32 *)(addr))[1] = u.u32[1];
        ((uint32 *)(addr))[2] = u.u32[2];
        ((uint32 *)(addr))[3] = u.u32[3];
    }
    => the last two else statements are removed.
```
or the interpreter's code may be very large when the macros are expanded.

core/iwasm/common/wasm_runtime_common.h
```diff
@@ -264,7 +271,92 @@ STORE_U16(void *addr, uint16_t value)
...
+
+    if ((addr1 & (uintptr_t)7) == 0) {
+        u.u64[0] = ((uint64 *)addr)[0];
+        u.u64[1] = ((uint64 *)addr)[1];
+    }
+    else if ((addr1 & (uintptr_t)3) == 0) {
+        u.u32[0] = ((uint32 *)addr)[0];
+        u.u32[1] = ((uint32 *)addr)[1];
+        u.u32[2] = ((uint32 *)addr)[2];
+        u.u32[3] = ((uint32 *)addr)[3];
```
wenyongh, line 340: same as above, seems we can:
```C
    else {
        bh_assert((addr1 & (uintptr_t)3) == 0);
        u.u32[0] = ((uint32 *)addr)[0];
        u.u32[1] = ((uint32 *)addr)[1];
        u.u32[2] = ((uint32 *)addr)[2];
        u.u32[3] = ((uint32 *)addr)[3];
    }
```

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -5666,7 +5691,16 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                     case SIMD_v128_load64_splat:
                     case SIMD_v128_store:
                     {
-                        wasm_set_exception(module, "unsupported SIMD opcode");
+                        uint32 offset, addr;
+                        offset = read_uint32(frame_ip);
+                        addr = GET_OPERAND(uint32, I32, 2);
```
wenyongh, line 5696: I guess it should be `addr = GET_OPERAND(uint32, I32, 0); frame_ip += 2;`?

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -5666,7 +5691,16 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+
+                        V128 data;
+                        data = POP_V128();
+                        frame_ip += 2;
```
wenyongh, line 5700: Should remove this line?

build-scripts/config_common.cmake
```diff
@@ -321,6 +324,9 @@ if (WAMR_BUILD_SIMD EQUAL 1)
   else ()
     message ("     SIMD disabled due to not supported on target RISCV64")
   endif ()
+  if(WAMR_BUILD_TARGET MATCHES "AARCH64.*" OR "ARM.*")
+    add_definitions (-DWAMR_BUILD_SIMDE=1)
```
loganek, line 328: Shouldn't that be dependent on the `WAMR_BUILD_SIMDE` CMake flag? Also, to follow the pattern, consider renaming this definition to `WASM_ENABLE_SMIDE`

core/iwasm/common/wasm_runtime_common.h
```diff
@@ -264,7 +271,64 @@ STORE_U16(void *addr, uint16_t value)
     ((uint8_t *)(addr))[0] = u.u8[0];
     ((uint8_t *)(addr))[1] = u.u8[1];
 }
+
+#define STORE_V128(addr, value)                         \
```
loganek, line 275: why is this a macro, and not an inline function like the LOAD_V128?

product-mini/platforms/linux-sgx/CMakeLists.txt
```diff
@@ -68,6 +68,11 @@ if (NOT DEFINED WAMR_BUILD_LIB_RATS)
   set (WAMR_BUILD_LIB_RATS 0)
 endif()
 
+if (NOT DEFINED WAMR_BUILD_SIMDE)
```
loganek, line 71: why is that only for linux-sgx, and not the other platforms too?

core/iwasm/interpreter/wasm_opcode.h
```diff
@@ -783,7 +783,7 @@ typedef enum WASMAtomicEXTOpcode {
 #define SET_GOTO_TABLE_ELEM(opcode) [opcode] = HANDLE_OPCODE(opcode)
 
 #if (WASM_ENABLE_JIT != 0 || WASM_ENABLE_FAST_INTERP != 0) \
-    && WASM_ENABLE_SIMD != 0
+    && WASM_ENABLE_SIMD != 0 && WAMR_BUILD_SIMDE != 0
```
loganek, line 786: is that needed?

core/iwasm/libraries/simde/simde.cmake
```diff
@@ -0,0 +1,22 @@
+# simde is a header only library
```
loganek, line 1: please add the copyright note

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -5646,27 +5650,151 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
 #endif
                 goto call_func_from_entry;
             }
-#if WASM_ENABLE_SIMD != 0
+#if WASM_ENABLE_SIMD != 0 && WAMR_BUILD_SIMDE != 0
```
loganek, line 5653: again, not sure if it ever makes sense to enable SIMDE without enabling SIMD, so we could probably define those flags in a way that only SIMDE check here is required

core/iwasm/interpreter/wasm_opcode.h
```diff
@@ -783,7 +783,7 @@ typedef enum WASMAtomicEXTOpcode {
 #define SET_GOTO_TABLE_ELEM(opcode) [opcode] = HANDLE_OPCODE(opcode)
 
 #if (WASM_ENABLE_JIT != 0 || WASM_ENABLE_FAST_INTERP != 0) \
-    && WASM_ENABLE_SIMD != 0
+    && WASM_ENABLE_SIMDE != 0
```
loganek, line 786: Why this change? Should this be for SIMD in general, even if SIMDE is disabled?

loganek (approved):
LGTM

core/iwasm/libraries/simde/simde.cmake
```diff
@@ -0,0 +1,24 @@
...
+    GIT_REPOSITORY  https://github.com/simd-everywhere/simde
+    GIT_TAG v0.8.2
+)
+
+FetchContent_GetProperties(simde)
+if (NOT simde_POPULATED)
+    message("-- Fetching simde ..")
+    FetchContent_MakeAvailable(simde)
+    include_directories("${simde_SOURCE_DIR}")
+endif()
```
lum1n0us, line 24:
```cmake
FetchContent_Declare(...)
FetchContent_MakeAvailable(simde)
```

is enough. `FetchContent_MakeAvailable()` will check POPULATED and others.

core/iwasm/common/wasm_runtime_common.h
```diff
@@ -264,7 +271,67 @@ STORE_U16(void *addr, uint16_t value)
     ((uint8_t *)(addr))[0] = u.u8[0];
     ((uint8_t *)(addr))[1] = u.u8[1];
 }
+
```
lum1n0us, line 274: IIUC, the key assumption for both, `STORE_V128` and `LOAD_V128`, functions to work is that WAMR runs on aligned addresses. If I remember right, WAMR operates in unaligned mode by default, on x86 and x86-64 at least(?). 

Please correct me if it is no longer the case. If yes, I would suggest adding a `WASM_ENABLE_xxx` or just using `WASM_ENABLE_SIMDE` to differentiate it from the others.

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -5646,27 +5650,151 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+    } while (0)
+
             HANDLE_OP(WASM_OP_SIMD_PREFIX)
             {
                 GET_OPCODE();
 
                 switch (opcode) {
                     /* Memory */
                     case SIMD_v128_load:
+                    {
```
lum1n0us, line 5675: In my view, wouldn't it be better to move all the new ~1000 lines to a new, separate file? The original core/iwasm/interpreter/wasm_interp_fast.c is already large. let's not make it any bigger.

jammar1: I agree - I'm happy to follow up with a PR that moves things around later, but I have a few other branches that I'd need to rebase etc so would be good to merge this? This is just a feature branch so I think it's okay if it's a bit messy for now?

jammar1:
@wenyongh I think I've addressed your comments, would you be able to review this? Thanks

core/iwasm/libraries/simde/simde.cmake
```diff
@@ -0,0 +1,27 @@
+# Copyright (C) 2024 Amazon Inc.  All rights reserved.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+# simde is a header only library
+
+set (LIB_SIMDE_DIR ${CMAKE_CURRENT_LIST_DIR})
+
+if (WAMR_BUILD_TARGET MATCHES "AARCH64.*" OR "ARM.*")
+  add_definitions (-DWASM_ENABLE_SIMDE=1)
+endif()
+else ()
```
wenyongh, line 10: Invalid `else()`, should remove L9.

core/iwasm/libraries/simde/simde.cmake
```diff
@@ -0,0 +1,27 @@
...
+  add_definitions (-DWASM_ENABLE_SIMDE=0)
```
wenyongh, line 11: Suggest to add `#ifndef WASM_ENABLE_SIMDE #define WASM_ENABLE_SIMDE 0 #endif` in core/config.h, and remove this line.

core/iwasm/interpreter/wasm_opcode.h
```diff
@@ -782,7 +782,9 @@ typedef enum WASMAtomicEXTOpcode {
 
 #define SET_GOTO_TABLE_ELEM(opcode) [opcode] = HANDLE_OPCODE(opcode)
 
-#if (WASM_ENABLE_JIT != 0 || WASM_ENABLE_FAST_INTERP != 0) \
+#if (WASM_ENABLE_JIT != 0                                           \
+     || (WASM_ENABLE_FAST_INTERP != 0 && defined(WASM_ENABLE_SIMDE) \
+         && WASM_ENABLE_SIMDE != 0))                                \
```
wenyongh, line 787: Suggest to add `#ifndef WASM_ENABLE_SIMDE #define WASM_ENABLE_SIMDE 0 #endif` in core/config.h, so as to make WASM_ENABLE_SIMDE always defined, and then change this line to `|| (WASM_ENABLE_FAST_INTERP != 0 && WASM_ENABLE_SIMDE != 0)`

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -5646,27 +5650,151 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
 #endif
                 goto call_func_from_entry;
             }
-#if WASM_ENABLE_SIMD != 0
+#if defined(WASM_ENABLE_SIMDE) && WASM_ENABLE_SIMDE != 0
```
wenyongh, line 5653: Suggest to add #ifndef WASM_ENABLE_SIMDE #define WASM_ENABLE_SIMDE 0 #endif in core/config.h, so as to make WASM_ENABLE_SIMDE always defined, and then change this line to `#if WASM_ENABLE_SIMDE != 0`

core/iwasm/interpreter/wasm_interp_fast.c
wenyongh: @jammar1 I added several comments, could you have a look? I am fine to merge this PR to the branch `dev/simd_for_interp` first but we should address the issues then. Do you want to merge this PR first, or merge it after you address the new comments?

jammar1: @wenyongh I've addressed the comments and CI/ the checks have passed. Could you merge it if you're happy with the changes? Thanks

lum1n0us (approved):
LGTM

wenyongh (commented):
LGTM

wenyongh:
> @wenyongh I think I've addressed your comments, would you be able to review this? Thanks

Sure, merged. A suggestion is to change `WASM_ENABLE_SIMDE` to `WASM_ENABLE_LIB_SIMDE`, one is the former makes confusing with `WASM_ENABLE_SIMD`, the other is that we use cmake variable `WAMR_BUILD_LIB_SIMDE`, it is better to make it consistent. How do you think?

PR #3860, opened 2024-10-15
Fix quadratic runtime for duplicate export name detection

sjamesr:
Previously, the loader would check the name of a new export against all existing exports, leading to a quadratic running time.

This change makes the loader parse the entire export section. The exports are then sorted by name, then adjacent exports are checked for uniqueness.

As a side effect, the exports will appear in the parsed module in alphabetical order, not load order. This could be useful for improving the performance of the various wasm_runtime_lookup functions (binary instead of linear search).


sjamesr:
Breaks tests, will work on it

Issue #3407, opened 2024-05-09
"duplicate export name" check is O(n^2)


yamt:
"duplicate export name" check in load_export_section is too slow for modules with many exports.

wenyongh:
Maybe we can create the export list first, and then sort the list by name, and then check whether the contiguous two exports have the same name, and report error if yes?

yamt:
> Maybe we can create the export list first, and then sort the list by name, and then check whether the contiguous two exports have the same name, and report error if yes?

maybe.
for the other runtime, i do something along the line. https://github.com/yamt/toywasm/blob/115e895007bb77b12f6a632a3f4354644b236c56/lib/module.c#L2122-L2143

wenyongh:
OK, got it, thanks.

wenyongh:
@yamt I tried to refine the code to quick-search (bsearch) the wasm/aot export with name, by sorting the export list by name after loading the list. But it runs failed when running the wasm-c-api sample, I found that it is caused by:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/ea582fbc07468b2ef11a13d5ac18058047e50ef7/samples/wasm-c-api/src/callback.c#L157

This file was ported from:
https://github.com/WebAssembly/wasm-c-api/blob/main/example/callback.c#L133

In which it assumes that the export order should not be changed and `exports.data[0]` stores the related data of wasm file's export 0. So it seems that sorting export list with name is not a good way. How about creating an array of `<name, export_index>` pair and sort it by name instead?

yamt:
> @yamt I tried to refine the code to quick-search (bsearch) the wasm/aot export with name, by sorting the export list by name after loading the list. But it runs failed when running the wasm-c-api sample, I found that it is caused by:
> 
> https://github.com/bytecodealliance/wasm-micro-runtime/blob/ea582fbc07468b2ef11a13d5ac18058047e50ef7/samples/wasm-c-api/src/callback.c#L157

i tend to think it's a bug in the sample. but i agree probably it's safer to preserve the order.

> This file was ported from: https://github.com/WebAssembly/wasm-c-api/blob/main/example/callback.c#L133
> 
> In which it assumes that the export order should not be changed and `exports.data[0]` stores the related data of wasm file's export 0. So it seems that sorting export list with name is not a good way. How about creating an array of `<name, export_index>` pair and sort it by name instead?

it makes sense.

PR #3861, opened 2024-10-16
Fix quadratic runtime for duplicate export name detection

sjamesr:
Previously, the loader would check the name of a new export against all existing exports, leading to a quadratic running time.

This change makes the loader parse the entire export section. The exports are then sorted by name, then adjacent exports are checked for uniqueness. Export order is preserved.

### DIFF
--- a/core/iwasm/aot/aot_loader.c
+++ b/core/iwasm/aot/aot_loader.c
@@ -4,9 +4,8 @@
  */
 
 #include "aot_runtime.h"
-#include "bh_common.h"
-#include "bh_log.h"
 #include "aot_reloc.h"
+#include "bh_platform.h"
 #include "../common/wasm_runtime_common.h"
 #include "../common/wasm_native.h"
 #include "../common/wasm_loader_common.h"
@@ -2753,6 +2752,48 @@ destroy_exports(AOTExport *exports)
     wasm_runtime_free(exports);
 }
 
+static int
+cmp_export_name(const void *a, const void *b)
+{
+    return strcmp(*(char **)a, *(char **)b);
+}
+
+static bool
+check_duplicate_exports(AOTModule *module, char *error_buf,
+                        uint32 error_buf_size)
+{
+    uint32 i;
+    bool result = false;
+    char *names_buf[32], **names = names_buf;
+    if (module->export_count > 32) {
+        names = loader_malloc(module->export_count * sizeof(char *), error_buf,
+                              error_buf_size);
+        if (!names) {
+            return result;
+        }
+    }
+
+    for (i = 0; i < module->export_count; i++) {
+        names[i] = module->exports[i].name;
+    }
+
+    qsort(names, module->export_count, sizeof(char *), cmp_export_name);
+
+    for (i = 1; i < module->export_count; i++) {
+        if (!strcmp(names[i], names[i - 1])) {
+            set_error_buf(error_buf, error_buf_size, "duplicate export name");
+            goto cleanup;
+        }
+    }
+
+    result = true;
+cleanup:
+    if (module->export_count > 32) {
+        wasm_runtime_free(names);
+    }
+    return result;
+}
+
 static bool
 load_exports(const uint8 **p_buf, const uint8 *buf_end, AOTModule *module,
              bool is_load_from_file_buf, char *error_buf, uint32 error_buf_size)
@@ -2760,7 +2801,7 @@ load_exports(const uint8 **p_buf, const uint8 *buf_end, AOTModule *module,
     const uint8 *buf = *p_buf;
     AOTExport *exports;
     uint64 size;
-    uint32 i, j;
+    uint32 i;
 
     /* Allocate memory */
     size = sizeof(AOTExport) * (uint64)module->export_count;
@@ -2774,14 +2815,6 @@ load_exports(const uint8 **p_buf, const uint8 *buf_end, AOTModule *module,
         read_uint32(buf, buf_end, exports[i].index);
         read_uint8(buf, buf_end, exports[i].kind);
         read_string(buf, buf_end, exports[i].name);
-
-        for (j = 0; j < i; j++) {
-            if (!strcmp(exports[i].name, exports[j].name)) {
-                set_error_buf(error_buf, error_buf_size,
-                              "duplicate export name");
-                return false;
-            }
-        }
 
         /* Check export kind and index */
         switch (exports[i].kind) {
@@ -2827,6 +2860,12 @@ load_exports(const uint8 **p_buf, const uint8 *buf_end, AOTModule *module,
             default:
                 set_error_buf(error_buf, error_buf_size, "invalid export kind");
                 return false;
+        }
+    }
+
+    if (module->export_count > 0) {
+        if (!check_duplicate_exports(module, error_buf, error_buf_size)) {
+            return false;
         }
     }
 
--- a/core/iwasm/interpreter/wasm_loader.c
+++ b/core/iwasm/interpreter/wasm_loader.c
@@ -4,8 +4,7 @@
  */
 
 #include "wasm_loader.h"
-#include "bh_common.h"
-#include "bh_log.h"
+#include "bh_platform.h"
 #include "wasm.h"
 #include "wasm_opcode.h"
 #include "wasm_runtime.h"
@@ -3184,6 +3183,12 @@ fail:
     return false;
 }
 
+static int
+cmp_export_name(const void *a, const void *b)
+{
+    return strcmp(*(char **)a, *(char **)b);
+}
+
 static bool
 load_import_section(const uint8 *buf, const uint8 *buf_end, WASMModule *module,
                     bool is_load_from_file_buf, bool no_resolve,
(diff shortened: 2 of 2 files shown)


core/iwasm/interpreter/wasm_loader.c
```diff
@@ -4054,17 +4062,47 @@ load_global_section(const uint8 *buf, const uint8 *buf_end, WASMModule *module,
...
+    }
+
+    qsort(names, module->export_count, sizeof(char *), cmp_export_name);
+
+    for (uint32 i = 1; i < module->export_count; i++) {
+        if (!strcmp(names[i], names[i - 1])) {
+            set_error_buf(error_buf, error_buf_size, "duplicate export name");
+            goto cleanup;
+        }
+    }
```
wenyongh, line 4088: This change will allocate memory and qsort the array each time, not sure whether it improves the performance a lot? We ever discussed it before, and I think a better way is to create the name to index map array (element index in the exports array) in the wasm module, and use the binary search algorithm to lookup export with the name: get the element index in the exports array, use it to get the export and then use the `export->index` to get the function instance in wasm instance.

And here, seems we can establish the map array with binary search algorithm also: get the right place the insert the new export, and then insert it.

ps. https://github.com/bytecodealliance/wasm-micro-runtime/issues/3407.

sjamesr:
In a module with 100,000 exports, the load time goes from 20+ seconds to a couple hundred milliseconds on my machine. In the case of 10,000 exports, it's a couple hundred milliseconds down to tens of milliseconds.

This change will qsort the array once at the end of export section loading.

I was initially hoping that sorting the exports themselves by name would allow a binary search implementation in all the `wasm_runtime_lookup_...` functions, but as you observed in #3407, this causes other problems. Your solution would require keeping 2*ExportCount pointers in memory while the module is loaded, is this OK?

wenyongh:
> In a module with 100,000 exports, the load time goes from 20+ seconds to a couple hundred milliseconds on my machine. In the case of 10,000 exports, it's a couple hundred milliseconds down to tens of milliseconds.
> 
> This change will qsort the array once at the end of export section loading.
> 

OK, got it, it really improves a lot. It only allocates memory one time and qsort the array one time.

> I was initially hoping that sorting the exports themselves by name would allow a binary search implementation in all the `wasm_runtime_lookup_...` functions, but as you observed in #3407, this causes other problems. Your solution would require keeping 2*ExportCount pointers in memory while the module is loaded, is this OK?

Keeping `2 * export_count` pointers really consumes a lot of memory. IIUC, maybe we can just use a uint32 array (e.g. `uint32 *export_iths_of_sorted_name`) to save each export's ith of a temporary sorted export array, which is sorted by name. And then use binary-search to lookup to export, and get the export name with `exports[export_iths_of_sorted_name[i]` in the lookup.

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -3,6 +3,8 @@
  * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
  */
 
+#include <stdlib.h>
```
wenyongh, line 6: had better remove this line, should be unnecessary

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -4054,17 +4062,47 @@ load_global_section(const uint8 *buf, const uint8 *buf_end, WASMModule *module,
...
+{
+    char **names = loader_malloc(module->export_count * sizeof(char *),
+                                 error_buf, error_buf_size);
+    bool result = false;
+
+    if (!names) {
+        return result;
+    }
+
+    for (uint32 i = 0; i < module->export_count; i++) {
```
wenyongh, line 4077: had better declare `uint32 i` at the beginning of this function, we usually do that way since here it is C source code.

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -4054,17 +4062,47 @@ load_global_section(const uint8 *buf, const uint8 *buf_end, WASMModule *module,
...
+        names[i] = module->exports[i].name;
+    }
+
+    qsort(names, module->export_count, sizeof(char *), cmp_export_name);
+
+    for (uint32 i = 1; i < module->export_count; i++) {
```
wenyongh, line 4083: same as above

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -4173,6 +4202,10 @@ load_export_section(const uint8 *buf, const uint8 *buf_end, WASMModule *module,
         }
     }
 
+    if (!check_duplicate_exports(module, error_buf, error_buf_size)) {
+        return false;
+    }
```
wenyongh, line 4207: had better put the three lines before L4203, we only need to check it when export count is larger than 0.

wenyongh: And could you apply this optimization for AOT? See
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/aot/aot_loader.c#L2778-L2784

wenyongh:
@sjamesr I got an idea to improve the performance of wasm_runtime_lookup_function and submitted PR #3865, could you help check it? Thanks.

core/iwasm/interpreter/wasm_loader.c
sjamesr: curious, variable declarations can occur other than at the top of blocks since C99. Other parts of the code make use of C99 features, e.g. a designated initializer in `core/iwasm/aot/aot_runtime.c`. What is the reason that you stick to this style?

sjamesr:
#3865 was more like my first approach of sorting the exports in place, but I couldn't figure out why the CI was failing. You've fixed that, so I think that approach is obviously better

core/iwasm/interpreter/wasm_loader.c
wenyongh: Yes, declaring variables in the middle of a function is good, I just think that declaring a variable in a for loop is not a good coding style in C, it is declared not so explicitly and may make other variables be shadowed. And a little afraid that some compilers may report warning or error.

wenyongh:
> #3865 was more like my first approach of sorting the exports in place, but I couldn't figure out why the CI was failing. You've fixed that, so I think that approach is obviously better

But #3865 only optimizes wasm_runtime_lookup_function, it doesn't refine the wasm/aot loading process. I am not sure why you close this PR? The idea looks good.

sjamesr:
I misunderstood your other change, I thought you had a better way to fix performance in the loader as well, I'll reopen this one and address your comments.

core/iwasm/interpreter/wasm_loader.c
sjamesr: Our internal "include-what-you-use" tools will complain if it's not here (required by qsort.h), it should be a nop if it's already included elsewhere, right?

core/iwasm/interpreter/wasm_loader.c
sjamesr: Fixed

core/iwasm/interpreter/wasm_loader.c
sjamesr: Fixed

core/iwasm/interpreter/wasm_loader.c
sjamesr: Fixed, also made the change for AOT

sjamesr:
@wenyongh This should be good for another look, thank you for your comments

core/iwasm/interpreter/wasm_loader.c
wenyongh: Got it, we normally include all the system header files in platform layer, e.g. platform_internal.h, and don't include them in vmcore files. I think here we can change `#include "bh_common.h", #includde "bh_log.h"` below to `#include 'bh_platform.h"`. bh_platform will include all the platform header files. Could you help change it?

So does aot_loader.c.

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -4054,17 +4062,48 @@ load_global_section(const uint8 *buf, const uint8 *buf_end, WASMModule *module,
     return false;
 }
 
+static bool
+check_duplicate_exports(WASMModule *module, char *error_buf,
+                        uint32 error_buf_size)
+{
+    uint32 i;
+    char **names = loader_malloc(module->export_count * sizeof(char *),
+                                 error_buf, error_buf_size);
```
wenyongh, line 4071: If to reduce the chance of calling loader_malloc which may spend more time and be easier to get failed, we can define a bufer and only allocate memroy when the buffer isn't large enough.
```C
    char *names_buf[32], **names = names_buf;

    if (module->export_count > 32) {
        names = loader_malloc(...);
        ...
    }
    ...
```

sjamesr: Done

sjamesr: Took care of both comments, thank you!

core/iwasm/aot/aot_loader.c
```diff
@@ -2753,14 +2752,57 @@ destroy_exports(AOTExport *exports)
     wasm_runtime_free(exports);
 }
 
+static int
+cmp_export_name(const void *a, const void *b)
+{
+    return strcmp(((WASMExport *)a)->name, ((WASMExport *)b)->name);
```
wenyongh, line 2758: Should be `*(char **)a)`, `*(char **)b`?

core/iwasm/aot/aot_loader.c
```diff
@@ -2753,14 +2752,57 @@ destroy_exports(AOTExport *exports)
...
+{
+    uint32 i;
+    bool result = false;
+    char *names_buf[32], **names = names_buf;
+    if (module->export_count > 32) {
+        names = loader_malloc(module->export_count * sizeof(char *), error_buf,
+                              error_buf_size);
+    }
+
+    if (!names) {
```
wenyongh, line 2773: had better put check just after `names = loader_malloc`

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -3184,6 +3183,12 @@ load_memory(const uint8 **p_buf, const uint8 *buf_end, WASMMemory *memory,
     return false;
 }
 
+static int
+cmp_export_name(const void *a, const void *b)
+{
+    return strcmp(((WASMExport *)a)->name, ((WASMExport *)b)->name);
```
wenyongh, line 3189: had better `*(char **)a`

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -4054,17 +4059,54 @@ load_global_section(const uint8 *buf, const uint8 *buf_end, WASMModule *module,
...
+    uint32 i;
+    bool result = false;
+    char *names_buf[32], **names = names_buf;
+
+    if (module->export_count > 32) {
+        names = loader_malloc(module->export_count * sizeof(char *), error_buf,
+                              error_buf_size);
+    }
+
+    if (!names) {
```
wenyongh, line 4075: same as above

wenyongh (approved):
LGTM with minor comments

core/iwasm/interpreter/wasm_loader.c
sjamesr: Ooops! Fixed, thank you

lum1n0us (approved):
LGTM

Issue #3816, opened 2024-09-26
API for custom native stack registration


mkokryashkin:
#### Feature
Currently my application runs WASM modules withing custom C fibers. These fibers have their own stacks, different from their thread stack. WAMR, on the other hand, heavilly relies on pthread API for native stack boundary checking.
It would be nice to have a handle like
```
wasm_runtime_register_native_stack_boundary(uint8_t *stack_boundary)
```

So it is possible to register your own native stack boundary and use WAMR within a fiber.

#### Implementation
We just need to set `exec_env->native_stack_boundary` to a custom value.

wenyongh:
Hi, it is interested to support fibers. Currently the `exec_env->native_stack_boundary` is updated in each time `wasm_call_function/aot_call_function`, which calls `wasm_exec_env_set_thread_info` to get native stack boundary from os and update it. Per my understanding, you want to use the registered native stack boundary in `wasm_call_function/aot_call_function`? Seems it is better to store the registered native stack boundary in a new field of exec_env and use it if it is not NULL in `wasm_exec_env_set_thread_info`?

mkokryashkin:
> Seems it is better to store the registered native stack boundary in a new field of exec_env and use it if it is not NULL in `wasm_exec_env_set_thread_info`?

Yes, that is a great option and it adds a lot of flexibility.

wenyongh:
I uploaded PR #3862, could you check whether it resolves your issue? You can call wasm_runtime_set_native_stack_boundary to set the native stack boundary to exec_env.

mkokryashkin:
It works nicely, thank you very much!

wenyongh:
Close this issue since https://github.com/bytecodealliance/wasm-micro-runtime/pull/3862 was merged.

PR #3862, opened 2024-10-16
Allow to set native stack boundary to exec_env

wenyongh:
p.s. https://github.com/bytecodealliance/wasm-micro-runtime/issues/3816

### DIFF
--- a/core/iwasm/common/wasm_exec_env.c
+++ b/core/iwasm/common/wasm_exec_env.c
@@ -282,8 +282,11 @@ wasm_exec_env_set_thread_info(WASMExecEnv *exec_env)
     os_mutex_lock(&exec_env->wait_lock);
 #endif
     exec_env->handle = os_self_thread();
-    exec_env->native_stack_boundary =
-        stack_boundary ? stack_boundary + WASM_STACK_GUARD_SIZE : NULL;
+    if (exec_env->user_native_stack_boundary)
+        exec_env->native_stack_boundary = exec_env->user_native_stack_boundary;
+    else
+        exec_env->native_stack_boundary =
+            stack_boundary ? stack_boundary + WASM_STACK_GUARD_SIZE : NULL;
     exec_env->native_stack_top_min = (void *)UINTPTR_MAX;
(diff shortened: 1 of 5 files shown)


core/iwasm/common/wasm_exec_env.c
```diff
@@ -282,8 +282,11 @@ wasm_exec_env_set_thread_info(WASMExecEnv *exec_env)
     os_mutex_lock(&exec_env->wait_lock);
 #endif
     exec_env->handle = os_self_thread();
-    exec_env->native_stack_boundary =
-        stack_boundary ? stack_boundary + WASM_STACK_GUARD_SIZE : NULL;
+    if (exec_env->user_native_stack_boundary)
+        exec_env->native_stack_boundary = exec_env->user_native_stack_boundary;
```
yamt, line 286: i suspect it's simpler to add WASM_STACK_GUARD_SIZE here.
or, at least document it prominently on the api doc of wasm_runtime_set_native_stack_boundary to ensure the user aware of it.

wenyongh: Yes, seems it is simpler to add it. I just thought the default value of WASM_STACK_GUARD_SIZE depends on platforms, and it may be more flexible for developer to decide what the boundary should be. Both are good to me, and agree to add comment about WASM_STACK_GUARD_SIZE in the API, event when WASM_STACK_GUARD_SIZE is added here.

lum1n0us (approved):
LGTM.

core/iwasm/common/wasm_exec_env.c
xujuntwt95329: > I still think that it should be better for developer to choose what to do: if the behavior controlled by runtime, then all the two options are controlled by runtime, otherwise all the two options had better be controlled by developer, or it may be confusing for him. I updated the PR to add some comments, please help review again. Thanks.

Agree, since it's an API to give the control to the user, then we should leave everything to the user.

xujuntwt95329 (approved):
LGTM

PR #3871, opened 2024-10-22
Exclude fuzz test python and npm packages in scoreboard scan

TianlongLiang:
Ignore some vulnerabilities reported in the scoreboard CI since they are only used in fuzz test. They won't be included in the actual product.


lum1n0us (approved):
LGTM.

wenyongh (approved):

PR #3879, opened 2024-10-24
Enable assertions in test_wamr.sh script

lum1n0us:


lum1n0us:
There already has been a `-g build iwasm with debug version`

Issue #3837, opened 2024-10-04
current status of AOT_CURRENT_VERSION


Zzzabiyaka:
Hi!

I see that we mention that upon breaking update we bump AOT_CURRENT_VERSION

Looking at git blame I am not sure we did that when we introduced version 2.x.y (which I suppose breaks the ABI for AOT) 

Is there still an intent to keep this flag reflecting the "AOT ABI" version or presumably we always will be bumping major version of WAMR instead when we break AOT ABI in the future ?

wenyongh:
@Zzzabiyaka please refer to [doc/semantic_version.md](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/doc/semantic_version.md) and [WAMR-2.0.0 release notes](https://github.com/bytecodealliance/wasm-micro-runtime/releases/tag/WAMR-2.0.0), the last time we upgraded AOT_CURRENT_VERSION was due to the merging of new feature GC and memory64. Normally we will try our best to keep the backward compatibility of AOT ABI and not to upgrade AOT_CURRENT_VERSION, but if it is really hard to maintain (e.g. new important feature breaks the ABI), we will upgrade it.

yamt:
> the last time we upgraded AOT_CURRENT_VERSION was due to the merging of new feature GC and memory64.

actually the last time we bumped it was for reference-types. https://github.com/bytecodealliance/wasm-micro-runtime/pull/612
it was in 2021.

we didn't bump it for GC or memory64.

i guess we have to admit we often forget to bump AOT_CURRENT_VERSION. :-)

lum1n0us:
> actually the last time we bumped it was for reference-types. https://github.com/bytecodealliance/wasm-micro-runtime/pull/612
 it was in 2021.
> we didn't bump it for GC or memory64.

@TianlongLiang @wenyongh next release? or make a 2.2.1?

TianlongLiang:
Yeah, I think it's a good idea to bump it in the next release since GC and memory64 changed the ABI.

wenyongh:
> > actually the last time we bumped it was for reference-types. #612
> > it was in 2021.
> > we didn't bump it for GC or memory64.
> 
> @TianlongLiang @wenyongh next release? or make a 2.2.1?

[...]

PR #3880, opened 2024-10-24
Bump AOT_CURRENT_VERSION for WAMR 2.x (gc, memory64)

yamt:
Maybe it's too late because we have already made a few releases since then.
But this might still help users who haven't upgraded to WAMR 2.x yet. Also, for the purpose of the versioning, it's safer to bump needlessly than missing necessary bumps.

Fixes https://github.com/bytecodealliance/wasm-micro-runtime/issues/3837


wenyongh (approved):

lum1n0us:
LGTM

lum1n0us (approved):
LGTM

PR #3882, opened 2024-10-24
Refine looking up aot function with index

wenyongh:


lum1n0us:
Is it because `DefPointer(WASMExportFuncInstance *, export_functions);` now are sorted by names instead of function indexes?

wenyongh:
> Is it because `DefPointer(WASMExportFuncInstance *, export_functions);` now are sorted by names instead of function indexes?

No, event if the export_functions isn't sorted by names, it doesn't mean that it will be sorted by function index, it's in random order. This PR's optimization is suitable no matter whether export_functions is sorted by names or not.

lum1n0us:
Totally true. It sticks to the order given by the export section context, which is based on toolchains and seems random now.

So, we now need to sort the list `export_functions` by name and store them in that order, and then search by both name and function index.

wenyongh:
> Totally true. It sticks to the order given by the export section context, which is based on toolchains and seems random now.
> 
> So, we now need to sort the list `export_functions` by name and store them in that order, and then search by both name and function index.

Yes, sort `export_functions` by name, and create another map array whose element is `<func_idx, export_ith>` and sort the array by func_idx.

lum1n0us (approved):
LGTM

xujuntwt95329 (approved):
LGTM

PR #3884, opened 2024-10-25
Fix code scanning alert no. 1322: Pointer overflow check

lum1n0us:
Fixes [https://github.com/bytecodealliance/wasm-micro-runtime/security/code-scanning/1322](https://github.com/bytecodealliance/wasm-micro-runtime/security/code-scanning/1322)

To fix the problem, we need to avoid performing pointer arithmetic that could result in overflow. Instead, we should compare the index directly with the length of the array. Specifically, we should check if `bytes` is within the bounds of the memory region defined by `heap->base_addr` and `heap->size`.

- Replace the pointer arithmetic check `addr + bytes > addr` with a check that ensures `bytes` is within the valid range.
- Use the difference between the end of the memory region and the base address to perform the range check safely.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/core/iwasm/common/wasm_memory.c
+++ b/core/iwasm/common/wasm_memory.c
@@ -423,7 +423,7 @@ is_native_addr_in_shared_heap(WASMModuleInstanceCommon *module_inst,
 
     if (heap && addr >= heap->base_addr
         && addr + bytes <= heap->base_addr + heap->size
-        && addr + bytes > addr) {
+        && (size_t)bytes <= heap->base_addr + heap->size - addr) {
         return true;
     }
     return false;


lum1n0us:
@TianlongLiang

TianlongLiang:
I think we should keep the original check. The `addr` is an arbitrary value, the rhs and  could underflow and therefore condition holds true, the `addr + bytes` can still overflow, consider such example:

```C
heap->base_addr = 0x1000
heap->size = 0x1000 (so the heap ranges from 0x1000 to 0x1FFF)
addr = 0xFFFFFFFFFFFFFFFE (a large near-maximal value for a size_t, just below the maximum for a 64-bit integer)
bytes = 0x5 (a small value that will cause an overflow when added to addr)
```

heap (true)
addr >= heap->base_addr (true)
addr + bytes <= heap->base_addr + heap->size (true, overflow)
(size_t)bytes <= heap->base_addr + heap->size - addr (true, underflow)

lum1n0us:
If `addr` and `heap->base_addr` are integers, that will be OK. But, ...

When checking for integer overflow, you may often write tests like p + i < p. This works fine if p and i are unsigned integers, since any overflow in the addition will cause the value to simply "wrap around." However, **using this pattern when p is a pointer is problematic because pointer overflow has undefined behavior according to the C and C++ standards**. If the addition overflows and has an undefined result, the comparison will likewise be undefined; it may produce an unintended result, or may be deleted entirely by an optimizing compiler.

IIUC, `addr + bytes <= heap->base_addr + heap->size` shares the same pointer overflow problem.

TianlongLiang:
So how about converting them to unsigned int first and doing the comparison to avoid the undefined behavior?

lum1n0us:
That's a way. 

I am going to create a new PR. don't know how to update this one. :)

PR #3886, opened 2024-10-27
Fix out of bounds issue in is_native_addr_in_shared_heap function

lum1n0us:
When checking for integer overflow, you may often write tests like p + i < p. This works fine if p and i are unsigned integers, since any overflow in the addition will cause the value to simply "wrap around." However, using this pattern when p is a pointer is problematic because pointer overflow has undefined behavior according to the C and C++ standards. If the addition overflows and has an undefined result, the comparison will likewise be undefined; it may produce an unintended result, or may be deleted entirely by an optimizing compiler.

### DIFF
--- a/core/iwasm/common/wasm_memory.c
+++ b/core/iwasm/common/wasm_memory.c
@@ -420,13 +420,31 @@ is_native_addr_in_shared_heap(WASMModuleInstanceCommon *module_inst,
                               uint8 *addr, uint32 bytes)
 {
     WASMSharedHeap *heap = get_shared_heap(module_inst);
+    uintptr_t base_addr;
+    uintptr_t addr_int;
+    uintptr_t end_addr;
 
-    if (heap && addr >= heap->base_addr
-        && addr + bytes <= heap->base_addr + heap->size
-        && addr + bytes > addr) {
-        return true;
+    if (!heap) {
+        return false;
     }
-    return false;
+
+    base_addr = (uintptr_t)heap->base_addr;
+    addr_int = (uintptr_t)addr;
+    if (addr_int < base_addr) {
+        return false;
+    }
+
+    end_addr = addr_int + bytes;
+    /* Check for overflow */
+    if (end_addr <= addr_int) {
+        return false;
+    }
+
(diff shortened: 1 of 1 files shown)


core/iwasm/common/wasm_memory.c
```diff
@@ -420,13 +420,28 @@ is_native_addr_in_shared_heap(WASMModuleInstanceCommon *module_inst,
...
     WASMSharedHeap *heap = get_shared_heap(module_inst);
+    uintptr_t base_addr = (uintptr_t)heap->base_addr;
+    uintptr_t addr_int = (uintptr_t)addr;
+    uintptr_t end_addr = addr_int + bytes;
 
-    if (heap && addr >= heap->base_addr
-        && addr + bytes <= heap->base_addr + heap->size
-        && addr + bytes > addr) {
-        return true;
+    if (!heap) {
```
wenyongh, line 427: Should check heap before accessing `heap->base_addr`, L423.

core/iwasm/common/wasm_memory.c
```diff
@@ -420,13 +420,28 @@ is_native_addr_in_shared_heap(WASMModuleInstanceCommon *module_inst,
...
+        return false;
     }
-    return false;
+
+    if (addr_int < base_addr) {
+        return false;
+    }
+
+    // Check for overflow
```
wenyongh, line 435: had better use `/* .. */` to add the comment, and how about mentioning `integer overflow`?

core/iwasm/common/wasm_memory.c
```diff
@@ -420,13 +420,28 @@ is_native_addr_in_shared_heap(WASMModuleInstanceCommon *module_inst,
...
+    if (end_addr < addr_int) {
```
wenyongh, line 436: Is it better to use `<=` to reject `bytes == 0`?

wenyongh (approved):
LGTM

TianlongLiang:
LGTM

xujuntwt95329 (approved):
LGTM

PR #3889, opened 2024-10-28
fix(ios): Remove `float-abi` flag

xdoardo:
Compiling `wamr` targeting `iOS` results in
```
clang: error: unsupported option '-mfloat-abi=' for target 'aarch64-apple-ios'
```

This small patch removes this setting in the related `CMakeLists.txt` file.


wenyongh (approved):

lum1n0us (approved):
LGTM

PR #3891, opened 2024-10-29
Refactor AOT loader to support compatible versions

lum1n0us:
This commit refactors the AOT loader in `aot_loader.c` to support compatible versions of the AOT_CURRENT_VERSION constant. Previously, the loader only accepted the exact AOT_CURRENT_VERSION value, but now it also accepts version 3. This change ensures that the runtime can load modules AoT-compiled with different versions of wamrc as long as they have compatible AOT_CURRENT_VERSION values.

Related to #3880

### DIFF
--- a/core/iwasm/aot/aot_loader.c
+++ b/core/iwasm/aot/aot_loader.c
@@ -4236,6 +4236,16 @@ fail:
 }
 
 static bool
+aot_compatible_version(uint32 version)
+{
+    /*
+     * refer to "AoT-compiled module compatibility among WAMR versions" in
+     * ./doc/biuld_wasm_app.md
+     */
+    return version == 4 || version == 3;
+}
+
+static bool
 load(const uint8 *buf, uint32 size, AOTModule *module,
      bool wasm_binary_freeable, bool no_resolve, char *error_buf,
      uint32 error_buf_size)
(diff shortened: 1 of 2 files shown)


core/iwasm/aot/aot_loader.c
```diff
@@ -4235,6 +4235,12 @@ create_sections(AOTModule *module, const uint8 *buf, uint32 size,
     return false;
 }
 
+static bool
+aot_compatible_version(uint32 version)
+{
+    return version == AOT_CURRENT_VERSION || version == 3;
```
Zzzabiyaka, line 4241: Just checking whether I got it right 

I think this code implies that aot version 3 is compatible with everything including newer versions 

Is it something we would like to commit to?

lum1n0us: > aot version 3 is compatible with everything including newer versions

Not everything but only 4. So maybe this is better?
```suggestion
    return version == 4 || version == 3;
```

---

in *build_wasm_app.md*:

| WAMR version | AOT_CURRENT_VERSION | Compatible AOT version |
| ------------ | ------------------- | ---------------------- |
| next         | 4                   | 3,4                    |

TianlongLiang:
LGTM

xujuntwt95329 (approved):
LGTM

wenyongh (approved):

PR #3892, opened 2024-11-01
Wasm loader enhancement: check code size in code entry 

TianlongLiang:
Add wasm loader check: in code entry, the code size should match the size of vec(locals) + expr, and expr should end with opcode end

### DIFF
--- a/core/iwasm/interpreter/wasm_loader.c
+++ b/core/iwasm/interpreter/wasm_loader.c
@@ -3614,7 +3614,7 @@ load_function_section(const uint8 *buf, const uint8 *buf_end,
              * + expr(at least 1 for opcode end). And expressions are encoded by
              * their instruction sequence terminated with an explicit 0x0B
              * opcode for end. */
-            if (p_code_end - 1 < p_code || *(p_code_end - 1) != WASM_OP_END) {
+            if (p_code_end <= p_code || *(p_code_end - 1) != WASM_OP_END) {
                 set_error_buf(
                     error_buf, error_buf_size,
                     "section size mismatch: function body END opcode expected");
@@ -15848,15 +15848,12 @@ re_scan:
     }
 
     if (loader_ctx->csp_num > 0) {
-        if (cur_func_idx < module->function_count - 1)
-            /* Function with missing end marker (between two functions) */
-            set_error_buf(error_buf, error_buf_size, "END opcode expected");
-        else
-            /* Function with missing end marker
-               (at EOF or end of code sections) */
-            set_error_buf(error_buf, error_buf_size,
-                          "unexpected end of section or function, "
-                          "or section size mismatch");
+        /* unmatched end opcodes result from unbalanced control flow structures,
+         * for example, br_table with inconsistent target count (1 declared, 2
+         * given), or simply superfluous end opcodes */
+        set_error_buf(
+            error_buf, error_buf_size,
+            "unexpected end opcodes from unbalanced control flow structures");
         goto fail;
     }
 
--- a/core/iwasm/interpreter/wasm_loader.c
+++ b/core/iwasm/interpreter/wasm_loader.c
@@ -3610,6 +3610,17 @@ load_function_section(const uint8 *buf, const uint8 *buf_end,
 #endif
             }
 
+            /* Code size in code entry can't be smaller than size of vec(locals)
+             * + expr(at least 1 for opcode end). And expressions are encoded by
+             * their instruction sequence terminated with an explicit 0x0B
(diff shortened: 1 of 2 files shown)


core/iwasm/interpreter/wasm_loader.c
```diff
@@ -3610,6 +3610,17 @@ load_function_section(const uint8 *buf, const uint8 *buf_end,
...
+             * + expr(at least 1 for opcode end). And expressions are encoded by
+             * their instruction sequence terminated with an explicit 0x0B
+             * opcode for end. */
+            if (p_code_end - 1 < p_code || *(p_code_end - 1) != WASM_OP_END) {
+                set_error_buf(
+                    error_buf, error_buf_size,
+                    "section size mismatch: function body END opcode expected");
+                return false;
+            }
+
```
lum1n0us, line 3623: For me, it is better check `WASM_OP_END` when learning `code_size`. It will be L3572

wenyongh: Whether the code ends with `WASM_OP_END` will be checked in the end of wasm_loader_prepare_bytecode, seems here we don't need to check again?
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_loader.c#L15839-L15850

And is it better to use `p_code_end <= p_code` instead of `p_code_end - 1 < p_code`?

And had better also add similar check (bh_assert) in wasm_mini_loader.c.

lum1n0us: My suggestion is not right. Here is how wasm-interp in wabt execute:

``` shell
$ /opt/wabt-1.0.36/bin/wasm-interp -v ./clusterfuzz-testcase-minimized-wasm_mutator_fuzz_loader-5465932459212800 
BeginModule(version: 1)
  BeginTypeSection(6)
    OnTypeCount(1)
    OnFuncType(index: 0, params: [f32], results: [i32])
  EndTypeSection
  BeginFunctionSection(2)
    OnFunctionCount(1)
    OnFunction(index: 0, sig_index: 0)
  EndFunctionSection
  BeginCodeSection(24)                       // not check END here as I guessed
    OnFunctionBodyCount(1)        
    BeginFunctionBody(0, size:2)
    OnLocalDeclCount(1)                      // check local here
    OnLocalDecl(index: 0, count: 32, type: f64)   
                                             // Expected the END for function body before 
                                             // reading next?
000001b: error: function body must end with END opcode 
```

TianlongLiang:
So how about I modify the loader logic as follows:

1. check code size > len(vec(locals)) + 1 and check the END opcode at the end of the function body in `load_function_section` to early stop some mutated wasm file loader checking
2. in `wasm_loader_prepare_bytecode`, at the end only check for unbalanced control flow structures now, since it won't be unmatched code size or missing end opcodes.

core/iwasm/interpreter/wasm_loader.c
wenyongh: You mean that wabt checks whether the last byte of the code is `END` first? It may be good, but it is a little strange that normally the loader checks the bytecode one by one, if the last byte is not `END`, it may be an invalid bytecode (e.g. a unknown opcode or a malformed byte of a bytecode), so it is hard to say what we should report. wabt's behavior may be not standard compatible, could we build out the interpreter of wasm spec and use it to test?

lum1n0us: > You mean that wabt checks whether the last byte of the code is END first

No. I used to think so. But the verbose information shows another way.

TianlongLiang: >  could we build out the interpreter of wasm spec and use it to test?
The spec interpreter output differs from wabt. 
```bash
./wasm -t -d ../../wasm-micro-runtime/clusterfuzz.wasm
-- Running ("(input \"../../wasm-micro-runtime/clusterfuzz.wasm\")")...
-- Parsing...
-- Running...
-- Loading (../../wasm-micro-runtime/clusterfuzz.wasm)...
-- Decoding...
-- Running...
-- Decoding...
-- Error: 
../../wasm-micro-runtime/clusterfuzz.wasm:0x2e: decoding error: unexpected end of section or function
```

wenyongh: OK, if the spec interpreter also checks whether the code ends with op END first, this PR is good.

wenyongh (approved):
LGTM

lum1n0us (approved):
LGTM

PR #3894, opened 2024-11-04
Enable ref types by default

Zzzabiyaka:
After Rust 1.82.0 was released enabling reftypes by default for rust wasm target https://blog.rust-lang.org/2024/09/24/webassembly-targets-change-in-default-target-features.html#enabling-reference-types-by-default I think this would be helpful for users to have this enabled by default in runtime as well, so that rust file compiled to wasm with the default flags can be run on wamr compiled with the default flags

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -113,8 +113,8 @@ if (NOT DEFINED WAMR_BUILD_SIMD)
 endif ()
 
 if (NOT DEFINED WAMR_BUILD_REF_TYPES)
-  # Disable reference types by default
-  set (WAMR_BUILD_REF_TYPES 0)
+  # Enable reference types by default
+  set (WAMR_BUILD_REF_TYPES 1)
 endif ()
 
 set (WAMR_ROOT_DIR ${CMAKE_CURRENT_SOURCE_DIR})
--- a/doc/build_wamr.md
+++ b/doc/build_wamr.md
@@ -223,7 +223,7 @@ Currently we only profile the memory consumption of module, module_instance and 
 > See [basic sample](../samples/basic/src/main.c) for a usage example.
 
 #### **Enable reference types feature**
-- **WAMR_BUILD_REF_TYPES**=1/0, default to disable if not set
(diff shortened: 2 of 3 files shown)


loganek:
I'm not against this change, however, I think by default WAMR tries to be as lean as possible. Not sure if @wenyongh and others have any thoughts on that.

lum1n0us:
I'm also struggling with this. On one side, ref-types are now part of the specification and should be supported by default by any WebAssembly runtime. On the other side, we need to keep the binary size as small as possible.

Here's a suggestion:
- We always release binaries of iwasm and wamrc that are fully compatible with the specification and offer better performance. This way, users can easily run WebAssembly code from Rust.
- (Optional) We provide separate compilation groups for full specification compatibility and for minimal binary size through something like CMakePresets.json. This allows developers to switch quickly.
- By default, we enable full specification compatibility. I mean enable ref-types by default.

Let's make a vote ?

wenyongh:
It is good to me if it brings convenience and has little impact on footprint and performance. And yes, not sure what others' opinions are, maybe a vote is better.

wenyongh:
BTW, if to enable ref-types by default, had better also change the CMakeLists.txt under `product-mini/platforms/windows` and `product-mini/platforms/cosmopolitan`.

Zzzabiyaka:
@wenyongh, how do we run the vote?

should I just ask all TSC members to 👍  or 👎 the initial pr comment?

lum1n0us (approved):
LGTM

lum1n0us:
https://github.com/bytecodealliance/wasm-micro-runtime/pull/3913#issue-2671535298

> The recent version of the rust toolchain will emit ref types opcodes, need to enable this feature in the iwasm build.

It is a strong reason we need this one now.

Zzzabiyaka:
Yes, @lum1n0us, that was the reason for this pr in the first place and it's in the linked article 

Maybe I should have been clearer about the reason

xujuntwt95329 (approved):
LGTM

PR #3897, opened 2024-11-08
fix(uwp): Gate NTSTATUS definition behind WINAPI_PARTITION_DESKTOP for UWP build 

bafadumi:
UWP build showed that this API wasn’t gated and is not supported by UWP. This error was discovered when porting WAMR2.2.0

Error : `unknown type name 'NTSTATUS'; did you mean 'LSTATUS'? `

### DIFF
--- a/core/shared/platform/windows/win_clock.c
+++ b/core/shared/platform/windows/win_clock.c
@@ -10,9 +10,11 @@
 #define NANOSECONDS_PER_SECOND 1000000000ULL
 #define NANOSECONDS_PER_TICK 100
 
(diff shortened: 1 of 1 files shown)


lum1n0us (approved):
LGTM

loganek:
I see the usage of this code was already gated in https://github.com/bytecodealliance/wasm-micro-runtime/blob/397f6633491052a3d4607039f6a87834b9a8524e/core/shared/platform/windows/win_clock.c#L63 so I think it's a valid change.

loganek (approved):

PR #3899, opened 2024-11-09
GlobalValueSet was moved to IRPartitionLayer recently, but we have a …

sjamesr:
…local definition anyway

### DIFF
--- a/core/iwasm/compilation/aot_orc_extra.cpp
+++ b/core/iwasm/compilation/aot_orc_extra.cpp
@@ -177,7 +177,7 @@ LLVMOrcLLLazyJITBuilderSetJITTargetMachineBuilder(
     LLVMOrcDisposeJITTargetMachineBuilder(JTMP);
 }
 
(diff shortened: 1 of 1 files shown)


sjamesr:
This resolves a compilation error against recent revisions of LLVM

lum1n0us:
> GlobalValueSet was moved to IRPartitionLayer recently

In which llvm release? WAMR is somewhat sensitive to the version of LLVM. Currently WAMR depends on LLVM 15.x

sjamesr:
It was moved about a month ago in https://github.com/llvm/llvm-project/commit/04af63b267c391a4b0a0fb61060f724f8b5bc2be. Internally at Google we build WAMR against LLVM at approximately HEAD.

In the above change, GlobalValueSet is now `using GlobalValueSet = std::set<const GlobalValue *>;` inside `IRPartitionLayer`.

I think my change is safe, to my non-expert eyes the definition of GlobalValueSet is the same before and after this change.

lum1n0us (approved):
LGTM

PR #3901, opened 2024-11-11
[instantiation linking] refactor wasm_table_inst_t

lum1n0us:
Use `WASMTableInstance` and (`#define AOTModuleInstance WASMModuleInstance`).

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -725,6 +725,7 @@ tables_instantiate(AOTModuleInstance *module_inst, AOTModule *module,
 #if WASM_ENABLE_GC != 0
         table->elem_ref_type.elem_ref_type = table_type->elem_ref_type;
 #endif
+        table->is_table64 = table_type->flags & TABLE64_FLAG;
 
 #if WASM_ENABLE_MULTI_MODULE != 0
         /* Set all elements to -1 or NULL_REF to mark them as uninitialized
@@ -788,6 +789,7 @@ tables_instantiate(AOTModuleInstance *module_inst, AOTModule *module,
 #if WASM_ENABLE_GC != 0
         table->elem_ref_type.elem_ref_type = table_type->elem_ref_type;
 #endif
+        table->is_table64 = table_type->flags & TABLE64_FLAG;
 
         /* Set all elements to -1 or NULL_REF to mark them as uninitialized
          * elements */
@@ -5493,7 +5495,8 @@ aot_const_str_set_insert(const uint8 *str, int32 len, AOTModule *module,
 #if WASM_ENABLE_DYNAMIC_AOT_DEBUG != 0
 AOTModule *g_dynamic_aot_module = NULL;
 
-void __attribute__((noinline)) __enable_dynamic_aot_debug(void)
+void __attribute__((noinline))
+__enable_dynamic_aot_debug(void)
 {
     /* empty implementation. */
 }
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -953,7 +953,8 @@ memories_deinstantiate(AOTModuleInstance *module_inst)
 
         memory_deinstantiate(memory);
 
-#if WASM_ENABLE_MULTI_MODULE == 0 && WASM_ENABLE_SHARED_MEMORY != 0
+#if WASM_ENABLE_MULTI_MODULE == 0
+#if WASM_ENABLE_SHARED_MEMORY != 0
         /* for spawned only */
         if (!shared_memory_is_shared(memory)) {
             continue;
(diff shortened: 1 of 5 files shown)


lum1n0us:
@bnason-nf 

It's not just for consistency. It's also beneficial for tasks such as creating a table from the host, importing an exported table from an instance, sharing table elements between the host and WebAssembly, between different WebAssembly modules, and for setting or getting a table from the host or another module. Additionally, it simplifies the management of ownership for `wasm_table_inst_t`.

lum1n0us:
I think there are two methods (possibly more). Please share your thoughts.

- A new function `wasm_runtime_get_export_index(const wasm_module_t, wasm_import_export_kind_t, const char*)` retrieves the `export_index` using the name. An existing function `wasm_runtime_get_export_type()` returns `wasm_export_t` from the `export_index`. From `wasm_export_t`, since it's known to be a table, you can get `wasm_table_type_t`. Then, a series of existing functions `wasm_table_type_get_XXX(const wasm_table_type_t)` can complete the tasks. We could certainly add some helper functions to shorten this process.

- Or A new function `wasm_table_type_t wasm_runtime_get_table_type(xxx, wasm_table_inst_t)` provides the `wasm_table_type_t` for a specified `wasm_table_inst_t`.

bnason-nf:
I guess the second way sounds nicer to me, but I don't have a strong opinion.

core/iwasm/aot/aot_runtime.c
```diff
@@ -704,6 +708,8 @@ tables_instantiate(AOTModuleInstance *module_inst, AOTModule *module,
             tbl_inst->elem_ref_type.elem_ref_type =
                 module->tables[i].table_type.elem_ref_type;
 #endif
+            tbl_inst->is_table64 = table->table_type.flags & TABLE64_FLAG;
+            tbl_inst->is_shared = table->table_type.flags & SHARED_TABLE_FLAG;
```
lum1n0us, line 712: Still necessary. L751 will use that.

lum1n0us:
Because there is no member such as `WASMTableType` in `WASMTableInstanc`e, there is always an issue with allocating and freeing resources for an API like `wasm_table_type_t wasm_get_table_type(..., wasm_table_inst_t, ...)`.

Later, I understood that the actual issue isn't obtaining a `wasm_table_type_t` for any `wasm_table_inst_t`, but rather for an exported one. If the `wasm_table_inst_t` was created by the host, the type details are already known. So here is `wasm_runtime_get_export_type_by_name()`.

lum1n0us:
BTW, for the same purpose and idea, also need to refactor `wasm_global_inst_t`. If agreed, will do it another PR

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -7827,43 +7842,27 @@ wasm_runtime_create_table(WASMModuleCommon *const module,
...
     if (module->module_type == Wasm_Module_Bytecode) {
-        wasm_table_inst_t *out_table =
-            runtime_malloc(sizeof(wasm_table_inst_t), NULL, NULL, 0);
-        if (!out_table)
-            return NULL;
-
         WASMTableInstance *wasm_table =
             wasm_create_table((WASMModule *)module, type);
         if (!wasm_table) {
             wasm_runtime_free(wasm_table);
```
wenyongh, line 7848: had better remove this line, which frees a null pointer

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -7827,43 +7842,27 @@ wasm_runtime_create_table(WASMModuleCommon *const module,
...
     if (module->module_type == Wasm_Module_AoT) {
-        wasm_table_inst_t *out_table =
-            runtime_malloc(sizeof(wasm_table_inst_t), NULL, NULL, 0);
-        if (!out_table)
-            return NULL;
-
         AOTTableInstance *aot_table =
             aot_create_table((AOTModule *)module, type);
         if (!aot_table) {
             wasm_runtime_free(aot_table);
```
wenyongh, line 7861: same as above

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -282,7 +283,8 @@ memories_deinstantiate(WASMModuleInstance *module_inst)
         memory_deinstantiate(memory);
 #endif
 
-#if WASM_ENABLE_MULTI_MODULE == 0 && WASM_ENABLE_SHARED_MEMORY != 0
+#if WASM_ENABLE_MULTI_MODULE == 0
+#if WASM_ENABLE_SHARED_MEMORY != 0
         /* for spawned only */
         if (!shared_memory_is_shared(memory)) {
             continue;
```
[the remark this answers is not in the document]
lum1n0us: Yes. need to `wasm_runtime_free(memory)`. 

`continue` is still necessary because `shared_memory_get_reference()` works only when ` bh_assert(shared_memory_is_shared(memory));`.

core/iwasm/common/wasm_runtime_common.c
lum1n0us: ~ So confused that why I used a `wasm_runtime_free()` here? 😕

wenyongh (approved):
LGTM

PR #3905, opened 2024-11-12
Fix global initialization

sjamesr:
While resolving linked globals, WAMR tries to copy the linked global's initial value into the destination global in the current module. However, a bug in the implementation causes the copy to be done from the InitializerExpression struct, not from its WASMValue field.

This did not come up in WAMR's spec test runner because those are built with WASM_ENABLE_SPEC_TEST, which means these globals are resolved as builtins, not linked globals, which goes through a different (presumably not faulty) path.


lum1n0us (approved):
👍   nice catch.

sjamesr:
Thank you! I won't lie, it took a good day of debugging to figure this one out. :smile:

xujuntwt95329 (approved):
LGTM

no1wudi (approved):

PR #3907, opened 2024-11-12
don't ignore wasm_resolve_import_func return value

sjamesr:
Previously the loader would ignore the return value of wasm_resolve_import_func, which would mean it would report loading success even if there were unresolved function imports.


sjamesr:
I'm guessing this is working as intended, the functions may be unlinked at load time but may be linked later when other modules are loaded.

PR #3914, opened 2024-11-19
[instantiation linking] Create and import WASMGlobalInstance

lum1n0us:


core/iwasm/compilation/aot_emit_variable.c
```diff
@@ -149,57 +150,156 @@ aot_compile_op_tee_local(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
...
         global_type = VALUE_TYPE_GC_REF;
 
-    offset = I32_CONST(global_offset);
-    if (!(global_ptr = LLVMBuildInBoundsGEP2(comp_ctx->builder, INT8_TYPE,
-                                             func_ctx->aot_inst, &offset, 1,
-                                             "global_ptr_tmp"))) {
-        aot_set_last_error("llvm build in bounds gep failed.");
-        return false;
-    }
+    global_ptr = get_global_value_addr(comp_ctx, func_ctx, global_idx);
```
wenyongh, line 302: The original code directly gets the global offset at the compilation time and loads global one time in execution. Per my understanding, `get_global_value_addr` will get the global address by several loads, it will reduce the performance? It seems that it is only needed for import global and only needed when multi-module is enabled?

core/iwasm/aot/aot_loader.c
```diff
@@ -2182,9 +2185,12 @@ load_import_globals(const uint8 **p_buf, const uint8 *buf_end,
                 tmp_global.global_data_linked;
             import_globals[i].is_linked = true;
         }
-#else
-        import_globals[i].is_linked = false;
+        else
 #endif
+#endif /* WASM_ENABLE_MULTI_MODULE != 0 */
```
wenyongh, line 2190: When multi-module isn't enabled and libc-builtin is enabled, the global isn't linked by default, is it expected? Or will the iwasm's main.c link the global when instantiating?

core/iwasm/aot/aot_runtime.h
```diff
@@ -127,6 +132,9 @@ typedef struct AOTModuleInstanceExtra {
     DefPointer(uint8 *, shared_heap_base_addr_adj);
     MemBound shared_heap_start_off;
 
+    DefPointer(AOTGlobalInstance *, globals);
```
wenyongh, line 135: Had better also add `bh_static_assert(offsetof(AOTModuleInstanceExtra, globals) == 24);` in aot_runtime.c, like the `bh_static_assert` for `shared_heap_start_off`:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/aot/aot_runtime.c#L62

core/iwasm/compilation/aot_emit_variable.c
```diff
@@ -149,57 +150,156 @@ aot_compile_op_tee_local(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
     return aot_compile_op_set_or_tee_local(comp_ctx, func_ctx, local_idx, true);
 }
 
+/*TODO: should be optimized by moving globals to WASMModuleInstance */
+LLVMValueRef
+get_global_from_wasm_inst(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
+                          uint32 global_idx)
+{
+    /* WASMModuleInstance->e */
+    uint32 e_offset_val = offsetof(WASMModuleInstance, e);
```
wenyongh, line 159: The `WASMModuleInstanceExtra` structure is put at the end of `WASMModuleInstance`, we can use a fixed offset to get it and its fields, like `offset_u32 = get_module_inst_extra_offset(comp_ctx)` and then `offset = offset_u32 + offsetof(AOTModuleInstanceExtra, globals)` for AOT or `offset = offset_u32 + offsetof(WASMModuleInstanceExtra, globals)` for JIT. By this way, we can save one load operation, refer to `create_shared_heap_info`:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/compilation/aot_llvm.c#L1528-L1545

core/iwasm/compilation/aot_emit_variable.c
```diff
@@ -149,57 +150,156 @@ aot_compile_op_tee_local(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
...
+    LLVMValueRef e_offset = I32_CONST(e_offset_val);
+    LLVMValueRef e_ptr_u8 =
+        LLVMBuildInBoundsGEP2(comp_ctx->builder, INT8_TYPE, func_ctx->aot_inst,
+                              &e_offset, 1, "e_ptr");
+    LLVMValueRef e_ptr =
+        LLVMBuildBitCast(comp_ctx->builder, e_ptr_u8, OPQ_PTR_TYPE, "e_ptr");
+    LLVMValueRef e =
+        LLVMBuildLoad2(comp_ctx->builder, OPQ_PTR_TYPE, e_ptr, "e");
```
wenyongh, line 167: had better check return value of e_offset, e_ptr_u8, e_ptr, e

core/iwasm/compilation/aot_emit_variable.c
```diff
@@ -149,57 +150,156 @@ aot_compile_op_tee_local(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
...
+    else {
+        /* AOTModuleInstanceExtra->globals */
+        globals_offset_val = offsetof(AOTModuleInstanceExtra, globals);
+    }
+    LLVMValueRef globals_offset = I32_CONST(globals_offset_val);
+    LLVMValueRef globals_ptr = LLVMBuildInBoundsGEP2(
+        comp_ctx->builder, INT8_TYPE, e, &globals_offset, 1, "globals_ptr");
+
+    LLVMValueRef globals =
+        LLVMBuildLoad2(comp_ctx->builder, OPQ_PTR_TYPE, globals_ptr, "globals");
```
wenyongh, line 183: had better check return value

core/iwasm/compilation/aot_emit_variable.c
```diff
@@ -149,57 +150,156 @@ aot_compile_op_tee_local(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
...
+    return global_ptr;
+}
+
+static LLVMValueRef
+get_global_value_addr(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
+                      uint32 global_idx)
+{
+    /* WASMGlobalInstance[global_idx] */
+    LLVMValueRef global_ptr =
+        get_global_from_wasm_inst(comp_ctx, func_ctx, global_idx);
```
wenyongh, line 200: same as above, had better check return values in this function.

core/iwasm/compilation/aot_emit_variable.c
wenyongh: OK, since performance and code size are important for the developers, hope the new feature impacts them as less as we can.

core/iwasm/aot/aot_loader.c
lum1n0us: YES. only apply `wasm_native_lookup_libc_builtin_global()` when executing loading linking. Otherwise, `core/iwasm/libraries/libc-builtin/spec_test_builtin_wrapper.c` will take over

wenyongh (approved):
LGTM

PR #3916, opened 2024-11-20
add testcases for shared heap and fix POP_MEM_OFFSET of memory64

WenLY1:
fix POP_MEM_OFFSET of memory64:
When WAMR_BUILD_MEMORY64 is enabled but is_memory64 is false, if the addr is larger than 0x80000000 and the addr is directly converted to uint64 from int32, the result is wrong, as follows

### DIFF
--- a/core/iwasm/interpreter/wasm_interp_classic.c
+++ b/core/iwasm/interpreter/wasm_interp_classic.c
@@ -593,8 +593,8 @@ wasm_interp_get_frame_ref(WASMInterpFrame *frame)
 #endif
 
 #if WASM_ENABLE_MEMORY64 != 0
-#define POP_MEM_OFFSET() (is_memory64 ? POP_I64() : POP_I32())
-#define POP_TBL_ELEM_IDX() (is_table64 ? POP_I64() : POP_I32())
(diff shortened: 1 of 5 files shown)


core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -590,7 +590,7 @@ wasm_interp_get_frame_ref(WASMInterpFrame *frame)
 #endif
 
 #if WASM_ENABLE_MEMORY64 != 0
-#define POP_MEM_OFFSET() (is_memory64 ? POP_I64() : POP_I32())
+#define POP_MEM_OFFSET() (is_memory64 ? POP_I64() : (uint32)POP_I32())
```
TianlongLiang, line 593: Maybe it's a good idea to PR directly to the main branch, and in main `POP_TBL_ELEM_IDX()` also has this problem.

wenyongh: Agree, branch `dev/shared_heap` had been merged to main. And main has resolved the CI wamr-ide test failure.

WenLY1: Done

lum1n0us (approved):
LGTM

PR #3919, opened 2024-11-20
Fix replacement value not being correct

Zzzabiyaka:
Changing this call allows to correctly implement replace opcode 

with existing call the following wat was failing 

```
    v128.const i32x4 10 11 12 13
    i32.const 15
    i32x4.replace_lane 3
    i32x4.extract_lane 3
    call $assert_true_i32
```

with replace on my branch looking like that

```
                    case SIMD_i32x4_replace_lane:
                    {
                        uint8_t lane = *frame_ip++;
                        int rep = POP_I32();
                        V128 v = POP_V128();

                        v.i32x4[lane] = rep;
                        addr_ret = GET_OFFSET();
                        PUT_V128_TO_ADDR(frame_lp + addr_ret, v);
                        break;
                    }
 ```

### DIFF
--- a/core/iwasm/interpreter/wasm_loader.c
+++ b/core/iwasm/interpreter/wasm_loader.c
@@ -15217,17 +15217,27 @@ re_scan:
                                                     error_buf_size)) {
                             goto fail;
                         }
+#if WASM_ENABLE_FAST_INTERP != 0
+                        emit_byte(loader_ctx, lane);
+#endif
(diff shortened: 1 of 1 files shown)


Zzzabiyaka:
@wenyongh @loganek Could you please take a look at this PR as I am not sure I have the full range of context about this to confidently say this is the right fix 

Thanks

wenyongh (approved):
LGTM

loganek (commented):
LGTM

loganek (approved):

Issue #2764, opened 2023-11-14
passive/declarative segments should be dropped on instantiation


yamt:
it doesn't seem implement correctly, at least for the classic interpreter.

yamt:
i guess i will take a look after fixing https://github.com/bytecodealliance/wasm-micro-runtime/issues/2735

sjamesr:
I just came across this while trying to eliminate spec test failures, is this as straightforward as something like:

```
diff --git a/core/iwasm/interpreter/wasm_runtime.c b/core/iwasm/interpreter/wasm_runtime.c
--- a/core/iwasm/interpreter/wasm_runtime.c
+++ b/core/iwasm/interpreter/wasm_runtime.c
@@ -2467,8 +2467,9 @@ wasm_instantiate(WASMModule *module, WAS
             goto fail;
         }
         for (i = 0; i < module->table_seg_count; i++) {
-            if (wasm_elem_is_active(module->table_segments[i].mode))
-                bh_bitmap_set_bit(module_inst->e->common.elem_dropped, i);
+          if (wasm_elem_is_active(module->table_segments[i].mode) ||
+              wasm_elem_is_declarative(module->table_segments[i].mode))
+            bh_bitmap_set_bit(module_inst->e->common.elem_dropped, i);
         }
     }
 #endif

```

PR #3922, opened 2024-11-21
Drop declarative elements on module instantiation

sjamesr:
This may be sufficient to address https://github.com/bytecodealliance/wasm-micro-runtime/issues/2764. It certainly makes a few more spec tests pass for me.

### DIFF
--- a/core/iwasm/interpreter/wasm_runtime.c
+++ b/core/iwasm/interpreter/wasm_runtime.c
@@ -2467,7 +2467,8 @@ wasm_instantiate(WASMModule *module, WASMModuleInstance *parent,
             goto fail;
         }
         for (i = 0; i < module->table_seg_count; i++) {
-            if (wasm_elem_is_active(module->table_segments[i].mode))
+            if (wasm_elem_is_active(module->table_segments[i].mode)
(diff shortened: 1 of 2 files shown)


core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -2467,7 +2467,8 @@ wasm_instantiate(WASMModule *module, WASMModuleInstance *parent,
             goto fail;
         }
         for (i = 0; i < module->table_seg_count; i++) {
-            if (wasm_elem_is_active(module->table_segments[i].mode))
+            if (wasm_elem_is_active(module->table_segments[i].mode)
+                || wasm_elem_is_declarative(module->table_segments[i].mode))
```
wenyongh, line 2471: Should also apply the change to aot_runtime.c?
https://github.com/bytecodealliance/wasm-micro-runtime/blob/62aca1727908848c194dfe52c7d62397f5df8441/core/iwasm/aot/aot_runtime.c#L1907-L1910

wenyongh (approved):

lum1n0us (approved):
LGTM

PR #3923, opened 2024-11-21
Fix WASI Path Mapping Processing

dpjohnst:
Filesystem paths can be mapped from the host path to a guest path using the format `<guest-path>::<host-path>`.

Previously `strtok` was used to find the `::` delimiter. Unfortunately `strtok` processes each delimiter character individually. This meant that the code was ~equivalent to `strtok(mapping_copy, ":")` which breaks with Windows-style paths (E.g. `C:\my_path\`).

To fix this `strstr` is used to search for the exact delimiter.

### DIFF
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -3631,8 +3631,18 @@ wasm_runtime_init_wasi(WASMModuleInstanceCommon *module_inst,
 
         bh_memcpy_s(mapping_copy, max_len, map_dir_list[i],
                     (uint32)(strlen(map_dir_list[i]) + 1));
-        map_mapped = strtok(mapping_copy, "::");
-        map_host = strtok(NULL, "::");
+
+        const char *delim = "::";
+        char *delim_pos = strstr(mapping_copy, delim);
+        if (delim_pos) {
+            *delim_pos = '\0';
+            map_mapped = mapping_copy;
+            map_host = delim_pos + strlen(delim);
+        }
+        else {
+            map_mapped = NULL;
+            map_host = NULL;
+        }
 
         if (!map_mapped || !map_host) {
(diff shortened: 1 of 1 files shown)


core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3631,8 +3631,18 @@ wasm_runtime_init_wasi(WASMModuleInstanceCommon *module_inst,
...
+        char *delim_pos = strstr(mapping_copy, delim);
+        if (delim_pos) {
+            *delim_pos = '\0';
+            map_mapped = mapping_copy;
+            map_host = delim_pos + strlen(delim);
+        }
+        else {
+            map_mapped = NULL;
+            map_host = NULL;
+        }
```
lum1n0us, line 3645: the whole `else` block can be removed. L3619 has initialized both to NULL.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3631,8 +3631,14 @@ wasm_runtime_init_wasi(WASMModuleInstanceCommon *module_inst,
...
                     (uint32)(strlen(map_dir_list[i]) + 1));
-        map_mapped = strtok(mapping_copy, "::");
-        map_host = strtok(NULL, "::");
+
+        const char *delim = "::";
+        char *delim_pos = strstr(mapping_copy, delim);
+        if (delim_pos) {
+            *delim_pos = '\0';
+            map_mapped = mapping_copy;
+            map_host = delim_pos + strlen(delim);
```
wenyongh, line 3640: map_host may be an empty string "", do you think we should check whether `strlen(map_host) == 0`, and if yes, set map_host to NULL to let runtime report error and exit?

wenyongh (approved):

lum1n0us (approved):
LGTM

PR #3924, opened 2024-11-22
Use plain assignment rather than bh_memcpy_s

sjamesr:


lum1n0us:
May I ask the advantage of making this change?

sjamesr:
It occurred to me that I should have used this approach in https://github.com/bytecodealliance/wasm-micro-runtime/pull/3905. It's safer (wouldn't compile if the types didn't match), more succinct, and is probably implemented as a memcpy by the compiler, so it's not really changing anything.

wenyongh (approved):

lum1n0us (approved):
LGTM

PR #3926, opened 2024-11-22
CMakeLists.txt: Do not require C++

midokura-xavi92:
By default, the project() CMake command defaults to C and C++. [1] Therefore, CMake might perform tests for both C and C++ compilers as part of the configuration phase.

However, this has the consequence of the configuration phase to fail if the system does not have a C++ toolchain installed, even if C++ is not really used by the project.

[1]: https://cmake.org/cmake/help/latest/command/project.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,7 @@ if(ESP_PLATFORM)
   return()
 endif()
 
-project (iwasm)
+project (iwasm LANGUAGES C)
 
 set (CMAKE_VERBOSE_MAKEFILE OFF)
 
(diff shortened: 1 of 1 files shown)


lum1n0us:
> even if C++ is not really used by the project.

Not true. like:

./libraries/wasi-nn/src/wasi_nn_tensorflowlite.cpp
./libraries/lib-wasi-threads/unit-test/test_tid_allocator.cpp
./compilation/aot_orc_extra2.cpp
./compilation/aot_llvm_extra2.cpp
./compilation/debug/dwarf_extractor.cpp
./compilation/aot_llvm_extra.cpp
./compilation/aot_orc_extra.cpp
./fast-jit/cg/x86-64/jit_codegen_x86_64.cpp

loganek:
@midokura-xavi92 I'm closing the PR as it looks like there are indeed C++ files for AOT/JIT modes.  Feel free to re-open the PR with more context though.

yamt:
i guess it isn't about the permission.
Screenshot 2024-12-11 at 19 09 25

yamt:
i don't know why. but github has always been that way as far as i know.
cf. https://github.com/isaacs/github/issues/361
while there seems to be a workaround, i guess it's simpler to create another PR and mention it here.

PR #3933, opened 2024-11-26
Add versioning support and update CMake configuration

lum1n0us:


lum1n0us:
@jirutka  Please inform me if platforms other than Linux require the SONAME feature.

jirutka:
The same applies to all Unix and Unix-like systems, so also BSDs and macOS.

yamt:
maybe you can do it for all platforms and let cmake do whatever appropriate?
https://cmake.org/cmake/help/latest/prop_tgt/SOVERSION.html

yamt:
build-tested for macOS.

master:
```
libiwasm.dylib:
        @rpath/libiwasm.dylib (compatibility version 0.0.0, current version 0.0.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1351.0.0)
```

with this PR:
```
libiwasm.dylib:
        @rpath/libiwasm.2.dylib (compatibility version 2.0.0, current version 2.2.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1351.0.0)
```

iirc, on macOS, a dylib can be loaded if
* the filename matches
* and compatibility version is larger than requested

core/version.h
```diff
@@ -3,9 +3,20 @@
  * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
  */
 
+/*
+ * version.h.in is a template file.  version.h is a generated file.
+ * Please do not edit both files directly.
+ *
+ * Any changes to the version should be done in the version.cmake file.
```
yamt, line 10: can you document the procedure?
eg. 1. update version.cmake, 2. run something (what?) to regen this file, 3. commit this generated file as well.

yamt (commented):
looks reasonable to me

lum1n0us:
@yamt @loganek @TianlongLiang @xujuntwt95329 @wenyongh @no1wudi 
Do you have any comments?

TianlongLiang (approved):
LGTM

doc/stability_release.md
```diff
@@ -0,0 +1,28 @@
...
+- *minor*: This number increases with the addition of new features. This encompasses not only MVP (Minimum Viable Product) or POST-MVP features but also WebAssembly System Interface (WASI) features and WAMR-specific features.
+- *patch*: This number is incremented for patches.
+
+## Legacy releases
+
+All previous versions (tags) will retain their current status. There will be no changes to existing release names and links.
+
+# Release Process
+
+WAMR has been deployed across various devices. A frequent release cycle would strain customers' testing resources and add extra deployment work. Two factors can trigger a new WAMR release:
```
loganek, line 15: > frequent release cycle would strain customers' testing resources

Not sure if I agree with that; customers can still choose to stick to a particular version or keep moving to the latest one. In fact, for some of the usecases it more frequent releases are better as it helps narrowing down any potential regressions which can only be seen when running WAMR at scale.

> and add extra deployment work

Agree with that, we should probably discuss automating the process within TSC as much as we can to streamline the releases.

core/version.h
```diff
@@ -3,9 +3,20 @@
  * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```
loganek, line 3: given we have `.in` file now, does this file still have to be checked-in?

lum1n0us: thanks for reminding.

doc/stability_release.md
lum1n0us: Are you proposing that we should streamline the release process (reduce manual work compared to the current method) and issue releases on a regular schedule, such as monthly?

core/version.h
lum1n0us: @no1wudi We are considering generating the *version.h* file from *version.h.in* during the CMake configuration process and removing version.h from the source code tree. [The Nuttx CI](https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/12705810342/job/35417551343?pr=3933) is failing because it uses make instead of CMake and cannot create the version.h file. Is there a method to still generate version.h using make?

lum1n0us: @loganek Several embedded platforms rely on Makefile instead of CMake and are unable to utilize `configure_file()` or other alternatives(like sed) to generate *version.h* from *version.h.in*. Therefore, it is advisable to continue maintaining version.h within the source code tree.

loganek: ok understood. I wonder though if we can ensure in CI that those two files (`version.h.in` and `version.h`) stay in sync?

lum1n0us: something like https://github.com/bytecodealliance/wasm-micro-runtime/pull/3933/files#diff-f275178c626161329dae1b5a6505c032a697ec141e65cdbca0df7b45112f63d3 ?

loganek: that's perfect, thanks!

doc/stability_release.md
loganek: yes, I think that'd be great to have predictable release schedule. Not sure when do we have the next TSC meeting, but let's add it to the agenda.

lum1n0us: February 3rd falls within the Chinese public holidays, so I plan to cancel the February community meeting. 

Perhaps we should open a new issue to discuss a predictable release schedule. We're completely on board with the idea. The only issue is carrying out the release tests, which are currently done manually. If you're interested, I think I could share the details of the release test cases.

core/version.h
lum1n0us: @loganek  If there are no further comments, shall we go ahead and merge the pull request?

loganek (approved):
Thanks for the changes, this looks good to me.

PR #3935, opened 2024-11-26
don't return an uninitialized trap if argv_to_results fails

sjamesr:
Currently, if argv_to_results fails (e.g. because an unsupported type is encountered), an non-null trap with an uninitialized message is returned.


lum1n0us (approved):
LGTM

wenyongh (approved):
LGTM

PR #3939, opened 2024-11-29
Fix incorrect assignment in win_file.c

dpjohnst:
win_error should be compared to ERROR_INVALID_HANDLE but was instead being assigned the value.


wenyongh (approved):

lum1n0us (approved):
LGTM

PR #3941, opened 2024-12-02
Improvements for platform thread APIs on Windows and Zephyr

TianlongLiang:
This PR includes the following modifications:

- On the Zephyr platform:
  - The `os_thread_wrapper` will call `os_thread_cleanup` to free `thread_data`, so when another thread tries to wait on its termination using `os_thread_join`, it's possible that thread is already terminated and won't be in `thread_data_list`. So simply print the warning and return `OK` if in `os_thread_join` find out the joining thread is not in the list.
- On Windows:
  - Unlike Zephyr, the `os_thread_cleanup` won't free the `thread_data` but set a flag, and the `thread_data` will still be in the list. So in `os_thread_join` if can't find the `thread_data` in the list, it must be a non-existent thread, returning an error in this case.
  - Release the resources of `thread_data` when the corresponding thread is joined, rather than delaying to release of them at the end altogether using `os_thread_sys_destroy

### DIFF
--- a/core/shared/platform/windows/win_thread.c
+++ b/core/shared/platform/windows/win_thread.c
@@ -88,6 +88,7 @@ thread_data_list_remove(os_thread_data *thread_data)
         p = p->next;
 
     if (p && p->next) {
+        bh_assert(p->next == thread_data);
         p->next = p->next->next;
         /* Release the resources in thread_data */
         os_cond_destroy(&thread_data->wait_cond);
--- a/core/shared/platform/windows/win_thread.c
+++ b/core/shared/platform/windows/win_thread.c
@@ -60,6 +60,62 @@ static DWORD thread_data_key;
 static void(WINAPI *GetCurrentThreadStackLimits_Kernel32)(PULONG_PTR,
                                                           PULONG_PTR) = NULL;
 
+static void
+thread_data_list_add(os_thread_data *thread_data)
+{
+    os_mutex_lock(&thread_data_list_lock);
+    /* If already in list, just return */
+    os_thread_data *p = &supervisor_thread_data;
+    while (p) {
+        if (p == thread_data) {
+            os_mutex_unlock(&thread_data_list_lock);
+            return;
+        }
+        p = p->next;
+    }
+    thread_data->next = supervisor_thread_data.next;
+    supervisor_thread_data.next = thread_data;
+    os_mutex_unlock(&thread_data_list_lock);
+}
+
+static void
+thread_data_list_remove(os_thread_data *thread_data)
+{
+    os_mutex_lock(&thread_data_list_lock);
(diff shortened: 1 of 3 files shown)


core/shared/platform/windows/win_thread.c
```diff
@@ -332,6 +390,7 @@ os_thread_join(korp_tid thread, void **p_retval)
     os_sem_wait(&curr_thread_data->wait_node.sem);
     if (p_retval)
         *p_retval = curr_thread_data->wait_node.retval;
+    thread_data_list_remove(thread_data);
```
lum1n0us, line 393: If `thread_data_list_remove()` thread_data now, what will happen if caller join this thread again? Can `os_thread_join()` figure out the thread has already terminated and return.

---
https://learn.microsoft.com/en-us/dotnet/api/system.threading.thread.join?view=net-9.0
>If the thread has already terminated when [Join](https://learn.microsoft.com/en-us/dotnet/api/system.threading.thread.join?view=net-9.0) is called, the method returns immediately

lum1n0us: Other limitation include:

> You should never call the [Join](https://learn.microsoft.com/en-us/dotnet/api/system.threading.thread.join?view=net-9.0) method of the [Thread](https://learn.microsoft.com/en-us/dotnet/api/system.threading.thread?view=net-9.0) object that represents the current thread from the current thread. This causes your app to become unresponsive because the current thread waits upon itself indefinitely,

> You cannot invoke Join on a thread that is in the [ThreadState.Unstarted](https://learn.microsoft.com/en-us/dotnet/api/system.threading.threadstate?view=net-9.0#system-threading-threadstate-unstarted) state.

TianlongLiang: > If thread_data_list_remove() thread_data now, what will happen if caller join this thread again? Can os_thread_join() figure out the thread has already terminated and return.

When it tries to join, L360-363 will return `BHT_ERROR`

> You should never call the [Join](https://learn.microsoft.com/en-us/dotnet/api/system.threading.thread.join?view=net-9.0) method of the [Thread](https://learn.microsoft.com/en-us/dotnet/api/system.threading.thread?view=net-9.0) object that represents the current thread from the current thread. This causes your app to become unresponsive because the current thread waits upon itself indefinitely

It will deadlock too, L390 sem wait forever

> You cannot invoke Join on a thread that is in the [ThreadState.Unstarted](https://learn.microsoft.com/en-us/dotnet/api/system.threading.threadstate?view=net-9.0#system-threading-threadstate-unstarted) state.

Our implementation uses `_beginthread` and didn't use Thread object, won't have this issue

wenyongh (approved):
LGTM

lum1n0us (approved):
LGTM

PR #3943, opened 2024-12-03
Add zero load opcodes

Zzzabiyaka:
tested using a simple wat again

### DIFF
--- a/core/iwasm/interpreter/wasm_interp_fast.c
+++ b/core/iwasm/interpreter/wasm_interp_fast.c
@@ -6391,14 +6391,8 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                         break;
                     }
 
-#define SIMD_LOAD_LANE_OP(register, width)                 \
+#define SIMD_LOAD_LANE_COMMON(vec, register, lane, width)  \
     do {                                                   \
-        uint32 offset, addr;                               \
-        offset = read_uint32(frame_ip);                    \
-        V128 vec = POP_V128();                             \
-        int32 base = POP_I32();                            \
-        offset += base;                                    \
-        int lane = *frame_ip++;                            \
         addr = GET_OPERAND(uint32, I32, 0);                \
         addr_ret = GET_OFFSET();                           \
         CHECK_MEMORY_OVERFLOW(width / 8);                  \
@@ -6409,6 +6403,17 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
             vec.register[lane] = *(uint##width *)(maddr);  \
         }                                                  \
         PUT_V128_TO_ADDR(frame_lp + addr_ret, vec);        \
+    } while (0)
+
+#define SIMD_LOAD_LANE_OP(register, width)                 \
+    do {                                                   \
+        uint32 offset, addr;                               \
+        offset = read_uint32(frame_ip);                    \
+        V128 vec = POP_V128();                             \
+        int32 base = POP_I32();                            \
+        offset += base;                                    \
+        int lane = *frame_ip++;                            \
+        SIMD_LOAD_LANE_COMMON(vec, register, lane, width); \
     } while (0)
 
                     case SIMD_v128_load8_lane:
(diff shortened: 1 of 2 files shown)


core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -6435,10 +6435,33 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+        uint32 offset, addr;                      \
+        offset = read_uint32(frame_ip);           \
+        int32 base = POP_I32();                   \
+        offset += base;                           \
+        addr = GET_OPERAND(uint32, I32, 0);       \
+        addr_ret = GET_OFFSET();                  \
+        CHECK_MEMORY_OVERFLOW(width / 8);         \
+        V128 v;                                   \
+        memset(&v, 0, sizeof(v));                 \
+        v.i32x4[0] = *(uint32_t *)(maddr);        \
```
Zzzabiyaka, line 6453: todo: fix for 64 bit

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -6435,10 +6435,38 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
+    do {                                           \
+        uint32 offset, addr;                       \
+        offset = read_uint32(frame_ip);            \
+        int32 base = POP_I32();                    \
+        offset += base;                            \
+        addr = GET_OPERAND(uint32, I32, 0);        \
+        addr_ret = GET_OFFSET();                   \
+        CHECK_MEMORY_OVERFLOW(width / 8);          \
+        V128 v;                                    \
+        memset(&v, 0, sizeof(v));                  \
```
Zzzabiyaka, line 6452: can probably be optimized to `memset(&v, width/8, sizeof(v) - width/8)`

loganek: this seems to be more concise:
```suggestion
        V128 v = {0};
```

Zzzabiyaka: This is probably more concise but doesn't allow to avoid setting first 4/8 bytes which means it does up to 50% more work to set the memory,
So I propose to go with memset still 

Let me know what you think

loganek: Not sure I understand your concerns. The entire structure will be initialized to 0, see https://stackoverflow.com/questions/11152160/initializing-a-struct-to-0

Zzzabiyaka: Yeah but I don't need first 4/8 bytes to be 0 initialised. I am going to overwrite them later.

In that case I don't want to zero them first for no reason and I only want to memset to 0 the part which will stay (remaining 12 or 8 bytes respectively)

loganek: The existing code resets all the bytes, doesn't skip any of that. Also, the improvement you propose looks to me more like a micro pre-mature optimization; unless you have any solid benchmark results that prove the difference, I don't think the code complexity is worth it.

Zzzabiyaka:
Hi @wenyongh @lum1n0us, could you take a look at this code please as @loganek should now be on Christmas holiday and the PR is quite simple in my opinion?

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -6435,10 +6435,37 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                     case SIMD_v128_store16_lane:
                     case SIMD_v128_store32_lane:
                     case SIMD_v128_store64_lane:
+                    {
+                        wasm_set_exception(module, "unsupported SIMD opcode");
+                        break;
+                    }
+#define SIMD_LOAD_ZERO_OP(width)                   \
```
lum1n0us, line 6442: It appears there is some code shared with `SIMD_LOAD_LANE_OP`(like after `int32 base...`). Is there a way we can minimize code duplication?

Zzzabiyaka: yeah that's a great idea actually 

what we can do is call `load_lane (0, width)` to load the 0 lane and then nullify other lanes

Zzzabiyaka:
@lum1n0us I refactored a bit, I think refactoring further will just bring a lot of complexity

lum1n0us:
> @lum1n0us I refactored a bit, I think refactoring further will just bring a lot of complexity

I'm quite fond of this version. It uncovers the logical connection between `load_lane` and `load_zero`.

lum1n0us (approved):
LGTM

Zzzabiyaka:
looks like the CI failure is unrelated (it's also in classic-interp while the PR only touches the fast one) so I think it should be merge-ready

loganek (approved):
LGTM, thanks

loganek:
The CI is green now after I re-ran it.

PR #3944, opened 2024-12-03
always include bh_mkstemp and bh_system

sjamesr:
Internally, Google builds WAMR with blaze rules, and these rules currently assume anything in shared/common libraries are shared by all configurations of WAMR. Fixing this substantially increases our build complexity.

If it doesn't introduce an unacceptable additional cost, it would be good if we could just include these symbols unconditionally, or perhaps move them closer to where they are used.

### DIFF
--- a/core/shared/utils/bh_common.c
+++ b/core/shared/utils/bh_common.c
@@ -166,7 +166,6 @@ wa_strdup(const char *s)
     return s1;
 }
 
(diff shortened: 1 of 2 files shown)


sjamesr:
I think the cost is around 300 bytes

sjamesr:
Breaks SGX, need to rethink this

PR #3955, opened 2024-12-11
set alignment 4 when loading multi return value

TianlongLiang:
If the return value is v128 type, it will set alignment 16 by default, which can cause potential unaligned memory access for agrv[]. So set alignment as 4 to accommodate all types (i32, i64, v128) of return values.

### DIFF
--- a/core/iwasm/compilation/aot_emit_function.c
+++ b/core/iwasm/compilation/aot_emit_function.c
@@ -2069,6 +2069,7 @@ call_aot_call_indirect_func(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
             aot_set_last_error("llvm build load failed.");
             return false;
         }
+        LLVMSetAlignment(value_rets[i], 4);
         cell_num += wasm_value_type_cell_num_internal(wasm_ret_types[i],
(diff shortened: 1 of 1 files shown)


core/iwasm/compilation/aot_emit_function.c
```diff
@@ -1832,6 +1832,7 @@ aot_compile_op_call(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
                 aot_set_last_error("llvm build load failed.");
                 goto fail;
             }
+            LLVMSetAlignment(ext_ret, 4);
```
wenyongh, line 1835: It seems there is similar issue in compiling opcode call_indirect and call_ref:

https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/compilation/aot_emit_function.c#L2696-L2698

https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/compilation/aot_emit_function.c#L3202-L3204

wenyongh (approved):

lum1n0us (approved):
LGTM

PR #3956, opened 2024-12-11
CMakeLists.txt: Do not require C++ by default

midokura-xavi92:
By default, the project() CMake command defaults to C and C++. [1] Therefore, CMake might perform tests for both C and C++ compilers as part of the configuration phase.

However, this has the consequence of the configuration phase to fail if the system does not have a C++ toolchain installed, even if C++ is not really used by the top-level project under the default settings.

Some configurations might still require a C++ toolchain, so `enable_language` is selectively called under such circumstances.

[1]: https://cmake.org/cmake/help/latest/command/project.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,7 @@ if(ESP_PLATFORM)
   return()
 endif()
 
-project (iwasm)
+project (iwasm LANGUAGES C)
 
 set (CMAKE_VERBOSE_MAKEFILE OFF)
 
--- a/core/iwasm/compilation/iwasm_compl.cmake
+++ b/core/iwasm/compilation/iwasm_compl.cmake
@@ -1,6 +1,7 @@
 set (IWASM_COMPL_DIR ${CMAKE_CURRENT_LIST_DIR})
 
 include_directories(${IWASM_COMPL_DIR})
+enable_language(CXX)
 
 if (WAMR_BUILD_DEBUG_AOT EQUAL 1)
     file (GLOB_RECURSE source_all
--- a/core/iwasm/fast-jit/iwasm_fast_jit.cmake
+++ b/core/iwasm/fast-jit/iwasm_fast_jit.cmake
@@ -9,6 +9,7 @@ if (WAMR_BUILD_FAST_JIT_DUMP EQUAL 1)
 endif ()
 
 include_directories (${IWASM_FAST_JIT_DIR})
+enable_language(CXX)
 
 if (WAMR_BUILD_TARGET STREQUAL "X86_64" OR WAMR_BUILD_TARGET STREQUAL "AMD_64")
     include(FetchContent)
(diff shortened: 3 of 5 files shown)


yamt:
how have you tested this? on which platform?
our CI coverage here is weak. https://github.com/bytecodealliance/wasm-micro-runtime/issues/2686

yamt (approved):
looks reasonable to me.

i bulid-tested this on macos with and without fast-jit.
(with https://github.com/bytecodealliance/wasm-micro-runtime/pull/3968)

lum1n0us:
Using the same reasoning, I suppose *core/iwasm/compilation/iwasm_compl.cmake* requires a similar modification.

yamt:
> Using the same reasoning, I suppose _core/iwasm/compilation/iwasm_compl.cmake_ requires a similar modification.

good point.

@midokura-xavi92 please test WAMR_BUILD_JIT=1 as well.

yamt:
> > Using the same reasoning, I suppose _core/iwasm/compilation/iwasm_compl.cmake_ requires a similar modification.
> 
> Not really: all files that `include(${IWASM_DIR}/compilation/iwasm_compl.cmake`) always do so from their own `CMakeLists.txt` (e.g.: [`wamr-compiler/CMakeLists.txt`](https://github.com/bytecodealliance/wasm-micro-runtime/blob/f8f37c8ebbcd618f28d7373c21d76bfecd44a8e3/wamr-compiler/CMakeLists.txt#L17-L20)), and also from their own `project`, whose languages always default to `C CXX`, according to [the documentation](https://cmake.org/cmake/help/latest/command/project.html):

it's included via runtime_lib.cmake.

top-level cmake -> runtime_lib.cmake -> iwasm_compl.cmake

> > @midokura-xavi92 please test WAMR_BUILD_JIT=1 as well.
> 
> That would require me to build the `wamr-compiler` from source:

you don't need to build wamr-compiler. just llvm.
if you are lucky, you might be able to use the existing llvm binary by setting LLVM_DIR.
 
> ```
> $ cmake -B build -DCMAKE_EXPORT_COMPILE_COMMANDS=1 -DCMAKE_BUILD_TYPE=Debug
> CMake Error at build-scripts/config_common.cmake:106 (message):
>   Cannot find LLVM dir: /home/xavier/wasm-micro-runtime/core/deps/llvm/build
> Call Stack (most recent call first):
>   build-scripts/runtime_lib.cmake:169 (include)
>   CMakeLists.txt:122 (include)
> ```
> 
> I am not against it, but building LLVM and friends from source will take some time. I will let you know once this is done, but in the meantime please take the rationale above into consideration.

as our CI in this area is weak, manual testing is a must.

alternatively you can improve our CI coverage. it's probably a better idea anyway. :-)

lum1n0us (approved):
LGTM

Issue #3687, opened 2024-08-05
CI doesn't cover shared library


yamt:
cf. https://github.com/bytecodealliance/wasm-micro-runtime/pull/3655#issuecomment-2268359593

PR #3655, opened 2024-07-22
Set compile symbol visibility to hidden in cmake

bnason-nf:

PR #3957, opened 2024-12-11
wasm_export.h: Use "default" visibility

midokura-xavi92:
Since the top-level `CMakelists.txt` is appending `-fvisibility=hidden` to the compile options, no public symbols are exported by default. This forbids users from linking against the shared library.

Using `gcc`/`clang` attributes [1], it is possible to override the definition for `WASM_RUNTIME_API_EXTERN` so that only required symbols are correctly exported.

- Before the patch:

```
$ nm build/libiwasm.so | grep -e 'wasm_runtime_get_version'
00000000000419e9 t wasm_runtime_get_version
```

- After the patch:

```
$ nm build/libiwasm.so | grep -e 'wasm_runtime_get_version'
0000000000043c69 T wasm_runtime_get_version
```

[1]: https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#Common-Function-Attributes

### DIFF
--- a/core/iwasm/include/wasm_export.h
+++ b/core/iwasm/include/wasm_export.h
@@ -23,6 +23,8 @@
 #else
 #define WASM_RUNTIME_API_EXTERN __declspec(dllimport)
 #endif
+#elif defined(__GNUC__) || defined(__clang__)
+#define WASM_RUNTIME_API_EXTERN __attribute__((visibility("default")))
 #else
 #define WASM_RUNTIME_API_EXTERN
 #endif


lum1n0us:
Please review the changes at https://github.com/bytecodealliance/wasm-micro-runtime/pull/3655. I trust you'll find the information useful.

lum1n0us:
@bnason-nf @yamt Please review this. Ensure that it does not conflict with your previous plans.

lum1n0us (approved):
LGTM

bnason-nf:
Thanks for the heads up @lum1n0us. I think this should be ok, but I would need to test more to be sure. I think it's fine to merge and we can revisit later if it causes any problems for me.

yamt:
> @bnason-nf @yamt Please review this. Ensure that it does not conflict with your previous plans.

honestly speaking, i don't remember the previous discussion well.
this patch itself looks reasonable to me.

@midokura-xavi92 if you care these stuff, you might want to improve the CI coverage.
iirc, our CI coverage on shared library and relevant things is pretty low.
cf. https://github.com/bytecodealliance/wasm-micro-runtime/issues/3687

yamt (approved):

PR #3958, opened 2024-12-15
[fuzzing] Enable instantiation

lum1n0us:
- use a large seed to create larger binaries.
- enable instantiation.

### DIFF
--- a/tests/fuzz/wasm-mutator-fuzz/wasm_mutator_fuzz.cc
+++ b/tests/fuzz/wasm-mutator-fuzz/wasm_mutator_fuzz.cc
@@ -13,31 +13,39 @@
 
 using namespace std;
 
-extern "C" WASMModuleCommon *
-wasm_runtime_load(uint8 *buf, uint32 size, char *error_buf,
-                  uint32 error_buf_size);
-
-extern "C" WASMModuleInstanceCommon *
-wasm_runtime_instantiate(WASMModuleCommon *module, uint32 stack_size,
-                         uint32 heap_size, char *error_buf,
-                         uint32 error_buf_size);
-
 extern "C" int
 LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size)
 {
     /* libfuzzer don't allow us to modify the given Data, so we copy the data
      * here */
     std::vector<uint8_t> myData(Data, Data + Size);
+
     /* init runtime environment */
     wasm_runtime_init();
-    wasm_module_t module =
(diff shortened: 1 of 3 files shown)


tests/fuzz/wasm-mutator-fuzz/wasm_mutator_fuzz.cc
```diff
@@ -13,31 +13,41 @@
...
+        wasm_runtime_load((uint8_t *)myData.data(), Size, error_buf, 120);
+    if (!module) {
+        std::cout << "[LOADING] " << error_buf << std::endl;
+        wasm_runtime_destroy();
+        /* return SUCCESS because the failure has been handled */
+        return 0;
+    }
+
+    wasm_module_inst_t inst = wasm_runtime_instantiate(
+        module, 8 * 1024 * 1024, 16 * 1024 * 1024, error_buf, 120);
```
wenyongh, line 37: The input heap size and stack size are fixed, maybe we can also let the fuzz tool randomly generate them in the future. Now it looks good.

wenyongh (approved):

tests/fuzz/wasm-mutator-fuzz/wasm_mutator_fuzz.cc
```diff
@@ -13,31 +13,41 @@
...
+        std::cout << "[INSTANTIATE] " << error_buf << std::endl;
         wasm_runtime_unload(module);
+        wasm_runtime_destroy();
+        /* return SUCCESS because the failure has been handled */
+        return 0;
     }
-    /* destroy runtime environment */
-    wasm_runtime_destroy();
 
+    std::cout << "PASS" << std::endl;
```
xujuntwt95329, line 46: Should we also execute the instance? Does wasm-mutator tell us which funcs are exported? Or does it export a fixed function to be invoked?

lum1n0us: - Use `wasm-tools smith` to create ~10 WebAssembly modules for each set of features, including both valid and invalid ones.
- Employ these as the seed corpus.
- Allow libfuzzer to perform its function.

xujuntwt95329: Got it. We can firstly run this to check the coverage and further optimize the pipeline according to the results

xujuntwt95329 (approved):
LGTM

PR #3959, opened 2024-12-15
[fuzzing] execute every exported function

lum1n0us:
after https://github.com/bytecodealliance/wasm-micro-runtime/pull/3984

### DIFF
--- a/tests/fuzz/wasm-mutator-fuzz/wasm_mutator_fuzz.cc
+++ b/tests/fuzz/wasm-mutator-fuzz/wasm_mutator_fuzz.cc
@@ -150,7 +150,7 @@ execute_export_functions(wasm_module_t module, wasm_module_inst_t inst)
                 }
             }
 
-            std::cout << ")";
+            std::cout << ")" << std::endl;
         }
 
         bool ret = wasm_runtime_call_wasm_a(exec_env, func, result_count,
--- a/tests/fuzz/wasm-mutator-fuzz/wasm_mutator_fuzz.cc
+++ b/tests/fuzz/wasm-mutator-fuzz/wasm_mutator_fuzz.cc
@@ -13,70 +13,73 @@
 
 using namespace std;
 
-/*
- * there is a unsigned integer overflow in
- * /usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/random.tcc:416
- *
- * use srand() and rand() instead
- */
-
-static int32_t
-generate_random_int32()
+static bool
+is_supported_val_kind(wasm_valkind_t kind)
 {
-    return (int32_t)rand();
-}
-
-static int64_t
-generate_random_int64()
-{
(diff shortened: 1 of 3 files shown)


yamt:
for a similar fuzzier (https://github.com/yamt/toywasm/tree/master/examples/fuzz)
i ended up with having a small limit for the number of functions to execute
because a module can have a ton of exported functions.
i'm not sure if the same concern applies to this fuzzier. just FYI.

lum1n0us:
> i ended up with having a small limit for the number of functions to execute because a module can have a ton of exported functions.

May I ask the negative consequences of executing too many functions?

yamt:
> > i ended up with having a small limit for the number of functions to execute because a module can have a ton of exported functions.
> 
> May I ask the negative consequences of executing too many functions?

it takes very long. practically ~forever.

lum1n0us:
Just FYI:

If the issue is due to an infinite loop, the `--fuel` option can be utilized to prevent the endless execution scenario.

yamt:
> Just FYI:
> 
> If the issue is due to an infinite loop, the `--fuel` option can be utilized to prevent the endless execution scenario.

does the fuel thing "accumulate" among function calls?
otherwise it isn't a solution.

lum1n0us:
>   -f, --fuel <FUEL>  This is roughly the number of loop iterations and function calls that will be executed before a trap is raised to prevent infinite loops.

IIUC, yes.

tests/fuzz/wasm-mutator-fuzz/wasm_mutator_fuzz.cc
```diff
@@ -13,6 +13,149 @@
...
+    return kind == WASM_I32 || kind == WASM_I64 || kind == WASM_F32
+           || kind == WASM_F64 || kind == WASM_EXTERNREF
+           || kind == WASM_FUNCREF;
+}
+
+static wasm_val_t
+pre_defined_val(wasm_valkind_t kind)
+{
+    if (kind == WASM_I32) {
+        return wasm_val_t{ .kind = WASM_I32, .of = { .i32 = 2025 } };
```
xujuntwt95329, line 28: Would it be better to generate random value to cover more value range?

lum1n0us: 😄 It makes it easier to replicate the issue that led to the failure. In a fuzzing issue report, you typically only get the error information generated by XSAN, with no accompanying logs. Using random values as function parameters can make it challenging to identify the exact combination that triggered the problem.

🤔 Additionally, I'm looking for a way to run opcodes with random operands. It appears to be a variant of the `LLVMFuzzerTestOneInput()` function. Perhaps I should create a new test `LLVMFuzzerTestOneInput()` to execute each opcode individually, allowing the fuzzer to supply random operands.

TianlongLiang (approved):
LGTM

yamt (approved):

PR #3973, opened 2024-12-19
top-level cmake: link llvm libraries to our shared library

yamt:
This fixes link errors seen on my environment.
(macOS 15.2, x86-64, Xcode 16.2)

Tested as:
```
mkdir b
cd b
cmake -D WAMR_BUILD_JIT=1 -D LLVM_DIR=/usr/local/opt/llvm@19/lib/cmake/llvm ..
make
```


lum1n0us (approved):
Using `INTERFACE` should be fine. However, using `PUBLIC` doesn't make a significant difference. LGTM.

wenyongh (approved):

yamt:
> Using `INTERFACE` should be fine. However, using `PUBLIC` doesn't make a significant difference. LGTM.

my understanding is that INTERFACE doesn't link the specified libraries when building the target.
when building a shared library, we need to link these libraries.

Issue #3977, opened 2024-12-19
WASI-SDK 25 call_indrect doesn't work on Darwin


woodsmc:
Thanks for filing a bug or defect report! Please fill out the TODOs below.

### Subject of the issue

WASI SDK 25 generates a call indirect which uses a ref type format. It doesn't parse correctly on the default build of iwasm for Darwin. Executing the generated wasm with the default iwasm for Darwin results in : ``WASM module load failed: zero byte expected` 

### Test case

Upload the related wasm file, wast file or the source files if you can.
```c
#define WASM_EXPORT(A) __attribute__((export_name(A)))

typedef int(*TFunction)(int);

int function(int x) {
	x *= 25;
	return x;
}

int work(int x, TFunction ptr) {
	return (*ptr)(x);
}

WASM_EXPORT("test") int test(int x) {
	return work(x, &function);
}
```
Compile this with wasi-sdk-25 and attempt to execute with iwasm results in an error message - ``WASM module load failed: zero byte expected` 

### Your environment

* Host OS: MACOS
* WAMR version, platform, cpu architecture, running mode, etc.

### Steps to reproduce

Tell us how to reproduce this bug or defect.

Build the code above with wasi-sdk-25, attempt to run with default iwasm on darwin. Receive error message as below

```
iwasm -f test ./indirect.wasm 25
```
And the error message ``WASM module load failed: zero byte expected` is displayed.

### Expected behavior

Tell us what should happen: The function should execute

### Actual behavior

Tell us what happens instead: WAMR displays error message instead.

### Extra Info

Anything else you'd like to add?

woodsmc:
I think turning on reference types resolves this issue. This PR changes the default CMake file for Darwin to do this. https://github.com/bytecodealliance/wasm-micro-runtime/pull/3978

It appeared to be turned on by default on the linux build at the moment, just not on Darwin.

loganek:
Hi @woodsmc I think it's reasonable to enable ref types by default. Having said that, I think we could do a better job in terms of error reporting and have clear message when iwasm was compiled with ref types disabled but the binary was compiled with ref types enabled. I can make the change once I'm back from holiday (January next year) so let's keep the issue open until the error handling is improved so I don't forget about it :)

woodsmc:
[...]

PR #3978, opened 2024-12-19
Add reference type support by default for darwin to support WASI-SDK-25

woodsmc:
This resolved the issue [WASI-SDK 25 `call_indrect` doesn't work on Darwin](https://github.com/bytecodealliance/wasm-micro-runtime/issues/3977), by turning on reference types by default for iwasm on Darwin.

The error message is triggered by the way call indirect is rendered by WASI-SDK25, which has changed from WASI-SDK24. Turning on the reference checks allows iwasm on Darwin use the WASI-SDK generated code directly.

*Updated with additional information*

### DIFF
--- a/product-mini/platforms/darwin/CMakeLists.txt
+++ b/product-mini/platforms/darwin/CMakeLists.txt
@@ -91,6 +91,11 @@ if (NOT DEFINED WAMR_BUILD_SIMD)
   set (WAMR_BUILD_SIMD 1)
 endif ()
 
+if (NOT DEFINED WAMR_BUILD_REF_TYPES)
+  # Enable reference types by default
+  set (WAMR_BUILD_REF_TYPES 1)
+endif ()
+
 if (NOT DEFINED WAMR_BUILD_DEBUG_INTERP)
   # Disable Debug feature by default
   set (WAMR_BUILD_DEBUG_INTERP 0)


loganek (approved):
LGTM

yamt (commented):
do you know any reasons this should be per-platform?
otherwise i guess it's better to put this in config_common.cmake
as we do for WAMR_BUILD_BULK_MEMORY.

lum1n0us:
> do you know any reasons this should be per-platform? otherwise i guess it's better to put this in config_common.cmake as we do for WAMR_BUILD_BULK_MEMORY.

completely agree with that. @wenyongh @loganek @tianlong @no1wudi what's your opinion?

lum1n0us (approved):
LGTM

wenyongh:
> > do you know any reasons this should be per-platform? otherwise i guess it's better to put this in config_common.cmake as we do for WAMR_BUILD_BULK_MEMORY.
> 
> completely agree with that. @wenyongh @loganek @tianlong @no1wudi what's your opinion?

Ref-types feature is a relatively complex/big feature and will increase lots of footprint if it is enabled, and it seems that it is seldom used by some platforms, I guess it is better to let the platform's CMakeListx.txt decide how to set it. There were similar changes in PR [#3894](https://github.com/bytecodealliance/wasm-micro-runtime/pull/3894/files). If we decide to enable it by default in `config_common.cmake`, had better revert #3894 also.

wenyongh (approved):

woodsmc:
I think it would make sense to enable for the desktop development platforms. For new users coming to WAMR seeing the output of RUST or the WASI-SDK run is important for evaluating the platform. For experienced users optimizing for size is key. For the RTOS based platforms I'd guess that these should be built without Ref-types.

Mentioning the options selected during the build process - and the options not selected would be really great improvement.

Additionally we should (I haven't yet) make sure that we can turn off ref-types in the WASI-SDK to support those platforms that don't want this feature.

loganek:
> Mentioning the options selected during the build process - and the options not selected would be really great improvement.

AFAIK some of the options are printed when generating CMake file, but some of them are not. Also, it'd be good to print the enabled features when user calls `iwasm --help` or `iwasm -v`. Would you mind open an issue so we can track the improvement?

lum1n0us:
> > Mentioning the options selected during the build process - and the options not selected would be really great improvement.
> 
> AFAIK some of the options are printed when generating CMake file, but some of them are not. Also, it'd be good to print the enabled features when user calls `iwasm --help` or `iwasm -v`. Would you mind open an issue so we can track the improvement?

I am creating [a new document about which wasm proposals are on-by-default and which are off-by-default](https://github.com/bytecodealliance/wasm-micro-runtime/blob/ee647abfece921f541c0b66f16de4ee76a20cf91/doc/stability_wasm_proposals.md). Are you suggesting we should output something like that during cmake configuration and execution by `iwasm` ?

lum1n0us:
> I think it would make sense to enable for the desktop development platforms. For new users coming to WAMR seeing the output of RUST or the WASI-SDK run is important for evaluating the platform. For experienced users optimizing for size is key. For the RTOS based platforms I'd guess that these should be built without Ref-types.

In that case, it should include *linux*, *macos* and *windows*, right?

lum1n0us:
I am going to merge this PR first and leave it open to continue our discussion.

PR #3983, opened 2024-12-24
Optimize memory initialization handling in AOT loader

no1wudi:
Save memory if the file buffer is always exist before exit.

### DIFF
--- a/core/iwasm/aot/aot_loader.c
+++ b/core/iwasm/aot/aot_loader.c
@@ -967,13 +967,29 @@ destroy_import_memories(AOTImportMemory *import_memories)
     wasm_runtime_free(import_memories);
 }
 
+/**
+ * Free memory initialization data segments.
+ *
+ * @param module the AOT module containing the data
+ * @param data_list array of memory initialization data segments to free
+ * @param count number of segments in the data_list array
+ */
+
 static void
-destroy_mem_init_data_list(AOTMemInitData **data_list, uint32 count)
+destroy_mem_init_data_list(AOTModule *module, AOTMemInitData **data_list,
+                           uint32 count)
 {
     uint32 i;
+    /* Free each memory initialization data segment */
     for (i = 0; i < count; i++)
-        if (data_list[i])
+        if (data_list[i]) {
+            /* If the module owns the binary data, free the bytes buffer */
+            if (module->is_binary_freeable && data_list[i]->bytes)
+                wasm_runtime_free(data_list[i]->bytes);
+            /* Free the data segment structure itself */
             wasm_runtime_free(data_list[i]);
+        }
+    /* Free the array of data segment pointers */
     wasm_runtime_free(data_list);
 }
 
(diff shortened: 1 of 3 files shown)


wenyongh:
This is to make the memory init data available and wasm/aot module can be used again when `module->is_binary_freeable` is true and the wasm/aot binary is freed after loading?

no1wudi:
Yes, the AOT binary should be available before runtime exit

core/iwasm/aot/aot_loader.c
```diff
@@ -1026,8 +1030,21 @@ load_mem_init_data_list(const uint8 **p_buf, const uint8 *buf_end,
         data_list[i]->offset.init_expr_type = init_value.init_expr_type;
         data_list[i]->offset.u = init_value.u;
         data_list[i]->byte_count = byte_count;
-        read_byte_array(buf, buf_end, data_list[i]->bytes,
-                        data_list[i]->byte_count);
+        data_list[i]->bytes = NULL;
+        if (module->is_binary_freeable) {
```
wenyongh, line 1034: There is similar handling in wasm loader, but it clones memory init data only when both `is_load_from_file_buf` and `module->is_binary_freeable` are true:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_loader.c#L5791
note `is_load_from_file_buf` is passed to this fuction's caller `load_init_data_section`.

Do you think here we had better handle `data_lists[i]->bytes` like wasm_loader.c?

no1wudi: Yes, it's better to handle them by same way

no1wudi: But, pass the `is_load_from_file_buf ` will change many places, I think `module->is_binary_freeable` is enough for many use case

wenyongh (approved):
LGTM

core/iwasm/compilation/aot.c
```diff
@@ -60,27 +63,39 @@ aot_create_mem_init_data_list(const WASMModule *module)
...
         data_list[i]->is_passive = module->data_segments[i]->is_passive;
         data_list[i]->memory_index = module->data_segments[i]->memory_index;
 #endif
         data_list[i]->offset = module->data_segments[i]->base_offset;
         data_list[i]->byte_count = module->data_segments[i]->data_length;
-        memcpy(data_list[i]->bytes, module->data_segments[i]->data,
-               module->data_segments[i]->data_length);
+        data_list[i]->bytes = NULL;
+        /* Allocate memory for AOT compiler is OK, because the data segment
+         * is small and the host memory is enough */
```
lum1n0us, line 82: If that's the case, why not maintain the original? But still, the modification was made.

TianlongLiang (approved):
LGTM

core/iwasm/compilation/aot.c
lum1n0us: Are you suggesting that malloc()/free() should better match the 
implied convention of the data structure? If using a variable-length array as we did previously, we should allocate one block of memory. If using a pointer as we do now, we should allocate memory twice.

no1wudi: Using variable-length arrays is not an issue; it's just that in the current scenario, most of the time we don't even need to allocate memory a second time. We can simply point the pointer to the data buffer.

lum1n0us: I completely agree the intent of this PR. 

My uncertainty lies with the changes made in aot.c. As you mentioned, the AOT compiler doesn't need to *set the pointer to the data buffer*. From my perspective, this suggests that there is no need to alter aot.c, as it does not align with the same intention as aot_loader.c. But if this is about upholding an implied convention of programming, then I believe it is acceptable.

no1wudi: The key point is the structure that `uint8* bytes; // NEW` at the end can not be used with variable-length, it's totally different with `uint8 bytes[1]`, so I must modifty the `aot.c`.

lum1n0us: Oh, I see.

lum1n0us (approved):
LGTM

PR #3996, opened 2024-12-31
heap_base and data_end have to be local globals

lum1n0us:


TianlongLiang (approved):
LGTM

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -5924,6 +5924,14 @@ load_from_sections(WASMModule *module, WASMSection *sections,
     for (i = 0; i < module->export_count; i++, export ++) {
         if (export->kind == EXPORT_KIND_GLOBAL) {
             if (!strcmp(export->name, "__heap_base")) {
+                /* has to be local. can't be an import one */
+                if (module->import_global_count
+                    && export->index <= module->import_global_count) {
```
wenyongh, line 5929: Should change `<=` to `<`?

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -5938,6 +5946,14 @@ load_from_sections(WASMModule *module, WASMSection *sections,
                 }
             }
             else if (!strcmp(export->name, "__data_end")) {
+                /* has to be local. can't be an import one */
+                if (module->import_global_count
+                    && export->index <= module->import_global_count) {
```
wenyongh, line 5951: ditto

core/iwasm/interpreter/wasm_loader.c
wenyongh: And could you also apply changes to wasm_mini_loader.c?

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -5924,6 +5924,14 @@ load_from_sections(WASMModule *module, WASMSection *sections,
     for (i = 0; i < module->export_count; i++, export ++) {
         if (export->kind == EXPORT_KIND_GLOBAL) {
             if (!strcmp(export->name, "__heap_base")) {
+                /* has to be local. can't be an import one */
+                if (module->import_global_count
+                    && export->index <= module->import_global_count) {
+                    set_error_buf(error_buf, error_buf_size,
+                                  "invalid __heap_base global index");
+                    return false;
```
wenyongh, line 5932: If a wasm module import a global named `__heap_base`, seems it should be allowed and we had better not block loading it? How about just ignoring it instead of returning false? e.g., change `if (!strcmp(export->name, "__heap_base"))` to `if (!strcmp(export->name, "__heap_base") && export->index >= module->import_global_count)`

wenyongh (approved):
LGTM

PR #3998, opened 2025-01-02
Add an example of how to embed WAMR in Zephyr user mode 

TianlongLiang:
Rough ideas of how to use WAMR in Zephyr user mode 

- Link WAMR as a separate library in the Zephyr application
- In the Zephyr application, a memory buffer will be created and granted to WAMR runtime for its global variables and all heap allocation(for both runtime and wasm app)


product-mini/platforms/zephyr/user-mode/lib-wamr-zephyr/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+if (NOT DEFINED WAMR_BUILD_LIBC_BUILTIN)
+  # Enable libc builtin support by default
+  set (WAMR_BUILD_LIBC_BUILTIN 1)
+endif ()
+
+if (NOT DEFINED WAMR_BUILD_LIBC_WASI)
+  # Disable libc wasi support by default
+  set (WAMR_BUILD_LIBC_WASI 0)
+endif ()
+
```
lum1n0us, line 33: In my understand, other combinations(like `WAMR_BUILD_LIBC_BUILTIN=1`+`WAMR_BUILD_LIBC_WASI=1`) will cause troubles.  If so, we might not allow users to modify both.

product-mini/platforms/zephyr/user-mode/lib-wamr-zephyr/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+if (WAMR_BUILD_TARGET STREQUAL "RISCV64_LP64" OR WAMR_BUILD_TARGET STREQUAL "RISCV32_ILP32")
+  set (WAMR_BUILD_FAST_INTERP 1)
+endif ()
+
+# Override the global heap usage
+if (NOT DEFINED WAMR_BUILD_GLOBAL_HEAP_POOL)
+  set (WAMR_BUILD_GLOBAL_HEAP_POOL 1)
+endif ()
+
```
lum1n0us, line 42: same story like above. we should not allow users to modify `WAMR_BUILD_GLOBAL_HEAP_POOL`

product-mini/platforms/zephyr/user-mode/src/main.c
```diff
@@ -0,0 +1,62 @@
...
+iwasm_user_mode(void)
+{
+    struct k_mem_partition *wamr_domain_parts[] = { &wamr_partition,
+                                                    &z_libc_partition };
+
+    printk("wamr_partition start addr: %p, size: %zu\n", wamr_partition.start,
+           wamr_partition.size);
+
+    /* Initialize the memory domain with single WAMR partition */
+    k_mem_domain_init(&wamr_domain, 2, wamr_domain_parts);
```
lum1n0us, line 35: need to process the return value of `k_mem_domain_init()`

product-mini/platforms/zephyr/user-mode/src/main.c
```diff
@@ -0,0 +1,62 @@
...
+
+    k_tid_t tid =
+        k_thread_create(&iwasm_user_mode_thread, iwasm_user_mode_thread_stack,
+                        MAIN_THREAD_STACK_SIZE, iwasm_main, NULL, NULL, NULL,
+                        MAIN_THREAD_PRIORITY, K_USER, K_FOREVER);
+
+    /* Grant WAMR memory domain access to user mode thread */
+    k_mem_domain_add_thread(&wamr_domain, tid);
```
lum1n0us, line 43: need to process the return value of `k_mem_domain_add_thread()`

product-mini/platforms/zephyr/user-mode/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+```C
+K_APPMEM_PARTITION_DEFINE(wamr_partition);
+```
+
+When creating a Zephyr thread, set the thread option to `K_USER` and the timeout to `K_FOREVER`. This can ensure that the `wamr_partition` is granted access to the thread before starting it with `k_thread_start`.
+
+### Advantage of using WAMR runtime in Zephyr user mode thread
+
+In a user-mode Zephyr thread, the application can only access a restricted partition of memory it granted to. It creates a sandbox for the WAMR runtime to run in, and the WAMR runtime can only access that memory space, meaning that all global variables in the WAMR runtime and both runtime and wasm app heap memory will allocated from it. In this way, an extra layer of security is added to the wasm application on top of the wasm sandbox provided by WAMR.
```
wenyongh, line 46: `will allocated from it` -> `will be allocated from it`?

product-mini/platforms/zephyr/simple/CMakeLists.txt
```diff
@@ -31,10 +31,8 @@ if (NOT DEFINED WAMR_BUILD_LIBC_BUILTIN)
   set (WAMR_BUILD_LIBC_BUILTIN 1)
 endif ()
 
-if (NOT DEFINED WAMR_BUILD_LIBC_WASI)
-  # Disable libc wasi support by default
-  set (WAMR_BUILD_LIBC_WASI 0)
-endif ()
+# Disable libc wasi support on Zephyr
+set (WAMR_BUILD_LIBC_WASI 0)
```
wenyongh, line 35: The orginal lines already disable libc-wasi by default, must we use hardcode here? Note that zephyr file/socket support is under development in branch `dev/zephyr_file_socket`.

product-mini/platforms/zephyr/user-mode/lib-wamr-zephyr/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+  set (WAMR_BUILD_AOT 1)
+endif ()
+
+if (NOT DEFINED WAMR_BUILD_LIBC_BUILTIN)
+  # Enable libc builtin support by default
+  set (WAMR_BUILD_LIBC_BUILTIN 1)
+endif ()
+
+# Disable libc wasi support on Zephyr
+set (WAMR_BUILD_LIBC_WASI 0)
```
wenyongh, line 30: Same as above

product-mini/platforms/zephyr/simple/CMakeLists.txt
TianlongLiang: The libc-wasi synchronization primitive will also need some work, but I think for now maybe keeping the original config and adding more comments is preferred against hardcode.

wenyongh (approved):
LGTM

Issue #3980, opened 2024-12-22
Feature Request: Wasm Proposal Details at Compile and Runtime


woodsmc:
> > Mentioning the options selected during the build process - and the options not selected would be really great improvement.
> 
> AFAIK some of the options are printed when generating CMake file, but some of them are not. Also, it'd be good to print the enabled features when user calls `iwasm --help` or `iwasm -v`. Would you mind open an issue so we can track the improvement?

I am creating [a new document about which wasm proposals are on-by-default and which are off-by-default](https://github.com/bytecodealliance/wasm-micro-runtime/blob/ee647abfece921f541c0b66f16de4ee76a20cf91/doc/stability_wasm_proposals.md). Are you suggesting we should output something like that during cmake configuration and execution by `iwasm` ? 

Might need to add the library case

_Originally posted by @lum1n0us in https://github.com/bytecodealliance/wasm-micro-runtime/issues/3978#issuecomment-2558411602_

lum1n0us:
I have created a https://github.com/bytecodealliance/wasm-micro-runtime/pull/3989 for further discussion.

loganek:
> We could suggest to the compiler / tool chain folks that they think about including a custom section with a list of the features enabled by the compiler when the .wasm file was produced. It would be easy to remove this custom section to reduce binary size, but it would provide useful information to the user.

Not sure if that's needed. Is there any feature that's currently not detectable?

lum1n0us:
From my perspective, it's a method to inform users about the combination of features they are currently using and to confirm whether it aligns with their expectations, rather than to trace back through the build environment.

loganek:
> From my perspective, it's a method to inform users about the combination of features they are currently using and to confirm whether it aligns with their expectations, rather than to trace back through the build environment.

Right, in that case I think it makes sense to have the custom section as suggested by @woodsmc

woodsmc:
Yeah, speaking from personal experience, there has been a bunch of times where I've compiled code to .wasm, and discovered it won't execute in the host environment. I have spent hours trying to debug this, only to discover that the .wasm needed features which were not turned on in the host environment. 

[...]

PR #4000, opened 2025-01-02
Error message improvement

loganek:
Improve error message in the scenario where the runtime was built with ref types disabled but the module uses reference types feature.

See #3980


lum1n0us (approved):
LGTM

wenyongh (approved):
LGTM

PR #4003, opened 2025-01-03
[fuzzing] Use software bound-check during fuzzing

lum1n0us:


TianlongLiang (approved):
LGTM

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -1682,7 +1682,9 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                 ret_offset = prev_frame->ret_offset;
 
                 for (ret_idx = 0,
-                    off = sizeof(int16) * (func_type->result_count - 1);
+                    off = func_type->result_count
+                              ? sizeof(int16) * (func_type->result_count - 1)
+                              : 0;
```
wenyongh, line 1687: will the original code cause fuzz test failed? When `func_type->result_count` is 0, the initial value of `off` is -2, but I am not sure whether it leads to unexpected behavior, since `ret_idx < func_type->result_count` wont be met and the loop body won't be executed.

wenyongh, line 1687: will the original code cause fuzz test failed? When `func_type->result_count` is 0, the initial value of `off` is -2, but I am not sure whether it leads to unexpected behavior, since `ret_idx < func_type->result_count` won't be met and the loop body won't be executed?

lum1n0us: `off` is a `uint32`, and `off = sizeof(int16) * (func_type->result_count - 1);` leads to a `unsigned integer overflow` when `func_type->result_count` is 0.

Another potential solution is to declare the variable `off` as a signed integer rather than an unsigned integer.

wenyongh (approved):
LGTM

PR #4005, opened 2025-01-03
implement local and function calls for v128 in the fast interpreter

Zzzabiyaka:
add fast versions of local.set/get 

add additional opcodes for these operations

add v128 values to handling of function params and return values 

**why?**

These changes allowed us to run spec tests for simd as before that, functions with v128 params / return values could not be called

### DIFF
--- a/core/iwasm/interpreter/wasm_interp_fast.c
+++ b/core/iwasm/interpreter/wasm_interp_fast.c
@@ -1699,6 +1699,13 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                                         GET_OPERAND(uint64, I64, off));
                         ret_offset += 2;
                     }
+#if WASM_ENABLE_SIMDE != 0
+                    else if (ret_types[ret_idx] == VALUE_TYPE_V128) {
+                        PUT_V128_TO_ADDR(prev_frame->lp + ret_offset,
+                                         GET_OPERAND_V128(off));
+                        ret_offset += 4;
+                    }
+#endif
 #if WASM_ENABLE_GC != 0
                     else if (wasm_is_type_reftype(ret_types[ret_idx])) {
                         PUT_REF_TO_ADDR(prev_frame->lp + ret_offset,
@@ -3536,6 +3543,24 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                 HANDLE_OP_END();
             }
 
+#if WASM_ENABLE_SIMDE != 0
+            HANDLE_OP(EXT_OP_SET_LOCAL_FAST_V128)
+            HANDLE_OP(EXT_OP_TEE_LOCAL_FAST_V128)
+            {
+                /* clang-format off */
+#if WASM_CPU_SUPPORTS_UNALIGNED_ADDR_ACCESS != 0
+                local_offset = *frame_ip++;
+#else
+                local_offset = *frame_ip;
+                frame_ip += 2;
+#endif
+                /* clang-format on */
+                PUT_V128_TO_ADDR((uint32 *)(frame_lp + local_offset),
+                                 GET_OPERAND_V128(0));
+                frame_ip += 2;
+                HANDLE_OP_END();
+            }
+#endif
             HANDLE_OP(WASM_OP_GET_GLOBAL)
             {
                 global_idx = read_uint32(frame_ip);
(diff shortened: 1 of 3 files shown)


core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -3536,6 +3541,24 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                 HANDLE_OP_END();
             }
 
+#if WASM_ENABLE_SIMDE != 0
```
Zzzabiyaka, line 3544: `#if WASM_ENABLE_SIMDE != 0` this should be `#if WASM_ENABLE_SIMD != 0` but I'll fix it once we make the spec tests work and are ready to launch

Zzzabiyaka: I do think we might not need WASM_ENABLE_SIMDE flag in the first place, it probably should be set based on fast-interp + simd

core/iwasm/interpreter/wasm_opcode.h
```diff
@@ -278,6 +278,14 @@ typedef enum WASMOpcode {
     DEBUG_OP_BREAK = 0xdc, /* debug break point */
 #endif
 
+#if (WASM_ENABLE_JIT != 0                                         \
+     || (WASM_ENABLE_FAST_INTERP != 0 && WASM_ENABLE_SIMDE != 0)) \
+    && WASM_ENABLE_SIMD != 0
```
wenyongh, line 283: Not sure whether it requires WASM_ENABLE_JIT and WASM_ENABLE_SIMD? Can we change to `#if WASM_ENABLE_FAST_INTERP != 0 && WASM_ENABLE_SIMDE != 0`?

core/iwasm/interpreter/wasm_opcode.h
```diff
@@ -790,6 +798,15 @@ typedef enum WASMAtomicEXTOpcode {
 #define SET_GOTO_TABLE_SIMD_PREFIX_ELEM()
 #endif
 
+#if (WASM_ENABLE_FAST_INTERP != 0 && WASM_ENABLE_SIMDE != 0) \
+    && WASM_ENABLE_SIMD != 0
```
wenyongh, line 802: Same as above

core/iwasm/interpreter/wasm_opcode.h
```diff
@@ -1021,7 +1038,7 @@ typedef enum WASMAtomicEXTOpcode {
         SET_GOTO_TABLE_ELEM(WASM_OP_MISC_PREFIX),    /* 0xfc */ \
         SET_GOTO_TABLE_SIMD_PREFIX_ELEM()            /* 0xfd */ \
         SET_GOTO_TABLE_ELEM(WASM_OP_ATOMIC_PREFIX),  /* 0xfe */ \
-        DEF_DEBUG_BREAK_HANDLE()                                \
+        DEF_DEBUG_BREAK_HANDLE() DEF_EXT_V128_HANDLE()          \
```
wenyongh, line 1041: Had better put `DEF_EXT_V128_HANDLE()` in a single line

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -12944,10 +12944,21 @@ wasm_loader_prepare_bytecode(WASMModule *module, WASMFunction *func,
...
                             emit_byte(loader_ctx, (uint8)local_offset);
                         }
-                        else {
+                        else if (is_64bit_type(local_type)) {
                             emit_label(EXT_OP_SET_LOCAL_FAST_I64);
                             emit_byte(loader_ctx, (uint8)local_offset);
                         }
+#if WASM_ENABLE_SIMDE != 0
+                        else if (local_type == VALUE_TYPE_V128) {
+                            emit_label(EXT_OP_SET_LOCAL_FAST_V128);
```
wenyongh, line 12953: Here we only handle `case WASM_OP_SET_LOCAL`, seems we should we also handle `case WASM_OP_TEE_LOCAL`

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -1699,6 +1699,11 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                                         GET_OPERAND(uint64, I64, off));
                         ret_offset += 2;
                     }
+                    else if (ret_types[ret_idx] == VALUE_TYPE_V128) {
```
wenyongh, line 1702: Had better wrap these lines with `#if WASM_ENABLE_SIMDE != 0`?

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -6954,6 +6999,7 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                     }
                     case SIMD_i32x4_add:
                     {
+
```
wenyongh, line 7002: had better remove unused line

core/iwasm/interpreter/wasm_opcode.h
Zzzabiyaka: yeah I agree but this one is forcefully corrected by clang-format so I guess I'll have to leave it like that

Zzzabiyaka:
looks like the CI failure is due to github infra and not my code, I think we're merge-ready

core/iwasm/interpreter/wasm_opcode.h
```diff
@@ -278,6 +278,12 @@ typedef enum WASMOpcode {
     DEBUG_OP_BREAK = 0xdc, /* debug break point */
 #endif
 
+#if WASM_ENABLE_FAST_INTERP != 0 && WASM_ENABLE_SIMDE != 0
+    EXT_OP_SET_LOCAL_FAST_V128 = 0xdd,
+    EXT_OP_TEE_LOCAL_FAST_V128 = 0xde,
+    EXT_OP_COPY_STACK_TOP_V128 = 0xdf,
```
wenyongh, line 284: Seems that EXT_OP_COPY_STACK_TOP_V128 is unneeded? It isn't emitted in wasm_loader.c.

core/iwasm/interpreter/wasm_opcode.h
wenyongh: OK, got it, thanks.

Zzzabiyaka:
closing, please see #4056

PR #4007, opened 2025-01-06
Prefer to get & set exception from exec_env_tls

paul-hoang:
Exceptions are currently stored in module instance, but if most of them can be stored at the exec_env that would allow using a single module instance in parallel without locking and handle the exception separately.

Using a single module instance in parallel might not be usual, but it's possible from what I understand and tested WAMR, at least for specific use cases (e.g. no writing to globals or wasm memory, only reading from that and writing to wasm stack). Please correct me if I'm missing anything.

This PR proposes a change in how we store and retrieve exception from a given module instance:
- If the exec_env_tls is present, get/set exception from that and not touching the module instance.
- If not, keep getting/setting from the module instance. This is for exceptions that happens before an exec_env is created.

### DIFF
--- a/core/iwasm/common/wasm_exec_env.h
+++ b/core/iwasm/common/wasm_exec_env.h
@@ -33,6 +33,8 @@ typedef struct WASMJmpBuf {
 } WASMJmpBuf;
 #endif
 
+#define EXCEPTION_BUF_LEN 128
+
 /* Execution environment */
 typedef struct WASMExecEnv {
     /* Next thread's exec env of a WASM module instance. */
@@ -166,6 +168,9 @@ typedef struct WASMExecEnv {
         /* The WASM stack. */
         uint8 bottom[1];
     } wasm_stack_u;
+
+    /* The exception buffer for current execution environment. */
+    char cur_exception[EXCEPTION_BUF_LEN];
 } WASMExecEnv;
 
 #if WASM_ENABLE_MEMORY_PROFILING != 0
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -2965,6 +2965,22 @@ wasm_runtime_get_exec_env_singleton(WASMModuleInstanceCommon *module_inst_comm)
 static void
 wasm_set_exception_local(WASMModuleInstance *module_inst, const char *exception)
 {
+#ifdef OS_ENABLE_HW_BOUND_CHECK
+    WASMExecEnv *exec_env_tls = wasm_runtime_get_exec_env_tls();
+    // Prefer setting the exception in the current thread's exec env.
+    if (exec_env_tls) {
+        if (exception) {
+            snprintf(exec_env_tls->cur_exception,
+                     sizeof(module_inst->cur_exception),
+                     "Exception: %s", exception);
+        } else {
+            exec_env_tls->cur_exception[0] = '\0';
+        }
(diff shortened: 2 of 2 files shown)


core/iwasm/common/wasm_exec_env.h
```diff
@@ -166,6 +168,9 @@ typedef struct WASMExecEnv {
         /* The WASM stack. */
         uint8 bottom[1];
     } wasm_stack_u;
+
+    /* The exception buffer for current execution environment. */
+    char cur_exception[EXCEPTION_BUF_LEN];
```
loganek, line 173: Not sure if we can safely add anything  below `wasm_stack_u.bottom` as this is the region used for the WASM stack frames.

paul-hoang: I see, I can move it to anywhere in `WASMExecEnv` struct. Is after `module_inst` OK?

lum1n0us:
>  would allow using a single module instance in parallel ... Using a single module instance in parallel might not be usual, 

I'm not entirely clear on the concept of "parallel" here. Does it mean using a single instance across multiple threads?

paul-hoang:
> I'm not entirely clear on the concept of "parallel" here. Does it mean using a single instance across multiple threads?

Yes, it means use a single instance across multiple threads at the same time without locking.

yamt:
> Using a single module instance in parallel might not be usual, but it's possible from what I understand and tested WAMR, at least for specific use cases (e.g. no writing to globals or wasm memory, only reading from that and writing to wasm stack). Please correct me if I'm missing anything.

currently it isn't possible.
cf. https://github.com/bytecodealliance/wasm-micro-runtime/issues/2512

> This PR proposes a change in how we store and retrieve exception from a given module instance:

i don't see what's the point of the change. it just makes it more inconsistent, doesn't it?

lum1n0us:
Indeed, it is a very unusual scenario and somewhat conflicts with some fundamental principles of WebAssembly, possibly the weak wasm thread model or the threads proposal or others, I can't quite remember which.

paul-hoang:
I see your points, @lum1n0us and @yamt . Maybe WAMR is currently written in a threading model where module instance is basically 1:1 to exec_env (as mentioned in #2512). But I think a "pure" exported wasm function (i.e. not writing to wasm globals and wasm memory) can benefit a lot from being called in parallel using a single module instance because creating a module instance is expensive in both memory and time.

"pure" wasm function is indeed unusual, and refactoring/rewriting major part of WAMR to efficiently support it correctly is not a high priority. I'll probably close this PR and maintain a similar internal patch on my side if needed.

lum1n0us:
>  a "pure" exported wasm function (i.e. not writing to wasm globals and wasm memory) can benefit a lot from being called in parallel using a single module instance because creating a module instance is expensive in both memory and time.

Just out of curiosity, is this just a theoretical exercise, or is there an actual case in question?

paul-hoang:
> > a "pure" exported wasm function (i.e. not writing to wasm globals and wasm memory) can benefit a lot from being called in parallel using a single module instance because creating a module instance is expensive in both memory and time.
> 
> Just out of curiosity, is this just a theoretical exercise, or is there an actual case in question?

There is an actual case. I can't share more details, but a wasm module can be used as a scripting language where most of its work is to call host functions to do something, so the wasm can be "pure".

Closing this PR since as mentioned in a previous comment.

PR #4010, opened 2025-01-07
add hashmap to get_const_offset

sjamesr:


sjamesr:
This change has worse performance than HEAD for small modules, but it is a substantial improvement for modules with large numbers of constants (I saw a 18% speedup in loading a module with 10,000 consts).

Ideas:
- Right now there are at least 3 allocations per insertion. Right now the hashmap keys and const buffer memory cannot be shared, because the constant buffer will be realloc'd when it grows, potentially invalidating the pointers in the hashmap. Perhaps the constant buffer could be a linked list of chunks of N constants. This way, more constant buffer space could be added without reallocating the existing buffer.

- hash map uses chaining to resolve collisions. Moving to open addressing might reduce the number of collisions.

- maybe the loader should be smarter about using a hash map only when the number of constants grows beyond a certain threshold

- maybe toolchains should emit table references instead of `{i32,i64,f32,f64}.const` instructions, thus moving the de-dup/compression step to the assembler/linker and out of the loader

wenyongh:
> This change has worse performance than HEAD for small modules, but it is a substantial improvement for modules with large numbers of constants (I saw a 18% speedup in loading a module with 10,000 consts).
> 

@sjamesr thanks for your effort to improve the performance, I haven't read the detailed code to understand the optimization, but it seems that a 18% speedup is not as expected. I tried to improve the perpformance with another PR (see #4012), and tested the standalone case in `tests/standalone/test-printf` which has more than 20000 consts, experiments show that the PR can greatly improve the performance for this case: the load time is reduced from ~550ms to ~5ms. You can tested it with the below command:
```bash
cd tests/standalone/test-printf
time iwasm --heap-size=102400 --stack-size=8096000 test_printf_builtin.wasm
# or modify the source code to print the execution time of wasm_loader_load manually
```
I also tested you patch, but it didn't improve the loading time for the test-print case, instead, it is even worse. So I think maybe my PR is better? And cound you test my PR for your case? Thanks in advance.

sjamesr:
This does indeed show a substantial improvement in our benchmark, a 15% speedup in loading a module with ~1000 consts. After this, the largest single cost in our benchmark is `read_leb`.

Thank you for doing this!

My change adds a bh_hash_map of `Const -> offset`, returning the existing offset if it's there or adding a new entry to the map if it's not. Moving bh_hash_map to open addressing and away from chaining might have a substantial impact by reducing the number of allocations, and this map is used elsewhere in the code. But for now I think I'll drop my change and see if we can improve `read_leb`.

wenyongh:
OK, welcome, I refactored some code and applied similar changes to wasm mini loader. If you are to drop this PR, then I will let others help review my PR.

For reading leb in wasm loader of fast interpreter, one optimization I can think is that we don't need to check the leb format again in the second scan since the checks have been done in the first scan. I also submitted another PR (see #4012) for it, but it doesn't improve performance a lot for the case test-printf. Maybe you can have a try for your case.

Another optimization is to refine the `bh_read_leb` function for the first scan, but I have no idea yet, maybe you can have a try also.

sjamesr:
https://github.com/bytecodealliance/wasm-micro-runtime/pull/4012

PR #4011, opened 2025-01-07
Improve stack consistency by ensuring sufficient space for dummy offsets

lum1n0us:
One more corner case: if the `frame_offset` increases and becomes equal to the `frame_offset_boundary` after the last assignment within the for loop.


xujuntwt95329 (approved):
LGTM

TianlongLiang (approved):
LGTM

no1wudi (approved):

PR #4012, opened 2025-01-09
Refine getting const offsets in wasm loader of fast-interp

wenyongh:
Getting const offsets in the wasm loader of fast interpreter may take a lot of
time, especially when there are lots of const opcodes in a wasm function,
since the loader creates a temp const list and searchs whether a const is
in the list by traversing the elements of the list one by one, and it also uses
a `Const` struct which is big as the list element.

This PR creates two const lists instead, whose element is int64 and int32. And
it treats i64/f64 the same and treats i32/f32 the same. In the firt time scan,
it just appends element to the lists, and in the second time scan, it sorts the lists
and remove duplicated elements first, and then binary search them for
duplicated consts. Experiments show that it improves the performance a lot
when there are a lot of consts.

### DIFF
--- a/core/iwasm/interpreter/wasm_loader.c
+++ b/core/iwasm/interpreter/wasm_loader.c
@@ -9538,12 +9538,31 @@ wasm_loader_get_const_offset(WASMLoaderContext *ctx, uint8 type, void *value,
                              uint32 error_buf_size)
 {
     if (!ctx->p_code_compiled) {
-        /* TODO: handle when total const cell num is overflow */
         if (type == VALUE_TYPE_I64 || type == VALUE_TYPE_F64) {
-            I64ConstNode *i64_const_node =
-                loader_malloc(sizeof(I64ConstNode), error_buf, error_buf_size);
-            if (!i64_const_node)
+            I64ConstNode *i64_const_node;
+
+            /* No slot left, emit const instead */
+            if (ctx->i64_const_num * 2 + ctx->i32_const_num > INT16_MAX - 2) {
+                *offset = 0;
+                return true;
+            }
+
+            /* Traverse the list if the const num is small */
+            if (ctx->i64_const_num > 0 && ctx->i64_const_num < 10) {
+                i64_const_node = ctx->i64_const_list;
+                while (i64_const_node) {
+                    if (i64_const_node->value == *(int64 *)value) {
+                        *offset = -1;
+                        return true;
+                    }
+                    i64_const_node = i64_const_node->next;
+                }
+            }
+
+            if (!(i64_const_node = loader_malloc(sizeof(I64ConstNode),
+                                                 error_buf, error_buf_size))) {
                 return false;
+            }
             i64_const_node->value = *(int64 *)value;
             i64_const_node->next = ctx->i64_const_list;
             ctx->i64_const_list = i64_const_node;
@@ -9551,10 +9570,31 @@ wasm_loader_get_const_offset(WASMLoaderContext *ctx, uint8 type, void *value,
         }
         else {
             bh_assert(type == VALUE_TYPE_I32 || type == VALUE_TYPE_F32);
-            I32ConstNode *i32_const_node =
-                loader_malloc(sizeof(I32ConstNode), error_buf, error_buf_size);
-            if (!i32_const_node)
+
+            I32ConstNode *i32_const_node;
+
+            /* No slot left, emit const instead */
+            if (ctx->i64_const_num * 2 + ctx->i32_const_num > INT16_MAX - 1) {
+                *offset = 0;
+                return true;
+            }
+
+            /* Traverse the list if the const num is small */
+            if (ctx->i32_const_num > 0 && ctx->i32_const_num < 10) {
+                i32_const_node = ctx->i32_const_list;
+                while (i32_const_node) {
+                    if (i32_const_node->value == *(int32 *)value) {
+                        *offset = -1;
+                        return true;
+                    }
+                    i32_const_node = i32_const_node->next;
+                }
+            }
+
+            if (!(i32_const_node = loader_malloc(sizeof(I32ConstNode),
+                                                 error_buf, error_buf_size))) {
                 return false;
+            }
             i32_const_node->value = *(int32 *)value;
             i32_const_node->next = ctx->i32_const_list;
             ctx->i32_const_list = i32_const_node;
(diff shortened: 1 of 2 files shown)


xujuntwt95329 (approved):
LGTM

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -356,9 +355,14 @@ memory_realloc(void *mem_old, uint32 size_old, uint32 size_new, char *error_buf,
 {
     uint8 *mem_new;
     bh_assert(size_new > size_old);
+
+    if ((mem_new = wasm_runtime_realloc(mem_old, size_new))) {
+        memset(mem_new + size_old, 0, size_new - size_old);
+        return mem_new;
+    }
+
```
lum1n0us, line 363: > The realloc() function changes the size of the memory block pointed to by ptr to size bytes.  The contents will be  unchanged in the range from the start of the region up to the minimum of the old and new sizes.  If the new size is larger than the old size, the added memory will not be initialized.  If ptr is NULL, then the  call  is equivalent  to  malloc(size),  for  all values of size; if size is equal to zero, and ptr is not NULL, then the call is equivalent to free(ptr). 

Need to prevent from `free()` by rejecting the case of `mem_old!=NULL` and `size_new==0`. especially in release building.


Beside being `free()`, a returned NULL from `realloc()` means *out of memory*, IIUC. In that case, `loader_malloc()` will be failed too. So I guess we might don't need `loader_malloc()` as a failed-safer.

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -7950,10 +7954,12 @@ typedef struct WASMLoaderContext {
     int16 preserved_local_offset;
 
     /* const buffer */
-    uint8 *const_buf;
-    uint16 num_const;
-    uint16 const_cell_num;
-    uint32 const_buf_size;
+    int64 *i64_consts;
```
lum1n0us, line 7957: I am wondering how about something like `8_bytes_consts`? Ultimately, we ended up storing a mix of i64 and f64 values, rather than only i64.

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -8238,10 +8240,15 @@ wasm_loader_ctx_init(WASMFunction *func, char *error_buf, uint32 error_buf_size)
         goto fail;
     loader_ctx->frame_offset_boundary = loader_ctx->frame_offset_bottom + 32;
 
-    loader_ctx->num_const = 0;
-    loader_ctx->const_buf_size = sizeof(Const) * 8;
-    if (!(loader_ctx->const_buf = loader_malloc(loader_ctx->const_buf_size,
-                                                error_buf, error_buf_size)))
+    loader_ctx->i64_const_size = 8;
```
lum1n0us, line 8243: size -> count?

loganek:
Thanks for the change. Would you provide a bit more details on the experiments you performed? What exactly did you benchmark and how much was the improvement?

core/iwasm/interpreter/wasm_loader.c
wenyongh: > 
> Need to prevent from `free()` by rejecting the case of `mem_old!=NULL` and `size_new==0`. especially in release building.
> 

In memory reallocation here, there won't be case `mem_old != NULL && size_new == 0`, I think we can just add an assertion here.

> Beside being `free()`, a returned NULL from `realloc()` means _out of memory_, IIUC. In that case, `loader_malloc()` will be failed too. So I guess we might don't need `loader_malloc()` as a failed-safer.

The `wasm_runtime_malloc` can be configured to various allocators during runtime initialization, it seems good to ignore loader_malloc if it is configured to some allocations. But developer may provide its allocator without realloc function, so I think we had better call load_malloc again.

core/iwasm/interpreter/wasm_loader.c
wenyongh: We treat f64 as i64 by reading its content with `*(int64 *)&f64`, and we sort and binary search the array with i64 elements, I think it is good to name it as `i64_consts`. We can add some comments for it.

core/iwasm/interpreter/wasm_loader.c
wenyongh: `count` also makes confusing, I think `i64_const_max_num` is better.

core/iwasm/interpreter/wasm_loader.c
wenyongh: `bh_assert(size_new > size_old);` has been added in L357, so `size_new` must be larger than 0, no need to add new assertion.

wenyongh:
> Thanks for the change. Would you provide a bit more details on the experiments you performed? What exactly did you benchmark and how much was the improvement?

I tested the standalone case `tests/standalone/test-printf` under WAMR, which has more than 30000 i64.const/i32.const opcodes:
```bash
cd <wamr_root>/roduct-mini/platforms/linux
mkdir build && cd build
cmake .. && make -j         # iwasm is built with fast-interp enabled
cd <wamr_root>/tests/standalone/test-printf
time iwasm --heap-size=102400 --stack-size=809600 test_printf_builtin.wasm > a.log
```

Or do some modifications to just test the wasm loader many times, e.g. apply below patch (test loader 101 times):
```diff
diff --git a/product-mini/platforms/posix/main.c b/product-mini/platforms/posix/main.c
index af50223a..6326f270 100644
--- a/product-mini/platforms/posix/main.c
+++ b/product-mini/platforms/posix/main.c
@@ -921,6 +921,15 @@ main(int argc, char *argv[])
                                    module_destroyer_callback);
 #endif
 
+    for (uint32 i = 0; i < 100; i++) {
+        wasm_module = wasm_runtime_load(wasm_file_buf, wasm_file_size,
+                                          error_buf, sizeof(error_buf));
+        wasm_runtime_unload(wasm_module);
+        wasm_runtime_free(wasm_file_buf);
+        wasm_file_buf =
+              (uint8 *)bh_read_file_to_buffer(wasm_file, &wasm_file_size);
+    }
+
     /* load WASM module */
     if (!(wasm_module = wasm_runtime_load(wasm_file_buf, wasm_file_size,
                                           error_buf, sizeof(error_buf)))) {
@@ -928,6 +937,7 @@ main(int argc, char *argv[])
         goto fail2;
     }
 
+#if 0
 #if WASM_ENABLE_DYNAMIC_AOT_DEBUG != 0
     if (!wasm_runtime_set_module_name(wasm_module, wasm_file, error_buf,
                                       sizeof(error_buf))) {
@@ -1041,6 +1051,7 @@ fail4:
     wasm_runtime_deinstantiate(wasm_module_inst);
 
 fail3:
+#endif
     /* unload the module */
     wasm_runtime_unload(wasm_module);
```
And then
```bash
time iwasm --heap-size=102400 --stack-size=809600 test_printf_builtin.wasm
```

My experiment shows that the time of 101 times loading can be reduced from ~39s to ~0.5s. Note that the testing also includes wasm_runtime_unload and bh_read_file_to_buffer, but I think their time should be relatively small.

lum1n0us (approved):
LGTM

loganek (commented):
LGTM

PR #4017, opened 2025-01-10
Refine read leb in wasm loader of fast interpreter

wenyongh:
For the wasm loader of fast interpreter, since the leb format has been checked
in the first time traversing, we can skip the check in the second time traversing.
This PR calls read_leb_quick instead of read_leb_xxx in the second time scan to
improve the loader's performance.

### DIFF
--- a/core/iwasm/interpreter/wasm_loader.c
+++ b/core/iwasm/interpreter/wasm_loader.c
@@ -197,7 +197,6 @@ check_buf1(const uint8 *buf, const uint8 *buf_end, uint32 length,
         res = (int32)res64;                                             \
     } while (0)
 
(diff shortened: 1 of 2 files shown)


lum1n0us (approved):
LGTM.

Just a thought – perhaps a leb-cache (mapping from leb values to constant values) would be more effective. It could be shared across different running modes.

TianlongLiang (approved):
LGTM

wenyongh:
> LGTM.
> 
> Just a thought – perhaps a leb-cache (mapping from leb values to constant values) would be more effective. It could be shared across different running modes.

leb128 is encoded with variant length, the length of bytes can not be pre-known, and the so-called value may have more than 4 bytes for uint32/int32 and more than 8 bytes for int64, I just have no good idea to create the map.

PR #4018, opened 2025-01-13
[instantiation linking] Optimize aot func inst list

lum1n0us:
https://github.com/bytecodealliance/wasm-micro-runtime/pull/3947#discussion_r1904834621

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -787,8 +787,7 @@ functions_instantiate(AOTModuleInstance *module_inst, AOTModule *module,
                       const WASMExternInstance *imports, uint32 import_count,
                       char *error_buf, uint32 error_buf_size)
 {
-    uint32 function_count = module->import_func_count + module->func_count;
-    uint64 total_size = sizeof(AOTFunctionInstance) * (uint64)function_count;
+    uint64 total_size = sizeof(AOTFunctionInstance) * module->import_func_count;
 
     AOTFunctionInstance *functions =
         runtime_malloc(total_size, error_buf, error_buf_size);
@@ -815,7 +814,7 @@ functions_instantiate(AOTModuleInstance *module_inst, AOTModule *module,
                 set_error_buf_v(error_buf, error_buf_size,
                                 "unknown import module \"%s\"",
                                 import->module_name);
-                return NULL;
+                goto fail;
             }
 
             function->import_func_inst = aot_lookup_function(
@@ -824,7 +823,7 @@ functions_instantiate(AOTModuleInstance *module_inst, AOTModule *module,
                 set_error_buf_v(error_buf, error_buf_size,
                                 "unknown import function \"%s\"",
                                 import->func_name);
-                return NULL;
+                goto fail;
             }
         }
 
@@ -860,7 +859,7 @@ functions_instantiate(AOTModuleInstance *module_inst, AOTModule *module,
             LOG_ERROR(
                 "mismatched import memory name: expect \"%s\", got \"%s\"",
                 import->func_name, extern_inst->field_name);
-            return NULL;
+            goto fail;
         }
 
         /* from other .wasm */
@@ -877,18 +876,10 @@ functions_instantiate(AOTModuleInstance *module_inst, AOTModule *module,
 #endif
     }
 
-    for (uint32 i = 0; i < module->func_count; i++, function++) {
-        function->is_import_func = false;
-        function->func_index = i + module->import_func_count;
-
-        uint32 ftype_index = module->func_type_indexes[i];
-        function->u.func.func_type = (AOTFuncType *)module->types[ftype_index];
-        function->u.func.func_ptr = module->func_ptrs[i];
-    }
-
-    bh_assert((uint32)(function - functions) == function_count);
-
     return functions;
+fail:
+    wasm_runtime_free(functions);
+    return NULL;
 }
 
 /**
@@ -1754,7 +1745,7 @@ init_func_ptrs(AOTModuleInstance *module_inst, AOTModule *module,
     func_ptrs = (void **)module_inst->func_ptrs;
     for (i = 0; i < module->import_func_count; i++, func_ptrs++) {
         AOTFunctionInstance *func =
-            aot_locate_function_instance(module_inst, i);
+            aot_locate_import_function_instance(module_inst, i);
         bh_assert(func->is_import_func);
 
         if (func->call_conv_wasm_c_api) {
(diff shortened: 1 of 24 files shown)


core/iwasm/aot/aot_runtime.c
```diff
@@ -760,6 +782,106 @@ globals_instantiate(AOTModuleInstance *module_inst, AOTModule *module,
...
+
+        /* from c_api (loading) */
+        function->call_conv_wasm_c_api = import->call_conv_wasm_c_api;
+#else
+        const WASMExternInstance *extern_inst =
+            wasm_runtime_get_extern_instance(imports, import_count,
+                                             WASM_IMPORT_EXPORT_KIND_FUNC, i);
+        if (!extern_inst) {
+            LOG_DEBUG("no import function(%s, %s) from imports list, might "
+                      "provied by wasm_native",
```
wenyongh, line 838: provided

core/iwasm/aot/aot_runtime.c
```diff
@@ -1621,13 +1744,29 @@ init_func_ptrs(AOTModuleInstance *module_inst, AOTModule *module,
     /* Set import function pointers */
     func_ptrs = (void **)module_inst->func_ptrs;
     for (i = 0; i < module->import_func_count; i++, func_ptrs++) {
-        *func_ptrs = (void *)module->import_funcs[i].func_ptr_linked;
-        if (!*func_ptrs) {
-            const char *module_name = module->import_funcs[i].module_name;
-            const char *field_name = module->import_funcs[i].func_name;
-            LOG_WARNING("warning: failed to link import function (%s, %s)",
-                        module_name, field_name);
```
wenyongh, line 1629: This is an intentional warning, it may be important to developed, could we keep it in the new code if link failed?

core/iwasm/aot/aot_runtime.c
```diff
@@ -1621,13 +1744,29 @@ init_func_ptrs(AOTModuleInstance *module_inst, AOTModule *module,
...
+            continue;
         }
+
+        AOTImportFunc *import_func = func->u.func_import;
+
+        LOG_DEBUG("use wasm_native linked functions for (%s,%s)",
+                  import_func->module_name, import_func->func_name);
+
+        *func_ptrs = import_func->func_ptr_linked;
+        bh_assert(*func_ptrs);
```
wenyongh, line 1769: Per my understanding, the import function doesn't have to be linked and runtime can print warning here, and during execution, the aot code or runtime throws exception "failed to call unlinked import function". The original behavior is like that, we had better not change it?

core/iwasm/aot/aot_runtime.c
```diff
@@ -2300,15 +2384,25 @@ aot_instantiate(AOTModule *module, AOTModuleInstance *parent,
                               imports, import_count, error_buf, error_buf_size))
         goto fail;
 
+    extra->function_count = module->import_func_count + module->func_count;
+    if (extra->function_count > 0) {
+        extra->import_functions =
+            functions_instantiate(module_inst, module, imports, import_count,
```
wenyongh, line 2390: how about renaming it to `import_functions_instantiate`?

core/iwasm/common/wasm_extern_inst.c
wenyongh, line 1: If this file isn't needed, had better remove it?

core/iwasm/common/wasm_extern_inst.h
wenyongh, line 1: ditto

core/iwasm/common/wasm_c_api.c
```diff
@@ -4953,6 +4937,70 @@ wasm_instance_new_with_args(wasm_store_t *store, const wasm_module_t *module,
...
+
+    if (import_types.num_elems > imports_src->num_elems) {
+        LOG_ERROR("module has %d imports but only %d provided. Placeholders "
+                  "are even allowed.\n",
+                  import_types.num_elems, imports_src->num_elems);
+        return false;
+    }
+
+    *imports_dst =
+        malloc_internal(sizeof(WASMExternInstance) * imports_src->num_elems);
```
wenyongh, line 4971: how about `(uint64)sizeof(WASMExternInstance)`?

core/iwasm/aot/aot_runtime.h
```diff
@@ -225,7 +238,7 @@ typedef struct AOTModule {
 
     /* function info */
     uint32 func_count;
-    /* func pointers of AOTed (un-imported) functions */
+    /* func pointers of AOTed (imported + local) functions */
```
wenyongh, line 241: how about also changing comment of L243?

core/iwasm/aot/aot_runtime.c
```diff
@@ -5894,4 +6013,26 @@ aot_destroy_global(AOTGlobalInstance *global)
...
     wasm_runtime_free(global);
-}
\ No newline at end of file
+}
+
+AOTFunctionInstance *
+aot_create_function_empty(const AOTModule *module)
+{
+    AOTFunctionInstance *function = runtime_malloc(
+        sizeof(AOTFunctionInstance) + sizeof(AOTImportFunc), NULL, 0);
```
wenyongh, line 6022: A little confused why need to plus `sizeof(AOTImportFunc)`?

lum1n0us: I am considering that this would enable host APIs (defined in wasm_export.h) to create an `AOTFunctionInstance` from scratch and pass it to the runtime for execution. The runtime would then use the `u.func_import` field to retrieve some information. However, there might be room for optimization in subsequent pull requests.

core/iwasm/aot/aot_runtime.h
lum1n0us: `func_type_indexes` still only stores for local(un-imported) functions.

core/iwasm/aot/aot_runtime.c
wenyongh: That brings convenience to developers and is a common handle in many VMs like jvm. In many cases the unlinked import function isn't called and developer can still call other functions. No developer reports it a problem, I strongly suggest to keep this. If you want to fix this, had better open another PR to let others to review/discuss and not fix it in this PR.

lum1n0us:
> In many cases the unlinked import function isn't called and developer can still call other functions. 

However, this does not align with the specification definition. I believed the purpose of creating the "instantiation linking" series of pull requests was to align with the specification requirements. Otherwise, our main branch code is functioning quite well.

core/iwasm/aot/aot_runtime.c
lum1n0us: > In many cases the unlinked import function isn't called and developer can still call other functions.

However, this does not align with the specification definition. I believed the purpose of creating the "instantiation linking" series of pull requests was to align with the specification requirements. Otherwise, our main branch code is functioning quite well.

wenyongh:
> > In many cases the unlinked import function isn't called and developer can still call other functions.
> 
> However, this does not align with the specification definition. I believed the purpose of creating the "instantiation linking" series of pull requests was to align with the specification requirements. Otherwise, our main branch code is functioning quite well.

It reports the error during the execution time but not instantiation time, it doesn't strictly obey the spec. But sometimes the user experience is more important, in fact, we also do some special handlings in other places. I still think it is good to keep it, it also reports error and it is more friendly to developers.

wenyongh (commented):
LGTM with minor comments

PR #4033, opened 2025-01-17
Copy callstack API

g0djan:
New WAMR public API to copy runtime call stack frames.

CAUTION: this APIs is not thread safe, that's why it's hidden behind feature flag for now. If you need to call it from another thread ensure the passed exec_env is suspended.

## Our use case
Sometimes WAMR runtime gets stuck in production and we have no data where in the code compiled to WASM it happens. We currently only track such situations in a separate native thread. To increase visibility into the problem we developed internal solution that requires presence of this API in WAMR. If a separate thread finds that the WASM VM thread has stuck, it interrupts it with a user defined signal and calls this API to collect callstack. The main complexity is maintaining [async-signal-safety](https://man7.org/linux/man-pages/man7/signal-safety.7.html) and avoiding segfaults. For that we're maintaining[ atomic copies](https://en.cppreference.com/w/cpp/utility/program/signal) of `exec_env`, `exec_env->module_inst`, `exec_env->module_inst->module`. Those copies are always set to NULL before the referenced memory is freed. Before a call to this API those copies are always checked for validity. In our use case scenario we guarantee ourselves only absence of crashes but we realize that the frame data that we collect might be invalidated due to a signal interruption. However it's highly unlikely and is not a concern for us.

## Have we tried existing WAMR APIs for our usecase?
Yes, we've tried suggested by maintainers `wasm_cluster_suspend_thread` and `wasm_runtime_terminate`.
1. In our production runtime often recovers from being stuck, so wasm_runtime_terminate is not a good option for us to report the call stack
2. The `wasm_cluster_suspend_thread` doesn't suit us either. Even if it did we'd still need API to iterate over stackframes.


core/iwasm/aot/aot_runtime.c
```diff
@@ -10,6 +10,7 @@
 #include "../common/wasm_runtime_common.h"
 #include "../common/wasm_memory.h"
 #include "../interpreter/wasm_runtime.h"
+#include <stdbool.h>
```
loganek, line 13: wamr defines its own boolean type, you should use that instead of stdbool

core/iwasm/aot/aot_runtime.c
```diff
@@ -4104,6 +4105,48 @@ aot_frame_update_profile_info(WASMExecEnv *exec_env, bool alloc_frame)
...
+{
+/*
+* Note for devs: please refrain from such modifications inside of aot_iterate_callstack
+ * - any allocations/freeing memory
+ * - dereferencing any pointers other than: exec_env, exec_env->module_inst,
+ * exec_env->module_inst->module, pointers between stack's bottom and top_boundary
+ * For more details check wasm_iterate_callstack in wasm_export.h
+*/
+    if (!is_tiny_frame(exec_env)) {
+        //TODO: support standard frames
```
loganek, line 4119: should we print an error/warning here? Or ideally, can we provide an implementation of that? I think there's already a code in this file to iterate through both types of frames so can be used as a reference

core/iwasm/aot/aot_runtime.c
```diff
@@ -4104,6 +4105,48 @@ aot_frame_update_profile_info(WASMExecEnv *exec_env, bool alloc_frame)
...
+    bool is_top_index_in_range = top_boundary >= top && top >= (bottom + sizeof(AOTTinyFrame));
+    if (!is_top_index_in_range) {
+        return;
+    }
+    bool is_top_aligned_with_bottom = (unsigned long)(top - bottom) % sizeof(AOTTinyFrame) == 0;
+    if (!is_top_aligned_with_bottom) {
+        return;
+    }
+
+    AOTTinyFrame* frame = (AOTTinyFrame*)(top - sizeof(AOTTinyFrame));
```
loganek, line 4135: I wonder if we want to iterate from top to bottom or from bottom to top? bottom pointer is constant but the top one might change during the signal; so if you iterate from top, you might actually start with some very high address if you're unluckly because the pointer will be invalid.

core/iwasm/include/wasm_export.h
```diff
@@ -864,6 +867,37 @@ wasm_runtime_create_exec_env(wasm_module_inst_t module_inst,
 WASM_RUNTIME_API_EXTERN void
 wasm_runtime_destroy_exec_env(wasm_exec_env_t exec_env);
 
+
+typedef bool (*wasm_frame_callback)(void*, wasm_frame_ptr_t);
```
loganek, line 871: It'd be good to document the meaning of the bool return type.

core/iwasm/include/wasm_export.h
```diff
@@ -864,6 +867,37 @@ wasm_runtime_create_exec_env(wasm_module_inst_t module_inst,
...
+ * for wasm_frame_t from wasm_c_api.h
+ *
+ * Note: The function is async-signal-safe if called with verified arguments. 
+ * Meaning it's safe to call it from a signal handler even on a signal interruption
+ * from another thread if next variables hold valid pointers
+ * - exec_env
+ * - exec_env->module_inst
+ * - exec_env->module_inst->module
+ *
+ * Note for devs: please refrain from such modifications inside of this call
```
loganek, line 889: I think it's fine to just keep it in the `.c` file, no need to expose it publicly.

core/iwasm/interpreter/wasm_interp.h
```diff
@@ -26,6 +26,8 @@ typedef struct WASMInterpFrame {
     /* Instruction pointer of the bytecode array.  */
     uint8 *ip;
 
+    uint32 func_index;
```
loganek, line 29: why do we need it? we already have function above.

core/iwasm/include/wasm_export.h
```diff
@@ -864,6 +867,37 @@ wasm_runtime_create_exec_env(wasm_module_inst_t module_inst,
...
+ * - any allocations/freeing memory
```
denizsokmen, line 890: I think it's up to the developer to provide a safe environment to do these actions. I may as well call this from within wasm code via a native api just to print call stack arbitrarily while it's perfectly safe to malloc anything. I think the only thing this api needs to document/ensure is that it's async-signal-safe to traverse the call stack

core/iwasm/aot/aot_runtime.c
g0djan: I'll take a look if I can implement it as well

core/iwasm/interpreter/wasm_interp.h
g0djan: so now it's calculated like this https://github.com/bytecodealliance/wasm-micro-runtime/blob/b6dea221a609606ce6c3dcaa0f4d0399ea5bc4dc/core/iwasm/interpreter/wasm_interp_fast.c#L1210
which requires dereferencing `module_inst->e->functions`. Always or almost always those pointers remain the same.

But if `wasm_table_get_func_inst` is called then functions pointer gets updated as I can see in https://github.com/bytecodealliance/wasm-micro-runtime/blame/b6dea221a609606ce6c3dcaa0f4d0399ea5bc4dc/core/iwasm/aot/aot_runtime.c#L1369

core/iwasm/include/wasm_export.h
g0djan: Yes, it's for sure can be called just through a native api without these restrictions. Maybe we can just mark it as that I just think that more info for the user is better if they'd like to call it from a signal handler

core/iwasm/include/wasm_export.h
g0djan: Okay, the modifications are done there anyway

core/iwasm/include/wasm_export.h
g0djan: Removed this bit from here, kept only in the .c file implementation

core/iwasm/aot/aot_runtime.c
loganek: > it's better to handle cases where the stack is big and it's better to process only top 100 frames for example

I don't have a hard data on that but I guess in "average application" stack traces aren't expected to be that big? Or can we make the `100` configurable? 

> for wasm it's much harder to go bottom to top because there are no pointers that point to next frame and I haven't found a way to calculate start of the next frame on the stack manually

Agree; however, AOTTinyFrame has a fixed size, so navigating top/bottom is just by moving pointer by sizeof(AOTTinyFrame) bytes.

g0djan: Also before that I do a check that `top` is aligned with bottom, so even it can be invalid it does the same thing as if we started from the bottom and can be controlled through the callback the same way

core/iwasm/interpreter/wasm_interp.h
g0djan: if it's always the case not clear why [this line](https://github.com/bytecodealliance/wasm-micro-runtime/blame/b6dea221a609606ce6c3dcaa0f4d0399ea5bc4dc/core/iwasm/aot/aot_runtime.c#L1369) exists then

core/iwasm/aot/aot_runtime.c
g0djan: I added support for standard frames, the same way as it's done for wasm

core/iwasm/interpreter/wasm_interp.h
g0djan: ok, it seems if this code `module->import_function_count + module->function_count` can never be 0 then yeah it's never modified

But can there be a wasm module with this sum equal 0?

g0djan: I reread the logic there
- [the pointer modified](https://github.com/bytecodealliance/wasm-micro-runtime/blame/b6dea221a609606ce6c3dcaa0f4d0399ea5bc4dc/core/iwasm/aot/aot_runtime.c#L1369) only if pointer is not allocated and `module->import_func_count + module->func_count` equals 0.
- but [the pointer is not allocated](https://github.com/bytecodealliance/wasm-micro-runtime/blob/b6dea221a609606ce6c3dcaa0f4d0399ea5bc4dc/core/iwasm/interpreter/wasm_runtime.c#L2555) only if `module->import_func_count + module->func_count` equals 0.

=> this branch of code with allocating `functions` inside of `wasm_table_get_func_inst` is never executed in fact

g0djan: I removed thios extra field and added an explanation comment, why it's safe to calculate manually

g0djan:
@loganek addressed all your comments and rebased to fix the checks. The last failing check is CI issue and there's another PR that will fix it.
Let me know if you have more comments

core/iwasm/aot/aot_runtime.c
```diff
@@ -4104,6 +4104,105 @@ aot_frame_update_profile_info(WASMExecEnv *exec_env, bool alloc_frame)
...
+     */
+    WASMModuleInstance *module_inst =
+        (WASMModuleInstance *)wasm_exec_env_get_module_inst(exec_env);
+    AOTFrame *cur_frame = (AOTFrame *)wasm_exec_env_get_cur_frame(exec_env);
+    uint8 *top_boundary = exec_env->wasm_stack.top_boundary;
+    uint8 *bottom = exec_env->wasm_stack.bottom;
+
+    WASMCApiFrame record_frame;
+    while (cur_frame && (uint8_t *)cur_frame >= bottom
+           && (uint8_t *)cur_frame + sizeof(AOTFrame) <= top_boundary) {
```
loganek, line 4172: just for a clarity, AOTFrame has a variable size so this check is not exactly right, but I think it's good enough for this purpose.

g0djan: >has a variable size

Size of this structure can vary during one session?

g0djan: ok, I figured you meant `uint32 size = (uint32)offsetof(AOTFrame, lp)`

Added it

loganek: No, I meant for the GC case the frame size might have variable size as it depends on e.g. locals

g0djan: got it, modified it and added comment that standard frames needs future work to work when GC is enabled

g0djan:
@lum1n0us could you take a look at this PR?

g0djan:
@yamt  could you please take a look at this PR?

core/iwasm/include/wasm_export.h
```diff
@@ -864,6 +867,38 @@ wasm_runtime_create_exec_env(wasm_module_inst_t module_inst,
...
+ *
+ * Usage: In the callback to read frames fields use APIs
+ * for wasm_frame_t from wasm_c_api.h
+ *
+ * Note: The function is async-signal-safe if called with verified arguments.
+ * Meaning it's safe to call it from a signal handler even on a signal
+ * interruption from another thread if next variables hold valid pointers
+ * - exec_env
+ * - exec_env->module_inst
+ * - exec_env->module_inst->module
```
yamt, line 892: IMO, making this complex functionality async-signal-safe is too much maintenance burden, especially when wamr itself doesn't rely on the property at all.

given that you need to suspend the target thread anyway, why don't you call this from another thread?

g0djan: >async-signal-safe is too much maintenance burden

Yeah, we understand that and I kept it simple as much as possible. Basically non async-signal-safe implementation would be different only in a few checks removed and there wouldn't be comments that I added in the code.

Particularly this comment about validity of pointers is a theoretical problem atm. We don't know any platform yet where updating pointer variable might be interrupted by a signal, possibly after launch we will see that it never happens.

>given that you need to suspend the target thread anyway, why don't you call this from another thread?

Do you mean using `wasm_cluster_suspend_thread`? 
I tried that and there're 2 problems for us:
- Now there’s no awaiting till thread actually gets suspended
- Suspension happens only after certain checks so we're not getting stacktraces that we need. E.g. if there's sleep somewhere, there won't be sleep in stacktrace reported, it'd report calls after sleep has finished. If there's a deadlock stacktrace won't be reported at all

g0djan: @yamt @lum1n0us

g0djan: @yamt @TianlongLiang @wenyongh Hey guys could someone reply here please? Sorry for pinging multiple times

yamt: > > async-signal-safe is too much maintenance burden
> 
> Yeah, we understand that and I kept it simple as much as possible. Basically non async-signal-safe implementation would be different only in a few checks removed and there wouldn't be comments that I added in the code.
> 
> Particularly this comment about validity of pointers is a theoretical problem atm. We don't know any platform yet where updating pointer variable might be interrupted by a signal, possibly after launch we will see that it never happens.
> 
> > given that you need to suspend the target thread anyway, why don't you call this from another thread?
> 
> Do you mean using `wasm_cluster_suspend_thread`?

i was not thinking about a specific way to suspend a thread.
this api is not expected to work on a running thread, is it?

> I tried that and there're 2 problems for us:
> 
>     * Now there’s no awaiting till thread actually gets suspended
> 
>     * Suspension happens only after certain checks so we're not getting stacktraces that we need. E.g. if there's sleep somewhere, there won't be sleep in stacktrace reported, it'd report calls after sleep has finished. If there's a deadlock stacktrace won't be reported at all

wrt how to suspend a thread, i thought you were planning to use a signal to interrupt blocking system calls, right?
to avoid the signal affecting the target application behavior, depending on the kind of system calls, you might need to somehow wrap the system calls.
you might want to adapt the existing blocking-syscall wrapper api: https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/common/wasm_blocking_op.c
after all, i suspect we will end up with something which shares the guts of the underlying machinery with wasm_runtime_terminate.

or, maybe things like ptrace can be more flexible for your purpose.

lum1n0us:
- It looks like we could refactor the existing APIs `wasm_runtime_dump_call_stack_to_buf()` and `wasm_runtime_dump_call_stack()` using the new API `wasm_iterate_callstack()`. At least print.
- Rather than "interrupting it with a user-defined signal and calling this API to collect the call stack," I would prefer if the runtime, whether interpreter or AOT, could receive the signal and begin dumping the call stack on its own. Even better would be if the runtime could detect an infinite loop, terminate it, and then dump the call stack.  With something like `--fuel` or `--timeout`. https://github.com/bytecodealliance/wasm-micro-runtime/pull/4072#issuecomment-2652465530
- I strongly hope this API can be made thread-safe. As a general solution, it would be extremely valuable.

g0djan:
>It looks like we could refactor the existing APIs wasm_runtime_dump_call_stack_to_buf() and wasm_runtime_dump_call_stack() using the new API wasm_iterate_callstack(). At least print.

Should be possible yeah, I can try

>I would prefer if the runtime, whether interpreter or AOT, could receive the signal and begin dumping the call stack on its own.

It's possible to bring this logic to WAMR, but if we are to do this I would like to have wasm_iterate_callstack API available first and then open a separate PR for dumping callstack on signal interruption

>I strongly hope this API can be made thread-safe

This doesn't go well with async-signal-safety, that's why it's not part of it. But it seems you'd like to have a general approach to inspect stack in a safe manner rather than thread-safety itself

core/iwasm/include/wasm_export.h
g0djan: >this api is not expected to work on a running thread, is it?

It's not. But I don't see that much benefit in calling this API from another thread compared to signal handler. Ensuring validity of pointers that are about to be dereferenced is the main restriction and it remains due to asynchronous nature of signal delivery

g0djan: >depending on the kind of system calls, you might need to somehow wrap the system calls
is there an example that I could check out?

g0djan:
>It looks like we could refactor the existing APIs wasm_runtime_dump_call_stack_to_buf() and wasm_runtime_dump_call_stack() using the new API wasm_iterate_callstack(). At least print.

I started refactoring that, but it comes down to replacing this loop for [wasm_interp](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_runtime.c#L4355)
 or [aot](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/aot/aot_runtime.c#L4247) with iterate_callstack call and internals of the loop are moved inside of the callback

But it doesn't make code simpler imo, it just adds more boilerplate internally. Do you see it useful/meaningful to do so @lum1n0us ?

g0djan:
@lum1n0us I addressed security concerns and made API read-only by reducing it to a copy API instead of using user defined callback. Also I hid the feature behind a feature flag to avoid accidental use of a non thread safe API

Could you please take a look at it?

core/iwasm/include/wasm_export.h
```diff
@@ -126,6 +126,22 @@ typedef WASMFunctionInstanceCommon *wasm_function_inst_t;
 struct WASMMemoryInstance;
 typedef struct WASMMemoryInstance *wasm_memory_inst_t;
 
+
+typedef struct wasm_frame_t {
```
lum1n0us, line 130:
```suggestion
typedef struct WASMCApiFrame{
```

core/iwasm/include/wasm_export.h
```diff
@@ -126,6 +126,22 @@ typedef WASMFunctionInstanceCommon *wasm_function_inst_t;
...
+    uint32_t func_index;
+    uint32_t func_offset;
+    const char *func_name_wp;
+
+    uint32_t *sp;
+    uint8_t *frame_ref;
+    uint32_t *lp;
+} WASMCApiFrame;
+
+typedef struct wasm_frame_t *wasm_frame_ptr_t;
```
lum1n0us, line 143:
```suggestion
typedef struct WASMCApiFrame *wasm_frame_t;
```

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -7,6 +7,7 @@
 #include "bh_common.h"
 #include "bh_assert.h"
 #include "bh_log.h"
+#include "wasm_export.h"
```
lum1n0us, line 10: *wasm_export.h* is for host. no need to include it here

core/iwasm/include/wasm_export.h
```diff
@@ -864,6 +880,34 @@ wasm_runtime_create_exec_env(wasm_module_inst_t module_inst,
...
+ * @param buffer the buffer of size equal length * sizeof(frame) to copy frames to
+ * @param length the number of frames to copy
+ * @param skip_n the number of frames to skip from the top of the stack
+ *
+ * @return number of copied frames
+ */
+WASM_RUNTIME_API_EXTERN uint32_t
+wasm_copy_callstack(const wasm_exec_env_t exec_env, wasm_frame_ptr_t buffer,
+                       const uint32_t length, const uint32_t skip_n);
+
```
lum1n0us, line 910: Are `length` and `skin_n` intended to track the resources consumed by buffer, such as copying the first half of the call stack at one time and the remaining half at another time? If that's the case, wouldn't it be quite challenging to prevent the call stack from changing between two consecutive copy operations?

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -5,6 +5,8 @@
 
 #include "wasm_runtime.h"
 #include "wasm.h"
+#include "wasm_exec_env.h"
+#include "wasm_export.h"
```
lum1n0us, line 9: No need to include more.

core/iwasm/interpreter/wasm_runtime.h
```diff
@@ -12,6 +12,7 @@
 #include "bh_hashmap.h"
 #include "../common/wasm_runtime_common.h"
 #include "../common/wasm_exec_env.h"
+#include "wasm_export.h"
```
lum1n0us, line 15: No need to include more.

lum1n0us (commented):
There are a few questions regarding the design of the APIs

core/iwasm/include/wasm_export.h
g0djan: > such as copying the first half of the call stack at one time and the remaining half at another time?

Yes, the intention is to be able to be able to copy entire callstack in batches even if memory where it's being copied is limited

> Wouldn't it be quite challenging to prevent the call stack from changing between two consecutive copy operations?

If thread is paused/suspended/interrupted then it's all good. It's the case for us and should be the case for any user of this API in its current state anyway

core/iwasm/include/wasm_export.h
g0djan: I'll remove aliasing to `wasm_frame_t` completely since WASMCApiFrame can be used, to avoid confusion with `wasm_frame_t` from wasm_c_api.h

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -4195,7 +4196,57 @@ wasm_get_module_inst_mem_consumption(const WASMModuleInstance *module_inst,
...
+    uint32 count = 0;
+
+    WASMCApiFrame record_frame;
+    while (cur_frame && (uint8_t *)cur_frame >= bottom
+           && (uint8_t *)cur_frame + sizeof(WASMInterpFrame) <= top_boundary
+           && count < (skip_n + length)) {
+        if (!cur_frame->function) {
+            cur_frame = cur_frame->prev_frame;
+            continue;
+        }
```
lum1n0us, line 4228: I observed that the PR only skips this case in interpreter mode. Is this intentional?

lum1n0us (changes requested):
just a few comments.

g0djan:
>just a few comments.

addressed all

core/iwasm/aot/aot_runtime.c
```diff
@@ -10,6 +10,7 @@
 #include "../common/wasm_runtime_common.h"
 #include "../common/wasm_memory.h"
 #include "../interpreter/wasm_runtime.h"
+#include <string.h>
```
lum1n0us, line 13: No need to include new headers.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1740,6 +1740,42 @@ wasm_runtime_destroy_exec_env(WASMExecEnv *exec_env)
...
+    }
+#endif
+
+#if WASM_ENABLE_AOT != 0
+    if (module_inst->module_type == Wasm_Module_AoT) {
+        return aot_copy_callstack(exec_env, buffer, length, skip_n, error_buf,
+                                  error_buf_size);
+    }
+#endif
+#endif
```
lum1n0us, line 1775: An explicit return value is required when all `if` fail.

core/iwasm/interpreter/wasm_runtime.c
lum1n0us: Do `aot_copy_callstack_xxx()` need the similar `cur_frame->function == NULL` checking?

core/iwasm/aot/aot_runtime.c
```diff
@@ -4103,6 +4104,136 @@ aot_frame_update_profile_info(WASMExecEnv *exec_env, bool alloc_frame)
...
+        buffer[count - skip_n] = record_frame;
+        cur_frame = cur_frame->prev_frame;
+        ++count;
+    }
+#else
+/*
+ * TODO: add support for standard frames when GC is enabled
+ * now it poses a risk due to variable size of the frame
+ */
+#endif
```
lum1n0us, line 4209: `return count` at the end like `aot_copy_callstack_tiny_frame()` ?

lum1n0us (changes requested):

g0djan:
Addressed

core/iwasm/aot/aot_runtime.c
lum1n0us: Even an empty implementation for GC should have a default return value.

The `return count` needs to be moved to after the `#endif` at line 4207. (Also, move `uint32 count` to line 4174 before `#if`).

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -5,6 +5,8 @@
 
 #include "wasm_runtime.h"
 #include "wasm.h"
+#include "wasm_c_api.h"
+#include "wasm_exec_env.h"
```
lum1n0us, line 9: Both headers should not be added.

lum1n0us (changes requested):
Last round, keep fighting 💪

g0djan:
@lum1n0us can you please take a look?

lum1n0us (approved):
LGTM

lum1n0us:
@xujuntwt95329 @TianlongLiang @wenyongh   Please take a look and feel free to leave your comments.

TianlongLiang (approved):
LGTM

g0djan:
Thank you!

PR #4064, opened 2025-02-06
In wasm32, fix potential conversion overflow when enlarging 65536 pages

TianlongLiang:
Fix conversion error when the total_size_new is 4G(0x100000000) and the old is 0

### DIFF
--- a/core/iwasm/common/wasm_memory.c
+++ b/core/iwasm/common/wasm_memory.c
@@ -1389,7 +1389,7 @@ wasm_enlarge_memory_internal(WASMModuleInstanceCommon *module,
     if (full_size_mmaped) {
 #ifdef BH_PLATFORM_WINDOWS
         if (!os_mem_commit(memory->memory_data_end,
-                           (uint64)(total_size_new - total_size_old),
+                           total_size_new - total_size_old,
                            MMAP_PROT_READ | MMAP_PROT_WRITE)) {
             ret = false;
             goto return_func;
@@ -1397,12 +1397,12 @@ wasm_enlarge_memory_internal(WASMModuleInstanceCommon *module,
 #endif
 
         if (os_mprotect(memory->memory_data_end,
-                        (uint64)(total_size_new - total_size_old),
+                        total_size_new - total_size_old,
                         MMAP_PROT_READ | MMAP_PROT_WRITE)
             != 0) {
 #ifdef BH_PLATFORM_WINDOWS
             os_mem_decommit(memory->memory_data_end,
-                            (uint64)(total_size_new - total_size_old));
+                            total_size_new - total_size_old);
 #endif
             ret = false;
             goto return_func;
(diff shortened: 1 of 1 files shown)


core/iwasm/common/wasm_memory.c
```diff
@@ -1389,20 +1389,20 @@ wasm_enlarge_memory_internal(WASMModuleInstanceCommon *module,
     if (full_size_mmaped) {
 #ifdef BH_PLATFORM_WINDOWS
         if (!os_mem_commit(memory->memory_data_end,
-                           (mem_offset_t)(total_size_new - total_size_old),
+                           (uint64)(total_size_new - total_size_old),
```
yamt, line 1392: these casts don't seem necessary to me.

TianlongLiang: It's about a corner case:
```webassembly
(module
  (memory (;0;) 0)
  (export "main" (func 0))
  (func (;0;) (result)
    block (result) ;; label = @1
      i32.const 65536
      memory.grow
      i64.load align=2
      drop
    end
  )
)
```
When in wasm32, the mem_offset_t is uint32, and total_size_new - total_size_old will be 0x100000000 and overflow as 0. Resulting in incorrectly throwing memory OOB exception for hardware boundary check.

yamt: i meant, because total_size_old and total_size_new are uint64, there are no point to cast it to uint64 again.
you can just remove the mem_offset_t casts.

TianlongLiang: You are right! Good catch

lum1n0us (approved):
LGTM

yamt (approved):
lgtm

lum1n0us:
But, with the declaration of `int os_mprotect(void *addr, size_t size, int prot)` ?

lum1n0us (commented):
Reconsidering the approach if `size_t` is a 4-byte type.

yamt:
> But, with the declaration of `int os_mprotect(void *addr, size_t size, int prot)` ?

it's a valid concern. can you file a bug?

TianlongLiang:
> > But, with the declaration of `int os_mprotect(void *addr, size_t size, int prot)` ?
> 
> it's a valid concern. can you file a bug?

In this pr, os_mprotect is only used for hardware boundary checks which are only enabled on 64-bit platforms, I think it's fine to leave it like this.

yamt:
> > > But, with the declaration of `int os_mprotect(void *addr, size_t size, int prot)` ?
> > 
> > 
> > it's a valid concern. can you file a bug?
> 
> In this pr, os_mprotect is only used for hardware boundary checks which are only enabled on 64-bit platforms, I think it's fine to leave it like this.

i think you are right that it's ok for this pr.

otoh, size_t os_mmap and friends take in general seems like a problem for eg. wasm_allocate_linear_memory.

PR #4067, opened 2025-02-06
Make sure that touching pages will not fail due to misconfigured protection

loganek:
When threads are initialized with `wasm_runtime_init_thread_env()`, the first few pages (configured using `STACK_OVERFLOW_CHECK_GUARD_PAGE_COUNT`) are configured to be not readable/writable (for the stack protection). The protection is then removed in `destroy_stack_guard_pages()`, which is automatically called when `os_thread_wrapper` is used for creating a new thread.
However, the wrapper might not always be used (and in fact, is not in some of the examples like [here](https://github.com/bytecodealliance/wasm-micro-runtime/tree/main/samples/terminate/src/main.c)) when the thread is created outside of WAMR. To avoid crashes in `touch_pages` (caused by trying to write to protected memory), we'll mark the whole stack readable/writable before touching the pages.

### DIFF
--- a/core/shared/platform/common/posix/posix_thread.c
+++ b/core/shared/platform/common/posix/posix_thread.c
@@ -524,8 +524,20 @@ static uint32
 touch_pages(uint8 *stack_min_addr, uint32 page_size)
 {
     uint8 sum = 0;
+    bool protect_configured = false;
     while (1) {
         volatile uint8 *touch_addr = (volatile uint8 *)os_alloca(page_size / 2);
+
+        if (!protect_configured) {
+            if (os_mprotect(stack_min_addr,
+                            touch_addr + page_size / 2 - stack_min_addr,
+                            MMAP_PROT_READ | MMAP_PROT_WRITE)
+                != 0) {
+                return 0;
+            }
+            protect_configured = true;
+        }
+
         if (touch_addr < stack_min_addr + page_size) {
             sum += *(stack_min_addr + page_size - 1);
             break;


lum1n0us:
> caused by trying to write to protected memory

Not quite following which line is intended to write to protected pages.

if using the code mentioned above, it seems that `init_stack_guard_pages()` are paired with `destroy_stack_guard_pages()`.

```c

static void *
runner_with_sigleton_exec_env(void *vp)
{
    wasm_module_inst_t inst = vp;
    bool ok = wasm_runtime_init_thread_env();  //->init_stack_guard_pages()
    assert(ok);
    wasm_application_execute_main(inst, 0, NULL);
    wasm_runtime_destroy_thread_env();  //->destroy_stack_guard_pages()
    return inst;
}
```

loganek:
oh you're right, in the example above indeed there's a call to `destroy_stack_guard_pages`, sorry I missed it. Looks like making both calls is an explicit requirement then, and it cannot be relaxed?

lum1n0us:
Perhaps we could either have 3rd parties use an Amazon-defined `start_routine()` and employ a callback to execute 3rd party code, or use something akin to [`pthread_cleanup_xxx()`](https://man7.org/linux/man-pages/man3/pthread_cleanup_push.3.html)?

lum1n0us:
@loganek  If there are no additional comments, I plan to close the PR.

loganek:
OK we can close the PR for now.

PR #4070, opened 2025-02-08
Unit test and sample update for shared heap enhancement

TianlongLiang:


samples/shared-heap/wasm-apps/CMakeLists.txt
```diff
@@ -33,6 +33,8 @@ set (CMAKE_EXE_LINKER_FLAGS
       -Wl,--export=my_shared_heap_malloc           \
       -Wl,--export=my_shared_heap_free             \
       -Wl,--export=print_buf                       \
+      -Wl,--export=produce_str                     \
+      -Wl,--export=consume_str                     \
```
lum1n0us, line 37: Just FYI, `__attribute__((export_name("XXX")))` can replace `--export=XXX`, especially for a function.

samples/shared-heap/README.md
```diff
@@ -0,0 +1,50 @@
...
+./shared_heap_chain_test
+```
+
+## How to use shared heap
+
+The shared heap is an advanced feature in WAMR that gives the user flexibility to share data between multiple WASM instances(it will be the same address mapping for different WASM instance). But it also heavily relies on the user's implementation to manage the shared data correctly. The following are some takeaway points to help the user use the shared heap correctly.
+
+### Create and manage shared heap
+
+You can create a shared heap by calling the `wasm_runtime_create_shared_heap(SharedHeapInitArgs *init_args)` API. And based on the init_args, you can create a shared heap in two ways:
```
lum1n0us, line 36: You can create a shared heap by calling the `wasm_runtime_create_shared_heap(SharedHeapInitArgs *init_args)` API. And based on the `init_args`, you can create a shared heap in two ways:

lum1n0us, line 36:
```suggestion
You can create a shared heap by calling the `wasm_runtime_create_shared_heap(SharedHeapInitArgs *init_args)` API. And based on the `init_args`, you can create a shared heap in two ways:
```

samples/shared-heap/README.md
```diff
@@ -0,0 +1,50 @@
...
+For the shared heap chain sample. It chains a pre-allocated heap and a normal shared heap to one chain(linked list) as a whole and attaches/detaches all together, and pass the WASM address directly between two WASM instances. Use the following commands to run the sample:
+
+```bash
+cd build
+./shared_heap_chain_test
+```
+
+## How to use shared heap
+
+The shared heap is an advanced feature in WAMR that gives the user flexibility to share data between multiple WASM instances(it will be the same address mapping for different WASM instance). But it also heavily relies on the user's implementation to manage the shared data correctly. The following are some takeaway points to help the user use the shared heap correctly.
```
lum1n0us, line 32:
```suggestion
The shared heap is an advanced feature in WAMR that gives the user flexibility to share data between multiple WASM instances(it will be the same address mapping for different WASM instance) or between WebAssembly and the host without incurring any copy overhead. The shared heap can be regarded as an extension of the WebAssembly linear memory. But it also heavily relies on the user's implementation to manage the shared data correctly. The following are some takeaway points to help the user use the shared heap correctly.
```

samples/shared-heap/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Create and manage shared heap
+
+You can create a shared heap by calling the `wasm_runtime_create_shared_heap(SharedHeapInitArgs *init_args)` API. And based on the init_args, you can create a shared heap in two ways:
+
+1. WAMR managed shared heap: when only `init_args.size` is given and `init_args.pre_allocated_addr` stays as NULL, WAMR will allocate a shared heap with the given size. The shared heap will be managed by WAMR, the wasm app or WAMR runtime can dynamically manage memory from it by calling `wasm_runtime_shared_heap_malloc()` and `wasm_runtime_shared_heap_free()` on demand. Only the memory allocated from the shared heap is valid and can be shared between different WASM instances, not the unallocated part of shared heap memory. And it will be automatically freed when runtime is destroyed(when `wasm_runtime_destroy()` is called).
```
lum1n0us, line 38:
```suggestion
1. WAMR managed shared heap: when only `init_args.size` is given and `init_args.pre_allocated_addr` stays as NULL, WAMR will allocate a shared heap(not from the linear memory) with the given size. The shared heap will be managed by WAMR, the wasm app or host(WAMR users) can dynamically manage memory from it by calling `wasm_runtime_shared_heap_malloc()` and `wasm_runtime_shared_heap_free()` on demand. Only the memory allocated from the shared heap is valid and can be shared between different WASM instances, not the unallocated part of shared heap memory. And it will be automatically freed when runtime is destroyed(when `wasm_runtime_destroy()` is called).
```

samples/shared-heap/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+2. Preallocated shared heap: the user can also use a pre-allocated memory(it can be allocated from the system heap, or is a static global buffer, the correctness of its accessibility and size needs to be ensured by the user) as a shared heap by giving `init_args.pre_allocated_addr` and `init_args.size`. This type of shared heap is used as a one big chunk, and it can to share data directly. And the pre-allocated memory will relies on user to manage its life cycle.
```
lum1n0us, line 40:
```suggestion
2. Preallocated shared heap: the user can also use a pre-allocated memory(it can be allocated from the system heap, or is a static global buffer, the correctness of its accessibility and size needs to be ensured by the user) as a shared heap by giving `init_args.pre_allocated_addr` and `init_args.size`. This kind of shared heap serves as an area for data exchange, primarily between the host and WebAssembly. Any data within this area can be directly accessed by both sides (assuming the layout of the data structure is known). For instance, the host can store large structured variables in this space, allowing the WebAssembly application to operate on them without the need for copying. And the pre-allocated memory will relies on user to manage its life cycle.
```

samples/shared-heap/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+After creation, the shared heap can be attached to a WASM instance by calling `wasm_runtime_attach_shared_heap(wasm_module_inst_t module_inst, wasm_shared_heap_t shared_heap)`. And it can be detached by calling `wasm_runtime_detach_shared_heap(wasm_module_inst_t module_inst)`. So that the data sharing can only happen between the WASM instances that have the same shared heap attached, complete by user's choice.
```
lum1n0us, line 42:
```suggestion
After creation, the shared heap can be attached to a WASM instance(an additional segment appended to the end of the linear memory) by calling `wasm_runtime_attach_shared_heap(wasm_module_inst_t module_inst, wasm_shared_heap_t shared_heap)`. And it can be detached by calling `wasm_runtime_detach_shared_heap(wasm_module_inst_t module_inst)`. So that the data sharing can only happen between the WASM instances that have the same shared heap attached, complete by user's choice.
```

PR #4071, opened 2025-02-08
Prevent mmap size overflow on 32 bit platform for memory.grow

TianlongLiang:
As discussed in https://github.com/bytecodealliance/wasm-micro-runtime/pull/4064, on 32-bit platform, memory.grow 65536 can cause incorrect conversion when `wasm_mremap_linear_memory`  pass the u64 size to mmap as size_t

### DIFF
--- a/core/iwasm/common/wasm_memory.c
+++ b/core/iwasm/common/wasm_memory.c
@@ -1253,6 +1253,12 @@ wasm_mremap_linear_memory(void *mapped_mem, uint64 old_size, uint64 new_size,
     bh_assert(new_size > 0);
     bh_assert(new_size > old_size);
 
+#if UINTPTR_MAX == UINT32_MAX
+    if (new_size == 4 * (uint64)BH_GB) {
+        return NULL;
+    }
+#endif
+
     if (mapped_mem) {
         new_mem = os_mremap(mapped_mem, old_size, new_size);
     }


core/iwasm/common/wasm_memory.c
```diff
@@ -1253,6 +1253,12 @@ wasm_mremap_linear_memory(void *mapped_mem, uint64 old_size, uint64 new_size,
     bh_assert(new_size > 0);
     bh_assert(new_size > old_size);
 
+#if UINTPTR_MAX == UINT32_MAX
+    if (new_size == 4 * (uint64)BH_GB) {
```
wenyongh, line 1257: How about `new_size > UINT32_MAX` since the larger size should also be blocked?

lum1n0us: Not quite sure about this modification. Here are some knowns:

- In wasm32, [the memory limit value range is 2**16](https://webassembly.github.io/spec/core/valid/types.html#memory-types)
- In wasm64, [the memory limit value range is either 2**16 or 2**48, depends on address type](https://github.com/WebAssembly/memory64/blob/main/proposals/memory64/Overview.md#validation)

So, maybe we should prevent wrong casting of 4 bytes `size_t` at Wasm-semantic level instead of 32bit/64bit platform level only?

wenyongh: You mean that when the platform is 64-bit and the wasm memory is 32-bit, we should also block the mmap with `size >= 4GB`? Or we should block such mmap for memory32, no matter the platform is 64-bit or 32-bit. Sounds reasonable.

lum1n0us: maybe something like:

```
#if WASM32
  // new_size should not be greater than 2**16
#endif

#if WASM64
  // new_size should not be greater than 2**48
#endif
```

Due to [the safeguards in config_common.cmake](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/build-scripts/config_common.cmake#L322-#L329), there's no need to be concerned about enabling memory64 on a 32-bit platform.

TianlongLiang: I added this specific check to handle some corner cases, consider the following scenario:
```wast
(module
  (memory (;0;) 65536 65536)
  (export "main" (func 0))
  (func (;0;) (result)
    block (result) ;; label = @1
      unreachable
    end
  )
)
```
If this wasm32 runs on 32-bit platform, it should give something like: `WASM module instantiates failed: allocate linear memory failed`, similar when the page count is 65500 or 65535, for the system can't mmap/mprotect such size memory. But currently the `new_size` will overflow when passed as uint32, and it can instantiate successfully, so I only check for 4G, the rest cases will be handled by mmap/mprotect and simply report error if mmap/mprotect size exceeds what the system can give. 
For wasm32 on a 64-bit platform, the system can mmap/mprotect 65536 pages, no need to add a check, the current way is fine.

lum1n0us:
> If this wasm32 runs on 32-bit platform, it should give something like: WASM module instantiates failed: allocate linear memory failed, similar when the page count is 65500 or 65535, for the system can't mmap/mprotect such size memory. 

We may need to ensure that the test at [`(module (memory 0 65536))`](https://github.com/WebAssembly/spec/blob/main/test/core/memory.wast#L8) runs smoothly without any exceptions on both 32-bit and 64-bit platforms, provided there are sufficient hardware resources. If the memory limit is not compatible with the hardware conditions, we should inform the user. Technically, technologies like Physical Address Extension (PAE) enable a 32-bit system to access more than 4GB of physical memory.

By the way, we have defined `#define DEFAULT_MEM64_MAX_PAGES UINT32_MAX`, which is less than what the proposal specifies `2**48`. Was this done intentionally?

TianlongLiang:
>  If the memory limit is not compatible with the hardware conditions, we should inform the user

mmap failure will log a corresponding error: `mmap failed with errno: 12, hint: (nil), size: 4294705152, prot: 0, flags: 34` but I observed that in [some cases](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/shared/platform/common/posix/posix_memmap.c#L66), it might fail silently.

I plan to cover all the mmap failures and log such errors, what do you think?

> #define DEFAULT_MEM64_MAX_PAGES UINT32_MAX, which is less than what the proposal specifies 2**48

Yes, this is for ABI compatibility, so that in AOT don't need to change the type of page_count, it can still stay as u32

lum1n0us:
https://github.com/WebAssembly/spec/issues/1869

It appears that for a 32-bit platform, it is acceptable to encounter a failure when attempting to instantiate a memory with 65536 pages.

lum1n0us:
> I plan to cover all the mmap failures and log such errors, what do you think?

Please log all failures.

core/iwasm/common/wasm_memory.c
lum1n0us:
```suggestion
    if (new_size >= 4 * (uint64)BH_GB) {
```

lum1n0us (approved):
LGTM

core/shared/platform/common/posix/posix_memmap.c
```diff
@@ -281,6 +285,10 @@ os_mprotect(void *addr, size_t size, int prot)
     if (prot & MMAP_PROT_EXEC)
         map_prot |= PROT_EXEC;
 
+    if (mprotect(addr, request_size, map_prot) == -1) {
```
wenyongh, line 288: L292 also calls mprotect again, seems we should change it to `return 0`?
And here is it better to change `== -` to `! = 0`?

core/shared/platform/common/posix/posix_memmap.c
```diff
@@ -281,6 +285,10 @@ os_mprotect(void *addr, size_t size, int prot)
...
+        printf("mprotect failed\n");
```
wenyongh, line 289: had better use os_printf

core/shared/platform/common/posix/posix_memmap.c
TianlongLiang: My bad, `os_mprotect` shouldn't be modified, I forgot to delete those lines

wenyongh (approved):
LGTM

loganek (approved):
LGTM

PR #4072, opened 2025-02-10
Allow finer control of start function on instantiation

sjamesr:
Before this change, wasm_runtime_instantiate would execute the start function unconditionally, correctly implementing the spec. However, if there is an infinite loop in the start function, there was no way to interrupt execution.

This change introduces a parameter to InstantiationArgs indicating whether the start function should be immediately invoked. If not, wasm_runtime_instantiate returns a module instance that is initialized except for the start function.

This change adds a wasm_runtime_instantiate_run_start_func function which performs this last instantiation step. The user may prepare a timeout in advance of calling this to guard against expensive/infinite loops.

This change also demonstrates a possible technique for doing this in iwasm.c.


yamt:
> Ideally, the termination should originate from the runtime itself, whether it's the interpreter, JIT, or AOT, rather than from iwasm. I'm considering a solution that's more intrinsic to the runtime, such as using a counter (which might be less precise) to estimate the remaining fuel or time left after each opcode or `block`(maybe every round of a loop).

please note that the timeout thing is NOT the sole motivation of wasm_runtime_terminate.
what we (well, i) wanted to have was a functionality to terminate instances on async events in general. (eg. on user request)
i just added the timeout option as a demo of the api. it was not the main motivation of the api.

fuel-like approaches would work for my use cases only if it allows to resume terminated instance. it likely makes host functions very tricky. (i have implemented a similar approach in the other runitme: https://github.com/yamt/toywasm/blob/master/doc/check_interrupt.md. my gut feeling is that it isn't suitable for wamr.)

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -2333,6 +2333,25 @@ wasm_set_running_mode(WASMModuleInstance *module_inst, RunningMode running_mode)
     return set_running_mode(module_inst, running_mode, false);
 }
 
+bool
+wasm_runtime_instantiate_run_start_func(wasm_module_inst_t module_inst,
```
yamt, line 2337: this name is a bit confusing because it doesn't run wasi's entry point, `_start`.
how about wasm_runtime_execute_instantiate_functions?

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -2333,6 +2333,25 @@ wasm_set_running_mode(WASMModuleInstance *module_inst, RunningMode running_mode)
...
+                                        wasm_exec_env_t exec_env,
+                                        char *error_buf,
+                                        uint32_t error_buf_size)
+{
+    const bool is_sub_inst = parent != NULL;
+    if (!execute_post_instantiate_functions((WASMModuleInstance *)module_inst,
+                                            is_sub_inst, exec_env)) {
+        set_error_buf(error_buf, error_buf_size,
+                      ((WASMModuleInstance *)module_inst)->cur_exception);
+        wasm_runtime_deinstantiate(module_inst);
```
yamt, line 2348: as wasm_runtime_instantiate succeeded at this point,
isn't it simpler to leave wasm_runtime_deinstantiate to api users?

PR #4075, opened 2025-02-10
[gc] Subtyping fix

vouillon:
The code was not handling the case where the right-hand side is a non-nullable ref type with a common heap type. For instance:
```
(ref eq) <: (ref any)
```

### DIFF
--- a/core/iwasm/common/gc/gc_type.c
+++ b/core/iwasm/common/gc/gc_type.c
@@ -1145,6 +1145,15 @@ wasm_reftype_is_subtype_of(uint8 type1, const WASMRefType *ref_type1,
                 return true;
             else {
                 int32 heap_type = ref_type1->ref_ht_common.heap_type;
+                // We dont care whether type2 is nullable or not. So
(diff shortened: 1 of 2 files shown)


xujuntwt95329 (approved):
LGTM

core/iwasm/common/gc/gc_type.c
```diff
@@ -1188,19 +1197,15 @@ wasm_reftype_is_subtype_of(uint8 type1, const WASMRefType *ref_type1,
                 }
 #endif
                 else if (heap_type == HEAP_TYPE_NONE) {
-                    /* (ref none) */
-                    /* TODO */
-                    bh_assert(0);
+                    return wasm_is_reftype_supers_of_none(type2, NULL, types,
+                                              type_count);
```
xujuntwt95329, line 1201: Coding style: should indent to the column of the first parameter in above line. You can use clang-format to auto format your patch

lum1n0us (approved):
LGTM

PR #4079, opened 2025-02-12
Unit test:type matching issue and code redundancy

yangkun27:
This pull request contains some changes about unit test

- In unit test wasm_vm.cc, it seems better to define the type of argv as uint32 to match the parameter of the function wasm_runtime_call_wasm.
- In test interpreter_test.cc, `ret =` is redundant.


no1wudi (approved):

lum1n0us (approved):
LGTM

PR #4080, opened 2025-02-12
Add a conditional check for the macro __STDC_VERSION__

yangkun27:
In some cases, if __cplusplus is defined, it may cause the warning "__STDC_VERSION__" not defined, it's better to check whether the macro __STDC_VERSION__  is defined firstly.


lum1n0us (approved):
LGTM

no1wudi (approved):

PR #4090, opened 2025-02-17
Fix spec tests when WASM_CPU_SUPPORTS_UNALIGNED_ADDR_ACCESS is 0

jammar1:
This fixes the SIMD spec tests when WASM_CPU_SUPPORTS_UNALIGNED_ADDR_ACCESS is set to 0.
After this is merged I will raise PR to enable CI actions with SIMD interpreter support.

### DIFF
--- a/core/iwasm/interpreter/wasm_interp_fast.c
+++ b/core/iwasm/interpreter/wasm_interp_fast.c
@@ -5843,60 +5843,58 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                         PUT_V128_TO_ADDR(frame_lp + addr_ret, LOAD_V128(maddr));
                         break;
                     }
-#define SIMD_LOAD_OP(simde_func, element_size, num_elements) \
-    do {                                                     \
-        uint32 offset, addr;                                 \
-        offset = read_uint32(frame_ip);                      \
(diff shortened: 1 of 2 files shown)


core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -5843,60 +5843,60 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
-        PUT_V128_TO_ADDR(frame_lp + addr_ret, result);       \
-                                                             \
+#define SIMD_LOAD_OP(simde_func, width)                \
+    do {                                               \
+        uint32 offset, addr;                           \
+        offset = read_uint32(frame_ip);                \
+        addr = GET_OPERAND(uint32, I32, 0);            \
+        frame_ip += 2;                                 \
+        addr_ret = GET_OFFSET();                       \
+        CHECK_MEMORY_OVERFLOW(width / 8);              \
```
loganek, line 5853: is that correct? e.g. for `SIMD_v128_load8x8_s` we'll only check if single byte can fit into the address? Shouldn't we check for 16bytes-slot?

loganek (approved):

loganek:
LGTM

Issue #4088, opened 2025-02-15
build wasm file to AOT, report error Unsupported opcode: 0x8 WASM_OP_THROW


ericCHensssssssss:
when i build wasm file to aot, it report errors. it seems not support WASM_OP_THROW

 ./wamrc --enable-gc -o test.aot --target=armv7 test.wasm

Create AoT compiler with:
  target:        armv7
  target cpu:
  target triple: armv7-pc-linux-gnu
  cpu features:
  opt level:     3
  size level:    3
  output format: AoT file
Unsupported opcode: 0x8

yamt:
it's expected because currently exception-handling is only implemented for the classic interpreter.

ericCHensssssssss:
Thanks for your quick response，i compile Kotlin to WebAssembly ，do you plan to support the exception-handling  feature in AOT，

lum1n0us:
I haven't heard of any further plans from the community regarding exception handling in AOT. This is partly because the exception-handling proposal underwent significant changes at the end of last year, and the version implemented by WAMR is now considered "legacy exception handling." This makes it difficult to extend the current interpreter's solution to AOT mode.

ericCHensssssssss:
i want run Kotlin program  at embbed devices，so aot is aot is really  important for me，if not support aot，it will slow use classic interpreter

lum1n0us:
Is there a way, like some command line options, for the Kotlin toolchain to compile .kt files to .wasm without utilizing exception handling opcodes?

ericCHensssssssss:
ok. goog idea, it can change the kotlin compiler chain . 
If i remove exception handling . is AOT support GC feature?

ericCHensssssssss:
@lum1n0us   when i build wasm from Kotlin，use the follow option              "-Xwasm-use-traps-instead-of-exceptions",
    wasmrc can compile the wasm to aot file success。
    how to run the aot，  
     when i run     ./iwasm  Sys.aot   
i meet the error：
 warning: failed to allocate memory for gc object
AOT module instantiate failed: create struct object failed

ericCHensssssssss:
@lum1n0us  Thanks. it worked when i removed the EH,  

and how can i run the aot. is the same as wasm file 

./iwasm -f test Sys.aot 12

when i run it. error produced

unhandled SIGSEGV, si_addr: 0x53
zsh: abort      ./iwasm -f test  Sys.aot 12

[...]

PR #4094, opened 2025-02-19
fix: when load aot init expr,no type_idx set.

chenliuyang1989:
This is case error when run aot ,the type is default 0

     error at File gc_object.c line 91
     bh_assert(rtt_type->type_flag == WASM_TYPE_STRUCT;


lum1n0us:
I believe the PR looks good. However, I'm quite interested in the failure. Would you mind sharing details about the test case?

chenliuyang1989:
my colleage raised one issue，

this  link：https://github.com/bytecodealliance/wasm-micro-runtime/issues/4088

this is the line which produce error

image

lum1n0us (approved):
LGTM

PR #4095, opened 2025-02-19
fix(aot_emit_aot_file): prevent buffer emission for zero byte_count

lum1n0us:
if using a debug building of wamrc to run spec test. there will be: core/iwasm/compilation/aot_emit_aot_file.c:1794:13: runtime error: null pointer passed as argument 2, which is declared to never be null

```wat
(module binary
  "\00asm" "\01\00\00\00"
  "\05\03\01"                          ;; Memory section with 1 entry
  "\00\00"                             ;; no max, minimum 0
  "\0b\07\01"                          ;; Data section with 1 entry
  "\80\00"                             ;; Memory index 0, encoded with 2 bytes
  "\41\00\0b\00"                       ;; (i32.const 0) with contents ""
)
```


TianlongLiang (approved):
LGTM

no1wudi (approved):

PR #4099, opened 2025-02-23
wasm_loader allocates more spaces for elements

### LINKED ISSUE #4096: [gc]  array as a field in struct object is not supported in constant init expr
### Steps to reproduce

1. Unpack zip [globals.zip](https://github.com/user-attachments/files/18892034/globals.zip)

2. Build `iwasm` with the following command:
```
cmake .. -DWAMR_BUILD_GC=1 -DWAMR_BUILD_EXCE_HANDLING=1 -DWAMR_BUILD_TAIL_CALL=1 && make
```

3.  Run `./iwasm globals.wasm`

### Expected behavior

It should print something like:
```
line 1
line 2
line 3
```

### Actual behavior

I get the following output:
```
WASM module instantiate failed: array as a field in struct object is not supported in constant init expr
unhandled SIGSEGV, si_addr: 0x21
Abandon (core dumped)
```

lum1n0us:
fix #4096

### DIFF
--- a/core/iwasm/interpreter/wasm_loader.c
+++ b/core/iwasm/interpreter/wasm_loader.c
@@ -1084,8 +1084,9 @@ load_init_expr(WASMModule *module, const uint8 **p_buf, const uint8 *buf_end,
                                     goto fail;
                                 }
 
-                                size = sizeof(WASMArrayNewInitValues)
-                                       + sizeof(WASMValue) * len_val.i32;
+                                size =
+                                    sizeof(WASMArrayNewInitValues)
+                                    + sizeof(WASMValue) * (uint64)len_val.i32;
                                 if (!(array_init_values = loader_malloc(
                                           size, error_buf, error_buf_size))) {
                                     goto fail;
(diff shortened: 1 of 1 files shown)


core/iwasm/interpreter/wasm_loader.c
```diff
@@ -1073,23 +1074,24 @@ load_init_expr(WASMModule *module, const uint8 **p_buf, const uint8 *buf_end,
...
                                         &const_expr_ctx, NULL, VALUE_TYPE_I32,
                                         NULL, NULL, &len_val, error_buf,
                                         error_buf_size)) {
-                                    destroy_init_expr_data_recursive(
-                                        module, array_init_values);
                                     goto fail;
                                 }
+
+                                size = sizeof(WASMArrayNewInitValues)
+                                       + sizeof(WASMValue) * len_val.i32;
```
wenyongh, line 1088: had better change `len_val.i32` to `(uint64)len_val.i32` to avoid possible integer overflow on 32-bit platform? Since in 32-bit, sizeof(..) is 32 bit.

wenyongh (approved):
LGTM with minor issue

TianlongLiang (approved):
LGTM

vouillon:
Thanks! But this is not a complete fix to #4096. The example no longer crashes, but one still gets an error message.

PR #4105, opened 2025-02-24
fix: add dispose of the debug information builder

Jiax-cn:
Dispose the DIBuilder when destroying complation information to avoid memory leak.


lum1n0us (approved):
LGTM

TianlongLiang (approved):
LGTM

PR #4110, opened 2025-02-24
feat: add wasm_memory_get_init_page_count()

Jiax-cn:
This API provides access to the initial memory size, as a supplement to #3786.


lum1n0us:
IMU, we can always obtain the initial page value from a memory type of a module rather than from the memory of an instance.

```c
WASM_RUNTIME_API_EXTERN void
wasm_runtime_get_export_type(const wasm_module_t module, int32_t export_index,
                             wasm_export_t *export_type);

WASM_RUNTIME_API_EXTERN uint32_t
wasm_memory_type_get_init_page_count(const wasm_memory_type_t memory_type);

```

Jiax-cn:
@lum1n0us  This is useful, thank you.

PR #4112, opened 2025-02-25
Fix build issues when compiling WAMRC as a cross-compiler

peter-tatrai:
### Description  
This pull request addresses issues encountered when compiling WAMRC to be a cross-compiler—running on a PC while targeting an embedded ARM platform. The key fixes include:  

### Fixes in This PR  
- **-m32 compiler flag**  
  - The `-m32` flag was previously added unconditionally for 32-bit targets, causing issues for non-x86 platforms.  
  - Now, `-m32` is only added when targeting a 32-bit x86 platform (`X86_32`), ensuring it is only applied where supported.  

- **Improved PIC Flag Handling:**  
  - The logic for setting PIC (`-fPIC`) based on `WAMR_BUILD_PLATFORM` was overly complex.  
  - This is now handled in CMake instead of using platform-specific conditions.  

- **AOT Relocation Functions:**  
  - WAMRC previously included platform-specific headers (e.g., `<arm_neon.h>`) that are not available on a PC.  
  - Since these AOT relocation functions are only used when executing AOT-compiled code (not by WAMRC itself), dummy versions have been introduced.  
  - A more complete solution would be to split the code to avoid linking these functions altogether, but that would require more extensive changes.  

- **SIMD Compilation Issues:**  
  - WAMRC attempted to include target platform-specific SIMD headers (e.g., ARM headers) when compiling on a different host platform (e.g., x86).  
  - To address this, an option has been added to disable SIMD in WAMRC when cross-compiling.


calvin2021y:
hi @peter-tatrai 

`<arm_neon.h>` is required at runtime ?  if so maybe allow user set sysroot and triple  by args will fix the problem.

peter-tatrai:
The header file is required at compile-time, but this is not about cross-compiling wamrc—rather, it's about building a cross-compiler. In this scenario, we have both a host and a target architecture (e.g., host: x86_64, target: armv7). However, the CMake files currently use a single variable, WAMR_BUILD_PLATFORM, for both, even though it should only define the target platform. This causes issues, such as the one mentioned above.

With these commits and `-DWAMR_BUILD_INVOKE_NATIVE_GENERAL=1 -DWAMR_BUILD_SIMD=0 -DWAMR_BUILD_TARGET=ARMv7`, it is possible to build wamrc as a runnable binary on a PC while cross-compiling AOT modules for armv7.

wamr-compiler/CMakeLists.txt
```diff
@@ -275,6 +271,8 @@ else ()
   message ("-- Lib wasi-threads disabled")
 endif ()
 
+set (WAMR_AOT_COMPILER 1)
```
lum1n0us, line 274: a similar naming pattern `WAMR_BUILD_WAMR_COMPILER`

wamr-compiler/CMakeLists.txt
```diff
@@ -132,21 +138,11 @@ endif ()
...
-      set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC")
-      set (CMAKE_SHARED_LIBRARY_LINK_C_FLAGS "${CMAKE_SHARED_LIBRARY_LINK_C_FLAGS} -fPIC")
-      set (CMAKE_SHARED_LIBRARY_LINK_CXX_FLAGS "${CMAKE_SHARED_LIBRARY_LINK_CXX_FLAGS} -fPIC")
-    endif ()
-  else ()
-    add_definitions (-m32)
-    set (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -m32")
-    set (CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -m32")
-  endif ()
+# Add -m32 flag if compiling on 64-bit system for 32-bit x86 target
```
lum1n0us, line 141: I noticed that the information regarding *Add -fPIC flag if build as 64-bit* has been deleted. Was this done on purpose?

core/iwasm/aot/arch/aot_reloc_dummy.c
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright (C) 2020 Intel Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+
+#include "aot_reloc.h"
+#include <stdlib.h>
```
lum1n0us, line 7: better remove <stdlib.h>. *bh_platform.h* in *aot_runtime.h* will do.

wamr-compiler/CMakeLists.txt
```diff
@@ -398,4 +396,4 @@ else()
                          ${UV_A_LIBS})
 endif()
 
-install (TARGETS wamrc DESTINATION bin)
+install (TARGETS wamrc)
```
lum1n0us, line 399: Was this done on purpose? Others might need it.

wamr-compiler/CMakeLists.txt
ptatrai: Yes, it is set by cmake now. I added POSITION_INDENPENDENT_CODE property to remaining libs: vmlib and aotclib on line 377

ptatrai: It will be set if it is supported by compiler and it will use appropriate flag for that compiler. So it will be set in more cases than in the original code.

wamr-compiler/CMakeLists.txt
ptatrai: Yes, I changed it to fix issues during packaging in Yocto. Since CMake 3.14 (which is already set as the minimum required version in this CMake file), the install command automatically adds DESTINATION ${CMAKE_INSTALL_BINDIR} if not specified. Including GNUInstallDirs ensures that CMAKE_INSTALL_BINDIR is set correctly, with a fallback to bin. Since this is a cache variable, users can override it to meet specific needs, which is often necessary for packaging.

wamr-compiler/CMakeLists.txt
ptatrai: Done.

core/iwasm/aot/arch/aot_reloc_dummy.c
ptatrai: removed

lum1n0us (approved):
LGTM. 

---

@yamt @no1wudi @loganek @bnason-nf please review the PR and share your comments. ☕

TianlongLiang (approved):
LGTM

PR #4113, opened 2025-02-26
[AOT] Memory Info Restore Mechanism with Better Performance

Jiax-cn:
I found that my program runs slower in WAMR AOT mode compared to other WASM runtimes, e.g., WAVM. I compared their LLVM IRs and found that WAMR emits more load operations of memory base.

In WAMR, functions with `mem_space_unchanged` keep memory base address in `mem_base_addr` of `AOTMemInfo`, while others keep **the address of** memory base address in that field. When emit instructions like load/store, the former use the base address directly, while the later should load base address from its address at first. This reload is redundant when there is no possibility of changing memory between two consecutive load/store instructions:
```llvm ir
%mem_base0 = load ptr, ptr %mem_base_addr_offset, align 8 ; reload
; load/store on %mem_base0
%mem_base1 = load ptr, ptr %mem_base_addr_offset, align 8 ; redundant reload
; load/store on %mem_base1
```

Optimization passes won’t recognize this redundancy because the reloaded memory base is accessed within the context.

In WAVM, the base address is reloaded when the memory possibly changes, e.g., after calling another function or after `memory.grow`. This can be redundant if there are no subsequent load/store instructions, but the dead code elimination pass handles this：
```llvm ir
%mem_base0 = load ptr, ptr %mem_base_addr_offset, align 8 ; reload
; load/store on %mem_base0
; emit memory.grow
%mem_base1 = load ptr, ptr %mem_base_addr_offset, align 8 ; reload
; load/store on %mem_base1
```

## Performance

Here is a sample C++ program `substr.cc`:

```C++
#include <string>
extern "C" void TestPerformance(const std::string& s) {
    s.substr(5);
}

int main() {
    for (int i = 0; i < 10007989; i++) {
        TestPerformance("Hello World");
    }
    return 0;
}
```

Compiled with emcc (version: `3.1.59 (0e4c5994eb5b8defd38367a416d0703fd506ad81)`)

```
emcc -O3 -g2 -s EXPORTED_FUNCTIONS='["_TestPerformance"]' ./substr.cc -o substr.wasm
```

Then ran wamrc and iwasm(linux) and compared the performance:

product-mini/platforms/posix/main.c:
```c
#include <time.h>
static const void *
app_instance_main(wasm_module_inst_t module_inst)
{
    const char *exception;
    struct timespec s, e;
    clock_gettime(CLOCK_MONOTONIC, &s);
    wasm_application_execute_main(module_inst, app_argc, app_argv);
    clock_gettime(CLOCK_MONOTONIC, &e);
    printf("cost: %ld us", e.tv_sec * 1000000 + e.tv_nsec / 1000 - s.tv_sec * 1000000 - s.tv_nsec / 1000);
    exception = wasm_runtime_get_exception(module_inst);
    return exception;
}
```

### result:

commit e3dcf4f8acfe58a04d54319bef5fc3728bd612bf
```
cost: 349913 us
```

commit 3f268e5150ad2e592f02ac0d467452c38aaa6ae7
```
cost: 308787 us
```

IR(optimized) comparison:

image

### DIFF
--- a/core/iwasm/compilation/aot_compiler.c
+++ b/core/iwasm/compilation/aot_compiler.c
@@ -999,6 +999,7 @@ static bool
 aot_compile_func(AOTCompContext *comp_ctx, uint32 func_index)
 {
     AOTFuncContext *func_ctx = comp_ctx->func_ctxes[func_index];
+    WASMModule *module = comp_ctx->comp_data->wasm_module;
     LLVMValueRef func_index_ref;
     uint8 *frame_ip = func_ctx->aot_func->code, opcode, *p_f32, *p_f64;
     uint8 *frame_ip_end = frame_ip + func_ctx->aot_func->code_size;
@@ -1230,6 +1231,8 @@ aot_compile_func(AOTCompContext *comp_ctx, uint32 func_index)
                 read_leb_uint32(frame_ip, frame_ip_end, func_idx);
                 if (!aot_compile_op_call(comp_ctx, func_ctx, func_idx, false))
                     return false;
+                if (module->functions[func_index]->has_memory_operations)
+                    restore_memory_info(comp_ctx, func_ctx);
                 break;
             }
 
@@ -1250,6 +1253,8 @@ aot_compile_func(AOTCompContext *comp_ctx, uint32 func_index)
                 if (!aot_compile_op_call_indirect(comp_ctx, func_ctx, type_idx,
                                                   tbl_idx))
                     return false;
+                if (module->functions[func_index]->has_memory_operations)
+                    restore_memory_info(comp_ctx, func_ctx);
                 break;
             }
 
@@ -1420,6 +1425,8 @@ aot_compile_func(AOTCompContext *comp_ctx, uint32 func_index)
                 if (!aot_compile_op_call_ref(comp_ctx, func_ctx, type_idx,
                                              false))
                     return false;
+                if (module->functions[func_index]->has_memory_operations)
+                    restore_memory_info(comp_ctx, func_ctx);
                 break;
             }
 
@@ -2092,6 +2099,8 @@ aot_compile_func(AOTCompContext *comp_ctx, uint32 func_index)
                 read_leb_uint32(frame_ip, frame_ip_end, mem_idx);
                 if (!aot_compile_op_memory_grow(comp_ctx, func_ctx))
                     return false;
+                if (module->functions[func_index]->has_memory_operations)
+                    restore_memory_info(comp_ctx, func_ctx);
                 break;
 
             case WASM_OP_I32_CONST:
--- a/core/iwasm/compilation/aot_emit_memory.c
+++ b/core/iwasm/compilation/aot_emit_memory.c
@@ -78,9 +78,6 @@ get_memory_check_bound(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
             return NULL;
     }
 
-    if (func_ctx->mem_space_unchanged)
-        return mem_check_bound;
-
     if (!(mem_check_bound = LLVMBuildLoad2(
               comp_ctx->builder,
               (comp_ctx->pointer_size == sizeof(uint64)) ? I64_TYPE : I32_TYPE,
@@ -164,17 +161,15 @@ aot_check_memory_overflow(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
     }
 
     /* Get memory base address and memory data size */
-    if (func_ctx->mem_space_unchanged
 #if WASM_ENABLE_SHARED_MEMORY != 0
-        || is_shared_memory
-#endif
-    ) {
+    if (is_shared_memory)
         mem_base_addr = func_ctx->mem_info[0].mem_base_addr;
-    }
-    else {
+    else
+#endif
+    {
         if (!(mem_base_addr = LLVMBuildLoad2(
                   comp_ctx->builder, OPQ_PTR_TYPE,
-                  func_ctx->mem_info[0].mem_base_addr, "mem_base"))) {
+                  func_ctx->mem_info[0].mem_base_addr, "mem_base_addr"))) {
             aot_set_last_error("llvm build load failed.");
             goto fail;
         }
@@ -1015,16 +1010,11 @@ get_memory_curr_page_count(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx)
 {
     LLVMValueRef mem_size;
 
-    if (func_ctx->mem_space_unchanged) {
-        mem_size = func_ctx->mem_info[0].mem_cur_page_count_addr;
-    }
-    else {
-        if (!(mem_size = LLVMBuildLoad2(
-                  comp_ctx->builder, I32_TYPE,
-                  func_ctx->mem_info[0].mem_cur_page_count_addr, "mem_size"))) {
-            aot_set_last_error("llvm build load failed.");
-            goto fail;
-        }
+    if (!(mem_size = LLVMBuildLoad2(
+              comp_ctx->builder, I32_TYPE,
+              func_ctx->mem_info[0].mem_cur_page_count, "mem_size"))) {
+        aot_set_last_error("llvm build load failed.");
+        goto fail;
     }
 
     return LLVMBuildIntCast(comp_ctx->builder, mem_size,
@@ -1165,16 +1155,14 @@ check_bulk_memory_overflow(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
 #if WASM_ENABLE_SHARED_MEMORY != 0
     bool is_shared_memory = comp_ctx->comp_data->memories[0].flags & 0x02;
 
-    if (func_ctx->mem_space_unchanged || is_shared_memory) {
-#else
-    if (func_ctx->mem_space_unchanged) {
-#endif
+    if (is_shared_memory)
         mem_base_addr = func_ctx->mem_info[0].mem_base_addr;
-    }
-    else {
+    else 
+#endif
+    {
         if (!(mem_base_addr = LLVMBuildLoad2(
                   comp_ctx->builder, OPQ_PTR_TYPE,
-                  func_ctx->mem_info[0].mem_base_addr, "mem_base"))) {
+                  func_ctx->mem_info[0].mem_base_addr, "mem_base_addr"))) {
             aot_set_last_error("llvm build load failed.");
             goto fail;
         }
@@ -1206,16 +1194,11 @@ check_bulk_memory_overflow(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
         }
     }
 
-    if (func_ctx->mem_space_unchanged) {
-        mem_size = func_ctx->mem_info[0].mem_data_size_addr;
-    }
-    else {
-        if (!(mem_size = LLVMBuildLoad2(
-                  comp_ctx->builder, I64_TYPE,
-                  func_ctx->mem_info[0].mem_data_size_addr, "mem_size"))) {
-            aot_set_last_error("llvm build load failed.");
-            goto fail;
-        }
+    if (!(mem_size = LLVMBuildLoad2(
+              comp_ctx->builder, I64_TYPE,
+              func_ctx->mem_info[0].mem_data_size, "mem_size"))) {
+        aot_set_last_error("llvm build load failed.");
+        goto fail;
     }
 
     ADD_BASIC_BLOCK(check_succ, "check_succ");
--- a/core/iwasm/compilation/aot_llvm.c
+++ b/core/iwasm/compilation/aot_llvm.c
@@ -1187,22 +1187,63 @@ create_local_variables(const AOTCompData *comp_data,
 }
 
 static bool
+restore_from_addr(const AOTCompContext *comp_ctx, const char *name,
+                  LLVMTypeRef type, LLVMValueRef src, LLVMValueRef dest) {
+    LLVMValueRef item;
+    if (!(item = LLVMBuildLoad2(comp_ctx->builder, type, src, name))) {
+        aot_set_last_error("llvm build load failed");
+        return false;
+    }
+    if (!LLVMBuildStore(comp_ctx->builder, item, dest)) {
+        aot_set_last_error("llvm build store failed");
+        return false;
+    }
+    return true;
+}
+
+bool
+restore_memory_info(const AOTCompContext *comp_ctx, AOTFuncContext *func_ctx) {
+    LLVMTypeRef ptr_type = (comp_ctx->pointer_size == sizeof(uint64)) ? I64_TYPE
+                                                                      : I32_TYPE;
+    if (!restore_from_addr(comp_ctx, "mem_base_addr", OPQ_PTR_TYPE,
+                           func_ctx->mem_info[0].mem_base_addr_addr,
+                           func_ctx->mem_info[0].mem_base_addr)
+        || !restore_from_addr(comp_ctx, "mem_data_size", I64_TYPE,
+                              func_ctx->mem_info[0].mem_data_size_addr,
+                              func_ctx->mem_info[0].mem_data_size)
+        || !restore_from_addr(comp_ctx, "mem_cur_page_count", I32_TYPE,
+                              func_ctx->mem_info[0].mem_cur_page_count_addr,
+                              func_ctx->mem_info[0].mem_cur_page_count)
+        || !restore_from_addr(comp_ctx, "mem_bound_check_1byte", ptr_type,
+                              func_ctx->mem_info[0].mem_bound_check_1byte_addr,
+                              func_ctx->mem_info[0].mem_bound_check_1byte)
+        || !restore_from_addr(comp_ctx, "mem_bound_check_2bytes", ptr_type,
+                              func_ctx->mem_info[0].mem_bound_check_2bytes_addr,
+                              func_ctx->mem_info[0].mem_bound_check_2bytes)
+        || !restore_from_addr(comp_ctx, "mem_bound_check_4bytes", ptr_type,
+                              func_ctx->mem_info[0].mem_bound_check_4bytes_addr,
+                              func_ctx->mem_info[0].mem_bound_check_4bytes)
+        || !restore_from_addr(comp_ctx, "mem_bound_check_8bytes", ptr_type,
+                              func_ctx->mem_info[0].mem_bound_check_8bytes_addr,
+                              func_ctx->mem_info[0].mem_bound_check_8bytes)
+        || !restore_from_addr(comp_ctx, "mem_bound_check_16bytes", ptr_type,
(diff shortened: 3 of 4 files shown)


lum1n0us:
Thank you for the keen observation and the intriguing analysis of the root cause.

IIUC, the rationale behind storing `mem_base_addr` in `AOTMemInfo` is to avoid unnecessary reloading. 

```c
// in aot_check_memory_overflow()
    /* Get memory base address and memory data size */
    if (func_ctx->mem_space_unchanged
        || is_shared_memory
#endif
    ) {
        mem_base_addr = func_ctx->mem_info[0].mem_base_addr;  // This branch should be used in the majority of cases.
    }
    else {
        if (!(mem_base_addr = LLVMBuildLoad2(
                  comp_ctx->builder, OPQ_PTR_TYPE,
                  func_ctx->mem_info[0].mem_base_addr, "mem_base"))) {
            aot_set_last_error("llvm build load failed.");
            goto fail;
        }
    }
```

Using the address of the memory base address does not allow optimization passes to recognize the pattern and decide to eliminate superfluous load instructions.

However, I concur that the conditions for setting `mem_space_unchanged` to true are so stringent that it rarely occurs in practice, particularly due to the `has_op_func_call` condition.  If you simply set `mem_space_unchanged` to `true` in `create_memory_info()`, you would achieve the desired improvement. 

```c
// in create_memory_info
    bool mem_space_unchanged = true; // (!func->has_op_memory_grow && !func->has_op_func_call) || (!module->possible_memory_grow);
```

Therefore, I believe the concept of "reloading the base address when the memory might change" is excellent.

If you're in agreement with my perspective, we can begin refactoring the PR by concentrating on "reloading the base address when the memory might change" and eliminating "keeping the address of the memory base address."

Jiax-cn:
@lum1n0us Completely agree

lum1n0us (changes requested):
> If you're in agreement with my perspective, we can begin refactoring the PR by concentrating on "reloading the base address when the memory might change" and eliminating "keeping the address of the memory base address."

Jiax-cn:
> > If you're in agreement with my perspective, we can begin refactoring the PR by concentrating on "reloading the base address when the memory might change" and eliminating "keeping the address of the memory base address."

I agree. Let me know what I can do to help with the refactoring.

lum1n0us:
I believe the approach should be:
- To remove all changes related to "keeping the address of the memory base address" in `AOTMemInfo`.
- To retain (or perhaps slightly refactor) the enhancement concerning "reloading the base address when the memory might change."

lum1n0us:
You're correct; having readable names for AOT functions in the IR and generated code would indeed make debugging more user-friendly. However, the reality is that .wasm files often lack a named section. Typically, in the interest of minimizing binary size, debug information is stripped, which means we need at least two naming systems to handle scenarios with and without a name section.

On the other hand, generated function names are a common assumption across AOT and JIT running modes and their supporting tools. Therefore, unless there's a comprehensive solution available, we might prefer to stick with additional scripts for the time being.

core/iwasm/compilation/aot_emit_memory.c
```diff
@@ -164,17 +161,15 @@ aot_check_memory_overflow(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
...
         mem_base_addr = func_ctx->mem_info[0].mem_base_addr;
-    }
-    else {
+    else
+#endif
+    {
         if (!(mem_base_addr = LLVMBuildLoad2(
                   comp_ctx->builder, OPQ_PTR_TYPE,
-                  func_ctx->mem_info[0].mem_base_addr, "mem_base"))) {
+                  func_ctx->mem_info[0].mem_base_addr, "mem_base_addr"))) {
```
lum1n0us, line 172: I assume the main purpose of this PR is to minimize or eliminate `load` instructions in memory operations. However, the changes eliminated all the fast accesses(`if` branch) but retained the slow ones(`else` branch). Does this actually address the original problem?

Jiax-cn: These two `else` branches are different. In this PR, `mem_info[0].mem_base_addr` comes from `LLVMBuildAlloca`, which allocates memory on the stack (often in registers due to `mem2reg` optimization or in cache), whereas the previous version loads from global memory.

Jiax-cn: Your thought is correct, but you have overlooked the optimization of `LLVMBuildAlloca` variables by `mem2reg`.

> In this PR, mem_info[0].mem_base_addr comes from LLVMBuildAlloca, which allocates memory on the stack (often in registers due to mem2reg optimization or in cache).

For example:
```ir
define i32 @foo(i32 %x) {
entry:
  %y = alloca i32
  store i32 %x, i32* %y
  %val = load i32, i32* %y
  ret i32 %val
}
```

The `load/store` of `%y` be optimized:
```
define i32 @foo(i32 %x) {
entry:
  ret i32 %x
}
```

Similarly, 

```
%mem_base_addr = alloca ptr, align 8
%mem_base_addr_offset = getelementptr inbounds i8, ptr %aot_inst, i32 376
%mem_base_addr1 = load ptr, ptr %mem_base_addr_offset, align 8
store ptr %mem_base_addr1, ptr %mem_base_addr, align 8

%mem_base_addr9 = load ptr, ptr %mem_base_addr, align 8
%maddr = getelementptr inbounds i8, ptr %mem_base_addr9, i64 %offset1
```

can be optimized to:

```
%mem_base_addr_offset = getelementptr inbounds i8, ptr %aot_inst, i32 376
%mem_base_addr1 = load ptr, ptr %mem_base_addr_offset, align 8

%maddr = getelementptr inbounds i8, ptr %mem_base_addr1, i64 %offset1
```

In this PR, if you dump the IR of substr.wasm above, you will see that the load/store operations of `%mem_base_addr` are also optimized. As a result, the final outcome remains the same as in the original code.

image


image

lum1n0us: I've devised a straightforward test case to assess the enhancement.  Here is the code:

```wat
(module
  (memory 5 10)

  (func $store32 (export "store32")  (param i32 i32)
    (i32.store (local.get 0) (local.get 1))
  )

  (func $load32 (export "load32")  (param i32) (result i32)
    (i32.load (local.get 0))
  )

  (func (export "load_store") (param i32 i32) (result i32)
    (local i32)

    (i32.load (local.get 0))
    (local.tee 2)
    (i32.store (local.get 1))

    (i32.load (local.get 1))
    (local.get 2)
    (i32.eq)
  )

  (func (export "load_grow_store") (param i32 i32) (result i32)
    (local i32)

    (i32.load (local.get 0))
    (local.tee 2)
    (i32.store (local.get 1))

    (memory.grow (i32.const 1))
    (drop)

    (i32.load (local.get 1))
    (local.get 2)
    (i32.eq)
  )

  (func (export "load_store_w_func") (param i32 i32) (result i32)
    (local i32)

    (local.get 0)
    (call $load32)

    (local.tee 2)
    (local.get 1)
    (call $store32)

    (i32.load (local.get 1))
    (local.get 2)
    (i32.eq)
  )
  
  (func (export "load_grow_store_w_func") (param i32 i32) (result i32)
    (local i32)

    (local.get 0)
    (call $load32)

    (local.tee 2)
    (local.get 1)
    (call $store32)

    (memory.grow (i32.const 1))
    (drop)

    (i32.load (local.get 1))
    (local.get 2)
    (i32.eq)
  )
)
``` 

And I used the following command: `--bounds-checks=1 --format=llvmir-op`, to create optimized llvmir. The `--bounds-checks=1` is employed to apply the `noinline` attribute.

Several intriguing findings emerged from the comparison of the before and after scenarios:

- Look at f0, f1, and f2. These are elementary cases involving load and store. As previously mentioned, the mem2reg optimization refines alloca variables, allowing the revised version to produce no additional IR compared to the original version.
- Now, consider f4 and f5. I believed they presented issues that this PR aims to address. Clearly, as seen in f5, there is no necessity to reload the memory base address after calling f1, as there is no memory growth in f4. This PR should eliminate that redundant loading. However, the modified version maintains the status quo.

:up: If I'm mistaken, please correct me.

This leads to my confusion: if there is no difference for basic cases and no enhancement for redundant loading, what is the rationale for changing?

Jiax-cn: This PR is to eliminate redundant memory info loads between multiple load/store instructions when the memory remains unchanged.

I use the following WAST example to compare the (optimized) IR generated by this PR with the original code, identifying specific scenarios where this optimization applies. I don’t need the `noinline` attribute, so I only used the `--format=llvmir-op` option.

```wast
(module
  (memory 5 10)

  (func (export "load_load") (param i32 i32) (result i32)
    (i32.load (local.get 0))
    (i32.load (local.get 1))
    (i32.eq)

    (memory.grow (i32.const 1))
    (drop)
  )

  (func (export "load_store") (param i32 i32)
    (i32.load (local.get 0))
    (i32.store (local.get 1))

    (memory.grow (i32.const 1))
    (drop)
  )

  (func (export "store_store") (param i32 i32)
    (i32.store (local.get 0) (i32.const 42))
    (i32.store (local.get 1) (i32.const 42))

    (memory.grow (i32.const 1))
    (drop)
  )

  (func (export "store_load") (param i32 i32) (result i32)
    (i32.store (local.get 0) (i32.const 42))
    (i32.load (local.get 1))

    (memory.grow (i32.const 1))
    (drop)
  )
)
```

Same IR in f0 and f1.
image

In f3 and f4, the IR generated by this PR is different.
image

This PR primarily optimizes store-store and store-load scenarios. As for load-load and load-store scenarios, I believe they might have been optimized.

In your example, f0\f1 shows this PR won't produce no addtional IR compared to the original version. The memory in f2 is unchanged. f3 is the load-store case. f4 and f5 contains two call and one load instruction. Therefore, their IRs remains the same.

lum1n0us:
Let me summarize it for us.

IMU, there are several scenarios we need to examine closely, such as single-load, single-store, load-load, load-store, store-store, store-load, load-store-grow-load-store, load-store-call-load-store, and load-store-call-grow-load-store. The rationale for the last three scenarios is that in the original implementation, the key condition for controlling the reloading of the memory base address is `func->has_op_memory_grow && !func->has_op_func_call`.

After merging both test scripts, I believe we can address all the aforementioned cases. For your information, since the cases are quite straightforward, particularly as functions being called, the compilation process will likely inline these functions, causing the last two cases to lose the call. This is the reason I recommend using --bounds-checks=1 to turn off the inline optimization.
The improvements in the pull request are as follows.

| case                            | before | after | FIXED? |
| ------------------------------- | ------ | ----- | ------ |
| single-load                     | N      | N     | N/A    |
| single-store                    | N      | N     | N/A    |
| load-load                       | N      | N     | N/A    |
| load-store                      | N      | N     | N/A    |
| store-store                     | Y      | N     | Y      |
| store-load                      | Y      | N     | Y      |
| load-store-grow-load-store      | N      | N     | N/A    |
| load-store-call-load-store      | Y      | Y     | N✨   |
| load-store-call-grow-load-store | Y      | Y     | N✨   |

***N*** means no redundant load/store
***Y*** means redundant load/store

PR has improved two cases and left two cases involving a `call`

===


Use `load-store-call-load-store` as an example

BEFORE:

```llvmir

;; THIS IS BEFORE

define internal i32 @"aot_func_internal#7"(ptr nocapture readonly %exec_env, i32 %0, i32 %1) unnamed_addr #0 {
func_begin:
  %aot_inst_addr = getelementptr inbounds ptr, ptr %exec_env, i64 2
  %aot_inst25 = load i64, ptr %aot_inst_addr, align 8
  %2 = inttoptr i64 %aot_inst25 to ptr
  %mem_base_addr_offset = getelementptr inbounds i8, ptr %2, i64 376
  %bound_check_4bytes_offset = getelementptr inbounds i8, ptr %2, i64 440
  %cur_exception = getelementptr inbounds i8, ptr %2, i64 104
  %addr_i64 = zext i32 %0 to i64
  %mem_check_bound = load i64, ptr %bound_check_4bytes_offset, align 8
  %cmp = icmp ult i64 %mem_check_bound, %addr_i64
  br i1 %cmp, label %got_exception, label %check_succ

check_succ:                                       ; preds = %func_begin
  %mem_base = load ptr, ptr %mem_base_addr_offset, align 8
  %maddr = getelementptr inbounds i8, ptr %mem_base, i64 %addr_i64
  store i32 42, ptr %maddr, align 1
  %addr_i642 = zext i32 %1 to i64
  %mem_check_bound4 = load i64, ptr %bound_check_4bytes_offset, align 8
  %cmp5 = icmp ult i64 %mem_check_bound4, %addr_i642
  br i1 %cmp5, label %got_exception, label %check_succ6

check_succ6:                                      ; preds = %check_succ
  %call = tail call i32 @"aot_func#1"(ptr %exec_env, i32 %0)
  %exce_value = load i8, ptr %cur_exception, align 1
  %cmp9 = icmp eq i8 %exce_value, 0
  br i1 %cmp9, label %check_exce_succ, label %common.ret

check_exce_succ:                                  ; preds = %check_succ6
  tail call void @"aot_func#0"(ptr %exec_env, i32 %call, i32 %1)
  %exce_value11 = load i8, ptr %cur_exception, align 1
  %cmp12 = icmp eq i8 %exce_value11, 0
  br i1 %cmp12, label %func_end, label %common.ret

common.ret:                                       ; preds = %check_succ6, %check_exce_succ, %got_exception, %func_end
  %common.ret.op = phi i32 [ %data24, %func_end ], [ 0, %got_exception ], [ 0, %check_exce_succ ], [ 0, %check_succ6 ]
  ret i32 %common.ret.op

func_end:                                         ; preds = %check_exce_succ
  %mem_base15 = load ptr, ptr %mem_base_addr_offset, align 8     ; <== 💥 redundant  reloading
  %maddr18 = getelementptr inbounds i8, ptr %mem_base15, i64 %addr_i64
  store i32 42, ptr %maddr18, align 1
  %mem_base20 = load ptr, ptr %mem_base_addr_offset, align 8    ; <== 💥 redundant  reloading 
  %maddr23 = getelementptr inbounds i8, ptr %mem_base20, i64 %addr_i642
  %data24 = load i32, ptr %maddr23, align 1
  br label %common.ret

got_exception:                                    ; preds = %check_succ, %func_begin
  tail call void @aot_set_exception_with_id(ptr %2, i32 2)
  br label %common.ret
}

```

AFTER:

```llvmir

;; THIS IS AFTER

define internal i32 @"aot_func_internal#7"(ptr nocapture readonly %exec_env, i32 %0, i32 %1) unnamed_addr #0 {
func_begin:
  %aot_inst_addr = getelementptr inbounds ptr, ptr %exec_env, i64 2
  %aot_inst78 = load i64, ptr %aot_inst_addr, align 8
  %2 = inttoptr i64 %aot_inst78 to ptr
  %3 = getelementptr inbounds i8, ptr %2, i64 440
  %mem_bound_check_4bytes684 = load i64, ptr %3, align 8
  %cur_exception = getelementptr inbounds i8, ptr %2, i64 104
  %addr_i64 = zext i32 %0 to i64
  %cmp = icmp ult i64 %mem_bound_check_4bytes684, %addr_i64
  br i1 %cmp, label %got_exception, label %check_succ

check_succ:                                       ; preds = %func_begin
  %4 = getelementptr inbounds i8, ptr %2, i64 376
  %mem_base_addr181 = load i64, ptr %4, align 8
  %5 = inttoptr i64 %mem_base_addr181 to ptr
  %maddr = getelementptr inbounds i8, ptr %5, i64 %addr_i64
  store i32 42, ptr %maddr, align 1
  %addr_i6411 = zext i32 %1 to i64
  %cmp14 = icmp ult i64 %mem_bound_check_4bytes684, %addr_i6411
  br i1 %cmp14, label %got_exception, label %check_succ15

check_succ15:                                     ; preds = %check_succ
  %call = tail call i32 @"aot_func#1"(ptr %exec_env, i32 %0)
  %exce_value = load i8, ptr %cur_exception, align 1
  %cmp18 = icmp eq i8 %exce_value, 0
  br i1 %cmp18, label %check_exce_succ, label %common.ret

check_exce_succ:                                  ; preds = %check_succ15
  tail call void @"aot_func#0"(ptr %exec_env, i32 %call, i32 %1)
  %exce_value36 = load i8, ptr %cur_exception, align 1
  %cmp37 = icmp eq i8 %exce_value36, 0
  br i1 %cmp37, label %func_end, label %common.ret

common.ret:                                       ; preds = %check_succ15, %check_exce_succ, %got_exception, %func_end
  %common.ret.op = phi i32 [ %data65, %func_end ], [ 0, %got_exception ], [ 0, %check_exce_succ ], [ 0, %check_succ15 ]
  ret i32 %common.ret.op

func_end:                                         ; preds = %check_exce_succ
  %mem_base_addr4267 = load i64, ptr %4, align 8    ; <== 💥 redundant  reloading
  %6 = inttoptr i64 %mem_base_addr4267 to ptr
  %maddr59 = getelementptr inbounds i8, ptr %6, i64 %addr_i64
  store i32 42, ptr %maddr59, align 1

  %maddr64 = getelementptr inbounds i8, ptr %6, i64 %addr_i6411
  %data65 = load i32, ptr %maddr64, align 1
  br label %common.ret

got_exception:                                    ; preds = %check_succ, %func_begin
  tail call void @aot_set_exception_with_id(ptr %2, i32 2)
  br label %common.ret
}
```

Jiax-cn:
> the key condition for controlling the reloading of the memory base address is `func->has_op_memory_grow && !func->has_op_func_call`.

IIUC, the key condition is `func->has_op_memory_grow || func->has_op_func_call`, since both memory growth and function calls can result in memory changes. Therefore, reloading **once** after each call is not redundant in your `load–store–call–load–store` example.

lum1n0us:
Yes. In the original version, `func->has_op_memory_grow || func->has_op_func_call` was a firm rule. However, it led to unnecessary extra reloading since not every function call involves a memory.grow. I recall that several earlier changes in this pull request were aimed at recalculating the linear memory base address only after a memory.grow, which I thought could alleviate the situation if it encompassed all scenarios. But it seems the latest version has abandoned this approach.

Jiax-cn:
Recalculating the linear memory base address only after a `memory.grow` may not be sufficient. If the callee performs a memory grow, the caller must also update `mem_base` accordingly. So I just reload after all calls and memory grow operations in ALL my commits.

Maybe you mean a fine-grained reloading? like analyze the call structure and remove reloads on paths without any `memory.grow` instructions.

core/iwasm/compilation/aot_compiler.c
```diff
@@ -2092,6 +2099,8 @@ aot_compile_func(AOTCompContext *comp_ctx, uint32 func_index)
                 read_leb_uint32(frame_ip, frame_ip_end, mem_idx);
                 if (!aot_compile_op_memory_grow(comp_ctx, func_ctx))
                     return false;
+                if (module->functions[func_index]->has_memory_operations)
+                    restore_memory_info(comp_ctx, func_ctx);
```
yamt, line 2103: maybe we can defer this until the next br/load/store?

PR #4114, opened 2025-02-26
Fix loading AOT on aarch64

Jiax-cn:
My `target_info.arch` is "aarch64", but `check_machine_info()` compares it with "aarch64v".


lum1n0us (approved):
LGTM

TianlongLiang (approved):
LGTM

PR #4117, opened 2025-02-27
Add wasm_runtime_get_func_name_from_index() api

Jiax-cn:
WAMR already implements this function, but its scope is limited to the source file and only applies to the module instance. I reimplemented this because it is clearer to obtain function names in the following scenarios, where the (WASM/AOT) module has not yet been instantiated:

1. `aot_create_perf_map()` This creates /tmp/perf-<pid>.map for linux-perf.
2. `aot_add_llvm_func1()` By doing so, LLVM can include function names in the IR and machine code, allowing us to see them in the call stack when debugging AOT with LLDB/GDB. 

If this is useful, I would be happy to submit PRs for these two scenarios.

### DIFF
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -7828,3 +7828,108 @@ wasm_runtime_is_underlying_binary_freeable(WASMModuleCommon *const module)
 
     return true;
 }
+
+#if WASM_ENABLE_AOT != 0
+#if WASM_ENABLE_CUSTOM_NAME_SECTION != 0
+static const char *
+aot_lookup_func_name(const char **func_names, uint32 *func_indexes,
+                 uint32 func_index_count, uint32 func_index)
+{
+    int64 low = 0, mid;
+    int64 high = func_index_count - 1;
+
+    if (!func_names || !func_indexes || func_index_count == 0)
+        return NULL;
+
+    while (low <= high) {
+        mid = (low + high) / 2;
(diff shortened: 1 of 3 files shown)


lum1n0us:
If the goal is to enhance the readability of the performance map, perhaps we should consider enhancing the functionality of [process_folded_data.py](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/test-tools/flame-graph-helper/process_folded_data.py).

---

```c
WASM_RUNTIME_API_EXTERN const uint8_t *
wasm_runtime_get_custom_section(const wasm_module_t module_comm,
                                const char *name, uint32_t *len);
```
can retrieve the content of the name section. With this information, the host can complete the mapping between function indices and their names. Personally, I think using an index to retrieve a name is an infrequent operation.

Jiax-cn:
> You're correct; having readable names for AOT functions in the IR and generated code would indeed make debugging more user-friendly. However, the reality is that .wasm files often lack a named section. Typically, in the interest of minimizing binary size, debug information is stripped, which means we need at least two naming systems to handle scenarios with and without a name section.
> 
> On the other hand, generated function names are a common assumption across AOT and JIT running modes and their supporting tools. Therefore, unless there's a comprehensive solution available, we might prefer to stick with additional scripts for the time being.

_Originally posted by @lum1n0us in https://github.com/bytecodealliance/wasm-micro-runtime/issues/4113#issuecomment-2717543365_

This PR has been reviewed and discussed. I am closing it but will keep my branch for future reference.

PR #4119, opened 2025-02-28
Use log instead of using assertion in aot loader 

TianlongLiang:


lum1n0us (approved):
ignore this one

lum1n0us (changes requested):
```
if (mem < INT32_MAX) {
  return mem;
}
LOG_WARNING(...)
// try with another options

```

no1wudi (approved):

Issue #2163, opened 2023-05-01
support for gas metering


rockwotj:
Other Wasm runtimes have support for gas metering or fuel that allow Wasm code to have an upper bound on computation. Wasmtime's implementation even supports yielding. It'd be great to have this support in wamr

https://docs.rs/wasmtime/latest/wasmtime/struct.Config.html#method.consume_fuel

PhaniKb:
This would be a really nice feature to have.

efJerryYang:
@lum1n0us Hi, this one may need to be closed and it has been resolved by #4122

PR #4122, opened 2025-03-02
feat: add instruction metering support with execution limit

### LINKED ISSUE #3927: Instruction metering
# Feature

Similar to the Wasmer runtime, it would be beneficial to implement instruction metering, allowing control to be handed back to the caller after executing a specified number of instructions.

# Benefit

This would enable a cooperative scheduler to manage how WebAssembly (Wasm) code is executed, preventing system overload.

# Implementation

Introduce a function like `wasm_func_call_with_metering` that returns an after a given number of executed instruction, a boolean indicating whether the function reached its end.

# Alternatives

None identified.

AlixANNERAUD:
Resolve #3927 and #2163.

### DIFF
--- a/core/config.h
+++ b/core/config.h
@@ -706,4 +706,8 @@
 #define WASM_ENABLE_AOT_VALIDATOR 0
 #endif
 
+#ifndef WASM_INSTRUCTION_METERING 
+#define WASM_INSTRUCTION_METERING 1
+#endif
+
 #endif /* end of _CONFIG_H_ */
--- a/core/iwasm/interpreter/wasm_interp_classic.c
+++ b/core/iwasm/interpreter/wasm_interp_classic.c
@@ -1516,10 +1516,13 @@ wasm_interp_call_func_import(WASMModuleInstance *module_inst,
             }                                                                 \
             os_mutex_unlock(&exec_env->wait_lock);                            \
         }                                                                     \
+        CHECK_INSTRUCTION_LIMIT();                                           \                                                                \
         goto *handle_table[*frame_ip++];                                      \
     } while (0)
 #else
-#define HANDLE_OP_END() FETCH_OPCODE_AND_DISPATCH()
+#define HANDLE_OP_END()                                                   \
+    CHECK_INSTRUCTION_LIMIT();                                           \
+    FETCH_OPCODE_AND_DISPATCH()                                           
 #endif
 
 #else /* else of WASM_ENABLE_LABELS_AS_VALUES */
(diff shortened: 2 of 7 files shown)


AlixANNERAUD:
Currently, there is no implementation on the AOT side since I'm not sure if this is really possible since we are dealing with compiled / native code.

Regarding the classic/fast interpreter, I'm note sure about the save/restore of the context. It seems that `prev_frame` is saved in `exec_env` and reused upon invocation. However, I'm unsure whether `prev_frame` alone is sufficient or if we also need to back up `frame` (and maybe more).

core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -1562,6 +1568,16 @@ get_global_addr(uint8 *global_data, WASMGlobalInstance *global)
 #endif
 }
 
+#if WASM_INSTRUCTION_METERING != 0
+#define CHECK_INSTRUCTION_LIMIT() \
+    if (instructions_left == 0) { \
+        goto return_func;         \
```
lum1n0us, line 1574: Thank you for contributing.

I totally agree with the concept of checking the meter in `HANDLE_OP_END`.

However, I have reservations about this part. Clearly, running out of meter should not be equated with a normal return. It's more akin to being suspended or interrupted. Additionally, unlike a normal return, there is no return value for subsequent processing. Perhaps a special exception is needed.

Another one, When `instructions_left == -1`, IIUC, this signifies unlimited instructions. We should bypass the calculation to minimize performance degradation, as this check occurs after every opcode execution.

AlixANNERAUD: Thank you for your feedback.

I’ve made the necessary corrections. However, I’m still struggling to see how this could be done differently for the comparison. The logic seems necessary to handle the specific cases effectively.

core/config.h
```diff
@@ -710,4 +710,8 @@
 #define WASM_ENABLE_AOT_VALIDATOR 0
 #endif
 
+#ifndef WASM_INSTRUCTION_METERING
+#define WASM_INSTRUCTION_METERING 1
```
TianlongLiang, line 714: I think we should disable them by default since there is still overhead when enabled, and add CMake cache variable to turn it on/off.

AlixANNERAUD: I fixed it and add doc.

TianlongLiang (approved):
LGTM

lum1n0us (approved):
LGTM

PR #4124, opened 2025-03-03
Expose WAMR_BUILD_GC_HEAP_SIZE_DEFAULT as a CMake option

swankjesse:
This is wired through to the GC_HEAP_SIZE_DEFAULT constant.

Also honor this value when configuring the engine with the wasm_c_api.

### DIFF
--- a/core/iwasm/common/wasm_c_api.c
+++ b/core/iwasm/common/wasm_c_api.c
@@ -392,8 +392,8 @@ wasm_engine_new_internal(wasm_config_t *config)
 
     WASM_C_DUMP_PROC_MEM();
 
-    /* wasm_config_t->MemAllocOption -> RuntimeInitArgs->MemAllocOption */
     init_args.gc_heap_size = GC_HEAP_SIZE_DEFAULT;
+    /* wasm_config_t->MemAllocOption -> RuntimeInitArgs->MemAllocOption */
     init_args.mem_alloc_type = config->mem_alloc_type;
     memcpy(&init_args.mem_alloc_option, &config->mem_alloc_option,
            sizeof(MemAllocOption));
--- a/core/iwasm/common/wasm_c_api.c
+++ b/core/iwasm/common/wasm_c_api.c
@@ -392,7 +392,6 @@ wasm_engine_new_internal(wasm_config_t *config)
 
     WASM_C_DUMP_PROC_MEM();
 
-    init_args.gc_heap_size = GC_HEAP_SIZE_DEFAULT;
     /* wasm_config_t->MemAllocOption -> RuntimeInitArgs->MemAllocOption */
     init_args.mem_alloc_type = config->mem_alloc_type;
     memcpy(&init_args.mem_alloc_option, &config->mem_alloc_option,
--- a/core/iwasm/common/wasm_c_api.c
+++ b/core/iwasm/common/wasm_c_api.c
@@ -393,6 +393,7 @@ wasm_engine_new_internal(wasm_config_t *config)
     WASM_C_DUMP_PROC_MEM();
 
     /* wasm_config_t->MemAllocOption -> RuntimeInitArgs->MemAllocOption */
+    init_args.gc_heap_size = GC_HEAP_SIZE_DEFAULT;
     init_args.mem_alloc_type = config->mem_alloc_type;
     memcpy(&init_args.mem_alloc_option, &config->mem_alloc_option,
            sizeof(MemAllocOption));
--- a/core/shared/mem-alloc/mem_alloc.cmake
+++ b/core/shared/mem-alloc/mem_alloc.cmake
@@ -24,6 +24,10 @@ if (WAMR_BUILD_GC_CORRUPTION_CHECK EQUAL 0)
     add_definitions (-DBH_ENABLE_GC_CORRUPTION_CHECK=0)
 endif ()
 
+if (DEFINED WAMR_BUILD_GC_HEAP_SIZE_DEFAULT)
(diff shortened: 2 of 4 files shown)


swankjesse:
Alternately, it’d be great to expose the gc_heap_size as option in `wasm_c_api.h` directly. That’d be ideal as it’d give callers the ability to set a heap size at runtime instead of at build time.

core/iwasm/common/wasm_c_api.c
```diff
@@ -393,6 +393,7 @@ wasm_engine_new_internal(wasm_config_t *config)
     WASM_C_DUMP_PROC_MEM();
 
     /* wasm_config_t->MemAllocOption -> RuntimeInitArgs->MemAllocOption */
+    init_args.gc_heap_size = GC_HEAP_SIZE_DEFAULT;
```
lum1n0us, line 396:
```patch
++    init_args.gc_heap_size = GC_HEAP_SIZE_DEFAULT;
    /* wasm_config_t->MemAllocOption -> RuntimeInitArgs->MemAllocOption */
--    init_args.gc_heap_size = GC_HEAP_SIZE_DEFAULT;
```

core/shared/mem-alloc/mem_alloc.cmake
```diff
@@ -24,6 +24,10 @@ if (WAMR_BUILD_GC_CORRUPTION_CHECK EQUAL 0)
     add_definitions (-DBH_ENABLE_GC_CORRUPTION_CHECK=0)
 endif ()
 
+if (DEFINED WAMR_BUILD_GC_HEAP_SIZE_DEFAULT)
+    add_definitions ("-DGC_HEAP_SIZE_DEFAULT=${WAMR_BUILD_GC_HEAP_SIZE_DEFAULT}")
+endif ()
```
lum1n0us, line 29: Please include the description and instructions for using this new compilation option in the document, doc/build_wamr.md.

lum1n0us (changes requested):

core/iwasm/common/wasm_c_api.c
```diff
@@ -392,6 +392,7 @@ wasm_engine_new_internal(wasm_config_t *config)
 
     WASM_C_DUMP_PROC_MEM();
 
+    init_args.gc_heap_size = GC_HEAP_SIZE_DEFAULT;
```
swankjesse, line 395: **FYI: Behavior change!**

Previously `init_args.gc_heap_size` was initialized to `0`. It’d ultimately end up using `GC_HEAP_SIZE_MIN` (4 kB).

With this change `init_args.gc_heap_size` is initialized to `GC_HEAP_SIZE_DEFAULT` (128 kB), which is what it’ll end up using.

This changes behavior even for users who don’t set a custom heap size.

doc/build_wamr.md
```diff
@@ -141,6 +141,7 @@ cmake -DWAMR_BUILD_PLATFORM=linux -DWAMR_BUILD_TARGET=ARM
 
 ### **Enable Garbage Collection**
 - **WAMR_BUILD_GC**=1/0, default to disable if not set
+- **WAMR_BUILD_GC_HEAP_SIZE_DEFAULT**=n, default to 128 kB (131072) if not set
```
TianlongLiang, line 144:
```suggestion

### **Set the Garbage Collection heap size**
- **WAMR_BUILD_GC_HEAP_SIZE_DEFAULT**=n, default to 128 kB (131072) if not set
```

core/iwasm/common/wasm_c_api.c
lum1n0us: You're correct; it does present a bit of a challenge. And this gives me second thoughts about the PR.

- Both `wasm_instantiate()` and `aot_instantiate()` will adjust the value of `gc_heap_size` to fall within the range of `[GC_HEAP_SIZE_MIN, GC_HEAP_SIZE_MAX]`. This means that even if you set a custom `GC_HEAP_SIZE_DEFAULT` on the command line, you might end up with a different value during execution.
- If we aim to unify the behavior for both *wasm_c_api.h* and *wasm_export.h* users to avoid confusion, I think a better place for the modification might be within `wasm_runtime_full_init_internal()`, perhaps with a line like `if (init_args->gc_heap_size > 0) gc_heap_size_default = init_args->gc_heap_size`.
- Typically, we manually adjust the value in config.h rather than through the command line. I'm not certain if we should introduce `GC_HEAP_SIZE_DEFAULT` as a command-line option at this point.

swankjesse: >  perhaps with a line like if (init_args->gc_heap_size > 0) gc_heap_size_default = init_args->gc_heap_size.

Good idea. Done.

swankjesse: I don’t have a preference on whether to expose GC heap configuration on the command line. I’m working on embedding it!

lum1n0us: In light of that, would you mind removing the changes in *core/shared/mem-alloc/mem_alloc.cmake* and *doc/build_wamr.md*? This will ensure that manually editing *config.h* remains the only method for setting a customized value when necessary.

swankjesse: While I am embedding wamr, I would prefer to avoid editing the source in order to configure my build. It’ll make it easiest to track upstream.

lum1n0us (approved):
LGTM

TianlongLiang (approved):
LGTM

no1wudi (approved):

PR #4135, opened 2025-03-11
Don't fail to compile if feature macros aren't defined

sjamesr:
This makes the output assume the feature is not enabled if the macro is not defined at all, rather than just being set to zero.

I know of at least one build that does not use CMake and does not set the macro if the feature is not enabled.

### DIFF
--- a/product-mini/platforms/common/wasm_proposal.c
+++ b/product-mini/platforms/common/wasm_proposal.c
@@ -5,6 +5,8 @@
 
 #include <stdio.h>
 
+#include "bh_platform.h"
+
 void
 wasm_proposal_print_status(void)
 {


lum1n0us:
>I know of at least one build that does not use CMake and does not set the macro if the feature is not enabled.

🤔  *core/config.h* should always define them

sjamesr:
That's much easier, thanks :)

product-mini/platforms/common/wasm_proposal.c
```diff
@@ -5,6 +5,8 @@
 
 #include <stdio.h>
 
+#include "../../../core/config.h"
```
lum1n0us, line 8:
```suggestion
#include "bh_platform.h"
```

Usually, we `#include "bh_platform.h"` which `#include "../platform/include/platform_common.h"` which `#include "../../../config.h"`

lum1n0us (changes requested):

product-mini/platforms/common/wasm_proposal.c
sjamesr: Done, thanks

lum1n0us (approved):
LGTM

TianlongLiang (approved):
LGTM

PR #4138, opened 2025-03-13
Fix iwasm build error when WAMR_BUILD_WASI_NN enabled

zhen9910:
A recent change on ./product-mini/platforms/linux/CMakeLists.txt renamed libiwasm to vmlib, but wasi-nn.cmake still wants to link libiwasm.so. Replace libiwasm with vmlib in wasi-nn.cmake to resolve iwasm build error when WAMR_BUILD_WASI_NN enabled.


TianlongLiang (approved):
LGTM

lum1n0us (approved):
LGTM

PR #4161, opened 2025-03-25
Fix few integer overflowing

lum1n0us:
- fix(interpreter): correct offset calculations in wasm_loader_get_const_offset function
- fix(mem-alloc): update offset calculation in gc_migrate for memory migration


core/shared/mem-alloc/ems/ems_kfc.c
```diff
@@ -236,6 +236,13 @@ gc_migrate(gc_handle_t handle, char *pool_buf_new, gc_size_t pool_buf_size)
         return GC_ERROR;
     }
 
+    if ((uintptr_t)base_addr_new > (uintptr_t)heap->base_addr) {
+        offset = (uintptr_t)base_addr_new - (uintptr_t)heap->base_addr;
```
xujuntwt95329, line 240: The offset is used in adjust_ptr, if we use absolute value here, the adjust_ptr seems will return a wrong address?

core/shared/mem-alloc/ems/ems_kfc.c
```diff
@@ -208,8 +208,23 @@ gc_get_heap_struct_size()
...
+     * `(intptr_t)`
+     */
+    uintptr_t offset_val = 0;
+#if UINTPTR_MAX == UINT64_MAX
+    offset_val = labs(offset);
+#else
+    offset_val = abs(offset);
+#endif
+
+    *p_ptr = (uint8 *)((uintptr_t)(*p_ptr) + offset_val);
```
TianlongLiang, line 227: Should we use `- offset_val` for negative offset?

lum1n0us: thanks for reminding.

TianlongLiang (approved):
LGTM

core/shared/mem-alloc/ems/ems_kfc.c
```diff
@@ -208,8 +208,28 @@ gc_get_heap_struct_size()
...
-    if (*p_ptr)
-        *p_ptr = (uint8 *)((intptr_t)(*p_ptr) + offset);
+    if ((!*p_ptr)) {
+        return;
+    }
+
+    /*
+     * to resolve a possible signed integer overflow issue
+     * when p_ptr is over 0x8000000000000000  by not using
+     * `(intptr_t)`
```
yamt, line 218: isn't it enough to use unsigned addition?

xujuntwt95329 (approved):
LGTM

PR #4162, opened 2025-03-25
Remove indirect-load for constants on Xtensa Target to improve performance

dongsheng28849455:
In XIP mode on Xtensa platform, compared with non XIP, there may be some performance degradation. The Coremark test results on the ESP32/Nuttx platform are:
> Non-XIP:  114.94
> XIP: 83

According to the handling of constants by the Xtensa backend, even without indrect load, there will be no relocation. Because the backend does not perform fixup processing on immediate numbers when processing jump/branch instructions, Refer to [getL32RTargetEncoding](https://github.com/espressif/llvm-project/blob/xtensa_release_19.1.2/llvm/lib/Target/Xtensa/MCTargetDesc/XtensaMCCodeEmitter.cpp#L325)
So for the Xtensa platform, the indirct loading is not so necessary, after removing it, the performance is improve as below(Coremark):
> Opt XIP: 112.9


core/iwasm/compilation/aot_emit_const.c
```diff
@@ -12,6 +12,7 @@ aot_compile_op_i32_const(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
 {
     LLVMValueRef value;
 
+#if !defined(BUILD_TARGET_XTENSA)
```
no1wudi, line 15: How about implement this by modify the intrinsic flags instead?
https://github.com/bytecodealliance/wasm-micro-runtime/blob/8acaaa2892758eb309759a4ae760f2b9e25b9785/core/iwasm/aot/aot_intrinsic.c#L870-L884

core/iwasm/compilation/aot_emit_conversion.c
```diff
@@ -347,7 +347,9 @@ aot_compile_op_i32_trunc_f32(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
 
     POP_F32(value);
 
+#if !defined(BUILD_TARGET_XTENSA)
```
no1wudi, line 350: How about add more intrinsic flags for conversions?

core/iwasm/compilation/aot_emit_conversion.c
```diff
@@ -347,17 +347,8 @@ aot_compile_op_i32_trunc_f32(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
...
-            max_value = F32_CONST(2147483648.0f);
-        }
-        else {
-            min_value = F32_CONST(-1.0f);
-            max_value = F32_CONST(4294967296.0f);
-        }
-    }
-    else {
+    if (comp_ctx->is_indirect_mode
+        && aot_intrinsic_check_capability(comp_ctx, "f32.const")) {
```
lum1n0us, line 351: Perhaps there ought to be a function to eliminate a particular intrinsic capability, such as `remove_intrinsic_capability(comp_ctx, AOT_INTRINSIC_FLAG_F32_CMP)`?

lum1n0us (approved):
LGTM

PR #4169, opened 2025-03-31
dwarf_extractor.cpp: add macro control

TianlongLiang:
Add macro control to be compatible with lower version toolchain


xujuntwt95329 (approved):
LGTM

PR #4175, opened 2025-04-01
Update cmake min. requirement to 3.14

lum1n0us:
3.14 is used and tested by liinux mini-product

it is able to fix https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/14174679441/job/39706594675?pr=4171

```
CMake Error at CMakeLists.txt:4 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 # Copyright (C) 2019 Intel Corporation.  All rights reserved.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-cmake_minimum_required (VERSION 3.0)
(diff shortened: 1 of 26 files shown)


TianlongLiang (approved):
LGTM

no1wudi (approved):

PR #4177, opened 2025-04-01
fix format specifier warning on 32bit builds

sjamesr:


sjamesr:
I think the errors are related to the test environment, not the change itself

core/iwasm/aot/aot_perf_map.c
```diff
@@ -3,6 +3,8 @@
  * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
  */
 
+#include <inttypes.h>
```
lum1n0us, line 6: Not include a new header. If required, please add a new definition about `PRIuPTR` in *core/shared/platform/include/platform_common.h*.

lum1n0us (changes requested):

core/iwasm/aot/aot_perf_map.c
sjamesr: Can you help me understand why we don't want to include this standard (since C99) header?

sjamesr: there was already a PRIxPTR in platform_common, hopefully that does the trick

yamt (approved):
lgtm

core/iwasm/aot/aot_perf_map.c
yamt: > Can you help me understand why we don't want to include this standard (since C99) header?

platform_common.h includes inttypes.h where available for you.
it also provides fallback definitions of PRIxPTR etc.
i dunno where the fallback is actually necessary though.

lum1n0us (approved):
LGTM

Issue #4103, opened 2025-02-24
add x86_64-w64-windows-gnu/aarch64-w64-windows-gnu/i686-w64-windows-gnu aot triple support


calvin2021y:
current wamrc only support mscv  triple,   can not made it work with x86_64-w64-windows-gnu.  (you can get this from https://github.com/mstorsjo/llvm-mingw/releases/download/20250114/llvm-mingw-20250114-msvcrt-ubuntu-20.04-x86_64.tar.xz)

calvin2021y:
I guess allow pass triple from arguments for wamrc will be a lot easy, and support all future arch & target.

if pass triple, then arch and target can be infered.

TianlongLiang:
You can use option `--target-abit=gnu` to change target abi, should it be sufficient for your usage?
```bsh
./wamrc.exe --target-abi=help
Supported ABI: gnu eabi eabihf gnueabihf msvc ilp32 ilp32f ilp32d lp64 lp64f lp64d
```

calvin2021y:
I guess it must be `x86_64-w64-windows-gnu`,  with `--target-abit=gnu` I get `x86_64-pc-linux-gnu`

 ```sh
./wamrc.exe --target-abi=gnu --target=x86_64 --enable-tail-call --format=aot -o fib.win fib.wasm
Create AoT compiler with:           
  target:        x86_64             
  target cpu:                       
  target triple: x86_64-pc-linux-gnu
  cpu features:                     
  opt level:     3                  
  size level:    3                  
  output format: AoT file           
Compile success, file fib.win was generated.
```

crashed when  load aot

```sh
./tests_wamr.exe             
=================================================================
==3320==ERROR: AddressSanitizer: access-violation on unknown address 0x01cf501c0004 (pc 0x7ff8f68f4678 bp 0x006411efb8b0 sp 0x006411efb828 T0)
==3320==The signal is caused by a WRITE memory access.                                                                                        
    #0 0x7ff8f68f4677 in memset+0x77 (C:\Windows\SYSTEM32\ntdll.dll+0x1800a4677)                                                              
    #1 0x7ff8b3ff2cc0 in _asan_memset+0x90 (C:\Users\admin\libclang_rt.asan_dynamic-x86_64.dll+0x180042cc0)                                   
    #2 0x7ff7f0b34206  (C:\Users\admin\tests_wamr.exe+0x140214206)                                                                            
[...]

PR #4183, opened 2025-04-06
Fix the error of AOT mode on the "i386-windows-msvc" platform

w4454962:
It took a lot of crashes to find the wrong location and fix it.

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -3285,8 +3285,25 @@ aot_call_indirect(WASMExecEnv *exec_env, uint32 tbl_idx, uint32 table_elem_idx,
             cell_num += wasm_value_type_cell_num(ext_ret_types[i]);
         }
 
+#if WASM_ENABLE_AOT_STACK_FRAME != 0
+        void *prev_frame = get_top_frame(exec_env);
+        if (!is_frame_per_function(exec_env)
+            && !aot_alloc_frame(exec_env, func_idx)) {
+            if (argv1 != argv1_buf)
+                wasm_runtime_free(argv1);
+            return false;
+        }
+#endif
         ret = invoke_native_internal(exec_env, func_ptr, func_type, signature,
                                      attachment, argv1, argc, argv);
+#if WASM_ENABLE_AOT_STACK_FRAME != 0
+        /* Free all frames allocated, note that some frames
+           may be allocated in AOT code and haven't been
+           freed if exception occurred */
+        while (get_top_frame(exec_env) != prev_frame)
+            aot_free_frame(exec_env);
+#endif
+
         if (!ret) {
             if (argv1 != argv1_buf)
                 wasm_runtime_free(argv1);
@@ -3333,7 +3350,7 @@ aot_call_indirect(WASMExecEnv *exec_env, uint32 tbl_idx, uint32 table_elem_idx,
            frame-per-function mode the frame is allocated at the
            beginning of the function. */
         if (!is_frame_per_function(exec_env)
-            && !aot_alloc_frame(exec_env, table_elem_idx)) {
+            && !aot_alloc_frame(exec_env, func_idx)) {
             return false;
         }
 #endif
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -3327,8 +3327,25 @@ aot_call_indirect(WASMExecEnv *exec_env, uint32 tbl_idx, uint32 table_elem_idx,
         return true;
     }
     else {
+#if WASM_ENABLE_AOT_STACK_FRAME != 0
+        void *prev_frame = get_top_frame(exec_env);
+        /* Only allocate frame for frame-per-call mode; in the
+           frame-per-function mode the frame is allocated at the
+           beginning of the function. */
+        if (!is_frame_per_function(exec_env)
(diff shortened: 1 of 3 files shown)


lum1n0us:
Could you give some details about the crash and the analysis of the root cause?

w4454962:
1. In the Windows x86-32 environment, the function names returned by LLVMGetSymbolName automatically include an "_" prefix. WAMR had several places missing the logic to skip this prefix during code compilation. I supplemented these missing parts, otherwise the generated AOT code would be incorrect, leading to program unresponsiveness and crashes.
﻿
2. The "wasm_runtime_call_indirect" function crashes when using newly created "wasm_runtime_create_exec_env". The crash occurs because cur_frame is null. Therefore, I added frame creation before calling aot_call_indirect and released it after completion to resolve this crash.
﻿
3. Added missing __alldiv and __allrem functions that were reported as unresolved during compilation.
﻿
4. The machine code generated by LLVM for calling symbols under x86 assumes cdecl calling convention (caller cleans the stack). However, LLVM-generated code doesn't balance the stack pointer. Changed the function convention to "__stdcall" (callee cleans the stack) to resolve abnormal crashes caused by stack imbalance.
﻿
After addressing these issues, my code now runs normally and produces consistent results with the DLL implementation.

w4454962:
Additionally, it is very difficult to find error code in AOT mode, as suspicious locations need to be identified from the assembly code generated by LLVM in order to determine the error point. Very difficult.

lum1n0us (commented):
Perhaps we should create a new helper to wrap `LLVMGetSymbolName()` and packed undecorate underscore operations, and land the new function to replace corresponding callings. Seems similar operations are spreading in `aot_emit_aot_file.c`.

> crash occurs because cur_frame is null. 

Could you provide me with more details about this crash? I am attempting to fully understand the issue.

w4454962:
> 也许我们应该创建一个新的助手来包装`LLVMGetSymbolName()`和打包去装饰下划线操作，并放置新函数来替换相应的调用。似乎类似的操作正在传播`aot_emit_aot_file.c`。
> 
> > 由于 cur_frame 为空，因此发生崩溃。
> 
> 您能为我提供有关此次崩溃的更多详细信息吗？我正在尝试全面了解该问题。

The null cur_frame crash occurs in a  AOT callback function when a new env is created without initializing cur_frame. The LLVM-generated machine code in the callback then accesses this uninitialized pointer, triggering a wild pointer crash‌

w4454962:
> 也许我们应该创建一个新的助手来包装`LLVMGetSymbolName()`和节省去装饰下划线操作，并放置新函数来替换相应的调用。似乎类似的操作正在传播`aot_emit_aot_file.c`。

I have seen a lot of code that judges the '_' underline, some at compile time and some at runtime, and dare not encapsulate it casually. But I also think it should be packaged and unified.

w4454962:
> > 也认为应该打包统一。
> 
> 那么，您想尝试一下然后更新 PR 吗？

I don't plan to update the PR because the current fix is already working properly. Optimizing code design is a separate topic, and I am not familiar enough with this project. It should be done by someone with strong coding skills, as there are too many codes for judging '_' and '_'.

lum1n0us:
If that's the case, would you mind uploading something that can help us reproduce the crash?

w4454962:
> If that's the case, would you mind uploading something that can help us reproduce the crash?

I'm sorry I can't do that, the runtime environment my project depends on is quite complex. In fact, you can reproduce these crashes just by writing a function with clock or the operator functions mentioned in the PR. This is an obvious common-sense error, isn't it?

calvin2021y:
hi @w4454962 

Thanks for the hard work to location the problem, I guess https://github.com/bytecodealliance/wasm-micro-runtime/issues/4103 crash also cause by this problem.

@lum1n0us 

I think we should force llvm generated machine code  use `__cdecl`.   This will fix msvc problem, and allow `aot_reloc_x86_32.c` work for `clang` & `gcc` and linux, macOS, BSD.

And I think windows x86_64 also need same fix (and maybe arm32/arm64/arm64ec).

I guess add option to set `vendor_sys` also will be very helpful like @TianlongLiang suggestion.

w4454962:
> hi @w4454962
> 
> Thanks for the hard work to location the problem, I guess #4103 crash also cause by this problem.
> 
> @lum1n0us
> 
> I think we should force llvm generated machine code use `__cdecl`. This will fix msvc problem, and allow `aot_reloc_x86_32.c` work for `clang` & `gcc` and linux, macOS, BSD.
> 
> And I think windows x86_64 also need same fix (and maybe arm32/arm64/arm64ec).
> 
> I guess add option to set `vendor_sys` also will be very helpful like @TianlongLiang suggestion.

The "Windows x86_64" function convention has been unified into one, and it is important to note that the first four parameters are passed from registers to the function, and a shadow space of 0×20 is provided for each function, which is different from the details that need to be paid attention to in 32-bit

calvin2021y:
hi @w4454962 

Thanks a lot for the explain.

I guess for the x86 arch,  we should keep use `__cdecl` for `aot_reloc_x86_32.c` and `llvm` generate code,  unless it cause unfixable problem. (In order to unify with other platforms),  so I guess if we can change wamrc llvm logic to fix the problem will be a better choice.

`__cdecl`  and `__stdcall` also generate diff symbal name,  maybe relate to `LLVMGetSymbolName` result.

lum1n0us (approved):
LGTM

TianlongLiang (approved):
LGTM

PR #4185, opened 2025-04-08
cmake: Enhance target selection for ARM architectures with FPU

no1wudi:
Improve the target selection logic for ARM architectures in the NuttX platform configuration.
* Added support for FPU detection in THUMB and ARM targets
* Ensured correct target is set based on architecture and configuration options


lum1n0us (approved):
LGTM

TianlongLiang (approved):
LGTM

PR #4189, opened 2025-04-10
debug-engine: fix a few type mismatches

yamt:
debug-engine: fix a few type mismatches

Note: on some targets, int32_t is a long.
for example, GCC shipped with the recent ESP-IDF has such a
configuration. [1]

[1] https://github.com/apache/nuttx/issues/15755#issuecomment-2635652808

cf.
https://github.com/apache/nuttx/pull/16022
https://docs.espressif.com/projects/esp-idf/en/stable/esp32/migration-guides/release-5.x/5.0/gcc.html#espressif-toolchain-changes

also, use strict prototypes as it's complained by the same version of GCC.

### DIFF
--- a/core/iwasm/libraries/debug-engine/handler.c
+++ b/core/iwasm/libraries/debug-engine/handler.c
@@ -34,7 +34,7 @@ static char *tmpbuf;
 static korp_mutex tmpbuf_lock;
 
 int
-wasm_debug_handler_init()
+wasm_debug_handler_init(void)
 {
     int ret;
     tmpbuf = wasm_runtime_malloc(MAX_PACKET_SIZE);
@@ -51,7 +51,7 @@ wasm_debug_handler_init()
 }
 
 void
-wasm_debug_handler_deinit()
+wasm_debug_handler_deinit(void)
 {
     wasm_runtime_free(tmpbuf);
     tmpbuf = NULL;
--- a/core/iwasm/libraries/debug-engine/handler.c
+++ b/core/iwasm/libraries/debug-engine/handler.c
@@ -204,8 +204,7 @@ handle_general_query(WASMGDBServer *server, char *payload)
     if (!strcmp(name, "Supported")) {
         os_mutex_lock(&tmpbuf_lock);
         snprintf(tmpbuf, MAX_PACKET_SIZE,
-                 "qXfer:libraries:read+;PacketSize=%" PRIx32 ";",
-                 MAX_PACKET_SIZE);
+                 "qXfer:libraries:read+;PacketSize=%x;", MAX_PACKET_SIZE);
         write_packet(server, tmpbuf);
         os_mutex_unlock(&tmpbuf_lock);
     }
@@ -384,7 +383,7 @@ send_thread_stop_status(WASMGDBServer *server, uint32 status, korp_tid tid)
 
     if (status == 0) {
         os_mutex_lock(&tmpbuf_lock);
-        snprintf(tmpbuf, MAX_PACKET_SIZE, "W%02x", status);
+        snprintf(tmpbuf, MAX_PACKET_SIZE, "W%02" PRIx32, status);
         write_packet(server, tmpbuf);
         os_mutex_unlock(&tmpbuf_lock);
         return;
@@ -400,8 +399,9 @@ send_thread_stop_status(WASMGDBServer *server, uint32 status, korp_tid tid)
 
     os_mutex_lock(&tmpbuf_lock);
     // TODO: how name a wasm thread?
-    len += snprintf(tmpbuf, MAX_PACKET_SIZE, "T%02xthread:%" PRIx64 ";name:%s;",
-                    gdb_status, (uint64)(uintptr_t)tid, "nobody");
+    len += snprintf(tmpbuf, MAX_PACKET_SIZE,
+                    "T%02" PRIx32 "thread:%" PRIx64 ";name:%s;", gdb_status,
+                    (uint64)(uintptr_t)tid, "nobody");
     if (tids_count > 0) {
         len += snprintf(tmpbuf + len, MAX_PACKET_SIZE - len, "threads:");
         while (i < tids_count) {
@@ -624,7 +624,8 @@ void
 handle_get_write_memory(WASMGDBServer *server, char *payload)
 {
     size_t hex_len;
-    int32 offset, act_len;
+    int offset;
+    int32 act_len;
     uint64 maddr, mlen;
     char *buff;
     bool ret;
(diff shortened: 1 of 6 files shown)


lum1n0us:
> Note: on some targets, int32_t is a long.

How about `uint32`? will it be a problem too ?

yamt:
> > Note: on some targets, int32_t is a long.
> 
> How about `uint32`? will it be a problem too ?

uint32 is an alias of uint32_t. the code this PR fixes is using uint32/int32.

theoretically, it's possible to make uint32 and uint32_t different. but i guess it causes far more issues than fixes.

yamt:
> So, should we use `unsigned int` to replace `uint32` in the function signatures in files located in core/iwasm/libraries/debug-engine/?

why do you think so?

> Plus, I noticed some `int32` in function signatures have been replaced and some are not, including several static variables.

are you suggesting it's a problem? how?

lum1n0us:
I suspect the issue arose when I saw the changes in *core/iwasm/libraries/debug-engine/gdbserver.\**. It appears to be an incomplete modification. Please help to me to understand the situation.

- If treating "int32 like a long" is the problem, then by the same reasoning, "uint32" could also be problematic.
- Plus, some int32 in function signatures have been replaced and some are not, including several static variables.

```shell
$ egrep -rnc "uint32|int32" core/iwasm/libraries/debug-engine
core/iwasm/libraries/debug-engine/packets.c:1
core/iwasm/libraries/debug-engine/debug_engine.cmake:0
core/iwasm/libraries/debug-engine/utils.c:3
core/iwasm/libraries/debug-engine/handler.c:20
core/iwasm/libraries/debug-engine/handler.h:1
core/iwasm/libraries/debug-engine/gdbserver.c:4
core/iwasm/libraries/debug-engine/debug_engine.h:15
core/iwasm/libraries/debug-engine/debug_engine.c:48
core/iwasm/libraries/debug-engine/gdbserver.h:2
core/iwasm/libraries/debug-engine/utils.h:5
core/iwasm/libraries/debug-engine/packets.h:0
```

yamt:
we should not assume `int32` or `int32_t` is `int` because they can be `long` or even other types.

for example, `os_socket_bind` takes `int *port`. so i changed `port` to `int`.

lum1n0us:
> we should not assume int32 or int32_t is int because they can be long or even other types.

Totally agree. If treating "int32 like a long" is the problem, then by the same reasoning, "uint32" could also be problematic. Right?

yamt:
> > we should not assume int32 or int32_t is int because they can be long or even other types.
> 
> Totally agree. If treating "int32 like a long" is the problem, then by the same reasoning, "uint32" could also be problematic. Right?

yes.

lum1n0us:
> > > we should not assume int32 or int32_t is int because they can be long or even other types.
> > 
> > 
> > Totally agree. If treating "int32 like a long" is the problem, then by the same reasoning, "uint32" could also be problematic. Right?
> 
> yes.

Then, two problems:
- should we use unsigned int to replace uint32 in the function signatures in files located in core/iwasm/libraries/debug-engine?
- There are still `int32` in function signatures in *core/iwasm/libraries/debug-engine* should we leave them as they are or use `int` to replace them as well?

yamt:
> > > > we should not assume int32 or int32_t is int because they can be long or even other types.
> > > 
> > > 
> > > Totally agree. If treating "int32 like a long" is the problem, then by the same reasoning, "uint32" could also be problematic. Right?
> > 
> > 
> > yes.
> 
> Then, two problems:
> 
>     * should we use unsigned int to replace uint32 in the function signatures in files located in core/iwasm/libraries/debug-engine?
> 
>     * There are still `int32` in function signatures in _core/iwasm/libraries/debug-engine_ should we leave them as they are or use `int` to replace them as well?

i don't see any reasons to replace them all.
they are not problems as far as they are used consistently.

lum1n0us:
Certainly. Just out of curiosity, if the next contributor intends to add a new function in *core/iwasm/libraries/debug-engine/gdbserver.h*, should they use `int/unsigned int` or `int32/uint32`? What's the general guideline here?

yamt:
> Certainly. Just out of curiosity, if the next contributor intends to add a new function in _core/iwasm/libraries/debug-engine/gdbserver.h_, should they use `int/unsigned int` or `int32/uint32`? What's the general guideline here?

nothing specific to debug-engine.
the common problematic pattern is `int32 *` vs `int *`.

lum1n0us (approved):
LGTM

core/iwasm/libraries/debug-engine/handler.c
```diff
@@ -205,7 +205,7 @@ handle_general_query(WASMGDBServer *server, char *payload)
         os_mutex_lock(&tmpbuf_lock);
         snprintf(tmpbuf, MAX_PACKET_SIZE,
                  "qXfer:libraries:read+;PacketSize=%" PRIx32 ";",
-                 MAX_PACKET_SIZE);
+                 (uint32)MAX_PACKET_SIZE);
```
xiaoxiang781216, line 208: why not change PRIx32 to %xd

yamt: do you mean `%x`?
it's a valid alternative.
well, theoretically MAX_PACKET_SIZE can be of other type as it's user-overridable.
but i don't have a strong preference for one way or another.

xiaoxiang781216: > do you mean `%x`?

Yes.

> it's a valid alternative. well, theoretically MAX_PACKET_SIZE can be of other type as it's user-overridable.

I assume MAX_PACKET_SIZE  doesn't add any suffix(the integer constant without any suffix is int).

> but i don't have a strong preference for one way or another.

 I prefer the code doesn't have the cast as much as possible.

yamt: > > do you mean `%x`?
> 
> Yes.
> 
> > it's a valid alternative. well, theoretically MAX_PACKET_SIZE can be of other type as it's user-overridable.
> 
> I assume MAX_PACKET_SIZE doesn't add any suffix(the integer constant without any suffix is int).
> 
> > but i don't have a strong preference for one way or another.
> 
> I prefer the code doesn't have the cast as much as possible.

ok. i'm fine with either ways. let me change it.

yamt: done.

xiaoxiang781216 (approved):
LGTM

xiaoxiang781216:
@no1wudi could you merge this patch to unblock ci on nuttx side?

TianlongLiang (approved):
LGTM

PR #4193, opened 2025-04-13
Remove the dlen to optimize it.

lum1n0us:
There are two reasons for this optimization:

- The value of dlen can equal 0x1_0000_0000, even in wasm32 mode, because it is derived from (4G-0). This results in a truncation when it is passed to b_memmove_s(). Consequently, s1max becomes 0 and n is greater than s1max. To correct this, a longer type is required.
- The dlen is only used to check if there is enough space in b_memmove_s(). However, from a different angle, after confirming that both src+len and dst+len are within the memory range, we can be assured and there is no need for this explicit check.


xujuntwt95329 (approved):
LGTM

PR #4196, opened 2025-04-14
fix false native stack overflow detections with HW_BOUND_CHECK

yamt:
In call_wasm_with_hw_bound_check/call_native_with_hw_bound_check, ensure to set up the stack boundary (wasm_exec_env_set_thread_info) before checking the overflow.

It seems that the problem was introduced by:
https://github.com/bytecodealliance/wasm-micro-runtime/pull/2940


TianlongLiang (approved):
LGTM

lum1n0us (approved):
LGTM

PR #4197, opened 2025-04-14
fix: Remove unused variables in SIMD_v128_const case

no1wudi:
Fix compiler warnings about unused variables `high` and `low` in the `SIMD_v128_const` case. These variables are only needed inside the `WASM_ENABLE_FAST_INTERP != 0` conditional block, but were incorrectly declared outside of it, causing unused variable warnings.


lum1n0us (approved):
LGTM

TianlongLiang (approved):
LGTM

PR #4202, opened 2025-04-16
Insert a missing % character into the format string.

lum1n0us:
leads to wrong type of arguments to formatting function.

### DIFF
--- a/core/iwasm/aot/aot_perf_map.c
+++ b/core/iwasm/aot/aot_perf_map.c
@@ -31,8 +31,15 @@ get_func_size(const AOTModule *module, struct func_info *sorted_func_ptrs,
 static int
 compare_func_ptrs(const void *f1, const void *f2)
 {
-    return (intptr_t)((struct func_info *)f1)->ptr
-           - (intptr_t)((struct func_info *)f2)->ptr;
+    uintptr_t ptr1 = (uintptr_t)((struct func_info *)f1)->ptr;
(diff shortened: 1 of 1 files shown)


core/iwasm/aot/aot_perf_map.c
```diff
@@ -31,8 +31,8 @@ get_func_size(const AOTModule *module, struct func_info *sorted_func_ptrs,
 static int
 compare_func_ptrs(const void *f1, const void *f2)
 {
-    return (intptr_t)((struct func_info *)f1)->ptr
-           - (intptr_t)((struct func_info *)f2)->ptr;
+    return (int)((intptr_t)((struct func_info *)f1)->ptr
+                 - (intptr_t)((struct func_info *)f2)->ptr);
```
yamt, line 35: the original code looks broken. this change hides the issue.

lum1n0us: 🤔 it is better?

```c
{
    void *ptr1 = ((struct func_info *)f1)->ptr;
    void *ptr2 = ((struct func_info *)f2)->ptr;

    if (ptr1 < ptr2)
        return -1;
    else if (ptr1 > ptr2)
        return 1;
    else
        return 0;
}
```

yamt (approved):
lgtm

TianlongLiang (approved):
LGTM

Issue #4173, opened 2025-03-31
WAMR `iwasm` interpreter produces incorrect output on SIMD WebAssembly module


khagankhan:
### Subject of the issue

WAMR simple interpreter `iwasm` produces a different result from other runtimes most likely because of SIMD instructions.

### Test case

[discrepancy.wat.txt](https://github.com/user-attachments/files/19538375/discrepancy.wat.txt)

[discrepancy.wasm.txt](https://github.com/user-attachments/files/19538383/discrepancy.wasm.txt)

Please remove .txt extesnions when running the WASM modules.

### Your environment

* `uname -a`:  x86_64
* `lsb_release -a`: Ubuntu 22.04.5 LTS
* `iwasm --versions`: iwasm 2.2.0
* Wamr built: 
```bash
cd /path/to/wasm-micro-runtime/product-mini/platforms/linux/
mkdir build && cd build
cmake -DBUILD_SHARED_LIBS=ON ..
make -j
```

## Steps to reproduce

Please download the `discrepancy.wat.txt` and `discrepancy.wasm.txt` files. Remove `.txt` extensions and run it with iwasm:

```bash
wasm-micro-runtime/product-mini/platforms/linux/build/iwasm -f _main discrepancy.wasm
```

## Expected behavior

iwasm is supposed to produce value zero (0) as `_main` return like other runtimes:

```bash
wasmtime run --invoke _main discrepancy.wasm
warning: using `--invoke` with a function that returns values is experimental and may break in the future
0
```

## Actual behavior

iwasm differs from other runtimes:

```bash
CRC mismatch detected! Wasmer: bf56788d, WABT: bf56788d, Wasmtime: bf56788d, WAMR: 9379f56d, WasmEdge: bf56788d
```
I reduced the test case and removed the host functions that I have used for differential testing. And I think it stems from SIMD instructions. When you run discrepancy.wasm iwasm will produce 1 instead of 0:

```bash
wasm-micro-runtime/product-mini/platforms/linux/build/iwasm -f _main discrepancy.wasm
0x1:i32
```

Despite shrinking the test case using wasm-tools and manually, the best I got was 200 lines. However, most functions can be easily inlined and unreachable ones may be removed. Please tell me if more info is needed.

## Extra Info
Wasmoi fuzzer, an Xsmith-based fuzzer produced the snippet of code that found the issue. https://www.flux.utah.edu/project/xsmith

lum1n0us:
[...]

PR #4203, opened 2025-04-17
Add missing V128 handling in WASM_OP_BR, reported in #4173

jammar1:


loganek:
Thanks, that looks good. Do you think it'd be good to add tests either in WAMR or to the spec test repo?

jammar1:
I added a small example in the linked issue. I can add that as a test but it's hard to see how it can re-occur. I'll also need to add a similar statement for when WASM_ENABLE_GC is true.

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -1053,6 +1053,11 @@ copy_stack_values(WASMModuleInstance *module, uint32 *frame_lp, uint32 arity,
                         frame_lp + dst_offsets[0],                          \
                         GET_I64_FROM_ADDR(frame_lp + src_offsets[0]));      \
                 }                                                           \
+                else if (cells[0] == 4) {                                   \
+                    PUT_V128_TO_ADDR(                                       \
+                        frame_lp + dst_offsets[0],                          \
+                        GET_V128_FROM_ADDR(frame_lp + src_offsets[0]));     \
+                }                                                           \
```
lum1n0us, line 1060: better to have an `else` branch to explicitly capture other cases.

jammar1: What would we want to do in that case? wasm_set_exception?

lum1n0us: If the `else` represents an unexpected case, setting `got_exception` is reasonable. Otherwise, leaving comments is fine by me

core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -1013,6 +1013,22 @@ copy_stack_values(WASMModuleInstance *module, uint32 *frame_lp, uint32 arity,
...
+                    if (src_offsets[0] >= 0) {                             \
+                        CLEAR_FRAME_REF((unsigned)src_offsets[0]);         \
+                        CLEAR_FRAME_REF((unsigned)(src_offsets[0] + 1));   \
+                        CLEAR_FRAME_REF((unsigned)(src_offsets[0] + 2));   \
+                        CLEAR_FRAME_REF((unsigned)(src_offsets[0] + 3));   \
+                        SET_FRAME_REF((unsigned)dst_offsets[0]);           \
+                        SET_FRAME_REF((unsigned)(dst_offsets[0] + 1));     \
+                        SET_FRAME_REF((unsigned)(dst_offsets[0] + 2));     \
+                        SET_FRAME_REF((unsigned)(dst_offsets[0] + 3));     \
+                    }                                                      \
```
lum1n0us, line 1030: better to have an else branch to explicitly capture other cases.

core/iwasm/interpreter/wasm_interp_fast.c
jammar1: I'm happy to add it, but there's already an assert on the possible values of cells in core/iwasm/interpreter/wasm.h line 1246, in wasm_value_type_size_internal so not sure it's necessary.

lum1n0us: I've been considering those negative fuzzing scenarios. But you're correct. They certainly require a better design to prevent such issues from the start.

lum1n0us (approved):
LGTM

loganek (approved):
LGTM

Issue #4192, opened 2025-04-12
Memory leak in libc-wasi posix.c fd_prestats_destroy


redrezo:
### Subject of the issue
There is a memory leak in `fd_prestats_destroy`. The lock is only released if `pt->prestats` is set. However `fd_prestats_init` always creates this lock and initializes prestats to NULL.
file: `core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c`

```c
bool
fd_prestats_init(struct fd_prestats *pt)
{
    if (!rwlock_initialize(&pt->lock))
        return false;
    pt->prestats = NULL;
    pt->size = 0;
    pt->used = 0;
    return true;
}
```
```c
void
fd_prestats_destroy(struct fd_prestats *pt)
{
    if (pt->prestats) {
        for (uint32 i = 0; i < pt->size; i++) {
            if (pt->prestats[i].dir != NULL) {
                wasm_runtime_free((void *)pt->prestats[i].dir);
            }
        }
        // rwlock_destroy(&pt->lock);
        wasm_runtime_free(pt->prestats);
    }
    rwlock_destroy(&pt->lock); // <- moved it out of the if to fix it locally
}
```

### Your environment

esp32s3 esp-idf

### Steps to reproduce

Found by instantiating and directly deinstantiating a module.

### Expected behavior

The memory to be freed ;-)

### Actual behavior

about 216 bytes are leaked on the esp-idf platform

TianlongLiang:
You are right about this bug. It would cause undefined behavior without destruction. Would you submit a PR to fix it?

PR #4205, opened 2025-04-18
Fix potiential memory leak in wasi libc

TianlongLiang:
Fix issues like https://github.com/bytecodealliance/wasm-micro-runtime/issues/4192


wenyongh (approved):

lum1n0us (approved):
LGTM

PR #4212, opened 2025-04-24
Enable support for version specification inside `samples/cmake/FindWASISDK.cmake`

Zelzahn:
Currently, it is not possible to select a specific WASI SDK version to use in a sample. Thus, I modified the `FindWASISDK.cmake` to let me do this. 

I'm no CMake expert, so feedback is greatly appreciated.

### DIFF
--- a/samples/cmake/FindWASISDK.cmake
+++ b/samples/cmake/FindWASISDK.cmake
@@ -4,9 +4,23 @@
 include(FindPackageHandleStandardArgs)
 
 file(GLOB WASISDK_SEARCH_PATH "/opt/wasi-sdk-*")
+
+function(check_version validator_result_var item)
+  if(WASISDK_FIND_VERSION)
+    if(item MATCHES "${WASISDK_FIND_VERSION}")
+      set(${validator_result_var} TRUE PARENT_SCOPE)
+    else()
+      set(${validator_result_var} FALSE PARENT_SCOPE)
+    endif()
+  else()
+    set(${validator_result_var} TRUE PARENT_SCOPE)
+  endif()
+endfunction()
+
 find_path(WASISDK_HOME
   NAMES share/wasi-sysroot
   PATHS ${WASISDK_SEARCH_PATH}
+  VALIDATOR check_version
   NO_DEFAULT_PATH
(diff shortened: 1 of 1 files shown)


lum1n0us:
> have multiple wasi-sdk's installed, 
I have to admit this is something new to me. Wow.

If you refer to [the installation script for CI](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/.github/actions/install-wasi-sdk-wabt/action.yml), you will find the assumption about wasi-sdk is that there is always a symlink named */opt/wasi-sdk* targeting a directory */opt/wasi-sdk-xxx*. With that, I guess we can modify the FindWASISDK in the following ways:

- Find (the symlink) "/opt/wasi-sdk" instead of (the directory) "/opt/wasi-sdk-*".
- Follow the symlink of */opt/wasi-sdk* to determine the actual */opt/wasi-sdk-xxx*, and use xxx as the version name to compare with the requirement.
- 
FindWASISDK.cmake should not search for the correct wasi-sdk if there are multiple versions. Developers should prepare for that.

samples/cmake/FindWASISDK.cmake
```diff
@@ -4,9 +4,23 @@
...
+    endif()
+  else()
+    set(${validator_result_var} TRUE PARENT_SCOPE)
+  endif()
+endfunction()
+
 find_path(WASISDK_HOME
   NAMES share/wasi-sysroot
   PATHS ${WASISDK_SEARCH_PATH}
+  VALIDATOR check_version
```
yamt, line 23: VALIDATOR is documented as a cmake 3.25 feature.
https://cmake.org/cmake/help/latest/command/find_path.html

in most places we still use `cmake_minimum_required (VERSION 3.14)`.

Zelzahn:
Yeah, I need to have multiple `wasi-sdk`'s installed because I'm toying with some old projects. 

> FindWASISDK.cmake should not search for the correct wasi-sdk if there are multiple versions. Developers should prepare for that.

I agree, I propose that I then close this PR and open a new one where I modify `FindWASISDK.cmake` to follow the symlink?

PR #4218, opened 2025-04-25
fix print_help when libc wasi is enabled

Septa2112:
With WASI libc enabled, the default maximum heap size in bytes  is 0 KB. Therefore, I updated `print_help` to improve the accuracy of the help message.


TianlongLiang (approved):
LGTM

PR #4213, opened 2025-04-25
build-scripts/build_llvm.py: bump to llvm 19

yamt:

PR #4219, opened 2025-04-25
LLVM: don't verify instcombine fixpoint

yamt:
LLVM 18 and later, instcombine perfoms only one iteration. it performs extra "verify fixpoint" operation when instcombine is specified in certain ways, including how we do so here. a problem is that the verification raises a fatal error when it finds we didn't reach a fixpoint:

    LLVM ERROR: Instruction Combining did not reach a fixpoint
    after 1 iterations

while it should be rare, it's quite normal not to reach a fixpoint. this commit fixes the issue by simply disabing the verification.

cf. https://github.com/llvm/llvm-project/commit/41895843b5915bb78e9d02aa711fa10f7174db43


yamt:
extracted from https://github.com/bytecodealliance/wasm-micro-runtime/pull/4213 as i feel it's independent.

lum1n0us (approved):
LGTM

PR #4220, opened 2025-04-25
LLVMCreateTargetMachineWithOpts: disable large data

yamt:
for x86-64, llvm 17 and later sometimes uses "l" prefix for data sections.
cf. https://github.com/llvm/llvm-project/commit/43249378da67319906cf04f2c6cd38df141f3bf6

because our aot file emitter/loader doesn't support such sections, it ends up with load-time errors solving symbols like ".lrodata".

this commit fixes it by avoid placing data in the large data sections.

references:
https://groups.google.com/g/x86-64-abi/c/jnQdJeabxiU https://github.com/llvm/llvm-project/commit/1feb00a28c9fdab162da08a15fcc9d088a36c352


yamt:
extracted from https://github.com/bytecodealliance/wasm-micro-runtime/pull/4213 as i feel it's independent.

lum1n0us (approved):
LGTM

TianlongLiang (approved):
LGTM

PR #4222, opened 2025-04-27
Add error handling for sgx ci

lum1n0us:
> Process completed with exit code 143.

It is a known issue with GitHub-hosted runners. Usually, increasing the swap file can help avoid it. However, sometimes error 143 still occurs. To prevent confusion, let's capture error 143 and allow the CI to pass.

### DIFF
--- a/.github/workflows/compilation_on_sgx.yml
+++ b/.github/workflows/compilation_on_sgx.yml
@@ -290,6 +290,28 @@ jobs:
 
       - name: run spec tests
         run: |
+          set +e
           source /opt/intel/sgxsdk/environment
-          ./test_wamr.sh ${{ matrix.test_option }} -t ${{ matrix.running_mode }}
+          attempts=0
+          max_attempts=3
+
+          while [ $attempts -lt $max_attempts ]; do
+            ./test_wamr.sh ${{ matrix.test_option }} -t ${{ matrix.running_mode }}
+            exitcode="$?"
+
+            if [ $exitcode -eq 0 ]; then
+              echo "Spec test passed"
+              exit 0
+            elif [ $exitcode -ne 143 ]; then
+              echo "Spec test failed with error code $exitcode"
+              exit 1
+            fi
+
+            echo "$exitcode is a known GitHub-hosted runner issue"
+            echo "::notice::Re-running the spec test due to error code 143"
+            attempts=$((attempts + 1))
+          done
+
+          echo "::notice::Report an error with code 143 in SGX CI after $max_attempts attempts"
+          exit 143
         working-directory: ./tests/wamr-test-suites


TianlongLiang (approved):
LGTM

.github/workflows/compilation_on_sgx.yml
```diff
@@ -290,6 +290,22 @@ jobs:
...
         run: |
+          set +e
           source /opt/intel/sgxsdk/environment
           ./test_wamr.sh ${{ matrix.test_option }} -t ${{ matrix.running_mode }}
+          exitcode="$?"
+
+          if [ $exitcode -eq 143 ]; then
+            echo "$exitcode is a known GitHub-hosted runner issue"
+            echo "::notice::Bypass an error with code 143 in SGX CI"
+            exit 0
```
yamt, line 301: doesn't this risk us to merge broken changes?

lum1n0us: I believe it will not introduce any other unexpected breaking changes.

- Only bypass exit code 143. Other non-zero codes will still be captured.
- Spec_test scripts, all.py and runtest.py, will clamp return codes, and 143 is not one of them.
- 143 is not an errno in errno.h

yamt: well, my concern is:

1. someone submits a PR with broken changes
2. the ci happened to pass because of this workaround
3. reviewers don't notice it and merge

yamt:
> It is a known issue with GitHub-hosted runners. Usually, increasing the swap file can help avoid it.

do you have a reference to a bug ticket or something?

.github/workflows/compilation_on_sgx.yml
lum1n0us: Based on my observations, if there is an error from the spec-test itself, 143 won't be generated. I believe the 143 code will not cover other errors. Those accidental 143 errors are quite annoying, so I want to relieve our daily work. If directly bypassing is too aggressive, how about automatically rerunning if it fails? 

- https://blog.logto.io/automatic-github-workflow-rerun
- https://github.com/actions/runner-images/issues/6680

xujuntwt95329 (approved):
LGTM

.github/workflows/compilation_on_sgx.yml
yamt: > Those accidental 143 errors are quite annoying, so I want to relieve our daily work.

i agree.

> If directly bypassing is too aggressive, how about automatically rerunning if it fails?

it seems nicer.

TianlongLiang:
> > It is a known issue with GitHub-hosted runners. Usually, increasing the swap file can help avoid it.
> 
> do you have a reference to a bug ticket or something?

https://github.com/actions/runner-images/issues/6680, there is also an approach we can try: adding `-large` suffix to ubuntu runner(`runs-on: ubuntu-latest-large //THIS LINE HERE`)

yamt:
> > > It is a known issue with GitHub-hosted runners. Usually, increasing the swap file can help avoid it.
> > 
> > 
> > do you have a reference to a bug ticket or something?
> 
> [actions/runner-images#6680](https://github.com/actions/runner-images/issues/6680), there is also an approach we can try: adding `-large` suffix to ubuntu runner(`runs-on: ubuntu-latest-large //THIS LINE HERE`)

large runners require some extra money, don't they?

lum1n0us:
Larger runners have more resources than the standard GitHub-hosted runners, but they require access granted by Enterprise administrators or organization owners. Additionally, they are not free.

- https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#larger-runners
- https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners/managing-larger-runners#allowing-repositories-to-access-larger-runners
- https://docs.github.com/en/billing/managing-billing-for-your-products/managing-billing-for-github-actions/about-billing-for-github-actions#per-minute-rates

yamt:
maybe this is not working as intended?
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/15381146513/job/43272172295?pr=4307

PR #4223, opened 2025-04-28
Use execve() to replace system()

lum1n0us:
- Direct Execution: execve() directly executes a program, bypassing the shell. This avoids vulnerabilities like shell injection, which can occur with system() if user input is not properly sanitized.
- Controlled Environment: With execve(), you can explicitly specify the environment variables for the new process, providing better control over the execution context.
- No Shell Overhead: execve() does not invoke a shell, reducing the risk of unintended behavior caused by shell features or configurations.
- Predictable Behavior: execve() only executes the specified program, whereas system() relies on the shell, which may interpret commands differently based on the shell's configuration or environment.


core/shared/utils/bh_common.c
```diff
@@ -167,15 +167,39 @@ wa_strdup(const char *s)
...
+
+        /* The `argv[]` must be terminated by a NULL pointer. */
+        if (argv[argc - 1] != NULL) {
+            return -1;
+        }
+    }
 
 #if !(defined(_WIN32) || defined(_WIN32_))
-    ret = system(cmd);
+    ret = execve(pathname, argv, envp);
```
yamt, line 194: does this work at all? i suppose you need fork+execve+wait.

core/shared/utils/bh_common.c
```diff
@@ -167,15 +167,39 @@ wa_strdup(const char *s)
 }
 
 #if WASM_ENABLE_WAMR_COMPILER != 0 || WASM_ENABLE_JIT != 0
+/* need to make sure that The `argv[]` must be terminated by a NULL pointer. */
 int
-bh_system(const char *cmd)
+bh_system(const char *pathname, char *const argv[], int argc)
```
yamt, line 172: i feel the name bh_system is confusing if it doesn't work like system().

core/shared/utils/bh_common.c
lum1n0us: > execve() executes the program referred to by pathname.  This causes the program that is currently being run by the calling process to be replaced with a new program, with newly initialized stack, heap, and (initialized and uninitialized) data segments.

IMU, in our scenarios, there is no need to fork a new process and run commands. The default behavior is sufficient

core/shared/utils/bh_common.c
lum1n0us: Got you points. I will rename it to bh_execve().

core/shared/utils/bh_common.c
yamt: i don't think so. wamrc has other things to do after invoking these external commands.
have you tested?

lum1n0us: Yes, I used commands like `WMRC_LLC_COMPILER=/SDF/051_llvm_releases/clang+llvm-15.0.6-x86_64-linux-gnu-ubuntu-18.04/bin/clang WAMRC_LLC_FLAGS="-mcmodel=medium" /workspaces/wasm-micro-runtime/wamr-compiler/build/wamrc -o xyz.aot ./xyz.wasm`.

xujuntwt95329: I‘m also curious about how this works. The man page says:
> execve() does not return on success, and the text, initialized data, uninitialized data (bss),  and  stack  of
       the calling process are overwritten according to the contents of the newly loaded program.

That means the later instructions after execve will not be executed.

lum1n0us: Thanks for reminder. Now I see both of your points.

PR #4224, opened 2025-04-28
wamrc: add --disable-llvm-jump-tables option

yamt:
while ideally a user should not need to care this kind of optimization details, in reality i guess it's sometimes useful. both of clang and GCC expose a similar option.  (-fno-jump-tables)


lum1n0us (approved):
LGTM.

no1wudi (approved):

PR #4225, opened 2025-04-28
platform/nuttx: Fix dcache operation in os_dcache_flush

no1wudi:
Replace up_invalidate_dcache_all() with up_flush_dcache_all() in os_dcache_flush() to properly flush the data cache instead of just invalidating it. This ensures that any modified data in the cache is written back to memory before execution.


lum1n0us (approved):
LGTM

wenyongh (approved):
LGTM

PR #4227, opened 2025-04-28
set default value of `WAMR_BUILD_REF_TYPES` to 1 in standalone cases

Septa2112:
During testing of the standalone, it was found that the `test_module_malloc` case, which is quite old, did not set `WAMR_BUILD_REF_TYPES=1`, resulting in test failures. 

It is considered to set a default value for `WAMR_BUILD_REF_TYPES` in `build-scripts/config_common.cmake` to prevent such situations.


yamt:
are you using llvm-19?

Septa2112:
@yamt I'm not using llvm-19. This is just something I discovered during testing when the `standalone/test_module_malloc` failed. After adding `-DWAMR_BUILD_REF_TYPES=1` to the test case, everything worked fine.

Whether to add a default value is debatable, but it seems necessary to add this option to the test case?

yamt:
> @yamt I'm not using llvm-19. This is just something I discovered during testing when the `standalone/test_module_malloc` failed. After adding `-DWAMR_BUILD_REF_TYPES=1` to the test case, everything worked fine.

test_module_malloc uses /opt/wasi-sdk to build a module.
which version of wasi-sdk is it?

> Now it seems that some checks of unit tests are failing. Whether to add a default value is debatable, but it seems necessary to add this option to `test_module_malloc`?

the lastest iteration of the topic i remember is: https://github.com/bytecodealliance/wasm-micro-runtime/pull/3978
at that time, we chose not to enable it globally.

Septa2112:
@yamt Sorry, that's my mistake. My `wasi-sdk` version is 25, and it shows LLVM 19, as shown below. And I also looked into the issue #3978, the reference types option definitely shouldn't be enabled by default.
```
25.0
wasi-libc: 574b88da4815
llvm: ab4b5a2db582
llvm-version: 19.1.5
config: f992bcc08219
```
------
But without any changes, the test case fails, which shouldn't happen.
```bash
Create AoT compiler with:
  target:        x86_64
  target cpu:    rocketlake
  target triple: x86_64-unknown-linux-gnu
  cpu features:
  opt level:     3
  size level:    3
  output format: AoT file
Compile success, file wasm-app/test.aot was generated.
AOT module load failed: reference types is not enabled in this build

```

Septa2112:
This change has been moved to `CMakeLists.txt`, and corresponding modifications have been made to other standalone cases.

TianlongLiang (approved):
LGTM

PR #4229, opened 2025-04-28
Improve readlinkat_dup() to handle symlink size correctly

lum1n0us:


core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -1222,43 +1222,56 @@ __wasi_errno_t
...
-    size_t len_org = len;
+    __wasi_errno_t error;
+    struct __wasi_filestat_t stat = { 0 };
+    size_t buf_len;
+
+    /*
+     * use fstatat to get a better estimation
+     * If path is a symbolic link, do not dereference it:
+     * instead return information about the link itself,
+     * like lstat().
```
yamt, line 1233: i'm not sure if this change is an improvement.
this adds another race window to the path_get logic, which is security-sensitive and already complex.

do you have actual use cases where this change matters?

lum1n0us: Initially, I encountered the same race condition problem as the original implementation. This is simply because it is a try-read-loop, making it easy to modify the symlink or target content between two `os_readlinkat()` calls. 

The addition of `fstat()` is only used to get a better starting point than 32. If `st_size` is small, everything returns to the original state. However, if `st_size` is larger, it can save several loop cycles. I feel it is better than a wild guess(of 32).

xujuntwt95329 (approved):
LGTM

TianlongLiang (approved):
LGTM

PR #4231, opened 2025-04-29
bypass vptr sanitizer

lum1n0us:
LLVM, by default, disables the use of C++'s built-in Run-Time Type Information. This decision is primarily driven by concerns about code size and efficiency.

But '-fsanitize=vptr' not allowed with '-fno-rtti'.


xujuntwt95329 (approved):
LGTM

PR #4238, opened 2025-05-02
fix return types of our 64-bit clz/ctz/popcount intrinsics

yamt:
the corresponding LLVM intrinsics' return types are same as their first argument. eg. i64 for llvm.cttz.i64.
cf. https://llvm.org/docs/LangRef.html#llvm-cttz-intrinsic

this commit changes the return types of our versions of the intrinsics to match llvm versions as our aot compiler, specifically __call_llvm_intrinsic, assumes.

strictly speaking, this is a potential AOT ABI change. however, I suppose it isn't a problem for many of 64-bit ABIs out there, where (lower half of) a 64-bit register is used to return a 32-bit value anyway.  (for such ABIs, this commit would fix the upper 32-bit value of the register.)

### DIFF
--- a/core/iwasm/aot/aot_intrinsic.c
+++ b/core/iwasm/aot/aot_intrinsic.c
@@ -194,7 +194,7 @@ aot_intrinsic_clz_i32(uint32 type)
     return num;
 }
 
-uint32
+uint64
 aot_intrinsic_clz_i64(uint64 type)
 {
     uint32 num = 0;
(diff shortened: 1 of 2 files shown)


yamt:
restarted a failed job, sgx 143

wenyongh (approved):
LGTM

xujuntwt95329 (approved):
LGTM

PR #4239, opened 2025-05-02
samples/wasm-c-api: skip aot compilation unless necessary

yamt:


samples/wasm-c-api/CMakeLists.txt
```diff
@@ -129,7 +129,7 @@ if (${WAT2WASM_VERSION} VERSION_LESS 1.0.26)
   set(WAT2WASM_FLAGS "--enable-reference-types")
 endif ()
 
-if(${WAMR_BUILD_AOT} EQUAL 1)
+if(${WAMR_BUILD_AOT} EQUAL 1 AND ${WAMR_BUILD_INTERP} EQUAL 0)
```
lum1n0us, line 132: Would it be better if we provide users with a hint like this somewhere in this CMakeLists.txt?

``` cmake
message("In this sample, we can only enable a single option between DWAMR_BUILD_INTERP and DWAMR_BUILD_AOT.")
```

xujuntwt95329 (approved):
LGTM

samples/wasm-c-api/CMakeLists.txt
yamt: i feel it belongs to README.

lum1n0us (approved):
LGTM

PR #4240, opened 2025-05-02
teach aot emitter/loader about .srodata and .srodata.cst* sections

yamt:
LLVM 19 and later started to use srodata ("small read only data") sections for RISCV.  cf. https://github.com/llvm/llvm-project/pull/82214 this commit makes our aot emitter/loader deal with those sections.

an alternative would be to disable small data sections completely by setting the "SmallDataLimit" module attribute to zero. however, i feel this commit is more straightforward and consisitent as we are already dealing with sdata sections.


lum1n0us (approved):
LGTM.

---
Maybe we can take some benefits from the *srodata* sections. Let's keep it that way.

wenyongh (approved):
LGTM

PR #4248, opened 2025-05-06
riscv: avoid llvm.cttz.i32/i64 for xip

yamt:
LLVM 16 and later expands cttz intrinsic to a table lookup, which involves some relocations. (unless ZBB is available, in which case the native instructions are preferred over the table-based lowering.)

cf. https://reviews.llvm.org/D128911


lum1n0us (approved):
LGTM

no1wudi (approved):

PR #4249, opened 2025-05-06
samples/wasm-c-api: remove unused valgrind detection

yamt:
* it's unused

* valgrind is basically a linux-only software. it isn't a good idea to make it a hard requirement. if we want to use valgrind, it's better to introduce a separate option to control it.


lum1n0us (approved):
LGTM

TianlongLiang (approved):
LGTM

xujuntwt95329 (approved):
LGTM

PR #4251, opened 2025-05-07
disable compiler to prevent get_current_target() crash

kylo5aby:
fixes a crash issue discovered  where the runtime
aborts when WAMR_BUILD_AOT=1 and WAMR_BUILD_WAMR_COMPILER=1 are both enabled.


yamt:
> when AOT and compiler both enabled

afaik, currently we don't have anything using both of them.
if you have a use case and want to add a support of such a configuration, what about adding samples and/or tests?
otherwise, i guess it will be broken again soon.

lum1n0us (approved):
LGTM

lum1n0us (changes requested):

TianlongLiang (approved):
LGTM

PR #4252, opened 2025-05-07
JIT: don't join worker threads twice

yamt:
in case of WASM_ENABLE_LAZY_JIT==0, compile_jit_functions should have already joined these threads. joining them again here is an undefined behavior.


lum1n0us (approved):
LGTM

TianlongLiang (approved):
LGTM

Issue #4244, opened 2025-05-05
Error in Python Language Binding build Instructions ? - Fails to Generate Shared Object


woodsmc:
### Instructions for Building the WAMR API for python do not work / contain missing information / typeo's
I've been attempting to create python bindings by hand outside of the build container, and I'm running into some issues. It looks like there maybe missing steps or typeos in the instructions that have been provided. I wanted to write down what I'm seeing. If I am missing something, please do let me know.

#### Details
When you follow the instructions given here: 
https://github.com/bytecodealliance/wasm-micro-runtime/tree/main/language-bindings/python/wamr-api
The tooling fails with an error message, the bindings are not generated.

### Test case
Follow the instructions below and on the [page](https://github.com/bytecodealliance/wasm-micro-runtime/tree/main/language-bindings/python/wamr-api).

### Your environment
* Host OS 
   * Windows Subsystem for Linux, Ubuntu 24.04.2 LTS
   * Python 3.12.3
* WAMR version: Current head branch

### Steps to reproduce
Following the instructions on the page https://github.com/bytecodealliance/wasm-micro-runtime/tree/main/language-bindings/python/wamr-api. 

### Expected behavior

The command:
```bash
pip install -r requirements.txt
```
Should work.
The command to build the native lib and update bindings should work:
```bash
# In WAMR root directory
bash language-bindings/python/utils/create_lib.sh
```
This should build and copy the libiwasm into the package.

### Actual behavior

The command:
```bash
pip install -r requirements.txt
```
Is missing the directory to execute it in. Since the other documentation explicitly states a directory, so should the instructions for this. When executed the following error message appears:

```
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    ...
```

To get this to work the following additional steps are required:
```bash
python3 -m venv venv
source venv/bin/activate     
pip install -r requirements.txt
```

After fixing this the following step also fails:
```bash
# In WAMR root directory
bash language-bindings/python/utils/create_lib.sh
```
The following messages are produced (copied exactly below):

```bash
-- The C compiler identification is GNU 13.3.0
-- The CXX compiler identification is GNU 13.3.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
[...]

PR #4255, opened 2025-05-07
Modifying build flags to ensure libiwasm.so is built

woodsmc:
I added the ` -DBUILD_SHARED_LIBS=ON \` flag to the bash script to ensure that the `libiwasm.so` file was built. This seems to resolve the issues highlighted in issue https://github.com/bytecodealliance/wasm-micro-runtime/issues/4244

I *think* that with this PR we can now close this issue.

The language bindings appear to be correctly generated.

(Just a thought ) -> Do we need any testing for these built scripts? - Perhaps this should be another issue?

### DIFF
--- a/language-bindings/python/utils/create_lib.sh
+++ b/language-bindings/python/utils/create_lib.sh
@@ -17,6 +17,7 @@ cmake \
     -DWAMR_BUILD_LIB_PTHREAD=1 \
     -DWAMR_BUILD_LIB_WASI_THREADS=1 \
     -DWAMR_BUILD_LIB_WASI=1 \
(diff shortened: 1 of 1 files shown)


lum1n0us:
> Do we need any testing for these built scripts? - Perhaps this should be another issue?

You are right. We should add some testing for bindings into the CI.

lum1n0us (approved):
LGTM

TianlongLiang (approved):
LGTM

woodsmc:
Thank you. I'll close the parent issue and open a second on creating test tooling for the build scripts / instructions to ensure that they stay consistent between updates.

PR #4092, opened 2025-02-18
initialize WASI stdio handles to invalid for better error handling

lum1n0us:
fix #4081

PR #4260, opened 2025-05-08
updating WASI stdio handle initialization and build options for UVWASI

lum1n0us:
fix https://github.com/bytecodealliance/wasm-micro-runtime/pull/4092#issuecomment-2861772863


TianlongLiang (approved):
LGTM

PR #4262, opened 2025-05-08
handle multi-byte types correctly in ref.null const expr

kylo5aby:
When processing `ref.null` constant expressions, multi-byte types were incorrectly
being handled even though they are valid types. This could lead to
incorrect reftype initialization for multi-byte types.

This change ensures that both invalid types (!is_byte_a_type) AND valid multi-byte
types are processed through the `type_idx` initialization path


lum1n0us (approved):
LGTM

PR #4264, opened 2025-05-08
Don't call os_thread_get_stack_boundary unless we actually use it

sjamesr:
Previously, if the user sets their own stack boundary, we still compute the thread stack boundary (which is expensive), then immediately discard the result. This change makes the expensive call only if we need it for sure.


lum1n0us (approved):
LGTM

raiderss:
Amazing work! As someone from Eyes Store developing FiveM resources, I find this really inspiring.

TianlongLiang (approved):
LGTM

PR #4261, opened 2025-05-08
Refactor fast-interpreter SIMD compilation flags

lum1n0us:
- enable SIMD flag by default unless hardware limitation
- use SIMDE flag to control fast-interpreter behavior

PR #4266, opened 2025-05-10
Check for WASM_ENABLE_SIMDE in a couple more places

sjamesr:
For WAMR users who don't use cmake, it's possible that WASM_ENABLE_SIMD is set when WASM_ENABLE_SIMDE isn't. This was causing build failures.

### DIFF
--- a/core/iwasm/interpreter/wasm_opcode.h
+++ b/core/iwasm/interpreter/wasm_opcode.h
@@ -790,15 +790,14 @@ typedef enum WASMAtomicEXTOpcode {
 #endif
 #define SET_GOTO_TABLE_ELEM(opcode) [opcode] = HANDLE_OPCODE(opcode)
 
-#if (WASM_ENABLE_JIT != 0 || WASM_ENABLE_FAST_INTERP != 0) \
-    && WASM_ENABLE_SIMD != 0
+#if WASM_ENABLE_SIMDE != 0
 #define SET_GOTO_TABLE_SIMD_PREFIX_ELEM() \
     SET_GOTO_TABLE_ELEM(WASM_OP_SIMD_PREFIX),
 #else
 #define SET_GOTO_TABLE_SIMD_PREFIX_ELEM()
 #endif
 
-#if (WASM_ENABLE_FAST_INTERP != 0) && WASM_ENABLE_SIMD != 0
+#if WASM_ENABLE_SIMD != 0 && WASM_ENABLE_SIMDE != 0
 #define DEF_EXT_V128_HANDLE()                                       \
     SET_GOTO_TABLE_ELEM(EXT_OP_SET_LOCAL_FAST_V128),     /* 0xdd */ \
         SET_GOTO_TABLE_ELEM(EXT_OP_TEE_LOCAL_FAST_V128), /* 0xde */ \


sjamesr:
I'm sorry for burdening you with these kinds of pull requests. Ideally we'd be using CMake as well, but that's unfortunately not an option. I do think this change is reasonable though, these symbols aren't defined if SIMDE isn't available, and that should be OK.

core/iwasm/interpreter/wasm_opcode.h
```diff
@@ -791,14 +791,15 @@ typedef enum WASMAtomicEXTOpcode {
...
+    && WASM_ENABLE_SIMD != 0 && WASM_ENABLE_SIMDE != 0
 #define SET_GOTO_TABLE_SIMD_PREFIX_ELEM() \
     SET_GOTO_TABLE_ELEM(WASM_OP_SIMD_PREFIX),
 #else
 #define SET_GOTO_TABLE_SIMD_PREFIX_ELEM()
 #endif
 
-#if (WASM_ENABLE_FAST_INTERP != 0) && WASM_ENABLE_SIMD != 0
+#if (WASM_ENABLE_FAST_INTERP != 0) && WASM_ENABLE_SIMD != 0 \
+    && WASM_ENABLE_SIMDE != 0
```
lum1n0us, line 802: After #4260, `WASM_ENABLE_SIMDE` represents fast-interpreter + SIMD support, so I guess  `#if WASM_ENABLE_SIMDE !=0 ` is sufficient here

core/iwasm/interpreter/wasm_opcode.h
```diff
@@ -791,14 +791,15 @@ typedef enum WASMAtomicEXTOpcode {
 #define SET_GOTO_TABLE_ELEM(opcode) [opcode] = HANDLE_OPCODE(opcode)
 
 #if (WASM_ENABLE_JIT != 0 || WASM_ENABLE_FAST_INTERP != 0) \
-    && WASM_ENABLE_SIMD != 0
+    && WASM_ENABLE_SIMD != 0 && WASM_ENABLE_SIMDE != 0
```
lum1n0us, line 794: because of the workaround, https://github.com/bytecodealliance/wasm-micro-runtime/blob/WAMR-2.2.0/core/iwasm/interpreter/wasm_interp_classic.c#L6584C1-L6589C7, maybe we can simplify this to `#if WASM_ENABLE_SIMD !=0 || WASM_ENABLE_SIMDE != 0`

sjamesr:
Done, thank you!

sjamesr:
I'm not sure how to resolve the regression this caused, I could use some pointers here.

lum1n0us:
> I'm not sure how to resolve the regression this caused

What kind of regression are we discussing?

sjamesr:
I was just referring to the test failures I got after making your suggested changes, e.g. https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/14980041423/job/42081755685

I'm not sure what combination of #defines are used for each binary target (e.g. wamr, iwasm, etc).

lum1n0us:
Those were my unclear comments. I had assumed that this PR would be based on #4261. I guess you can merge this PR into 4260 locally and try your approach, like in this comparison: https://github.com/bytecodealliance/wasm-micro-runtime/compare/e37d6ba9a50182ea11b1063be64b48364714d49a..5c305beb5b284eb7abd3b9df2217a886d1f66ffd. 

I am going to push the merging of #4260, and then this PR will be ready to merge.

sjamesr:
Ahh, now I understand! Thank you. I rebased my pull request on #4261 and made the changes you suggested, seems to work. :)

lum1n0us (approved):
LGTM

TianlongLiang (approved):
LGTM

PR #4267, opened 2025-05-12
Add "load_by_name" API at wasi-nn 

HongxiaWangSSSS:
When using WASI-NN ,we want to reduce copying the AI model from the host to the WASM by using "load_by_name" API.
We also want to use it to improve the performance, keep the safety and WASI-NN also supports this method on different backends.

I test with 3 tflite models.(x86_64, Ubuntu 22.04)

image

_Both [coco_ssd_mobilenet_v1](https://storage.googleapis.com/download.tensorflow.org/models/tflite/coco_ssd_mobilenet_v1_1.0_quant_2018_06_29.zip) and [coco_ssd_mobilenet_v3](http://download.tensorflow.org/models/object_detection/ssd_mobilenet_v3_small_coco_2020_01_14.tar.gz) are for detection, its file size and input tensor size is different.  
[mobilenet_v2](https://github.com/frogermcs/TFLite-Checker/blob/master/android/ImageClassificationTester/app/src/main/assets/mobilenet_v2_1.0_224.tflite) is for classification and size is more bigger._

The time consumed does not show a linear growth as the file size increases. 
For most cases, load_by_name will more faster whether the load tflite or the entire inference process (load+ set input +compute +get output).

### DIFF
--- a/core/iwasm/libraries/wasi-nn/include/wasi_nn.h
+++ b/core/iwasm/libraries/wasi-nn/include/wasi_nn.h
@@ -30,7 +30,7 @@ load(graph_builder_array *builder, graph_encoding encoding,
     __attribute__((import_module("wasi_nn")));
 
 wasi_nn_error
-load_by_name(const char *name, graph *g)
+load_by_name(char *name, uint32_t name_len, graph *g)
     __attribute__((import_module("wasi_nn")));
 
 /**
--- a/core/iwasm/libraries/wasi-nn/src/wasi_nn.c
+++ b/core/iwasm/libraries/wasi-nn/src/wasi_nn.c
@@ -697,6 +697,7 @@ static NativeSymbol native_symbols_wasi_nn[] = {
     REG_NATIVE_FUNC(get_output, "(ii*i*)i"),
 #else  /* WASM_ENABLE_WASI_EPHEMERAL_NN == 0 */
     REG_NATIVE_FUNC(load, "(*ii*)i"),
+    REG_NATIVE_FUNC(load_by_name, "(*i*)i"),
     REG_NATIVE_FUNC(init_execution_context, "(i*)i"),
     REG_NATIVE_FUNC(set_input, "(ii*)i"),
     REG_NATIVE_FUNC(compute, "(i)i"),
--- a/core/iwasm/libraries/wasi-nn/src/wasi_nn_tensorflowlite.cpp
+++ b/core/iwasm/libraries/wasi-nn/src/wasi_nn_tensorflowlite.cpp
@@ -85,12 +85,9 @@ is_valid_graph(TFLiteContext *tfl_ctx, graph g)
         NN_ERR_PRINTF("Invalid graph: %d >= %d.", g, MAX_GRAPHS_PER_INST);
         return runtime_error;
     }
-    if (tfl_ctx->models[g].model_pointer == NULL) {
+    if (tfl_ctx->models[g].model_pointer == NULL
+        && tfl_ctx->models[g].model == NULL) {
         NN_ERR_PRINTF("Context (model) non-initialized.");
-        return runtime_error;
-    }
-    if (tfl_ctx->models[g].model == NULL) {
-        NN_ERR_PRINTF("Context (tflite model) non-initialized.");
         return runtime_error;
     }
     return success;
--- a/core/iwasm/libraries/wasi-nn/test/utils.c
+++ b/core/iwasm/libraries/wasi-nn/test/utils.c
@@ -9,6 +9,7 @@
 
 #include <stdio.h>
 #include <stdlib.h>
+#define USE_WASM_LOAD_BY_NAME 1
 
 wasi_nn_error
 wasm_load(char *model_name, graph *g, execution_target target)
@@ -58,7 +59,7 @@ wasm_load(char *model_name, graph *g, execution_target target)
 wasi_nn_error
 wasm_load_by_name(const char *model_name, graph *g)
 {
-    wasi_nn_error res = load_by_name(model_name, g);
+    wasi_nn_error res = load_by_name(model_name, strlen(model_name), g);
     return res;
 }
 
@@ -108,7 +109,12 @@ run_inference(execution_target target, float *input, uint32_t *input_size,
               uint32_t num_output_tensors)
 {
     graph graph;
+
+#if USE_WASM_LOAD_BY_NAME == 0
     if (wasm_load(model_name, &graph, target) != success) {
+#else
+    if (wasm_load_by_name(model_name, &graph) != success) {
+#endif
         NN_ERR_PRINTF("Error when loading model.");
         exit(1);
     }


core/iwasm/libraries/wasi-nn/src/wasi_nn.c
```diff
@@ -697,6 +697,7 @@ static NativeSymbol native_symbols_wasi_nn[] = {
     REG_NATIVE_FUNC(get_output, "(ii*i*)i"),
 #else  /* WASM_ENABLE_WASI_EPHEMERAL_NN == 0 */
     REG_NATIVE_FUNC(load, "(*ii*)i"),
+    REG_NATIVE_FUNC(load_by_name, "(*i*)i"),
```
lum1n0us, line 700: Why not use `-DWASM_ENABLE_WASI_EPHEMERAL_NN=1`?

core/iwasm/libraries/wasi-nn/src/wasi_nn_tensorflowlite.cpp
```diff
@@ -85,14 +85,11 @@ is_valid_graph(TFLiteContext *tfl_ctx, graph g)
         NN_ERR_PRINTF("Invalid graph: %d >= %d.", g, MAX_GRAPHS_PER_INST);
         return runtime_error;
     }
-    if (tfl_ctx->models[g].model_pointer == NULL) {
+    if (tfl_ctx->models[g].model_pointer == NULL
+        && tfl_ctx->models[g].model == NULL) {
```
lum1n0us, line 89: The original version can output different information based on various invalid argument cases. Is there a specific reason we need to merge them?

core/iwasm/libraries/wasi-nn/include/wasi_nn.h
```diff
@@ -30,7 +30,7 @@ load(graph_builder_array *builder, graph_encoding encoding,
     __attribute__((import_module("wasi_nn")));
 
 wasi_nn_error
-load_by_name(const char *name, graph *g)
+load_by_name(char *name, uint32_t name_len, graph *g)
```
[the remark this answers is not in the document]
HongxiaWangSSSS: yes, its a bugfix,

core/iwasm/libraries/wasi-nn/src/wasi_nn_tensorflowlite.cpp
HongxiaWangSSSS: If call load_by_name(), there is no need to save the tflite buf to model_pointer.
https://github.com/HongxiaWangSSSS/wasm-micro-runtime/blob/c5414fd28baf973e3c95db1318de4d26f88007d3/core/iwasm/libraries/wasi-nn/src/wasi_nn_tensorflowlite.cpp#L141C49-L141C68

HongxiaWangSSSS: I'm not sure why not free model_point after below operation in load()
https://github.com/HongxiaWangSSSS/wasm-micro-runtime/blob/c5414fd28baf973e3c95db1318de4d26f88007d3/core/iwasm/libraries/wasi-nn/src/wasi_nn_tensorflowlite.cpp#L151

lum1n0us: It is required to validate `TFLitesContext.models[g]` for both cases, using `load()` and `load_by_name()`. It will not be acceptable if the change disables one of these cases.

core/iwasm/libraries/wasi-nn/src/wasi_nn.c
lum1n0us: If use `-DWAMR_BUILD_WASI_EPHEMERAL_NN=1` during compilation, you will be able to use the set of APIs, including `load_by_name()`. There is no need to change this line.

core/iwasm/libraries/wasi-nn/src/wasi_nn_tensorflowlite.cpp
HongxiaWangSSSS: Whether it is `load `or `load_by_name`, the check of `models[g].model_pointer`does not seem to be necessary, just make sure the `models[g].model` is not NULL maybe is enough.
Do you have any idea?

core/iwasm/libraries/wasi-nn/src/wasi_nn.c
HongxiaWangSSSS: If `-DWAMR_BUILD_WASI_EPHEMERAL_NN=1` must be added , I think there is no need to add this line.
Currently, it is possible to use the default `wasi-nn` instead of `wasi_ephemeral_nn`.

lum1n0us: Yes. Please do it.

core/iwasm/libraries/wasi-nn/test/utils.c
```diff
@@ -108,7 +109,12 @@ run_inference(execution_target target, float *input, uint32_t *input_size,
               uint32_t num_output_tensors)
 {
     graph graph;
+
+#if USE_WASM_LOAD_BY_NAME == 0
```
lum1n0us, line 113: Once we have two API set in *wasi_nn.h* separated by `WASM_ENABLE_WASI_EPHEMERAL_NN`:
- Wrap `wasm_load_by_name()` with `WASM_ENABLE_WASI_EPHEMERAL_NN`.
- There is no need to introduce `USE_WASM_LOAD_BY_NAME`.
- Use `WASM_ENABLE_WASI_EPHEMERAL_NN` to decide whether to use `wasm_load()` or `wasm_load_by_name()`.

core/iwasm/libraries/wasi-nn/include/wasi_nn.h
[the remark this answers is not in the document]
lum1n0us: Since we have two sets of APIs for historical reasons, we might remove one in another PR. For now, let's ensure both are functional.

- I suggest we use `WASM_ENABLE_WASI_EPHEMERAL_NN` on the wasm side.
- With this flag, we declare two sets of APIs in *wasi_nn.h*. Please align with the content of `native_symbols_wasi_nn` in *wasi_nn.c*

core/iwasm/libraries/wasi-nn/test/utils.c
HongxiaWangSSSS: updated!

core/iwasm/libraries/wasi-nn/src/wasi_nn_tensorflowlite.cpp
HongxiaWangSSSS: after this [operation](https://github.com/HongxiaWangSSSS/wasm-micro-runtime/blob/c5414fd28baf973e3c95db1318de4d26f88007d3/core/iwasm/libraries/wasi-nn/src/wasi_nn_tensorflowlite.cpp#L151),  the `models[g].model_pointer`'s connect has been saved in `models[g].model`.

core/iwasm/libraries/wasi-nn/src/wasi_nn.c
ayakoakasaka: Test code at
https://github.com/bytecodealliance/wasm-micro-runtime/blob/c018b8ab987e28e1ed92c1dc7280a3a777bd8a6d/core/iwasm/libraries/wasi-nn/test/utils.c#L61

has a different signature with `wasi_ephemeral_nn` load_by_name at https://github.com/bytecodealliance/wasm-micro-runtime/blob/c018b8ab987e28e1ed92c1dc7280a3a777bd8a6d/core/iwasm/libraries/wasi-nn/src/wasi_nn.c#L692



So I doubt switching to `wasi_ephemeral_nn`

core/iwasm/libraries/wasi-nn/include/wasi_nn.h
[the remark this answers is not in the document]
lum1n0us: > Where can I find the function prototypes for wasi_ephemeral_nn?

By passing a compilation flag during the configuration phase, like `cmake -DWAMR_BUILD_WASI_EPHEMERAL_NN=1`

> For functions like get_output, the signatures are different,

I think wasi-nn has switched from witx to wit to describe their APIs.

In fact, the most popular API in the wasi-nn field is the Rust API. We aim to maintain the most compatibility with two widely used APIs: [bytecodealliance wasi-nn](https://github.com/bytecodealliance/wasi-nn) and [wasmedge wasi-nn](https://github.com/second-state/wasmedge-wasi-nn). You will find both interfaces with host below:

- https://github.com/bytecodealliance/wasi-nn/blob/556890b121dd1171665d835aba4d04a7e29e37dc/rust/src/generated.rs#L303-L313
- https://github.com/second-state/wasmedge-wasi-nn/blob/ggml/rust/src/generated.rs#L383-L404

---

*FYI, Some interfaces of wasmedge-wasi-nn are not supported simply because they are not used in [WasmEdge-WASINN-examples ggml](https://github.com/second-state/WasmEdge-WASINN-examples/tree/master/wasmedge-ggml)*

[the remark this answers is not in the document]
lum1n0us: @HongxiaWangSSSS I suggest we use `WASM_ENABLE_WASI_EPHEMERAL_NN` on the wasm side and there will be two sets for *wasi_ephemeral_nn* and *wasi_nn*. *wasi_nn* will be deprecated in the future.

```c
wasi_nn_error
load(graph_builder_array *builder, graph_encoding encoding,
     execution_target target, graph *g)
    __attribute__((import_module("wasi_ephemeral_nn")));

wasi_nn_error
load_by_name(const char *name, uint32_t len, graph *g)
    __attribute__((import_module("wasi_ephemeral_nn")));

wasi_nn_error
load_by_name_with_config(const char *name, uint32_t name_len, void *config, uint32_t config_len, graph *g)     __attribute__((import_module("wasi_ephemeral_nn")));;

wasi_nn_error
init_execution_context(graph g, graph_execution_context *exec_ctx)     __attribute__((import_module("wasi_ephemeral_nn")));;

wasi_nn_error
set_input(graph_execution_context ctx, uint32_t index, tensor *tensor)     __attribute__((import_module("wasi_ephemeral_nn")));;

wasi_nn_error
compute(graph_execution_context ctx)     __attribute__((import_module("wasi_ephemeral_nn")));;

wasi_nn_error
get_output(graph_execution_context ctx, uint32_t index,
           tensor_data output_tensor, uint32_t *output_tensor_size)     __attribute__((import_module("wasi_ephemeral_nn")));;

#else

wasi_nn_error
load(graph_builder_array *builder, graph_encoding encoding,
     execution_target target, graph *g)
    __attribute__((import_module("wasi_nn")));

wasi_nn_error
init_execution_context(graph g, graph_execution_context *ctx)
    __attribute__((import_module("wasi_nn")));

wasi_nn_error
set_input(graph_execution_context ctx, uint32_t index, tensor *tensor)
    __attribute__((import_module("wasi_nn")));

wasi_nn_error
compute(graph_execution_context ctx) __attribute__((import_module("wasi_nn")));

wasi_nn_error
get_output(graph_execution_context ctx, uint32_t index,
           tensor_data output_tensor, uint32_t *output_tensor_size)
    __attribute__((import_module("wasi_nn")));
#endif
```

core/iwasm/libraries/wasi-nn/test/utils.c
```diff
@@ -58,7 +58,7 @@ wasm_load(char *model_name, graph *g, execution_target target)
 wasi_nn_error
 wasm_load_by_name(const char *model_name, graph *g)
 {
-    wasi_nn_error res = load_by_name(model_name, g);
+    wasi_nn_error res = load_by_name(model_name, strlen(model_name), g);
```
lum1n0us, line 61: better be
```c
wasi_nn_error
wasm_load_by_name(const char *model_name, graph *g)
{
    wasi_nn_error res = load_by_name(model_name, strlen(model_name), g);
    return res;
}
#endif 

```

core/iwasm/libraries/wasi-nn/include/wasi_nn.h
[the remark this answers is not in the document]
ayakoakasaka: Is there a mismatch in the function signature?
https://github.com/bytecodealliance/wasm-micro-runtime/blob/c018b8ab987e28e1ed92c1dc7280a3a777bd8a6d/core/iwasm/libraries/wasi-nn/src/wasi_nn.c#L697

and


> wasi_nn_error
> get_output(graph_execution_context ctx, uint32_t index,
>            tensor_data output_tensor, uint32_t *output_tensor_size)     __attribute__((import_module("wasi_ephemeral_nn")));

core/iwasm/libraries/wasi-nn/src/wasi_nn.c
HongxiaWangSSSS: > Test code at
> 
> https://github.com/bytecodealliance/wasm-micro-runtime/blob/c018b8ab987e28e1ed92c1dc7280a3a777bd8a6d/core/iwasm/libraries/wasi-nn/test/utils.c#L61
> 
> has a different signature with `wasi_ephemeral_nn` load_by_name at
> 
> https://github.com/bytecodealliance/wasm-micro-runtime/blob/c018b8ab987e28e1ed92c1dc7280a3a777bd8a6d/core/iwasm/libraries/wasi-nn/src/wasi_nn.c#L692
> 
I think this is a bug and it has never been called befor, and I changed to https://github.com/bytecodealliance/wasm-micro-runtime/blob/ff4ed2ff66ba015019fc53f2e08b5426ee2592fa/core/iwasm/libraries/wasi-nn/test/utils.c#L61

core/iwasm/libraries/wasi-nn/include/wasi_nn.h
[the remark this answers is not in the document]
ayakoakasaka: If so, we need to define it at wasi_nn.h 

```
wasi_nn_error
 get_output(graph_execution_context ctx, uint32_t index,
           tensor_data output_tensor, uint32_t output_tensor_len, uint32_t *output_tensor_size)     __attribute__((import_module("wasi_ephemeral_nn")));
```
But the  backend definition doesn't look like it matches (signature might be ok)

https://github.com/bytecodealliance/wasm-micro-runtime/blob/c018b8ab987e28e1ed92c1dc7280a3a777bd8a6d/core/iwasm/libraries/wasi-nn/src/wasi_nn_tensorflowlite.cpp#L366-L368

core/iwasm/libraries/wasi-nn/src/wasi_nn.c
ayakoakasaka: So, does that mean that the ephemeral version is meant to be compatible with Rust (especially WasmEdge), whereas the non-ephemeral one doesn't need to be?
If that's the case, wouldn't it make sense to add `load_by_name` to the wasi_nn.h header, which is a straightforward C interpretation of the witx specification?

core/iwasm/libraries/wasi-nn/include/wasi_nn.h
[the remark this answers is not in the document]
lum1n0us: Yes. I am suggestion this https://github.com/bytecodealliance/wasm-micro-runtime/pull/4267#discussion_r2108026395 as the new content of wasi_nn.h

core/iwasm/libraries/wasi-nn/src/wasi_nn.c
lum1n0us: Yes. `load_by_name()` needs a new signature.

ayakoakasaka: https://github.com/bytecodealliance/wasm-micro-runtime/pull/4267#discussion_r2084586337
said this is for C langage, so why not keep it for `non-wasi_ephemeral_nn` to provide the same performance ?

PR #4268, opened 2025-05-12
Bypass wamr_ide-related components from the release process.

lum1n0us:
Mostly because of some observations:
- There is no actual usage reported.
- Both ide-ext and ide-docker-image have not been maintained for quite a while.
- At the very least, there is no need to recompile it every time when there are no modifications.

### DIFF
--- a/.github/workflows/release_process.yml
+++ b/.github/workflows/release_process.yml
@@ -113,7 +113,7 @@ jobs:
       runner: macos-13
       upload_url: ${{ needs.create_release.outputs.upload_url }}
       ver_num: ${{ needs.create_tag.outputs.new_ver }}
-  
+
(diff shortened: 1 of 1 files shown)


TianlongLiang:
I agree with the PR. I think the existing version of the VSCode extension is good enough for most usage. If there is a concrete request for a newer wamr feature in VSCode extension. We can reenable it and update it by then.

> At the very least, there is no need to recompile it every time when there are no modifications.

The VSCode extension itself won't change most of the time. The reason we recompiled it previously is that the iwasm/wamrc inside Docker needs to change.

TianlongLiang (approved):

xujuntwt95329 (approved):
LGTM

PR #4269, opened 2025-05-12
product-mini/platforms/windows: set C++17 explicitly

yamt:
The recent LLVM uses std::optional, which is C++17.


lum1n0us (approved):
LGTM

TianlongLiang (approved):
LGTM

PR #4271, opened 2025-05-12
Generic exceptions handling callback included

borjatelladotecnalia:
In order to be able to catch  "trap" errors from native side and reload the WAMR application or trigger corrective actions such as "reload factory settings" for us this type of notifications are very usefull. We took as reference the "enlarge memory exception" mechanism and try to keep it simple.

### DIFF
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -91,6 +91,9 @@ wasm_runtime_destroy_registered_module_list(void);
 
 #define E_TYPE_XIP 4
 
+static generic_error_callback_t generic_error_cb;
+static void *generic_error_user_data;
+
 static uint8
 val_type_to_val_kind(uint8 value_type);
 
@@ -3016,6 +3019,8 @@ wasm_set_exception_local(WASMModuleInstance *module_inst, const char *exception)
         module_inst->cur_exception[0] = '\0';
     }
     exception_unlock(module_inst);
+    if (generic_error_cb)
+      generic_error_cb((WASMModuleInstanceCommon *)module_inst, ((AOTModuleInstance *)module_inst)->cur_exec_env, "");
 }
 
 void
@@ -7869,4 +7874,12 @@ wasm_runtime_is_underlying_binary_freeable(WASMModuleCommon *const module)
 #endif /* WASM_ENABLE_AOT != 0 */
 
     return true;
+}
+
+void
+wasm_runtime_set_generic_error_callback(
+  const generic_error_callback_t callback, void *user_data)
+{
+    generic_error_cb = callback;
+    generic_error_user_data = user_data;
 }
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -3020,7 +3020,8 @@ wasm_set_exception_local(WASMModuleInstance *module_inst, const char *exception)
     }
     exception_unlock(module_inst);
     if (generic_error_cb)
-      generic_error_cb((WASMModuleInstanceCommon *)module_inst, ((AOTModuleInstance *)module_inst)->cur_exec_env, "");
+        generic_error_cb((WASMModuleInstanceCommon *)module_inst,
+                         ((AOTModuleInstance *)module_inst)->cur_exec_env, "");
 }
 
 void
@@ -7877,8 +7878,8 @@ wasm_runtime_is_underlying_binary_freeable(WASMModuleCommon *const module)
 }
 
 void
-wasm_runtime_set_generic_error_callback(
-  const generic_error_callback_t callback, void *user_data)
+wasm_runtime_set_generic_error_callback(const generic_error_callback_t callback,
+                                        void *user_data)
 {
     generic_error_cb = callback;
     generic_error_user_data = user_data;
--- a/core/iwasm/include/wasm_export.h
+++ b/core/iwasm/include/wasm_export.h
@@ -2102,12 +2102,19 @@ typedef void (*enlarge_memory_error_callback_t)(
     uint32_t memory_index, enlarge_memory_error_reason_t failure_reason,
     wasm_module_inst_t instance, wasm_exec_env_t exec_env, void *user_data);
 
+
+typedef void (*generic_error_callback_t)(
+    wasm_module_inst_t instance, wasm_exec_env_t exec_env, void *user_data);
 /**
  * Setup callback invoked when memory.grow fails
  */
 WASM_RUNTIME_API_EXTERN void
 wasm_runtime_set_enlarge_mem_error_callback(
     const enlarge_memory_error_callback_t callback, void *user_data);
+
+WASM_RUNTIME_API_EXTERN void
+wasm_runtime_set_generic_error_callback(
+    const generic_error_callback_t callback, void *user_data);
 
 /*
  * module instance context APIs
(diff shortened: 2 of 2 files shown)


yamt:
> In order to be able to catch "trap" errors from native side and reload the WAMR application or trigger corrective actions such as "reload factory settings" for us this type of notifications are very usefull. We took as reference the "enlarge memory exception" mechanism and try to keep it simple.

i don't see what's the point of the api.
an embedder can use wasm_runtime_get_exception to handle traps.

borjatelladotecnalia:
> > In order to be able to catch "trap" errors from native side and reload the WAMR application or trigger corrective actions such as "reload factory settings" for us this type of notifications are very usefull. We took as reference the "enlarge memory exception" mechanism and try to keep it simple.
> 
> i don't see what's the point of the api. an embedder can use wasm_runtime_get_exception to handle traps.

Our use case is more or less the following: A native embedded application written in C, running on ThreadX RTOS, embeds a WASM runtime to execute a WebAssembly module in the background. The WASM module exports functions used by the native application to enqueue/push data. The WASM module periodically computes results in its dedicated thread. If invalid data or processing errors occur, a native callback function is invoked directly from the WASM runtime environment to notify the host system

**Functional Flow:**

1. **Initialization:**
   * Native code starts `wasm_thread`, sets up the WASM runtime, and loads the module.
   * A C function pointer (the callback) is registered as a host function import, exposed to the WASM side.

2. **Data Input:**
    * Native app calls exported WASM functions to pass data (e.g., via `enqueue_value(data)`).

3. **Computation Cycle:**
   * The WASM module sleeps or waits for a time slice (every *n* seconds), then processes the buffered input.

4. **Error Detection and Callback Trigger:**
   * If the input state is invalid or a computation fails:
     * The WASM code calls an imported host function (the registered C callback).
     * The native-side callback is invoked from within the WASM thread context.
     * This function may log the error, update system state, or trigger recovery actions.

What we need is to stop the WASM module when an error occurs and notify the native side. The native application is unaware of the internal behavior of the WASM module, which runs autonomously in the background. As I understand it (though I may be wrong), using wasm_runtime_get_exception would require the native side to poll the WASM module, which doesn't fit our use case.

yamt:
> > > In order to be able to catch "trap" errors from native side and reload the WAMR application or trigger corrective actions such as "reload factory settings" for us this type of notifications are very usefull. We took as reference the "enlarge memory exception" mechanism and try to keep it simple.
> > 
> > 
> > i don't see what's the point of the api. an embedder can use wasm_runtime_get_exception to handle traps.
> 
> Our use case is more or less the following: A native embedded application written in C, running on ThreadX RTOS, embeds a WAMR runtime to execute a WebAssembly module in the background. The WASM module exports functions used by the native application to enqueue/push data. The WASM module periodically computes results in its dedicated thread. If invalid data or processing errors occur, a native callback function is invoked directly from the WASM runtime environment to notify the host system
> 
> **Functional Flow:**
> 
>     1. **Initialization:**
>        
>        * Native code starts `wasm_thread`, sets up the WASM runtime, and loads the module.
>        * A C function pointer (the callback) is registered as a host function import, exposed to the WASM side.
> 
>     2. **Data Input:**
>        
>        * Native app calls exported WASM functions to pass data (e.g., via `enqueue_value(data)`).
> 
>     3. **Computation Cycle:**
>        
>        * The WASM module sleeps or waits for a time slice (every _n_ seconds), then processes the buffered input.
> 
>     4. **Error Detection and Callback Trigger:**
>        
>        * If the input state is invalid or a computation fails:
>          
>          * The WASM code calls an imported host function (the registered C callback).
>          * The native-side callback is invoked from within the WASM thread context.
>          * This function may log the error, update system state, or trigger recovery actions.
> 
> 
> What we need is to stop the WASM module when an error occurs and notify the native side. The native application is unaware of the internal behavior of the WASM module, which runs autonomously in the background. As I understand it (though I may be wrong), using wasm_runtime_get_exception would require the native side to poll the WASM module, which doesn't fit our use case.

wasm_runtime_get_exception is expected to be used after an api call to execute wasm bytecode returns.
eg.
```shell
wasm_application_execute_main(... );
if (wasm_runtime_get_exception(... )) {
  // got an exception, call your handler.
}
```

note that thread-mgr propagates the exception to every threads in the group. (cf. wasm_cluster_set_exception)

borjatelladotecnalia:
> > > > In order to be able to catch "trap" errors from native side and reload the WAMR application or trigger corrective actions such as "reload factory settings" for us this type of notifications are very usefull. We took as reference the "enlarge memory exception" mechanism and try to keep it simple.
> > > 
> > > 
> > > i don't see what's the point of the api. an embedder can use wasm_runtime_get_exception to handle traps.
> > 
> > 
> > Our use case is more or less the following: A native embedded application written in C, running on ThreadX RTOS, embeds a WAMR runtime to execute a WebAssembly module in the background. The WASM module exports functions used by the native application to enqueue/push data. The WASM module periodically computes results in its dedicated thread. If invalid data or processing errors occur, a native callback function is invoked directly from the WASM runtime environment to notify the host system
> > **Functional Flow:**
> > ```
> > 1. **Initialization:**
> >    
> >    * Native code starts `wasm_thread`, sets up the WASM runtime, and loads the module.
> >    * A C function pointer (the callback) is registered as a host function import, exposed to the WASM side.
> > 
> > 2. **Data Input:**
> >    
> >    * Native app calls exported WASM functions to pass data (e.g., via `enqueue_value(data)`).
> > 
> > 3. **Computation Cycle:**
> >    
> >    * The WASM module sleeps or waits for a time slice (every _n_ seconds), then processes the buffered input.
> > 
> > 4. **Error Detection and Callback Trigger:**
> >    
> >    * If the input state is invalid or a computation fails:
> >      
> >      * The WASM code calls an imported host function (the registered C callback).
> >      * The native-side callback is invoked from within the WASM thread context.
> >      * This function may log the error, update system state, or trigger recovery actions.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > What we need is to stop the WASM module when an error occurs and notify the native side. The native application is unaware of the internal behavior of the WASM module, which runs autonomously in the background. As I understand it (though I may be wrong), using wasm_runtime_get_exception would require the native side to poll the WASM module, which doesn't fit our use case.
> 
> wasm_runtime_get_exception is expected to be used after an api call to execute wasm bytecode returns. eg.
> 
> ```shell
> wasm_application_execute_main(... );
> if (wasm_runtime_get_exception(... )) {
>   // got an exception, call your handler.
> }
> ```
> 
> note that thread-mgr propagates the exception to every threads in the group. (cf. wasm_cluster_set_exception)

Well, I think that I found the origin of my missunderstanding, at the begining of our project I used to test with the structure described above and tried some  "trapy" code.
```shell

 6  float div_to_zero (int value, int div_value)
  7 {
  8     float new_value = value / div_value;
  9     printf ("Divsion done %d",div_value);
 10     return new_value;
 11 }

 57 int main(int argc, char **argv)
 58 {
 59     for (int i = 0 ; i < 15;i++)
 60         div_to_zero(10, (10-i));
 61     printf ("\nAll divisions done\n");
 65     return 0;
 66 }
```
It seems that if the code is compiled with the  **-O3** clang flag, the compiler is wise enough to strip that line when i=10 and the exception calculating the new_value (line 8) is not triggered and the execution is not stopped,  on the other hand if compiled with **-O0** the exception is triggered and the execution is stopped.

lum1n0us:
Great. It seems there is no need to discuss this PR any further. Let's close it.

PR #4274, opened 2025-05-13
Set CMAKE_OSX_SYSROOT when building lldb

lum1n0us:
CMake 4 no longer sets the CMAKE_OSX_SYSROOT variable by default, causing the lldb build to fail after all GitHub-hosted runners have been upgraded to CMake 4.

As a workaround, the variable is set using CMake command line options. There is a PR to fix this issue in the llvm-project:
https://github.com/llvm/llvm-project/pull/138020. We might want to remove this workaround after that PR has been merged.

---

With this PR, I am able to successfully build lldb with the patch. https://github.com/lum1n0us/wasm-micro-runtime/actions/runs/14989528618/job/42109722679


TianlongLiang (approved):
LGTM

xujuntwt95329 (approved):
LGTM

PR #4275, opened 2025-05-13
add .cmake-format.yml

Septa2112:


lum1n0us:
Let's close it and reopen it once we find a smooth way to integrate it into the repository without causing significant modifications.

PR #4278, opened 2025-05-15
 Fix SIMD load lane to avoid incompatible pointer types

lum1n0us:


core/iwasm/interpreter/wasm_interp_fast.c
```diff
@@ -6514,17 +6515,17 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
...
-        else {                                             \
-            vec.register[lane] = *(uint##width *)(maddr);  \
-        }                                                  \
-        PUT_V128_TO_ADDR(frame_lp + addr_ret, vec);        \
+#define SIMD_LOAD_LANE_COMMON(vec, register, lane, width)            \
+    do {                                                             \
+        addr_ret = GET_OFFSET();                                     \
+        CHECK_MEMORY_OVERFLOW(width / 8);                            \
+        if (width == 64) {                                           \
+            vec.register[lane] = GET_I64_FROM_ADDR((uint32 *)maddr); \
```
loganek, line 6523: could you explain the reason for this change? The macro is defined as:
```
#define GET_I64_FROM_ADDR(addr) (*(int64 *)(addr))
```
and `maddr` is already a `uint8*` and in the macro it's casted to the `int64` pointer.

lum1n0us:
```c
#if WASM_CPU_SUPPORTS_UNALIGNED_ADDR_ACCESS != 0
  #define GET_I64_FROM_ADDR(addr) (*(int64 *)(addr))
#else
  static inline int64
  GET_I64_FROM_ADDR(uint32 *addr)    // <- goes to "incompatible pointer type"
  {
      union {
          int64 val;
          uint32 parts[2];
      } u;
      u.parts[0] = addr[0];
      u.parts[1] = addr[1];
      return u.val;
  }
#endif
```

loganek: oh, didn't notice the other branch. Thanks!

loganek (approved):
LGTM

TianlongLiang (approved):
LGTM

Issue #4129, opened 2025-03-05
[RFC] Predictable and (almost) fully automated WAMR releases


loganek:
#### Feature

I'd like to discuss with the community standardizing release cadence of WAMR releases. 

#### Benefit

The customers can know in advance any WAMR releases and plan accordingly. Customers no longer have to request ad-hoc requests.

#### Implementation

As a first step we need to make sure all of the tests are automated and publicly available, so Intel engineers no longer have to be overloaded with release requests; any maintainer can coordinate the release. Based on the discussion we had in the community meetings, there are two major actions to achieve this:
1. decouple WAMR IDE from WAMR runtime, as IDE right now requires a bunch of manual tests (and the effort to automate them would likely be significant)
2. enable tests that require QEMU 

Whereas it's unknown yet whether 2. can be implemented completely using github actions, I think the problem overall is solvable one way or another (e.g. by using instances from 3rd party cloud providers etc.) so whoever picks this up can figure out the implementation detail.
The 1. can look a bit concerning, so I'd like to clarify that I'm not suggesting for it to completely decouple IDE from the WAMR project but instead keep it around with the same versioning schema, but make releases on best-effort basis. 

The regular releases will include any changes that happened since the previous release, and if the particular feature will not be ready, it'll have to wait for another release (hence I'm proposing relatively frequent releases).

When it comes to the versioning, I propose incrementing MINOR version by default and change it when:
* there are mainly bugfixes and minor improvements, only the PATCH number should be increased
* for any breaking changes, MAJOR version should be incremented

I guess the preference would be not to change the MAJOR version too often, so for that we should likely maintain a dev branch where eventually all the breaking changes are included, and then added to the main branch through a single pull request to make sure they all are included in a single release.

This is just a proposal and I do understand it might not touch on all the potential problems and details. I'm open for feedback and suggestions; feel free to leave your comment below.

no1wudi:
> 1. decouple WAMR IDE from WAMR runtime, as IDE right now requires a bunch of manual tests (and the effort to automate them would likely be significant)

Agreed. As a tool used during the development phase, the cost of rolling out IDE updates to end-users is significantly lower compared to the core runtime, allowing for a faster and more flexible release cadence.

> enable tests that require QEMU

If possible, we should incorporate as many automated tests as we can. Ideally, if quality can be guaranteed, we should be able to perform a release operation at any commit point. 
[...]

PR #4279, opened 2025-05-15
Fixed unit tests on X86_32

Septa2112:
- Fixed bugs where some unit cases could not run when `WAMR_BUILD_TARGET=X86_32`.
- Skipped some unit cases that are not supported when `WAMR_BUILD_TARGET=X86_32`.


loganek:
Would it make sense to add those to the CI to avoid future regressions?

lum1n0us:
Sure. We should add it to the CI. By the way, I am planning [a CI improvement strategy](https://github.com/bytecodealliance/wasm-micro-runtime/issues/4129#issuecomment-2871236940) here.

tests/unit/CMakeLists.txt
```diff
@@ -42,19 +43,27 @@ enable_testing()
...
 add_subdirectory(linear-memory-aot)
 add_subdirectory(aot-stack-frame)
 add_subdirectory(linux-perf)
 add_subdirectory(gc)
-add_subdirectory(memory64)
 add_subdirectory(tid-allocator)
-add_subdirectory(shared-heap)
\ No newline at end of file
+
+if (NOT WAMR_BUILD_TARGET STREQUAL "X86_32")
```
loganek, line 56: not sure if excluding all of those tests by default is the right approach here. For example, I guess `running-modes` in this case are excluded because they require fast-jit (which isn't supported on 32bit target). However, fast-jit is optional in this test, and can be disabled for 32bit mode; it'd require changes in ./tests/unit/running-modes/CMakeLists.txt, something like:
```
-set(WAMR_BUILD_FAST_JIT 1)
+if(NOT WAMR_BUILD_TARGET STREQUAL "X86_32")
+  set(WAMR_BUILD_FAST_JIT 1)
+endif()

```

Haven't checked the other tests, but I presume similar strategy can be applied at least to some of them.

TianlongLiang: I took a look at the disabled test suites:
```cmake
  # should enable 32-bit llvm when X86_32
  add_subdirectory (aot)
  add_subdirectory (custom-section)
  add_subdirectory (compilation)

  # Fast-JIT or mem64 is not supported on X86_32
  add_subdirectory (running-modes)
  add_subdirectory (memory64)
  add_subdirectory (shared-heap)

  # HW_BOUND_CHECK is not supported on X86_32
  add_subdirectory (runtime-common)
```
I think we can indeed ignore most of them at this point. We normally wouldn't compile aot file on the 32-bit platform, and enabling those test suites on the 32-bit platform requires the 32-bit LLVM libraries, which will complicate the CIs and add more artifacts. So we can ignore `aot`, `custom-section`, `compilation`, `running-modes`(if we ignore fast-jit, we will test switch between llvm-jit and interp). The `memory64` and `runtime-common` weren't supported on the 32-bit platform.  So that's only left for `shared-heap`, I have already enabled it in another PR and will merge soon.

Septa2112:
I have added these tests to the CI, and they have passed successfully.

lum1n0us (approved):
LGTM

TianlongLiang (approved):
LGTM

PR #4286, opened 2025-05-20
Fix Compiler Error C2491

lum1n0us:
> Data, static data members, and functions can be declared as `dllimports`
  but not defined as `dllimports`.

https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2491?view=msvc-170


TianlongLiang (approved):
LGTM

yamt (approved):

Issue #4280, opened 2025-05-16
When using msvc to compile the sample in samples\basic, there is an error: "definition of dllimport function not allowed"


UradaSources:
When using msvc to compile the sample in samples\basic, there is an error: "definition of dllimport function not allowed"

i am trying to compile the samples\basic sample program in Windows using the Visual Studio CMake folder mode, but the compilation always fails, and the compiler reports C2491 errors, which seem to be caused by the macro:
`#define WASM_RUNTIME_API_EXTERN __declspec(dllimport)`

the location is 2 functions under the core\iwasm\common\wasm_runtime_common.c file
```
WASM_RUNTIME_API_EXTERN bool
wasm_runtime_resolve_symbols(WASMModuleCommon *module)

WASM_RUNTIME_API_EXTERN bool
wasm_runtime_is_underlying_binary_freeable(WASMModuleCommon *const module)
```

There are also a lot of C4273 dll link inconsistency warnings. Did I do something wrong?

Also, does the project officially support clang under Windows? I tried to compile the project with mingw-clang, and there were always a lot of strange errors

```
[proc] 执行命令: "C:\Program Files\CMake\bin\cmake.EXE" --build d:/lch/Private/wasm_test/wasm-micro-runtime/samples/basic/build --config Debug --target all --
[build] [10/88   1% :: 1.886] Building C object CMakeFiles/vmlib.dir/D_/lch/Private/wasm_test/wasm-micro-runtime/core/shared/platform/windows/win_socket.c.obj
[build] In file included from D:/lch/Private/wasm_test/wasm-micro-runtime/core/shared/platform/windows/win_socket.c:6:
[build] In file included from D:/lch/Private/wasm_test/wasm-micro-runtime/core/shared/platform/windows/../include/platform_api_vmcore.h:9:
[build] In file included from D:/lch/Private/wasm_test/wasm-micro-runtime/core/shared/platform/windows/../include/platform_common.h:13:
[build] D:/lch/Private/wasm_test/wasm-micro-runtime/core/shared/platform/windows/platform_internal.h:14:9: warning: unknown pragma ignored [-Wunknown-pragmas]
[build]    14 | #pragma warning(disable : 5105)
[build]       |         ^
[build] In file included from D:/lch/Private/wasm_test/wasm-micro-runtime/core/shared/platform/windows/win_socket.c:9:
[...]

PR #4287, opened 2025-05-22
Enable runtime API exposure for MSVC builds

lum1n0us:


samples/basic/src/free_buffer_early.c
```diff
@@ -5,6 +5,7 @@
 
 #include "wasm_export.h"
 #include "bh_read_file.h"
+#include "bh_platform.h"
```
yamt, line 8: is this related to the rest of changes in this PR?

lum1n0us: If trying to compile samples/basic on Windows(mentioned in #4280) after patching `COMPILING_WASM_RUNTIME_API`, this modification can fix an error caused by `optarg` and `getopt()`.

yamt: why bh_platfrom.h, not bh_getopt.h as nearby files do?

yamt (approved):
lgtm

PR #4294, opened 2025-05-25
Enhance type checking for function types in loader

lum1n0us:
Especially when GC is enabled, a valid item of `module->types` needs additional checks before casting to `WASMFuncType`.

Also, avoid overflowing if reftype_map_count is 0.

Additionally, correctly set IN_OSS_FUZZ based on CFLAGS_ENV for sanitizer configuration. Update ASan and UBSan messages for clarity in non-oss-fuzz environments.

Since GC hasn't been enabled in it, the mini loader doesn't share the same problem.


wenyongh (approved):
LGTM

PR #4300, opened 2025-05-28
add heaptype check for GC when `ref.null`

Septa2112:
Update `reftype` validation in `load_init_expr` when `ref.null`.

### DIFF
--- a/core/iwasm/interpreter/wasm_loader.c
+++ b/core/iwasm/interpreter/wasm_loader.c
@@ -808,19 +808,24 @@ load_init_expr(WASMModule *module, const uint8 **p_buf, const uint8 *buf_end,
             {
                 uint8 type1;
 
+#if WASM_ENABLE_GC == 0
                 CHECK_BUF(p, p_end, 1);
                 type1 = read_uint8(p);
 
-#if WASM_ENABLE_GC == 0
                 cur_value.ref_index = NULL_REF;
                 if (!push_const_expr_stack(&const_expr_ctx, flag, type1,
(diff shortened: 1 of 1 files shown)


Septa2112:
The previous logic was indeed not compliant with the GC spec, which is why I submitted the latest patch. 

---
## Explanation of the Latest Changes
- According to [Link 1](https://webassembly.github.io/gc/core/valid/instructions.html#xref-syntax-instructions-syntax-instr-ref-mathsf-ref-null-mathit-ht), we must ensure that the heap type is valid when `ref.null`. 
- According to [Link 2](https://webassembly.github.io/gc/core/valid/types.html#heap-types), a heap type is considered valid if it is either a **concrete heap type** or an **abstract heap type**. 


However, in this function, the check for abstract heap types (`absheaptype`) was clearly missing, so this condition needs to be added explicitly in the `if` statement.

### The added logic is

```C
// Step 1: Read heap type
read_leb_int32(p, p_end, heap_type);

// Step 2: Check if it's not an absheaptype
if (!wasm_is_valid_heap_type(heap_type)){
    //Step 3: check type_idx  (concrete heap type)
}
else {
    ...
}

```

### technical details

* **When GC is disabled**, no change is needed.
* **When GC is enabled**, heap types in WAMR are LEB-encoded values ([Link 3](https://webassembly.github.io/gc/core/appendix/index-types.html)). Therefore, we must use `read_leb_int32` to parse the heap type correctly. And we can compute the original `type1` using `type1 = (uint8)((int32)0x80 + heap_type);`.

PR #4302, opened 2025-05-29
loader: handle nullable heap reference types in import section

kylo5aby:
- Add support for reading nullable heap reference types in the import section 
of WASM modules.
- Check for nullable heap reference types in table imports


lum1n0us (approved):
LGTM. Although, I might think `skip_leb_int32()` is slightly better.

PR #4307, opened 2025-06-02
fix wasi-nn abi definitions

yamt:
sync with a more appropriate version of the definitions.

as we use the "wasi_ephemeral_nn", which is p1-based, it seems
more appropriate to use definitions from witx, not wit.

it's a bit unfortunate p2-based wasi-nn made gratuitous changes
like this from p1.

note: this is an ABI change.


yamt:
> Need to sync up the content at https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/libraries/wasi-nn/README.md#wasm.

which part of it? this PR contains an update of README.md.

lum1n0us:
Oops, missed that part. It's all good, no need to change anything.

PR #4309, opened 2025-06-02
wasi-nn: remove "backends" argument from detect_and_load_backend()

yamt:
it seems meaningless and quite confusing to access a table with two aliases ("lookup" and "backends") within a function.

no functional changes are intended.

### DIFF
--- a/core/iwasm/libraries/wasi-nn/src/wasi_nn.c
+++ b/core/iwasm/libraries/wasi-nn/src/wasi_nn.c
@@ -342,7 +342,6 @@ graph_encoding_to_backend_lib_name(graph_encoding encoding)
 
 static bool
 detect_and_load_backend(graph_encoding backend_hint,
-                        struct backends_api_functions *backends,
                         graph_encoding *loaded_backend)
 {
     if (backend_hint > autodetect)
@@ -365,7 +364,7 @@ detect_and_load_backend(graph_encoding backend_hint,
     if (!backend_lib_name)
         return false;
 
-    return prepare_backend(backend_lib_name, backends + backend_hint);
+    return prepare_backend(backend_lib_name, lookup + backend_hint);
 }
 
 /* WASI-NN implementation */
@@ -410,7 +409,7 @@ wasi_nn_load(wasm_exec_env_t exec_env, graph_builder_array_wasm *builder,
     }
 
     graph_encoding loaded_backend = autodetect;
-    if (!detect_and_load_backend(encoding, lookup, &loaded_backend)) {
+    if (!detect_and_load_backend(encoding, &loaded_backend)) {
         res = invalid_encoding;
         NN_ERR_PRINTF("load backend failed");
         goto fail;
@@ -468,7 +467,7 @@ wasi_nn_load_by_name(wasm_exec_env_t exec_env, char *name, uint32_t name_len,
     NN_DBG_PRINTF("[WASI NN] LOAD_BY_NAME %s...", name);
 
     graph_encoding loaded_backend = autodetect;
-    if (!detect_and_load_backend(autodetect, lookup, &loaded_backend)) {
+    if (!detect_and_load_backend(autodetect, &loaded_backend)) {
         NN_ERR_PRINTF("load backend failed");
         return invalid_encoding;
     }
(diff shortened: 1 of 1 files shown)


lum1n0us:
The initial idea is to minimize the usage of global variables by having fewer functions aware of the existence of the `lookup` variable, if it ultimately can't be avoided. So, I hope we can try another approach to remove the global variable. Maybe, just maybe, we can move that table into `WASINNContext`?

yamt:
> The initial idea is to minimize the usage of global variables by having fewer functions aware of the existence of the `lookup` variable, if it ultimately can't be avoided. So, I hope we can try another approach to remove the global variable. Maybe, just maybe, we can move that table into `WASINNContext`?

do you imply to allow loading (dlopen/dlysm) a backend multiple times within a process?
i don't think it works well without significant rework. after all, the symbol table is usually global.

lum1n0us:
> do you imply to allow loading (dlopen/dlysm) a backend multiple times within a process?

We do have a requirement to run multiple runtime instances (multiple `wasm_runtime_full_init()` results with varying `init_args`) in a system without process isolation, and it encourages us to reduce the number of global variables. 

I understand it's a bit hard to imagine that multiple wasi-nn backends are required, even with that multiple runtime requirement. However, eventually, if other components in the source code tree follow the policy of minimizing global variables, I believe wasi-nn should do the same. My opinion is to leave the original code unmodified and make changes when implementing that feature, or perform the optimization now if the resources allow.

yamt:
> > do you imply to allow loading (dlopen/dlysm) a backend multiple times within a process?
> 
> We do have a requirement to run multiple runtime instances (multiple `wasm_runtime_full_init()` results with varying `init_args`) in a system without process isolation, and it encourages us to reduce the number of global variables.

in that case this table should be moved to a structure to describe a runtime, say `struct wasm_runtime`, not `WASINNContext`.

> I understand it's a bit hard to imagine that multiple wasi-nn backends are required, even with that multiple runtime requirement. However, eventually, if other components in the source code tree follow the policy of minimizing global variables, I believe wasi-nn should do the same. My opinion is to leave the original code unmodified and make changes when implementing that feature, or perform the optimization now if the resources allow.

actually this PR as it is would help it because it reduces the amount of code referencing the global variable. (from 4 functions to 1)

lum1n0us (approved):
`WASINNContext` was just a wild guess. Don't take it seriously. 

I have to admit, you are right; the amount of code referencing is reduced. Let's follow your approach and see how it goes.

yamt:
the ci failure looks interesting. maybe worth investigation.
it doesn't seem related to this PR though.
```
Internal ctest changing into directory: /Users/runner/work/wasm-micro-runtime/wasm-micro-runtime/samples/wasm-c-api/build
Test project /Users/runner/work/wasm-micro-runtime/wasm-micro-runtime/samples/wasm-c-api/build
      Start  1: Test_callback
 1/10 Test  #1: Test_callback ....................   Passed    0.04 sec
      Start  2: Test_callback_chain
 2/10 Test  #2: Test_callback_chain ..............   Passed    0.08 sec
      Start  3: Test_empty_imports
 3/10 Test  #3: Test_empty_imports ...............   Passed    0.02 sec
      Start  4: Test_global
 4/10 Test  #4: Test_global ......................   Passed    0.10 sec
      Start  5: Test_hello
 5/10 Test  #5: Test_hello .......................   Passed    0.04 sec
      Start  6: Test_hostref
 6/10 Test  #6: Test_hostref .....................***Exception: Illegal  0.04 sec
Initializing...
Loading binary...
Compiling module...
[00:15:25:111 - 7FF856588A00]: failed to link function (, f): WASM module instantiate failed: unknown import
Creating callback...
Instantiating module...
Extracting exports...
Creating host references...
Accessing global...
call_v_r... okay
call_r_v... okay
call_v_r... okay
call_r_v... okay
call_v_r... okay
call_r_v... okay
call_v_r... okay
Accessing table...
call_i_r... okay
call_i_r... okay
call_ir_v... okay
call_ir_v... okay
call_i_r... okay
call_i_r... okay
call_ir_v... okay
call_i_r... okay
call_i_r... okay
Accessing function...
call_r_r... Calling back...
> > 0x0
okay
call_r_r... Calling back...
> > 0x1
okay
call_r_r... 

      Start  7: Test_memory
 7/10 Test  #7: Test_memory ......................   Passed    0.04 sec
      Start  8: Test_reflect
 8/10 Test  #8: Test_reflect .....................   Passed    0.09 sec
      Start  9: Test_table
 9/10 Test  #9: Test_table .......................   Passed    0.09 sec
      Start 10: Test_trap
10/10 Test #10: Test_trap ........................   Passed    0.07 sec

90% tests passed, 1 tests failed out of 10

Total Test time (real) =   0.69 sec

The following tests FAILED:
	  6 - Test_hostref (ILLEGAL)
Errors while running CTest
Error: Process completed with exit code 8.
```

yamt:
> the ci failure looks interesting. maybe worth investigation. it doesn't seem related to this PR though.

i filed an issue and restarted the job.
https://github.com/bytecodealliance/wasm-micro-runtime/issues/4316

Issue #4314, opened 2025-06-02
wasi-nn updates a global without locks


yamt:
wasi-nn seems to allow multiple modules to modify [a global](https://github.com/bytecodealliance/wasm-micro-runtime/blob/2a303861cc916dc182b7fecaa0aacc1b797e7ac6/core/iwasm/libraries/wasi-nn/src/wasi_nn.c#L32) without any serializations.
maybe it's mostly safe if dlopen() provides reference count (as macOS and linux do. i haven't checked what posix says about this) and the cpu does not tear pointer stores and maintain total ordering.
however, even for such platforms, it's probably nicer to have an explicit locking.
after all, this logic is not performance critical. simplicity is more important.

PR #4319, opened 2025-06-03
wasi-nn: protect the backend lookup table with a lock

yamt:
this would avoid potential issues when multiple instances happen to make an attempt to load a backend at the same time.

Fixes: https://github.com/bytecodealliance/wasm-micro-runtime/issues/4314


TianlongLiang (approved):
LGTM

PR #4336, opened 2025-06-05
Add Zephyr setup action and CI on Zephyr

lum1n0us:


lum1n0us:
See if I can learn something from https://github.com/project-ocre/ocre-runtime/blob/main/.github/workflows/build.yml.

PR #4350, opened 2025-06-10
Enable aot memory64 sw bounds checks by default

TianlongLiang:


wamr-compiler/main.c
```diff
@@ -137,9 +137,10 @@ print_help()
     printf("                              3 - Small code model\n");
     printf("  -sgx                      Generate code for SGX platform (Intel Software Guard Extensions)\n");
     printf("  --bounds-checks=1/0       Enable or disable the bounds checks for memory access:\n");
-    printf("                              by default it is disabled in all 64-bit platforms except SGX and\n");
-    printf("                              in these platforms runtime does bounds checks with hardware trap,\n");
-    printf("                              and by default it is enabled in all 32-bit platforms\n");
+    printf("                              By default, it is disabled on all 64-bit platforms except SGX or when\n");
+    printf("                              memory64 is enabled. In these platform and feature combinations,\n");
+    printf("                              the runtime performs bounds checks with a hardware trap;\n");
+    printf("                              By default, it is enabled on all 32-bit platforms or when using memory64.\n");
```
lum1n0us: My version:
This flag controls bounds checking with a software check. On 64-bit platforms, it is disabled by default, using a hardware trap if supported, except when SGX or memory64 is enabled, which defaults to a software check. On 32-bit platforms, the flag is enabled by default, using a software check due to the lack of hardware support.

core/iwasm/compilation/aot_llvm.c
```diff
@@ -3204,6 +3204,21 @@ aot_create_comp_context(const AOTCompData *comp_data, aot_comp_option_t option)
 
 #if WASM_ENABLE_WAMR_COMPILER != 0
     WASMModule *wasm_module = (WASMModule *)comp_data->wasm_module;
+    bool is_memory64 = false;
+
+    /* TODO: multi-memories for now assuming the memory idx type is
```
lum1n0us:
```suggestion
    /* TODO: multi-memories for now assuming the flag of a memory type is
```

PR #4359, opened 2025-06-12
improve installation steps for wasi-sdk and wabt on Windows

lum1n0us:


.github/actions/install-wasi-sdk-wabt/action.yml
```diff
@@ -30,51 +30,96 @@ runs:
...
+        choco install -y wget
+
+        mkdir -p /opt/wasi-sdk
+        mkdir -p /opt/wabt
+
+        echo "Downloading wasi-sdk for Windows..."
+        wget -O wasi-sdk.tar.gz --progress=dot:giga https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-25/wasi-sdk-25.0-x86_64-windows.tar.gz
+
+        echo "Extracting wasi-sdk..."
+        tar --strip-components=1 -xf wasi-sdk.tar.gz -C /opt/wasi-sdk
```
yamt, line 114: maybe linux/macos can use this too?

yamt (approved):

.github/actions/install-wasi-sdk-wabt/action.yml
lum1n0us: `--strip-components=1`? Both(on linux and macos) are using soft links and need to maintain their top-level directory names. FYI, Windows might also need to keep two paths, `/opt/wasi-sdk-*` and `/opt/wasi-sdk`, if there is no global modification regarding PATHS for the wabt and wasi-sdk toolchain.

yamt: >  FYI, Windows might also need to keep two paths, /opt/wasi-sdk-* and /opt/wasi-sdk, if there is no global modification regarding PATHS for the wabt and wasi-sdk toolchain.

ok. let's see how it goes.

Issue #4288, opened 2025-05-22
distribute headers for our extensions for wasm module developers


yamt:
we have a few things which define the abi/api for wasm modules.
including:
```
core/iwasm/libraries/wasi-nn/include/wasi_nn.h
core/iwasm/libraries/lib-socket/inc/wasi_socket_ext.h
```

it would be more user-friendly to distribute those files as a separate package/tarball/whatever
because a developer of wasm module usually doesn't want to have a copy of the whole wamr source tree
just to use those apis.

lum1n0us:
[WAMR_SDK](https://github.com/bytecodealliance/wasm-micro-runtime/tree/main/wamr-sdk) was intended to be a toolchain that allows users to follow WAMR's private WASM headers. It initially included three features: wasi-nn (very early version), pthread, and sockets. Later, wasi developed its own thread implementation, wasi-thread, and the use of pthread in WAMR is not encouraged. Similarly, wasi-nn is also not encouraged due to the creation of an (almost) [official Rust API](https://github.com/bytecodealliance/wasi-nn).

Undoubtedly, we were considering not maintaining the WAMR_SDK before this issue was raised.

WAMR_SDK was designed to work with `WAMR_BUILD_LIBC_BUILTIN` (and app-framework, which is also not maintained; you might find it in .github/workflows/build_wamr_sdk.yml). However, I guess now we are hoping it's in a form like wasi-libc + socket. Please let me know your comments about WAMR_SDK.

yamt:
> [WAMR_SDK](https://github.com/bytecodealliance/wasm-micro-runtime/tree/main/wamr-sdk) was intended to be a toolchain that allows users to follow WAMR's private WASM headers. It initially included three features: wasi-nn (very early version), pthread, and sockets. Later, wasi developed its own thread implementation, wasi-thread, and the use of pthread in WAMR is not encouraged. Similarly, wasi-nn is also not encouraged due to the creation of an (almost) [official Rust API](https://github.com/bytecodealliance/wasi-nn).
> 
> Undoubtedly, we were considering not maintaining the WAMR_SDK before this issue was raised.
> 
> WAMR_SDK was designed to work with `WAMR_BUILD_LIBC_BUILTIN` (and app-framework, which is also not maintained; you might find it in .github/workflows/build_wamr_sdk.yml). However, I guess now we are hoping it's in a form like wasi-libc + socket. Please let me know your comments about WAMR_SDK.

[...]

PR #4360, opened 2025-06-12
wamr-wasi-extensions: add lib-socket things

yamt:
cf. https://github.com/bytecodealliance/wasm-micro-runtime/issues/4288


yamt:
> BTW, should we also add CI for test wamr-wasi-extensions in the future? Not sure whether the wasi-nn can be added easily, but I think socket should be relatively easy.

it's on my todo list.
at least it should be build-tested.

lum1n0us (approved):

PR #4361, opened 2025-06-13
wasi_nn_openvino.c: remove pre/postprocessing and layout assumptions

yamt:
as wasi-nn doesn't have these concepts, the best we can do without risking breaking certain applications here is to pass through tensors as they are.

this matches wasmtime's behavior.

tested with:

* wasmtime classification-example (with this change, this example fails on tensor size mismatch instead of implicitly resizing it.)

* license-plate-recognition-barrier-0007, a converted version with non-fp32 output. [1] (with this change, this model outputs integers as expected.)

[1] https://github.com/openvinotoolkit/open_model_zoo/tree/cd7ebe313b69372763e76b82e5d24935308fece4/models/public/license-plate-recognition-barrier-0007

### DIFF
--- a/core/iwasm/libraries/wasi-nn/src/wasi_nn_openvino.c
+++ b/core/iwasm/libraries/wasi-nn/src/wasi_nn_openvino.c
@@ -32,8 +32,6 @@ typedef struct {
     void *weight_data;
     ov_tensor_t *weights_tensor;
     ov_model_t *model;
-    /* add prepostprocess */
-    ov_model_t *new_model;
     ov_compiled_model_t *compiled_model;
     ov_infer_request_t *infer_request;
     ov_tensor_t *input_tensor;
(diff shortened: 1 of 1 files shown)


yamt:
the ci failed with the infamous SGX 143.
i filed https://github.com/bytecodealliance/wasm-micro-runtime/issues/4363 and restarted the job.

lum1n0us:
@yamt Would you mind adding license-plate-recognition-barrier-0007into https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/libraries/wasi-nn/test/Dockerfile.wasi-nn-smoke (in a new PR if prefer)?

yamt:
> @yamt Would you mind adding license-plate-recognition-barrier-0007into https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/libraries/wasi-nn/test/Dockerfile.wasi-nn-smoke

i used https://github.com/bytecodealliance/wasm-micro-runtime/pull/4373 to run the model.
i haven't used core/iwasm/libraries/wasi-nn/test. maybe i will take a look later.

> (in a new PR if prefer)?

yes.

lum1n0us (approved):

Issue #4265, opened 2025-05-09
Send an empty/error reply when the debugger fails.


lum1n0us:
Here if we directly return without any response, the debugger client will wait forever. A better way may be sending an empty/error reply (like `write_packet(server, "");` or `write_packet(server, "E01");`) to tell the server there is an error.

https://sourceware.org/gdb/current/onlinedocs/gdb.html/Standard-Replies.html#Standard-Replies

_Originally posted by @xujuntwt95329 in https://github.com/bytecodealliance/wasm-micro-runtime/pull/4234#discussion_r2074619453_

PR #4362, opened 2025-06-13
send an empty/error reply from server

yviansu:
This problem relevant to https://github.com/bytecodealliance/wasm-micro-runtime/issues/4265 
Here I added a function to send the empty and error reply, will reply before the server returns.

**Attention: this feature has not been tested yet since no related test cases have been found.**

### DIFF
--- a/core/iwasm/libraries/debug-engine/handler.c
+++ b/core/iwasm/libraries/debug-engine/handler.c
@@ -180,7 +180,7 @@ process_wasm_global(WASMGDBServer *server, char *args)
    Not tested yet, but it should work.
  */
 static void
-fail(WASMGDBServer *server, const char *err)
+send_reply(WASMGDBServer *server, const char *err)
 {
     if (!err || !*err)
         write_packet(server, "");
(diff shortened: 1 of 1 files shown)


core/iwasm/libraries/debug-engine/handler.c
```diff
@@ -175,6 +175,19 @@ process_wasm_global(WASMGDBServer *server, char *args)
     os_mutex_unlock(&tmpbuf_lock);
 }
 
+/* TODO: let server send an empty/error reply.
+   Original issue: 4265
+   Not tested yet, but it should work.
+ */
+static void
+fail(WASMGDBServer *server, const char *err)
```
TianlongLiang: Maybe it's better to rename it to `send_error_reply` or something, fail as a function name is a little strange since WAMR often uses fail as a goto label

TianlongLiang (approved):
LGTM with a minor comment

core/iwasm/libraries/debug-engine/handler.c
yamt: +1

PR #4372, opened 2025-06-17
improve logic of `heap_type` validation when `ref.null`

Septa2112:
PR #4300 introduced the rationale for validating heap_type. 

This patch moves the validation before the computation of type1 to prevent potential overflow.

### DIFF
--- a/core/iwasm/interpreter/wasm_loader.c
+++ b/core/iwasm/interpreter/wasm_loader.c
@@ -1057,11 +1057,9 @@ load_init_expr(WASMModule *module, const uint8 **p_buf, const uint8 *buf_end,
                     }
                     cur_ref_type.ref_ht_common.ref_type =
                         (uint8)((int32)0x80 + heap_type);
-                    if (!push_const_expr_stack(
-                            &const_expr_ctx, flag, cur_ref_type.ref_type, NULL,
-                            0, &cur_value, error_buf, error_buf_size))
-                    if (!push_const_expr_stack(&const_expr_ctx, flag, type1,
-                                               NULL, 0, &cur_value,
+                    if (!push_const_expr_stack(&const_expr_ctx, flag,
+                                               cur_ref_type.ref_type, NULL, 0,
+                                               &cur_value,
 #if WASM_ENABLE_EXTENDED_CONST_EXPR != 0
                                                NULL,
 #endif


Septa2112:
Further improved the heap type check logic:

1. Moved the heap type validation ahead of the `if (!is_byte_a_type...)` check. Depending on whether `heap_type >= 0`, different methods are used to get `type1`.

2. The `type_index` check is now performed only when `heap_type >= 0`, to ensure that `type1` is correctly retrieved in this case.

3. Why not checking `type_index`  when `heap_type < 0`? Because in the current WAMR implementation, if `wasm_is_valid_heap_type(heap_type)` is `true`, then the condition `if (!is_byte_a_type(type1) || wasm_is_type_multi_byte_type(type1))` will always be `false`.

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -843,21 +843,29 @@ load_init_expr(WASMModule *module, const uint8 **p_buf, const uint8 *buf_end,
...
+                    type1 = cur_ref_type.ref_type;
+                }
+                else {
+                    if (!wasm_is_valid_heap_type(heap_type)) {
+                        set_error_buf_v(error_buf, error_buf_size,
+                                        "unknown type %d", heap_type);
+                        goto fail;
+                    }
+                    type1 = (uint8)((int32)0x80 + heap_type);
+                }
```
lum1n0us: better have a comment @L867 to hint that `if heap_type < 0`, `if (!is_byte_a_type(type1) || wasm_is_type_multi_byte_type(type1))` will be false.

lum1n0us: Another suggestion is to merge those `if...else..` blocks, like:

```c
if (heap_type >= 0) {
    if (!check_type_index(module, module->type_count, heap_type,
                          error_buf, error_buf_size)) {
        goto fail;
    }
    wasm_set_refheaptype_typeidx(&cur_ref_type.ref_ht_typeidx,
                                 true, heap_type);
    if (!push_const_expr_stack(&const_expr_ctx, flag,
                               cur_ref_type.ref_type,
                               &cur_ref_type, 0, &cur_value,
                               error_buf, error_buf_size))
        goto fail;

}
else {
    if (!wasm_is_valid_heap_type(heap_type)) {
        set_error_buf_v(error_buf, error_buf_size,
                        "unknown type %d", heap_type);
        goto fail;
    }
    type1 = (uint8)((int32)0x80 + heap_type);
    if (!push_const_expr_stack(&const_expr_ctx, flag, type1,
                               NULL, 0, &cur_value, error_buf,
                               error_buf_size))
        goto fail;
}
```

Septa2112: The related comment has been added.

Septa2112: Addressed.

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -843,28 +843,49 @@ load_init_expr(WASMModule *module, const uint8 **p_buf, const uint8 *buf_end,
...
+                    }
                     wasm_set_refheaptype_typeidx(&cur_ref_type.ref_ht_typeidx,
-                                                 true, type_idx);
-                    if (!push_const_expr_stack(&const_expr_ctx, flag,
-                                               cur_ref_type.ref_type,
-                                               &cur_ref_type, 0, &cur_value,
-                                               error_buf, error_buf_size))
-                        goto fail;
+                                                 true, heap_type);
+                    type1 = cur_ref_type.ref_type;
```
lum1n0us: IIUC, after `wasm_set_refheaptype_typeidx()`, `type1` is always `REF_TYPE_HT_NULLABLE`, then it will always pass `!is_byte_a_type() || wasm_is_type_multi_byte_type()`. So there is no need to have a `else` branch in L865.

Septa2112: Addressed.

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -830,32 +830,49 @@ load_init_expr(WASMModule *module, const uint8 **p_buf, const uint8 *buf_end,
...
+                 * or the type C.types[typeidx] must be defined in the context.
+                 */
+                if (heap_type >= 0) {
+                    if (!check_type_index(module, module->type_count, heap_type,
+                                          error_buf, error_buf_size)) {
+                        goto fail;
+                    }
+                }
+                else {
+                    if (!wasm_is_valid_heap_type(heap_type)) {
```
lum1n0us, line 1053: Guess `heap_type` should be `type1`.

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -830,32 +830,49 @@ load_init_expr(WASMModule *module, const uint8 **p_buf, const uint8 *buf_end,
             case INIT_EXPR_TYPE_REFNULL_CONST:
             {
                 uint8 type1;
-
-#if WASM_ENABLE_GC == 0
+#if WASM_ENABLE_GC != 0
+                const uint8 *p_copy = p;
+                int32 heap_type;
+                read_leb_int32(p_copy, p_end, heap_type);
+#endif
```
lum1n0us: If following the spec, s33 should be a type index. might rename it as type_index better.

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -830,32 +830,49 @@ load_init_expr(WASMModule *module, const uint8 **p_buf, const uint8 *buf_end,
...
                 CHECK_BUF(p, p_end, 1);
                 type1 = read_uint8(p);
 
+#if WASM_ENABLE_GC == 0
```
lum1n0us: like above, in L839, rename `type1` to `heap_type` or `abs_heaptype`.

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -830,32 +830,49 @@ load_init_expr(WASMModule *module, const uint8 **p_buf, const uint8 *buf_end,
...
+                    }
+                }
+                else {
+                    if (!wasm_is_valid_heap_type(heap_type)) {
+                        set_error_buf_v(error_buf, error_buf_size,
+                                        "unknown type %d", heap_type);
+                        goto fail;
+                    }
+                }
+
```
lum1n0us: after `wasm_is_valid_heap_type()`, not sure we still need `is_byte_a_type()` and `wasm_is_type_multi_byte_type()`.

core/iwasm/interpreter/wasm_loader.c
lum1n0us: Another thing, L872 will read it again. Seems we can remove either one.

Septa2112: I think it might not be appropriate to rename `heap_type` to `type_idx` here:
1. `type_idx` has already been defined as a `uint32` at the beginning of this function.
2. Based on the surrounding lines of code in function `wasm_loader_prepare_bytecode` when `WASM_OP_REF_NULL`, I think it's clearer and more consistent to keep the name `heap_type`.
https://github.com/bytecodealliance/wasm-micro-runtime/blob/18d4227ab6593bb1870aac025d351cce9e3c1041/core/iwasm/interpreter/wasm_loader.c#L13154-L13169

core/iwasm/interpreter/wasm_loader.c
Septa2112: Likewise, considering the same code context in function `wasm_loader_prepare_bytecode` when `WASM_OP_REF_NULL` , it might be better to keep the check for `int32 heap_type` here.

lum1n0us: But in this case, not reusing exactly the same code. Maybe it's possible by just copying the content from `wasm_loader_prepare_bytecode()`. However, don't use mixed logic here.

core/iwasm/interpreter/wasm_loader.c
lum1n0us: In this `case`, there is also a variable named `type1`, which is a bad name. Keeping two variables with unclear meanings will make the code difficult to understand.

lum1n0us (changes requested):

lum1n0us (approved):

Issue #4313, opened 2025-06-02
wasi-nn WASINNContext management seems broken


yamt:
`wasm_runtime_get_wasi_nn_ctx` associates `WASINNContext` to `wasm_module_inst_t`
with a hash, using the pointer value of `wasm_module_inst_t` as a key.
the context and association is left as they are until runtime shutdown. (`wasm_runtime_destroy`)
it seems unsafe because a `wasm_module_inst_t` pointer does not identify a module instance.
once the pointer is freed by `wasm_runtime_deinstantiate`, the same pointer value can be used for an unrelated module instance.

kylo5aby:
the root cause is the hash map approach it used isn't satisfied this scenario, and then it causes unsafe key and lifecycle mismatch. 

iiuc, here we can replace the hash map approach with WAMR's native module instance context system, following the same pattern used by wasi itself.

in `wasi_nn.c`, use a safe context key:

```c
static void *g_wasi_nn_context_key = NULL;

bool wasi_nn_initialize()
{
    // create context key with destructor callback
    g_wasi_nn_context_key = wasm_native_create_context_key(wasi_nn_context_dtor);
   ...
}
```

once module_inst is destroyed, callback `wasi_nn_context_dtor` will destroyed related wasi nn context.

rewrite context management:
```c
static WASINNContext *
wasm_runtime_get_wasi_nn_ctx(wasm_module_inst_t instance)
{
    // use WAMR's native context system
    WASINNContext *wasi_nn_ctx = 
        (WASINNContext *)wasm_native_get_context(instance, g_wasi_nn_context_key);
    
    if (wasi_nn_ctx == NULL) {
         ...
        wasm_native_set_context(instance, g_wasi_nn_context_key, wasi_nn_ctx);
    }

    return wasi_nn_ctx;
}
```

do you have any suggestions?

yamt:
> iiuc, here we can replace the hash map approach with WAMR's native module instance context system, following the same pattern used by wasi itself.

i agree.
to take the approach, we should make wasi-nn host code thread safe first, because it would make a context shared among threads.
cf. https://github.com/bytecodealliance/wasm-micro-runtime/issues/2430

kylo5aby:
> > iiuc, here we can replace the hash map approach with WAMR's native module instance context system, following the same pattern used by wasi itself.
> 
[...]

Issue #2430, opened 2023-08-07
wasi-nn + threads


yamt:
wasm_deinstantiate and aot_deinstantiate call wasi_nn_destroy only for !is_sub_inst.

on the other hand, wasm_runtime_get_wasi_nn_ctx seems to allocate the context for each instances.

probaby it just means that (this implementation of?) wasi-nn is not compatible with threads.

tonibofarull:
For the wasi-nn we have a new (not yet PR'd) implementation where the context is managed within the library.
The motivations are the following:
1. Allowing the user to build is as a native library.
2. Avoid coupling with WAMR as much as possible.

yamt:
> What would be required to support it with threads?

i don't know.
probably
* share the context among instances. (wasi-threads and lib-pthread are using instance per thread model)
* add some locks

yamt:
>     * share the context among instances. (wasi-threads and lib-pthread are using instance per thread model)
> 
>     * add some locks

these have been implemented since then.

Issue #4386, opened 2025-06-18
os_mutex_trylock


yamt:
our platform api doesn't have os_mutex_trylock.
is it intentional? eg. impossible/inefficient for some platforms?

TianlongLiang:
I think it's intentional, but not for the reason that it's impossible/inefficient for some platforms. The other APIs, such as `os_mutex_lock` in `platform_api_vmcore.h` are required platform APIs to support a certain WAMR feature, in this case, the thread manager, and we don't need os_mutex_trylock to do that; it's not a must, so we don't add it.

yamt:
> I think it's intentional, but not for the reason that it's impossible/inefficient for some platforms. The other APIs, such as `os_mutex_lock` in `platform_api_vmcore.h` are required platform APIs to support a certain WAMR feature, in this case, the thread manager, and we don't need os_mutex_trylock to do that; it's not a must, so we don't add it.

ok.
it makes sense to keep the platform api minimum.

for the record, i recently happened to have a use case. (https://github.com/bytecodealliance/wasm-micro-runtime/pull/4387)
i don't claim this use case alone is enough reason to add trylock though.

maybe we can revisit when/if more use cases come up.
i'm going to close this for now.

PR #4387, opened 2025-06-18
wasi-nn: add minimum serialization on WASINNContext

yamt:
currently this is not necessary because context (WASINNContext) is local to instance. (wasm_module_instance_t)

i plan to make a context shared among instances in a cluster when fixing https://github.com/bytecodealliance/wasm-micro-runtime/issues/4313. this is a preparation for that direction.

an obvious alternative is to tweak the module instance context APIs to allow declaring some kind of contexts instance-local. but i feel, in this particular case, it's more natural to make "wasi-nn handles" shared among threads within a "process".

note that, spec-wise, how wasi-nn behaves wrt threads is not defined at all because wasi officially doesn't have threads yet. i suppose, at this point, that how wasi-nn interacts with wasi-threads is something we need to define by ourselves, especially when we are using an outdated wasi-nn version.

with this change, if a thread attempts to access a context while another thread is using it, we simply make the operation fail with the "busy" error. this is intended for the mimimum serialization to avoid problems like crashes/leaks/etc. this is not intended to allow parallelism or such.

no functional changes are intended at this point yet.

cf.
https://github.com/bytecodealliance/wasm-micro-runtime/issues/4313
https://github.com/bytecodealliance/wasm-micro-runtime/issues/2430

### DIFF
--- a/core/iwasm/libraries/wasi-nn/src/wasi_nn.c
+++ b/core/iwasm/libraries/wasi-nn/src/wasi_nn.c
@@ -98,6 +98,8 @@ wasi_nn_ctx_destroy(WASINNContext *wasi_nn_ctx)
     NN_DBG_PRINTF("-> is_model_loaded: %d", wasi_nn_ctx->is_model_loaded);
     NN_DBG_PRINTF("-> current_encoding: %d", wasi_nn_ctx->backend);
 
+    bh_assert(!wasi_nn_ctx->busy);
+
     /* deinit() the backend */
     if (wasi_nn_ctx->is_backend_ctx_initialized) {
         wasi_nn_error res;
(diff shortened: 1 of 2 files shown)


TianlongLiang (approved):

lum1n0us (approved):
Just saying, I used to think you would prefer to use atomic variables to achieve similar functionality.

yamt:
> Just saying, I used to think you would prefer to use atomic variables to achieve similar functionality.

originally i wanted to use trylock. (and filed https://github.com/bytecodealliance/wasm-micro-runtime/issues/4386)

direct uses of atomics would make it a bit difficult to read/maintain. it can be done later if/when it turns out desirable.

Issue #4376, opened 2025-06-17
wasi_nn_tensorflowlite.cpp get_output seems to return the number of float32


yamt:
glancing at the code, wasi_nn_tensorflowlite.cpp seems to return the number of fp32 values via output_tensor_size.
i believe it should return number of bytes.

well, at least for wasi_ephemeral_nn.
i'm not sure what should do for the legacy wamr-specific "wasi_nn" api.
it seems the size of tensor sizes has always been inconsistent for "wasi_nn". cf. https://github.com/bytecodealliance/wasm-micro-runtime/issues/3223

yamt:
https://github.com/bytecodealliance/wasm-micro-runtime/pull/4390 fixed this for wasi_ephemeral_nn.
the legacy abi was left intact for now.

PR #4390, opened 2025-06-19
wasi_nn_tensorflowlite.cpp: fix get_output return size

yamt:
it should be byte size, not the number of (fp32) values.

i'm ambivalent about how to deal with the compatibility for the legacy wamr-specific "wasi_nn". for now, i avoided changing it. (so that existing tests using the legacy abi, namely test_tensorflow.c and test_tensorflow_quantized.c, pass as they are.) if we have any users who still want to use the legacy abi, i suppose they consider the compatibility is more important than the consistency with other backends.

cf. https://github.com/bytecodealliance/wasm-micro-runtime/issues/4376

### DIFF
--- a/core/iwasm/libraries/wasi-nn/src/wasi_nn_tensorflowlite.cpp
+++ b/core/iwasm/libraries/wasi-nn/src/wasi_nn_tensorflowlite.cpp
@@ -389,23 +389,34 @@ get_output(void *tflite_ctx, graph_execution_context ctx, uint32_t index,
         return too_large;
     }
 
-    uint32_t model_tensor_size = 1;
-    for (int i = 0; i < (int)tensor->dims->size; ++i)
-        model_tensor_size *= (uint32_t)tensor->dims->data[i];
-
-    if (*output_tensor_size < model_tensor_size) {
-        NN_ERR_PRINTF("Insufficient memory to copy tensor %d", index);
-        return too_large;
-    }
-
     if (tensor->quantization.type == kTfLiteNoQuantization) {
         NN_DBG_PRINTF("No quantization information");
-        float *ot =
-            tfl_ctx->interpreters[ctx].interpreter->typed_output_tensor<float>(
-                index);
-
-        int size = model_tensor_size * sizeof(float);
-        bh_memcpy_s(output_tensor, size, ot, size);
+#if WASM_ENABLE_WASI_EPHEMERAL_NN != 0
+        if (*output_tensor_size < tensor->bytes) {
+            NN_ERR_PRINTF("Insufficient memory to copy tensor %d", index);
+            return too_large;
+        }
+#else
+        /*
+         * for now, maintain the bug-to-bug compatibility with the old abi,
+         * where the size here is the number of fp32, not bytes.
+         */
+        if (*output_tensor_size < tensor->bytes / sizeof(float)) {
+            NN_ERR_PRINTF("Insufficient memory to copy tensor %d", index);
+            return too_large;
+        }
+#endif
+        bh_memcpy_s(output_tensor, *output_tensor_size, tensor->data.data,
+                    tensor->bytes);
+#if WASM_ENABLE_WASI_EPHEMERAL_NN != 0
+        *output_tensor_size = tensor->bytes;
+#else
+        /*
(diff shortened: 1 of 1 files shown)


yamt:
restrarting [SGX 143](https://github.com/bytecodealliance/wasm-micro-runtime/issues/4363)
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/15746783829/job/44384528925?pr=4390

core/iwasm/libraries/wasi-nn/src/wasi_nn_tensorflowlite.cpp
```diff
@@ -414,6 +410,16 @@ get_output(void *tflite_ctx, graph_execution_context ctx, uint32_t index,
             NN_ERR_PRINTF("Quantization per channel is not supported");
             return runtime_error;
         }
+
+        uint32_t model_tensor_size = 1;
+        for (int i = 0; i < (int)tensor->dims->size; ++i)
+            model_tensor_size *= (uint32_t)tensor->dims->data[i];
+
+        if (*output_tensor_size / sizeof(float) < model_tensor_size) {
```
lum1n0us, line 434: based on L439, there should be two checks(for consistency)? like

```c
  if (*output_tensor_size / sizeof(float) < model_tensor_size) {
#else
  if (*output_tensor_size < model_tensor_size) {
#endif 
```

core/iwasm/libraries/wasi-nn/src/wasi_nn_tensorflowlite.cpp
```diff
@@ -389,23 +389,19 @@ get_output(void *tflite_ctx, graph_execution_context ctx, uint32_t index,
...
+        if (*output_tensor_size < tensor->bytes) {
+            NN_ERR_PRINTF("Insufficient memory to copy tensor %d", index);
+            return too_large;
+        }
+        bh_memcpy_s(output_tensor, *output_tensor_size, tensor->data.data,
+                    tensor->bytes);
+#if WASM_ENABLE_WASI_EPHEMERAL_NN != 0
+        *output_tensor_size = tensor->bytes;
+#else
+        *output_tensor_size = tensor->bytes / sizeof(float);
```
lum1n0us, line 418: if `bh_memcpy_s()` actually write `tensor->bytes` bytes, L403 should be `*output_tensor_size = tensor->bytes;`?

yamt: this ifdef block is for compatibility. (for the legacy "wasi_nn", it returns the number of floats)

core/iwasm/libraries/wasi-nn/src/wasi_nn_tensorflowlite.cpp
yamt: yes. i will fix. thank you.

yamt:
> Not familiar with the LLM model, it seems there are assumptions about the format under different tensor->quantization.type.

i'm not quite happy with these assumptions because it makes us incompatible with wasmedge.

> For me, it would be better to add comments to make it easier for readers to follow.

totally agree.

core/iwasm/libraries/wasi-nn/src/wasi_nn_tensorflowlite.cpp
yamt: i added a few comments

core/iwasm/libraries/wasi-nn/src/wasi_nn_tensorflowlite.cpp
yamt: fixed.

core/iwasm/libraries/wasi-nn/src/wasi_nn_tensorflowlite.cpp
lum1n0us: if  `output_tensor_size` is the number of fp32, L394 would be something like `if (*output_tensor_size < tensor->bytes/sizeof(float))` when `WASM_ENABLE_WASI_EPHEMERAL_NN == 0` ?

yamt: right. fixed. thank you.

PR #4392, opened 2025-06-19
wamr-wasi-extensions/socket: disable reference-types

yamt:
and add a comment to explain why.

### DIFF
--- a/wamr-wasi-extensions/socket/CMakeLists.txt
+++ b/wamr-wasi-extensions/socket/CMakeLists.txt
@@ -13,6 +13,12 @@ target_include_directories(wamr-wasi-socket
                            $<BUILD_INTERFACE:${wasi_socket_header_dir}>
                            $<INSTALL_INTERFACE:include>)
 
+# as this is a library, be extra conservative about wasm features
+# to improve compatibilities. as this particular library is just a
+# simple static stub, extra wasm features won't benefit us much anyway.
+# note that LLVM-19 enables reference-types by default.
+target_compile_options(wamr-wasi-socket PRIVATE -mno-reference-types)
+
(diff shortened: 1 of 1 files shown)


lum1n0us (commented):
Reference-types is an MVP feature, and toolchains and runtimes should support it. Besides compatibility, is there any other reason we should disable reference-types?

yamt:
> Reference-types is an MVP feature, and toolchains and runtimes should support it.

it's a post-mvp feature as far as i know.

> Besides compatibility, is there any other reason we should disable reference-types?

no.

lum1n0us:
Yes, to be more accurate, it is a finished proposals. https://github.com/WebAssembly/proposals/blob/main/finished-proposals.md
Now it is a part of WebAssembly Spec.

yamt:
anyway, it isn't ubiquitous. even we ourselves don't enable it everywhere.
also, it doesn't benefit this particular library.
if you have reasons to prefer enabling it, please explain.

lum1n0us:
> even we ourselves don't enable it everywhere.

after https://github.com/bytecodealliance/wasm-micro-runtime/pull/3894, WAMR should always support reference type by default.

---

There is no obvious, specific reason to reject this modification. So, let's merge it first?

yamt:
> > even we ourselves don't enable it everywhere.
> 
> after #3894, WAMR should always support reference type by default.

my impression is that it's far from always.
for example, it's disabled by default for nuttx.

anyway, regardless of our default, some people prefer to disable it for size reasons.

> There is no obvious, specific reason to reject this modification. So, let's merge it first?

yes, please.

lum1n0us (approved):

PR #4394, opened 2025-06-19
CI: build wamr-wasi-extensions

yamt:


.github/workflows/wamr_wasi_extensions.yml
```diff
@@ -0,0 +1,57 @@
...
+          mkdir dist
+          ./build_libs.sh $(pwd)/dist/wamr-wasi-extensions
+        working-directory: wamr-wasi-extensions
+
+      - name: Build wamr-wasi-extensions samples
+        run: |
+          ./build_samples.sh $(pwd)/dist/wamr-wasi-extensions
+        working-directory: wamr-wasi-extensions
+
+      - name: Upload artifacts
```
yamt, line 51: @lum1n0us
eventually i want to ship this zip file as a part of release. can you guide me how it can be done?

TianlongLiang: For other release artifacts, take iwasm as an example, firstly define a reusable workflow: .github/workflows/build_iwasm_release.yml to build iwasm as an artifact, use actions/upload-release-asset@v1 to upload the artifact. and then use it in realease_process.yml, pass different input to build for different platforms

yamt: thank you. i will take a look.

yamt: i read build_iwasm_release.yml a bit.
is the binary we release (eg. build_iwasm_release.yml) completely independent from what we tested (eg. compilation_on_macos.yml) right now?

TianlongLiang: > i read build_iwasm_release.yml a bit. is the binary we release (eg. build_iwasm_release.yml) completely independent from what we tested (eg. compilation_on_macos.yml) right now?

Yes, it's independent

yamt: ok. then maybe i will leave this PR for on.pull_request and prepare another yml file for release workflow.

.github/workflows/wamr_wasi_extensions.yml
```diff
@@ -0,0 +1,57 @@
...
+        if: matrix.os == 'macos-14'
```
lum1n0us, line 52: Only available for macos-14?

yamt: i only upload it for macos-14 because the artifact is identical regardless of matrix.os.

lum1n0us (approved):

PR #3694, opened 2024-08-07
Overlook the access modes of STDIN, STDOUT and STDERR

lum1n0us:
https://github.com/bytecodealliance/wasm-micro-runtime/issues/3686

Issue #3686, opened 2024-08-05
[WASI] Should we overlook or restrict the access modes for stdin/stdout/stderr?


lum1n0us:
We've observed that in certain situations, the access modes for STDIN, STDOUT and STDERR do not behave as anticipated. For example, STDOUT may be read-only, while STDIN may be write-only.

It's important that we look into why these unexpected access modes are occurring. However, I'm wondering if we need to take steps to block these incorrect settings in *core/iwasm/libraries/libc-wasi*

Specifically, when determining file descriptor rights in the function `fd_determine_type_rights()`, should we assign fixed and unchangeable rights to STDIN, STDOUT and STDERR?

wenyongh:
I saw the PR #3694 sets the STDIN rights with:
```
(__WASI_RIGHT_FD_ADVISE | __WASI_RIGHT_FD_FILESTAT_GET | \
  __WASI_RIGHT_FD_READ | __WASI_RIGHT_POLL_FD_READWRITE)
```
not sure will it disable the write right for STDIN and disable the read right for STDOUT? It seems the similar issues were reported before, but after discussion with @yamt, we decided not to change the code:
https://github.com/bytecodealliance/wasm-micro-runtime/issues/2645#issuecomment-1790387978
https://github.com/bytecodealliance/wasm-micro-runtime/pull/2606

yamt:
> We've observed that in certain situations, the access modes for STDIN, STDOUT and STDERR do not behave as anticipated. For example, STDOUT may be read-only, while STDIN may be write-only.
> 
> It's important that we look into why these unexpected access modes are occurring. However, I'm wondering if we need to take steps to block these incorrect settings in _core/iwasm/libraries/libc-wasi_

yes, please investigate the situation further before making changes to libc-wasi.

lum1n0us:
The problem originated from integrating WAMR with runwasi, which aims to run a WebAssembly runtime within containerd. I haven't pinpointed the exact issue yet, but there are some unanticipated behaviors occurring within their sandbox environment. Like, `fcntl(STDOUT, F_GETFL, 0)` returns `O_RDONLY`. The unusual result will [strip off write bits](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c#L518-L519). 

Then I made an trial on linux.
``` bash
$ cat demo.c
int main() {
    int ret;

[...]

PR #4395, opened 2025-06-20
move stdio fd handle checking in fd_determine_type_rights to ensure type is properly set before return

SneezingCactus:
Before this change, fd_determine_type_rights would return early upon dealing with a stdio file handle without giving it any type (therefore being left as a __WASI_FILETYPE_UNKNOWN). This would end up having repercussions on wasi-sdk's implementation of vfprintf, which doesn't register line breaks properly if the file it's writing to (stdout in the case of printf) isn't a tty, which is mainly checked by seeing if the file is a CHARACTER_DEVICE.


lum1n0us:
> This would end up having repercussions on wasi-sdk's implementation of vfprintf, which doesn't register line breaks properly if the file it's writing to (stdout in the case of printf) isn't a tty, which is mainly checked by seeing if the file is a CHARACTER_DEVICE.

Would you mind pointing out the source code in wasi-libc related to that?

> `fd_determine_type_rights` would return early upon dealing with a stdio file handle without giving it any type

IIUC, `*type = buf.st_filetype;` is the key to ⬆️ problem. So how about moving L527~L540 to L469 and return directly.

yamt:
> > This would end up having repercussions on wasi-sdk's implementation of vfprintf, which doesn't register line breaks properly if the file it's writing to (stdout in the case of printf) isn't a tty, which is mainly checked by seeing if the file is a CHARACTER_DEVICE.
> 
> Would you mind pointing out the source code in wasi-libc related to that?

https://github.com/WebAssembly/wasi-libc/blob/1377a93acea031acd9b6e77a830360d88477319c/libc-bottom-half/sources/isatty.c

> > `fd_determine_type_rights` would return early upon dealing with a stdio file handle without giving it any type
> 
> IIUC, `*type = buf.st_filetype;` is the key to ⬆️ problem. So how about moving L527~L540 to L469 and return directly.

i feel it's better to revert https://github.com/bytecodealliance/wasm-micro-runtime/pull/3694 and investigate the original issue because it was based on a broken experiment: https://github.com/bytecodealliance/wasm-micro-runtime/issues/3686#issuecomment-2998903186

lum1n0us:
I feel this PR is primarily about ensuring `*type` is assigned as an output parameter, which the original code missed. We might want to reopen the discussion on 3686.

yamt:
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/15771043456/job/44455986874?pr=4395

PR #4397, opened 2025-06-20
Turn warnings into errors in CI

lum1n0us:
- Apply global warnings in config_common.cmake instead of maintaining them in separate files.
- Enable errors during CI when building iwasm and wamrc


wamr-compiler/CMakeLists.txt
```diff
@@ -198,13 +198,43 @@ if (NOT (CMAKE_C_COMPILER MATCHES ".*clang.*" OR CMAKE_C_COMPILER_ID MATCHES ".*
...
-  set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra -Wformat -Wformat-security \
-                                       -ffunction-sections -fdata-sections \
-                                       -Wno-unused-parameter -Wno-pedantic")
+  set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -ffunction-sections -fdata-sections")
   # Remove the extra spaces for better make log
   string (REGEX REPLACE "  *" " " CMAKE_C_FLAGS ${CMAKE_C_FLAGS})
 endif()
 
+# global additional warnings. Keep those options consistent with config_common.cmake.
+if (MSVC)
```
loganek: feels like a duplication of config_common.cmake - would it make sense to create warnings.cmake file and move the common code there, so it can be included when needed?

build-scripts/config_common.cmake
```diff
@@ -81,6 +81,38 @@ elseif (WAMR_BUILD_TARGET MATCHES "THUMB.*")
...
+  add_compile_options(/W4)
+else ()
+  # refer to https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html
+  add_compile_options(-Wall -Wextra -Wformat -Wformat-security -Wshadow)
+  # -pedantic causes warnings like "ISO C forbids initialization between function pointer and ‘void *’" which
+  #   requires by NativeSymbol
+  #
+  # -fpermissive causes warnings like "-fpermissive is valid for C++/ObjC++ but not for C"
+  #
+  # _Static_assert and _Alignof requires c11 or later
```
yamt: is this c11 comment intended to be here? why?

lum1n0us: Thank you for reminding me. I am inclined to start a discussion here, as we are using `_Static_assert` and `_Alignof`(and/or atomic), primarily in WASI. Do we need to designate C11 as our standard, or should we continue using a lower version, like C99, for cross-platform compatibility?

---

code in `wamr-compiler/CMakeLists.txt` 

```cmake
if (WAMR_BUILD_LIBC_WASI EQUAL 1)
  # Enable _Static_assert
  set (CMAKE_C_STANDARD 11)
  if (MSVC)
    add_compile_options(/experimental:c11atomics)
  endif()
else()
  set (CMAKE_C_STANDARD 99)
endif()
```

yamt: for modern linux/macOS, i guess it doesn't matter.
i have no idea about the situation of windows.
embedded platforms might have extra restrictions.

i personally prefer conservative options. ie. keep the use of the latest features limited/exceptional.

@loganek do you have opinions? iirc you were using an ancient version of gcc.

loganek: Yes, we do use gcc 4.9 for some platforms. Having said that, specifically `_Static_assert` and `_Alignof` are supported by gcc 4.9 (with a note that `_Static_assert` always requires both parameters).

TianlongLiang (approved):

lum1n0us:
@yamt @loganek  we are planning to merge this pull request, which will impact the current continuous integration or release process. Please provide your feedback.

build-scripts/warnings.cmake
```diff
@@ -0,0 +1,28 @@
+# Copyright (C) 2019 Intel Corporation.  All rights reserved.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# global additional warnings. Keep those options consistent with wamr-compiler/CMakeLists.txt.
```
yamt: i don't understand this comment. wamr-compiler/CMakeLists.txt includes this file, doesn't it?

lum1n0us: Oops, thanks for reminding

lum1n0us:
depends on https://github.com/bytecodealliance/wasm-micro-runtime/pull/4582

yamt (approved):

PR #4400, opened 2025-06-23
Shared heap enhancements for Interpreter and AOT

### LINKED ISSUE #4042: [RFC] Shared heap enhancements
# Feature: Shared heap enhancements

Propose two enhancements:
1. Shared heap created from preallocated memory buffer: The user can create a shared heap from a pre-allocated buffer and see that memory region as one large chunk; there's no need to dynamically manage it(malloc/free). The user needs to make sure the native address and size of that memory region are valid.
2. Introduce shared heap chain: The user can create a shared heap chain, from the wasm app point of view, it's still a continuous memory region in wasm app's point of view while in the native it can consist of multiple shared heaps (each of which is a continuous memory region). For example, one 500MB shared heap 1 and one 500 MB shared heap 2 form a chain, in Wasm's point of view, it's one 1GB shared heap.

## Benefit

After these enhancements, the data sharing between wasm apps, and between hosts can be more efficient and flexible. Admittedly shared heap management can be more complex for users, but it's similar to the zero-overhead principle. No overhead will be imposed for the users who don't use the shared heap enhancement or don't use the shared heap at all. 

## High level design

- New option for `SharedHeapInitArgs` to init a shared heap from the preallocated buffer

- One new export API to create a shared heap chain:

  ```C
  WASMSharedHeap *
  wasm_runtime_chain_shared_heaps(WASMSharedHeap *head, WASMSharedHeap 
  *body)
  ```

- The behavior of existing shared heap APIs: `wasm_runtime_attach_shared_heap` and `wasm_runtime_detach_shared_heap` will attach/detach the shared heap chain instead of the single shared heap. 

## Some implementation choices

1. In a shared heap chain, only one of the shared heap can be malloc/free. The rest of them have to be pre-allocated shared heap for management simplicity.

2. To create a shared heap chain, the related shared heaps can't be attached to any wasm apps. 

3. No other export API added, how the pre-allocated shared heap's address passes between wasms and host depends on runtime management, for example, using native API.

## Alternatives

The alternative can be instantiation linking + multi-memory to achieve a similar functionality. Further comparison can be arranged when the WAMR instantiation linking is ready and multi-memory is well supported in the upstream toolchain. My current estimation is that the performance won't be too much gap between the alternative and this enhancement, but the management can be potentially easier.

TianlongLiang:
For interpreter and AOT mode:

- Added a new type of shared heap, which can be initialized from a preallocated buffer.
- Introduced shared heap chain, allowing multiple shared heaps to be chained together; and added new export APIs to create/disconnect a shared heap chain.
- Updated related document and sample code, added more unit test cases.


core/iwasm/common/wasm_memory.c
```diff
@@ -233,6 +248,212 @@ wasm_runtime_create_shared_heap(SharedHeapInitArgs *init_args)
...
+    heap_handle_exist = head->heap_handle != NULL;
+
+    os_mutex_lock(&shared_heap_list_lock);
+    if (head->attached_count != 0 || body->attached_count != 0) {
+        LOG_WARNING("To create shared heap chain, all shared heap need to be "
+                    "detached first.");
+        os_mutex_unlock(&shared_heap_list_lock);
+        return NULL;
+    }
+    for (cur = shared_heap_list; cur; cur = cur->next) {
```
TianlongLiang, line 270: If the head is already the head of another chain, and create a new chain with this API, the older pointer `WASMSharedHeap * old_chain` will point to this new chain, better disallow this to avoid unwanted behavior

core/iwasm/common/wasm_memory.c
```diff
@@ -177,39 +177,54 @@ wasm_runtime_create_shared_heap(SharedHeapInitArgs *init_args)
         goto fail1;
     }
 
-    if (!(heap->heap_handle =
-              runtime_malloc(mem_allocator_get_heap_struct_size()))) {
-        goto fail2;
-    }
-
     size = align_uint(size, os_getpagesize());
```
TianlongLiang, line 180: move L186 check before use size

core/iwasm/common/wasm_memory.c
```diff
@@ -324,26 +523,38 @@ wasm_runtime_detach_shared_heap_internal(WASMModuleInstanceCommon *module_inst)
...
             (WASMModuleInstanceExtra *)((WASMModuleInstance *)module_inst)->e;
+        if (e->shared_heap != NULL) {
+            os_mutex_lock(&shared_heap_list_lock);
+            e->shared_heap->attached_count--;
+            os_mutex_unlock(&shared_heap_list_lock);
+        }
         e->shared_heap = NULL;
-#if WASM_ENABLE_JIT != 0
 #if UINTPTR_MAX == UINT64_MAX
         e->shared_heap_start_off.u64 = UINT64_MAX;
```
TianlongLiang, line 544: Add comments: to check a corner case:

app_offset >= shared_heap_start && app_offset <= shared_heap_end - bytes + 1 when bytes=1 and both e->shared_heap_start_off and e->shared_heap_end_off is 0xffffffff

core/iwasm/common/wasm_memory.c
```diff
@@ -385,71 +596,96 @@ wasm_runtime_get_shared_heap(WASMModuleInstanceCommon *module_inst_comm)
...
 
-    if (!is_memory64) {
-        if (app_offset >= heap->start_off_mem32
-            && app_offset <= UINT32_MAX - bytes + 1) {
-            return true;
-        }
+    shared_heap_start =
+        (uint64)get_last_used_shared_heap_start_offset(module_inst);
+    shared_heap_end = (uint64)get_last_used_shared_heap_end_offset(module_inst);
+    if (app_offset >= shared_heap_start
```
TianlongLiang: move 619 underflow check here, check it first

core/iwasm/common/wasm_memory.c
```diff
@@ -385,71 +596,96 @@ wasm_runtime_get_shared_heap(WASMModuleInstanceCommon *module_inst_comm)
...
-        if (app_offset >= heap->start_off_mem64
-            && app_offset <= UINT64_MAX - bytes + 1) {
+
+    /* Early stop for app start address not in the shared heap(chain) at all */
+    shared_heap_start =
+        is_memory64 ? heap->start_off_mem64 : heap->start_off_mem32;
+    shared_heap_end = is_memory64 ? UINT64_MAX : UINT32_MAX;
+    if (app_offset < shared_heap_start
+        || app_offset > shared_heap_end - bytes + 1
+        || bytes - 1 > shared_heap_end) {
```
TianlongLiang: same as above

core/iwasm/common/wasm_memory.c
```diff
@@ -385,71 +596,96 @@ wasm_runtime_get_shared_heap(WASMModuleInstanceCommon *module_inst_comm)
...
+        goto fail;
+    }
+
+    /* Find the exact shared heap that app addr is in, and update last used
+     * shared heap info in module inst extra */
+    for (cur = heap; cur; cur = cur->chain_next) {
+        shared_heap_start =
+            is_memory64 ? cur->start_off_mem64 : cur->start_off_mem32;
+        shared_heap_end = shared_heap_start - 1 + cur->size;
+        if (app_offset >= shared_heap_start
```
TianlongLiang: same as above

core/iwasm/common/wasm_memory.c
```diff
@@ -814,12 +1070,12 @@ wasm_runtime_validate_app_str_addr(WASMModuleInstanceCommon *module_inst_comm,
...
-              + (memory_inst->is_memory64
-                     ? (app_str_offset - shared_heap->start_off_mem64)
-                     : (app_str_offset - shared_heap->start_off_mem32));
-        str_end = (char *)shared_heap->base_addr + shared_heap->size;
+        shared_heap_end_off =
+            get_last_used_shared_heap_end_offset(module_inst_comm);
+        shared_heap_base_addr_adj =
+            (char *)get_last_used_shared_heap_base_addr_adj(module_inst_comm);
+        str = shared_heap_base_addr_adj + app_str_offset;
+        str_end = shared_heap_base_addr_adj + shared_heap_end_off;
```
TianlongLiang: should be `shared_heap_end_off+1`

core/iwasm/common/wasm_memory.c
```diff
@@ -974,29 +1222,17 @@ wasm_runtime_addr_native_to_app(WASMModuleInstanceCommon *module_inst_comm,
 
     bounds_checks = is_bounds_checks_enabled(module_inst_comm);
```
TianlongLiang, line 1232: this logic should moved to cmake config, to double check

core/iwasm/common/wasm_memory.c
```diff
@@ -1113,36 +1349,16 @@ wasm_check_app_addr_and_convert(WASMModuleInstance *module_inst, bool is_str,
...
+            (WASMModuleInstanceCommon *)module_inst);
+        shared_heap_end_off = get_last_used_shared_heap_end_offset(
+            (WASMModuleInstanceCommon *)module_inst);
+        native_addr = shared_heap_base_addr_adj + (uintptr_t)app_buf_addr;
 
-        /* The whole string must be in the linear memory */
+        /* The whole string must be in the shared heap */
         str = (const char *)native_addr;
-        str_end = (const char *)shared_heap->base_addr + shared_heap->size;
+        str_end = (const char *)shared_heap_base_addr_adj + shared_heap_end_off;
```
TianlongLiang: should be shared_heap_end_off + 1

TianlongLiang (commented):
Review checkpoint 1/2

samples/shared-heap/src/shared_heap_chain.c
```diff
@@ -0,0 +1,321 @@
...
+    uint8 *buf;
+
+    wasm_runtime_call_wasm(exec_env, func, 2, argv);
+
+    if (wasm_runtime_get_exception(module_inst)) {
+        printf("Failed to call function: %s\n",
+               wasm_runtime_get_exception(module_inst));
+        return false;
+    }
+    if (free_on_fail && argv[0] == 0) {
```
lum1n0us:
```suggestion
    if (argv[0] == 0) {
```

samples/shared-heap/src/shared_heap_chain.c
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    /* allocate memory by calling my_shared_heap_malloc function and send it
+       to wasm app2 */
+    for (i = 0; i < 8; i++) {
+        argv[0] = 1024 * (i + 1);
+        argv[1] = i + 1;
+        if (!produce_data(module_inst, exec_env, queue,
+                          my_shared_heap_malloc_func, argv, 1024 * (i + 1),
+                          true)) {
```
lum1n0us, line 79: Maybe we should write something into the msg payload here to allow L135 to output and even examine it.

samples/shared-heap/wasm-apps/test2.c
```diff
@@ -16,3 +15,14 @@ print_buf(char *buf)
     printf("wasm app2's wasm func received buf: %s\n\n", buf);
     shared_heap_free(buf);
 }
+
+void
+consume_str(char *buf)
```
lum1n0us, line 20: Just a minor opinion: it's better if there is an expected character as input so `consume_str()` can examine it by itself.

tests/unit/shared-heap/shared_heap_test.cc
```diff
@@ -26,50 +28,50 @@ struct ret_env {
...
+load_wasm(char *wasm_file_tested, unsigned int app_heap_size,
+          ret_env &ret_module_env)
 {
-    std::string wasm_mem_page = wasm_file_tested;
-    const char *wasm_file = strdup(wasm_mem_page.c_str());
-    wasm_module_inst_t wasm_module_inst = nullptr;
-    wasm_module_t wasm_module = nullptr;
-    wasm_exec_env_t exec_env = nullptr;
-    unsigned char *wasm_file_buf = nullptr;
+    const char *wasm_file = strdup(wasm_file_tested);
```
lum1n0us: need to free `wasm_file` somewhere

tests/unit/shared-heap/shared_heap_test.cc
```diff
@@ -92,149 +94,956 @@ destroy_module_env(struct ret_env module_env)
...
+
+static void
+create_test_shared_heap(uint8 *preallocated_buf, size_t size,
+                        WASMSharedHeap **shared_heap_res)
+{
+    SharedHeapInitArgs args = { 0 };
+    WASMSharedHeap *shared_heap = nullptr;
+    args.pre_allocated_addr = preallocated_buf;
+    args.size = size;
+    shared_heap = wasm_runtime_create_shared_heap(&args);
```
lum1n0us, line 228: Who is going to release this `shared_heap` at the end of every case(if the case `create_test_shared_heap()`?

tests/unit/shared-heap/shared_heap_test.cc
```diff
@@ -26,50 +28,50 @@ struct ret_env {
     char error_buf[128];
 };
 
-struct ret_env
-load_wasm(char *wasm_file_tested, unsigned int app_heap_size)
+static void
+destroy_module_env(struct ret_env module_env);
+
```
lum1n0us, line 33: Might want to use setup() and teardown() to simplify and unify test procedures.

tests/unit/shared-heap/shared_heap_test.cc
```diff
@@ -92,149 +94,956 @@ destroy_module_env(struct ret_env module_env)
...
+    SharedHeapInitArgs args = { 0 };
+    WASMSharedHeap *shared_heap = nullptr;
+    uint32 argv[1] = { 0 }, BUF_SIZE = os_getpagesize();
+    uint8 preallocated_buf[BUF_SIZE];
+
+    /* create a preallocated shared heap */
+    args.pre_allocated_addr = preallocated_buf;
+    args.size = BUF_SIZE;
+    shared_heap = wasm_runtime_create_shared_heap(&args);
+    if (!shared_heap) {
```
lum1n0us, line 200: Who is going to release this shared_heap at the end of every case(if the case create_test_shared_heap()?

tests/unit/shared-heap/shared_heap_test.cc
```diff
@@ -92,149 +94,956 @@ destroy_module_env(struct ret_env module_env)
...
+    memset(&args, 0, sizeof(args));
+    args.pre_allocated_addr = preallocated_buf2;
+    args.size = size2;
+    shared_heap2 = wasm_runtime_create_shared_heap(&args);
+    if (!shared_heap2) {
+        FAIL() << "Create preallocated shared heap failed.\n";
+    }
+
+    *shared_heap_chain =
+        wasm_runtime_chain_shared_heaps(shared_heap, shared_heap2);
```
lum1n0us, line 262: Who is going to unchain and release this shared_heap_chain at the end of every case(if the case `create_test_shared_heap_chain()`)?

tests/unit/shared-heap/wasm-apps/test_addr_conv.c
```diff
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
  */
 
-#include <stdio.h>
+#define NULL 0
```
lum1n0us: this is wired

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -7883,3 +7883,40 @@ wasm_runtime_is_underlying_binary_freeable(WASMModuleCommon *const module)
...
+        bytes = 1;
+    }
+
+    /* Find the exact shared heap that app addr is in, and update last used
+     * shared heap info in func context */
+    for (cur = heap; cur; cur = cur->chain_next) {
+        shared_heap_start =
+            is_memory64 ? cur->start_off_mem64 : cur->start_off_mem32;
+        shared_heap_end = shared_heap_start - 1 + cur->size;
+        if (app_offset >= shared_heap_start
```
TianlongLiang: check underflow first

core/iwasm/compilation/aot_emit_memory.c
```diff
@@ -111,17 +145,428 @@ ffs(int n)
...
+            aot_set_last_error("llvm build load failed");             \
+            goto fail;                                                \
+        }                                                             \
+    } while (0)
+
+/* Update last used shared heap info(alloc ptr) in function ctx:
+ * 1. shared_heap_start_off 2. shared_heap_end_off 3. shared_heap_base_addr_adj
+ */
+bool
+aot_check_shared_heap_chain(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
```
TianlongLiang: rename to update

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -7883,3 +7883,40 @@ wasm_runtime_is_underlying_binary_freeable(WASMModuleCommon *const module)
 
     return true;
 }
+
+#if WASM_ENABLE_SHARED_HEAP != 0
+bool
+wasm_runtime_update_last_used_shared_heap(WASMModuleInstanceCommon *module_inst,
```
TianlongLiang: Rename it to check_and_update

core/iwasm/compilation/aot_llvm.c
```diff
@@ -1517,73 +1517,153 @@ create_memory_info(const AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
...
+            goto fail;                                                         \
+        }                                                                      \
+        if (!LLVMBuildStore(comp_ctx->builder, load_val, func_ctx->field)) {   \
+            aot_set_last_error("llvm build store failed");                     \
+            goto fail;                                                         \
+        }                                                                      \
+    } while (0)
+
 static bool
 create_shared_heap_info(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx)
```
TianlongLiang, line 1565: In the future, maybe should update func ctx shared heap alloca value back to module instance extra when the function end

core/iwasm/compilation/aot_llvm.c
```diff
@@ -1517,73 +1517,153 @@ create_memory_info(const AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
...
+              comp_ctx->pointer_size == sizeof(uint64) ? I64_NEG_ONE
+                                                       : I32_NEG_ONE,
+              "head_start_off_minus_one"))) {
+        aot_set_last_error("llvm build load failed");
+        goto fail;
     }
 
+    /* if there is attached shared heap(s), the value will be valid start_off-1,
+     * otherwise it will be UINT32_MAX/UINT64_MAX, so during the bounds checks,
+     * when has attached shared heap: offset > start_off - 1 => offset >= offset
```
TianlongLiang: should be offset >= start_off

TianlongLiang (commented):
Review 2/2

tests/unit/shared-heap/shared_heap_test.cc
TianlongLiang: runtime will destory the shared heap list and free the resources

tests/unit/shared-heap/shared_heap_test.cc
TianlongLiang: runtime will destroy the shared heap list and free the resources

tests/unit/shared-heap/shared_heap_test.cc
TianlongLiang: no need to unchain, runtime will destroy the shared heap list and free the resources

lum1n0us (approved):

xujuntwt95329 (approved):

PR #4402, opened 2025-06-23
loader: add type index checking for aot loader

kylo5aby:


TianlongLiang:
Maybe consider adding `WASM_ENABLE_AOT_VALIDATOR` macro to wrap the modified code, IIUC, these checks are only for caution, normal aot compiled from wamrc won't have this problem.

kylo5aby:
> Maybe consider adding `WASM_ENABLE_AOT_VALIDATOR` macro to wrap the modified code, IIUC, these checks are only for caution, normal aot compiled from wamrc won't have this problem.

resolved

lum1n0us (approved):

TianlongLiang (approved):

PR #4403, opened 2025-06-23
wasi_nn_llamacpp.c: make this compilable

yamt:


core/iwasm/libraries/wasi-nn/src/wasi_nn_llamacpp.c
```diff
@@ -2,6 +2,9 @@
  * Copyright (C) 2019 Intel Corporation.  All rights reserved.
  * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
  */
+
+#include <stdlib.h>
```
lum1n0us, line 6: How about `#include "bh_platform.h"` and using os_xxx APIs? Just in case there are cross-platform requirements, although I highly doubt it.

yamt: if you feel strongly, i can make the change.
however, i don't think it improves the cross-platform situation much as it's still using malloc/free via llama.cpp, cjson etc.

PR #4377, opened 2025-06-17
Dev/zephyr file socket

srberard:
This is still a work in progress, but committing now to make the code available for others who might want to review, test, or build on it while it's still being developed.

Fixes include the following (Zephyr target):
* Improved Socket support
* Fixes to stdout/console output
* Fixes for fstat, fstatat, and unlink
* Code changes to fix numerous build warnings.

This branch still needs to be updated to the latest trunk and validation on non-Zephyr platforms.

PR #4406, opened 2025-06-24
Update zephyr file socket branch

lum1n0us:
This is based on https://github.com/bytecodealliance/wasm-micro-runtime/pull/4377. 

@srberard Please help me confirm this update. This PR is supposed to contain all and only your latest modifications. If it is as expected, I will merge this PR while ignoring the CI results. I hope we can use new PRs to fix those CI failures.

### DIFF
--- a/core/iwasm/common/wasm_application.c
+++ b/core/iwasm/common/wasm_application.c
@@ -289,6 +289,7 @@ wasm_application_execute_main(WASMModuleInstanceCommon *module_inst, int32 argc,
     exec_env = wasm_runtime_get_exec_env_singleton(module_inst);
     if (exec_env) {
         wasm_runtime_dump_mem_consumption(exec_env);
+        (WASMModuleInstance *)module_inst->cur_exception 
     }
 #endif
 
@@ -712,7 +713,10 @@ execute_func(WASMModuleInstanceCommon *module_inst, const char *name,
             }
             case VALUE_TYPE_F32:
             {
-                os_printf("%.7g:f32", *(float32 *)(argv1 + k));
+                // Explicit cast to double to avoid warning.
+                // Float arguments are promoted to double in variadic
+                // functions per section 6.5.2.2 of the C99 standard.
+                os_printf("%.7g:f32", (double) *(float32 *)(argv1 + k));  
                 k++;
                 break;
             }
--- a/core/iwasm/common/wasm_application.c
+++ b/core/iwasm/common/wasm_application.c
@@ -289,7 +289,7 @@ wasm_application_execute_main(WASMModuleInstanceCommon *module_inst, int32 argc,
     exec_env = wasm_runtime_get_exec_env_singleton(module_inst);
     if (exec_env) {
         wasm_runtime_dump_mem_consumption(exec_env);
-        (WASMModuleInstance *)module_inst->cur_exception 
+        (WASMModuleInstance *)module_inst->cur_exception
     }
 #endif
 
@@ -716,7 +716,7 @@ execute_func(WASMModuleInstanceCommon *module_inst, const char *name,
                 // Explicit cast to double to avoid warning.
                 // Float arguments are promoted to double in variadic
                 // functions per section 6.5.2.2 of the C99 standard.
-                os_printf("%.7g:f32", (double) *(float32 *)(argv1 + k));  
+                os_printf("%.7g:f32", (double)*(float32 *)(argv1 + k));
                 k++;
                 break;
             }
--- a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
+++ b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
@@ -1904,10 +1904,10 @@ convert_timestamp(__wasi_timestamp_t in, os_timespec *out)
 #else
     out->tv_nsec = (long)(in % 1000000000);
 #endif
-    in /= 1000000000;
+    __wasi_timestamp_t temp = in / 1000000000;
 
     // Clamp to the maximum in case it would overflow our system's time_t.
-    out->tv_sec = (time_t)in < BH_TIME_T_MAX ? (time_t)in : BH_TIME_T_MAX;
+    out->tv_sec = (time_t)temp < BH_TIME_T_MAX ? (time_t)temp : BH_TIME_T_MAX;
 }
 
 __wasi_errno_t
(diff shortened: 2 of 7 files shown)


core/iwasm/common/wasm_application.c
```diff
@@ -289,6 +289,7 @@ wasm_application_execute_main(WASMModuleInstanceCommon *module_inst, int32 argc,
     exec_env = wasm_runtime_get_exec_env_singleton(module_inst);
     if (exec_env) {
         wasm_runtime_dump_mem_consumption(exec_env);
+        (WASMModuleInstance *)module_inst->cur_exception
```
lum1n0us: not finished.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -2212,7 +2212,7 @@ wasmtime_ssp_poll_oneoff(wasm_exec_env_t exec_env, struct fd_table *curfds,
                 if (error == 0) {
                     // Proper file descriptor on which we can poll().
                     pfds[i] = (os_poll_file_handle){
-                        .fd = fos[i]->file_handle,
+                        .fd = fos[i]->file_handle->fd,
```
lum1n0us, line 2215: `file_handle`'s type is `os_file_handle` which is `typedef int os_file_handle`. So a compilation error will be raised: ` error: invalid type argument of ‘->’ (have ‘os_file_handle’ {aka ‘int’})`.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -2845,23 +2845,37 @@ wasmtime_ssp_sock_recv_from(wasm_exec_env_t exec_env, struct fd_table *curfds,
...
     }
 
+    // If the source address is not NULL, the caller is requesting the source
+    // address to be returned if the protocol supports it.  As such, we convert
+    // the format of the structure pass in prior to the call to the OS
+    // implementation.  If the value is NULL, the POSIX standard states that
+    // the address is not returned.
+    if (src_addr != NULL) {
+        sockaddr_ptr = &sockaddr;
+        wasi_addr_to_bh_sockaddr(src_addr, &sockaddr);
```
lum1n0us, line 2863: There is no need to translate the content of `src_addr`

[`src_addr` will be filled upon return. Before the call, its content is unpredictable. ](https://github.com/bytecodealliance/wasm-micro-runtime/pull/4391)

core/shared/platform/zephyr/zephyr_file.c
```diff
@@ -67,20 +68,63 @@ K_MUTEX_DEFINE(desc_array_mutex);
...
+    if (s == NULL) {
+        return NULL;
+    }
+
+    size_t len = strlen(s) + 1; // Include null terminator
+    char *dup = malloc(len);
+    if (dup == NULL) {
+        return NULL; // Allocation failed
+    }
+
```
lum1n0us, line 90: May need `memset` to zero out the buffer.

core/shared/platform/zephyr/zephyr_file.c
```diff
@@ -715,13 +802,18 @@ os_renameat(os_file_handle old_handle, const char *old_path,
 
     GET_FILE_SYSTEM_DESCRIPTOR(old_handle->fd, ptr);
 
-    char *path = strdup(new_path);
+    char *path = duplicate_string(new_path);
```
lum1n0us, line 806: need to free `path` somewhere

core/shared/platform/zephyr/zephyr_socket.c
```diff
@@ -1031,14 +1044,15 @@ os_socket_get_broadcast(bh_socket_t socket, bool *is_enabled)
...
 
-    va_start(args, request);
-    ret = zsock_ioctl(handle->fd, request, args);
-    va_end(args);
+    // va_start(args, request);
+    // ret = zsock_ioctl(handle->fd, request, args);
+    // va_end(args);
 
-    return ret;
+    // return ret;
```
lum1n0us, line 1055: need to remove commented code

lum1n0us (commented):
I just listed all my doubts. We can have a discussion, but please fix them in a new PR if you want to. I will merge this PR to the dev branch once its commits are confirmed.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
srberard: In #3633, @lucasAbadFr changed for Zephyr to be the following:
```
typedef struct zephyr_handle *os_file_handle;
```

I'll see what I can do to unify this better to align with the POSIX/Linux implementation.

core/iwasm/common/wasm_application.c
srberard: Hi @lum1n0us, I don't understand what you mean by the above.  I don't believe this file was updated as part of this Zephyr work.  Can you explain more what you would like to see changed here?

core/shared/platform/zephyr/zephyr_file.c
srberard: It shouldn't be needed here.  We allocate the string length plus 1 for the terminator.  If the allocation fails, the malloc failed and the buffer was not allocated; NULL is returned to indicate error.  If the malloc succeeded, memory is allocated and a memcpy is performed for the full length of the buffer (which is strlen + 1).  This will copy the full string plus the NULL terminator.  A memset prior would not cause an issue but would work done unnecessarily (perf impact).

In the code above, I'm using `malloc` when I should probably use `os_malloc` instead.  Existing code uses the `BH_MALLOC` macro (which calls `os_malloc`).  I've update the code to use this instead.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
srberard: Removed.

core/shared/platform/zephyr/zephyr_file.c
dkouba-atym: I just tested this change to BH_MALLOC and it results in a hard-fault on my device. I have not tracked down the root cause, however it appears to be when the duplicated string is accessed.

dkouba-atym: Apologies, the above can be ignored. My issue turned out to be something else memory related in my setup.

core/iwasm/common/wasm_application.c
lum1n0us: I used to think it was a typo and that we should just remove it.

PR #4409, opened 2025-06-24
Fix few shadow warnings

lum1n0us:
- declaration of ‘memidx’ shadows a previous local
- declaration of ‘count’ shadows a previous local

### DIFF
--- a/core/iwasm/interpreter/wasm_interp_classic.c
+++ b/core/iwasm/interpreter/wasm_interp_classic.c
@@ -4088,7 +4088,7 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                     case WASM_OP_STRING_ENCODE_LOSSY_UTF8_ARRAY:
                     case WASM_OP_STRING_ENCODE_WTF8_ARRAY:
                     {
-                        uint32 start, array_len, count;
+                        uint32 start, array_len;
                         int32 bytes_written;
                         EncodingFlag flag = WTF8;
                         WASMArrayType *array_type;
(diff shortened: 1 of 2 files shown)


core/iwasm/interpreter/wasm_interp_classic.c
```diff
@@ -4088,7 +4088,7 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
                     case WASM_OP_STRING_ENCODE_LOSSY_UTF8_ARRAY:
                     case WASM_OP_STRING_ENCODE_WTF8_ARRAY:
                     {
-                        uint32 start, array_len, count;
+                        uint32 start, array_len;
```
yamt, line 4091: the other way around (ie. keep the variables with smaller scope) might be better
(at least easier to read for human IMO)

what do you think?

lum1n0us: TRUE in the case of `count` in *wasm_interp_classic.c*. However, handling `memidx` in *wasm_loader.c* seems to be another problem. `memidx` is a widely used variable in various opcode handlers. If we follow the rule of keeping variables with smaller scope, we might have to repeatedly declare it many times. It seems there is no proper rule for a large switch-case structure.

yamt: ok. although i prefer smaller scope, there are pros and cons and i don't have any data to back up my claim.

yamt (approved):

TianlongLiang (approved):

PR #4411, opened 2025-06-25
wasi-nn: make the host use the wasi_ephemeral_nn version of tensor_data

yamt:
the motivations:

* make the actual input size available to the backends. (currently the backends have to make a guess from shape/type.)

* make the host logic look a bit similar to wasi_ephemeral_nn.

this is a backend api/abi change.

### DIFF
--- a/core/iwasm/libraries/wasi-nn/src/wasi_nn_openvino.c
+++ b/core/iwasm/libraries/wasi-nn/src/wasi_nn_openvino.c
@@ -402,7 +402,7 @@ set_input(void *ctx, graph_execution_context exec_ctx, uint32_t index,
                       shape_info);
 
         CHECK_OV_STATUS(ov_tensor_create_from_host_ptr(input_type, input_shape,
-                                                       wasi_nn_tensor->data,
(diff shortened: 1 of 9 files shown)


TianlongLiang (approved):

core/iwasm/libraries/wasi-nn/src/wasi_nn_openvino.c
```diff
@@ -460,14 +460,14 @@ get_output(void *ctx, graph_execution_context exec_ctx, uint32_t index,
 
     CHECK_OV_STATUS(ov_tensor_get_byte_size(ov_tensor, &byte_size), ret);
 
-    if (byte_size > *output_tensor_size) {
+    if (byte_size > output_tensor->size) {
```
lum1n0us, line 463: May I ask about the difference between output_tensor->size and output_tensor_size?

yamt: this PR makes output_tensor_size output-only.

PR #4413, opened 2025-06-25
Collective fix

TianlongLiang:
- Fix typos
- Fix bugs in unit test and bh_vector.c

### DIFF
--- a/core/shared/utils/bh_vector.c
+++ b/core/shared/utils/bh_vector.c
@@ -35,8 +35,8 @@ extend_vector(Vector *vector, size_t length)
     if (length <= vector->max_elems)
         return true;
 
-    if (length < vector->size_elem * 3 / 2)
-        length = vector->size_elem * 3 / 2;
+    if (length < vector->max_elems * 3 / 2)
+        length = vector->max_elems * 3 / 2;
 
     if (!(data = alloc_vector_data(length, vector->size_elem))) {
         return false;
@@ -194,12 +194,12 @@ bh_vector_append(Vector *vector, const void *elem_buf)
         goto just_return;
     }
 
-    /* make sure one more slot is used by the thread who allocas it */
+    /* make sure one more slot is used by the thread who allocates it */
     if (vector->lock)
         os_mutex_lock(vector->lock);
 
     if (!extend_vector(vector, vector->num_elems + 1)) {
-        LOG_ERROR("Append ector elem failed: extend vector failed.\n");
+        LOG_ERROR("Append vector elem failed: extend vector failed.\n");
         goto unlock_return;
     }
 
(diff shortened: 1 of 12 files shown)


core/shared/utils/bh_vector.c
```diff
@@ -35,8 +35,8 @@ extend_vector(Vector *vector, size_t length)
     if (length <= vector->max_elems)
         return true;
 
-    if (length < vector->size_elem * 3 / 2)
-        length = vector->size_elem * 3 / 2;
+    if (length < vector->max_elems * 3 / 2)
+        length = vector->max_elems * 3 / 2;
```
yamt: IMO, this warrants a separate PR.

TianlongLiang: fair point, I was actually pretty surprised to see this and check this several times

tests/unit/memory64/memory64_atomic_test.cc
```diff
@@ -56,6 +56,8 @@ class memory64_atomic_test_suite : public testing::TestWithParam<RunningMode>
         if (exec_env)
             wasm_runtime_destroy_exec_env(exec_env);
         if (module_inst)
+            wasm_runtime_deinstantiate(module_inst);
+        if (module)
             wasm_runtime_unload(module);
```
yamt, line 61: i'm not sure if it's appropriate to unload module here.
isn't it simpler to make all clean up logic to TearDown?

tests/fuzz/wasm-mutator-fuzz/server/app/main.py
```diff
@@ -204,7 +204,7 @@ def New_fuzzing():
         # curd.set_error_status_to(list(map(lambda x: x.id, error_list)), db)
         # Fuzzing.query.filter_by(id=fuzz.id).delete()
         fuzz.data = {'error': "Clone repo Error"}
-        db.commit()
+        db.session.commit()
```
yamt, line 207: is this a functional change?

core/shared/utils/bh_vector.c
yamt: heh. i reviewed it several times too.

the bug is not critical as it might seems because usually it merely causes slight over-allocation and frequent realloc. ("slight" because size_elem is usually small.)

tests/fuzz/wasm-mutator-fuzz/server/app/main.py
TianlongLiang: This line should be a typo, the previous lines(such as L194) all use `db.session.commit()`

tests/unit/memory64/memory64_atomic_test.cc
TianlongLiang: Yes, this resource management in some unit test suites is far from optimal. But I think I will address them in another PR focusing on that issue.

yamt: ok

tests/fuzz/wasm-mutator-fuzz/server/app/main.py
yamt: ok

lum1n0us (approved):

PR #4417, opened 2025-06-26
Improve run.py of regression

Septa2112:
1. Fix script interruption on case failure
2. improve statistics logic
3. enable select specific issue ids

### DIFF
--- a/tests/regression/ba-issues/run.py
+++ b/tests/regression/ba-issues/run.py
@@ -254,17 +254,26 @@ def process_and_run_test_cases(
     total = len(passed_ids) + len(failed_ids)
     passed = len(passed_ids)
     failed = len(failed_ids)
-    issue_ids_should_test = (
-        issue_ids_should_test if issue_ids_should_test else "no more"
+
+    format_issue_ids_should_test = (
+        " ".join("#" + str(x) for x in issue_ids_should_test)
+        if issue_ids_should_test
+        else "no more"
     )
-    json_only_ids = json_only_ids if json_only_ids else "no more"
+    format_json_only_ids = (
+        " ".join("#" + str(x) for x in json_only_ids) if json_only_ids else "no more"
+    )
+
     print(f"####################################")
     print(f"==== Test results ====")
     print(f"  Total:  {total}")
     print(f"  Passed: {passed}")
     print(f"  Failed: {failed}")
-    print(f"  Left issues in folder: {issue_ids_should_test}")
-    print(f"  Cases in JSON but not found in folder: {json_only_ids}")
+    if not selected_ids:
+        print(f"  Left issues in folder: {format_issue_ids_should_test}")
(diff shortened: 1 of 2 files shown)


TianlongLiang (approved):

tests/regression/ba-issues/run.py
```diff
@@ -214,33 +217,79 @@ def process_and_run_test_cases(data: Dict[str, Dict]):
...
+        "-i",
+        "--issues",
+        type=str,
+        help="Comma separated list of issue ids to run, e.g. 1,2,3. Default: all.",
+    )
+    args = parser.parse_args()
+
+    selected_ids = None
+    if args.issues:
+        selected_ids = [int(x) for x in args.issues.split(",") if x.strip().isdigit()]
```
lum1n0us, line 291: If I recall correctly, `--type=list` can do the trick. But might still need to convert `str` to `int`.

tests/regression/ba-issues/run.py
```diff
@@ -214,33 +217,79 @@ def process_and_run_test_cases(data: Dict[str, Dict]):
...
+                    failed_ids.add(issue_id)
 
     total = len(passed_ids) + len(failed_ids)
     passed = len(passed_ids)
     failed = len(failed_ids)
-    issue_ids_should_test = (
-        issue_ids_should_test if issue_ids_should_test else "no more"
+
+    format_issue_ids_should_test = (
+        " ".join("#" + str(x) for x in issue_ids_should_test)
```
lum1n0us: Just a hint, usually, using f-strings has better performance than string concatenation.

tests/regression/ba-issues/run.py
```diff
@@ -214,33 +217,79 @@ def process_and_run_test_cases(data: Dict[str, Dict]):
...
+        if issue_ids_should_test
+        else "no more"
+    )
+    format_json_only_ids = (
+        " ".join("#" + str(x) for x in json_only_ids) if json_only_ids else "no more"
```
lum1n0us: Just a hint, usually, using f-strings has better performance than string concatenation.

tests/regression/ba-issues/run.py
```diff
@@ -214,33 +217,79 @@ def process_and_run_test_cases(data: Dict[str, Dict]):
...
     )
+
+    print(f"####################################")
     print(f"==== Test results ====")
-    print(f"  Total: {total}")
+    print(f"  Total:  {total}")
```
lum1n0us: you might want to add one more space before `Total:` and let all three lines aligned at `:` and remove the space before `{total}`

tests/regression/ba-issues/run.py
Septa2112: For example, `type=list` will convert `-i 2833` to `['2','8','3','3']`. So this seems like an acceptable solution for now.

lum1n0us (approved):

PR #4418, opened 2025-06-26
Avoid executable stack by marking that it's not required.

jkroonza:
Also refer to: https://github.com/fluent/fluent-bit/issues/10513


jkroonza:
Other assembly files may be affected too, but this is the one that causes problems for me.

yamt:
> Other assembly files may be affected too, but this is the one that causes problems for me.

can you consider to apply this to other *.s files?
at least `invokeNative_em64_simd.s`, which is used by the default configuration for the target.

jkroonza:
I can definitely add to all, but I'm not in a position to test all, is that acceptable?

yamt:
> I can definitely add to all, but I'm not in a position to test all, is that acceptable?

i think it's fine.

jkroonza:
Is there a way to cherry pick this over to 1.3 from here or do I push a separate PR again?

core/iwasm/common/arch/invokeNative_osx_universal.s
```diff
@@ -15,4 +15,6 @@
 #else
 #include "invokeNative_em64_simd.s"
 #endif
-#endif
\ No newline at end of file
+#endif#if defined(__linux__) && defined(__ELF__)
```
yamt: this seems broken.

also, this file is for osx.
i don't think you need to apply the change to this file.

jkroonza: I missed that in my final eyeball check.

Sorry about that.  I did remove the extra lines now, but my editor insists on adding newlines, and I suspect it's good to have the newline there, so if it's acceptable, let it sneak in, otherwise I'll get that out (will be a bit more work but happy to action).

core/iwasm/common/arch/invokeNative_mingw_x64.s
```diff
@@ -55,3 +55,6 @@ cycle_end:
     call *%r10
     leave
     ret
+#if defined(__linux__) && defined(__ELF__)
+.section .note.GNU-stack,"",%progbits
+#endif
```
yamt: i suspect these mingw files are for windows.

yamt (approved):

core/iwasm/common/arch/invokeNative_mingw_x64.s
jkroonza: reverted the mingw files.  It shouldn't have an effect in that case of compiling for windows ... but still, it could cause confusion.

core/iwasm/common/arch/fneh.txt
```diff
@@ -0,0 +1,3 @@
+#if defined(__linux__) && defined(__ELF__)
+.section .note.GNU-stack,"",%progbits
+#endif
```
yamt, line 3: i suspect this file was added unintentionally.

PR #4425, opened 2025-06-27
CI: add wamr_wasi_extensions to the release assets

yamt:
you can find an example of the release asset at:
https://github.com/yamt/wasm-micro-runtime/releases/download/WAMR-2.3.999/wamr-wasi-extensions-2.3.999.zip

note: this is a static library for wasm32-wasi. no need to provide per host OS (macOS, ubuntu, etc) binaries.

### DIFF
--- a/.github/workflows/build_wamr_wasi_extensions.yml
+++ b/.github/workflows/build_wamr_wasi_extensions.yml
@@ -1,0 +1,57 @@
+# Copyright (C) 2019 Intel Corporation.  All rights reserved.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+name: build wamr_wasi_extensions release
+
+on:
+  workflow_call:
+    inputs:
+      upload_url:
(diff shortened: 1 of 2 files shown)


TianlongLiang:
I think on Windows, a static library would be `.lib` format(at least for MSVC)

yamt:
> I think on Windows, a static library would be `.lib` format(at least for MSVC)

this library is for wasm32-wasi target, not windows native.

TianlongLiang:
OK, got it, thanks

TianlongLiang (approved):

yamt:
> you can find an example of the release asset at: https://github.com/yamt/wasm-micro-runtime/releases/download/WAMR-2.3.999/wamr-wasi-extensions-2.3.999.zip

you can find an example to consume this binary at https://github.com/yamt/wasm-micro-runtime/commit/a3fd5ba5204449e1c115d1627ced99bd71180bff

PR #4429, opened 2025-06-27
Fix potential integer overflow issues

lum1n0us:
It is reported as "Multiplication result converted to larger type". And "Multiplication result may overflow 'Type A' before it is converted to 'Type B'." Type A is a larger type than Type B.

Since the conversion applies after the multiplication, arithmetic overflow may still occur.

> The rule flags every multiplication of two non-constant integer expressions
> that is (explicitly or implicitly) converted to a larger integer type. The
> conversion is an indication that the expression would produce a result that
> would be too large to fit in the smaller integer type.

### DIFF
--- a/core/iwasm/compilation/aot_emit_aot_file.c
+++ b/core/iwasm/compilation/aot_emit_aot_file.c
@@ -302,8 +302,8 @@ get_init_expr_size(const AOTCompContext *comp_ctx, const AOTCompData *comp_data,
 
             /* array_elem_type + type_index + len + elems */
             size += sizeof(uint32) * 3
-                    + wasm_value_type_size_internal(array_type->elem_type,
-                                                    comp_ctx->pointer_size)
+                    + (uint64)wasm_value_type_size_internal(
+                          array_type->elem_type, comp_ctx->pointer_size)
                           * value_count;
             break;
         }
--- a/core/iwasm/compilation/aot_emit_function.c
+++ b/core/iwasm/compilation/aot_emit_function.c
@@ -347,7 +347,8 @@ call_aot_invoke_c_api_native(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
 
     /* Get &c_api_func_imports[func_idx], note size of CApiFuncImport
        is pointer_size * 3 */
-    offset = I32_CONST((comp_ctx->pointer_size * 3) * import_func_idx);
+    offset = I32_CONST((unsigned long long)comp_ctx->pointer_size * 3
+                       * import_func_idx);
     CHECK_LLVM_CONST(offset);
     c_api_func_import =
         LLVMBuildInBoundsGEP2(comp_ctx->builder, INT8_TYPE, c_api_func_imports,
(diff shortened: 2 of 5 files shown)


TianlongLiang (approved):

core/iwasm/compilation/aot_emit_aot_file.c
```diff
@@ -302,8 +302,8 @@ get_init_expr_size(const AOTCompContext *comp_ctx, const AOTCompData *comp_data,
 
             /* array_elem_type + type_index + len + elems */
             size += sizeof(uint32) * 3
-                    + wasm_value_type_size_internal(array_type->elem_type,
-                                                    comp_ctx->pointer_size)
+                    + (uint64)wasm_value_type_size_internal(
+                          array_type->elem_type, comp_ctx->pointer_size)
```
yamt, line 306: does this make differences when this function (get_init_expr_size) returns uint32?

yamt:
> It is reported as "Multiplication result converted to larger type". And "Multiplication result may overflow 'Type A' before it is converted to 'Type B'." Type A is a larger type than Type B.

reported by what?

core/iwasm/compilation/aot_emit_function.c
```diff
@@ -347,7 +347,8 @@ call_aot_invoke_c_api_native(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
 
     /* Get &c_api_func_imports[func_idx], note size of CApiFuncImport
        is pointer_size * 3 */
-    offset = I32_CONST((comp_ctx->pointer_size * 3) * import_func_idx);
+    offset = I32_CONST((unsigned long long)comp_ctx->pointer_size * 3
+                       * import_func_idx);
```
yamt, line 351: is it safe to pass a value potentially larger than i32 max to I32_CONST?

lum1n0us:
> reported by what?

https://codeql.github.com/codeql-query-help/cpp/cpp-integer-multiplication-cast-to-long/

core/iwasm/compilation/aot_emit_aot_file.c
lum1n0us: `get_init_expr_size()` returns a type of `uint32` now. So, there will be no difference.

```c
static uint32
get_init_expr_size(const AOTCompContext *comp_ctx, const AOTCompData *comp_data,
                   InitializerExpression *expr)
```

core/iwasm/compilation/aot_emit_function.c
lum1n0us: Based on the declaration of `LLVMConstInt()`, `N` is an unsigned long long, and usually, a value larger than `INT32_MAX` or `UINT32_MAX` won't cause trouble.

```cpp
LLVMValueRef LLVMConstInt(LLVMTypeRef IntTy, unsigned long long N,
                          LLVMBool SignExtend) {
```

yamt: glacing at llvm logic, it seems ok as LLVMConstInt seems to end up with APInt with implicitTrunc=true.

core/iwasm/compilation/aot_emit_aot_file.c
yamt: so, this change is only to appease codeql?

yamt (approved):

core/iwasm/compilation/aot_emit_aot_file.c
lum1n0us: YES.

PR #4434, opened 2025-06-30
wasi_nn_openvino.c: avoid self-assignment warning

yamt:


lum1n0us (approved):

lum1n0us:
BTW, do you think we need to add the self-assignment warning flag globally and enable it by default?

yamt:
> BTW, do you think we need to add the self-assignment warning flag globally and enable it by default?

-Wself-assign is enabled by -Wall here.
isn't it the case for you?

```
Apple clang version 17.0.0 (clang-1700.0.13.5)
Target: x86_64-apple-darwin24.5.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

PR #4440, opened 2025-07-01
Add validation for recursive type count in  loader

kylo5aby:


core/iwasm/interpreter/wasm_loader.c
```diff
@@ -2033,6 +2033,12 @@ load_type_section(const uint8 *buf, const uint8 *buf_end, WASMModule *module,
             if (flag == DEFINED_TYPE_REC) {
                 read_leb_uint32(p, p_end, rec_count);
 
+                if (rec_count == 0) {
+                    set_error_buf(
+                        error_buf, error_buf_size,
+                        "recursive type count should be greater than 0");
```
lum1n0us: Not sure about that `rec_count` can't be 0. Would you mind pointing out the relevant content in the GC spec? https://webassembly.github.io/gc/core/valid/modules.html

kylo5aby: > Not sure about that `rec_count` can't be 0. Would you mind pointing out the relevant content in the GC spec? https://webassembly.github.io/gc/core/valid/modules.html

I think gc spec doesn't describe it explicitly as mentioned in [spec](https://github.com/WebAssembly/gc/blob/main/proposals/gc/MVP.md#:~:text=a%20group%20of%20recursive,0%2C%20...%2C%20(rec%20%3Csubtype%3E*).N). But in this [section](https://github.com/WebAssembly/gc/blob/main/proposals/gc/MVP.md#:~:text=a%20rec%20definition,of%20individual%20types.), it says that 

> a rec definition defines a group of mutually recursive types that can refer to each other; it thereby defines several type indices at a time
a single type definition, as in Wasm before this proposal, is reinterpreted as a short-hand for a recursive group containing just one type

that means for a rec, it must has at least 2 types, otherwise, it should be a single type.

for the related implementation in [binaryen](https://github.com/WebAssembly/binaryen/blob/ec1d0b68b317538cc43ab7c769704dfbebcba2c3/src/wasm/wasm-binary.cpp#L253C1-L258C6), only emit `rec` flag when the number of types greater than 1.

lum1n0us (approved):

lum1n0us (approved):

Issue #4364, opened 2025-06-13
wasi configurations in wasm_module_t


yamt:
`wasm_runtime_set_wasi_args` / `wasm_runtime_set_wasi_args_ex` sets wasi configurations on `wasm_module_t`.
it isn't intuitive because those configuartions logically belong to `wasm_module_inst_t`.
ideally it would be nice to allow to instantiate multiple `wasm_module_inst_t` from a single `wasm_module_t`.

one way to fix it is to make them a part of `InstantiationArgs`.
it's unfortunate `InstantiationArgs` is defined in a way which can't be extended without breaking the ABI.

(i thought i filed this issue long time ago. but i couldn't find it. sorry if it's duplicated.)

lum1n0us:
You are right. The current design doesn't accommodate usage scenarios like varying WASI contexts for multiple instances (either continuously in one thread or parallelly across multiple threads) of a module.

I prefer to add a new API like `wasm_runtime_instantiate_with_wasi(wasm_module_t, InstantiationArgs*, WasiConfigArgs*, char*, uint32_t)` and a new set of helpers for building `WasiConfigArgs`.

lum1n0us:
#4256

yamt:
> [#4256](https://github.com/bytecodealliance/wasm-micro-runtime/issues/4256)

oh, was it this May? i thought it was at least a few years ago. it seems my memory is severely broken.

yamt:
> You are right. The current design doesn't accommodate usage scenarios like varying WASI contexts for multiple instances (either continuously in one thread or parallelly across multiple threads) of a module.
> 
> I prefer to add a new API like `wasm_runtime_instantiate_with_wasi(wasm_module_t, InstantiationArgs*, WasiConfigArgs*, char*, uint32_t)` and a new set of helpers for building `WasiConfigArgs`.

how about:
```c
struct InstantiationArgs2;

bool
wasm_runtime_instantiation_args_create(struct InstantiationArgs2 **p);
void
wasm_runtime_instantiation_args_destroy(struct InstantiationArgs2 *p);

void
wasm_runtime_instantiation_args_set_default_stack_size(
    struct InstantiationArgs2 *p, uint32 v);

WASM_RUNTIME_API_EXTERN wasm_module_inst_t
wasm_runtime_instantiate_ex2(const wasm_module_t module,
                             const struct InstantiationArgs2 *args,
                             char *error_buf, uint32_t error_buf_size);
```

[...]

Issue #4331, opened 2025-06-05
wasi-nn: load_by_name seems insecure


yamt:
our load_by_name implementation seems to trust host filename given by wasm modules blindly.
i consider it a security issue.

other runtimes (wasmtime, wasmedge) seem to have a concept of model registry, preloaded models, etc.
to me, it seems like a more reasonable design we should follow.

lum1n0us:
> our load_by_name implementation seems to trust host filename given by wasm modules blindly.

Yes, there is no similar consideration in the current WAMR wasi-nn implementation. I think it is useful, and it would be great if the community could contribute this feature.

yamt:
IMO, it's a bug, not feature.

yamt:
i guess this should be a per-instance configuration like wasi mapdir.
a dictionary of name -> loaded model.

yamt:
wasmedge seems to have --nn-preload cli option, which takes Name:Encode:Target:Path:Path:Path:...
eg. --nn-preload default:GGML:AUTO:llama-2-7b-chat-q5_k_m.gguf

wasmtime seems to have `--wasi=nn-graph=<encoding>::<path>`

yamt:
i'm thinking embedder apis like the following:
```c
struct wasi_nn_graph_registry;

wasi_nn_error
wasi_nn_graph_registry_create(struct wasi_nn_graph_registry **registryp);

void
wasi_nn_graph_registry_destroy(struct wasi_nn_graph_registry *registry);

// the implementation might eagerly preload the given graph, or might defer
// it until wasi-nn load-by-name is actually called for the name.
// the users of this api should not rely on either behaviors.
wasi_nn_error
wasi_nn_graph_registry_register_graph(struct wasi_nn_graph_registry *registry,
                                      const char *name, const char *builders[],
                                      uint32 nbuilders, const char *ecnoding,
                                      const char *target);

// this might fail if the graph is currently in-use. or this might just mark
// the graph "unregistered" so that it will be ignored for later operations.
// the users of this api should not rely on either behaviors.
wasi_nn_error
wasi_nn_graph_registry_unregister_graph(struct wasi_nn_graph_registry *registry,
                                        const char *name);

// TODO: somehow make wasm_runtime_instantiate_with_wasi associate
// wasi_nn_graph_registry to the instance.
// cf. https://github.com/bytecodealliance/wasm-micro-runtime/issues/4364
```

PR #4444, opened 2025-07-02
introduce wasm_runtime_instantiate_ex2

yamt:
at this point, just a slightly inefficiant functionality
equivalent of wasm_runtime_instantiate_ex.
however, unlike wasm_runtime_instantiate_ex, this one is designed
to be extendable without breaking the user-visible ABI.
because the definition of InstantiationArgs2 is not exposed to
users, we can safely add new members to it.

this commit also makes wasm_runtime_instantiate_ex a wrapper
of wasm_runtime_instantiate_ex2.

if this goes well, maybe it's a good idea to apply a similar
pattern to RuntimeInitArgs, LoadArgs, SharedHeapInitArgs, etc.
i started with InstantiationArgs just because i happen to have
a need to extend it for wasi-nn.

cf.
https://github.com/bytecodealliance/wasm-micro-runtime/issues/4364
https://github.com/bytecodealliance/wasm-micro-runtime/issues/4331

### DIFF
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -1654,14 +1654,70 @@ wasm_runtime_instantiate(WASMModuleCommon *module, uint32 stack_size,
                                              error_buf_size);
 }
 
+static void
+instantiation_args_set_defaults(struct InstantiationArgs2 *args)
+{
+    memset(args, 0, sizeof(*args));
+}
+
 WASMModuleInstanceCommon *
 wasm_runtime_instantiate_ex(WASMModuleCommon *module,
                             const InstantiationArgs *args, char *error_buf,
                             uint32 error_buf_size)
 {
+    struct InstantiationArgs2 v2;
+    instantiation_args_set_defaults(&v2);
+    v2.v1 = *args;
+    return wasm_runtime_instantiate_ex2(module, &v2, error_buf, error_buf_size);
+}
+
+bool
+wasm_runtime_instantiation_args_create(struct InstantiationArgs2 **p)
+{
+    struct InstantiationArgs2 *args = wasm_runtime_malloc(sizeof(*args));
+    if (args == NULL) {
+        return false;
+    }
+    instantiation_args_set_defaults(args);
+    *p = args;
+    return true;
+}
+
+void
+wasm_runtime_instantiation_args_destroy(struct InstantiationArgs2 *p)
+{
+    wasm_runtime_free(p);
+}
+
+void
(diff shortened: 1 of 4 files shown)


core/iwasm/include/wasm_export.h
```diff
@@ -732,6 +734,46 @@ wasm_runtime_instantiate_ex(const wasm_module_t module,
...
+
+/**
+ * Setter functions for the InstantiationArgs2 object.
+ */
+void
+wasm_runtime_instantiation_args_set_default_stack_size(
+    struct InstantiationArgs2 *p, uint32_t v);
+
+void
+wasm_runtime_instantiation_args_set_host_managed_heap_size(
```
yamt, line 759: i suspect this is the longest name in our api. :-)

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1654,14 +1654,70 @@ wasm_runtime_instantiate(WASMModuleCommon *module, uint32 stack_size,
...
+{
+    struct InstantiationArgs2 v2;
+    instantiation_args_set_defaults(&v2);
+    v2.v1 = *args;
+    return wasm_runtime_instantiate_ex2(module, &v2, error_buf, error_buf_size);
+}
+
+bool
+wasm_runtime_instantiation_args_create(struct InstantiationArgs2 **p)
+{
```
lum1n0us, line 1676: I believe it would be better to add some input parameter checking for those exported APIs.

lum1n0us (commented):
IIUC, there should be some APIs related to WASI. OR are they included in another PR?

yamt:
> IIUC, there should be some APIs related to WASI. OR are they included in another PR?

i plan to add them later.

core/iwasm/common/wasm_runtime_common.c
yamt: what kind of input parameter checks? for example?

yamt:
> > IIUC, there should be some APIs related to WASI. OR are they included in another PR?
> 
> i plan to add them later.

i meant another PR.

core/iwasm/common/wasm_runtime_common.c
lum1n0us: like `p==null`, `args==null`, and so on. Simply because those are APIs accessed by untrusted (host) code. However, the decision is yours.

yamt: well, i guess host code is trusted in general.

TianlongLiang (approved):

PR #4446, opened 2025-07-03
wasi_nn_llamacpp.c: explicitly reject unimplemented input index

yamt:
note: wasmedge seems to accept index 1 for metadata. we don't implement it.


yamt:
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/16040445096/job/45260837912?pr=4446

Issue #4447, opened 2025-07-03
wasi: unnecessarily strong assumptions about fd 0-2


yamt:
because the plain wasip1 dosen't have connect/listen,
sometimes embedders play a trick to pass an already connected/listening socket.

eg. https://github.com/yamt/garbage/tree/master/wasm/httpd

unfortunately these tricks don't work well for the current wamr if the embedder uses fd 0-2 for it
(as the above example does by default)
because it has unnecessarily strong assumptions about how fd 0-2 are used.

lum1n0us:
> sometimes embedders play a trick to pass an already connected/listening socket.

Please kindly highlight those lines.

yamt:
> > sometimes embedders play a trick to pass an already connected/listening socket.
> 
> Please kindly highlight those lines.

https://github.com/yamt/garbage/blob/4478d1d09d556347062ea6c9b1d161a93775b55e/wasm/httpd/test.sh#L10
https://github.com/yamt/garbage/blob/4478d1d09d556347062ea6c9b1d161a93775b55e/listenexec/main.c#L120

PR #4448, opened 2025-07-03
Revert "libc-wasi: Make rights of STDIN/STDOUT/STDERR fixed and overl…

yamt:
…ook their access modes (#3694)"

This reverts commit 67fa155878861699336380980c3601f733f56194.

because it broke certain use cases while
it isn't clear what it fixed.

tested with:
https://github.com/yamt/garbage/tree/master/wasm/httpd
https://github.com/yamt/garbage/tree/master/wasm/tty

cf.
https://github.com/bytecodealliance/wasm-micro-runtime/issues/4447
https://github.com/bytecodealliance/wasm-micro-runtime/issues/3686#issuecomment-2998903186

### DIFF
--- a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
+++ b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
@@ -459,27 +459,8 @@ fd_determine_type_rights(os_file_handle fd, __wasi_filetype_t *type,
                          __wasi_rights_t *rights_inheriting)
 {
     struct __wasi_filestat_t buf;
-    __wasi_errno_t error;
+    __wasi_errno_t error = os_fstat(fd, &buf);
 
-    if (os_is_stdin_handle(fd)) {
-        *rights_base = RIGHTS_STDIN;
-        *rights_inheriting = RIGHTS_STDIN;
-        return __WASI_ESUCCESS;
-    }
-
-    if (os_is_stdout_handle(fd)) {
-        *rights_base = RIGHTS_STDOUT;
-        *rights_inheriting = RIGHTS_STDOUT;
-        return __WASI_ESUCCESS;
-    }
-
-    if (os_is_stderr_handle(fd)) {
-        *rights_base = RIGHTS_STDERR;
-        *rights_inheriting = RIGHTS_STDERR;
-        return __WASI_ESUCCESS;
-    }
-
-    error = os_fstat(fd, &buf);
     if (error != __WASI_ESUCCESS)
         return error;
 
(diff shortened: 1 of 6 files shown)


lum1n0us:
I mostly agree that STDXXX can have variant `st_filetype` because of *those tricks*,

But, are you suggesting that removing read/write bits based on the access mode is acceptable according to STDXX? 

```c
switch (access_mode) {
    case WASI_LIBC_ACCESS_MODE_READ_ONLY:
        *rights_base &= ~(__wasi_rights_t)__WASI_RIGHT_FD_WRITE;
        break;
    case WASI_LIBC_ACCESS_MODE_WRITE_ONLY:
        *rights_base &= ~(__wasi_rights_t)__WASI_RIGHT_FD_READ;
        break;
}
```

yamt:
> But, are you suggesting that removing read/write bits based on the access mode is acceptable according to STDXX?

yes. i think it's the most straightforward behavior.
what's your concern?

yamt:
@lum1n0us
can we decide on this? (and/or https://github.com/bytecodealliance/wasm-micro-runtime/pull/4395)
or, do you have some more info? (eg. need a research on something)

lum1n0us:
The main concern is that after this modification, the code below will allow STDIN to be unreadable and STDOUT to be unwritable in some scenarios, which I don't think should be permitted.

``` c
switch (access_mode) {
    case WASI_LIBC_ACCESS_MODE_READ_ONLY:
        *rights_base &= ~(__wasi_rights_t)__WASI_RIGHT_FD_WRITE;
        break;
    case WASI_LIBC_ACCESS_MODE_WRITE_ONLY:
        *rights_base &= ~(__wasi_rights_t)__WASI_RIGHT_FD_READ;
        break;
}
```

In my view, the access mode (fcntl() & O_ACCMODE) of STDIN and STDOUT should remain untouchable. Therefore, the rights of STDIN and STDOUT should also be constant.

yamt:
> The main concern is that after this modification, the code below will allow STDIN to be unreadable and STDOUT to be unwritable in some scenarios, which I don't think should be permitted.
> 
> ```c
> switch (access_mode) {
>     case WASI_LIBC_ACCESS_MODE_READ_ONLY:
>         *rights_base &= ~(__wasi_rights_t)__WASI_RIGHT_FD_WRITE;
>         break;
>     case WASI_LIBC_ACCESS_MODE_WRITE_ONLY:
>         *rights_base &= ~(__wasi_rights_t)__WASI_RIGHT_FD_READ;
>         break;
> }
> ```
> 
> In my view, the access mode (fcntl() & O_ACCMODE) of STDIN and STDOUT should remain untouchable. Therefore, the rights of STDIN and STDOUT should also be constant.

do you mean that wasi should be [different from the most of posix platforms](https://github.com/bytecodealliance/wasm-micro-runtime/issues/3686#issuecomment-3010747490) in the regard? why?

Issue #4451, opened 2025-07-03
Performance degradation when changing the "iovs_len" from 1 to 0 in "fd_write" calling


hungryzzz:
### Symptom Description
Hi, I ran the `good.wasm` and `bad.wasm` attached in the `code.zip`, and surprisingly found that when I change the third parameter (`iovs_len`) of `fd_write` function call from `1` to `0`, the execution time of `Wamr` became `4x` longer. But intuitively, when copying a `0` length data to the fd, the execution time may be shorter.

The execution times in the `WasmEdge` (AOT) and `Wamr`(AOT) of both `good.wasm` and `bad.wasm` are as follows (measured by `time` tool):

For `good.wasm`:
- `WasmEdge`: `0.62s`
- `Wamr`: `0.65s`

For `bad.wasm`:
- `WasmEdge`: `0.62s`
- `Wamr`: `2.56s`

### Repreduce Steps

```shell
# run on Wamr
wamrc -o $filename-wamr-aot.wasm $filename.wasm
iwasm $filename-wamr-aot.wasm > case.out

# run on WasmEdge
wasmedgec $filename.wasm $filename-wasmedge-aot.wasm
wasmedge $filename-wasmedge-aot.wasm > case.out
```

### Environment
- Wamr commit: ee056d80766a4
- Operating system: Linux ringzzz-OptiPlex-Micro-Plus-7010 6.5.0-18-generic
- Architecture: Intel(R) Core(TM) i5-13500

[code.zip](https://github.com/user-attachments/files/21038065/code.zip)

yamt:
i guess it's because of the bh_log call https://github.com/bytecodealliance/wasm-micro-runtime/blob/ee056d80766a422f0500be6571e402a38b8087f0/core/iwasm/common/wasm_memory.c#L615-L616

can you try https://github.com/bytecodealliance/wasm-micro-runtime/pull/4452?

hungryzzz:
Hi, I have tried https://github.com/bytecodealliance/wasm-micro-runtime/pull/4452, and the performance degradation disappeared.
Thank you very much!

yamt:
> Hi, I have tried [#4452](https://github.com/bytecodealliance/wasm-micro-runtime/pull/4452), and the performance degradation disappeared. Thank you very much!

thank you for the confirmation.

PR #4452, opened 2025-07-04
wasi: avoid user-triggerable 0-sized allocations

yamt:
might fix https://github.com/bytecodealliance/wasm-micro-runtime/issues/4451

### DIFF
--- a/core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
+++ b/core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
@@ -375,6 +375,9 @@ wasi_fd_pread(wasm_exec_env_t exec_env, wasi_fd_t fd, iovec_app_t *iovec_app,
         return (wasi_errno_t)-1;
 
     total_size = sizeof(wasi_iovec_t) * (uint64)iovs_len;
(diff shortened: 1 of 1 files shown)


lum1n0us (approved):

core/iwasm/libraries/libc-wasi/libc_wasi_wrapper.c
```diff
@@ -375,6 +375,9 @@ wasi_fd_pread(wasm_exec_env_t exec_env, wasi_fd_t fd, iovec_app_t *iovec_app,
         return (wasi_errno_t)-1;
 
     total_size = sizeof(wasi_iovec_t) * (uint64)iovs_len;
+    if (total_size == 0) {
```
loganek, line 378: if `total_size` is zero (i.e. `iovs_len` is zero), can't we just return early, instead of doing the rest of the work in this function?

yamt: maybe we can.
but i don't think it makes sense to optimize this rarely-used case.
it's better to keep it consistent with the normal case. (eg. validate fd etc)

TianlongLiang (approved):

Issue #4457, opened 2025-07-07
[Regression] Failed to compile wasm binary  to aot file


Septa2112:
### Subject of the issue

During regression testing, I found that the current version of `wamrc` fails to compile the wasm file introduced in #2833 into an AOT file. 

### Test case

You can reproduce the issue using the wasm file from #2833, or with the following minimal test case `test.wat`:
```
(module
  (type (;0;) (func (result v128 v128 v128 v128)))
  (func (;0;) (type 0) (result v128 v128 v128 v128)
    (local v128)
    v128.const i32x4 0xbfe4b646 0xee2ae7a5 0x49ff9c4c 0x10fd4f0b
    local.set 0
    nop
    block  ;; label = @1
      v128.const i32x4 0x9a0abccb 0x3fb3e0ea 0xb4903313 0xef0b2868
      local.get 0
      i32.const 0
      select
      local.tee 0
      local.set 0
    end
    local.get 0
    local.get 0
    local.get 0
    local.get 0
    return)
  (memory (;0;) 65536 65536)
  (export "test" (func 0))
  (data (;0;) (i32.const -11152137) "test"))
```

### Your environment

* Ubuntu 22.04
* wamrc 2.3.1

### Steps to reproduce

```
cd /path/to/wasm-micro-runtime/wamr-compiler
mkdir build && cd build
cmake .. && make
./wamrc -o test.aot test.wasm
```

### Expected behavior
```
Create AoT compiler with:
  target:        x86_64
  target cpu:    rocketlake
  target triple: x86_64-unknown-linux-gnu
  cpu features:  
  opt level:     3
  size level:    3
  output format: AoT file
Compile success, file test.aot was generated.
```
### Actual behavior

```bash
Create AoT compiler with:
  target:        x86_64
  target cpu:    rocketlake
  target triple: x86_64-unknown-linux-gnu
  cpu features:  
  opt level:     3
  size level:    3
  output format: AoT file
Error: find block end addr failed.
```

Septa2112:
After investigation, it appears that this issue is related to the changes introduced in PR #4236.

As shown in the code below, when the opcode is `WASM_OP_SELECT_128`, it fails to recognize this opcode correctly and returns false because `FAST_INTERP` is not enabled in `wamrc`.
[...]

PR #4461, opened 2025-07-08
modify macro related to simde when WASM_OP_SELECT_128

Septa2112:
Refer to #4457

### DIFF
--- a/core/iwasm/interpreter/wasm_loader.c
+++ b/core/iwasm/interpreter/wasm_loader.c
@@ -13254,7 +13254,7 @@ re_scan:
                     uint8 opcode_tmp = WASM_OP_SELECT;
 
                     if (type == VALUE_TYPE_V128) {
-#if WASM_ENABLE_JIT != 0 || WASM_ENABLE_SIMDE != 0
+#if WASM_ENABLE_SIMDE != 0
                         opcode_tmp = WASM_OP_SELECT_128;
 #else
                         set_error_buf(error_buf, error_buf_size,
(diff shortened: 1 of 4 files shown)


lum1n0us (approved):

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -13256,8 +13254,7 @@ wasm_loader_prepare_bytecode(WASMModule *module, WASMFunction *func,
                     uint8 opcode_tmp = WASM_OP_SELECT;
 
                     if (type == VALUE_TYPE_V128) {
-#if WASM_ENABLE_JIT != 0 \
-    || WASM_ENABLE_FAST_INTERP != 0 && WASM_ENABLE_SIMD != 0
+#if WASM_ENABLE_JIT != 0 || WASM_ENABLE_SIMDE != 0
```
TianlongLiang: seems that `WASM_ENABLE_SIMDE != 0` is sufficient since `WASM_OP_SELECT_128 = 0xe2,` is only avaiable when `WASM_ENABLE_SIMDE != 0`

Zzzabiyaka:
I wish we could catch things like that easier 

Filled #4484 to see if we could do something in a more resilient way

core/iwasm/interpreter/wasm_loader.c
Septa2112: Addressed

TianlongLiang (approved):

PR #4465, opened 2025-07-08
appease a few compiler warnings (-Wstrict-prototypes)

yamt:


lum1n0us (approved):
@yamt Do you think we should include `-Wstrict-prototypes` in the global warnings? https://github.com/bytecodealliance/wasm-micro-runtime/pull/4397/

yamt:
> @yamt Do you think we should include `-Wstrict-prototypes` in the global warnings? #4397

maybe.
i noticed this during nuttx build, which has it's own set of compiler options.

PR #4466, opened 2025-07-08
wasi_socket_ext.c: fix the number of getaddrinfo results

yamt:
the case of zero results was especially broken.

### DIFF
--- a/core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
+++ b/core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
@@ -541,10 +541,10 @@ getaddrinfo(const char *node, const char *service, const struct addrinfo *hints,
         }
     } while (max_info_size > addr_info_size);
 
+    addr_info_size = max_info_size;
     if (addr_info_size == 0) {
         free(addr_info);
-        *res = NULL;
-        return __WASI_ERRNO_SUCCESS;
+        HANDLE_ERROR(__WASI_ERRNO_NOENT)
     }
 
     aibuf_res =
@@ -555,10 +555,6 @@ getaddrinfo(const char *node, const char *service, const struct addrinfo *hints,
     }
 
     *res = &aibuf_res[0].ai;
-
-    if (addr_info_size) {
-        addr_info_size = max_info_size;
-    }
 
     for (i = 0; i < addr_info_size; i++) {
         struct addrinfo *ai = &aibuf_res[i].ai;


core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
```diff
@@ -541,10 +541,10 @@ getaddrinfo(const char *node, const char *service, const struct addrinfo *hints,
         }
     } while (max_info_size > addr_info_size);
 
+    addr_info_size = max_info_size;
```
lum1n0us, line 544: IIUC, the `do...while...` loop above will assign a value (`>= max_info_size`) to `addr_info_size`. There is no need to assign it again.

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
```diff
@@ -541,10 +541,10 @@ getaddrinfo(const char *node, const char *service, const struct addrinfo *hints,
...
     if (addr_info_size == 0) {
         free(addr_info);
-        *res = NULL;
-        return __WASI_ERRNO_SUCCESS;
+        HANDLE_ERROR(__WASI_ERRNO_NOENT)
```
lum1n0us, line 547: I'm uncertain about that, mainly because I'm not sure when `addr_info_size` can be zero.

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
yamt: at this point, addr_info_size is the allocated buffer size and max_info_size is what __wasi_sock_addr_resolve actually returned.
it's more appropriate to use the latter for the following logic.

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
yamt: __wasi_sock_addr_resolve can return 0 entries.
getaddrinfo should not return 0 entries.

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
yamt: i observed the case max_info_size == 0 and addr_info_size != 0.
in that case, the original logic returns a zero-filled `res`, which is quite broken.

lum1n0us: > it's more appropriate to use the latter for the following logic.

the allocation of `aibuf_res`? Agree.

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
lum1n0us: Uncertain about the original design intent to treat `addr_info_size==0` as a success rather than a failure. The modification may disrupt the API's original return value. 

@loganek please help us out.

yamt: getaddrinfo is standard-defined.
whatever the original intention was, we should not provide incompatible api.

susv4-2018 says:
> Upon successful return of getaddrinfo(), the location to which res points shall refer to a linked list of addrinfo structures, each of which shall specify a socket address and information for use in creating a socket with which to use that socket address. The list shall include at least one addrinfo structure.

it seems clear that returning zero result is not conforming.

loganek: yeah looks like inconsistency with the documentation that should be fixed. I don't think we should be touching __wasi_sock_addr_resolve though, it's not the same as getaddrinfo so it's ok if it acts differently (plus that'd be backward incompatible change).

lum1n0us (approved):

PR #4469, opened 2025-07-09
posix os_socket_addr_resolve: relax compatibility check

yamt:
some getaddrinfo implementations sometimes return results with ai_protocol=0.

cf. https://github.com/apache/nuttx/issues/16693

standard-wise, i couldn't find anything about ai_protocol in susv4-2018 except the following text:

> The fields ai_family, ai_socktype, and ai_protocol shall be usable
> as the arguments to the socket() function to create a socket suitable
> for use with the returned address.

because ai_protocol is usually merely used to feed socket() and socket() usually accepts 0, it's probably standard-wise ok for getaddrinfo to return ai_protocol=0.
the major implementations of getaddrinfo (eg. kame) seem to return specific values like ai_protocol=IPPROTO_TCP though.

anyway, for the purpose of this function, there is little point to be strict on the host getaddrinfo behavior. this commit just relaxes the check to be friendly to those getaddrinfo implementations.

### DIFF
--- a/core/shared/platform/common/posix/posix_socket.c
+++ b/core/shared/platform/common/posix/posix_socket.c
@@ -373,8 +373,8 @@ is_addrinfo_supported(struct addrinfo *info)
         (info->ai_family == AF_INET || info->ai_family == AF_INET6)
         // Allow only UDP and TCP
         && (info->ai_socktype == SOCK_DGRAM || info->ai_socktype == SOCK_STREAM)
-        && (info->ai_protocol == IPPROTO_TCP
-            || info->ai_protocol == IPPROTO_UDP);
+        && (info->ai_protocol == IPPROTO_TCP || info->ai_protocol == IPPROTO_UDP
+            || info->ai_protocol == 0);
 }
 
 int


core/shared/platform/common/posix/posix_socket.c
```diff
@@ -373,8 +373,8 @@ is_addrinfo_supported(struct addrinfo *info)
         (info->ai_family == AF_INET || info->ai_family == AF_INET6)
         // Allow only UDP and TCP
         && (info->ai_socktype == SOCK_DGRAM || info->ai_socktype == SOCK_STREAM)
-        && (info->ai_protocol == IPPROTO_TCP
-            || info->ai_protocol == IPPROTO_UDP);
+        && (info->ai_protocol == IPPROTO_TCP || info->ai_protocol == IPPROTO_UDP
+            || info->ai_protocol == 0);
```
lum1n0us, line 377:
```
        ai_socktype
              This field specifies the preferred socket type, for example
              SOCK_STREAM or SOCK_DGRAM.  Specifying 0 in this field
              indicates that socket addresses of any type can be returned
              by getaddrinfo().

       ai_protocol
              This field specifies the protocol for the returned socket
              addresses.  Specifying 0 in this field indicates that
              socket addresses with any protocol can be returned by
              getaddrinfo().
```

[It](https://man7.org/linux/man-pages/man3/getaddrinfo.3.html) seems that both `ai_socktype` and `ai_protocol` use 0 to indicate something. Perhaps we should change both in this PR?

By the way, should we synchronize the modification with espidf_socket.c?

yamt: > ```
>         ai_socktype
>               This field specifies the preferred socket type, for example
>               SOCK_STREAM or SOCK_DGRAM.  Specifying 0 in this field
>               indicates that socket addresses of any type can be returned
>               by getaddrinfo().
> 
>        ai_protocol
>               This field specifies the protocol for the returned socket
>               addresses.  Specifying 0 in this field indicates that
>               socket addresses with any protocol can be returned by
>               getaddrinfo().
> ```
> 
> [It](https://man7.org/linux/man-pages/man3/getaddrinfo.3.html) seems that both `ai_socktype` and `ai_protocol` use 0 to indicate something. Perhaps we should change both in this PR?

the description you cited is about hints to getaddrinfo.
this code is checking is the result of getaddrinfo.
while they use the same C structure, they are not logically same.

> By the way, should we synchronize the modification with espidf_socket.c?

probably. i don't know why it has separate implementation.

yamt: > > By the way, should we synchronize the modification with espidf_socket.c?
> 
> probably. i don't know why it has separate implementation.

i filed an issue: https://github.com/bytecodealliance/wasm-micro-runtime/issues/4473

lum1n0us (approved):

PR #4472, opened 2025-07-10
Fix typos

TianlongLiang:
* Fix typos in runtime timer functions and in comments

### DIFF
--- a/core/shared/utils/runtime_timer.h
+++ b/core/shared/utils/runtime_timer.h
@@ -15,7 +15,7 @@ extern "C" {
 uint64
 bh_get_tick_ms(void);
 uint32
-bh_get_elpased_ms(uint32 *last_system_clock);
+bh_get_elapsed_ms(uint32 *last_system_clock);
 
 struct _timer_ctx;
 typedef struct _timer_ctx *timer_ctx_t;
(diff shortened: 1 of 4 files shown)


lum1n0us (approved):

core/shared/utils/runtime_timer.h
```diff
@@ -15,7 +15,7 @@ extern "C" {
 uint64
 bh_get_tick_ms(void);
 uint32
-bh_get_elpased_ms(uint32 *last_system_clock);
+bh_get_elapsed_ms(uint32 *last_system_clock);
```
yamt: this is an api breaking change.
wamr-app-framework seems using this. can you take care of it?

TianlongLiang: will do

loganek: what's the plan for addressing it? Are we leaving it as is, or we create a new API and mark the existing one as deprecated/to be removed in the next major release?

TianlongLiang: On my second thought, I prefer leaving it as is, I think it's too much of a fuss to break the API over a typo.

loganek (approved):
LGTM

yamt (approved):

Issue #4474, opened 2025-07-10
lib-socket getsockopt etc abuses __WASI_ERRNO_SUCCESS where 0 is more appropriate


yamt:

yamt:
fixed by https://github.com/bytecodealliance/wasm-micro-runtime/pull/4476

PR #4476, opened 2025-07-10
wasi_socket_ext.c: fix error reporting

yamt:
do not return __WASI_ERRNO_xxx error code where 0/-1 is more appropriate.

most of them were not harmful because the value of __WASI_ERRNO_SUCCESS is 0.

the one with __WASI_ERRNO_NOPROTOOPT in socket() is a user-visible bug.

tested with wamr-wasi-extensions/samples/socket-tcp-udp.

cf. https://github.com/bytecodealliance/wasm-micro-runtime/issues/4474

### DIFF
--- a/core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
+++ b/core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
@@ -191,7 +191,7 @@ bind(int sockfd, const struct sockaddr *addr, socklen_t addrlen)
     error = __wasi_sock_bind(sockfd, &wasi_addr);
     HANDLE_ERROR(error)
 
-    return __WASI_ERRNO_SUCCESS;
+    return 0;
 }
 
 int
@@ -212,7 +212,7 @@ connect(int sockfd, const struct sockaddr *addr, socklen_t addrlen)
     error = __wasi_sock_connect(sockfd, &wasi_addr);
     HANDLE_ERROR(error)
 
-    return __WASI_ERRNO_SUCCESS;
+    return 0;
 }
 
 int
@@ -220,7 +220,7 @@ listen(int sockfd, int backlog)
 {
     __wasi_errno_t error = __wasi_sock_listen(sockfd, backlog);
     HANDLE_ERROR(error)
-    return __WASI_ERRNO_SUCCESS;
+    return 0;
 }
 
 ssize_t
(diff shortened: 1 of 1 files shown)


TianlongLiang (approved):

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
```diff
@@ -385,7 +385,7 @@ socket(int domain, int type, int protocol)
         socktype = SOCKET_STREAM;
     }
     else {
-        return __WASI_ERRNO_NOPROTOOPT;
+        HANDLE_ERROR(__WASI_ERRNO_NOPROTOOPT)
```
loganek, line 388: this macro checks for error value (`if (error != __WASI_ERRNO_SUCCESS)`) but we never set the error value before so it might be garbage, and the result of that macro is non-deterministic in that case. I think the right thing to do is to do one of the two:
1. set error = __WASI_ERRNO_NOPROTOOPT, call `HANDLE_ERROR` (so the errno value is set) and return -1
2. define macro `#define SET_ERROR(error) do { errno = error; return -1; } while (0)` and use it here.

Having the macro would be useful, because it should be used everywhere where we call `return __WASI_ERRNO_*` - the socket() function should always return -1 on error and set `errno` value instead of returning the error itself.

loganek: We should also update the `return __WASI_ERRNO_NOPROTOOPT;` to use the macro, and potentially other places too.

yamt: > this macro checks for error value (`if (error != __WASI_ERRNO_SUCCESS)`) but we never set the error value before so it might be garbage, and the result of that macro is non-deterministic in that case.

i don't understand what you meant.

```c
        HANDLE_ERROR(__WASI_ERRNO_NOPROTOOPT)
```
should expand to
```c
    if (__WASI_ERRNO_NOPROTOOPT != __WASI_ERRNO_SUCCESS) {
        errno = __WASI_ERRNO_NOPROTOOPT;
        return -1;
    }
```

loganek: ah right, I got confused because the macro parameter name is the same as local variable name. This one is fine then, but we should still update line 378 in this file.

yamt: > we should still update line 378 in this file.

done. thank you.

loganek (approved):
lgtm

lum1n0us:
@yamt please resolve the conflict. Thank you.ae

yamt:
> @yamt please resolve the conflict. Thank you.ae

done

Issue #4487, opened 2025-07-14
wasi_nn_load_by_name buffer overrun


yamt:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/2a952b371a824d4c853c854e27a88051d28ef295/core/iwasm/libraries/wasi-nn/src/wasi_nn.c#L532

`name[name_len]` is past the end of the buffer.

also, it seems wrong to check the NUL termination here.
iirc, witx string doesn't contain NUL.

yamt:
https://github.com/bytecodealliance/wasm-micro-runtime/pull/4491 fixed this

PR #4491, opened 2025-07-14
wasi-nn: fix buffer overruns in load_by_name/load_by_name_with_config

yamt:
also, ensure NUL terminations here to simplify backend implementations. (note that, although names with '\0' in the middle are usually valid in wasm, wamr doesn't support them in general. we might revisit this later.)

also, add a missing address validation in load_by_name_with_config.

cf. https://github.com/bytecodealliance/wasm-micro-runtime/issues/4487

### DIFF
--- a/core/iwasm/libraries/wasi-nn/src/wasi_nn.c
+++ b/core/iwasm/libraries/wasi-nn/src/wasi_nn.c
@@ -507,20 +507,40 @@ fail:
     return res;
 }
 
+static wasi_nn_error
+copyin_and_nul_terminate(wasm_module_inst_t inst, char *name, uint32_t name_len,
+                         char **resultp)
+{
+    char *nul_terminated_name;
+    if (!wasm_runtime_validate_native_addr(inst, name, name_len)) {
+        return invalid_argument;
+    }
+    nul_terminated_name = wasm_runtime_malloc(name_len + 1);
+    if (nul_terminated_name == NULL) {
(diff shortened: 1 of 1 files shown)


core/iwasm/libraries/wasi-nn/src/wasi_nn.c
```diff
@@ -507,36 +507,57 @@ wasi_nn_load(wasm_exec_env_t exec_env, graph_builder_array_wasm *builder,
...
+    if (!wasm_runtime_validate_native_addr(inst, name, name_len)) {
+        return invalid_argument;
+    }
+    nul_terminated_name = wasm_runtime_malloc(name_len + 1);
+    if (nul_terminated_name == NULL) {
+        return runtime_error;
+    }
+    bh_memcpy_s(nul_terminated_name, name_len + 1, name, name_len);
+    nul_terminated_name[name_len] = '\0'; /* ensure NUL termination */
+    if (strlen(nul_terminated_name) != name_len) {
```
lum1n0us, line 524: This `if` is always false, right?

yamt: usually it's false.
however, not always.
consider (name="\0\0\0", name_len=3).

lum1n0us: Okay, 😆

PR #4496, opened 2025-07-16
Always use release builds of pre-built dependency libraries on Windows

kylo5aby:
On Windows platform, Debug builds of wamrc now use Release runtime 
library (/MD) instead of Debug runtime library (/MDd) to avoid 
linking conflicts with Release LLVM libraries.

This workaround resolves LNK2038 errors like:
- RuntimeLibrary mismatch: 'MD_DynamicRelease' vs 'MDd_DynamicDebug'

A better way would be to using debug LLVM libraries for debug wamrc builds, 
but debug LLVM builds cause CPack failures in GitHub CI, likely due to size constraints.

The change only affects Windows builds and preserves debugging 
capabilities for wamrc code while ensuring compatibility with 
pre-built Release LLVM libraries.

### DIFF
--- a/build-scripts/config_common.cmake
+++ b/build-scripts/config_common.cmake
@@ -99,6 +99,12 @@ if (WAMR_BUILD_JIT EQUAL 1)
     # Enable Lazy JIT by default
     set (WAMR_BUILD_LAZY_JIT 1)
   endif ()
+
+  # In Debug mode, always use release builds of pre-built dependency libraries
+  if (WAMR_BUILD_PLATFORM STREQUAL "windows" AND MSVC)
+    add_compile_options($<$<CONFIG:Debug>:/MD>)
+  endif()
+
   if (NOT DEFINED LLVM_DIR)
     set (LLVM_SRC_ROOT "${WAMR_ROOT_DIR}/core/deps/llvm")
     set (LLVM_BUILD_ROOT "${LLVM_SRC_ROOT}/build")
--- a/wamr-compiler/CMakeLists.txt
+++ b/wamr-compiler/CMakeLists.txt
@@ -159,6 +159,11 @@ if (WAMR_BUILD_DEBUG_AOT EQUAL 1)
 endif()
 
 # Enable LLVM
+# In Debug mode, always use release builds of pre-built dependency libraries
+if (WAMR_BUILD_PLATFORM STREQUAL "windows" AND MSVC)
+  add_compile_options($<$<CONFIG:Debug>:/MD>)
+endif()
+
 if (NOT WAMR_BUILD_WITH_CUSTOM_LLVM)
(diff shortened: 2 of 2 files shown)


wamr-compiler/CMakeLists.txt
```diff
@@ -161,6 +161,13 @@ endif()
 # Enable LLVM
 if (NOT WAMR_BUILD_WITH_CUSTOM_LLVM)
   set (LLVM_SRC_ROOT "${PROJECT_SOURCE_DIR}/../core/deps/llvm")
+
+  # In Debug mode, force using Release runtime library on Windows platform
+  if (WAMR_BUILD_PLATFORM STREQUAL "windows")
+    set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} /MD")
+    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} /MD")
+  endif()
+
```
lum1n0us: - Does iwasm+llvmjit have the same issue?
- Both pre-built libraries and libraries compiled from source share the same problem; we might want to add `/MD` for both `WAMR_BUILD_WITH_CUSTOM_LLVM` and `not WAMR_BUILD_WITH_CUSTOM_LLVM` scenarios.
- I personally suggest using `add_compile_options()` to achieve greater flexibility and compatibility.

kylo5aby: > * Does iwasm+llvmjit have the same issue?
> * Both pre-built libraries and libraries compiled from source share the same problem; we might want to add `/MD` for both `WAMR_BUILD_WITH_CUSTOM_LLVM` and `not WAMR_BUILD_WITH_CUSTOM_LLVM` scenarios.
> * I personally suggest using `add_compile_options()` to achieve greater flexibility and compatibility.

Added same logic for iwasm+llvmjit

build-scripts/config_common.cmake
```diff
@@ -99,6 +99,12 @@ if (WAMR_BUILD_JIT EQUAL 1)
     # Enable Lazy JIT by default
     set (WAMR_BUILD_LAZY_JIT 1)
   endif ()
+
+  # In Debug mode, force using LLVM Release runtime library on Windows platform
```
lum1n0us:
```suggestion
  # In Debug mode, always use release builds of pre-built dependency libraries
```

wamr-compiler/CMakeLists.txt
```diff
@@ -159,6 +159,11 @@ if (WAMR_BUILD_DEBUG_AOT EQUAL 1)
 endif()
 
 # Enable LLVM
+# In Debug mode, force using LLVM Release runtime library on Windows platform
```
lum1n0us:
```suggestion
# In Debug mode, always use release builds of pre-built dependency libraries
```

lum1n0us (approved):

Issue #4464, opened 2025-07-08
lib-socket getaddrinfo returns -1


yamt:
our lib-socket implementation of getaddrinfo returns -1 on error, storing the error code to errno.
however, standard-wise, getaddrinfo is supposed to return an EAI_xxx error code.

yamt:
fixed by https://github.com/bytecodealliance/wasm-micro-runtime/pull/4498

PR #4498, opened 2025-07-17
lib-socket: make getaddrinfo return EAI_ values

yamt:
cf. https://github.com/bytecodealliance/wasm-micro-runtime/issues/4464

### DIFF
--- a/core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
+++ b/core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
@@ -38,6 +38,13 @@ __errno_location(void);
         return -1;                       \
     }
 
+/* REVISIT: in many cases, EAI_SYSTEM may not be an ideal error code */
+#define GAI_HANDLE_ERROR(error)          \
+    if (error != __WASI_ERRNO_SUCCESS) { \
+        errno = error;                   \
+        return EAI_SYSTEM;               \
+    }
+
 static void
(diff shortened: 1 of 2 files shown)


core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
```diff
@@ -38,6 +38,13 @@ __errno_location(void);
         return -1;                       \
     }
 
+/* REVISIT: in many cases, EAI_SYSTEM may not be an ideal error code */
+#define GAI_HANDLE_ERROR(error)          \
```
lum1n0us, line 42: May I ask what GAI stands for?

lum1n0us (commented):
It seems to disrupt the recently modified ABIs that only return 0 and -1. Is it worthwhile?

core/iwasm/libraries/lib-socket/src/wasi/wasi_socket_ext.c
yamt: getaddrinfo.

yamt:
> It seems to disrupt the recently modified ABIs that only return 0 and -1. Is it worthwhile?

https://github.com/bytecodealliance/wasm-micro-runtime/pull/4476 is about traditional socket system calls like setsockopt.

this PR is about getaddrinfo, which is relatively new and has a different convention.

loganek (approved):
LGTM

lum1n0us (approved):

Issue #4504, opened 2025-07-18
extended const should have bumped AOT_CURRENT_VERSION


yamt:
[extended const](https://github.com/bytecodealliance/wasm-micro-runtime/pull/4432) changed the aot format in a way which affects very wide range of modules.
i guess it should have bumped AOT_CURRENT_VERSION.

the situation is a bit complex because we have released 2.4.0 a few days ago.

my suggestion is:
1. withdraw 2.4.0 (hopefully not too late)
2. bump AOT_CURRENT_VERSION
3. release 2.4.1

TianlongLiang:
You are right, we should bump AOT_CURRENT_VERSION and create a new release for it

yamt:
fixed by 2.4.1

PR #4511, opened 2025-07-24
bump AOT_CURRENT_VERSION for extended-const

yamt:
cf. https://github.com/bytecodealliance/wasm-micro-runtime/issues/4504


loganek (approved):
LGTM

PR #4512, opened 2025-07-24
Add regression test CI

Septa2112:
Add regression test CI on Ubuntu 22.04


Septa2112:
- The entire CI test run takes approximately 3 minutes. 
- The test run fails if there is at least one failed case (`failed > 0`).

PR #4514, opened 2025-07-25
Use venv to install llvm dependencies and run build script

loganek:
Installing pip packages system-wide is no longer recommended, therefore we create virtual environment for setting up the build environment for LLVM.

### DIFF
--- a/wamr-compiler/build_llvm.sh
+++ b/wamr-compiler/build_llvm.sh
@@ -3,5 +3,17 @@
 # Copyright (C) 2020 Intel Corporation. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-/usr/bin/env python3 -m pip install --user -r ../build-scripts/requirements.txt
+TEMP_DIR=$(mktemp -d)
+
+cleanup() {
+    local exit_code=$?
+    rm -rf "$TEMP_DIR"
+    exit $exit_code
+}
+
+trap cleanup EXIT INT TERM
+
+/usr/bin/env python3 -m venv --clear "$TEMP_DIR"
+source "$TEMP_DIR/bin/activate"
+/usr/bin/env python3 -m pip install -r ../build-scripts/requirements.txt
 /usr/bin/env python3 ../build-scripts/build_llvm.py "$@"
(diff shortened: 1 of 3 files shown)


TianlongLiang (approved):

TianlongLiang:
Maybe we should add venv/ to .gitignore? What do you think

wamr-compiler/build_llvm_arc.sh
```diff
@@ -3,5 +3,5 @@
 # Copyright (C) 2020 Intel Corporation. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-/usr/bin/env python3 -m pip install --user -r ../build-scripts/requirements.txt
-/usr/bin/env python3 ../build-scripts/build_llvm.py --platform arc "$@"
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+$SCRIPT_DIR/build_llvm.sh --platform arc "$@"
```
yamt, line 7: it's better to avoid assuming /bin/sh is bash.
what's wrong with "$0"?

loganek: ok, updated.

loganek:
> Maybe we should add venv/ to .gitignore? What do you think

There's no venv directory created as part of this process; instead, we create a venv folder in the system's temp directory.

TianlongLiang:
Got it, thanks

wamr-compiler/build_llvm.sh
```diff
@@ -3,5 +3,17 @@
...
+cleanup() {
+    local exit_code=$?
+    rm -rf "$TEMP_DIR"
+    exit $exit_code
+}
+
+trap cleanup EXIT INT TERM
+
+/usr/bin/env python3 -m venv --clear "$TEMP_DIR"
+source "$TEMP_DIR/bin/activate"
```
yamt, line 17: https://github.com/bytecodealliance/wasm-micro-runtime/pull/4541

wamr-compiler/build_llvm.sh
```diff
@@ -3,5 +3,17 @@
...
+
+cleanup() {
+    local exit_code=$?
+    rm -rf "$TEMP_DIR"
+    exit $exit_code
+}
+
+trap cleanup EXIT INT TERM
+
+/usr/bin/env python3 -m venv --clear "$TEMP_DIR"
```
yamt, line 16: why do you use /usr/bin/env?

PR #4520, opened 2025-07-29
Add validation for rec indices in aot loader

kylo5aby:
Ensure `rec_idx` within valid ranges to prevent accessing invalid type references.

### DIFF
--- a/core/iwasm/aot/aot_loader.c
+++ b/core/iwasm/aot/aot_loader.c
@@ -1805,7 +1805,12 @@ load_types(const uint8 **p_buf, const uint8 *buf_end, AOTModule *module,
         read_uint32(buf, buf_end, parent_type_idx);
         read_uint16(buf, buf_end, rec_count);
         read_uint16(buf, buf_end, rec_idx);
-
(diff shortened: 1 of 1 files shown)


core/iwasm/aot/aot_loader.c
```diff
@@ -1805,7 +1805,12 @@ load_types(const uint8 **p_buf, const uint8 *buf_end, AOTModule *module,
         read_uint32(buf, buf_end, parent_type_idx);
         read_uint16(buf, buf_end, rec_count);
         read_uint16(buf, buf_end, rec_idx);
-
+#if WASM_ENABLE_AOT_VALIDATOR != 0
+        if (rec_idx > i) {
```
lum1n0us, line 1809: Are we certain this type is REC?

kylo5aby: > Are we certain this type is REC?

This checks whether the type index in a rec group is valid.
If GC is enabled, even a single type will have a `rec_idx` of 0.
We do not check whether `parent_type_idx` is valid here, because it is already checked later.

lum1n0us (approved):

PR #4521, opened 2025-07-29
loader: add type index checking

kylo5aby:


core/iwasm/aot/aot_loader.c
```diff
@@ -1786,7 +1786,7 @@ load_types(const uint8 **p_buf, const uint8 *buf_end, AOTModule *module,
 
             read_uint32(buf, buf_end, j);
 #if WASM_ENABLE_AOT_VALIDATOR != 0
-            if (j >= module->type_count) {
+            if (j > i) {
```
lum1n0us, line 1790: It's better to add some comments to explain this assumption `j > i`.

kylo5aby: > It's better to add some comments to explain this assumption `j > i`.

Done

lum1n0us (approved):

PR #4522, opened 2025-07-29
Add validation for recursive type count in loader

kylo5aby:


core/iwasm/interpreter/wasm_loader.c
```diff
@@ -2499,6 +2496,13 @@ load_type_section(const uint8 *buf, const uint8 *buf_end, WASMModule *module,
 #endif /* end of WASM_ENABLE_GC == 0 */
     }
 
+    for (i = 0; i < module->type_count; i++) {
+        if (module->types[i] == NULL) {
+            set_error_buf_v(error_buf, error_buf_size, "unknown type %d", i);
```
lum1n0us, line 2501: Could you please explain the reason for this modification? Is it solely related to the error message?

kylo5aby: > Could you please explain the reason for this modification? Is it solely related to the error message?

for a valid module, all types should be initialized after `load_type_section`, for an invalid one (with a malformed type section), some types may remain uninitialized.
If such a type is later referenced in other sections, it may cause a crash. Therefore, I add a validation step here to ensure that all types are properly initialized.

lum1n0us: However, the original `module->types[type_idx] == NULL` can also catch those invalid cases and achieve validation. Is it a kind of code refactoring ?

kylo5aby: > However, the original `module->types[type_idx] == NULL` can also catch those invalid cases and achieve validation. Is it a kind of code refactoring ?

since this check is already performed in type section, repeating the same checks in other sections during loading seems duplicate.

PR #4537, opened 2025-08-04
Improve build scripts for bash compatibility and update README

Septa2112:
## Issue
Now running `./build_llvm.sh` will cause an issue: `./build_llvm.sh: 17: source: not found`. 

Because the script is executed with the system default shell (`sh`), which does not support the `source` command used in the script, it will cause the Python virtual environment not to activate actually.

## How to Fix
The script should either be explicitly run using `bash` (e.g., `bash ./build_llvm.sh`) or the script itself should specify `bash` as the default shell by adding the shebang line `#!/usr/bin/env bash` at the top.

## My improvements:

- Updated the README.
- Modified the shebang line.
- Added `set -e` to make the script exit immediately if any command fails.

### DIFF
--- a/wamr-compiler/build_llvm.sh
+++ b/wamr-compiler/build_llvm.sh
@@ -1,7 +1,9 @@
-#!/bin/sh
+#!/usr/bin/env bash
 
 # Copyright (C) 2020 Intel Corporation. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+set -e
 
 TEMP_DIR=$(mktemp -d)
 
--- a/wamr-compiler/README.md
+++ b/wamr-compiler/README.md
@@ -13,6 +13,7 @@ sudo apt-get install git build-essential cmake g++-multilib libgcc-9-dev lib32gc
 
 ```shell
 cd wamr-compiler
+# or explicitly with bash: bash ./build_llvm.sh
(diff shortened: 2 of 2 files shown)


wamr-compiler/build_llvm.sh
```diff
@@ -1,8 +1,10 @@
-#!/bin/sh
+#!/usr/bin/env bash
```
lum1n0us, line 1: There is no doubt that this can solve the problem. However, `#!/bin/bash` is commonly used in bash scripts within this project. It is difficult not to consider whether we should apply this modification to all other bash scripts to prevent similar default shell issues.

Personally, I believe the best approach is to verify the default shell and set it to bash when preparing the environment.

```bash
## identify available shells.
$ cat /etc/shells

## verify current default shell
$ echo "$SHELL"

## change the default shell.
$ chsh -s /path/to/new/shell
```

yamt:
@Septa2112 does https://github.com/bytecodealliance/wasm-micro-runtime/pull/4541 fix the issue for your env?

Septa2112:
@yamt yes, it works for my env. And your changes are cleaner and more reasonable. I'm considering closing my PR and adopting yours.

yamt:
> @yamt yes, it works for my env. And your changes are cleaner and more reasonable. I'm considering closing my PR and adopting yours.

thank you for confirmation.

PR #4546, opened 2025-08-06
Fix integer overflow in GC threshold calculation

kylo5aby:


core/shared/mem-alloc/ems/ems_gc_internal.h
```diff
@@ -338,8 +338,18 @@ typedef struct gc_heap_struct {
...
+    if (result > UINT32_MAX) {
+        /* Threshold factor can be greater than 1000 (100%), which means
+         *  GC will never be triggered. So heap->gc_threshold >
+         * APP_HEAP_SIZE_MAX is allowed
+         */
+        heap->gc_threshold = UINT32_MAX;
+    }
+    else {
+        heap->gc_threshold = (uint32_t)result;
+    }
```
lum1n0us: > `heap->gc_threshold_factor = GC_DEFAULT_THRESHOLD_FACTOR;`. 

IIUC, `gc_threshold_factor` is always be 300.

kylo5aby: > > `heap->gc_threshold_factor = GC_DEFAULT_THRESHOLD_FACTOR;`.
> 
> IIUC, `gc_threshold_factor` is always be 300.

`heap->total_free_size` would be a big value. Although it wouldn't exceed 2^30(`GC_HEAP_SIZE_MAX`),  but `heap->total_free_size*heap->gc_threshold_factor` might overflow `uint32`.

Since `gc_threshold_factor` is fixed and its value is 300, so I think just check `heap->total_free_size*heap->gc_threshold_factor` is OK. thanks for reminder

lum1n0us (approved):

Issue #4480, opened 2025-07-10
WAMR JIT mode incorrectly outputs type_mismatch


teyahb8:
### Subject of the issue

I executed the following test case with different runtimes. Other runtimes supporting typed function references would output the same number, including WAMR's AOT mode. But, WAMR's JIT mode outputs "type mismatch" error. Other runtime modes that don't support typed function references would output that "typed function references are not supported."

### Test case

```
(module
  (type (;0;) (func))
  (type (;1;) (func (result v128)))
  (export "main" (func 0))
  (export "to_test" (func 0))
  (func (;0;) (type 1) (result v128)
    v128.const i32x4 0x00000000 0x00000000 0x00000000 0xfd000000
    br 0
    unreachable
    unreachable
    unreachable
    unreachable
    unreachable
    unreachable
    unreachable
    unreachable
    unreachable
    unreachable
    i64.or
    ref.null 0
    unreachable
    i16x8.lt_u
  )
)
```

### Your environment

* Host OS: Ubuntu 22.04
* WAMR version (2.3.1), platform (Linux), cpu architecture (x86_64), running mode (JIT).

### Steps to reproduce

Run the command: `iwasm --llvm-jit -f main i16x8.lt_u_15_10_1752003706322386.wasm`

### Expected behavior

Should output "336294682933583715844663186250927177728".

### Actual behavior

Output is "type mismatch"

### Extra Info

Can you please confirm if this is a bug in WAMR's JIT mode? Thanks in advance.

teyahb8:
Were you able to reproduce this, @TianlongLiang?

Septa2112:
I seem to have encountered some issues while trying to reproduce this. If possible, could you attach the `wasm` binary you tested and share some more detailed steps, such as the compilation options you used? That would be helpful.

teyahb8:
Thanks for the feedback. Here are the compilation options:

```
cmake -DWAMR_BUILD_BULK_MEMORY=1 -DWAMR_BUILD_REF_TYPES=1 -DWAMR_BUILD_SIMD=1 -DWAMR_BUILD_JIT=1 -DLLVM_DIR=/usr/lib/llvm-18/lib/cmake/llvm ..
make -j
```

And, here is the testcase in .zip format since .wasm file could not be uploaded here.

**Testcase:** [testcase.zip](https://github.com/user-attachments/files/21432858/testcase.zip)

Septa2112:
[...]

PR #4547, opened 2025-08-07
Enhance wamrc build configuration to support optional gc

### LINKED ISSUE #4542: How to handle `ref.null 0`?
`ref.null 0` is valid under the GC specification but invalid under the current specification(only allow `externref` and `funcref`). This means a Wasm runtime should report a validation error when loading a Wasm module that contains `ref.null` 0 if GC is not enabled. The same applies to wamrc.

But the latest version of wamrc will not report a validation error. This is because:
- By design, unlike iwasm, wamrc is a full-feature component that uses execution switchers (command line options) instead of compilation flags to configure all supported features. So GC is always-on in wamrc.
- wamrc will validate the `ref.null` opcode in WebAssembly modules under the GC specification by always enabling the compilation flag `WASM_ENABLE_GC`. Although there is a `--enable-gc` option in wamrc's command line options which is able to disable gc during execution, this option is not passed to the WebAssembly loader and is only used during and after AOT loading.
- There should be an additional opcode validator in code generation (compilation) after WebAssembly loading in wamrc. However, in a specific case(#4480), `br 0` will skip all following opcodes, including `ref.null 0`.

In my opinion, there are still three options available to handle this `ref.null 0`.
a. pass command line options of wamrc to wasm loader. like an extra parameter in `wasm_runtime_load()`.
b. add opcode validation in `aot_validate_wasm()`
c. in CMakeLists.txt of wamrc, make `WASM_ENABLE_GC` configurable. 

From my perspective, options a. and b. are too burdensome since `ref.null` is the only opcode that currently requires extra attention. I prefer to use option c.

lum1n0us:
There will be two sets of binaries for iwasm and wamrc. One set without a suffix indicates there is no GC support. Another set with the suffixes "-gc" and "-gc-eh" includes GC support. Users cannot disable GC through command line options; they need to choose the appropriate set to use.

fix #4542 and #4480


lum1n0us:
NOT an ideal solution.

PR #4548, opened 2025-08-07
Add bounds checking to prevent local count from exceeding UINT16_MAX

kylo5aby:
Add bounds checking to prevent `ref_type_map_count `(which is come from the sum of every sub local count) from exceeding
UINT16_MAX.

### DIFF
--- a/core/iwasm/interpreter/wasm_loader.c
+++ b/core/iwasm/interpreter/wasm_loader.c
@@ -1799,6 +1799,11 @@ resolve_func_type(const uint8 **p_buf, const uint8 *buf_end, WASMModule *module,
         return false;
     }
     if (ref_type_map_count > 0) {
+        if (ref_type_map_count > UINT16_MAX) {
+            set_error_buf(error_buf, error_buf_size,
+                          "ref type count too large");
+            return false;
+        }
         total_size = sizeof(WASMRefTypeMap) * (uint64)ref_type_map_count;
         if (!(type->ref_type_maps =
                   loader_malloc(total_size, error_buf, error_buf_size))) {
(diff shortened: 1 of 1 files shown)


core/iwasm/interpreter/wasm_loader.c
```diff
@@ -3959,6 +3959,11 @@ load_function_section(const uint8 *buf, const uint8 *buf_end,
                           total_size, error_buf, error_buf_size))) {
                     return false;
                 }
+                if (ref_type_map_count > UINT16_MAX) {
```
TianlongLiang: Maybe we should move this condition before L3958, so no need for unnecessary malloc

TianlongLiang:
Also, will this issue also affect `resolve_struct_type` and `resolve_func_type` ?

kylo5aby:
> Also, will this issue also affect `resolve_struct_type` and `resolve_func_type` ?

resolved

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -1799,6 +1799,11 @@ resolve_func_type(const uint8 **p_buf, const uint8 *buf_end, WASMModule *module,
         return false;
     }
     if (ref_type_map_count > 0) {
+        if (ref_type_map_count > UINT16_MAX) {
```
lum1n0us, line 1802: ~~`ref_type_map_count` is an `uint32` and it is always `>0`. Perhaps we should rewrite from L1801 to L1806 the code by removing `if (ref_type_map_count > 0)`~~

lum1n0us (approved):

xujuntwt95329 (approved):
LGTM

PR #4549, opened 2025-08-08
Fix potential overflow in memory size calculation

kylo5aby:
After resizing memory size, `num_bytes_per_page` might be greater than `DEFAULT_NUM_BYTES_PER_PAGE`, 
hence check for integer overflow before `num_bytes_per_page += heap_size`.

Here is the detail error log:
```shell
wasm-micro-runtime/core/iwasm/interpreter/wasm_runtime.c:339:32: runtime error: unsigned integer overflow: 4294836224 + 16777216 cannot be represented in type 'unsigned int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /home/zhenwei/project/wasm-micro-runtime/core/iwasm/interpreter/wasm_runtime.c:339:32
```

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -1026,14 +1026,14 @@ memory_instantiate(AOTModuleInstance *module_inst, AOTModuleInstance *parent,
         /* If only one page and at most one page, we just append
            the app heap to the end of linear memory, enlarge the
            num_bytes_per_page, and don't change the page count */
-        heap_offset = num_bytes_per_page;
-        num_bytes_per_page += heap_size;
-        if (num_bytes_per_page < heap_size) {
+        if (heap_size > UINT32_MAX - num_bytes_per_page) {
             set_error_buf(error_buf, error_buf_size,
                           "failed to insert app heap into linear memory, "
                           "try using `--heap-size=0` option");
             return NULL;
         }
+        heap_offset = num_bytes_per_page;
+        num_bytes_per_page += heap_size;
     }
     else if (heap_size > 0) {
         if (init_page_count == max_page_count && init_page_count == 0) {
--- a/core/iwasm/interpreter/wasm_runtime.c
+++ b/core/iwasm/interpreter/wasm_runtime.c
@@ -335,14 +335,14 @@ memory_instantiate(WASMModuleInstance *module_inst, WASMModuleInstance *parent,
             /* If only one page and at most one page, we just append
                the app heap to the end of linear memory, enlarge the
                num_bytes_per_page, and don't change the page count */
-            heap_offset = num_bytes_per_page;
-            num_bytes_per_page += heap_size;
-            if (num_bytes_per_page < heap_size) {
+            if (heap_size > UINT32_MAX - num_bytes_per_page) {
                 set_error_buf(error_buf, error_buf_size,
                               "failed to insert app heap into linear memory, "
                               "try using `--heap-size=0` option");
                 return NULL;
             }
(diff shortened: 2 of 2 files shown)


TianlongLiang (commented):
Need to make the same modification for `aot_runtime.c` too. And if this overflow can happen only when the SHRUNK_MEMORY is enabled, maybe consider to add the macro around the condition.

kylo5aby:
> Need to make the same modification for `aot_runtime.c` too. And if this overflow can happen only when the SHRUNK_MEMORY is enabled, maybe consider to add the macro around the condition.

resolved. refers to:https://github.com/bytecodealliance/wasm-micro-runtime/blob/6b51c61f5e50573f672916a7df953a2704b823f4/core/iwasm/interpreter/wasm_loader.c#L6559C1-L6560C61. this check happened when MULTI_MODULE is disabled or SHRUNK_MEMORY enabled

lum1n0us:
> UndefinedBehaviorSanitizer: undefined-behavior: unsigned integer overflow: 4294836224 + 16777216 cannot be represented in type 'unsigned int'

IIUC, this reveals a more common question: which is better, pre-additional-check or post-additional-check to fix a potential integer overflow? A pre-additional-check involves using a check to prevent integer overflow from the very beginning. A post-additional-check involves using a check after addition to see if there is an overflow. Clearly, UBSan strongly prefers the pre-additional-check, and I believe GCC compilation warnings do the same.

The issue is that post-additional-check is widely used in source code. Should we conduct a comprehensive check to replace post-additional-check with pre-additional-check? Should this approach be extended to all four integer arithmetic operations?

Please let me know if I am overthinking this.

@yamt @loganek @TianlongLiang

kylo5aby:
for C/C++ spec, unsigned integer overflow is well defined,  and overflow of signed integers invokes undefined behavior. so the above code will perform well in GCC compiler, So overflow of signed integers  pre-additional-check should be better, and for overflow of unsigned integers, UBSan suggests it's not the intended value.

lum1n0us:
FYI: from https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html

`-fsanitize=unsigned-integer-overflow`: Unsigned integer overflow, where the result of an unsigned integer computation cannot be represented in its type. **_Unlike signed integer overflow, this is not undefined behavior, but it is often unintentional_**. This sanitizer does not check for lossy implicit conversions performed before such a computation

yamt:
> > UndefinedBehaviorSanitizer: undefined-behavior: unsigned integer overflow: 4294836224 + 16777216 cannot be represented in type 'unsigned int'
> 
> IIUC, this reveals a more common question: which is better, pre-additional-check or post-additional-check to fix a potential integer overflow? A pre-additional-check involves using a check to prevent integer overflow from the very beginning. A post-additional-check involves using a check after addition to see if there is an overflow. Clearly, UBSan strongly prefers the pre-additional-check, and I believe GCC compilation warnings do the same.
> 
> The issue is that post-additional-check is widely used in source code. Should we conduct a comprehensive check to replace post-additional-check with pre-additional-check? Should this approach be extended to all four integer arithmetic operations?
> 
> Please let me know if I am overthinking this.
> 
> @yamt @loganek @TianlongLiang

my suggestions:

* do NOT enable -fsanitize=unsigned-integer-overflow because it isn't an undefined behavior.
* for performance sensitive logic, use __builtin_add_overflow etc. provide something like https://github.com/yamt/toywasm/blob/9a5622791e99395e26e6e96cef830af3d91a1685/lib/platform.h#L176-L191 and encourage the use of them.

TianlongLiang (approved):

PR #4553, opened 2025-08-12
Fix a sanitizer env not defined issue

kylo5aby:


build-scripts/config_common.cmake
```diff
@@ -149,7 +149,7 @@ if (NOT DEFINED WAMR_BUILD_SANITIZER)
   set(WAMR_BUILD_SANITIZER $ENV{WAMR_BUILD_SANITIZER})
```
yamt: isn't it set here?

kylo5aby: > isn't it set here?

Actually, if env `WAMR_BUILD_SANITIZER` is undefined, this equivalent to ` set(WAMR_BUILD_SANITIZER)`, which means `WAMR_BUILD_SANITIZER` being  undefined as well. Alternatively we can use ` set(WAMR_BUILD_SANITIZER "$ENV{WAMR_BUILD_SANITIZER}")` to make  `WAMR_BUILD_SANITIZER` = '' when env `WAMR_BUILD_SANITIZER` is undefined.

yamt: oh, i didn't know cmake set can unset variable.

yes, it seems simpler to quote `$ENV{}`.

yamt (approved):

lum1n0us (approved):

PR #4563, opened 2025-08-14
add validation of dynamic_offset

Septa2112:
In Fast-Interpreter mode, `wasm_loader_pop_frame_offset` does not validate `dynamic_offset` before performing calculations. This can lead to an potential unsigned integer underflow in certain cases, for example when `cell_num_to_pop` exceeds `dynamic_offset`:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/cbb6d033375a07966ec218863e6c18cb55574e3e/core/iwasm/interpreter/wasm_loader.c#L9869

This patch adds a validation step for `dynamic_offset` to prevent such cases and improve code robustness.

### DIFF
--- a/core/iwasm/interpreter/wasm_loader.c
+++ b/core/iwasm/interpreter/wasm_loader.c
@@ -8426,6 +8426,14 @@ check_offset_pop(WASMLoaderContext *ctx, uint32 cells)
     return true;
 }
 
+static bool
+check_dynamic_offset_pop(WASMLoaderContext *ctx, uint32 cells)
+{
+    if (ctx->dynamic_offset < cells)
+        return false;
+    return true;
+}
+
 static void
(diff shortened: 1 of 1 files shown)


kylo5aby:
should also update mini loader part

lum1n0us:
This issue concerns an unsigned integer underflow, such as `(3-4)`, which is captured by the sanitizer. If there is no sanitizer, a subsequent checker will detect it and report an exception. Therefore, the malformed Wasm won't cause any problems during the instantiation and execution phases.

And according to [Clang's Undefined Behavior Sanitizer documentation](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html), including unsigned integer underflow in sanitizer options is more like asking the developer if this is intentional.

Issue #4560, opened 2025-08-14
Proposal: Add Github Actions workflow to auto-close stale issues


Septa2112:
Currently, WAMR does not have an automated workflow to close stale issues. This can lead to backlog of old and inactive issues, making it harder to manage the repository.

I would like to add a Github Actions workflow that automatically marks issues as stale after a certain period of inactivity (e.g., 60 days) and closes them if no further updates are made (e.g., after another 7 days).

### Benefits
- Keep the issue list focused on active and relevant topics
- Reduces the manual workload for maintainers.
- Encourages contributors to follow up on issues they care about.

### Additional Notes
- The time limits and exempted labels can be discussed.
- This workflow would only target issues.

### Possible implementation
We could use the [actions/stale](https://github.com/actions/stale) GitHub Action.

lum1n0us:
One additional requirement is that if an issue is labeled with *help wanted*, do not close it.

srberard:
I support efforts to improve project hygene.  As pointed out in other threads, there are over 500 issues in the repository.  I did some quick analysis of these and there are many that have not been updated in over a year (approximately 350).  In my opinion, this makes it much harder to us to understand what really needs attention.  

In principle, I support the approach of having things auto closed if they are untouched for a period of time.  I've seen this work well in other projects.  I would say any issue that's not been updated in 6 months or a year should be marked as stale.  I think giving folks 30 days or say before closing is reasonable.

That said, we've not had this in place yet so I think just enabling this now would not be the best idea.  Instead, I suggest that we do a review of the current set of open issues first.  Once we've done that, we can enable an automatic rule as described here.

PR #4565, opened 2025-08-18
add Github workflow: auto_close_issue

Septa2112:
Refer to: https://github.com/bytecodealliance/wasm-micro-runtime/issues/4560

### DIFF
--- a/.github/workflows/auto_close_issue.yml
+++ b/.github/workflows/auto_close_issue.yml
@@ -12,6 +12,7 @@ jobs:
     steps:
       - uses: actions/stale@v9
         with:
+          debug-only: true
           days-before-issue-stale: 30
           days-before-issue-close: 60
           stale-issue-label: "stale"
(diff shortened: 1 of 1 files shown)


.github/workflows/auto_close_issue.yml
```diff
@@ -0,0 +1,24 @@
...
+        with:
+          debug-only: true
+          days-before-issue-stale: 30
+          days-before-issue-close: 60
+          stale-issue-label: "stale"
+          stale-issue-message: "This issue is stale because it has been open for 30 days with no activity."
+          close-issue-message: "This issue was closed because it has been inactive for 60 days since being marked as stale."
+          days-before-pr-stale: -1
+          days-before-pr-close: -1
+          exempt-labels: "help wanted, good first issue, bug, enhancement"
```
lum1n0us, line 23: In my opinion, 

```suggestion
          exempt-labels: "help wanted"
```

.github/workflows/auto_close_issue.yml
```diff
@@ -0,0 +1,24 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v9
+        with:
+          debug-only: true
+          days-before-issue-stale: 30
+          days-before-issue-close: 60
```
yamt, line 17: if we want to do this, i'd suggest to use a longer threshold. eg. a few years.

i don't think we really need automation like this though. right now we only have ~500 open issues.

PR #4571, opened 2025-08-19
Add micro `WAMR_BUILD_LIME1` to enable minimal lime1 feature set

kylo5aby:
Currently, WAMR already supports a Lime1 feature super set. `WAMR_BUILD_LIME1` includes a minimal feature set for Lime1, making it more lightweight if only Lime1 features is needed


core/iwasm/compilation/aot_compiler.c
```diff
@@ -2532,6 +2532,8 @@ aot_compile_func(AOTCompContext *comp_ctx, uint32 func_index)
                             return false;
                         break;
                     }
+#endif /* WASM_ENABLE_BULK_MEMORY */
+#if WASM_ENABLE_BULK_MEMORY != 0 || WASM_ENABLE_LIME1 != 0
```
lum1n0us: IMU, `WASM_ENABLE_LIME1` is the name of a combination of flags. It should not be used directly in the source code.

build-scripts/config_common.cmake
```diff
@@ -251,6 +251,17 @@ if (NOT DEFINED WAMR_BUILD_EXTENDED_CONST_EXPR)
   set (WAMR_BUILD_EXTENDED_CONST_EXPR 0)
 endif ()
 
+if (NOT DEFINED WAMR_BUILD_LIME1)
+  set (WAMR_BUILD_LIME1 0)
+endif ()
```
lum1n0us, line 266: IMU, WASM_ENABLE_LIME1 is the name of a combination of flags. It should not be used directly in the source code. No need to define `WAMR_BUILD_LIME1`

build-scripts/config_common.cmake
```diff
@@ -251,6 +251,17 @@ if (NOT DEFINED WAMR_BUILD_EXTENDED_CONST_EXPR)
...
+
+if (WAMR_BUILD_LIME1 EQUAL 1)
+  set (WAMR_BUILD_EXTENDED_CONST_EXPR 1)
+  set (WAMR_BUILD_BULK_MEMORY 0)
+  set (WAMR_BUILD_SIMD 0)
+  set (WAMR_BUILD_REF_TYPES 0)
+endif ()
```
lum1n0us, line 276: A challenging question is how to handle *bulk-memory-opt* and *call-indirect-overlong*. IMU, an ideal approach would be to have all flags to match [all command line options of clang](https://clang.llvm.org/docs/ClangCommandLineReference.html#webassembly).

kylo5aby: > A challenging question is how to handle _bulk-memory-opt_ and _call-indirect-overlong_. IMU, an ideal approach would be to have all flags to match [all command line options of clang](https://clang.llvm.org/docs/ClangCommandLineReference.html#webassembly).

therefore, `WAMR_BUILD_BULK_MEMORY_OPT` and `WAMR_BUILD_CALL_INDIRECT_OVERLONG` compilation flags are needed. These represent two individual WASM features, similar to `WAMR_BUILD_EXTENDED_CONST`.

when enabling Lime1, all related WASM features must also be enabled in cmake (unrelated features should be disabled).

after this is done, each individual clang WASM feature  can be controlled independently, allowing for a minimal Lime1 feature set to be achieved.

yamt: i agree it's better to have separate options for individual features.

no1wudi: > i agree it's better to have separate options for individual features.

Agree too

TianlongLiang: Agree too

build-scripts/config_common.cmake
```diff
@@ -251,10 +259,30 @@ if (NOT DEFINED WAMR_BUILD_EXTENDED_CONST_EXPR)
...
 # Compilation options to marco
 ########################################
 
+if (WAMR_BUILD_LIME1 EQUAL 1)
+  set (WAMR_BUILD_BULK_MEMORY 0)
+  set (WAMR_BUILD_BULK_MEMORY_OPT 1)
+  set (WAMR_BUILD_REF_TYPES 0)
+  set (WAMR_BUILD_CALL_INDIRECT_OVERLONG 1)
+  set (WAMR_BUILD_EXTENDED_CONST_EXPR 1)
+  set (WAMR_BUILD_SIMD 0)
```
yamt: i guess WAMR_BUILD_LIME1 should not override something explicitly specified by user.
eg. `-DWAMR_BUILD_SIMD=1 -DWAMR_BUILD_LIME1=1` should not disable simd.
what do you think?

kylo5aby: > i guess WAMR_BUILD_LIME1 should not override something explicitly specified by user. eg. `-DWAMR_BUILD_SIMD=1 -DWAMR_BUILD_LIME1=1` should not disable simd. what do you think?

I think `WAMR_BUILD_LIME1` is intended to include a minimal lime1 feature set.  `relaxed simd` feature is not part of lime1, but it's enabled in WAMR by default. From the perspective of the `Lime1` feature set it should only open wasm1.0 plus some other features(e.g. bulk-memory-opt) when  `WAMR_BUILD_LIME1` flag is enabled. so imo, if `WAMR_BUILD_LIME1` flag is enabled, those default features should be disabled. cc @lum1n0us @TianlongLiang

build-scripts/config_common.cmake
```diff
@@ -219,6 +219,14 @@ if (NOT DEFINED WAMR_BUILD_BULK_MEMORY)
...
 endif ()
 
+if (NOT DEFINED WAMR_BUILD_BULK_MEMORY_OPT)
+  set (WAMR_BUILD_BULK_MEMORY_OPT 0)
+endif ()
+
+if (NOT DEFINED WAMR_BUILD_CALL_INDIRECT_OVERLONG)
+  set (WAMR_BUILD_CALL_INDIRECT_OVERLONG 0)
+endif ()
+
```
lum1n0us, line 231: Add introduction about both into *doc/build_wamr.md*

core/iwasm/compilation/aot_llvm.c
```diff
@@ -2724,6 +2724,9 @@ aot_create_comp_context(const AOTCompData *comp_data, aot_comp_option_t option)
     if (option->enable_bulk_memory)
         comp_ctx->enable_bulk_memory = true;
```
lum1n0us, line 2732: + `comp_ctx->enable_bulk_memory_opt = true;`?

core/iwasm/compilation/aot_llvm.c
```diff
@@ -2733,6 +2736,9 @@ aot_create_comp_context(const AOTCompData *comp_data, aot_comp_option_t option)
     if (option->enable_ref_types)
         comp_ctx->enable_ref_types = true;
```
lum1n0us, line 2744: + `comp_ctx->enable_call_indirect_overlong = true;`?

core/iwasm/compilation/aot_llvm.h
```diff
@@ -413,6 +413,9 @@ typedef struct AOTCompContext {
     /* Bulk memory feature */
     bool enable_bulk_memory;
 
+    /* Bulk memory opt feature */
```
lum1n0us:
```suggestion
    /* Bulk memory opt feature. will be enabled alongside the enable_bulk_memory*/
```

core/iwasm/compilation/aot_llvm.h
```diff
@@ -449,6 +452,9 @@ typedef struct AOTCompContext {
     /* Reference Types */
     bool enable_ref_types;
 
+    /* Call Indirect Overlong */
```
lum1n0us:
```suggestion
    /* Call Indirect Overlong. will be enabled alongside the enable_ref_types */
```

build-scripts/config_common.cmake
yamt: `WAMR_BUILD_xxx` usually just enables something necessary to support `xxx`.
to support lime1-targeted modules, we don't need to disable simd etc.

if you want stronger semantics, i guess it's better to give it a more obvious name.
eg. `WAMR_ONLY_MINIMUM_FEATURES_FOR_LIME1`

(btw, iirc WAMR_BUILD_SIMD does not implement relaxed-simd.)

kylo5aby: this seems reasonable. So when `WAMR_BUILD_LIME1 is enabled`,  WAMR will support Lime1-targeted modules. If users wish to support only the minimal Lime1 feature set, they can simply disable e.g. WAMR_BUILD_REF_TYPES and WAMR_BUILD_SIMD, which are enabled by default.

core/iwasm/compilation/aot_llvm.c
kylo5aby: > * `comp_ctx->enable_bulk_memory_opt = true;`?

if `option->enable_bulk_memory == true`, `option->enable_bulk_memory_opt` will also be true,  and `comp_ctx->enable_bulk_memory_opt` depends on `option->enable_bulk_memory_opt`.

core/iwasm/compilation/aot_llvm.c
kylo5aby: > * `comp_ctx->enable_call_indirect_overlong = true;`?

same as above

build-scripts/config_common.cmake
kylo5aby: resolved. thanks

lum1n0us (commented):
- compilation cases in CI.
- how to build a minimum building about lime1 in doc/build_wamr.md

kylo5aby:
> * compilation cases in CI.
> * how to build a minimum building about lime1 in doc/build_wamr.md

resolved.

lum1n0us (approved):

core/iwasm/compilation/aot_llvm.h
```diff
@@ -464,6 +471,9 @@ typedef struct AOTCompContext {
     /* Enable extended constant expression */
     bool enable_extended_const;
 
+    /* Enable Lime1 */
+    bool enable_lime1;
```
yamt: this field doesn't seem necessary.
an auto variable in wamr-compiler is enough.

lum1n0us:
@kylo5aby Please resolve conflicts.

kylo5aby:
> @kylo5aby Please resolve conflicts.

resolved.

lum1n0us:
wait for #4620

PR #4573, opened 2025-08-20
Add Android macro check for indirect mode

jammar1:
I had an issue running binaries built with XIP/indirect mode on Android because this check was missing. They run successfully after this.


loganek (approved):
LGTM

PR #4582, opened 2025-08-26
fix: cast file descriptor to size_t for comparison in fd_prestats

lum1n0us:
``` shell
comparison of integer expressions of different signedness:
  ‘__wasi_fd_t’ {aka ‘int’} and ‘size_t’ {aka ‘long unsigned int’}
  [-Werror=sign-compare]
  288 |     if (fd >= pt->size)
```


TianlongLiang (approved):

yamt (approved):
maybe int is enough for pt->size?

lum1n0us:
It could be, but it appears that it needs more changes.

PR #4588, opened 2025-08-29
Fix compilation error when MSVC is not set to C11 standard

vkensou:


core/shared/platform/include/platform_wasi_types.h
```diff
@@ -29,6 +29,9 @@
 #endif /* _Alignof */
 
 extern "C" {
+#elif defined(_MSC_VER) && !_CRT_HAS_C11
+#define _Static_assert static_assert
+#define _Alignof __alignof
```
lum1n0us, line 34: IMM, how about?

```c
#ifdef __cplusplus
extern "C" {
#endif

#ifndef _Static_assert
#define _Static_assert static_assert
#endif /* _Static_assert */

#ifndef _Alignof
#if defined(_MSC_VER) && !_CRT_HAS_C11
#define _Alignof __alignof
#else
#define _Alignof alignof
#endif
#endif /* _Alignof */

```

vkensou: ok

vkensou: This make a lot error. So I revert it.

lum1n0us: May I ask about the error? the modification I apologize for the confusion. My intention is to expand the compatibility of `_Static_assert` and `_Align_of`.

As far as I know, `if __cplusplus >= 201103L`, `static_assert` is a keyword. `if __STDC_VERSION__ >= 201112L`, `_Static_assert` is a keyword. Both `__GNUC__` and `__clang__` support `_Static_assert` since version 4.6, and `if _MSC_VER >= 1600` supports `_Static_assert` for both C++ and C.

Therefore, we might need to reorganize to cover all these cases.

vkensou: errors are here: https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/17377924957/job/49328417065

I am not familiar with gcc/clang.  Seems gcc and clang is ok when use c99. Only msvc must use c11.

sbwtw: I tested this commit on VS2022 and it's works.

lum1n0us (approved):

PR #4603, opened 2025-09-05
Fix compile warning -Wdouble-promotion on some platform

TianlongLiang:


yamt:
* doesn't `-Wdouble-promotion` warn on `return NAN` in f64_min/f64_max?
* do you plan to enable `-Wdouble-promotion` by default?

TianlongLiang:
> * doesn't `-Wdouble-promotion` warn on `return NAN` in f64_min/f64_max?

You are right about NAN., I plan to explicitly convert it like `return (double)NAN;`

> * do you plan to enable `-Wdouble-promotion` by default?

Yes, I plan to enable -Wdouble-promotion by default, I think it can benefit some embedded platforms.

yamt:
> > * doesn't `-Wdouble-promotion` warn on `return NAN` in f64_min/f64_max?
> 
> You are right about NAN., I plan to explicitly convert it like `return (double)NAN;`
> 
> > * do you plan to enable `-Wdouble-promotion` by default?
> 
> Yes, I plan to enable -Wdouble-promotion by default, I think it can benefit some embedded platforms.

ok. it makes sense.

core/iwasm/aot/aot_intrinsic.c
```diff
@@ -140,10 +140,10 @@ aot_intrinsic_copysign_f64(float64 a, float64 b)
 float32
 aot_intrinsic_fmin_f32(float32 a, float32 b)
 {
-    if (isnan(a) || isnan(b))
+    if (isnanf(a) || isnanf(b))
```
yamt: for which platforms is this necessary?
isnan macro usually works for both of double and float.

TianlongLiang: I am working on the Zephyr platform, but the same would be true for Alios, Riot, and Rt-thread. Given that they all use `core/shared/platform/common/math/math.c`. They are not macro in those platforms but function

yamt: > I am working on the Zephyr platform, but the same would be true for Alios, Riot, and Rt-thread. Given that they all use core/shared/platform/common/math/math.c. They are not macro in those platforms but function

i guess it should use a macro.

eg.
```
#define isnan(x) (sizeof(x) == sizeof(double) ? isnan_double(x) : isnan_float(x))
```

TianlongLiang: Good idea, I'm also considering this approach, in this way, it can greatly reduce LOC

TianlongLiang: It seems that we can't simply use a macro. In Ubuntu, macro is expanded to intrincs like `__builtin_isnan`, if we were to use `#define isnan(x) (sizeof(x) == sizeof(double) ? isnan_double(x) : isnan_float(x))` it will output warning anyway(even though we won't be actually do double promotion)

yamt: i guess you can cast on double side
```c
#define isnan(x) (sizeof(x) == sizeof(double) ? isnan_double((double)x) : isnan_float(x))
```

build-scripts/runtime_lib.cmake
```diff
@@ -163,7 +163,7 @@ endif ()
 ####################### Common sources #######################
 if (NOT MSVC)
     set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu99 -ffunction-sections -fdata-sections \
-                                         -Wall -Wno-unused-parameter -Wno-pedantic")
+                                         -Wall -Wdouble-promotion -Wno-unused-parameter -Wno-pedantic")
```
lum1n0us: After merging #4397, it should be in build-scripts/warnings.cmake

build-scripts/warnings.cmake
```diff
@@ -20,6 +20,9 @@ else ()
     $<$<COMPILE_LANGUAGE:C>:-Wincompatible-pointer-types>
     $<$<COMPILE_LANGUAGE:C>:-Wimplicit-function-declaration>
   )
+  add_compile_options (
```
yamt, line 23: why don't you combine this to the above one?

TianlongLiang: I plan to put warning flags that are mostly meant for benefit embedded systems mostly here. probably should add a comment here

PR #4604, opened 2025-09-05
Fix the source_all override in platform_api_math.cmake for alios, riot, rt-thread, and zephyr platforms

TianlongLiang:
Prevent source_all from being overwritten by platform_api_math.cmake


lum1n0us (approved):

lum1n0us:
@srberard Please be aware of this.

TianlongLiang:
I found out that other platforms also have this problem, and it's easier to rename the variable in platform_api_math.cmake.

yamt (approved):

Issue #4610, opened 2025-09-10
wasm_runtime_validate_app_str_addr is easy to misuse


yamt:
wasm_runtime_validate_app_str_addr (and "$" in signature) is easy to use in an insecure way.
eg. sock_addr_resolve mentioned in https://github.com/bytecodealliance/wasm-micro-runtime/issues/4463 is an example of insecure usage.

it depends on the contents of linear memory (nul termination) which is owned by user app.
with threads, the user app can modify the buffer contents immediately after the native function
checked the buffer with wasm_runtime_validate_app_str_addr.
even without threads, depending on the functionality the native function provides, it might be possible to trick the native function itself to modify the buffer.

my suggestion is to make wasm_runtime_validate_app_str_addr return the length of the validated string.
and probably introduce "$~", which yields two arguments to the native function.

Issue #4463, opened 2025-07-08
wasi: socket api seems to lack pointer validations


yamt:
* sock_recv/sock_recv_from/sock_send/sock_send_to doesn't seem to validate iovec
* sock_addr_resolve doesn't seem to validate addr_info
* sock_bind doesn't seem to validate addr
* maybe we should audit all socket apis.

yamt:
the following functions seem to lack necessary address validation.

sock_accept
sock_addr_resolve
sock_bind
sock_connect
sock_open
sock_recv_from
sock_recv
sock_send
sock_send_to

PR #4611, opened 2025-09-10
libc-wasi: add missing pointer validations to socket functions

yamt:
cf. https://github.com/bytecodealliance/wasm-micro-runtime/issues/4463

the fix for sock_addr_resolve is incomplete.
cf. https://github.com/bytecodealliance/wasm-micro-runtime/issues/4610


yamt:
> While there's a possibility of being out of the sandbox, most pointers target local variables in guest language functions.

i'm not sure what you mean.

yamt:
> > While there's a possibility of being out of the sandbox, most pointers target local variables in guest language functions.
> 
> i'm not sure what you mean.

malicious (or, innocent-but-buggy) guest programs can access host memory past the end of its linear memory.
this PR fixes some of the attack surfaces.

lum1n0us:
> malicious (or, innocent-but-buggy) guest programs can access host memory past the end of its linear memory.

Can't argue with that. 😆

PR #4613, opened 2025-09-11
wasi-nn: retire is_model_loaded flag

yamt:
this flag doesn't make much sense anymore because:
- backends validate given graph/ctx by themselves
- some of them support loading multiple models for a context

### DIFF
--- a/core/iwasm/libraries/wasi-nn/src/wasi_nn.c
+++ b/core/iwasm/libraries/wasi-nn/src/wasi_nn.c
@@ -66,7 +66,6 @@ wasi_nn_ctx_destroy(WASINNContext *wasi_nn_ctx)
     }
     NN_DBG_PRINTF("[WASI NN] DEINIT...");
     NN_DBG_PRINTF("Freeing wasi-nn");
-    NN_DBG_PRINTF("-> is_model_loaded: %d", wasi_nn_ctx->is_model_loaded);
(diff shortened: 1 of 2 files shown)


yamt:
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/17630748121/job/50097638718?pr=4613

lum1n0us:
Then, what about the scenario where `wasi_nn_init_execution_context()`, `wasi_nn_set_input()`, or other functions are called without first using `wasi_nn_load_by_name()`?

yamt:
> Then, what about the scenario where `wasi_nn_init_execution_context()`, `wasi_nn_set_input()`, or other functions are called without first using `wasi_nn_load_by_name()`?

there is no way to get a valid graph or execution-context without load.

lum1n0us:
> there is no way to get a valid graph or execution-context without load.

I mean do we need to prevent this scenario? Or just let the execution failed?

yamt:
> > there is no way to get a valid graph or execution-context without load.
> 
> I mean do we need to prevent this scenario? Or just let the execution failed?

eg. init_execution_context takes a graph. it just fails unless the user gives it a valid graph.
i don't see any need to have extra logic to prevent it.

lum1n0us:
IIUC, `is_model_initialized()` is an unnecessary check. The subsequent execution will naturally fail unless a valid graph is provided.

PR #4619, opened 2025-09-11
Ensure --addr-pool mask accepts numbers only

linear0211:
The `--addr-pool` option previously used `atoi` to parse the mask value.  
If non-numeric characters were included, `atoi` returned `0`, causing the pool to unintentionally accept all IP addresses.

This change ensures the mask accepts only numeric values, preventing unexpected behavior.

### DIFF
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -3812,7 +3812,7 @@ wasm_runtime_init_wasi(WASMModuleInstanceCommon *module_inst,
 
     /* addr_pool(textual) -> apool */
     for (i = 0; i < addr_pool_size; i++) {
-        char *cp, *address, *mask, *endptr;
+        char *cp, *address, *mask, *nextptr, *endptr;
         long mask_val;
         bool ret = false;
 
@@ -3823,14 +3823,20 @@ wasm_runtime_init_wasi(WASMModuleInstanceCommon *module_inst,
             goto fail;
         }
 
-        address = strtok(cp, "/");
-        mask = strtok(NULL, "/");
+#ifdef BH_PLATFORM_WINDOWS
+        address = strtok_s(cp, "/", &nextptr);
+        mask = strtok_s(NULL, "/", &nextptr);
+#else
+        address = strtok_r(cp, "/", &nextptr);
+        mask = strtok_r(NULL, "/", &nextptr);
+#endif
 
         if (!mask) {
             snprintf(error_buf, error_buf_size,
                      "Invalid address pool entry: %s, must be in the format of "
                      "ADDRESS/MASK",
                      addr_pool[i]);
+            wasm_runtime_free(cp);
             goto fail;
         }
 
@@ -3840,11 +3846,13 @@ wasm_runtime_init_wasi(WASMModuleInstanceCommon *module_inst,
         if (mask == endptr || *endptr != '\0') {
             snprintf(error_buf, error_buf_size,
                      "Invalid address pool entry: mask must be a number");
+            wasm_runtime_free(cp);
             goto fail;
         }
         if (errno != 0 || mask_val < 0) {
             snprintf(error_buf, error_buf_size,
                      "Init wasi environment failed: invalid mask number");
+            wasm_runtime_free(cp);
             goto fail;
         }
 


core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3833,7 +3834,15 @@ wasm_runtime_init_wasi(WASMModuleInstanceCommon *module_inst,
             goto fail;
         }
 
-        ret = addr_pool_insert(apool, address, (uint8)atoi(mask));
+        mask_val = strtol(mask, &endptr, 10);
```
lum1n0us, line 3844: IMU, `strtol()` can handle "a100" and "0x100", but how does it deal with "+100" and "-100"?

yamt: * a stricter way to check strtoxxx result is something like https://github.com/yamt/toywasm/blob/36e1f4aa9b390474da09f0e1b831eed702f67fd0/cli/str_to_uint.c#L11-L21
* i guess you can make addr_pool_insert reject masks inappropriate for the address family.

i feel addr_pool_insert is a bit awkward api as it takes `addr` as a string but `mask` as an integer. but it's another story.

linear0211: According to the [Linux man page](https://man7.org/linux/man-pages/man3/strtol.3.html):

> The string may begin with an arbitrary amount of white space (as determined by [isspace(3)](https://man7.org/linux/man-pages/man3/isspace.3.html)) followed by a single optional '+' or '-' sign.

So, `+100` and `-100` are parsed as positive 100 and negative 100.

I’ll also update the code to add stricter `strtol` checks and enforce mask range validation in `addr_pool_insert`.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3833,7 +3834,21 @@ wasm_runtime_init_wasi(WASMModuleInstanceCommon *module_inst,
...
-        ret = addr_pool_insert(apool, address, (uint8)atoi(mask));
+        errno = 0;
+        mask_val = strtol(mask, &endptr, 10);
+
+        if (mask == endptr || *endptr != '\0') {
+            snprintf(error_buf, error_buf_size,
+                     "Invalid address pool entry: mask must be a number");
+            goto fail;
+        }
+        if (errno != 0 || mask_val < 0 || mask_val > 128) {
```
lum1n0us:
```suggestion
        if (errno != 0 || mask_val < 0) {
```

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -3116,10 +3116,18 @@ addr_pool_insert(struct addr_pool *addr_pool, const char *addr, uint8 mask)
         next->type = IPv6;
         bh_memcpy_s(next->addr.ip6, sizeof(next->addr.ip6), target.ipv6,
                     sizeof(target.ipv6));
+        if (mask > 128) {
+            wasm_runtime_free(next);
+            return false;
+        }
```
lum1n0us, line 3121: remove L3108. Add `next->mask=mask` here.

core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
```diff
@@ -3116,10 +3116,18 @@ addr_pool_insert(struct addr_pool *addr_pool, const char *addr, uint8 mask)
...
     }
     else {
         next->type = IPv4;
         next->addr.ip4 = target.ipv4;
+        if (mask > 32) {
+            wasm_runtime_free(next);
+            return false;
+        }
```
lum1n0us, line 3130: remove L3108. Add `next->mask=mask` here.

lum1n0us (dismissed):
LGTM. With minor comments

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -3822,18 +3823,40 @@ wasm_runtime_init_wasi(WASMModuleInstanceCommon *module_inst,
...
 
-        address = strtok(cp, "/");
-        mask = strtok(NULL, "/");
+#ifdef BH_PLATFORM_WINDOWS
+        address = strtok_s(cp, "/", &nextptr);
+        mask = strtok_s(NULL, "/", &nextptr);
+#else
+        address = strtok_r(cp, "/", &nextptr);
+        mask = strtok_r(NULL, "/", &nextptr);
+#endif
```
yamt, line 3832: maybe we can have bh_strtok_r function for this?

lum1n0us: Agreed. However, I think we should use a new PR to implement that and merge it first. Does that sound good to you?

linear0211: Sounds good to me. Should I open a new PR for the implementation?

yamt: either ways are fine for me

lum1n0us: File https://github.com/bytecodealliance/wasm-micro-runtime/issues/4766 to track

PR #4621, opened 2025-09-12
Port gitbook document to main

TianlongLiang:
And remove some invalid app framework-related document links


lum1n0us (commented):
Not sure if we should enhance the GitBook in this PR or in future PRs. For example:
- There are over 25 samples. Should we introduce them?
- Few new features should be included.
- Some description should be sync up with latest code.

TianlongLiang:
I prefer to do it in future PRs. I will create an issue to list the enhancement tasks

lum1n0us (approved):

TianlongLiang:
@no1wudi @yamt would you agree to port first?

Issue #4561, opened 2025-08-14
Should we add a watchdog to prevent the similar crash in AOT mode?


Septa2112:
Refer to issue #4539 and PR #4559.

If we pre-link all import items to a watchdog function from the start, the watchdog function will report an error once executed, and follow the current path during loading to resolve the import section content. Therefore,

a) all import functions are executable and not NULL

b) and, unlinked import functions will report an error without additional checks during execution.

lum1n0us:
FYI:
- According to specification requirements, if the import section cannot be fully satisfied during the instantiation phase, the runtime should report an instantiation failure and halt further execution.
- Denials of service during Wasm execution are considered security vulnerabilities. This might be debatable, but it is a rule followed by Wasmtime and BA. While not necessarily something we must adhere to, it is a perspective worth considering.

PR #4625, opened 2025-09-15
Add a watchdog to prevent the similar crash in AOT mode

Septa2112:
Refer to #4561


core/iwasm/aot/aot_runtime.c
```diff
@@ -114,6 +114,12 @@ set_error_buf_v(char *error_buf, uint32 error_buf_size, const char *format, ...)
     }
 }
 
+static void
+aot_unlinked_import_func_trap(void)
+{
+    printf("Unlinked import function called\n");
```
lum1n0us: Instead of `printf()`, Code should raise an exception and allow the runtime to exit gracefully.

PR #4626, opened 2025-09-15
bump wamr version to 2.4.2

lum1n0us:


lum1n0us:
https://github.com/bytecodealliance/wasm-micro-runtime/actions/workflows/compilation_on_windows.yml

And

```
Test path_symlink_trailing_slashes failed
  [exit_code] 0 == 1
STDOUT:
Exception: unreachable

STDERR:
thread 'main' panicked at 'expected errno NOENT; got NOTDIR', src/bin/path_symlink_trailing_slashes.rs:7:9
```

It is not an issue caused by WAMR. Let's disregard it and proceed with the release.

Issue #4628, opened 2025-09-15
zephyr: The `simple-file` and `simple-http` samples don't compile.


lucasAbadFr:
### Subject of the issue

The `simple-file` and `simple-http` Zephyr samples fail to compile on current `main`.  
Both samples were originally introduced to demonstrate WASI libc support on Zephyr, but they are now broken due to missing includes and undeclared functions in the Zephyr platform layer.

### Your environment

* Host OS: Ubuntu 22.04
* WAMR: current `main`
* Zephyr: v3.7.0 (LTS)
* Platform: Zephyr
* CPU arch: ARM Cortex-M (tested on `frdm_mcxn947/mcxn947`)

### Steps to reproduce
Simply follow the sample readme.

### Expected behavior
Both samples should compile and run.

### Actual behavior
Compilation fails with errors such as:
```bash
error: invalid use of incomplete typedef 'os_timespec' {aka 'struct timespec'}
error: storage size of 'ts' isn't known
warning: implicit declaration of function 'nanosleep' [-Wimplicit-function-declaration]
error: storage size of 'ts' isn't known
```

### Extra info 
The root cause is that `os_timespec` is only forward-declared as `struct timespec`, which is not defined in Zephyr unless `CONFIG_POSIX_API=y`.
However, the Zephyr port of WAMR is intended to work without enabling the POSIX API.
As a result, the samples cannot currently build against Zephyr v3.7 LTS.

I plan to submit a PR to fix this issue.

anasimran101:
@lucasAbadFr
so whats the status? I have been trying to build this sample-file and test libc-wasi on zephyr. I am using zephyr v4.2.0 


building the sample simple with revision to main branch gives:

/zephyr/zephyr-base/modules/lib/wasm-micro-runtime/core/shared/platform/zephyr/platform_internal.h:253:26: error: field 'file' has incomplete type
  253 |         struct fs_file_t file;

WAMR_BUILD_LIBC_WASI is set to 0 or 1 doesnt matter.

Build is a success when i use WAMR-2.4.x with WAMR_BUILD_LIBC_WASI set to 0. 
if i try to set it to WAMR_BUILD_LIBC_WASI to 1, it spits out a long list of build error top one being:


/zephyr/zephyr-base/modules/lib/wasm-micro-runtime/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/blocking_op.h:60:65: error: unknown type name 'nfds_t'
[...]

PR #4629, opened 2025-09-15
Zephyr: fix the product mini compilation issue

lucasAbadFr:
**This pull request present a fix to #4628.**

It introduces the following changes:
- Add a new `os_nanosleep` API.
- Provide a naive but documented Zephyr implementation of `os_nanosleep`.
- Add documentation for previously introduced `os_` APIs in `platform_api_extension.h`.
- Fix build issues so that the `simple-file` and `simple-http` samples compile again.

>[!NOTE]
> At first I intended to change `os_ioctl` return type but didn't in the end.


core/shared/platform/common/posix/posix_sleep.c
```diff
@@ -6,6 +6,7 @@
 #include <time.h>
 
 #include "platform_api_extension.h"
+#include "libc_errno.h"
```
lum1n0us: It is not a standard libc header and leads to a compilation error:

```
fatal error: libc_errno.h: No such file or directory
```

lum1n0us (changes requested):

lum1n0us:
If it's alright, could I inquire about your experience with: How to set up CI on a Zephyr platform using GitHub Actions? Since there is no board, a simulation is required. So, which is better: nsim or QEMU?

core/shared/platform/common/posix/posix_sleep.c
lucasAbadFr: Yes I know, but it didn't anticipate this error.
In fact I tested building iwasm for Linux (file sample) and no other plateforms (except Zephyr obviously).

From what I see, the issue comes from this CMake logic:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/95f506a6e77d3ac7588eac7263f95558edfa7f3b/core/shared/platform/common/posix/platform_api_posix.cmake#L15-L22

Now that `posix_sleep.c` directly include `libc_errno.h`, we can build with `WAMR_BUILD_LIBC_WASI=0` but we still need this helper.

How do we manage this case ? We can:
* Use a `case` in `os_nanosleep` on `ret` for POSIX.
* Change the CMake logic (but I can't cleary indentify the impact)

lucasAbadFr:
I @lum1n0us ! Thank you for the quick review.

Honestly, I don't have any real experiences for setting a CI on the Zephyr platform using Github Actions. But I may share my insights about native sim vs QEMU.

From my understanding about simulation tools:
* `native_sim` / `native_posix`: They should be used for "smoke tests" because they build and run very quickly and are executed as a Linux process. But they don't fully replicate the Hardware (I don't think it is really a problem for WAMR). They are great for functional and build validation. IMO we should build and run the samples. Also I identified that the samples won't work as is but we need to modify the `WAMR_BUILD_TARGET`. We might have a specific CI configuration suited to the runner.

* `QEMU`: It emulates real hardware, so it is closer to actual devices. From what I see it is already supported in Zephyr see the revelant [documentation](https://docs.zephyrproject.org/latest/boards/qemu/index.html). It will be an heavier setup and maintenance effort (I think). The builds and runs will mechanicaly be slower.

* `Renode`: I find this project facinating but didn't really use it yet. It's a simulator that emulates boards and peripherals. The complexity is even heavier than QEMU but it offers detailled hardware modeling. It's probably overkill as zephyr already test there board [here](https://zephyr-dashboard.renode.io/).
 
Also the WAMR project could select a few (embedded) Hardware platforms to support for each of the (RTOS) paltforms supported. It could be aligned with the [embedded SIG](https://github.com/bytecodealliance/sig-embedded/tree/main) targeted (hardware) platforms.

But in any case at least trying to compile and run the samples should be pretty simple to put in place, and we should begin by that.

lucasAbadFr:
Hello @lum1n0us , 

I provided a fix, only macOS is failing with: 

```bash
Error: can not get prebuilt llvm libraries
Error: Process completed with exit code 1.
```

It seem related to #4830

Issue #4632, opened 2025-09-16
`path_symlink_trailing_slashes` is failing on windows


yamt:
eg.
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/17752229385/job/50449098314?pr=4606
```
Test path_symlink_trailing_slashes failed
  [exit_code] 0 == 1
STDOUT:
Exception: unreachable

STDERR:
thread 'main' panicked at 'expected errno NOENT; got NOTDIR', src/bin/path_symlink_trailing_slashes.rs:7:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

yamt:
maybe it's caused by the runner image upgrade? (windows-2022 -> windows-2025)

PR #4633, opened 2025-09-16
CI: use windows-2022 image for now

yamt:
github is currently rolling out windows-2025 image.
for some reasons, the "path_symlink_trailing_slashes" test case in
wasi testsuite fails on windows-2025 image.
someone familar with windows need to investigate what was the
key difference between 2022 and 2025.
until that happens, this commit makes our CI use windows-2022 image.

cf.
https://github.com/bytecodealliance/wasm-micro-runtime/issues/4632
https://github.com/actions/runner-images/issues/12677

### DIFF
--- a/.github/workflows/build_iwasm_release.yml
+++ b/.github/workflows/build_iwasm_release.yml
@@ -151,7 +151,7 @@ jobs:
         working-directory: ${{ inputs.cwd }}
 
       - name: Compress the binary on Windows
-        if: inputs.runner == 'windows-latest'
+        if: inputs.runner == 'windows-2022'
         run: |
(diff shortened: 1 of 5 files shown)


yamt:
https://github.com/bytecodealliance/wasm-micro-runtime/actions/runs/17753506930/job/50452650642?pr=4633

lum1n0us (commented):
LGTM. Let's wait for the CI results.

lum1n0us (approved):

lum1n0us:
I think we should also patch release/2.4.x, just to be safe.

yamt:
> I think we should also patch release/2.4.x, just to be safe.

i will consider it once this PR lands.

Issue #4609, opened 2025-09-09
WAMR does not have ARM64 Windows assembly for native invokes


cosmo0920:
#### Feature

In cmake files for building nativeInvokes for many platforms, there is no ARM64 Windows definition.
 
#### Benefit

Currently, ARM64 Windows is just a small portion of running Windows, but Microsoft surface and other first series of Copilot+PCs are already running on Snapdragon X Elite which is built on top of ARM64 technology.

#### Implementation

Like em64*.asm implementation, we need to implement assemblies which are fulfilled with MASM calling conventions to build with the MSVC ecosystem.

#### Alternatives

There is no alternatives. This is because these assembly could be one of the core features to run WebAssembly on each of the platforms, I suppose.

TianlongLiang:
Do you have a testing environment? If so, you can try to use GenAI to write similar asm code from other platforms. I think it's not that hard to write code, but hard to test on our end. It would be great if you could try it if you have a testing environment.

cosmo0920:
I have a testing environment which is ARM64 Windows built on top of Snapdragon X Elite. It's a really ARM64 Windows environment.

PR #4636, opened 2025-09-16
Implement invokeNative asm code for armasm64 assembler on ARM64 Windows

cosmo0920:
In ARM64 Windows, we need to use armasm64.exe instead of ml.exe.
So, we need to specify it manually inside of cmake files.

Fixes https://github.com/bytecodealliance/wasm-micro-runtime/issues/4609.


cosmo0920:
> Code LGTM, would you mind rebase after #4633?

Sure thing. Today is a public holiday here. So, I'll process tomorrow.

cosmo0920:
I rebased off the current main. So, CI tasks are all succeeded.

PR #4641, opened 2025-09-22
[UNIT] enhance complication unit test case

pugong2019:
- Add test_llvm_integration_optimization.cc with LLVM backend validation
    - Test optimization passes, code generation, and target machine setup
    - Cover PGO, vectorization, register allocation, and cross-compilation
    - Add llvm_test.wat/wasm for LLVM-specific compilation testing
    - Add test_integration_performance.cc for end-to-end workflow validation
    - Implement performance benchmarking and memory profiling tests
    - Add concurrent compilation, stress testing, and error recovery
    - Test multi-module compilation and deterministic output validation

### DIFF
--- a/tests/unit/enhanced_unit_test/compilation/test_integration_performance.cc
+++ b/tests/unit/enhanced_unit_test/compilation/test_integration_performance.cc
@@ -39,6 +39,16 @@ class IntegrationPerformanceTest : public testing::Test
 protected:
     void SetUp() override
     {
+        // Initialize WAMR runtime first
+        RuntimeInitArgs init_args;
+        memset(&init_args, 0, sizeof(RuntimeInitArgs));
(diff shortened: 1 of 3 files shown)


tests/unit/enhanced_unit_test/compilation/wasm-apps/llvm_test.wasm
lum1n0us, line 1: Usually, we don't commit binaries, like .wasm files. They should be generated from .wat files (which you have) during compilation before executing test cases.

lum1n0us (commented):
WIP. not finished.

lum1n0us:
*tests/unit/enhanced_unit_test/* ==> *tests/unit/enhanced-unit-test/*

PR #4643, opened 2025-09-22
Add missing condition for V128 in WASM_OP_TEE_LOCAL

jammar1:
I encountered a crash with this error "Error: Exception: invalid local type" and it seems like WASM_OP_TEE_LOCAL was missing a branch for the V128 type.


lum1n0us:
@jammar1 please refer to https://github.com/bytecodealliance/wasm-micro-runtime/pull/4645 and add something in [tests/regression/ba-issues/issues](https://github.com/bytecodealliance/wasm-micro-runtime/pull/4645/files/3dd98d2f161a55e0c69a35f65c8ae492465d43c2#diff-ec3b8a2ff2fe45bfba383c82a44d2005986c9c82224248c3ee4e6e22927790de)

tests/regression/ba-issues/issues/issue-4643/test.wat
```diff
@@ -0,0 +1,44 @@
+(module
+  (memory 1)
+  (export "memory" (memory 0))
+
+  (func $test
+    ;; Add 130 i64 locals (260 slots) to push v128 past offset 256
```
jammar1, line 6: We need to hit the slow local.tee implementation so we need local_offset > 256, see: https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/interpreter/wasm_loader.c#L13895

PR #4645, opened 2025-09-23
fix: false OOB in array.fill for interp

### LINKED ISSUE #4644: classic-interp: array.fill false OOB with full length
On the classic interpreter, `array.fill` fails with "out of bounds" exception when length is identical to the array size, which is unexpected. 

### Test case

```wat
(type $i32_array (array (mut i32)))

(func $main
  (local $arr (ref $i32_array))

  ;; 创建一个大小为10的i32数组，初始值为0
  (array.new $i32_array
    (i32.const 0)    ;; 初始值
    (i32.const 10))  ;; 数组大小
  (local.set $arr)

  ;; 使用 array.fill 填充数组
  (array.fill $i32_array
    (local.get $arr)  ;; 数组引用
    (i32.const 0)     ;; 起始索引
    (i32.const 42)    ;; 填充值
    (i32.const 10))   ;; 填充长度 - 填满整个数组
)
```

### Your environment

* Host OS: Linux
* WAMR `main` branch, x86-64, classic-interp

### Expected behavior

No exception

### Actual behavior

OOB

XeniaLu:
Fixes #4644

### DIFF
--- a/core/iwasm/interpreter/wasm_interp_classic.c
+++ b/core/iwasm/interpreter/wasm_interp_classic.c
@@ -3165,7 +3165,7 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
 
                         if (len > 0) {
                             if ((uint64)start_offset + len
-                                >= wasm_array_obj_length(array_obj)) {
+                                > wasm_array_obj_length(array_obj)) {
                                 wasm_set_exception(
                                     module, "out of bounds array access");
                                 goto got_exception;
--- a/core/iwasm/interpreter/wasm_interp_fast.c
+++ b/core/iwasm/interpreter/wasm_interp_fast.c
@@ -2562,7 +2562,7 @@ wasm_interp_call_func_bytecode(WASMModuleInstance *module,
 
                         if (len > 0) {
                             if ((uint64)start_offset + len
-                                >= wasm_array_obj_length(array_obj)) {
+                                > wasm_array_obj_length(array_obj)) {
                                 wasm_set_exception(
                                     module, "out of bounds array access");
                                 goto got_exception;


tests/regression/ba-issues/issues/issue-4644/test.wat
```diff
@@ -0,0 +1,24 @@
...
+  (local.set $arr)
+
+  ;; 使用 array.fill 填充数组
+  (array.fill $i32_array
+    (local.get $arr)  ;; 数组引用
+    (i32.const 0)     ;; 起始索引
+    (i32.const 42)    ;; 填充值
+    (i32.const 10))   ;; 填充长度 - 填满整个数组
+)
+
```
lum1n0us: 👍 love it

And if it's convenient for you, please write comments in English for better readability by others.

yamt:
doesn't the official spec test cover this very basic case? i guess it should be added...

XeniaLu:
> doesn't the official spec test cover this very basic case? i guess it should be added...

Good point! Indeed it doesn't:

https://github.com/WebAssembly/spec/blob/e1ce46897460b74548c5e96b1530f59ef5b366cf/test/core/gc/array_fill.wast#L61-L68

For an array of size 12, the official spec test only covered:

* offset=13, len=0 ❌
* offset=0, len=13 ❌
* offset=12, len=0 ✅
* offset=2, len=2 ✅

I'll add a 3 ones there:

* offset=0, len=12 ✅
* offset=0, len=1 ✅
* offset=8, len=4 ✅

XeniaLu:
Spec test added in https://github.com/WebAssembly/spec/pull/1998 and cherry-picked as a patch.

PR #4647, opened 2025-09-24
loader: fix block/loop ref params type checking

### LINKED ISSUE #4646: interp: checked wrong reftype popping block/loop parameters
When a `block`/`loop` has multiple different reference parameters, the `reftype` is not updated to each parameter when compared to the popped parameter value from the stack, leading to unexpected type mismatch exception.

### Test case

```wat
  (i32.const 10)
  (struct.new $struct_a ...)
  (i32.const 20)
  (struct.new $struct_b ...)

  (block (param i32 (ref $struct_a) i32 (ref $struct_b)) (result (ref $struct_c)) ... )
```

### Your environment

* Host OS: Linux
* WAMR 2.4.2, x86-64, classic-interp

### Steps to reproduce

Load a module with such `block` / `loop` instructions.

### Expected behavior

Load successfully.

### Actual behavior

```
Error loading wasm module: WASM module load failed: type mismatch: expect (ref ht) but got other
```

### Extra Info

Only in this block:

https://github.com/bytecodealliance/wasm-micro-runtime/blob/6450d872995e84a5febeac1254cdd80dad3b94c1/core/iwasm/interpreter/wasm_loader.c#L11955-L11956

was the `wasm_ref_type` a.k.a. `loader_ctx->ref_type_tmp` set properly:

https://github.com/bytecodealliance/wasm-micro-runtime/blob/6450d872995e84a5febeac1254cdd80dad3b94c1/core/iwasm/interpreter/wasm_loader.c#L11991-L12000

But not in this block (earlier):

https://github.com/bytecodealliance/wasm-micro-runtime/blob/6450d872995e84a5febeac1254cdd80dad3b94c1/core/iwasm/interpreter/wasm_loader.c#L11909-L11910

XeniaLu:
Fixes #4646


tests/regression/ba-issues/issues/issue-4646/test.wat
```diff
@@ -0,0 +1,31 @@
+;; 定义不同的引用类型
```
lum1n0us: 👍 love it

And if it's convenient for you, please write comments in English for better readability by others.

XeniaLu: aw sure - it was LLM-generated, so 😏

PR #4653, opened 2025-09-30
Add new cases in ARC relocation of AOT

Septa2112:


yamt:
although i have no idea about arc, the change looks reasonable, glancing at https://github.com/foss-for-synopsys-dwc-arc-processors/arc-ABI-manual/blob/master/arcv3-elf.md

PR #4345, opened 2025-06-09
Modify AOT static PGO to conform to llvm-18 and add a CI job to test static PGO on the coremark benchmark

TianlongLiang:

PR #4654, opened 2025-10-01
Update LLVM to 21.1.2.

wffurr:
This required minor changes to LLVMContext construction and PGOOptions.

### DIFF
--- a/build-scripts/build_llvm.py
+++ b/build-scripts/build_llvm.py
@@ -294,7 +294,7 @@ def main():
         "arc": {
             "repo": "https://github.com/llvm/llvm-project.git",
             "repo_ssh": "git@github.com:llvm/llvm-project.git",
-            "branch": "release/18.x",
+            "branch": "release/21.x",
         },
         "xtensa": {
             "repo": "https://github.com/espressif/llvm-project.git",
@@ -304,7 +304,7 @@ def main():
         "default": {
             "repo": "https://github.com/llvm/llvm-project.git",
             "repo_ssh": "git@github.com:llvm/llvm-project.git",
-            "branch": "release/18.x",
+            "branch": "release/21.x",
         },
     }
 
--- a/.github/workflows/spec_test_on_nuttx.yml
+++ b/.github/workflows/spec_test_on_nuttx.yml
@@ -37,6 +37,19 @@ jobs:
     permissions:
       contents: read
       actions: write
+    uses: jlumbroso/free-disk-space@v1.3.1
+    with:
+      # Deletes Android, Haskell, and .NET runtimes from the runner, freeing
+      # about 15 GB.  
+      android: true
+      dotnet: true
+      haskell: true
+      # Keeps the tool cache, some large packages, docker images, and swap.
+      # Some of these could also be removed if more space is needed.
+      tool-cache: false
+      large-packages: false
+      docker-images: false
+      swap-storage: false
     uses: ./.github/workflows/build_llvm_libraries.yml
     with:
       os: "ubuntu-22.04"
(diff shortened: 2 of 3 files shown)


build-scripts/build_llvm.py
```diff
@@ -294,17 +294,17 @@ def main():
...
             "repo": "https://github.com/llvm/llvm-project.git",
             "repo_ssh": "git@github.com:llvm/llvm-project.git",
-            "branch": "release/18.x",
+            "branch": "release/21.x",
         },
         "xtensa": {
             "repo": "https://github.com/espressif/llvm-project.git",
             "repo_ssh": "git@github.com:espressif/llvm-project.git",
-            "branch": "xtensa_release_18.1.2",
+            "branch": "xtensa_release_21.1.2",
```
yamt: no such a branch availabile

wffurr: Thanks - I should have checked that.  Looks like the new API for LLVMContext construction will have to get wrapped in an #ifdef LLVM version check, since xtensa's latest release is only LLVM 19.

wffurr:
Is this good to merge?  The test failures look like timeouts; one error is "fatal error: error writing to /tmp/cckAF38s.s: No space left on device" which doesn't seem relevant.  Thanks!

yamt:
> Is this good to merge? The test failures look like timeouts; one error is "fatal error: error writing to /tmp/cckAF38s.s: No space left on device" which doesn't seem relevant. Thanks!

isn't it possible llvm 21 build actually somehow requires more space and caused ENOSPC on the ci?
i guess someone needs to investigate a bit further. (sorry i have no time right now)

wffurr:
> > Is this good to merge? The test failures look like timeouts; one error is "fatal error: error writing to /tmp/cckAF38s.s: No space left on device" which doesn't seem relevant. Thanks!
> 
> isn't it possible llvm 21 build actually somehow requires more space and caused ENOSPC on the ci? i guess someone needs to investigate a bit further. (sorry i have no time right now)

I found a script to free up disk space on the GitHub Ubuntu runner.  Is that OK to use?  Seems to fix the out of space error.  If you'd prefer I didn't use the action from the marketplace, I can hoist that bit out into a script in the WAMR repo and use that instead.

lum1n0us:
@yamt @TianlongLiang @loganek What are your thoughts on upgrading LLVM to version 21?

yamt:
> @yamt @TianlongLiang @loganek What are your thoughts on upgrading LLVM to version 21?

i don't have strong opinions either ways.
it's a bit sad to use different versions for different targets. but i guess it doesn't matter much.
otoh, i don't see any strong reasons to update to 21 either.

TianlongLiang:
For ARC, I don't think we should upgrade LLVM to 21.x. For the default target, we do need to update static pgo-related data structures like in [this pr](https://github.com/bytecodealliance/wasm-micro-runtime/pull/4345). Personally, I think we need to update the static PGO first before actually upgrading LLVM.

wffurr:
We're using WAMR with LLVM head and have patches for LLVM's changes in 21 and 22 that break WAMR.  This is probably not the way the software is intended to be used.  I was hoping to update to something closer to head.

I could send the patch for just `#if LLVM_VERSION_MAJOR >= 21` in `aot_llvm.c` instead of updating the LLVM targets in the build script.

lum1n0us:
Thanks for the contribution.
 
At the moment, we do not see a strong enough reason to upgrade WAMR’s LLVM baseline to 21, so we are not planning to change the default LLVM version in build_llvm.py for now.
 
Also, the aot_llvm.c change has already been merged through #4937.
 
As for the other two changes in this PR, they do not seem closely related to our current needs either. So unless there is some additional context or a stronger motivation for keeping this PR open, we plan to close it.
 
Thanks again for the work and for putting this together.

PR #4655, opened 2025-10-02
zephyr: including fs_interface.h

soburi:
`fs_file_t` and `fs_dir_t` is defined in fs_interface.h. However, platform_internal.h does not currently include it, so the symbols can't be resolved.
We include it to resolve these definitions with
platform_internal.h standalone.


yamt:
are you aware of https://github.com/bytecodealliance/wasm-micro-runtime/pull/4629 ?

soburi:
@yamt 

> are you aware of #4629 ?

Ah, thank you.
I didn't notice.

PR #4656, opened 2025-10-03
build: fix LLVM 18 build on GCC 15

XeniaLu:
See also: referenced LLVM pull requests in patch files.


lum1n0us:
Is this due to some compilation errors raised by GCC 15(or higher only)? Do lower versions of GCC, like GCC 11, encounter the same issues under specific conditions?

XeniaLu:
It's only happening on GCC 15 (and higher, and potentially some corresponding versions of clang): https://github.com/llvm/llvm-project/pull/101761 and https://github.com/llvm/llvm-project/pull/123320

lum1n0us:
Will lower versions of GCC, like GCC 11 (default on Ubuntu 22.04), successfully compile LLVM libraries after applying those (llvm) patches?

XeniaLu:
Yes, it [builds](https://github.com/rustica-lang/rustica-engine/actions/runs/18445507286/job/52551660203#step:9:104) on Ubuntu [22.04 (GCC 11)](https://github.com/rustica-lang/rustica-engine/actions/runs/18445507286/job/52551660203#step:9:28) with [the same patches](https://github.com/rustica-lang/rustica-engine/blob/3126e3c7b81dc60010f3d5f41fb7eb8e77cde53a/subprojects/llvm.wrap#L7-L8).

build-scripts/build_llvm.py
```diff
@@ -23,9 +23,15 @@ def clone_llvm(dst_dir, llvm_repo, llvm_branch):
 
     if not llvm_dir.exists():
         GIT_CLONE_CMD = f"git clone --depth 1 --branch {llvm_branch} {llvm_repo} llvm"
-        print(GIT_CLONE_CMD)
+        print(f"cd {dst_dir} && {GIT_CLONE_CMD}")
         subprocess.check_output(shlex.split(GIT_CLONE_CMD), cwd=dst_dir)
 
+        patch_dir = pathlib.Path(__file__).parent.joinpath("llvm-patches").resolve()
```
yamt, line 29: i suspect this should be target dependent.

while we are currently using the similar version (llvm-18) for all targets, it has not always been the case.
and it will unlikely be the case in future.
cf. https://github.com/bytecodealliance/wasm-micro-runtime/pull/4654

PR #4658, opened 2025-10-06
Remove warnings relevant to Zephyr platform

kr-t:
This PR is intended to remove following warnings, when build in Zephyr application:
```c
wasm-micro-runtime/core/shared/platform/zephyr/platform_internal.h:293: warning: "CLOCK_MONOTONIC" redefined
  293 | #define CLOCK_MONOTONIC 4
wasm-micro-runtime/core/shared/platform/zephyr/zephyr_file.c: In function 'zephyr_fs_alloc_obj':
wasm-micro-runtime/core/shared/platform/zephyr/zephyr_file.c:123:25: warning: implicit declaration of function 'bh_strdup' [-Wimplicit-function-declaration]
  123 |             ptr->path = bh_strdup(path);
      |                         ^~~~~~~~~
wasm-micro-runtime/core/shared/platform/zephyr/zephyr_file.c:123:23: warning: assignment to 'char *' from 'int' makes pointer from integer without a cast [-Wint-conversion]
  123 |             ptr->path = bh_strdup(path);
      |                       ^
wasm-micro-runtime/core/shared/platform/zephyr/zephyr_file.c: In function 'os_renameat':
wasm-micro-runtime/core/shared/platform/zephyr/zephyr_file.c:853:35: warning: initialization of 'char *' from 'int' makes pointer from integer without a cast [-Wint-conversion]
  853 |             char *new_path_copy = bh_strdup(new_path);
      |                                   ^~~~~~~~~
[45/462] Building C object CMakeFiles/app.dir/wasm-micro-runtime/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c.obj
wasm-micro-runtime/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c: In function 'wasmtime_ssp_poll_oneoff':
wasm-micro-runtime/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c:2216:42: warning: initialization of 'os_file_handle' {aka 'struct zephyr_handle *'} from 'int' makes pointer from integer without a cast [-Wint-conversion]
 2216 |                     os_file_handle tfd = fos[i]->file_handle->fd;
      |                                          ^~~
wasm-micro-runtime/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c:2222:31: warning: initialization of 'int' from 'os_file_handle' {aka 'struct zephyr_handle *'} makes integer from pointer without a cast [-Wint-conversion]
 2222 |                         .fd = tfd,
      |                               ^~~
wasm-micro-runtime/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c:2222:31: note: (near initialization for '(anonymous).fd')
```

Result - no warnings. 

Tested in [ocre](https://github.com/project-ocre/ocre-runtime) as a [zephyr](https://github.com/zephyrproject-rtos/zephyr) application.

### DIFF
--- a/core/shared/platform/zephyr/zephyr_file.c
+++ b/core/shared/platform/zephyr/zephyr_file.c
@@ -6,6 +6,7 @@
 #include "platform_api_vmcore.h"
 #include "platform_api_extension.h"
 #include "libc_errno.h"
+#include "bh_common.h"
 
 #include <string.h>
 #include <stdlib.h>
--- a/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
+++ b/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c
@@ -2213,7 +2213,7 @@ wasmtime_ssp_poll_oneoff(wasm_exec_env_t exec_env, struct fd_table *curfds,
 
 // Temporary workaround (see PR#4377)
 #ifdef BH_PLATFORM_ZEPHYR
-                    os_file_handle tfd = fos[i]->file_handle->fd;
+                    int tfd = fos[i]->file_handle->fd;
 #else
                     os_file_handle tfd = fos[i]->file_handle;
 #endif
--- a/core/shared/platform/zephyr/platform_internal.h
+++ b/core/shared/platform/zephyr/platform_internal.h
@@ -290,7 +290,9 @@ typedef struct timespec os_timespec;
 #define CLOCK_REALTIME 1
 #endif
 
+#ifndef CLOCK_MONOTONIC
 #define CLOCK_MONOTONIC 4
+#endif
 
 static inline int
 os_sched_yield(void)


kr-t:
Update after the initial commit:

Apparently, the CI doesn't like that in Zephyr's case I used int instead of os_file_handle. I have tried another way of refactoring the logic of how we abstract the `os_file_handle tfd` variable, but each time I have successfully removed warnings, the CI was unhappy with my changes and I couldn't pass all checks. Therefore, the changes in this PR in its current state leaves us with following - a successful CI, reduced amount of warnings and following warnings left:
```C
wasm-micro-runtime/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c:2216:42: warning: initialization of 'os_file_handle' {aka 'struct zephyr_handle *'} from 'int' makes pointer from integer without a cast [-Wint-conversion]
 2216 |                     os_file_handle tfd = fos[i]->file_handle->fd;
      |                                          ^~~
wasm-micro-runtime/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c:2222:31: warning: initialization of 'int' from 'os_file_handle' {aka 'struct zephyr_handle *'} makes integer from pointer without a cast [-Wint-conversion]
 2222 |                         .fd = tfd,
      |                               ^~~
wasm-micro-runtime/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/posix.c:2222:31: note: (near initialization for '(anonymous).fd')
``` 

Fixing temporary workaround which was made in PR#4377 will probably help with these remaining warnings as well.

lum1n0us:
IMM, a better approach is to avoid using bh_xxx() functions in zephyr/zephyr_file.c. Instead, opt for a specific version that is compatible with Zephyr

kr-t:
Noted, will check out the alternatives.

kr-t:
Removed bh_strdup() function, therefore reducing the undeclared warning. 
Since zephyr supports strdup() only when CONFIG_POSIX_C_LIB_EXT is selected, I didn't want to force users to use this, so I replaced strdup() with more generic strcpy().

core/shared/platform/zephyr/zephyr_file.c
```diff
@@ -850,8 +852,10 @@ os_renameat(os_file_handle old_handle, const char *old_path,
     for (int i = 0; i < CONFIG_WASI_MAX_OPEN_FILES; i++) {
         struct zephyr_fs_desc *ptr = &desc_array[i];
         if (ptr->used && ptr->path && strcmp(ptr->path, abs_old_path) == 0) {
-            char *new_path_copy = bh_strdup(new_path);
+            size_t new_path_len = strlen(abs_new_path) + 1;
```
lum1n0us, line 855: It seems there is another modification to store the absolute path instead of the relative path, which might be better to split into a new PR.

core/shared/platform/zephyr/zephyr_file.c
```diff
@@ -850,8 +852,10 @@ os_renameat(os_file_handle old_handle, const char *old_path,
...
+            char *new_path_copy = BH_MALLOC(new_path_len);
             if (new_path_copy != NULL) {
```
lum1n0us, line 857: The error-handling branch seems a bit odd; we should always expect the update to succeed, right? I mean, if the allocation fails, shouldn't the entire rename operation fail as well?

core/shared/platform/zephyr/zephyr_file.c
kr-t: This PR was intended to fix two warning, redefinition of `CLOCK_MONOTONIC` and implicit declaration of function 'bh_strdup'. With first one parallelly being committed by @TianlongLiang in https://github.com/bytecodealliance/wasm-micro-runtime/pull/4668 , this PR can't be further split anymore.

kr-t: Or if you mean that I replaced bh_strdup in two places and it would be better to split them in two separate PRs, or I missed something, please let me know.

core/shared/platform/zephyr/zephyr_file.c
kr-t: Essentially, this was planed to be as simple as "replace the keyword causing warning without changing the currently implemented logic". But yes, I agree. I will add a failing branch and return error code.

core/shared/platform/zephyr/zephyr_file.c
lum1n0us: `new_path` -> `abs_new_path`. 

The modified version stores absolute paths instead of relative paths in `desc_array`, which looks good. It would be better to build absolute paths in `os_open_preopendir()` to ensure all paths are absolute. 

If this is intentional, it should be mentioned in the PR description. It's your decision whether to split this into another PR or keep it in the current one. I am okay with both options.

lum1n0us (approved):

PR #4661, opened 2025-10-08
Merge version update to 2.4.2 back into main branch

wffurr:
Merging the `release/2.4.x` branch back into `main` updates the version number in `main` to 2.4.2 and makes the `2.4.2` commit tag the newest "reachable" commit from `git describe`, e.g.

```
$ git describe --abbrev=0 --tags
WAMR-2.4.2
```

This is useful to let automated tooling know that CVE-2025-58749 is fixed in the `main` branch.


wffurr:
Thanks lum1n0us!  May I please have one more approval from a maintainer?  It would be great to get the version updated for our security vulnerability scanner.

wffurr:
Thanks for merging the PR!

> i guess we need a bit refinement on the release process.

I think the version bump needs to happen on main and then cherry pick that commit into the release branch.  I'm not a git expert, so there might be a simpler way to do this.

wffurr:
This did not work the way I thought it would.  I think the merge commit on `bytecodelliance:main` has to be an actual merge commit from the `release/2.4.x` branch for `git describe` to be able to reach the tagged commit from head.

Making that merge commit on `wffurr:main` and then sending a PR for it to `bytecodealliance:main` didn't create that connection.

I'm not any kind of git expert; I'm barely literate in it; so I don't have any good suggestions for your release process here.

To fix our internal issue with the vulnerability scanner and version detection tools, I changed our import tool to pull from the `release/2.4.x` branch instead of `main`.

Thanks for taking a look at this!

PR #4663, opened 2025-10-10
Prevent `BuildPhi` from encountering a null llvm_entry_block

lum1n0us:
There is a scenario where `aot_compile_op_block()` does not prepare a block for `if`. As a result, the return value of `LLVMBuildPhi()` in `push_aot_block_to_stack_and_pass_params()` will be dangling, leading to a memory leak as it cannot be released.

### DIFF
--- a/core/iwasm/compilation/aot_emit_control.c
+++ b/core/iwasm/compilation/aot_emit_control.c
@@ -495,15 +495,17 @@ push_aot_block_to_stack_and_pass_params(AOTCompContext *comp_ctx,
 
         /* Create param phis */
         for (i = 0; i < block->param_count; i++) {
-            SET_BUILDER_POS(block->llvm_entry_block);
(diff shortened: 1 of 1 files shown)


yamt:
> There is a scenario where `aot_compile_op_block()` does not prepare a block for `if`. As a result, the return value of `LLVMBuildPhi()` in `push_aot_block_to_stack_and_pass_params()` will be dangling, leading to a memory leak as it cannot be released.

* do you have a small example to hit the case?
* how does it end up with leaks? iirc, instruction nodes belong to the builder context and will be freed with it.

lum1n0us:
[4663.wat.txt](https://github.com/user-attachments/files/22919362/4663.wat.txt)

Reproduce it with wamrc, ASAN, and 4633.wasm.

Based on my observation, if an instruction isn't inserted into a basic block, LLVM might lose track of it, leading to a leak.

PR #4460, opened 2025-07-07
Add support for metadata.code.branch_hint section

Lukasdoe:
Since we're currently in the progress of upstreaming the generation of the code metadata branch hint section in llvm ([PR](https://github.com/llvm/llvm-project/pull/146230)), I also implemented support for this feature in wamr. 

One important consideration for the implementation was to keep the encoding of hints generic, so that more [compilation hints](https://github.com/WebAssembly/compilation-hints/blob/main/proposals/compilation-hints/Overview.md) could be added in the future.

PR #4664, opened 2025-10-10
Add support for metadata.code.call_targets section

Lukasdoe:
Following #4460, I also implemented [call targets](https://github.com/WebAssembly/compilation-hints/blob/main/proposals/compilation-hints/Overview.md#call-targets) hints in WAMR. Support in LLVM is tbd.

This also includes the generalization of our metadata parsing mechanism so that it'll be very easy to integrate new types of compilation metadata.

Basis: https://github.com/WebAssembly/compilation-hints/blob/main/proposals/compilation-hints/Overview.md

### DIFF
--- a/core/iwasm/compilation/aot_emit_control.c
+++ b/core/iwasm/compilation/aot_emit_control.c
@@ -279,7 +279,7 @@ aot_emit_branch_hint(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
         hint = hint->next;
     }
     if (hint != NULL) {
-        ((struct WASMCompilationHintBranchHint *)hint)->used = true;
+        hint->used = true;
         // same weight llvm MDBuilder::createLikelyBranchWeights assigns
         const uint32_t likely_weight = (1U << 20) - 1;
         const uint32_t unlikely_weight = 1;
--- a/core/iwasm/interpreter/wasm.h
+++ b/core/iwasm/interpreter/wasm.h
@@ -778,12 +778,7 @@ struct WASMCompilationHint {
     bool used;
 };
 struct WASMCompilationHintBranchHint {
-    struct WASMCompilationHint *next;
-    enum WASMCompilationHintType type;
-    uint32 offset;
-    bool used;
-
-    // custom field
+    struct WASMCompilationHint common;
     bool is_likely;
 };
 struct WASMCompilationHintCallTargetsHint {
@@ -791,12 +786,7 @@ struct WASMCompilationHintCallTargetsHint {
     uint32 call_frequency;
 };
 struct WASMCompilationHintCallTargets {
-    struct WASMCompilationHint *next;
-    enum WASMCompilationHintType type;
-    uint32 offset;
-    bool used;
-
-    // custom fields
+    struct WASMCompilationHint common;
     size_t target_count;
     struct WASMCompilationHintCallTargetsHint *hints;
 };
--- a/core/iwasm/interpreter/wasm_loader.c
+++ b/core/iwasm/interpreter/wasm_loader.c
@@ -5648,7 +5648,7 @@ handle_compilation_hint_branch_hint_processor(const uint8 *buf,
 {
     (void)module;
     struct WASMCompilationHintBranchHint *hint = hint_store;
-    hint->type = WASM_COMPILATION_HINT_BRANCH;
+    hint->common.type = WASM_COMPILATION_HINT_BRANCH;
     if (hint_size != 1) {
         set_error_buf_v(error_buf, error_buf_size,
                         "invalid branch hint size, expected 1, got %d.",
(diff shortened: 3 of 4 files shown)


core/iwasm/interpreter/wasm.h
```diff
@@ -765,21 +765,41 @@ struct WASMTag {
...
     struct WASMCompilationHint *next;
     enum WASMCompilationHintType type;
+    uint32 offset;
+    bool used;
 };
 struct WASMCompilationHintBranchHint {
     struct WASMCompilationHint *next;
     enum WASMCompilationHintType type;
     uint32 offset;
+    bool used;
```
yamt: i feel it's simpler to have `struct WASMCompilationHint common` instead of duplicating individual fields.

core/iwasm/compilation/aot_emit_control.c
```diff
@@ -272,14 +272,14 @@ aot_emit_branch_hint(AOTCompContext *comp_ctx, AOTFuncContext *func_ctx,
...
-            && ((struct WASMCompilationHintBranchHint *)hint)->offset
-                   == offset) {
+        if (hint->type == WASM_COMPILATION_HINT_BRANCH
+            && hint->offset == offset) {
             break;
         }
         hint = hint->next;
     }
     if (hint != NULL) {
+        ((struct WASMCompilationHintBranchHint *)hint)->used = true;
```
yamt: no need to cast as the `used` field is in WASMCompilationHint as well.

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -7500,13 +7621,35 @@ wasm_loader_unload(WASMModule *module)
...
+            // each code metadata section parsed adds a chain of hints to the
+            // function hints. That chain of WASMCompilationHint specializations
+            // is allocated as an array. We need to find the starts of the
+            // separate arrays to free them all properly.
+            struct WASMCompilationHint *curr = module->function_hints[i];
+            struct WASMCompilationHint *last_chain_start = curr;
+            while (curr != NULL) {
+                if (!curr->used) {
+                    printf("Unused hint for function %u, offset: %x\n",
+                           i + module->import_count, curr->offset);
```
yamt: we usually don't use printf.
maybe LOG_WARNING?

core/iwasm/compilation/aot_llvm_extra.cpp
```diff
@@ -433,3 +433,9 @@ aot_compress_aot_func_names(AOTCompContext *comp_ctx, uint32 *p_size)
     *p_size = compressed_str_len;
     return compressed_str;
 }
+
+uint64_t
+aot_func_name_hash(const char *name)
+{
+    return MD5Hash(StringRef(name));
```
yamt, line 440: is this function ubiquitously available?
i suspect it isn't.

core/iwasm/interpreter/wasm.h
```diff
@@ -765,21 +765,41 @@ struct WASMTag {
...
     DUMMY = 0,
-    WASM_COMPILATION_BRANCH_HINT = 0,
+    WASM_COMPILATION_HINT_BRANCH = 1,
+    WASM_COMPILATION_HINT_CALL_TARGETS = 2,
 };
 struct WASMCompilationHint {
     struct WASMCompilationHint *next;
     enum WASMCompilationHintType type;
+    uint32 offset;
+    bool used;
```
yamt, line 778: is this `used` field only for diagnostic purposes?

core/iwasm/interpreter/wasm_loader.c
```diff
@@ -5648,6 +5637,116 @@ handle_branch_hint_section(const uint8 *buf, const uint8 *buf_end,
...
+    else {
+        set_error_buf_v(error_buf, error_buf_size,
+                        "invalid branch hint value, expected 0 or 1, got %d",
+                        data);
+        goto fail;
+    }
+    return true;
+fail:
+    return false;
+}
```
yamt, line 5673: put a blank line between functions

core/iwasm/interpreter/wasm.h
Lukasdoe: more or less. It is definitely a hint that something is wrong if a hint is not used. We apply hints to all branch instructions, so if it is not used, we either missed it, or the referenced instruction does not exist (meaning an invalid compilation hint section).

However, since these hints are not mission-critical, I didn't want to assert that this is true.

core/iwasm/compilation/aot_llvm_extra.cpp
Lukasdoe: Can you please elaborate a bit more on that? I don't understand what your concern is, since I'm not sure whether you are hinting at the availability of the MD5Hash function, which is part of LLVM's MD5.h support header, or something else.

core/iwasm/interpreter/wasm.h
Lukasdoe: Good idea, thx!

core/iwasm/compilation/aot_llvm_extra.cpp
yamt: i didn't know llvm provides MD5Hash and assumed it was something from linux-specific library.
if it's from llvm, it's fine.

PR #4676, opened 2025-10-21
Fix mismatch of enum sizes between WASM and host

dkouba-atym:
While working with WAMR on Zephyr, I discovered that the sizes of a couple socket enums were not consistent between the host and inside of WASM. This resulted in an inability to create a UDP socket, as the layout of the enum was compressed (enums became 1 byte instead of 4) and the offset of `hints_enabled` was incorrect, and therefore all sockets created were TCP.

In WASM: `sizes: enum=4, hints_t=12, off(type)=0 off(family)=4 off(en)=8`
In Zephyr host: `sizes: enum=1, hints_t=3, off(type)=0 off(family)=1 off(en)=2`

As a result I've enforced the struct sizes using the same method it was done for a different enum and added the appropriate assertions to validate.

### DIFF
--- a/core/shared/platform/include/platform_wasi_types.h
+++ b/core/shared/platform/include/platform_wasi_types.h
@@ -533,19 +533,14 @@ assert_wasi_layout(offsetof(__wasi_subscription_t, userdata) == 0, "witx calcula
 assert_wasi_layout(offsetof(__wasi_subscription_t, u) == 8, "witx calculated offset");
 
 /* keep syncing with wasi_socket_ext.h */
-typedef enum {
-    /* Used only for sock_addr_resolve hints */
-    SOCKET_ANY = -1,
-    SOCKET_DGRAM = 0,
-    SOCKET_STREAM,
-} __wasi_sock_type_t;
 
 typedef uint16_t __wasi_ip_port_t;
 
 /* Ensure that __wasi_addr_type_t has a size of 4 byte (I32).
    However, it will not have the type safety of enum. */
 typedef uint32_t __wasi_addr_type_t;
-enum { IPv4 = 0, IPv6 };
+#define IPv4 (0)
+#define IPv6 (1)
 
 /* n0.n1.n2.n3 */
 typedef struct __wasi_addr_ip4_t {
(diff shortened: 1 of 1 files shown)


lum1n0us:
I believe we should synchronize this modification with core/iwasm/libraries/lib-socket/inc/wasi_socket_ext.h.

core/shared/platform/include/platform_wasi_types.h
```diff
@@ -589,20 +583,36 @@ typedef struct __wasi_addr_t {
     } addr;
 } __wasi_addr_t;
 
-typedef enum { INET4 = 0, INET6, INET_UNSPEC } __wasi_address_family_t;
+/* Force 32-bit wire width for cross-boundary fields */
+typedef int32_t __wasi_sock_type_t;
+enum { SOCKET_ANY = -1, SOCKET_DGRAM = 0, SOCKET_STREAM = 1 };
```
yamt: i agree it isn't appropriate to use enums to define an ABI.

it's better to retire all enums in this file.
ie.
```c
#define SOCKET_DGRAM 0
```

it's even better to stop using too generic names.
```c
typedef int32_t __wamr_socket_ext_sock_type_t;
#define WAMR_SOCKET_EXT_SOCK_TYPE_DGRAM 0
```
note that SOCKET_DGRAM is NOT wasi SOCK_DGRAM.

srberard: Hi @yamt,
Are you suggesting that we replace all of the enums with #defines?  For example:
```c
enum { IPv4 = 0, IPv6 };
```
with this:
```c
#define IPv4   0
#define IPv6   1
```

I'm happy to update the PR and make this change.  The current PR continues to use enums but fixes the typedef to use fixed sizes (following the pattern used for `__wasi_addr_type_t`).

I also agree with changing things like `SOCKET_DGRAM` to something like `WAMR_SOCKET_EXT_SOCK_TYPE_DGRAM ` as it is very confusing, particularly when having to debug.  The only reservation I have is that will likely break existing code that uses the `SOCKET_*` values.

lum1n0us: I believe what @yamt  means is to use `#define` instead of `enum`. Otherwise, there is still an implicit casting from `enum` to `int32_t`, which could be avoided. 

I am okay with using more readable names (perhaps due to poor memory).

yamt: exactly.
sorry for late response.

kr-t:
I have replaced enums to defines. Please let me know if you'd like to change some naming or anything else additionally in this PR.

kr-t:
Squashed and rebased to latest main.

PR #4681, opened 2025-10-27
docs: clarify that * in signatures does address conversion

joe-p:
It was non-obvious that `*` means the runtime does the app to native conversion automatically. With the current docstring, one might assume that the runtime only does boundary checks and not address conversion. Under this assumption, passing the already-native address to `wasm_runtime_addr_app_to_native` will lead to unexpected behavior


lum1n0us:
LGTM. It is better to mention that both `$` and `*` in the signature will trigger address validation and address conversion.

joe-p:
Ah good point, added in [b683a01](https://github.com/bytecodealliance/wasm-micro-runtime/pull/4681/commits/b683a01c9709a57fdac842fb543b2ebf81e5072d)

lum1n0us (approved):

Issue #4618, opened 2025-09-11
[RFC] Explicitly validating the pointers against null


lum1n0us:
> Avoid null pointer de-reference errors by explicitly validating the pointers against null before using the pointer

A Well-known rules of defensive programming. But does wamr need it? I knew we're applying caller guarantee rules. And it's a simple checker (but a lot).

There are 5 files in core/iwasm/includes and approximately 170 APIs in wasm_export.h, 1 API in lib_export.h, 15 APIs in aot_export.h, 65 APIs in gc_export.h, and 180 APIs in wasm_c_api.h. 
 
Additionally, `char*` should be considered as a pointer first before being treated as a string.

Furthermore, a return value is required after implementing pointer parameter checking. Approximately 80 APIs might need redesigning because they previously returned void.

lum1n0us:
A basic 💡 is:

Develop a new set of APIs, named something like xxx_checked, based on existing APIs. like `wasm_runtime_get_export_type_checked()`, `wasm_runtime_attach_shared_heap_checked()` and so on.

Each new API will act as a wrapper for its corresponding existing API. like

```c
#include "wasm_export.h"

static inline bool
wasm_runtime_attach_shared_heap_checked(wasm_module_inst_t module_inst,
                                        wasm_shared_heap_t shared_heap)
{
  CHECK_NULL_OR_RET_FALSE(module_inst);
  CHECK_NULL_OR_RET_FALSE(shared_heap);
  return wasm_runtime_attach_shared_heap(module_inst, shared_heap);
}
```

All new APIs will be placed in new files, named something like xxx_checked. like *wasm_export_checked.h*, *lib_export_checked.h* and so on.

no1wudi:
Can we embed it in exisitng API and control it via build flags, which is similar to fortity?

lum1n0us:
> Fortify is a suite of tools used for application security testing, which analyzes software code to find and fix security vulnerabilities before deployment. 

Do you mean this fortify?

> Can we embed it in existing API and control it via build flags?

@no1wudi  I'm not following. Would you mind giving an example?

yamt:
honestly speaking, i feel this is too much complexity for little or no benefits.

yamt:
as it's to detect programming errors, it's more appropriate to use assertions.
```c
bh_assert(module_inst != NULL);
```

lum1n0us:
> as it's to detect programming errors, it's more appropriate to use assertions.

[...]

PR #4682, opened 2025-10-29
A new set of APIs with input pointer NULL-checkers

lum1n0us:
- *ci/generate_checked_functions.py* is used to generate checked APIs from the original APIs.
- *.github/workflows/verify_checked_apis.yml* ensures that the checked APIs are always derived from the latest APIs.
- *samples/checked-api/* provides a quick example.
- *core/iwasm/include/\*_checked.h* contains the generated checked APIs.


lum1n0us:
Based on the discussion in issue #4618, it is not a good idea to maintain another set of APIs solely for parameter checking.

PR #4688, opened 2025-10-31
fix: remove stack size check from exec_env_check

joe-p:
As per [this blogpost](https://bytecodealliance.github.io/wamr.dev/blog/understand-the-wamr-stacks/) WAMR does not require a WASM stack in AoT/JIT mode, but `wasm_runtime_exec_env_check` still always expects the size to be > 0


joe-p:
Looks like the CI is failing due to a disk issue with the runner:

> fatal: cannot create directory at 'llvm/test/tools/llvm-mca/AArch64/Cortex/IPC': No space left on device

PR #4690, opened 2025-11-03
feat: add support for checking unsupported build configurations

lum1n0us:


tests/unit/shared-heap/CMakeLists.txt
```diff
@@ -10,7 +10,7 @@ add_definitions(-DRUN_ON_LINUX)
 set(WAMR_BUILD_APP_FRAMEWORK 0)
 set(WAMR_BUILD_AOT 1)
 set(WAMR_BUILD_INTERP 1)
-set(WAMR_BUILD_FAST_INTERP 1)
```
TianlongLiang, line 13: Fast interp does support shared heap

lum1n0us: Then, disable memory64?

```
set(WAMR_BUILD_FAST_INTERP 1)
#...
if(WAMR_BUILD_TARGET STREQUAL "X86_32")
    set(WAMR_BUILD_MEMORY64 0)
else()
    set(WAMR_BUILD_MEMORY64 1)   # <- FAST_INTERP doesn't support WAMR_BUILD_MEMORY64
endif()
set(WAMR_BUILD_SHARED_HEAP 1)
```

TianlongLiang: Good idea, let's disable them for now

PR #4697, opened 2025-11-10
core/iwasm: Add proper CFI directives to ARM Thumb invokeNative 

no1wudi:
This commit adds comprehensive Call Frame Information (CFI) directives to the ARM Thumb invokeNative assembly function to enable proper stack unwinding in GDB debugging sessions.

Key changes:
- Added .cfi_startproc and .cfi_endproc to define the function boundaries
- Implemented .cfi_def_cfa directives to establish the canonical frame address
- Added .cfi_offset directives for all callee-saved registers (lr, r7, r6, r5, r4)
- Established proper frame pointer (r7) usage for reliable stack unwinding
- Enhanced stack frame management with consistent alignment and cleanup

The CFI directives allow GDB to accurately track register values and unwind the call stack when debugging native function invocations, significantly improving the debugging experience for WAMR applications on ARM Thumb platforms.

### DIFF
--- a/core/iwasm/common/arch/invokeNative_thumb_vfp.s
+++ b/core/iwasm/common/arch/invokeNative_thumb_vfp.s
@@ -12,7 +12,7 @@ invokeNative:
         .globl _invokeNative
 _invokeNative:
 #endif /* end of BH_PLATFORM_DARWIN */
-
+        .cfi_startproc
 /*
  * Arguments passed in:
  *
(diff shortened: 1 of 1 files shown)


no1wudi:
Before this patch:
image
After this patch:
屏幕截图 2025-11-09 225551

lum1n0us:
Great work. Do you have any plans to expand similar functionality to other architectures, like x86, x86-64, and so on?

no1wudi:
> Great work. Do you have any plans to expand similar functionality to other architectures, like x86, x86-64, and so on?

I plan to implement it for arm/riscv first, and later for other platforms

TianlongLiang:
This feature will greatly facilitate the debugging process. But do you think we should conditionally include it, controlled by CMake config or something? Since it will add overhead when invoking a native function?

no1wudi:
> This feature will greatly facilitate the debugging process. But do you think we should conditionally include it, controlled by CMake config or something? Since it will add overhead when invoking a native function?

No "overhead" here, CFI directives is just annotation for debugger, will not generate actual code if don't touch the asm function itself

PR #4713, opened 2025-11-18
Adding additional text around the test requirement for code.

woodsmc:
Speaking to @lum1n0us I was suggesting that we change our process for accepting code contributions to require testing, unless there is a good reason not to have them. At the moment supplying a test is optional. 

I suggest we should change this to be a requirement -> if you change code supply a test. We could have contributors point to existing tests which cover their contribution, or supply new testing. If the code change is super tiny, like formatting, or variable name change, then code's existing functionality is testing with the existing tests.

If the code contribution is larger or affects logic, fixes a bug etc, then we should require a test. In the case of a bug, a test case to reproduce the issue, and the fix, so we can re-run the test case to show the issue cannot be reproduced.

This is my first attempt at the text for this, it's a little all over the place. But it's a start. I'd welcome your thoughts on the idea.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,8 +18,12 @@ Code changes
 ===================
 We Use Github Flow, So All Code Changes Happen Through Pull Requests. Pull requests are the best way to propose changes to the codebase. We actively welcome your pull requests:
 
-- If you've added code that should be tested, add tests. Ensure the test suite passes.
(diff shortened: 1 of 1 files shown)


woodsmc:
Yes, and following these guides, I'd have to add :

> This is a minor change explain the change, and as such is except from addition test contribution.

Yes, the irony of me not adding this disclaimer is high,... ;p

TianlongLiang (approved):
Good idea 👍

Issue #4703, opened 2025-11-13
Hardware bounds checking fails for multi-memory + multi-module


sbird-kick:
**Describe the bug**
Using two modules in which one calls the other, I can reproduce a scenario in out of bounds memory accesses are ignored if done by the second module. This does not happen if you disable hardware bounds checks.

**Version**
I am using iwasm-2.4.3. The commit is 7898af9f21c5f0623c1b555bbd952b158cc9acc4.

**To Reproduce**
1. Compile iwasm with these flags `-DWAMR_BUILD_MULTI_MODULE=1 -DWAMR_BUILD_MULTI_MEMORY=1 -DWAMR_DISABLE_HW_BOUND_CHECK=0`
2. Compile these two wat files using `wat2wasm --enable-multi-memory`
single_page_memory.wat -> simple_page_memory.wasm:
```
(module
  (func $other_main (import "double_page_memory" "main_2") (result i32))
  (memory (export "memory") 1)
  (func $main (export "main") (result i32)
      i32.const 0
      i32.load 
      call $other_main
      drop
  )
)
```

double_page_memory.wat -> double_page_memory.wasm:
```
(module 
  (memory $first_memory (export "memory_2") 100)
  (memory $second_memory (export "memory_3") 200)
  (func $main_2 (export "main_2") (result i32)
      
      i32.const 10213
      
      i32.const 6553400
      i32.load 
      
      i32.div_s 
      drop

      i32.const 10213
      
      i32.const 1055340000
      i32.load $second_memory
      
      i32.div_s 
      
  )
  (data 0 (i32.const 6553400) "12345678" )
  (data 1 (i32.const 10553400) "12345678" )
)
```

3. Run them using `iwasm-2.4.3 single_page_memory.wasm --module-path=$(pwd)`

4. Compile iwasm again using `-DWAMR_BUILD_MULTI_MODULE=1 -DWAMR_BUILD_MULTI_MEMORY=1 -DWAMR_DISABLE_HW_BOUND_CHECK=1` and run it again

**Expected behavior**
I should see an out of bounds memory access in both.

```
humzai@humzai-2313D:~/bug_report_iwasm$ iwasm-2.4.3 single_page_memory.wasm --module-path=$(pwd)
Exception: out of bounds memory access
```

**Actual Result**
I only see an out of bounds memory access when iwasm is compiled with `-DWAMR_DISABLE_HW_BOUND_CHECK=1`

**Desktop**

- Arch x86
[...]

PR #4714, opened 2025-11-20
Allow HW OOB exception propagate from callee to caller issue when multi-module is enabled

TianlongLiang:


TianlongLiang:
Will add a test case too

tests/regression/ba-issues/build_wamr.sh
```diff
@@ -60,4 +67,7 @@ build_iwasm "-DWAMR_BUILD_REF_TYPES=1 -DWAMR_BUILD_JIT=1 -DWAMR_BUILD_LIBC_WASI=
 # build multi-tier-jit iwasm for testing classic-interp, fast-jit, llvm-jit and multi-tier-jit with libc-wasi disabled
 build_iwasm "-DWAMR_BUILD_REF_TYPES=1 -DWAMR_BUILD_FAST_JIT=1 -DWAMR_BUILD_JIT=1 -DWAMR_BUILD_LIBC_WASI=0" "multi-tier-jit-wasi-disabled"
 
+# build default iwasm for testing multi-module
+build_iwasm "-DWAMR_BUILD_MULTI_MODULE=1 -DWAMR_BUILD_MULTI_MEMORY=1 -DWAMR_BUILD_AOT=0 -DWAMR_BUILD_FAST_INTERP=0" "multi-memory-multi-module" "false"
```
lum1n0us: better to have a description of all the parameters of `build_iwasm().

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -248,6 +249,15 @@ runtime_signal_handler(void *sig_addr)
...
             wasm_set_exception(module_inst, "out of bounds memory access");
+#if WASM_ENABLE_MULTI_MODULE != 0
+            if (jmpbuf_node->module_inst
+                && jmpbuf_node->module_inst
+                       != (WASMModuleInstanceCommon *)module_inst) {
+                wasm_runtime_propagate_exception_from_import(
+                    (WASMModuleInstanceCommon *)jmpbuf_node->module_inst,
+                    (WASMModuleInstanceCommon *)module_inst);
+            }
+#endif
```
yamt, line 260: i'm not sure why we need WASM_ENABLE_MULTI_MODULE handling in this low level logic.

exception propagation should be done regardless of hw bound check, shouldn't it?

lum1n0us: This specific case #4703 occurs only when an exception happens in the child module instance, and the main module instance fails to copy that exception, causing the problem.

We both agree that there is [another simple solution](https://github.com/bytecodealliance/wasm-micro-runtime/pull/4714#pullrequestreview-3533295845) for this issue.

However, in the short term, I can't determine under which circumstances it is necessary to switch module instances before calling a WASM function. Please enlighten me.

yamt: it's better to minimize the logic in signal handler.
can't you propagate the exception as wasm_exec_env_restore_module_inst does for some reasons? why not?

lum1n0us: Yes, that's one way to patch it.

PR #4730, opened 2025-11-25
Add runtime API for shared heap destroy and related test cases

TianlongLiang:
And for existing test cases, fix some compile warnings

### DIFF
--- a/tests/unit/shared-heap/shared_heap_test.cc
+++ b/tests/unit/shared-heap/shared_heap_test.cc
@@ -292,6 +292,37 @@ create_test_shared_heap(uint8 *preallocated_buf, size_t size,
     }
 }
 
+static WASMSharedHeap *
+create_preallocated_shared_heap(size_t size, uint8 *buffer)
+{
+    SharedHeapInitArgs args = {};
+    WASMSharedHeap *shared_heap = NULL;
+
+    memset(buffer, 0, size);
+    args.pre_allocated_addr = buffer;
+    args.size = size;
+
+    shared_heap = wasm_runtime_create_shared_heap(&args);
+    if (!shared_heap) {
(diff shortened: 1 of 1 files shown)


core/iwasm/common/wasm_memory.c
```diff
@@ -339,6 +358,95 @@ wasm_runtime_unchain_shared_heaps(WASMSharedHeap *head, bool entire_chain)
...
+        cur = shared_heap_list->next;
+        while (prev) {
+            if (head == cur) {
+                prev->next = head->next;
+                destroy_shared_heap_node(cur);
+                break;
+            }
+            prev = prev->next;
+            cur = cur->next;
+        }
```
lum1n0us, line 418: It seems that the process of looking for `prev` can be integrated into the above loop to ensure `head_found`.

But the current implementation is definitely functional. It is acceptable to keep it as is.

core/iwasm/common/wasm_memory.c
```diff
@@ -339,6 +358,95 @@ wasm_runtime_unchain_shared_heaps(WASMSharedHeap *head, bool entire_chain)
...
+        if (cur->chain_next == head)
+            is_chain_head = false;
+    }
+    if (!head_found) {
+        LOG_WARNING("Shared heap %p isn't tracked by runtime.", head);
+        os_mutex_unlock(&shared_heap_list_lock);
+        return false;
+    }
+    if (!is_chain_head) {
+        LOG_WARNING("Shared heap %p isn't the head of a shared heap chain.",
```
lum1n0us, line 396: Suppose there are two chains, A-B-C and B-C, and both need to be `wasm_runtime_destroy_shared_heap()` `entire_chain`. 

First, destroy the first chain, which is A-B-C. When deleting up to C, it is discovered that C is not the head, and then the condition `if (!is_chain_head)` will fail and return.

PR #4740, opened 2025-12-05
Add a runtime API for reset shared heap(chain)

TianlongLiang:


yamt:
can you elaborate use cases a bit?

TianlongLiang:
> can you elaborate use cases a bit?

For example, if a WASM module calls `shared_heap_malloc` a few times and an exception occurs before it gets a chance to free it, then the WAMR runtime-managed shared heap will have some dangling pointer(some memory block can't be allocated) until the runtime destroys and reinitializes. This api provides a lighter way to reset(allow the whole WAMR runtime-managed shared heap to become allocatable) the shared heaps without re-initializing the runtime.

yamt:
> > can you elaborate use cases a bit?
> 
> For example, if a WASM module calls `shared_heap_malloc` a few times and an exception occurs before it gets a chance to free it, then the WAMR runtime-managed shared heap will have some dangling pointer(some memory block can't be allocated) until the runtime destroys and reinitializes. This api provides a lighter way to reset(allow the whole WAMR runtime-managed shared heap to become allocatable) the shared heaps without re-initializing the runtime.

so,

* it's an equivalent of destroy + re-create
* not intended to be used for a chain being used by running instances

right?

tests/unit/shared-heap/shared_heap_test.cc
```diff
@@ -32,13 +32,13 @@ static void
...
 static bool
-load_wasm(char *wasm_file_tested, unsigned int app_heap_size,
+load_wasm(const char *wasm_file_tested, unsigned int app_heap_size,
           ret_env &ret_module_env)
 {
     char *wasm_file = strdup(wasm_file_tested);
     unsigned int wasm_file_size = 0;
     unsigned int stack_size = 16 * 1024, heap_size = app_heap_size;
-    char error_buf[128] = { 0 };
+    char error_buf[128] = {};
```
yamt, line 41: is this a change for "fix some warnings"?
what warning?

tests/unit/shared-heap/shared_heap_test.cc
```diff
@@ -998,12 +994,154 @@ TEST_F(shared_heap_test, test_shared_heap_chain_unchain)
...
+
+    offset_after_reset = wasm_runtime_shared_heap_malloc(
+        tmp_module_env.wasm_module_inst, buf_size, &native_ptr_after_reset);
+    ASSERT_NE(0u, offset_after_reset);
+    ASSERT_NE(nullptr, native_ptr_after_reset);
+
+    EXPECT_EQ(offset, offset_after_reset);
+    EXPECT_EQ(native_ptr, native_ptr_after_reset);
+
+    /* Only on some platform, the os_mmap will memset the memory to 0
```
yamt, line 1069: does this mean that in some cases shared memory may have uninitialized user-visible contents?
is it documented somewhere?

tests/unit/shared-heap/shared_heap_test.cc
TianlongLiang:
```
 warning: missing initializer for member ‘SharedHeapInitArgs::pre_allocated_addr’ [-Wmissing-field-initializers]
 1081 |     SharedHeapInitArgs args = { 0 };
 ```
I simply replaced all occurrences in this file.

I remember that we had a discussion about the choice between {0} and {}, for C++, {} should be a more future-proof version, right?

TianlongLiang:
> > > can you elaborate use cases a bit?
> > 
> > 
> > For example, if a WASM module calls `shared_heap_malloc` a few times and an exception occurs before it gets a chance to free it, then the WAMR runtime-managed shared heap will have some dangling pointer(some memory block can't be allocated) until the runtime destroys and reinitializes. This api provides a lighter way to reset(allow the whole WAMR runtime-managed shared heap to become allocatable) the shared heaps without re-initializing the runtime.
> 
> so,
> 
> * it's an equivalent of destroy + re-create
> * not intended to be used for a chain being used by running instances
> 
> right?

Yes, you are right :)

tests/unit/shared-heap/shared_heap_test.cc
TianlongLiang: Seems that on Linux it won't reset to 0(the commented-out lines will fail), though it should, given our implementation uses `MAP_ANONYMOUS.`, I will dig into it

lum1n0us:
@yamt I am going to merge this PR. Please feel free to continue the discussion above.

PR #4517, opened 2025-07-28
wasi_nn_tensorflowlite.cpp: make this compatible with wasmedge

yamt:
for wasi_ephemeral_nn,

* implement u8 input

* stop dealing with quantization.

  * wasi-nn doesn't have a concept of quantization or pre/post-processing. i can't think of any ways to make the backend perform zero-point/scale processing without risking to break other applications.

  * there seems to be applications which just use u8 inputs/outputs for a quantized model. (see [1] for an example.) for certain kinds of inputs/outputs, it usually just works.

this commit keeps the legacy wasi_nn logic intact for now.

tested with [1] with [2] applied.

WAMR with this patch:
```
Read graph weights, size in bytes: 3561598
[wasi_nn.c:297 WARNING] load_by_name_with_config() not found
[wasi_nn_tensorflowlite.cpp:272 WARNING] Default encoding is CPU.
Loaded graph into wasi-nn with ID: Graph#0
Read input tensor, size in bytes: 150528
   1.) [166](198)Aix galericulata
   2.) [34](1)Gallus gallus domesticus
   3.) [158](1)Coccothraustes coccothraustes
   4.) [778](1)Sitta europaea
   5.) [819](1)Anas platyrhynchos
```

wasmedge:
```
Read graph weights, size in bytes: 3561598
Loaded graph into wasi-nn with ID: Graph#0
Read input tensor, size in bytes: 150528
   1.) [166](198)Aix galericulata
   2.) [34](1)Gallus gallus domesticus
   3.) [158](1)Coccothraustes coccothraustes
   4.) [778](1)Sitta europaea
   5.) [819](1)Anas platyrhynchos
```

and "Aix galericulata" seems like a reasonable classification of the image to my eyes.

[1] https://github.com/second-state/WasmEdge-WASINN-examples/tree/67f174bab59d98c1b52f7367ec0928701dc998f9/tflite-birds_v1-image

[2] https://github.com/second-state/WasmEdge-WASINN-examples/pull/204

Related:
https://github.com/bytecodealliance/wasm-micro-runtime/issues/3555 https://github.com/bytecodealliance/wasm-micro-runtime/issues/2611

PR #4764, opened 2025-12-19
add a set of apis to configure wasi-nn via InstantiationArgs2

zhanheng1:
add a set of apis to configure wasi-nn via InstantiationArgs2:

1. Add a new command line option for wasi-nn: --wasi-nn-graph=encoding:target:model_path1:model_path2:...:model_pathN;
2. Add new structs to manage this cmd option: WASINNGlobalContext and wasi_nn_graph_registry;
3. Add 2 new parameters for load_by_name(encoding and target). load_by_name() can get encoding and target from WASINNGlobalContext and set into backend.
4. Update test codes.

### DIFF
--- a/core/iwasm/common/wasm_runtime_common.c
+++ b/core/iwasm/common/wasm_runtime_common.c
@@ -1816,20 +1816,32 @@ wasm_runtime_wasi_nn_registry_set_args(WASINNRegistry *registry,
 
     if ((sizeof(uint32_t *) * n_graphs) >= UINT32_MAX) {
         LOG_ERROR("Invalid size for wasm_runtime_malloc.");
-        return NULL;
+        return false;
     }
 
     registry->n_graphs = n_graphs;
-    registry->target =
-        (uint32_t **)wasm_runtime_malloc(sizeof(uint32_t *) * n_graphs);
-    registry->encoding =
-        (uint32_t **)wasm_runtime_malloc(sizeof(uint32_t *) * n_graphs);
-    registry->loaded =
-        (uint32_t **)wasm_runtime_malloc(sizeof(uint32_t *) * n_graphs);
-    registry->model_names =
-        (uint32_t **)wasm_runtime_malloc(sizeof(uint32_t *) * n_graphs);
-    registry->graph_paths =
-        (uint32_t **)wasm_runtime_malloc(sizeof(uint32_t *) * n_graphs);
+    if (!(registry->target =
+              (uint32_t **)wasm_runtime_malloc(sizeof(uint32_t *) * n_graphs))
+        || !(registry->encoding = (uint32_t **)wasm_runtime_malloc(
+                 sizeof(uint32_t *) * n_graphs))
+        || !(registry->loaded = (uint32_t **)wasm_runtime_malloc(
+                 sizeof(uint32_t *) * n_graphs))
+        || !(registry->model_names =
+                 (char **)wasm_runtime_malloc(sizeof(char *) * n_graphs))
+        || !(registry->graph_paths =
+                 (char **)wasm_runtime_malloc(sizeof(char *) * n_graphs))) {
+        if (registry->target)
+            wasm_runtime_free(registry->target);
+        if (registry->encoding)
+            wasm_runtime_free(registry->encoding);
+        if (registry->loaded)
+            wasm_runtime_free(registry->loaded);
+        if (registry->model_names)
+            wasm_runtime_free(registry->model_names);
+        if (registry->graph_paths)
+            wasm_runtime_free(registry->graph_paths);
+        return false;
+    }
     memset(registry->target, 0, sizeof(uint32_t *) * n_graphs);
     memset(registry->encoding, 0, sizeof(uint32_t *) * n_graphs);
     memset(registry->loaded, 0, sizeof(uint32_t *) * n_graphs);


core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1696,6 +1696,65 @@ wasm_runtime_instantiation_args_destroy(struct InstantiationArgs2 *p)
     wasm_runtime_free(p);
 }
 
+#if WASM_ENABLE_WASI_NN != 0 || WASM_ENABLE_WASI_EPHEMERAL_NN != 0
+WASINNArguments;
```
yamt: ?

core/iwasm/libraries/wasi-nn/include/wasi_nn_types.h
```diff
@@ -48,6 +48,7 @@ typedef enum {
     WASI_NN_ERROR_NAME(unsupported_operation),
     WASI_NN_ERROR_NAME(too_large),
     WASI_NN_ERROR_NAME(not_found),
+    WASI_NN_ERROR_NAME(not_loaded),
```
yamt: please don't make abi changes lightly.

core/iwasm/libraries/wasi-nn/src/wasi_nn_backend.h
```diff
@@ -17,7 +17,8 @@ load(void *ctx, graph_builder_array *builder, graph_encoding encoding,
      execution_target target, graph *g);
 
 __attribute__((visibility("default"))) wasi_nn_error
-load_by_name(void *tflite_ctx, const char *name, uint32_t namelen, graph *g);
+load_by_name(void *tflite_ctx, const char *name, uint32_t namelen,
+             graph_encoding encoding, execution_target target, graph *g);
```
yamt: why do you want to add encoding argument?

core/iwasm/include/wasm_export.h
```diff
@@ -796,6 +798,47 @@ wasm_runtime_instantiation_args_set_wasi_ns_lookup_pool(
     struct InstantiationArgs2 *p, const char *ns_lookup_pool[],
     uint32_t ns_lookup_pool_size);
 
+WASM_RUNTIME_API_EXTERN bool
+wasm_runtime_init_wasi_nn_global_ctx(wasm_module_inst_t module_inst,
```
yamt: please consider to use InstantiationArgs2 instead of adding new user-facing api configuration based on instance.

core/iwasm/common/wasm_runtime_common.h
```diff
@@ -545,6 +545,17 @@ typedef struct WASMModuleInstMemConsumption {
     uint32 exports_size;
 } WASMModuleInstMemConsumption;
 
+#if WASM_ENABLE_WASI_NN != 0 || WASM_ENABLE_WASI_EPHEMERAL_NN != 0
+typedef struct WASINNGlobalContext {
```
yamt: in what sense is this "global"?

core/iwasm/libraries/wasi-nn/src/wasi_nn.c
```diff
@@ -565,17 +607,94 @@ wasi_nn_load_by_name(wasm_exec_env_t exec_env, char *name, uint32_t name_len,
...
 
-    res = success;
+    bool is_loaded = false;
+    uint32 model_idx = 0;
+    char *global_model_path_i;
+    uint32_t global_n_graphs =
+        wasm_runtime_get_wasi_nn_global_ctx_ngraphs(wasi_nn_global_ctx);
+    // Assume filename got from user wasm app : max; sum; average; ...
+    // Assume file path got from user cmd opt: /your/path1/max.tflite;
+    // /your/path2/sum.tflite; ......
```
yamt: i don't fully understand what this super complex logic is doing.
i'd say it's simpler and more flexible to let user give the name explicitly when adding a model to a registry.

product-mini/platforms/posix/main.c
```diff
@@ -974,6 +1024,13 @@ main(int argc, char *argv[])
     libc_wasi_set_init_args(inst_args, argc, argv, &wasi_parse_ctx);
 #endif
 
+#if WASM_ENABLE_WASI_NN != 0 || WASM_ENABLE_WASI_EPHEMERAL_NN != 0
+    wasi_nn_graph_registry_create(&nn_registry);
+    wasi_nn_graph_registry_set_args(nn_registry, encoding, target, n_models,
+                                    model_paths);
```
yamt: it's better to allow multiple --wasi-nn-graph=.
eg. if you have two --wasi-nn-graph=, the registry would contain two models.
```
--wasi-nn-graph=name=modelA,encoding=...,target=....
--wasi-nn-graph=name=modelB,encoding=...,target=....
```
conceptually a registry would be a map indexed by a string.
```
{
    "modelA" -> Model(encodingA, targetA, filelistA, ...),
    "modelB" -> Model(encodingB, targetB, filelistB, ...),
}
```
when a wasm module calls load-by-name("modelA"), it will get Model(encodingA, targetA, filelistA, ...).

product-mini/platforms/posix/main.c
```diff
@@ -825,6 +844,37 @@ main(int argc, char *argv[])
...
         }
+#if WASM_ENABLE_WASI_NN != 0 || WASM_ENABLE_WASI_EPHEMERAL_NN != 0
+        else if (!strncmp(argv[0], "--wasi-nn-graph=", 16)) {
+            char *token;
+            char *saveptr = NULL;
+            int token_count = 0;
+            char *tokens[12] = { 0 };
+
+            // encoding:tensorflowlite|openvino|llama  target:cpu|gpu|tpu
+            // --wasi-nn-graph=encoding:target:model_file_path1:model_file_path2:model_file_path3:......
```
yamt: complex argument parsing logic like this can be in libc_wasi.c so that it's shared among platforms.

core/iwasm/common/wasm_runtime_common.h
SChina-CSC-SDD-Dev: InstantiationArgs2 will be freed soon after wasm_runtime_instantiate_ex2()(in posix/main.c for example), and user cannot use InstantiationArgs2 in wasm app later. 

So I add this new struct to handle these command line options, so load_by_name() can get these information(like target, encoding) through WASINNGlobalContext. 

WASINNGlobalContext will be registered into WASMModuleInstanceExtraCommon->context by wasm_native_set_context() after init.

product-mini/platforms/posix/main.c
zhanheng1: Thank you, the feature now supports multiple --wasi-nn-graphs arguments, and it works exactly as you described in this comment.

product-mini/platforms/posix/main.c
zhanheng1: They are now in libc_wasi.c

core/iwasm/libraries/wasi-nn/src/wasi_nn.c
zhanheng1: The changes have been made as your suggestions.

core/iwasm/libraries/wasi-nn/src/wasi_nn_backend.h
zhanheng1: I noticed that both 'encoding' and 'target' are configurable options, so I incorporated them into the code. Do you think this was inappropriate?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1795,6 +1795,83 @@ wasm_runtime_instantiation_args_set_wasi_ns_lookup_pool(
...
+{
+    if (!registry || !model_names || !encoding || !target || !graph_paths) {
+        return false;
+    }
+
+    registry->n_graphs = n_graphs;
+    registry->target = (uint32_t **)malloc(sizeof(uint32_t *) * n_graphs);
+    registry->encoding = (uint32_t **)malloc(sizeof(uint32_t *) * n_graphs);
+    registry->model_names = (uint32_t **)malloc(sizeof(uint32_t *) * n_graphs);
+    registry->graph_paths = (uint32_t **)malloc(sizeof(uint32_t *) * n_graphs);
```
yamt: please use wasm_runtime_malloc. we don't use malloc/free directly here.

also, check integer overflows.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1795,6 +1795,83 @@ wasm_runtime_instantiation_args_set_wasi_ns_lookup_pool(
...
+    memset(registry->target, 0, sizeof(uint32_t *) * n_graphs);
+    memset(registry->encoding, 0, sizeof(uint32_t *) * n_graphs);
+    memset(registry->model_names, 0, sizeof(uint32_t *) * n_graphs);
+    memset(registry->graph_paths, 0, sizeof(uint32_t *) * n_graphs);
+
+    for (uint32_t i = 0; i < registry->n_graphs; i++) {
+        registry->graph_paths[i] = strdup(graph_paths[i]);
```
yamt: use bh_strdup.

core/iwasm/common/wasm_runtime_common.h
```diff
@@ -612,11 +624,31 @@ WASMExecEnv *
...
+
+    char **graph_paths;
+    uint32_t n_graphs;
+} WASINNArguments;
+
+WASM_RUNTIME_API_EXTERN int
+wasi_nn_graph_registry_create(WASINNArguments **registryp);
+
+WASM_RUNTIME_API_EXTERN void
+wasi_nn_graph_registry_destroy(WASINNArguments *registry);
```
yamt: basically, functions used by iwasm command need to be in wasm_export.h.

core/iwasm/libraries/wasi-nn/src/wasi_nn_backend.h
yamt: in wasi-nn, at least in this implementation, `encoding` is basically same as `backend`.
i don't understand why a backend needs to be told its own encoding.

core/iwasm/libraries/wasi-nn/include/wasi_ephemeral_nn.h
```diff
@@ -7,6 +7,3 @@
 #define WASI_NN_NAME(name) wasi_ephemeral_nn_##name
 
 #include "wasi_nn.h"
-
-#undef WASM_ENABLE_WASI_EPHEMERAL_NN
-#undef WASI_NN_NAME
```
yamt, line 12: why?

core/iwasm/libraries/wasi-nn/test/utils.c
```diff
@@ -5,17 +5,16 @@
...
 #include "utils.h"
 #include "logger.h"
-#include "wasi_nn.h"
-
 #include <stdio.h>
 #include <stdlib.h>
 
-wasi_nn_error
-wasm_load(char *model_name, graph *g, execution_target target)
+WASI_NN_ERROR_TYPE
```
yamt: WASI_NN_ERROR_TYPE etc are implementation details.
they are not expected to be used in user programs like this.

product-mini/platforms/common/libc_wasi.c
```diff
@@ -177,3 +185,75 @@ libc_wasi_set_init_args(struct InstantiationArgs2 *args, int argc, char **argv,
...
+        }
+    }
+
+    if (token_count != 4) {
+        ret = LIBC_WASI_PARSE_RESULT_NEED_HELP;
+        printf("4 arguments are needed for wasi-nn.\n");
+        goto fail;
+    }
+
+    ctx->model_names[ctx->n_graphs] = strdup(tokens[0]);
```
yamt: why do you need strdup here?
argv is stable during the lifetime of this process, isn't it?

core/iwasm/libraries/wasi-nn/src/wasi_nn.c
```diff
@@ -565,17 +607,76 @@ wasi_nn_load_by_name(wasm_exec_env_t exec_env, char *name, uint32_t name_len,
...
+        char *global_model_path_i =
+            wasm_runtime_get_wasi_nn_global_ctx_graph_paths_i(
+                wasi_nn_global_ctx, model_idx);
+
+        graph_encoding encoding =
+            str2encoding(wasm_runtime_get_wasi_nn_global_ctx_encoding_i(
+                wasi_nn_global_ctx, model_idx));
+        execution_target target =
+            str2target(wasm_runtime_get_wasi_nn_global_ctx_target_i(
+                wasi_nn_global_ctx, model_idx));
```
yamt: i guess it's simpler to perform these conversions earlier. eg. in wasi_nn_graph_registry_set_args

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1795,6 +1795,83 @@ wasm_runtime_instantiation_args_set_wasi_ns_lookup_pool(
...
+            }
+        free(registry);
+    }
+}
+
+void
+wasm_runtime_instantiation_args_set_wasi_nn_graph_registry(
+    struct InstantiationArgs2 *p, WASINNArguments *registry)
+{
+    p->nn_registry = *registry;
```
yamt: structure copy? is this intended?

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -3300,6 +3300,18 @@ wasm_instantiate(WASMModule *module, WASMModuleInstance *parent,
...
 #endif
 
+#if WASM_ENABLE_WASI_NN != 0 || WASM_ENABLE_WASI_EPHEMERAL_NN != 0
+    /* Store graphs' path into ctx. Graphs will be loaded until user app calls
+     * load_by_name */
+    WASINNArguments *nn_registry = &args->nn_registry;
+    if (!wasm_runtime_init_wasi_nn_global_ctx(
+            (WASMModuleInstanceCommon *)module_inst, nn_registry->model_names,
+            nn_registry->encoding, nn_registry->target, nn_registry->n_graphs,
+            nn_registry->graph_paths, error_buf, error_buf_size)) {
```
yamt: i suppose it's simpler to pass a structure pointer than copying every members of the structure.

core/iwasm/common/wasm_runtime_common.h
yamt: it's a bit confusing as this structure is not really global. (vs our global state initialized by wasm_runtime_init.)
it's merely per instance (or a group of instances in case of wasi-threads), right?

i'd suggest to unify WASINNGlobalContext and WASINNArguments into a single structure and name it, say, WASINNRegistry.

core/iwasm/include/wasm_export.h
```diff
@@ -796,6 +798,37 @@ wasm_runtime_instantiation_args_set_wasi_ns_lookup_pool(
...
+wasm_runtime_get_wasi_nn_global_ctx_graph_paths_i(
+    WASINNGlobalContext *wasi_nn_global_ctx, uint32_t idx);
+
+WASM_RUNTIME_API_EXTERN uint32_t
+wasm_runtime_get_wasi_nn_global_ctx_loaded_i(
+    WASINNGlobalContext *wasi_nn_global_ctx, uint32_t idx);
+
+WASM_RUNTIME_API_EXTERN uint32_t
+wasm_runtime_set_wasi_nn_global_ctx_loaded_i(
+    WASINNGlobalContext *wasi_nn_global_ctx, uint32_t idx, uint32_t value);
```
yamt: these functions seems internal implementation details.
do you really want them to expose to users?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -8080,3 +8157,146 @@ wasm_runtime_check_and_update_last_used_shared_heap(
...
+
+uint32_t
+wasm_runtime_get_wasi_nn_global_ctx_loaded_i(
+    WASINNGlobalContext *wasi_nn_global_ctx, uint32_t idx)
+{
+    if (wasi_nn_global_ctx && (idx < wasi_nn_global_ctx->n_graphs))
+        return wasi_nn_global_ctx->loaded[idx];
+
+    return -1;
+}
```
yamt: i'd suggest to move majority of these wasi-nn specific stuff to libraries/wasi-nn/src.

core/iwasm/libraries/wasi-nn/src/wasi_nn.c
```diff
@@ -565,17 +607,76 @@ wasi_nn_load_by_name(wasm_exec_env_t exec_env, char *name, uint32_t name_len,
...
+    bool is_loaded = false;
+    uint32 model_idx = 0;
+    uint32_t global_n_graphs =
+        wasm_runtime_get_wasi_nn_global_ctx_ngraphs(wasi_nn_global_ctx);
+    for (model_idx = 0; model_idx < global_n_graphs; model_idx++) {
+        char *model_name = wasm_runtime_get_wasi_nn_global_ctx_model_names_i(
+            wasi_nn_global_ctx, model_idx);
 
-    res = success;
+        if (model_name && strcmp(nul_terminated_name, model_name) != 0) {
```
yamt: in what case model_name is NULL here?
isn't simpler to reject it earlier?

core/iwasm/libraries/wasi-nn/src/wasi_nn.c
```diff
@@ -565,17 +607,76 @@ wasi_nn_load_by_name(wasm_exec_env_t exec_env, char *name, uint32_t name_len,
...
+        // No enlarge for now
+        NN_ERR_PRINTF("No enough space for new model");
+        res = too_large;
+    }
+    else if (model_idx >= global_n_graphs) {
+        NN_ERR_PRINTF("Model %s is not loaded, you should pass its path "
+                      "through --wasi-nn-graph",
+                      nul_terminated_name);
+        res = not_found;
+    }
```
yamt, line 614: i suppose wasi_nn_load_by_name_with_config needs an equivalent.
i guess it's simpler to make them share a common logic.

core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -3300,6 +3300,18 @@ wasm_instantiate(WASMModule *module, WASMModuleInstance *parent,
     }
 #endif
 
+#if WASM_ENABLE_WASI_NN != 0 || WASM_ENABLE_WASI_EPHEMERAL_NN != 0
+    /* Store graphs' path into ctx. Graphs will be loaded until user app calls
+     * load_by_name */
```
yamt: i suppose aot needs an equivalent.

core/iwasm/libraries/wasi-nn/src/wasi_nn.c
yamt: that way we can reject invalid/unsupported things earlier.

core/iwasm/include/wasm_export.h
yamt: also, honestly speaking, i feel these accessor functions are excessive, if they are only used within wasi-nn implementation.

core/iwasm/libraries/wasi-nn/include/wasi_ephemeral_nn.h
zhanheng1: In wasi_ephemeral_nn.h, macros are defined and then wasi_nn.h is included immediately after, but these macro definitions are later undefined (#undef). This means that in utils.h, the macros might be in an undefined state, leading to compilation errors( if WASM_ENABLE_WASI_EPHEMERAL_NN != 0).

core/iwasm/common/wasm_runtime_common.h
zhanheng1: > it's merely per instance (or a group of instances in case of wasi-threads), right?

    Yes, WASINNGlobalContext is per instance.

> i'd suggest to unify WASINNGlobalContext and WASINNArguments into a single structure and name it, say, WASINNRegistry.

    Does this mean that the new data structure WASINNRegistry is still a member of InstantiationArgs2, but will not be freed together with InstantiationArgs2—requiring users to manually destroy it, for example, in posix/main.c? 
    After all, InstantiationArgs2 will be released before the WASM program runs, so the WASM program will not be able to locate InstantiationArgs2 during its execution.

core/iwasm/common/wasm_runtime_common.c
zhanheng1: Done

core/iwasm/common/wasm_runtime_common.c
zhanheng1: Done.

core/iwasm/common/wasm_runtime_common.c
zhanheng1: Fixed.

core/iwasm/interpreter/wasm_runtime.c
zhanheng1: Done

product-mini/platforms/common/libc_wasi.c
zhanheng1: Removed.

core/iwasm/libraries/wasi-nn/src/wasi_nn_backend.h
zhanheng1: Removed.

core/iwasm/libraries/wasi-nn/test/utils.c
zhanheng1: WASI_NN_ERROR_TYPE in utils.h/utils.c now replace to a specific new value type instead of macro.

core/iwasm/common/wasm_runtime_common.h
zhanheng1: Done.

core/iwasm/include/wasm_export.h
zhanheng1: Removed.

core/iwasm/common/wasm_runtime_common.c
zhanheng1: Done.

core/iwasm/common/wasm_runtime_common.h
zhanheng1: Done. WASINNGlobalContext and WASINNArguments are unified into a single structure WASINNRegistry.

core/iwasm/libraries/wasi-nn/src/wasi_nn.c
zhanheng1: Removed.

core/iwasm/interpreter/wasm_runtime.c
zhanheng1: Done

core/iwasm/libraries/wasi-nn/src/wasi_nn.c
zhanheng1: Done.

core/iwasm/libraries/wasi-nn/src/wasi_nn.c
zhanheng1: Done.

core/iwasm/libraries/wasi-nn/include/wasi_ephemeral_nn.h
yamt: wasi_ephemeral_nn.h header is meant to be used by applications which uses this version of api.
these macros are not meant to be used by applications. (WASM_ENABLE_WASI_EPHEMERAL_NN, WASI_NN_NAME, WASI_NN_ERROR_TYPE, etc)
they are supposed to use eg. wasi_ephemeral_nn_error.
see samples like wamr-wasi-extensions/samples/nn-cli/main.c

core/iwasm/libraries/wasi-nn/test/build.sh
```diff
@@ -22,6 +22,8 @@ CURR_PATH=$(cd $(dirname $0) && pwd -P)
 
 /opt/wasi-sdk/bin/clang \
     --target=wasm32-wasi \
+    -DWASM_ENABLE_WASI_NN=1 \
+    -DWASM_ENABLE_WASI_EPHEMERAL_NN=1 \
```
yamt: why do you enable WASM_ENABLE_WASI_EPHEMERAL_NN for this, but not for test_tensorflow_quantized.c the below?

have you run these tests at all?
if not, i'd suggest to leave these to use the legacy api for now.

core/iwasm/common/wasm_runtime_common.h
```diff
@@ -775,6 +798,20 @@ wasm_runtime_instantiation_args_set_wasi_ns_lookup_pool(
...
+wasm_runtime_instantiation_args_set_wasi_nn_registry(
+    struct InstantiationArgs2 *p, WASINNRegistry *registry);
+
+WASM_RUNTIME_API_EXTERN bool
+wasm_runtime_wasi_nn_registry_set_args(WASINNRegistry *registry,
+                                       const char **model_names,
+                                       const uint32_t **encoding,
+                                       const uint32_t **target,
+                                       uint32_t n_graphs,
+                                       const char **graph_paths);
```
yamt, line 812: i suppose these should be in wasm_export.h as it's used by the embedding applications like iwasm.

core/iwasm/include/wasm_export.h
```diff
@@ -796,6 +797,19 @@ wasm_runtime_instantiation_args_set_wasi_ns_lookup_pool(
     struct InstantiationArgs2 *p, const char *ns_lookup_pool[],
     uint32_t ns_lookup_pool_size);
 
+WASM_RUNTIME_API_EXTERN struct WASINNRegistry *
+wasm_runtime_get_wasi_nn_registry(const wasm_module_inst_t module_inst);
+
+WASM_RUNTIME_API_EXTERN void
+wasm_runtime_set_wasi_nn_registry(wasm_module_inst_t module_inst,
+                                  struct WASINNRegistry *wasi_ctx);
```
yamt: wasm_runtime_get_wasi_nn_registry/wasm_runtime_set_wasi_nn_registry are not supposed to be used by applications, are they?
i guess they should not be in wasm_export.h.

in wasm_export.h, put only api which is used directly by application. (eg. iwasm)
internal api should go to somewhere else, like wasm_runtime_common.h

core/iwasm/common/wasm_runtime_common.h
```diff
@@ -1427,6 +1464,19 @@ wasm_runtime_check_and_update_last_used_shared_heap(
...
+WASM_RUNTIME_API_EXTERN void
+wasm_runtime_set_wasi_nn_registry(WASMModuleInstanceCommon *module_inst,
+                                  WASINNRegistry *wasi_ctx);
+
+WASM_RUNTIME_API_EXTERN WASINNRegistry *
+wasm_runtime_get_wasi_nn_registry(WASMModuleInstanceCommon *module_inst_comm);
+
+WASM_RUNTIME_API_EXTERN void
+wasm_runtime_set_wasi_nn_registry(WASMModuleInstanceCommon *module_inst_comm,
+                                  WASINNRegistry *wasi_nn_ctx);
```
yamt, line 1473: two duplicated wasm_runtime_set_wasi_nn_registry prototypes in this file.

product-mini/platforms/posix/main.c
```diff
@@ -977,12 +1012,18 @@ main(int argc, char *argv[])
...
         wasm_module, inst_args, error_buf, sizeof(error_buf));
-    wasm_runtime_instantiation_args_destroy(inst_args);
     if (!wasm_module_inst) {
         printf("%s\n", error_buf);
         goto fail3;
     }
 
+#if WASM_ENABLE_WASI_NN != 0 || WASM_ENABLE_WASI_EPHEMERAL_NN != 0
+    wasm_runtime_wasi_nn_registry_create(&nn_registry);
+    wasi_nn_set_init_args(inst_args, nn_registry, &wasi_nn_parse_ctx);
```
yamt: wasi_nn_set_init_args vs inst_args look a bit confusing.
maybe you meant wasi_nn_set_inst_args?

core/iwasm/libraries/wasi-nn/test/build.sh
zhanheng1: > why do you enable WASM_ENABLE_WASI_EPHEMERAL_NN for this, but not for test_tensorflow_quantized.c the below?

Just forgot to enable them for test_tensorflow_quantized.c, thanks.

> have you run these tests at all?

Yes, I have tested them all with WASM_ENABLE_WASI_EPHEMERAL_NN enabled.

product-mini/platforms/posix/main.c
zhanheng1: Fixed.

core/iwasm/common/wasm_runtime_common.h
zhanheng1: Fixed.

core/iwasm/include/wasm_export.h
zhanheng1: Fixed.

core/iwasm/common/wasm_runtime_common.h
zhanheng1: Fixed.

core/iwasm/libraries/wasi-nn/include/wasi_ephemeral_nn.h
zhanheng1: Fixed.

core/iwasm/libraries/wasi-nn/include/wasi_nn_types.h
```diff
@@ -169,6 +169,7 @@ typedef enum WASI_NN_NAME(execution_target) {
     WASI_NN_TARGET_NAME(cpu) = 0,
     WASI_NN_TARGET_NAME(gpu),
     WASI_NN_TARGET_NAME(tpu),
+    WASI_NN_TARGET_NAME(unsupported_target),
```
yamt: where does this (unsupported_target) come from?
i feel uneasy to add wamr-specific extension.

core/iwasm/libraries/wasi-nn/test/build.sh
yamt: > Yes, I have tested them all with WASM_ENABLE_WASI_EPHEMERAL_NN enabled.

iirc, test_tensorflow_quantized.c is not supposed to work with WASM_ENABLE_WASI_EPHEMERAL_NN because it uses wamr-specific extension which we disabled for WASM_ENABLE_WASI_EPHEMERAL_NN.
cf. https://github.com/bytecodealliance/wasm-micro-runtime/pull/4517

do you mean just build-tested?

zhanheng1: ①If we build **WAMR** with both **WASM_ENABLE_WASI_EPHEMERAL_NN=1** and **WASM_ENABLE_WASI_NN=1**(default):
1. For test_tensorflow.c, WASM can get correct output while WASM_ENABLE_WASI_EPHEMERAL_NN==1, **with modifications in utils.c/utils.h in this PR**.

2. But for  test_tensorflow_quantized.c, yes just build-tested. WASM will report type errors for input while WASM_ENABLE_WASI_EPHEMERAL_NN==1.

② If we build **WAMR** with **WASM_ENABLE_WASI_EPHEMERAL_NN=0**(User must explicitly set this variable to 0.)  and **WASM_ENABLE_WASI_NN=1**:
1. For both test_tensorflow.c and test_tensorflow_quantized.c, with WASM_ENABLE_WASI_EPHEMERAL_NN==0, **with legacy apis in utils.c/utils.h**, WASM can get correct output;

So your advice is to keep the legacy APIs so that the tests pass under the second condition② described above？ If users wish to test the first case① after, they have to rewrite a test program themselves.

core/iwasm/libraries/wasi-nn/include/wasi_nn_types.h
zhanheng1: Um, that's an enum value we defined ourselves. I checked the wasi-nn proposal, and it's indeed not supported. I'll remove it.

core/iwasm/libraries/wasi-nn/test/build.sh
zhanheng1: Revert test files to the original version.

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1833,6 +1795,109 @@ wasm_runtime_instantiation_args_set_wasi_ns_lookup_pool(
...
+
+    registry->n_graphs = n_graphs;
+    registry->target =
+        (uint32_t **)wasm_runtime_malloc(sizeof(uint32_t *) * n_graphs);
+    registry->encoding =
+        (uint32_t **)wasm_runtime_malloc(sizeof(uint32_t *) * n_graphs);
+    registry->loaded =
+        (uint32_t **)wasm_runtime_malloc(sizeof(uint32_t *) * n_graphs);
+    registry->model_names =
+        (uint32_t **)wasm_runtime_malloc(sizeof(uint32_t *) * n_graphs);
```
yamt: model_names and graph_paths are char *, not uint32_t.
doesn't your compiler complain?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -1833,6 +1795,109 @@ wasm_runtime_instantiation_args_set_wasi_ns_lookup_pool(
...
+    }
+
+    if ((sizeof(uint32_t *) * n_graphs) >= UINT32_MAX) {
+        LOG_ERROR("Invalid size for wasm_runtime_malloc.");
+        return NULL;
+    }
+
+    registry->n_graphs = n_graphs;
+    registry->target =
+        (uint32_t **)wasm_runtime_malloc(sizeof(uint32_t *) * n_graphs);
```
yamt: should check integer overflow and allocation failure.

product-mini/platforms/posix/main.c
```diff
@@ -977,12 +1012,18 @@ main(int argc, char *argv[])
...
         wasm_module, inst_args, error_buf, sizeof(error_buf));
-    wasm_runtime_instantiation_args_destroy(inst_args);
     if (!wasm_module_inst) {
         printf("%s\n", error_buf);
         goto fail3;
     }
 
+#if WASM_ENABLE_WASI_NN != 0 || WASM_ENABLE_WASI_EPHEMERAL_NN != 0
+    wasm_runtime_wasi_nn_registry_create(&nn_registry);
+    wasi_nn_set_inst_args(inst_args, nn_registry, &wasi_nn_parse_ctx);
```
yamt, line 1022: shouldn't this be done before inst_args is used? ie. before wasm_runtime_instantiate_ex2

product-mini/platforms/posix/main.c
```diff
@@ -977,12 +1012,18 @@ main(int argc, char *argv[])
...
+    wasm_runtime_set_wasi_nn_registry(wasm_module_inst, nn_registry);
```
yamt, line 1023: wasm_runtime_set_wasi_nn_registry is not supposed to be used by the embedder, is it?

i suppose the registry is passed via inst_args.

product-mini/platforms/posix/main.c
zhanheng1: If I recall correctly, we previously discussed this issue. 
Since wasm_runtime_malloc() requires wasm_runtime_instantiate_ex2() to initialize the memory first, this logic should either be implemented inside wasm_runtime_instantiate_ex2(), or extracted as a separate function to be called after wasm_runtime_instantiate_ex2(). 
Originally, the function was placed in the middle of wasm_runtime_instantiate_ex2(), but it was later suggested that extracting it as a standalone function would be better, which is why the current version was adopted.

core/iwasm/common/wasm_runtime_common.c
zhanheng1: changed to char* now.

core/iwasm/common/wasm_runtime_common.c
zhanheng1: Checks has been added now.

product-mini/platforms/posix/main.c
zhanheng1: If I remember correctly, we've also discussed this issue before. 
Since inst_args is destroyed immediately in the next line of code, the WASM application cannot actually use it. 
The current version stores the registry in a slot within each WASM module's context, allowing the WASM app to retrieve the registry via the module's context. 
Therefore, wasm_runtime_set_wasi_nn_registry should be explicitly called in main.c.

product-mini/platforms/posix/main.c
yamt: i don't remember the discussion. was i involved?

`wasm_runtime_malloc` doesn't require instance memory. do you mean `wasm_module_malloc`?
i think it's fine to use `wasm_runtime_malloc` to allocate the registry. after all, the registry doesn't necessarily belong to a specific module or instance. i vaguely remember we had a brief discussion about possible use cases to shared a registry among instances.

i remember i suggested to make `wasm_runtime_instantiation_args_set_wasi_nn_registry` just assign a pointer.
that is,
```c
struct InstantiationArgs2 {
   ....
#if WASM_ENABLE_WASI_NN != 0 || WASM_ENABLE_WASI_EPHEMERAL_NN != 0
    WASINNRegistry *nn_registry;
#endif
}

void
wasm_runtime_instantiation_args_set_wasi_nn_registry(
    struct InstantiationArgs2 *p, WASINNRegistry *registry)
{
    p->nn_registry = registry;
}
```
and make `wasm_instantiate` and `aot_instantiate` call `wasm_runtime_set_wasi_nn_registry` internally to assign `args->nn_registry` to the new instance.

product-mini/platforms/posix/main.c
yamt: inst_args is, as its name suggests, only for wasm_runtime_instantiate_ex2.
it's expected necessary info (in this case a pointer to the registry object)
is copied during instantiation.

PR #4772, opened 2025-12-23
Move exception lock to ModuleInstance data.

vchigrin:
This lock acquired on each native function call. This cause performance impact on programs, containing many native function calls, and running in multithreaded environment.

### DIFF
--- a/core/iwasm/libraries/thread-mgr/thread_manager.c
+++ b/core/iwasm/libraries/thread-mgr/thread_manager.c
@@ -1530,31 +1530,16 @@ wasm_cluster_is_thread_terminated(WASMExecEnv *exec_env)
     return is_thread_terminated;
 }
 
-static WASMModuleInstanceExtraCommon *
-GetModuleinstanceCommon(WASMModuleInstance *module_inst)
-{
-#if WASM_ENABLE_AOT != 0
-    if (module_inst->module_type == Wasm_Module_AoT) {
-        return &((AOTModuleInstanceExtra *)module_inst->e)->common;
-    }
-    else {
-        return &module_inst->e->common;
-    }
-#else
-    return &module_inst->e->common;
-#endif
-}
-
 void
 exception_lock(WASMModuleInstance *module_inst)
 {
-    os_mutex_lock(&GetModuleinstanceCommon(module_inst)->exception_lock);
+    os_mutex_lock(&GetModuleInstanceExtraCommon(module_inst)->exception_lock);
 }
 
 void
 exception_unlock(WASMModuleInstance *module_inst)
 {
-    os_mutex_unlock(&GetModuleinstanceCommon(module_inst)->exception_lock);
+    os_mutex_unlock(&GetModuleInstanceExtraCommon(module_inst)->exception_lock);
 }
 
 void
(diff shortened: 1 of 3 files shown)


lum1n0us (changes requested):
- The PR has affected the layout of `WASMModuleInstanceExtra`, which will change key `offsetof` values in AOT files. Therefore, it requires a new `AOT_CURRENT_VERSION` in core/config.h.

- Obviously, there is a failed CI that should be fixed.

vchigrin:
Thank you so much! Fixed both issues, all CI checks are green now.

core/iwasm/libraries/thread-mgr/thread_manager.c
```diff
@@ -1537,22 +1530,31 @@ wasm_cluster_is_thread_terminated(WASMExecEnv *exec_env)
     return is_thread_terminated;
 }
 
+static WASMModuleInstanceExtraCommon *
+GetModuleinstanceCommon(WASMModuleInstance *module_inst)
```
lum1n0us: I would prefer to add this function into wasm_runtime_common.h and/or wasm_runtime_common.c.

test-tools/aot-analyzer/include/config.h
```diff
@@ -15,7 +15,7 @@
 #define WASM_CURRENT_VERSION 1
 
 #define AOT_MAGIC_NUMBER 0x746f6100
-#define AOT_CURRENT_VERSION 5
+#define AOT_CURRENT_VERSION 6
```
lum1n0us: May I ask why it needs a bump?

vchigrin: Sorry, my bad. Did not notice same definition in multiple files. Updated PR, now version changed in `core/config.h`, as you recommended.

vchigrin:
Thank you so much for review. 
Just out of curiosity, are these changes really break compatibility (and thus require AOT_CURRENT_VERSION bump)? I doubted a bit for two reasons:
1. [here](https://github.com/bytecodealliance/wasm-micro-runtime/blob/46472ee6c849382f4cb54db75b40bd9492bb6e9b/core/iwasm/aot/aot_runtime.c#L59-L64) are some static asserts that check offsets of some fields in `AOTModuleInstanceExtra`. When I started to work on this patch I thought that "preserving these offsets is all what required for binary compatibility. If these assertions did not fire, then version bump is not necessary". May be that assumption was wrong, though.
2. One recent PR [also changed](https://github.com/bytecodealliance/wasm-micro-runtime/pull/4400/changes#diff-1e917a6d1997c03d2c44ff72e6e5e117835fdf9cc6320e42acd4089a66ada8ecR119) layout of `AOTModuleInstanceExtra`, but it did not change `AOT_CURRENT_VERSION`.

Of course, you, as code owner, know better all nuances. I just write to ensure that we'll not force our users re-compile all their AOT programs without need.

lum1n0us:
The basic relationship is that `WASMModuleInstanceExtraCommon common` is a field within `AOTModuleInstanceExtra`. Subsequent fields, such as `functions`, will have their offsetof values changed. For protection, we tend to bump AOT_VER to let the AOT loader be aware of that.

PR#4400 was likely overlooked. Currently, the entire review system relies on manual review. Perhaps we could consider introducing other tools to assist.

PR #4773, opened 2025-12-26
update Dockerfile and devcontainer configuration

lum1n0us:
The goal is to keep the devcontainer image minimal and its build time as fast as possible.
- Remove all rarely used tools.
- Change the base image to the latest template.
- Synchronize the WASI-SDK and WABT versions with those used by the CI.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,5 +1,5 @@
 {
-  "name": "WAMR Dev",
+  "name": "WAMR-Dev",
   "build": {
     "dockerfile": "Dockerfile",
     "args": {


.devcontainer/Dockerfile
```diff
@@ -1,150 +1,43 @@
...
-ARG EMSDK_VER=3.1.43
-WORKDIR /opt/emsdk
-RUN  git pull \
-  && ./emsdk install ${EMSDK_VER} \
-  && ./emsdk activate ${EMSDK_VER} \
-  && echo "source /opt/emsdk/emsdk_env.sh" >> /root/.bashrc
-
-#
-# install wasi-sdk
 ARG WASI_SDK_VER=20
```
TianlongLiang: should we use matain the consistent version between `Dockerfile` and `devcontainer.json` for WASI SDK and WABT?

TianlongLiang (dismissed):
LGTM with a minor comment

.devcontainer/Dockerfile
lum1n0us: Technically, it's not necessary to maintain the same version, as the JSON value will override it like build-arg. However, you are correct that there is no way to know if someone is using the Dockerfile directly.

Issue #3164, opened 2024-02-16
AOT compiler `wamrc` doesn't support compiling to arm64 bit


penneryu:
penner@xxx build % ./wamrc --target=aarch64 -o wasm.aot wasm.wasm 
[18:08:57:013 - 1DEF6D000]: warning: a module with WASI apis should be either a command or a reactor
Create AoT compiler with:
  target:        aarch64v8
  target cpu:    
  target triple: aarch64v8-pc-linux-gnu
  cpu features:  
  opt level:     3
  size level:    1
  output format: AoT file
LLVM ERROR: Only small, tiny and large code models are allowed on AArch64
zsh: abort      ./wamrc --target=aarch64 -o wasm.aot wasm.wasm

Can I modify the source code in aot_llvm_extra2.cpp

TianlongLiang:
Hi, can you tell me which platform and which commit you are using? 
I tried it on my x86 Ubuntu20 with latest commit, it's fine:
```shell
./wamrc --target=aarch64 -o test.aot test.wasm
Create AoT compiler with:
  target:        aarch64v8
  target cpu:    
  target triple: aarch64v8-pc-linux-gnu
  cpu features:  
  opt level:     3
  size level:    3
  output format: AoT file
Compile success, file test.aot was generated.
```

penneryu:
```
#if defined(_WIN32) || defined(_WIN32_) || defined(__APPLE__) \
    || defined(__MACH__)
        if (!option.target_abi) {
            LOG_VERBOSE("Set size level to 1 for Windows or MacOS AOT file");
            option.size_level = 1;
        }
#endif
    }
```
I found the problem, we should differentiate apple m1 chip

TianlongLiang:
Good idea, maybe we should differentiate intel-based and arm-based MacOS. For now, you can try adding a command line argument to set manually set the size-level now to see whether it works:

```shell
./wamrc --size-level=3 --target=aarch64 -o test.aot test.wasm
```

eloparco:
For AOT on M1, usually I compile with `wamrc --size-level=0 -o wasm.aot wasm.wasm` (as suggested by the LLVM error `LLVM ERROR: Only small, tiny and large code models are allowed on AArch64`):
```bash
Create AoT compiler with:
  target:        arm64
  target cpu:    apple-m1
  target triple: arm64-pc-linux-gnu
  cpu features:  
  opt level:     3
  size level:    0
  output format: AoT file
[...]

PR #4775, opened 2025-12-29
fix(aot): reserve x18 register on macOS ARM64

teamchong:
## Problem

Apple reserves CPU register x18 for TLS on ARM64. When generating AOT code for aarch64 on macOS, LLVM may use x18, causing crashes when the AOT code runs on macOS ARM64 (M1/M2/M3).

## Solution

This patch makes three changes to properly handle macOS ARM64:

1. **Detect darwin/macho ABI** - When explicit ABI is specified as "darwin" or "macho", use `-apple-` vendor string
2. **Detect darwin/apple in default triple** - When auto-detecting platform, check for "darwin" or "apple" in the triple
3. **Add +reserve-x18 feature** - Tell LLVM to reserve x18 register when targeting aarch64 on macOS hosts

```c
#if (defined(__APPLE__) || defined(__MACH__)) && defined(BUILD_TARGET_AARCH64)
    if (is_aarch64) {
        // Append +reserve-x18 to features
    }
#endif
```

## Testing

- AOT compilation and execution on macOS ARM64 (M1/M2/M3) works correctly
- No crashes from x18 register conflicts

## Related

- https://github.com/bytecodealliance/wasm-micro-runtime/issues/3164
- https://github.com/bytecodealliance/wasm-micro-runtime/discussions/1944

### DIFF
--- a/tests/standalone/test-aot-x18-reserve/run.sh
+++ b/tests/standalone/test-aot-x18-reserve/run.sh
@@ -74,14 +74,14 @@ fi
 
 if [[ $1 != "--aot" ]]; then
     echo "============> run stress_registers.wasm (interpreter mode)"
-    echo "Running 100 iterations in interpreter mode..."
-    for i in $(seq 1 100); do
+    echo "Running 1000 iterations in interpreter mode..."
+    for i in $(seq 1 1000); do
         if ! ${IWASM_CMD} stress_registers.wasm 2>&1; then
             echo "FAILED: Crash at iteration $i"
             exit 1
         fi
     done
-    echo "PASSED: 100 iterations completed without crash"
+    echo "PASSED: 1000 iterations completed without crash"
 else
     echo "============> compile stress_registers.wasm to AOT"
 


lum1n0us:
Thank you for reporting that. 

Can you provide a test case to expose or verify the issue? If I understand correctly, macOS-15 with M1 (as mentioned in the [GitHub documentation](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories)) is a suitable platform, right?

teamchong:
> Thank you for reporting that.
> 
> Can you provide a test case to expose or verify the issue? If I understand correctly, macOS-15 with M1 (as mentioned in the [GitHub documentation](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories)) is a suitable platform, right?

I've added a test case that:
  1. Creates a WASM module that stresses register allocation (24 local variables)
  2. Compiles to AOT with -O3 optimization
  3. Runs 1000 iterations to verify no crashes

  Test results on M2 Mac:
  - Without fix: 828/1000 crash rate (SIGSEGV)
  - With fix: 0/1000 crash rate

  The test only runs on macOS ARM64 (macos-15 runner) and is skipped on other platforms.

PR #4778, opened 2026-01-04
fix: Wrong compiler parameter on MSVC

sbwtw:
fix wrong compiler param `-mindirect-branch-register` on MSVC

### DIFF
--- a/wamr-compiler/CMakeLists.txt
+++ b/wamr-compiler/CMakeLists.txt
@@ -328,7 +328,7 @@ endif ()
 #TODO: sync up WAMR_BUILD_SANITIZER in config_common.cmake
 # set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wconversion -Wsign-conversion")
 if (WAMR_BUILD_TARGET MATCHES "X86_.*" OR WAMR_BUILD_TARGET STREQUAL "AMD_64")
-  if (CMAKE_C_COMPILER MATCHES ".*gcc.*" OR CMAKE_C_COMPILER_ID MATCHES ".*GNU")
+  if (CMAKE_C_COMPILER_ID MATCHES ".*GNU")
     set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -mindirect-branch-register")
     # UNDEFINED BEHAVIOR, refer to https://en.cppreference.com/w/cpp/language/ub
     if(CMAKE_BUILD_TYPE STREQUAL "Debug" AND NOT WAMR_BUILD_JIT EQUAL 1)


lum1n0us (commented):
The `-mndirect-branch-register` is a Clang-specific option. I would suggest a shorter form like the one below.

```
if (CMAKE_C_COMPILER MATCHES ".*clang.*" OR CMAKE_C_COMPILER_ID MATCHES ".*Clang")
  set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -mindirect-branch-register")
endif ()
```

It seems we should reduce the repetition of this code segment. I suggest moving this piece of code(the optimized version) to config_common.cmake. I haven't tested this idea before, so please let me know if it doesn't work if you decide to give it a try.

sbwtw:
> The `-mndirect-branch-register` is a Clang-specific option. I would suggest a shorter form like the one below.
> 
> ```
> if (CMAKE_C_COMPILER MATCHES ".*clang.*" OR CMAKE_C_COMPILER_ID MATCHES ".*Clang")
>   set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -mindirect-branch-register")
> endif ()
> ```
> 
> It seems we should reduce the repetition of this code segment. I suggest moving this piece of code(the optimized version) to config_common.cmake. I haven't tested this idea before, so please let me know if it doesn't work if you decide to give it a try.

I moved these codes to config_common, and i found clang doesn't support `-mindirect-branch-register`. it's gcc only.

lum1n0us:
>  it's gcc only.

You are right. My mistaken.

build-scripts/config_common.cmake
```diff
@@ -79,8 +79,13 @@ if (WAMR_BUILD_TARGET MATCHES "ARM.*")
 elseif (WAMR_BUILD_TARGET MATCHES "THUMB.*")
   set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -mthumb")
   set (CMAKE_ASM_FLAGS "${CMAKE_ASM_FLAGS} -Wa,-mthumb")
+elseif (WAMR_BUILD_TARGET MATCHES "X86_.*" OR WAMR_BUILD_TARGET STREQUAL "AMD_64")
+    if (CMAKE_C_COMPILER MATCHES ".*gcc.*" OR CMAKE_C_COMPILER_ID MATCHES ".*GNU")
```
lum1n0us: `CMAKE_C_COMPILER` is a path for the compiler and might be set to `/usr/bin/cc` or link to others. Using `CMAKE_C_COMPILER_ID` is more accurate.

```suggestion
    if (CMAKE_C_COMPILER_ID STREQUAL "GNU")
```

build-scripts/config_common.cmake
```diff
@@ -79,8 +79,13 @@ if (WAMR_BUILD_TARGET MATCHES "ARM.*")
...
+      set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -mindirect-branch-register")
```
lum1n0us, line 84: suggest applying the same modification to cover `CMAKE_CXX_FLAGS`.

```
if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
  # ...
endif ()
```

product-mini/platforms/linux/CMakeLists.txt
```diff
@@ -140,9 +140,8 @@ check_pie_supported()
 set (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,--gc-sections")
 
 if (WAMR_BUILD_TARGET MATCHES "X86_.*" OR WAMR_BUILD_TARGET STREQUAL "AMD_64")
-  if (NOT (CMAKE_C_COMPILER MATCHES ".*clang.*" OR CMAKE_C_COMPILER_ID MATCHES ".*Clang"))
+  if (CMAKE_C_COMPILER MATCHES ".*gcc.*" OR CMAKE_C_COMPILER_ID MATCHES ".*GNU")
```
lum1n0us: Is it intentional to keep a copy here instead of using the content from config_common.cmake like others?

product-mini/platforms/linux/CMakeLists.txt
```diff
@@ -140,9 +140,8 @@ check_pie_supported()
...
     set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -mindirect-branch-register")
-    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mindirect-branch-register")
     # UNDEFINED BEHAVIOR, refer to https://en.cppreference.com/w/cpp/language/ub
```
lum1n0us: Is it intentional to remove `set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mindirect-branch-register")`?

sbwtw:
Hi @lum1n0us , Thanks for your review. and I resolved all issues. please have a look

wamr-compiler/CMakeLists.txt
```diff
@@ -328,7 +328,7 @@ endif ()
 #TODO: sync up WAMR_BUILD_SANITIZER in config_common.cmake
 # set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wconversion -Wsign-conversion")
 if (WAMR_BUILD_TARGET MATCHES "X86_.*" OR WAMR_BUILD_TARGET STREQUAL "AMD_64")
-  if (NOT (CMAKE_C_COMPILER MATCHES ".*clang.*" OR CMAKE_C_COMPILER_ID MATCHES ".*Clang" OR MSVC))
+  if (CMAKE_C_COMPILER MATCHES ".*gcc.*" OR CMAKE_C_COMPILER_ID MATCHES ".*GNU")
```
lum1n0us:
```suggestion
  if (CMAKE_C_COMPILER_ID MATCHES ".*GNU")
```

PR #4785, opened 2026-01-14
fix: disable unsigned integer overflow sanitization

lum1n0us:
ref https://github.com/bytecodealliance/wasm-micro-runtime/pull/4549#issuecomment-3218687294


TianlongLiang (approved):

lum1n0us:
Static PGO has been verified with version 18.1.8 and has passed.

PR #4799, opened 2026-01-22
language-bindings/go: upgrade yaml.v3

haitaohuang:
Upgrade github.com/stretchr/testify from v1.7.0 to v1.11.1, and upgrade gopkg.in/yaml.v3 to v3.0.1

This fixes CVE-2022-28948 in yaml v3.0.0


lum1n0us:
Usually, we don't post the patch for a security issue publicly. However, in this specific case, the targeted security advisory is not going to be accepted as a security issue because language bindings are not tier A support features and they are all experimental. Therefore, we will continue with the process.

haitaohuang:
@lum1n0us thanks. Will there be a new release tag downstream can point to?

PR #4800, opened 2026-01-22
Fix potential native stack overflow in struct/array object creation

kylo5aby:
Add native stack overflow detection to `wasm_array_obj_new()`,
`wasm_struct_obj_new()` to prevent potential stack overflow when those GC objects creation performs deep recursion.

### DIFF
--- a/core/iwasm/common/gc/gc_object.c
+++ b/core/iwasm/common/gc/gc_object.c
@@ -103,6 +103,10 @@ wasm_struct_obj_new_internal(void *heap_handle, WASMRttTypeRef rtt_type)
 WASMStructObjectRef
 wasm_struct_obj_new(WASMExecEnv *exec_env, WASMRttTypeRef rtt_type)
 {
+    if (!wasm_runtime_detect_native_stack_overflow(exec_env)) {
+        return NULL;
+    }
+
     void *heap_handle = get_gc_heap_handle(exec_env);
     return wasm_struct_obj_new_internal(heap_handle, rtt_type);
 }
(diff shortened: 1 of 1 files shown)


lum1n0us:
- Native stack overflow checking is a good practice. However, besides that, memory (allocated) is another resource that can be exhausted due to recursion. So, I am considering directly limiting the nesting level to protect against both stack overflow and out-of-memory issues. Please share your comments.
- We need a few regression test cases to cover these scenarios.

kylo5aby:
> * Native stack overflow checking is a good practice. However, besides that, memory (allocated) is another resource that can be exhausted due to recursion. So, I am considering directly limiting the nesting level to protect against both stack overflow and out-of-memory issues. Please share your comments.

for multiple levels of nesting, currently appears to handle this properly(just return `NULL`, to avoid heap overflow),  this behavior seems inherently equivalent to creating numerous `struct/array` objects. If I'm wrong, please let me know.

lum1n0us:
Yes and no. Theoretically, whether it's the stack or heap, if they happen to be exhausted at the time of these two `wasm_xxx_obj_new()` calls, they can gracefully return an error and exit. 

However, in practice, there are many functions and allocations along the chain of executing nested structs/arrays. Each point could potentially lead to resource exhaustion and an exception. Unless intercepted at the very beginning, it's challenging to catch all such instances during execution.

Still, don't forget to add a few regression test cases to cover these scenarios.

TianlongLiang:
> So, I am considering directly limiting the nesting level to protect against both stack overflow and out-of-memory issues. Please share your comments.

Agreed, I also think directly limiting the nesting level is a clearer standard, it can give more meaningful error info like "reach max recursion level" rather than "stack/heap overflow"

lum1n0us:
IIUC, it's not due to too many nested levels of a composite type. It's a problem related to too many recursive levels. It might be better to either look for optimization opportunities in the current recursive native stack overflow protection or consider marking this issue as "won't fix."

PR #4812, opened 2026-01-28
Fix a compilation error by clang-17

lum1n0us:
To fix below errors reported by clang-17
```
variable-sized object may not be initialized
```

clang-17 is the default version on MacOS Tahoe(26.2) on AppleM1


TianlongLiang:
I think we'd better not to use heap memory. I am not sure if it will be correctly collected when it's not in the happy path(e.g., when a ASSERT fail)

Issue #4816, opened 2026-01-30
Cannot correctly release a field (anyref) in a struct or an array in destroy_init_expr_data_recursive()


lum1n0us:
Assume there is a struct type declared as 

```
(type $struct_type (struct (field (mut i32)) (field (mut ref $struct_type))))
```

Additionally, there is a global defined as

```
(global $g_struct (mut (ref $struct_type)) (struct.new $struct_type (i32.const 10) (struct.new $struct_type (i32.const 20) (struct.new_default $struct_type))))
```

In the second field, `g_struct` is assigned a value created by `struct.new`, whose second field is also a struct created by `struct.new_default`.

In `load_init_expr()`, `g_struct` is represented by `WASMStructNewInitValues`. 

```
typedef struct WASMStructNewInitValues {
    uint32 type_idx;
    uint32 count;
    WASMValue fields[1];   // <- fields
} WASMStructNewInitValues;

typedef union WASMValue {   // <- only data.
    int32 i32;
    uint32 u32;
    uint32 global_index;
    uint32 ref_index;
    int64 i64;
    uint64 u64;
    float32 f32;
    float64 f64;
    V128 v128;
#if WASM_ENABLE_GC != 0
    wasm_obj_t gc_obj;
    uint32 type_index;
    struct {
        uint32 type_index;
        uint32 length;
    } array_new_default;
    /* pointer to a memory space holding more data, current usage:
     *  struct.new init value: WASMStructNewInitValues *
     *  array.new init value: WASMArrayNewInitValues *
     */
    void *data;
#endif
} WASMValue;
```

All field values are represented by `WASMValue`. When a `WASMStructNewInitValues` is assigned to another `WASMStructNewInitValues` as a field, it is packed into a `WASMValue` and, as a result, loses its type.

Therefore, when we try to determine the type of a field in a struct, such as in `destroy_init_expr_data_recursive()`, we have to refer to the type declaration, `$struct_type`. However, at this point, `$struct_type` provides an abstract type, `anyref`, and cannot offer more precise information. As a result, we cannot release the second field as a struct because we lack the necessary knowledge.

PR #4176, opened 2025-04-01
fix nested struct init value error in gc aot

xujuntwt95329:

PR #4817, opened 2026-01-30
Update the GC tiered level from B to C.

lum1n0us:
Several known limitations have been exposed, and a few might impact the estimation when trying to integrate that feature into a product.

- https://github.com/bytecodealliance/wasm-micro-runtime/issues/4542
- https://github.com/bytecodealliance/wasm-micro-runtime/issues/4816
- https://github.com/bytecodealliance/wasm-micro-runtime/pull/4176


TianlongLiang:
@no1wudi do you have any thoughts on this? @lum1n0us and I don't have too much bandwidth working on syncing the implementation with the newest GC proposal. I wonder whether you will be interested/needed the newest GC proposal. If you are not interested either, I think it's safe for us to update the GC proposal to tier C(won't affected the fact the maintenance of existing parts) as this PR does.

PR #4818, opened 2026-01-30
ci: allow .clang-tidy and fix exception type in guidelines check

kylo5aby:


ci/coding_guidelines_check.py
```diff
@@ -183,12 +183,14 @@ def check_file_name(path: Path) -> bool:
...
     - package-lock
     - vite-env.d
+    - .clang-tidy (standard config file for Clang-Tidy)
     """
     if path.stem in [
         "docker-compose",
         "package-lock",
         "vite-env.d",
         "osv-scanner",
+        ".clang-tidy",
```
lum1n0us, line 193: **/.clang-tidy. in case if a sub-dir has its own configuration

kylo5aby: > if path.stem in [
    "docker-compose",

here uses `path.stem` to match the filename itself. So currently it can handle the case `**/.clang-tidy`

PR #4827, opened 2026-02-05
validate heap_type in load_import_table_list

kylo5aby:
add validation to ensure `heap_type` to be either a valid type index (0 <= heap_type < type_count) or a valid common heap type via `wasm_is_valid_heap_type()`


lum1n0us:
We usually assume that WASM files can be malformed, but AOT files are not. This is mostly because, in WAMR's security model, the AOT compiler is in the trusted zone. Therefore, I can hold off on this patch.

PR #4828, opened 2026-02-05
Refactoring enclave sample

lum1n0us:


ci/install_sgx_for_applicaiton_developer.sh
```diff
@@ -0,0 +1,400 @@
+#!/bin/bash
```
TianlongLiang, line 1: Will we add more CI for sgx in the future? Otherwise, maybe we should put it in the build-scripts dir

TianlongLiang (approved):

ci/install_sgx_for_applicaiton_developer.sh
lum1n0us: Update SGX related documents in another PR.

PR #4835, opened 2026-02-12
Fix unchecked literal_size in AOT load_text_section

sumleo:
`load_text_section()` reads `literal_size` from the AOT section body but does not validate it against the remaining buffer before using it to compute `module->code` and `module->code_size`.

When `literal_size` exceeds `(buf_end - buf)`, `module->code` points past the section body and `code_size` underflows, leading to out-of-bounds access during module usage or crash during `aot_unload()`.

This patch adds a bounds check for `literal_size` immediately after reading it.


TianlongLiang:
> When `literal_size` exceeds `(buf_end - buf)`

I don't think this can be true, the WAMR assumption is that you are using wamrc to generate aot file, not whatever binary you fuzzed, so we don't do check in load time(like this `literal_size` in your case) since it's already done when aot file generation.

sumleo:
Got it.

PR #4836, opened 2026-02-12
Fix off-by-one in AOT func_index bounds checks

sumleo:
The AOT relocation loader validates `func_index` parsed from symbols like `aot_func#N` using:

```c
(func_index = (uint32)atoi(p)) > module->func_count
```

Since `func_ptrs` is allocated with exactly `func_count` elements (indices 0 to `func_count-1`), `func_index == func_count` is out of bounds but passes the `>` check.

This patch changes `>` to `>=` in all 4 affected locations in `do_text_relocation()` and `do_data_relocation()`.


TianlongLiang:
The modification is valid, but you may want to modify the unit test case, current it's not make any sense, I suggest modify to actually cover your modification, or simply remove it would be fine too.

sumleo:
Removed.

PR #4837, opened 2026-02-12
Fix missing IS_INVALID_TAGINDEX check in RETHROW handler

sumleo:
The RETHROW opcode handler reads `exception_tag_index` from the stack and directly accesses `module->module->tags[exception_tag_index]` without checking for `INVALID_TAGINDEX`.

When `CATCH_ALL` catches a cross-module exception with an unknown tag, it pushes `INVALID_TAGINDEX` (0xFFFFFFFF) onto the stack. If RETHROW then executes, it accesses `tags[0xFFFFFFFF]` — a massive out-of-bounds read.

The THROW handler correctly checks `IS_INVALID_TAGINDEX` before the array access. This patch adds the same check to the RETHROW handler: when the tag index is invalid, skip the `tags[]` access and set `cell_num_to_copy` to 0, allowing the exception to propagate to `CATCH_ALL` handlers.

### DIFF
--- a/tests/unit/exception-handling/exception_handling_test.cc
+++ b/tests/unit/exception-handling/exception_handling_test.cc
@@ -82,8 +82,8 @@ TEST_F(ExceptionHandlingTest, load_module_with_exception_handling)
         /* Export: "f" = func 0 */
         0x07, 0x05, 0x01, 0x01, 0x66, 0x00, 0x00,
         /* Code section */
-        0x0A, 0x09, 0x01,
-        0x07, 0x00,       /* body size=7, 0 locals */
+        0x0A, 0x08, 0x01,
+        0x06, 0x00,       /* body size=6, 0 locals */
         0x06, 0x40,       /* try (void) */
         0x19,             /* catch_all */
         0x0B,             /* end try */


tests/unit/wasm-interp/wasm_interp_test.cc
```diff
@@ -0,0 +1,168 @@
...
+    if (!module) {
+        /* If the module fails to load, it's likely because the tag section
+         * encoding is not matching the expected format. Skip in that case. */
+        GTEST_SKIP() << "Module load failed: " << error_buf;
+    }
+
+    wasm_module_inst_t inst = wasm_runtime_instantiate(
+        module, 8192, 8192, error_buf, sizeof(error_buf));
+
+    if (inst) {
```
lum1n0us: should have an assertion about inst.

`if(inst)` should be removed since `else` branch are not expected.

tests/unit/wasm-interp/wasm_interp_test.cc
```diff
@@ -0,0 +1,168 @@
...
+        /* If instantiation succeeds, verify function lookup works */
+        wasm_function_inst_t func =
+            wasm_runtime_lookup_function(inst, "f");
+        EXPECT_NE(func, nullptr) << "Function 'f' should be found";
+
```
lum1n0us: should be an assertion about `func`.

`if(func)` should be removed.

tests/unit/wasm-interp/wasm_interp_test.cc
```diff
@@ -0,0 +1,168 @@
...
+        if (func) {
+            /* Execute the function - should not crash */
+            bool ok = wasm_runtime_call_wasm(
+                wasm_runtime_create_exec_env(inst, 8192), func, 0, NULL);
+            /* The function may or may not succeed depending on the
+             * exact exception handling implementation details */
+            (void)ok;
```
lum1n0us: better have a return value for compare.

and use `wasm_runtime_get_exception` to check if necessary.

tests/unit/wasm-interp/wasm_interp_test.cc
sumleo: Done. Replaced `if(inst)` with `ASSERT_NE(inst, nullptr)` so the test fails immediately if instantiation doesn't succeed.

tests/unit/wasm-interp/wasm_interp_test.cc
sumleo: Done. Removed the `if(func)` conditional and upgraded `EXPECT_NE` to `ASSERT_NE` so the test aborts if the function isn't found.

tests/unit/wasm-interp/wasm_interp_test.cc
sumleo: Done. Now checking the return value of `wasm_runtime_call_wasm` with `ASSERT_TRUE(ok)` and including `wasm_runtime_get_exception(inst)` in the failure message. Also properly creating and destroying `exec_env`.

PR #4846, opened 2026-02-25
Fix relocation addend sign extension on 32-bit platforms

sumleo:
## Summary

On 32-bit platforms, `load_relocation_section` reads the relocation addend as `uint32` and zero-extends it to `uint64`. This corrupts negative addends — for example, an addend of -4 (`0xFFFFFFFC` as a 32-bit value) becomes `4294967292` instead of remaining `-4` (`0xFFFFFFFFFFFFFFFC` as a 64-bit value).

The Windows code path (around line 3616) already handles this correctly with proper sign extension.

## Changes

- Added a `(int64)(int32)` double cast when widening `addend32` to 64 bits, so negative values are sign-extended rather than zero-extended
- `addend32` remains `uint32` as expected by `read_uint32()`
- The `offset32` variable remains `uint32` since relocation offsets are unsigned

## Test plan

- Verified that the fix matches the existing Windows code path behavior
- The change only affects 32-bit platforms (`sizeof(void *) != 8`)
- Negative relocation addends (e.g., used for GOT-relative relocations) will now be preserved correctly


core/iwasm/aot/aot_loader.c
```diff
@@ -3868,11 +3868,12 @@ load_relocation_section(const uint8 *buf, const uint8 *buf_end,
                 read_uint64(buf, buf_end, relocation->relocation_addend);
             }
             else {
-                uint32 offset32, addend32;
+                uint32 offset32;
+                int32 addend32;
```
lum1n0us: L3873 `read_uint32()` expects `addend32` as an unsigned. No need to change its type to `int32`.

core/iwasm/aot/aot_loader.c
```diff
@@ -3868,11 +3868,12 @@ load_relocation_section(const uint8 *buf, const uint8 *buf_end,
...
                 read_uint32(buf, buf_end, offset32);
                 relocation->relocation_offset = (uint64)offset32;
                 read_uint32(buf, buf_end, addend32);
-                relocation->relocation_addend = (uint64)addend32;
+                relocation->relocation_addend = (int64)addend32;
```
lum1n0us: Yes. there is a risk when casting `uint64` to `int64`. But personally, I am curious that is it reported by a compiler or analysis tool?

core/iwasm/aot/aot_loader.c
sumleo: Done. Reverted `addend32` back to `uint32` and now sign-extend via a double cast `(int64)(int32)addend32` instead. This keeps `read_uint32()` happy while still preserving proper sign extension for negative addends on 32-bit platforms.

core/iwasm/aot/aot_loader.c
sumleo: This was found during code review. On 32-bit platforms, the relocation addend is read as a `uint32` and then cast to `uint64`. If the addend is negative (e.g., -4 stored as 0xFFFFFFFC), the unsigned widening produces 0x00000000FFFFFFFC instead of the correct 0xFFFFFFFFFFFFFFFC. The fix now uses `(int64)(int32)addend32` to properly sign-extend while keeping `addend32` as `uint32` for `read_uint32()`.

PR #4848, opened 2026-02-25
Fix wrong condition in Memory64 data segment loading

sumleo:
## Summary

In `load_data_segment_section`, the Memory64 code path manually inspects
import/defined memory arrays to determine the memory64 flag. The condition
`module->import_memory_count > 0` is incorrect when a module has both imported
and defined memories — a data segment targeting a defined memory
(`mem_index >= import_memory_count`) would still enter the import branch,
causing an out-of-bounds read.

## Fix

Replace the manual flag lookup with the existing `has_module_memory64()` helper,
which already encapsulates this logic correctly. Apply the same refactor to both
`wasm_loader.c` and `wasm_mini_loader.c`.

Before:
```c
uint8 memory_flag;
if (module->import_memory_count > 0) {
    memory_flag = module->import_memories[mem_index].u.memory.mem_type.flags;
}
else {
    memory_flag = module->memories[mem_index - module->import_memory_count].flags;
}
mem_offset_type = memory_flag & MEMORY64_FLAG ? VALUE_TYPE_I64 : VALUE_TYPE_I32;
```

After:
```c
mem_offset_type = has_module_memory64(module) ? VALUE_TYPE_I64 : VALUE_TYPE_I32;
```

## Impact

Fixes an out-of-bounds array access when a module with both imported and defined
memories has a data segment targeting a defined memory with Memory64 enabled.


core/iwasm/interpreter/wasm_loader.c
```diff
@@ -5159,7 +5159,7 @@ load_data_segment_section(const uint8 *buf, const uint8 *buf_end,
 #if WASM_ENABLE_MEMORY64 != 0
                 /* This memory_flag is from memory instead of data segment */
                 uint8 memory_flag;
-                if (module->import_memory_count > 0) {
+                if (mem_index < module->import_memory_count) {
```
TianlongLiang: better refactor this code snippet with `has_module_memory64` instead, and modify the `wasm_mini_loder.c` too

sumleo: Done. Replaced the manual import/defined memory flag lookup with `has_module_memory64(module)` in both `wasm_loader.c` and `wasm_mini_loader.c`.

lum1n0us (changes requested):
`has_module_memory64()` always refers to `memory[0]`. But in `load_data_segment_section()`, it should be `memory[mem_index]`.

it should be something like:
``` c
if (mem_index < module->import_memory_count) {
  memory_flag = module->import_memories[mem_index].u.memory.mem_type.flags;
}
else {
  memory_flag = module->memories[mem_index - module->import_memory_count].flags;
}

```

PR #4867, opened 2026-03-07
doc(security,aot): added security guidance around AoT binaries

srberard:
Adds explicit guidance distinguishing trusted AoT binaries from untrusted Wasm binaries, clarifies that malformed/manipulated AoT files are bugs not security issues, and cleans up wording in the security checklist section.

### DIFF
--- a/doc/security_need_to_know.md
+++ b/doc/security_need_to_know.md
@@ -21,18 +21,16 @@ If the AoT compiler and/or related tools emit an AoT binary that breaches the Wa
 ### Is this bug considered a security vulnerability?
 
 #### For someone who finds a problem
+If a bug **causes a crash or hang**, treat it as a possible security issue and report it to a security advisor. A maintainer will review it and change its category if needed. When in doubt, report it as a security issue.
 
-if a bug **results in crash or hang**, please treat it as a security problem and report it to a security advisor. The maintainer will look into it and change its category if needed. It is better safe than sorry.
+If the person reporting the issue can answer "Yes" to any question in the checklist below, report it as a security issue. Otherwise, the issue can be treated as a regular bug.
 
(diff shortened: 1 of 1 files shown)


doc/security_need_to_know.md
```diff
@@ -15,13 +14,25 @@ It is commonly stated that a security issue is an issue that:
...
+If the AoT compiler and/or related tools emit an AoT binary that causes a breach of the Wasm sandbox or a crash is considered a potential security issue.  It is assumed that the correct configuration and options are used when generating AoT binaries.  Misconfiguration or misuse of the tooling options, therefore, are not considered to be security issues.
+
 ### Is this bug considered a security vulnerability?
 
 #### For someone who finds a problem
 
 if a bug **results in crash or hang**, please treat it as a security problem and report it to a security advisor. The maintainer will look into it and change its category if needed. It is better safe than sorry.
 
-If the author of an issue(results in crash or hang) can go through the following checklist and answer all questions with "No", it is fine to mark it as a regular bug. If not, please report it as a security issue.
+If the author of an issue(results in crash or hang) can go through the checklist below and answer all questions with "No", it is fine to mark it as a regular bug. If not, please report it as a security issue.
```
lum1n0us: The original idea here is to use an easily recognizable characteristic to let the author finish a quick initial triage. Maybe a better description would be:
- If an issue is a hang or crash, report it as a security issue. 
- If an issue doesn't end up with a hang or crash, but allows any of the following questions to be answered as YES, report it as a security issue.

But is this too strict?

doc/security_need_to_know.md
```diff
@@ -15,13 +14,25 @@ It is commonly stated that a security issue is an issue that:
...
+
+Does the
```
lum1n0us: maybe we can provide a list with WASM/WAMR-specific questions, such as escaping the sandbox, exposing unauthorized WASI APIs, and so on.

srberard: Updated.  I simplified and updated the question list.

PR #4875, opened 2026-03-17
Add SYSTEM keyword to 3rd party libraries in CMake

TianlongLiang:
1. Enable the system include flag for third-party libraries to suppress corresponding compile warnings
2. Also, fix a wamrc compile warning


TianlongLiang:
Maybe after this, we can enable -Werror or at least one step closer to enabling it

TianlongLiang:
One implication for the use `SYSTEM` Keyword:
- It changes the Search Priority, compilers search standard include directories (-I) before system include directories (-isystem). This will result in the WAMR project's local headers being prioritized over third-party or system headers if there happens to be a naming collision.

It should be fine though

lum1n0us (approved):

PR #4895, opened 2026-03-31
fix: Protect mem consumption from null deref

lum1n0us:
- add debug assertions to wasm_get_module_inst_mem_consumption
- add debug assertions to aot_get_module_inst_mem_consumption
- nullify memories pointer after free in destruction path
- nullify memories pointer after free in AOT destruction
- docs: add precondition documentation for memory consumption functions

### DIFF
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -966,6 +966,7 @@ memories_deinstantiate(AOTModuleInstance *module_inst)
         }
     }
     wasm_runtime_free(module_inst->memories);
+    module_inst->memories = NULL;
 }
 
 static AOTMemoryInstance *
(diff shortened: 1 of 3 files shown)


core/iwasm/interpreter/wasm_runtime.c
```diff
@@ -3445,9 +3445,11 @@ wasm_deinstantiate(WASMModuleInstance *module_inst, bool is_sub_inst)
         (WASMModuleInstanceCommon *)module_inst);
 #endif
 
-    if (module_inst->memory_count > 0)
+    if (module_inst->memory_count > 0) {
         memories_deinstantiate(module_inst, module_inst->memories,
                                module_inst->memory_count);
+        module_inst->memories = NULL;
```
TianlongLiang, line 3451: I don't think it's worth setting NULL here? Given the fact that this module inst is about to be destroyed anyway? It will create an inconsistency with other data members, like L3455

core/iwasm/aot/aot_runtime.c
```diff
@@ -966,6 +966,7 @@ memories_deinstantiate(AOTModuleInstance *module_inst)
         }
     }
     wasm_runtime_free(module_inst->memories);
+    module_inst->memories = NULL;
```
TianlongLiang, line 969: I don't think it's worth setting NULL here? Given the fact that the module instance will be destroyed later anyway?

PR #4909, opened 2026-04-08
Fixed build configs and invoke for windows arm64

joevdh:
- Updated cmake build process to detect windows arm configuration, and use the marmasm assembler.
- Fixed the windows arm64 assembly to load the parameters correctly when invoking functions
- Note: AI assisted


core/iwasm/common/iwasm_common.cmake
```diff
@@ -4,12 +4,18 @@
 set (IWASM_COMMON_DIR ${CMAKE_CURRENT_LIST_DIR})
 
 include_directories (${IWASM_COMMON_DIR})
-if (MSVC AND WAMR_BUILD_PLATFORM STREQUAL "windows" AND WAMR_BUILD_TARGET MATCHES "AARCH64.*")
-  if (DEFINED ENV{VCToolsInstallDir})
+if (MSVC AND WAMR_BUILD_PLATFORM STREQUAL "windows" AND WAMR_BUILD_TARGET MATCHES "AARCH64.*" AND NOT WAMR_BUILD_WAMR_COMPILER)
```
lum1n0us, line 7: What specific build issue does skipping `WAMR_BUILD_WAMR_COMPILER` solve? The wamr-compiler also needs to compile for Windows ARM64, so it's not clear why this        
  scenario should be excluded from the ARM64 assembly settings. Could you provide more context on this?

core/iwasm/common/iwasm_common.cmake
```diff
@@ -4,12 +4,18 @@
...
+  if (CMAKE_ASM_MARMASM_COMPILER)
```
lum1n0us, line 8: Concern: `CMAKE_ASM_MARMASM_COMPILER` appears to be a newer CMake feature. Could you confirm:                                                                                    
  - Which CMake version introduced this variable?                                                                                                                                
  - The project currently requires CMake 3.14 minimum - is this compatible?

lum1n0us (changes requested):
The PR fixes invokeNative_armasm64.asm but does not modify invokeNative_armasm64_simd.asm.                                                                                     
                                                                                                                                                                                 
  Looking at iwasm_common.cmake, the SIMD version follows the same Windows ARM64 path:
```                                                                                           
  else()          
      # SIMD version                                                                                                                                                             
      if (WAMR_BUILD_PLATFORM STREQUAL "windows")                                                                                                                                
        if (MSVC)                                                                                                                                                                
          set (source_all ${c_source_all} ${IWASM_COMMON_DIR}/arch/invokeNative_armasm64_simd.asm)                                                                               
```
                                                                                                                                                                                 
  Both files likely share the same calling convention issue. Should invokeNative_armasm64_simd.asm be updated with the same register loading fixes (removing d4-d7, correcting   
  parameter indices)?

PR #4912, opened 2026-04-13
perf(wamrc): Replace slow symbol lookup with hash map in wamrc

jammar1:
Replace slow symbol lookup with hash map in AOT compilation.

The relocation symbol deduplication in aot_emit_aot_file.c used a linked list, requiring O(n) traversal per relocation to check for duplicates.

Replace with a lazily-allocated hash map for O(1) lookups. The linked list is preserved for ordered iteration during emission.

When testing against a fairly large Wasm binary (about 18MB) this reduced compilation time by about ~15%.

### DIFF
--- a/core/iwasm/compilation/aot_emit_aot_file.c
+++ b/core/iwasm/compilation/aot_emit_aot_file.c
@@ -45,6 +45,7 @@ typedef struct AOTSymbolList {
     AOTSymbolNode *head;
     AOTSymbolNode *end;
     uint32 len;
+    void *hash_table;
 } AOTSymbolList;
 
 /* AOT object data */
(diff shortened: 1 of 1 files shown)


lum1n0us (changes requested):
Thanks for the optimization! The performance improvement is meaningful.

However, I have a question about the design: currently, when a new symbol is not found, the code creates both an `AOTSymbolNode` (for the linked list) and an `AOTSymbolHashEntry` (for the hash table). This means **two separate allocations storing the same symbol and index**.

Is there a specific reason to keep the linked list if the hash table already provides O(1) lookup? The linked list seems to only be used for ordered iteration during destruction, which could also be done by iterating the hash table buckets?

jammar1:
We do currently use the list for symbol emission too, in `aot_emit_relocation_symbol_table`. The hash entry is just storing a pointer to the string so it doesn't use much extra memory. We could completely swap to just using a hash map if we also added str_len to it as long as the iteration order is consistent between the passes. Happy to do either.

core/iwasm/compilation/aot_emit_aot_file.c
```diff
@@ -989,46 +990,81 @@ get_relocation_groups_size(AOTObjectData *obj_data,
...
+            }
+            sym = sym->next;
+            idx++;
+        }
+    }
+    index = symbol_list->len;
+    {
+        AOTSymbolNode *sym = wasm_runtime_malloc(sizeof(AOTSymbolNode));
+        if (!sym)
+            return (uint32)-1;
```
lum1n0us, line 1045: `symbol_list->len` is 0 and `hash_table` points to a allocation and `aot_obj_data_destroy()` only releases `has_table` if `symbol_list.len != 0`. There will be a leak.

lum1n0us (changes requested):
My concern is that this approach maintains two data structures at the same time, which feels a bit heavy for WAMR’s lightweight design goals. I think we should justify that tradeoff more clearly, or consider whether a simpler solution could achieve a similar benefit.

PR #4936, opened 2026-05-06
Add custom data setter for InstantiationArgs2

mechakotik:
Add custom data setter for InsantiationArgs2, so user can provide custom data before post-instantiation code (such as `_initialize`) runs.


lum1n0us (approved):

mechakotik:
Just pinging in case this PR got lost. Anything I should address to move it forward?

PR #4944, opened 2026-05-14
Utilize C99 flexible array members to allow for FORTIFY/ASAN

satur9nine:
When building the code today with:

    -fsanitize=bounds,bounds-strict,vla-bound 
    -fsanitize-trap=bounds,bounds-strict,vla-bound 

WAMR crashes in various places when trying to use arrays of length one as if they are flexible array members. This change enables use of real C99+ flexible array members when the compiler supports it.

### DIFF
--- a/.github/workflows/compilation_on_android_ubuntu.yml
+++ b/.github/workflows/compilation_on_android_ubuntu.yml
@@ -369,6 +369,48 @@ jobs:
           ctest --output-on-failure
         working-directory: tests/unit
 
+  build_array_bounds_warning_check:
+    needs: [build_llvm_libraries_on_ubuntu_2204]
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-22.04]
+        include:
+          - os: ubuntu-22.04
+            llvm_cache_key: ${{ needs.build_llvm_libraries_on_ubuntu_2204.outputs.cache_key }}
+    steps:
+      - name: checkout
+        uses: actions/checkout@v6.0.2
+        with:
+          submodules: recursive
+
+      - name: Get LLVM libraries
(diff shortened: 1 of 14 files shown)


lum1n0us (changes requested):
Please ignore the SGX CI failure for now and focus on investigating the Ubuntu-side issue. Once that is fixed, this should be ready to merge.


---

I agree that the issue addressed here is real: using trailing [1] / [0] arrays as flexible array members is not the C99-standard form, and stricter bounds/FORTIFY-style checking can legitimately diagnose or trap on such patterns.

My guess is that some of these non-standard forms were originally kept for compiler compatibility, especially when the same headers may be consumed by C++ compilers. That said, in the codebase there also appear to be two different cases mixed together: some are true variable-length tail storage and should be normalized toward standard FAM usage where possible, while others are effectively just placeholders for layout/type-compatibility purposes and would be better expressed with a clearer non-array representation rather than continuing to look like tail arrays.

With that in mind, this PR looks more like a partial fix than a complete cleanup. It resolves some real problematic cases, but there are still a non-trivial number of similar definitions remaining under core/iwasm and core/shared that have not been corrected yet. I think it would help to frame this change explicitly as an incremental step, and track the remaining cases separately so we do not give the impression that the broader issue has been fully addressed.

satur9nine:
OK found one more small thing to fix and minor CI improvement; now CI is green.

satur9nine:
You said:

> I think it would help to frame this change explicitly as an incremental step...

Let me know if there's something specifically you want me to do in that regard in this PR.

satur9nine (commented):
Ready for review

satur9nine:
@lum1n0us Is there anything else you would like me to do in this PR?

PR #4949, opened 2026-05-21
fast-interp: legacy exception handling (try/catch/rethrow/delegate/tag)

matthargett:
Lifts the cmake guard at `build-scripts/unsupported_combination.cmake:67` that forbids
`WAMR_BUILD_EXCE_HANDLING=1 WAMR_BUILD_FAST_INTERP=1` and adds the matching dispatch
loop coverage in `wasm_interp_fast.c` — loader-side EH metadata table, runtime
EH-frame stack, catch-walk for `throw` / `rethrow`, `delegate` forward-to-outer,
tag-with-params payload routing, and result-typed `try`-region COPY-at-CATCH
alignment. I tried to keep each step bisectable.

Why we built this: we're replacing WasmEdge with WAMR fast-interp as the wasm
runtime in a pure-interpreter App-Store-eligible app, and a migration
blocker is `graphql-validation` compiled by [Porffor](https://github.com/CanadaHonk/porffor) — JS-to-wasm output that lowers
`try`/`catch`/`throw` to the wasm-exceptions section. Without EH enabled, fast-interp
rejects the binary at load with `invalid section id`; with EXCE_HANDLING +
CLASSIC_INTERP it loads but fast-interp is 1.3–1.8× faster on every benchmark we
ran.

Cross-microarch benchmarks: M4 MacBook / A14  iPhone 12 / A12 iPhone XS /
S8 Watch SE2 at
https://github.com/rebeckerspecialties/wasm-benchmark/blob/claude/relaxed-simd-diff-fuzz/README.md#cross-runtime-results-across-apple-silicon-e-cores
. Integration tests in our benchmark repo include a Porffor-compiled
`graphql-validation` workload that mirrors the real-world `try { visit(…) } catch (e)
{ if (e !== abortObj) throw e; }` shape and exercises every EH opcode the loader
emits. ASan + UBSan builds are part of the local dev loop.

Companion PR: relaxed-SIMD fast-interp opcode lowering, posted separately
(`f32x4.relaxed_madd` etc).

Validated existing benchmarks perform nearly exactly the same in terms of wallclock, throughput, cache, and branch predictor using CPU bottlneck template in xctrace.

### DIFF
--- a/build-scripts/unsupported_combination.cmake
+++ b/build-scripts/unsupported_combination.cmake
@@ -64,7 +64,17 @@ endfunction()
 
 if(WAMR_BUILD_EXCE_HANDLING EQUAL 1)
   check_aot_mode_error("Unsupported build configuration: EXCE_HANDLING + AOT")
-  check_fast_interp_error("Unsupported build configuration: EXCE_HANDLING + FAST_INTERP")
+  # FAST_INTERP + EXCE_HANDLING is supported for *throw-only* shapes:
+  # WASM modules that declare tags and execute throw / rethrow without
+  # ever entering a same-function try / catch handler. The throw
+  # propagates to the caller via the existing got_exception bailout
+  # path, exactly like any other trap. This covers Porffor (its
+  # JS-to-wasm compiler emits 0 try/catch handlers; every JS throw
+  # escapes to the host). Modules that contain WASM_OP_TRY / CATCH /
+  # CATCH_ALL / DELEGATE still load, but those handlers report
+  # "unsupported opcode" at runtime — see the WASM_OP_TRY handler in
+  # core/iwasm/interpreter/wasm_interp_fast.c. Full same-function
+  # try / catch lowering is the natural follow-up.
   check_fast_jit_error("Unsupported build configuration: EXCE_HANDLING + FAST_JIT")
   check_llvm_jit_error("Unsupported build configuration: EXCE_HANDLING + JIT")
 endif()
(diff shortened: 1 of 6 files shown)


matthargett:
Update: pushed [`0411662d`](https://github.com/rebeckerspecialties/wasm-micro-runtime/commit/0411662d4c2f52625a356ff0583c9d2d970cb00f) — MSVC `__builtin_expect` no-op shim. The Windows MSVC matrix that was failing with `LNK2019: unresolved external symbol __builtin_expect` is now green.

Of the remaining single CI failure (`build_regression_tests (ubuntu-22.04)`), the four failing tests (BA issues 2702, 2833, 270801, 270802) all execute under `mode: aot` / `runtime: iwasm-default` and crash with `exit code -4` (SIGILL) on the AOT binary. This PR doesn't touch AOT codegen or AOT runtime — only fast-interp. The same SIGILL-on-AOT pattern is visible on `main`'s recent `nightly_run` (`test (ubuntu-22.04, asan, aot, $WASI_TEST_OPTIONS)` + tsan variant — both failure, both running AOT-compiled wasm). My read is this is an upstream-wide CI-infrastructure issue introduced around the LLVM 22 bump (PR #4937) — happy to be told otherwise if I'm misreading. Either way, nothing this PR can fix.

matthargett:
This PR was meant to make a Porffor integration test that worked in wasmtime (using Porffor was itself suggested by wasmtime maintainers) run under WAMR so I could do side-by-side benchmarking on mobile devices. A couple of things changed: Porffor-generated Wasm using legacy EH—has also disappeared from current Porffor. Second, this was the legacy EH that isn't necessary since several other parts of wasm-emitting toolchains are using the new standard.

relaxed-simd is more important and still relevant, I'll revisit latest EH spec implementation alongside other features I need for my app runtime use case once that's merged.

PR #4951, opened 2026-05-22
Update LLVM from 18 to 22

stertl:
### Changes

Bumps the LLVM version from LLVM 18 (`llvmorg-18.1.8`) to LLVM 22 (`llvmorg-22.1.6`). The ARC architecture target branch is updated accordingly from `release/18.x` to `release/22.x`.

### Motivation

LLVM 22 delivers a significant AoT compilation speedup compared to LLVM 18. Compiling an example program ([example.c.zip](https://github.com/user-attachments/files/28149840/example.c.zip) — a large C file with many global variables) to AOT shows:

| `wamrc` version | Wall time |
|---|---|
| wamrc 2.4.4 (LLVM 18) | 66.79 s |
| wamrc with LLVM 22 | 12.59 s |

**~5× speedup** — the updated compiler takes roughly one fifth of the time of the previous version for this workload.

### Testing

```sh
# Compile C to Wasm
$ time clang-21 -target wasm32 -O3 -nostdlib example.c -Wl,--no-entry,--export=function -o example.wasm

# Compile Wasm to AoT with wamrc-2.4.4 with LLVM 18
$ time wamrc-2.4.4 -o example.aot example.wasm                                                                    
Create AoT compiler with:
  target:        x86_64
  target cpu:    skylake
  target triple: x86_64-unknown-linux-gnu
  cpu features:  
  opt level:     3
  size level:    3
  output format: AoT file
Compile success, file example.aot was generated.
wamrc -o example.aot example.wasm  66.79s user 0.04s system 99% cpu 1:06.83 total

# Compile Wasm to AoT with wamrc with LLVM22
$ time ./wamr-compiler/build/wamrc -o example.aot example.wasm
Create AoT compiler with:
  target:        x86_64
  target cpu:    skylake
  target triple: x86_64-unknown-linux-gnu
  cpu features:  
  opt level:     3
  size level:    3
  output format: AoT file
Compile success, file example.aot was generated.
/home/stefan/git/siemens-wasm-micro-runtime/wamr-compiler/build/wamrc -o    12.59s user 0.13s system 99% cpu 12.719 total
```

### DIFF
--- a/build-scripts/build_llvm.py
+++ b/build-scripts/build_llvm.py
@@ -299,7 +299,7 @@ def main():
         "arc": {
             "repo": "https://github.com/llvm/llvm-project.git",
             "repo_ssh": "git@github.com:llvm/llvm-project.git",
-            "branch": "release/18.x",
+            "branch": "release/22.x",
         },
         "xtensa": {
             "repo": "https://github.com/espressif/llvm-project.git",
@@ -309,7 +309,7 @@ def main():
         "default": {
             "repo": "https://github.com/llvm/llvm-project.git",
             "repo_ssh": "git@github.com:llvm/llvm-project.git",
-            "branch": "llvmorg-18.1.8",
+            "branch": "llvmorg-22.1.7",
         },
     }
 
--- a/core/iwasm/aot/aot_runtime.c
+++ b/core/iwasm/aot/aot_runtime.c
@@ -5069,8 +5069,13 @@ aot_dump_pgo_prof_data_to_buf(AOTModuleInstance *module_inst, char *buf,
     }
 
     prof_header.magic = 0xFF6C70726F667281LL;
+#if LLVM_VERSION_MAJOR >= 19
+    /* Version 10 */
+    prof_header.version = 0x000000000000000ALL;
+#else
     /* Version 9 */
     prof_header.version = 0x0000000000000009LL;
+#endif
     /* with VARIANT_MASK_IR_PROF (IR Instrumentation) */
     prof_header.version |= 0x1ULL << 56;
     /* with VARIANT_MASK_MEMPROF (Memory Profile) */
@@ -5082,6 +5087,11 @@ aot_dump_pgo_prof_data_to_buf(AOTModuleInstance *module_inst, char *buf,
     /* __llvm_prf_bits won't be used in PGO, set dummy value here */
     prof_header.num_prof_bitmaps = 0;
     prof_header.bitmap_delta = 0;
+#if LLVM_VERSION_MAJOR >= 19
+    /* vtable value profiling isn't used in PGO, set dummy values here */
+    prof_header.num_vtables = 0;
+    prof_header.vnames_size = 0;
+#endif
 
     if (!is_little_endian()) {
         aot_exchange_uint64((uint8 *)&prof_header.magic);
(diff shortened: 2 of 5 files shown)


lum1n0us:
@TianlongLiang  Please feel free to jump in 

- I re-ran all the failed CI jobs. There are still some compilation errors and AOT generation failures. Please look into these issues.
- During the CI build, the LLVM library is generating extra cache data. This might lead to storage pressure. I’ll fix this as soon as possible.
- Upgrading LLVM is a foundational change. In addition to making sure CI passes, please run local tests, including unit tests, regression tests, spec tests, and sample tests. Also, please do a manual check of the PGO feature.

woodsmc:
>Upgrading LLVM is a foundational change. In addition to making sure CI passes, please run local tests, including unit tests, regression tests, spec tests, and sample tests. Also, please do a manual check of the PGO feature.

We don't have access to all of the platforms, so the local tests are going to be limited to the platforms we've got access to - I think we can do this for Linux x86-64 / ARM64 ... 

Can someone else help with the other platforms? - Is there special github action we can trigger?

stertl:
- I ran all the tests from the `tests` folder (except `tests/fuzz`) on Linux x86_64, and with the committed changes they are all passing.
- As @lum1n0us suspected, the raw PGO data header structure was updated from version 9 to 10, adding two additional fields. Thank you for the hint.
  I put those changes behind compiler switches, but I am not sure how to properly obtain the `LLVM_VERSION_MAJOR` in `core/iwasm/aot/aot_runtime.h` without including an LLVM header, which is probably not desired. 
  Currently, I have inserted a `#define LLVM_VERSION_MAJOR 22` for testing.
  Any suggestions for a proper solution?
- For the fix of the RISC-V PCREL_HI20/LO12 relocations we would need someone with RISC-V experience to review this. 
  @no1wudi, maybe you have some suggestions on that, or would you like to continue on the original PR #4777, fixing the very same issue?
- I assume the two failing checks in the last execution are failing sporadically, as they did not fail in an earlier run and also sporadically fail on the main branch.

matthargett:
Flagging some adjacent CI-tooling work that overlaps with this LLVM bump, so we can coordinate rather than collide.

**Already merged — partly covers the review feedback here:**
- **#4988** frees ~20 GB of runner disk before the prebuilt-LLVM `actions/cache` restore in the multi-tier-jit `test` job — i.e. the "LLVM library generating extra cache data → storage pressure" flagged above, for the non-container Linux jobs. (It can't help `container:` jobs like `spec_test_on_qemu`: their steps run inside the nuttx image and can't free the host disk — see gaps below.)
- **#4980** un-broke the CodeQL nightly, which had been failing on the same external `clang+llvm-18.1.8.tar.xz` + `libtinfo5` fragility the CI carries for its LLVM.

**Open, and relevant to how this PR gets sequenced:**
- **#4992** — bumps the coding-guidelines formatter to **clang-format 21** on `ubuntu-24.04` + a one-time reformat. Worth landing before/alongside an LLVM bump: clang-format's output shifts across major versions, and 21 matches the LLVM in Xcode 26/27, so contributors (including on this PR) format identically to CI instead of chasing drift. Concretely, clang-format **18.1.3 vs 18.1.8** already format two files differently — so version alignment matters here.
- **#4982** — a diff-scoped clang-tidy-21 gate (a curated `.clang-tidy` update is planned so it stays low-noise).
- **#4981** — a per-PR libFuzzer smoke test.
- **#4983** (issue) — the umbrella + a suggested merge order: move CI runners to `ubuntu-24.04` and stop re-downloading the external LLVM tarball, which is the *systemic* source of the storage pressure.

**Gaps we've intentionally not filed PRs for yet (and why):**
- **Storage fix for `container:` jobs** (`spec_test_on_qemu`, etc.): #4988 covers only non-container jobs; container steps can't free the host disk from inside the image. The real fixes are #4983's native-LLVM (no multi-GB cache to restore) or a larger runner — held back because it's the bigger runner migration *and* it depends on which LLVM CI standardizes on.
- **The CI LLVM version itself**: #4983 originally suggested runner-native LLVM 18 for the CodeQL/fuzz builds. This PR's move to **LLVM 22** for AOT changes that calculus — it would be cleaner for the CI analysis builds and `wamrc` to converge on a single LLVM. We deliberately haven't filed the "native-LLVM for CodeQL/fuzz" PR to avoid colliding with this bump; happy to align once the target LLVM is settled here.

FWIW on the sporadic CI failures: I've independently seen the same `build_regression_tests` AOT flake (issue-2702 / 2833 / 270801 / 270802) on a *formatting-only* PR of ours (#4992) where it can't be causal — consistent with "also fails sporadically on main."

PR #4956, opened 2026-05-28
iwasm: ignore SIGPIPE

syborg64:
`SIGPIPE` is triggered when writing to a stream that has no readers. Its default behavior is to terminate the application.

- `SIGPIPE`'s rationale was to allow filter programs to exit even if the output of `write` was never checked. It's not relevant in wasm since we're not writing lazy scripts in C anymore, and libraries typically handle errors graciously.
- `EPIPE` is sufficient to tell users of `write` that the function failed
- [the sockets proposal](https://github.com/WebAssembly/WASI/blob/main/proposals/sockets/Posix-compatibility.md#writing-to-closed-streams-sigpipe-so_nosigpipe-) encourages ignoring this signal
- wasm has no concept of signals, so there is ultimately no 'correct' handling of `SIGPIPE`. It should just not affect wasm modules
- currently there are no workarounds like `SO_NOSIGPIPE` or `MSG_NOSIGNAL`

This PR ignores SIGPIPE ~~outright~~ *in product-mini/iwasm* because that's the cleanest way to do it.

### DIFF
--- a/product-mini/platforms/posix/main.c
+++ b/product-mini/platforms/posix/main.c
@@ -452,7 +452,7 @@ module_destroyer_callback(uint8 *buffer, uint32 size)
 #endif /* WASM_ENABLE_MULTI_MODULE */
 
 static int
-init_ignore_sigpipe()
+init_ignore_sigpipe(void)
 {
     struct sigaction sig_act;
     sig_act.sa_handler = SIG_IGN;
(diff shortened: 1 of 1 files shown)


yamt:
it's better for a library to avoid modifying process-wide settings like this unless absolutely necessary. how about leaving it to the embedder like iwasm command?

syborg64:
would that be added to `product-mini/platforms/posix/main.c` directly then ?
That would be alright by me.

In this case, would a build flag still be needed ? I can't imagine any reason to opt-out if it's *just* iwasm

PR #4973, opened 2026-07-01
addr2line: detect clang version and dispatch to modern or legacy resolver

TianlongLiang:
Refactor the wasm symbolicator to make its toolchain workarounds explicit. addr2line.py now detects the wasi-sdk's clang major version at startup and routes per-frame address resolution to one of:

- resolve_address_modern (clang >= 22, wasi-sdk 33+): single llvm-symbolizer call per address. The symbolizer handles address-to-name correctly on this clang.

- resolve_address_legacy (clang < 22 or unknown): llvm-symbolizer call + llvm-dwarfdump --lookup overlay to fix the outermost function name. Older clang versions emit wasm DWARF that confuses llvm-symbolizer's address-to-name resolver for some addresses (e.g., reports 'recurse' as 'free'). --lookup goes through a different DWARF traversal path that handles them.

The dispatch decision is logged once to stderr so users can see which path was taken without polluting stdout. Stdout is byte- identical between paths for non-buggy inputs.

Also includes accumulated improvements:
- --mode flag {interp,aot,fast-interp} for different runtime offset conventions (interp post-advance, aot at-instruction-start, fast-interp's transformed in-memory bytecode)
- Inline frame annotation "(inlined into <next>)" for clarity
- llvm-symbolizer preferred over llvm-addr2line for column info
- Fallback for offset=0 (trap at function entry; frame_ip not captured)
- Last-resort function-index name fallback when DWARF lacks PC ranges


lum1n0us (approved):
LGTM

PR #4982, opened 2026-07-05
Run clang-tidy 21 on pull-request changed lines

matthargett:
## What

Run the repository's existing `.clang-tidy` on pull requests — but only on the lines the PR changes.

The root `.clang-tidy` is already a maintained, strict config (`bugprone-*`, `cert-*`, `clang-analyzer-*`, ... with `WarningsAsErrors`), but no workflow runs it today, so the null-deref / narrowing-conversion / use-after-move class of bug it catches is currently caught only by a reviewer reading the diff.

## How

Mirrors the diff-scoping already used by the `git-clang-format` gate in `coding_guidelines.yml`:

- `permissions: contents: read`, `fetch-depth: 0`, diff computed from `${{ github.event.pull_request.base.sha }}..HEAD`.
- `clang-tidy-diff.py` turns the changed-line ranges into `-line-filter`, so a PR is only flagged for issues it introduces on the lines it touches — pre-existing issues on untouched lines stay silent.
- `compile_commands.json` comes from **configuring** the default linux `iwasm` build (interp + AOT runtime, **no LLVM required**); the database is emitted at configure time, so no build step is needed.
- Only changed sources that are **in `compile_commands.json`** are analyzed, each with its real compile flags. A changed file not in the default build — or a header, which has no translation unit of its own — is skipped rather than analyzed without context (which would otherwise produce false failures from missing includes or the wrong `#ifdef` branch).
- A PR that changes no C/C++ source **skips install / configure / lint entirely**, so docs-, test- and CI-only PRs cost almost no Action minutes while the job still reports success.

## clang-tidy version

Runs **clang-tidy 21 on `ubuntu-24.04`**, installed from apt.llvm.org (a small ~50 MB package, not a full LLVM toolchain download). clang-tidy 21 matches the LLVM shipped in Xcode 26/27 and the clang-format-21 gate proposed in #4992, so lint results are consistent whether a contributor builds/auto-formats on macOS/Xcode or on another LLVM-built platform.

## Curated `.clang-tidy`

Moving from clang-tidy 14 to 21 surfaces newer checks. Five are noise/style for this C codebase rather than bug-catchers, so they are disabled (on a full-tree scan they otherwise dominate the output with no correctness signal):

| check | why disabled |
|---|---|
| `misc-include-cleaner` | IWYU-style; ~1750 findings, unusable noise |
| `misc-header-include-cycle` | structural noise |
| `modernize-macro-to-enum` | opinionated for a C codebase |
| `readability-inconsistent-declaration-parameter-name` | cosmetic |
| `bugprone-assignment-in-if-condition` | flags WAMR's deliberate assignment-in-condition idiom |

The **bug-catchers stay on**, including `bugprone-narrowing-conversions`, `bugprone-sizeof-expression`, `bugprone-implicit-widening-of-multiplication-result`, `bugprone-multi-level-implicit-pointer-conversion`, `performance-*`, `clang-analyzer-*` and `cert-*` — the classes that have historically caught real portability / correctness bugs on this project.

## Residual on existing code (why the gate can land now)

A full-tree scan with this config reports roughly 600 diagnostics on the default-config translation units — dominated by `bugprone-narrowing-conversions` — concentrated in `core/iwasm/common` and `core/iwasm/interpreter`. **None of these block contributors**: the gate is diff-scoped, so they only surface when a PR actually edits those lines. The residual is cleaned up incrementally in per-subsystem follow-up PRs (ordered stack in #4983).

## Fork / upstream safety

Same properties as `coding_guidelines.yml`: only `contents: read`, diff from the base SHA — so it behaves identically for in-repo and forked pull requests.

## Notes

- Coverage is scoped to the default linux config's translation units; it can be broadened with additional configs.
- Trigger is `pull_request` only: the gate diffs against the PR base, so a manual (`workflow_dispatch`) run would have no meaningful base.

## Verification

Verified end-to-end against real CI on a throwaway fork PR:

- a deliberate `bugprone-narrowing-conversions` on a changed line of an in-database translation unit **fails** the job — confirming both that configure-only still parses the linux sources and that the diff-scoped fail path works;
- a PR that changes no C/C++ source **skips** the install / configure / lint steps and reports success (this PR's own run);
- a changed file not in the compilation database is skipped rather than mis-analyzed (diff intersected with `compile_commands.json`).


.github/workflows/clang_tidy.yml
```diff
@@ -0,0 +1,101 @@
...
+        run: |
+          # Default linux iwasm config (interp + AOT runtime + libc), no LLVM
+          # required. Covers the highest-traffic loader / interpreter / common
+          # sources; building also materializes any generated headers so
+          # clang-tidy can parse the changed translation units.
+          cmake -S product-mini/platforms/linux -B build \
+            -G Ninja \
+            -DCMAKE_BUILD_TYPE=Debug \
+            -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
+          cmake --build build --target iwasm --parallel
```
lum1n0us: For *compile_commands.json*, there is no need to `cmake --build build`

.github/workflows/clang_tidy.yml
```diff
@@ -0,0 +1,101 @@
...
+  clang_tidy:
+    runs-on: ubuntu-22.04
+    timeout-minutes: 30
+    steps:
+      - name: checkout
+        uses: actions/checkout@v4
+        with:
+          # full history so the pull_request base SHA is available for the diff
+          fetch-depth: 0
+
```
lum1n0us, line 46: Maybe add a quick-exit path for those PRs only affect documents, test cases, CIs to save action minutes?

.github/workflows/clang_tidy.yml
```diff
@@ -0,0 +1,101 @@
...
+      - name: generate compile_commands.json
+        run: |
+          # Default linux iwasm config (interp + AOT runtime + libc), no LLVM
+          # required. Covers the highest-traffic loader / interpreter / common
+          # sources; building also materializes any generated headers so
+          # clang-tidy can parse the changed translation units.
+          cmake -S product-mini/platforms/linux -B build \
+            -G Ninja \
+            -DCMAKE_BUILD_TYPE=Debug \
+            -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
```
lum1n0us, line 99: Generating  compile_commands.json  from the default Linux configuration does cover the highest-traffic sources, but it does not guarantee coverage of every file or every changed line in a PR. As a result, this workflow may miss some modified code paths or analyze files without the correct compilation context

.github/workflows/clang_tidy.yml
```diff
@@ -0,0 +1,101 @@
...
+
+          # -U0: exact changed-line ranges. clang-tidy-diff.py turns those into
+          # -line-filter, so only lines this PR touches are analyzed.
+          if git diff -U0 --no-color "${BASE_SHA}" HEAD -- \
+                '*.c' '*.cc' '*.cpp' '*.h' \
+              | python3 clang-tidy-diff.py \
+                  -p1 -path build \
+                  -clang-tidy-binary "$TIDY" \
+                  -j "$(nproc)" 2>&1 \
+              | tee clang-tidy.log ; then
```
lum1n0us, line 144: When  clang-tidy  is asked to analyze a changed file that is not present in  compile_commands.json , it may fall back to analyzing the file directly without the proper build context. That can produce false failures due to missing macro definitions, wrong conditional-compilation branches, missing include paths, or even diagnostics on unchanged lines in the modified file.

.github/workflows/clang_tidy.yml
lum1n0us, line 46: It may be worth adding a quick-exit path for PRs that only change documentation, test data, or other non-source files. In those cases, the workflow could skip the  clang-tidy  step entirely and save GitHub Actions minutes.

lum1n0us (commented):
Before turning this workflow into an enforced check, should we run a full  clang-tidy  scan on the current codebase and clean up any existing issues first? Otherwise, contributors may end up having to deal with pre-existing diagnostics while making unrelated changes in their PRs.

.github/workflows/clang_tidy.yml
matthargett: Good call — dropped the build. `compile_commands.json` is emitted at configure time, so the job now just configures (no `cmake --build`). Verified on a throwaway self-test PR that the linux TUs still parse without it: clang-tidy caught a deliberate narrowing conversion on a changed line, which means the `configure_file`-generated headers those TUs include are all present after configure alone. (3704755b)

.github/workflows/clang_tidy.yml
matthargett: Added. A cheap `detect changed C/C++ sources` step now runs first; if a PR touches no `*.c/*.cc/*.cpp`, the install / configure / lint steps are all skipped and the job still reports success — so a docs-, test- or CI-only PR costs almost no Action minutes and never blocks. Verified on this PR's own run: it changes no C source, so all four heavy steps were skipped and the job is green. (3704755b)

.github/workflows/clang_tidy.yml
matthargett: Right. The lint step now only analyzes files that are in `compile_commands.json`, each with its real compile command; a changed file outside the default build is skipped rather than analyzed without context. Coverage is intentionally scoped to the default linux config's translation units (the highest-traffic loader / interpreter / common sources) — noted in the PR description — and broadening it with additional configs is an easy follow-up if you'd like it.

.github/workflows/clang_tidy.yml
matthargett: Fixed — this was the important one. The step intersects the changed files with `compile_commands.json` and only runs clang-tidy on those, with `-p build` supplying each file's real flags. A changed source not in the default build (or a header, which has no translation unit of its own) is skipped instead of analyzed without context, so the missing-include / wrong-`#ifdef`-branch false failures you describe can't happen. (3704755b)

.github/workflows/clang_tidy.yml
matthargett: Done — see the `detect changed C/C++ sources` gate above: docs-, test- and non-source-only PRs now skip the whole install + configure + lint body, and the job still reports success. (3704755b)

matthargett:
> Before turning this workflow into an enforced check, should we run a full  clang-tidy  scan on the current codebase and clean up any existing issues first? Otherwise, contributors may end up having to deal with pre-existing diagnostics while making unrelated changes in their PRs.

I tested this locally, and it's a _big_ change. We can split it up and do it incrementally by subdir/component/owner, or do one rule per PR, or a combination of the two, or just do the big bang. I think the big bang sets a bad precedent we wouldn't want other contributions to replicate, and I don't have a strong opinion on which axes to break it up on.

when updating PC-Lint and eslint/prettier versions at previous jobs, beyond a pass to update disable/false-postiive comments, I let the reformats happen incrementally as changes came in (hence the approach in this PR). that said, it may not be great optics or overhead for a public/OSS project, so I'm flexible :)

lum1n0us:
Since default compilation flag set is not able to cover all pre-processors in source code, there is a potential problem that clang-tidy will not be able to scan modifications because `#if...#else...#endif` around them are closed. I am thinking about using https://github.com/wasm-micro-runtime/wasm-micro-runtime/pull/5064 to resolve it by figuring out the right compilation flag combination. Still working on it...

matthargett:
> Since default compilation flag set is not able to cover all pre-processors in source code, there is a potential problem that clang-tidy will not be able to scan modifications because `#if...#else...#endif` around them are closed. I am thinking about using #5064 to resolve it by figuring out the right compilation flag combination. Still working on it...

the single-config blind spot is real, and I ran into it several times while stabilizing CI on our fork's component-model integration branch this week. nearly every real defect lived behind a non-default flag (WASM_ENABLE_UVWASI vs the built-in libc-wasi context, a MULTI_MEMORY-guarded helper whose caller wasn't guarded, SHARED_MEMORY-only macros in a TU that didn't include the header). A diff-scoped scan under the default config is silent on all of those, so flag-tracer is attacking a real pain point for contributors :)

wrt keeping per-PR cost bounded once #5064 can map a diff to its guarding switches: rather than scanning the cross-product of combinations, union the guarding flags of the whole diff into a single extra configure (defaults + everything relevant enabled), and only partition when the tracer finds mutually exclusive flags in the set (e.g. LIBC_UVWASI vs the built-in libc-wasi, or interp-only vs AOT-only). In practice that's 1–3 clang-tidy passes per PR instead of 2^N, and the DB-intersection logic here stays unchanged — it just runs once per config.

while you iterate, can we land this as-is (only gives yellow status in CI on fail, not a required check to pass CI) with one minor improvement for accuracy: when a changed line sits inside a preprocessor region the current config disables, report "N changed lines not analyzed under this configuration" instead of passing silently? That keeps the check useful for the default-config majority today, makes the blind spot visible (insted of invisible), and gives flag-tracer a clean integration point. then we can swap the warning for real extra passes when it's ready. Happy to help test #5064 against this workflow once it's far enough along.

Issue #4983, opened 2026-07-05
CI: move runners to ubuntu-24.04 and use the runner-native LLVM 18 / clang tooling for reliability


matthargett:
## Summary

Move the CI workflows from `ubuntu-22.04` to `ubuntu-24.04`, and bump the pinned clang tooling (`clang-format` / `clang-tidy`, and the LLVM used by the CodeQL / fuzz builds) to the LLVM 18 that the 24.04 runner ships natively. The goal is reliability: several workflows currently depend on an externally-downloaded LLVM tarball plus a hard-coded `libtinfo5` `.deb` URL, and that combination is brittle.

## Why (this is already biting)

- `.github/scripts/codeql_buildscript.sh` downloads the `clang+llvm-18.1.8-...-ubuntu-18.04` release tarball and then installs `libtinfo5` from a **hard-coded** security-pocket URL (`libtinfo5_6.3-2ubuntu0.1_amd64.deb`). That exact file has rotated out of the pool (the runner's ncurses is now `6.3-2ubuntu0.2`), so the `wget` 404s, `libtinfo5` is missing, and linking `wamrc` against the tarball's `libomptarget.rtl.amdgpu` fails with undefined `NCURSES_TINFO_5` symbols. Net effect: **the nightly CodeQL `Analyze` job cannot finish today.** (Fixed defensively in #4980, but the root cause is the ubuntu-18.04 tarball needing a `libtinfo5` that modern runners don't ship.)
- The same hard-coded-`libtinfo5`-URL pattern was independently flagged on the new fuzz-smoke workflow (#4981), which reuses the same LLVM 18.1.8 tarball.
- `ci/coding_guidelines_check.py` pins `clang-format-14` / `git-clang-format-14`, and the new clang-tidy gate (#4982) pins `clang-tidy-14` to match — a version that is now two years behind and not the runner default.

## What ubuntu-24.04 gives us

The `ubuntu-24.04` runner image ships, preinstalled:

- **Clang / LLVM: 16.0.6, 17.0.6, 18.1.3** (and matching `clang-format` / `clang-tidy` at 16/17/18)
- GCC 12 / 13 / 14

[...]

PR #4990, opened 2026-07-10
ci: bump clang-format 14 → 20 and reformat

matthargett:
## Summary

Bumps the coding-guidelines formatter from **clang-format-14 to clang-format-20** and reformats the tree to match. This is the clang-**format** half of #4983's tooling modernization; the clang-tidy bump and the runner / native-LLVM parts follow separately.

## What changed

- **Commit 1 (config):** `ci/coding_guidelines_check.py` and `ci/pre_commit_hook_sample` pin clang-format-20; `coding_guidelines.yml` installs it from `apt.llvm.org` for the running Ubuntu release (it isn't preinstalled), so the pinned version no longer drifts with the runner image — the rot behind today's `-12`/`-14` mix. `.clang-format`'s version note is updated (it also fixes the `clang-formt` typo).
- **Commit 2 (reformat):** mechanical `clang-format-20 -i` over the coding-guidelines-checked sources, as a standalone commit for easy review.

## Noise: small

Only **42 of 441 files (~209 lines)** change — the code was already close to the clang-format-20 style, and `.clang-format` needed no option changes (zero config warnings at 20). The reformat is also version-flat: 39 files at clang-format-18, 43 at 21/22, so pinning 20 carries no penalty over the runner-preinstalled 18 while being an explicitly-pinned latest stable.

## Scope

Intentionally just clang-**format**. The clang-**tidy** bump to 20 is separate: at 20 the current `.clang-tidy` wildcards surface ~3,300 diagnostics on the tree (53% `misc-include-cleaner`), which needs config curation plus incremental cleanup — a follow-up tied to #4982, not bundled here.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,4 +1,4 @@
-# using [clang-formt-12 options](https://releases.llvm.org/12.0.0/tools/clang/docs/ClangFormatStyleOptions.html)
+# using [clang-format-20 options](https://releases.llvm.org/20.1.0/tools/clang/docs/ClangFormatStyleOptions.html)
 RawStringFormats:
   - Language: Cpp
     Delimiters:
--- a/.github/workflows/coding_guidelines.yml
+++ b/.github/workflows/coding_guidelines.yml
@@ -26,6 +26,20 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: install clang-format-20
+        run: |
+          # clang-format 20 is not preinstalled on the runner; install it (and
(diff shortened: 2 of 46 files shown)


matthargett:
Superseded by https://github.com/bytecodealliance/wasm-micro-runtime/pull/4991 — switched to ubuntu-24.04's preinstalled clang-format-18 (no external download) instead of installing clang-format-20 from apt.llvm.org.

Issue #2833, opened 2023-11-28
Different output when running a same binary


XinyuShe:
I used the AOT mode of different runtimes to run randomly generated wasm binaries, and the output from wamr and wasmedge are different from the others, and many binaries have similar situation
[filea12287_4.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/13489969/filea12287_4.zip)

```bash
root@4252f5ec38df:/a# iwasm --version
iwasm 1.2.3
root@4252f5ec38df:a# wamrc --version
wamrc 1.2.3
```
## AOT
```bash
-----------------wasmtime-----------------
Error: failed to run main module `/a/filea12287_4.cwasm`

Caused by:
    0: failed to invoke `main`
    1: error while executing at wasm backtrace:
           0: 0x3776 - <unknown>!<wasm function 25>
           1: 0x382f - <unknown>!<wasm function 26>
           2: 0x9f3e - <unknown>!<wasm function 79>
    2: memory fault at wasm address 0x100002305 in linear memory of size 0x100000000
    3: wasm trap: out of bounds memory access

-----------------wasmer-----------------
Compiler: cranelift
Target: x86_64-unknown-linux-gnu
✔ File compiled successfully to `/a/filea12287_4.wasmu`.
error: failed to run `/a/filea12287_4.wasmu`
╰─▶ 1: RuntimeError: out of bounds memory access
           at <unnamed> (/a/filea12287_4.wasm[25]:0x3776)
           at <unnamed> (/a/filea12287_4.wasm[26]:0x382f)
           at <unnamed> (/a/filea12287_4.wasm[79]:0x9f3e)

-----------------wamr-----------------
Create AoT compiler with:
  target:        x86_64
  target cpu:    znver3
  target triple: x86_64-unknown-linux-gnu
  cpu features:  
  opt level:     3
  size level:    3
  output format: AoT file
Compile success, file /a/filea12287_4.wasm.aot was generated.
<0x0000000000000000 0x0000000000000000>:v128,<0x0000000000000000 0x0000000000000000>:v128,<0x0000000000000000 0x0000000000000000>:v128,<0x0000000000000000 0x0000000000000000>:v128,<0x0000000000000000 0x0000000000000000>:v128,<0x0000000000000000 0x0000000000000000>:v128,<0x0000000000000000 0x0000000000000000>:v128

-----------------wasmedge-----------------
[2023-11-27 14:52:32.675] [info] compile start
[...]

Issue #2702, opened 2023-11-02
The implementation of instruction 'i64x2.mul' may have bugs


XinyuShe:
I constructed a wasm binary with an exported function called "main". I used the following command and compared the outputs of WAMR with other runtimes.

[filea105487.zip](https://github.com/bytecodealliance/wasm-micro-runtime/files/13233464/filea105487.zip)

The outputs of WAMR are different from other runtimes.
The outputs are as follows:

```
filea105487.wasm
-----------------wamr-----------------
Create AoT compiler with:
  target:        x86_64
  target cpu:    znver3
  target triple: x86_64-unknown-linux-gnu
  cpu features:  
  opt level:     3
  size level:    3
  output format: AoT file
Compile success, file filea105487.wasm.aot was generated.
<0x0000000000000000 0x0000000000000000>:v128,<0x0000000000000000 0x0000000000000000>:v128,<0x0000000000000000 0x0000000000000000>:v128,<0x0000000000000000 0x0000000000000000>:v128,<0x0000000000000000 0x0000000000000000>:v128,<0x0000000000000000 0x0000000000000000>:v128,<0x0000000000000000 0x0000000000000000>:v128
-----------------wasmedge-----------------
142531464396347235927115696178951
142531464396347235927115696178951
142531464396347235927115696178951
142531464396347235927115696178951
142531464396347235927115696178951
142531464396347235927115696178951
142531464396347235927115696178951
-----------------wasmtime-----------------
warning: using `--invoke` with a function that returns values is experimental and may break in the future
142531464396347235927115696178951
142531464396347235927115696178951
142531464396347235927115696178951
142531464396347235927115696178951
142531464396347235927115696178951
142531464396347235927115696178951
142531464396347235927115696178951
```

I have analyzed it and I think the error may occurred at line 77 of its wat file
![image](https://github.com/bytecodealliance/wasm-micro-runtime/assets/60457190/610e48c1-8b7a-41fd-9076-a8218a2668b3)

PR #4998, opened 2026-07-17
aot: pin host CPU features for the default native target (fix SIGILL on AVX-512-masked runners)

### LINKED ISSUE #4997: build_regression_tests: 4 SIMD-AOT ba-issue cases crash with SIGILL on ubuntu-22.04 (pre-existing, masked by the LLVM-cache failure)
## Summary

Four `tests/regression/ba-issues` SIMD / `v128` cases fail in `build_regression_tests (ubuntu-22.04)`, all crashing with **`SIGILL`** (illegal instruction, exit `-4`) when the `wamrc`-generated `.aot` runs:

| issue | expected | observed |
|---|---|---|
| #2833 | `<0x0…>:v128` ×7, ret 0 | ret **-4 (SIGILL)** |
| #270801 | `<0xffffffff00000000 …>:v128` ×7, ret 0 | ret **-4 (SIGILL)** |
| #270802 | `<0xffffffff00000000 …>:v128` ×7, ret 0 | ret **-4 (SIGILL)** |
| #2702 | `<0x0000070700000707 …>:v128` ×7, ret 0 | ret **-4 (SIGILL)** |

`Total: 106, Passed: 102, Failed: 4`. Reproduction:

```
wamrc -o out.aot tests/regression/ba-issues/issues/issue-2833/filea12287_4.wasm
iwasm --heap-size=0 -f main out.aot      # -> SIGILL
```

## Root cause: the AOT emits AVX-512 the runner can't execute

GitHub moved the ubuntu-22.04 image onto **AMD EPYC 9V74 (Genoa)** VMs whose `/proc/cpuinfo` advertises `avx avx2 fma f16c sse4_2 vaes vpclmulqdq` — **no `avx512*` flag** (AVX-512 is masked off in the VM).

But `wamrc`'s default host-targeted AOT (no `--target`/`--cpu`) selects the target with `LLVMGetHostCPUName()` — which classifies Genoa as an AVX-512-capable model — and leaves the feature string empty, so LLVM enables the model-implied AVX-512. The emitted `.aot` is full of AVX-512 (`vmovdqa64 %zmm0{%k1}{z}`, `vporq %zmm`, `vpternlogd`, `vextracti64x4`) and SIGILLs on the first one:

```
Program received signal SIGILL, Illegal instruction.
=> 0x40006878: vcmpordpd %xmm1,%xmm0,%k1      # opmask destination -> AVX-512 (EVEX)
```

The JIT/ORC target-machine setup already pairs the host CPU *name* with `LLVMGetHostCPUFeatures()` (CPUID-accurate); the AOT default path does not, so it targets the CPU *model* rather than the ISA actually enabled on the VM.

**Pre-existing and LLVM-version-independent** — reproduces identically on #4951 (LLVM 22) and current `main` (LLVM 18). On `main` it is currently masked by a separate LLVM-cache CI break (fixed in #4996); the `compilation on android, ubuntu-22.04` workflow last passed on 2026-05-19 (`5aebe728`), matching when the image moved onto these AMD VMs.

## Fix — #4998

In the default native AOT path, also query `LLVMGetHostCPUFeatures()` so codegen is capped at the runner's actual CPUID feature set (AVX-512 masked off), mirroring the JIT setup. Explicit `--cpu` / `--cpu-features` / `--target` builds are unaffected.

**Verified** on an EPYC 9V74 runner in my rebeckerspecialties fork: with the fix there is no AVX-512 in the emitted object, and the AOT runs to the expected `…:v128` output with exit 0 — no SIGILL.

matthargett:
## Problem

On the current ubuntu-22.04 (and macOS) GitHub-hosted runners, AOT-compiled modules that use SIMD crash at runtime with **SIGILL**. It shows up in `build_regression_tests` — four `ba-issues` SIMD cases (#2833, #270801, #270802, #2702) fail with exit `-4` — and is tracked in #4997.

## Root cause

When `wamrc` runs with no `--target` and no `--cpu`, `aot_create_comp_context` picks the target CPU with `LLVMGetHostCPUName()` but leaves the feature string empty:

```c
if (!triple && !cpu) {
    triple = LLVMGetDefaultTargetTriple();
    cpu    = LLVMGetHostCPUName();   /* e.g. "znver4" */
    /* features left empty */
}
```

Given a CPU *name* but no feature string, LLVM turns on **every feature that model implies — including AVX-512**. GitHub now backs some ubuntu-22.04 runners with **AMD EPYC 9V74 (Zen 4)** VMs whose reported CPU name implies AVX-512 but which have AVX-512 **masked out of cpuid**, so the AOT module ends up with AVX-512 instructions the running CPU can't execute.

Captured on a runner under gdb:

```
Program received signal SIGILL, Illegal instruction.
=> 0x40006878:  vcmpordpd %xmm1,%xmm0,%k1      # writes an AVX-512 mask register
```

with `vmovdqa64 %zmm0{%k1}{z}`, `vpternlogd`, `vextracti64x4` in the emitted object, while `/proc/cpuinfo` on that VM lists no `avx512*` flags.

## Fix

In that default native path, also query `LLVMGetHostCPUFeatures()` so the emitted code matches the features **actually enabled** on the host (AVX-512 masked off) — mirroring what the JIT target-machine setup already does. Explicit `--cpu` / `--cpu-features` and cross-target (`--target`) builds are unaffected.

## Verification

Rebuilt `wamrc` with the fix on an EPYC 9V74 runner and recompiled the failing case:

- **no AVX-512** instructions remain in the emitted object;
- the AOT runs to completion (**exit 0**) with the expected `…:v128` output — no SIGILL.

Fixes #4997.

### DIFF
--- a/core/iwasm/compilation/aot_llvm.c
+++ b/core/iwasm/compilation/aot_llvm.c
@@ -2632,7 +2632,7 @@ aot_create_comp_context(const AOTCompData *comp_data, aot_comp_option_t option)
     LLVMTargetRef target;
     char *triple = NULL, *triple_norm, *arch, *abi;
     char *cpu = NULL, *features, buf[128];
-    char *triple_norm_new = NULL, *cpu_new = NULL;
+    char *triple_norm_new = NULL, *cpu_new = NULL, *features_new = NULL;
     char *err = NULL, *fp_round = "round.tonearest",
          *fp_exce = "fpexcept.strict";
     char triple_buf[128] = { 0 }, features_buf[128] = { 0 };
@@ -3115,6 +3115,15 @@ aot_create_comp_context(const AOTCompData *comp_data, aot_comp_option_t option)
                 aot_set_last_error("llvm get host cpu name failed.");
                 goto fail;
             }
+            /* Also pin the host CPU's actually-enabled features. The CPU name
+             * alone (e.g. "znver4") makes LLVM enable every feature that model
+             * implies, including AVX-512; but a virtualized host can have those
+             * masked out of cpuid, so without the real feature string the AOT
+             * code may emit instructions the running CPU rejects with SIGILL.
+             * LLVMGetHostCPUFeatures() reflects what is actually enabled, and
+             * this mirrors the JIT target-machine setup. */
+            if (!features)
+                features = features_new = LLVMGetHostCPUFeatures();
         }
         else if (triple) {
             /* Normalize a target triple */
@@ -3499,6 +3508,9 @@ fail:
 
     if (cpu_new)
         LLVMDisposeMessage(cpu_new);
+
+    if (features_new)
+        LLVMDisposeMessage(features_new);
 
     if (!ret)
         aot_destroy_comp_context(comp_ctx);


core/iwasm/compilation/aot_llvm.c
```diff
@@ -3115,6 +3115,15 @@ aot_create_comp_context(const AOTCompData *comp_data, aot_comp_option_t option)
...
             }
+            /* Also pin the host CPU's actually-enabled features. The CPU name
+             * alone (e.g. "znver4") makes LLVM enable every feature that model
+             * implies, including AVX-512; but a virtualized host can have those
+             * masked out of cpuid, so without the real feature string the AOT
+             * code may emit instructions the running CPU rejects with SIGILL.
+             * LLVMGetHostCPUFeatures() reflects what is actually enabled, and
+             * this mirrors the JIT target-machine setup. */
+            if (!features)
+                features = features_new = LLVMGetHostCPUFeatures();
```
lum1n0us, line 3126: seems `features_new` is assigned but not used. May I ask for its purpose?

matthargett: it is used, just at the cleanup label. `features_new` is the owning pointer for the heap string returned by `LLVMGetHostCPUFeatures()`, so it can be released at `fail:` (the `if (features_new) LLVMDisposeMessage(features_new);` a bit further down). I kept it separate from `features` on purpose: downstream `features` gets repointed to non-owned memory — the `features_buf` stack buffer when `,+d` is appended, or the `"+d"`/`""` string literals — so freeing `features` directly wouldn't be safe. This mirrors how `cpu_new` (for `LLVMGetHostCPUName()`) and `triple_norm_new` are handled right above it. Without the separate variable the feature string would leak. lmk if you're seeing something I'm not, of course! :D

lum1n0us (approved):

PR #4999, opened 2026-07-17
ci: fix nuttx builds failing with 'no space left on device'

matthargett:
## Problem

`compilation on nuttx` is intermittently red on `main` and across PRs — every `build_iwasm_on_nuttx` leg fails with **`no space left on device`**, in two places: the container **image pull** (`Docker pull failed`, the dominant mode) and **during the build** while compiling the from-source `bloaty` (protobuf/abseil).

## Root cause

It's a **`container:` job** on the large `apache-nuttx-ci-linux` image, and it also builds `google/bloaty` (recursive protobuf/abseil submodules) from source on every matrix leg for the size report. A container job pulls its image **at job init, before any step runs**, so no step can free disk first — and the runner's default free space plus that from-source build doesn't fit.

## Fix

- **Drop `container:`.** Run on `ubuntu-latest`, remove the unused preinstalled toolchains (~25 GB) as step 1, then run the NuttX configure+build inside the apache image via `docker run` with the workspace mounted. The image is pulled *after* disk is freed (verified ~119 GB free), so both modes are fixed.
- **Build bloaty once + cache it.** A small `build_bloaty` job compiles the pinned bloaty in the image and caches the binary, keyed on **`BLOATY_REF`** — bump the ref and the cache key changes, so it rebuilds and refreshes automatically. The matrix `needs: build_bloaty` and restores the prebuilt binary, so 63 protobuf builds collapse to one. The size report (`bloaty -d compileunits --source-filter wamr/libc-builtin/libc-wasi`) is **byte-for-byte unchanged** — the MCU binary-size regression signal is fully preserved.
- Drop a **duplicated `wamr_config_option` matrix entry** (it was listed twice), removing a redundant leg.

## Verification (on a fork)

- Free-disk gives `/dev/root 145G 26G 119G 18% /`; the apache image pulls; **all board matrix legs build and link**, no ENOSPC.
- `build_bloaty` builds once and `Cache saved with key: bloaty-<ref>-ubuntu-latest`; the matrix legs `Cache restored from key: bloaty-<ref>-...` (restore, not rebuild) and the size report prints.

## CI cost

Measured on the fork runs: building bloaty (protobuf/abseil/…) from source adds **~3 min to every matrix leg** (~8–9.5 min per leg vs **~5.7 min** with a cache restore), and it was being done on all 56 legs. Doing it **once** (~7.6 min, and ~free on a warm cache) instead saves roughly **56 × 3 ≈ 170 Actions-minutes per run**; dropping the duplicated matrix leg saves ~7 × 5.7 ≈ 40 more — about **~200 Actions-minutes per nuttx run**.

At the current **~18 nuttx runs/week** on this repo that's roughly **~3,600 Actions-minutes (~60 hours) per week** (warm cache; a bloaty-ref bump adds one ~8-min rebuild that run). This is on top of no longer wasting minutes on the ENOSPC-failed runs.

### DIFF
--- a/.github/workflows/compilation_on_nuttx.yml
+++ b/.github/workflows/compilation_on_nuttx.yml
@@ -45,16 +45,62 @@ concurrency:
 
 env:
   WASI_SDK_PATH: "/opt/wasi-sdk"
+  NUTTX_IMAGE: "ghcr.io/apache/nuttx/apache-nuttx-ci-linux@sha256:d9261eacf6c6ebe656c571757751c803e8f04c3ae9b820320a5ea5dd57b7205a"
+  # Bump BLOATY_REF to update bloaty. The size-report cache below is keyed on
+  # it, so a new ref rebuilds bloaty and refreshes the cache automatically.
+  BLOATY_REF: "34f4a66559ad4938c1e629e9b5f54630b2b4d7b0"
 
 permissions:
   contents: read
 
 jobs:
-  build_iwasm_on_nuttx:
+  # Build bloaty once and cache it, keyed on BLOATY_REF (a new ref -> cache
+  # miss -> rebuild + refresh). The matrix below restores the prebuilt binary
+  # instead of every leg compiling bloaty and its protobuf/abseil submodules.
+  build_bloaty:
     runs-on: ubuntu-latest
-    container:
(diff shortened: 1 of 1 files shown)


lum1n0us (approved):
@no1wudi I'll go ahead and merge this PR — feel free to share any comments you may have!

PR #5008, opened 2026-07-23
Documentation Update For WAMR's New Home

woodsmc:
This is a documentation update that reflects that WAMR will no longer be a Bytecode Alliance project. This PR updates the documentation in three key ways:

1. It removes references to the Bytecode Alliance and statements that this is a Bytecode Alliance project.
2. It inherits the security practices that WAMR previously operated under as part of the Bytecode Alliance. Modified to reflect that this is now a WAMR only practice.
3. It inherits the code of conduct that WAMR previously operated under as part of the Bytecode Alliance. Modified to reflect that this is now a WAMR only practice.

**Note**: This does not change the URLs in the documentation. This ensures that this PR can be committed ahead of the migration. A further PR will be required to update the URLs. 

This early change is designed to allow review and approval of the above changes.


SECURITY.md
```diff
@@ -1,3 +1,42 @@
 # Security Policy
 
-Please refer to the [Bytecode Alliance security policy](https://bytecodealliance.org/security) for details on how to report security issues in WebAssembly Micro Runtime, our disclosure policy, and how to receive notifications about security issues.
+## Reporting a security bug in a the WAMR Project
```
yamt, line 3: this section looks like a modified copy of https://bytecodealliance.org/security.
i wonder if it's copyright-wise ok.

lum1n0us (approved):

SECURITY.md
lum1n0us: I guess not. And I assume this is not the final version. We will define our own version soon.

TianlongLiang (approved):

gitbook/home_page.md
```diff
@@ -1,6 +1,6 @@
 # Welcome
 
-Welcome to the home page of WAMR documentation, [WebAssembly Micro Runtime](https://github.com/bytecodealliance/wasm-micro-runtime) is an open-source project under [Bytecode Alliance](https://bytecodealliance.org/). As the name suggests, it is a lightweight standalone WebAssembly (WASM) runtime with a small footprint, high performance, and highly configurable features for applications across from embedded, IoT, edge Trusted Execution Environment (TEE), smart contract, cloud-native, and so on.
+Welcome to the home page of WAMR documentation, [WebAssembly Micro Runtime](https://github.com/bytecodealliance/wasm-micro-runtime) is an open-source project. As the name suggests, it is a lightweight standalone WebAssembly (WASM) runtime with a small footprint, high performance, and highly configurable features for applications across from embedded, IoT, edge Trusted Execution Environment (TEE), smart contract, cloud-native, and so on.
```
ttrenner, line 3: link is still on bytecodealliance

woodsmc: Yes. The links are intentionally unchanged. As per the original discussion. Allows changes to documentation before the transfer. Links will work while still hosted under BA as we prepare for change.

SECURITY.md
woodsmc: Correct. We inherit the same security process as currently used. In the past this was a link to the BA's process. In the future, it should be the same process, using WAMR specific channels. Hence why it copies the original BA text.

woodsmc: > I guess not. And I assume this is not the final version. We will define our own version soon.

Soon as in, before Monday? - Let's see if there is any objections from the BA. Ideally we'd implement the same process, and best practices. We should base this on the the original BA text.

PR #5010, opened 2026-07-25
Prefer OptionalPassInfoMixin

boomanaiden154:
LLVM 23 prefers {Optional,Required}PassInfoMixin rather than PassInfoMixin and explicitly (or not) setting isRequired. In LLVM 24, PassInfoMixin will be moved to a detail namespace and thus will cause a compilation failure.

### DIFF
--- a/core/iwasm/compilation/aot_llvm_extra.cpp
+++ b/core/iwasm/compilation/aot_llvm_extra.cpp
@@ -86,7 +86,12 @@ LLVM_C_EXTERN_C_END
 
 ExitOnError ExitOnErr;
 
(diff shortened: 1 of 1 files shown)


boomanaiden154:
Bump. Can I get a review on this? It should be a simple patch and without this, builds with ToT LLVM will start failing in the next week or two.

yamt (approved):
for other reviewers: https://github.com/llvm/llvm-project/pull/192120

boomanaiden154:
Bump. I'm going to put up PRs tonight to drop `PassInfoMixin`.

lum1n0us (approved):

TianlongLiang (approved):

PR #5011, opened 2026-07-27
Refactor ci trigger

lum1n0us:
Move pre-merge CI onto the contributor's fork and spend upstream minutes
only when a maintainer actually approves.

CI over a PR's lifecycle (assume the PR is updated N times = N pushes):

- Fork (contributor's own minutes): every push to a non-main/release/dev
  branch triggers the push-event CI. Over the PR that is N runs per
  workflow, each still subject to its on.push.paths filter (only the
  workflows whose files changed actually run).
- Upstream (project minutes): nothing runs on push. A `pull_request_review`
  (submitted) event fires the `gate` job; the first APPROVED review of the
  PR lets the real jobs run once, later approvals are skipped. Because a new
  push dismisses the stale approval, each "push -> re-approve" cycle allows
  at most one full upstream run. Best case a clean PR costs upstream exactly
  1 run; a PR re-approved after k re-pushes costs k upstream runs.

Required repository settings:

- Branch protection: "Dismiss stale pull request approvals when new commits
  are pushed" MUST be enabled. The gate counts APPROVED reviews and skips
  when >1; dismissal is what resets the count so a re-approval can trigger a
  fresh upstream run after new commits.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -7,7 +7,7 @@ assignees: ""
 ---
 
 **Is it a security vulnerability?**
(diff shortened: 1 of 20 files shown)


srberard (approved):
LGTM.  

It does look like this work needs to be updated to align with `main`.

PR #5037, opened 2026-08-11
fix(fast-jit): support shared-heap memory access

pymergetic:
## Summary

- Enable `WAMR_BUILD_SHARED_HEAP=1` + `WAMR_BUILD_FAST_JIT=1` by teaching fast-jit shared-heap address translation (previously hard-rejected in `unsupported_combination.cmake`).
- Memory loads/stores/atomics/bulk: gate on the shared-heap **chain-head** start so the linear hot path stays call-free; shared/chain addresses go through a small runtime helper. Absolute native `maddr` is used for memory ops (same model as AOT).
- Native imports with `*` / `*~`: zero-extend app offsets (`U32TOI64` instead of sign-extend) and reload the `jit_cache` out-param after `callnative` so shared-heap pointers near `0xFFFFF000` convert correctly.
- Extend `tests/unit/shared-heap` to exercise `Mode_Fast_JIT`.

## Test plan

- [x] Build product-mini with `FAST_JIT=1` + `SHARED_HEAP=1` (cmake guard lifted)
- [x] Shared-heap harness: advanced **72/72**, extreme **88/88** (HW + soft bounds)
- [x] Fallout: FAST_JIT without SHARED_HEAP; SHARED_HEAP with no heap attached; soft bounds — **27/27** each
- [x] Atomics + shared heap (`SHARED_MEMORY=1`): **52/52** (interp + fast-jit + parity)
- [ ] CI / `tests/unit/shared-heap` with LLVM available

## Notes

- Memory64 / multi-memory / SIMD remain unsupported with fast-jit (existing cmake constraints).
- Approach mirrors AOT’s shared-heap split (linear vs shared) but uses a helper for the cold shared/chain path to fit fast-jit’s BB/regalloc model.

Made with [Cursor](https://cursor.com)


core/iwasm/fast-jit/fe/jit_emit_function.c
```diff
@@ -313,9 +313,6 @@ jit_compile_op_call(JitCompContext *cc, uint32 func_idx, bool tail_call)
 
         ret = jit_cc_new_reg_I32(cc);
         func_params[0] = module_inst_reg = get_module_inst_reg(jit_frame);
-        func_params[4] = native_addr_ptr = jit_cc_new_reg_ptr(cc);
-        GEN_INSN(ADD, native_addr_ptr, cc->exec_env_reg,
```
TianlongLiang, line 317: What is the purpose of removing this parameter? Isn't this an ABI-breaking change

pymergetic: Not ABI-breaking — the 5th parameter wasn’t removed.

jit_check_app_addr_and_convert is still called with 5 args; func_params[4] is still &exec_env->jit_cache (void **p_native_addr). The pre-loop setup was only moved into the pointer-arg path: we rebuild that address immediately before callnative, and again after it before LDPTR, because the ptr reg is caller-saved and isn’t safe to keep live across the call. Same callee / arity / out-param; this is just so the converted native address is read back correctly (needed for shared-heap * / *~).

PR #5042, opened 2026-08-13
build: centralize external tool discovery

lum1n0us:
Move WASI-SDK, WABT, Binaryen, wamrc, and Emscripten discovery into modern Find modules and migrate CMake callers.

Use find_package(<Tool> [version] [REQUIRED|QUIET] [COMPONENTS ...]) with build-scripts on CMAKE_MODULE_PATH; consume exported targets or compatibility variables instead of fixed paths.


srberard (dismissed):

srberard (approved):
LGTM

PR #5047, opened 2026-08-14
zephyr: add socket fd flag handling via fcntl()

### LINKED ISSUE #5046: zephyr: add non-blocking socket flag support
**Problem description**

Socket-backed file descriptors did not correctly support WASI fd flag operations on Zephyr. The existing implementation relied on Zephyr filesystem descriptor handling to retrieve and update descriptor flags, which is only valid for regular file descriptors.

As a result, applications attempting to enable or query non-blocking mode on sockets through WASI APIs could receive incorrect behavior because socket descriptors are managed through the Zephyr networking stack rather than the filesystem layer.

---

**Describe the solution you'd like**

Add dedicated handling for socket-backed file descriptors in `os_file_get_fdflags()` and `os_file_set_fdflags()`.

When the handle represents a socket:

- Use `fcntl(F_GETFL)` to retrieve the socket status flags.
- Map `O_NONBLOCK` to the corresponding WASI flag (`__WASI_FDFLAG_NONBLOCK`).
- Use `fcntl(F_SETFL)` to enable or disable non-blocking operation when requested through WASI APIs.
- Return appropriate WASI error codes when the underlying socket operation fails.

Regular file descriptors should continue using the existing filesystem-based implementation to preserve current behavior.

---

**Describe alternatives you've considered**

- Extending the existing filesystem descriptor implementation to handle sockets. This was rejected because socket descriptors are not compatible with Zephyr filesystem APIs.
- Maintaining separate socket-specific code paths in higher layers. This would duplicate logic and break the abstraction provided by the platform layer.
- Ignoring fd flag management for sockets. This would leave WASI non-blocking socket support incomplete.

---

**Additional context**

This change improves WASI compatibility on Zephyr by correctly supporting non-blocking socket operation through `fd_fdstat_get` and `fd_fdstat_set_flags` related paths.

The implementation detects socket-backed handles via `handle->is_sock` and delegates flag management to the native socket API (`fcntl`) while preserving the existing behavior for filesystem descriptors.

This fixes issues encountered when applications configure sockets for non-blocking I/O and aligns socket flag handling with standard POSIX expectations.

groncarolonxp:
Handle fd flags for socket-backed file descriptors separately from Zephyr FS descriptors.

When the handle represents a socket, use fcntl(F_GETFL/F_SETFL) to retrieve and set flags, including support for O_NONBLOCK. This avoids using Zephyr FS APIs on sockets, which are not compatible.

Regular file descriptors keep the existing behavior based on zephyr_fs_desc and FS_O_APPEND.

Review of Zephyr socket non-blocking fix

Fixes #5046
Show more lines


core/shared/platform/zephyr/zephyr_file.c
```diff
@@ -9,6 +9,7 @@
 
 #include <string.h>
 #include <stdlib.h>
+#include <fcntl.h>
```
lum1n0us: `zephyr/posix/fcntl.h` ?

lum1n0us (commented):
Recently, zephyr samples have been modernized and please make sure they are workable with this PR. You will figure out how to run with the *product-mini/platforms/zephyr/README.md*.

core/shared/platform/zephyr/zephyr_file.c
groncarolonxp: hi @lum1n0us , thanks for having a look at this. I had the test running: found some issues and fixed it. this is a short summary of what i did   

1. ed164f77 - Socket fd flag handling via fcntl()
      Added separate handling for socket-backed file descriptors in zephyr_file.c's fdflags get/set logic, using fcntl(F_GETFL/F_SETFL) (including O_NONBLOCK support) instead of Zephyr FS APIs, which don't work on sockets.


2. 71060b3a - simple-http main.c cleanup
      Fixed main() to return -1 explicitly on wasm_runtime_full_init() failure (was relying on undefined fallthrough), and added a missing trailing newline.


3. 761a70e2 - update_wasm_header.sh helper script
      Added a script to automate rebuilding http_get.wasm and regenerating src/http_get.h, replacing the previous manual multi-step process; documented as a shortcut in the README.


4. ee722f03 - prj.conf Kconfig updates
      Enabled the heap memory pool (CONFIG_HEAP_MEM_POOL_SIZE) and POSIX support (CONFIG_POSIX_SYSTEM_INTERFACES, CONFIG_POSIX_TIMERS, CONFIG_POSIX_FD_MGMT) needed for fcntl()/dup()/lseek()/ftruncate() used by zephyr_file.c.


5. 7a40afbb - Fix socket misclassification in os_file_get_access_mode()
      Moved the is_sock check to the top of the function, before the stdin/stdout/stderr fd-number checks. Without CONFIG_POSIX_DEVICE_IO, a socket can legitimately occupy fd 0/1/2, and the old ordering would wrongly mark it
      read-only/write-only, causing send()/recv() to fail with __WASI_ENOTCAPABLE.

srberard (dismissed):

product-mini/platforms/zephyr/simple-http/update_wasm_header.sh
lum1n0us, line 1: To ensure cross-platform compatibility as required, we should probably avoid using Bash scripts here. I'd highly recommend going with either Python or a CMake script instead. For reference, there's a great example of `add_custom_command()` in *zephyr/simple* that you might find helpful to look at.

lum1n0us (changes requested):

product-mini/platforms/zephyr/simple-http/update_wasm_header.sh
groncarolonxp: cool, I removed the script, I saw the complete loop was already implemented in the cmake :)

groncarolonxp: [wasm-mod] Preparing HTTP GET request for http://192.0.2.10:8000/
[wasm-mod] sock = 6
[wasm-mod] connect rc = 0
[wasm-mod] send rc = 36
[wasm-mod] Response:

HTTP/1.0 200 OK
Server: SimpleHTTP/0.6 Python/3.10.12
Date: Tue, 25 Aug 2026 05:41:28 GMT
Content-type: text/html; charset=utf-8
Content-Length: 1567

<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN" "http://www.w3.org/TR/html4/strict.dtd">
<html>
<head>
<meta http-equiv="Content-Type" content="text/html; charset=utf-8">
<title>Directory listing for /</title>
</head>
<body>
<h1>Directory listing for /</h1>
<hr>
<ul>
[BLA BLA]
</ul>
<hr>
</body>
</html>
[wasm-mod] len = 0 break

[wasm-mod] Connection closed
main executed
wasi exit code: 0
PASS: the HTTP request completed
elapsed: 236ms

groncarolonxp: rebased on latest main and pruned unecessary commits

lum1n0us (approved):
LGTM

PR #5051, opened 2026-08-17
test_wamr.sh: fetch the wabt binary release with curl instead of wget

matthargett:
## What breaks

The `windows-2022` runner image no longer ships `wget` in its bash environment. Every spec-test leg that uses the `-b` wabt-binary-release path (e.g. `test (classic-interp, $THREADS_TEST_OPTIONS)` in `compilation_on_windows.yml`) now dies during setup:

```
./test_wamr.sh: line 529: wget: command not found
tar: wabt-tar.gz: Cannot open: No such file or directory
Can not find ./wabt/out/gcc/Release/wat2wasm.exe
spec tests FAILED
```

The workflow's last green run on main was 2026-07-27, which predates the runner-image update — the next windows run on main will hit this. (Observed on a fork running the identical workflow today.)

## Fix

Fetch the release with `curl -L`, which is preinstalled on all hosted runners including Windows. One line, no behavior change on the happy path.

### DIFF
--- a/.github/actions/install-wasi-sdk-wabt/action.yml
+++ b/.github/actions/install-wasi-sdk-wabt/action.yml
@@ -41,7 +41,7 @@ runs:
       run: |
         SDK_DIR="wasi-sdk-${SDK_VER}.0-x86_64-linux"
         echo "Downloading wasi-sdk for Ubuntu..."
-        sudo wget -O wasi-sdk.tar.gz --progress=dot:giga \
+        sudo curl -fsSL -o wasi-sdk.tar.gz \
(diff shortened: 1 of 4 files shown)


lum1n0us (approved):

lum1n0us (changes requested):
https://github.com/wasm-micro-runtime/wasm-micro-runtime/actions/runs/34582736825/job/103209882620?pr=5051

Oops, missing dependency. Zephyr CI runs in a custom container, so no curl out of the box. Let's install it on the fly for now

lum1n0us (dismissed):

lum1n0us:
> If you'd rather not depend on what the image ships, I can make setup_wabt try curl and fall back to wget instead — happy to do that if you prefer it.

SGTM! Let's give it a shot.

lum1n0us (approved):

PR #5052, opened 2026-08-17
tests/unit: fetch cmocka from the GitLab mirror, pinned by hash

matthargett:
## What breaks

`tests/unit/CMakeLists.txt` fetches cmocka from `git.cryptomilk.org`'s cgit snapshot endpoint, which generates archives on the fly and intermittently serves error pages under load. When several unit-test jobs run in parallel, FetchContent saves the error page as the tarball and the configure step dies before any test builds:

```
CMake Error: Problem extracting tar: .../cmocka-2.0.1.tar.gz
CMake Error: Problem with archive_read_open_file(): Unrecognized archive format
```

Observed repeatedly today across `unit tests (X86_64)` / `unit tests (X86_32)` legs on a fork running the identical workflow (while the same URL served a valid archive between runs — it is load-dependent).

## Fix

Fetch the same `cmocka-2.0.1` release from the project's official GitLab mirror (`gitlab.com/cmocka/cmocka`), whose archive endpoint is CDN-backed, and pin it with `URL_HASH SHA256=...`. I verified the extracted trees of the cryptomilk and GitLab archives are identical. The hash also turns any future bad download into a loud, immediate failure instead of an inscrutable extraction error.

### DIFF
--- a/tests/unit/CMakeLists.txt
+++ b/tests/unit/CMakeLists.txt
@@ -78,13 +78,15 @@ FetchContent_MakeAvailable(googletest)
 if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.24")
   FetchContent_Declare(
     cmocka
-    URL https://git.cryptomilk.org/projects/cmocka.git/snapshot/cmocka-2.0.1.tar.gz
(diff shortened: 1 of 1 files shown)


lum1n0us (approved):
LGTM

PR #5061, opened 2026-08-22
interpreter: load unaligned wasm headers safely

matthargett:
## Problem

`wasm_runtime_load` reads the Wasm magic and version by casting the caller's byte pointer to `uint32 *`. The public API accepts byte buffers without an alignment requirement, so a valid module at an odd address triggers undefined behavior and can fault on strict-alignment targets.

## Fix

Copy the two fixed-width header fields into aligned local storage before decoding them. The parsing contract and byte order are unchanged.

## Tests

`InterpreterTest.LoadsWasmFromUnalignedInputBuffer` loads a valid module from `storage.data() + 1`. It passes normally and with ASan+UBSan (`detect_stack_use_after_return=1`).

## Follow-on

Other fixed-width parser reads should use the same copy-based pattern if they can consume caller-owned byte buffers.


lum1n0us (changes requested):
Established helpers for unaligned buffer reads (especially for wider types) already exist in *wasm_runtime_common.h* and are widely adopted across the codebase. Following the same convention here when parsing the magic header is recommended.

matthargett:
> Established helpers for unaligned buffer reads (especially for wider types) already exist in _wasm_runtime_common.h_ and are widely adopted across the codebase. Following the same convention here when parsing the magic header is recommended.

ok, I did some deeper research and testing on my local devices. I agree that values in the cell area should go through the PUT_*_TO_ADDR / GET_*_FROM_ADDR accessors. I checked whether routing the CSP through them would cover this, and I'm not sure it does:
1. The misalignment isn't limited to the CSP. wasm_interp_interp_frame_size() only rounds to 4 bytes, so an odd-sized frame leaves the next frame's WASMInterpFrame header (prev_frame, function, ip, csp_bottom, etc) misaligned. For example, the entry frame in wasm_interp_call_wasm is sized `ret_cell_num > 2 ? ret_cell_num : 2`, so calling an export with `(result i32 i32 i32)` misaligns the callee frame (unless I'm missing something?). That happens on main and on this PR as it stands, since the padding is relative to frame_lp. Header fields are plain struct accesses throughout the runtime, so accessors don't seem practical there. (again ,tell me if I'm missing something!)
2. On x64/AArch64 the accessors are plain *(T *) derefs. AFAICS that means they'd only change behavior on 64-bit targets built with WASM_CPU_SUPPORTS_UNALIGNED_ADDR_ACCESS=0 (e.g. RISC-V64), where they'd add split accesses to every PUSH_CSP/POP_CSP_N.

I also looked at how other interpreters handle this, and none keeps pointer-bearing records like WASMBranchBlock among packed 4-byte cells:
* V8's DrumBrake packs 4-byte slots like we do. But it keeps frame state off the slot stack, precompiles branches, and reads wide slots with memcpy-based helpers.
* wasm3, which also uses 32-bit slots by default, rounds 64-bit slots and callee frame offsets to even slots instead.
* JSC's IPInt, Pulley, wasmi and WasmEdge align frames and slots by construction.

my suggestion based on my understanding would be to fix this at the layout level:
* round wasm_interp_interp_frame_size() up to sizeof(void *) so every frame starts pointer-aligned. This covers both interpreters, and I'll keep the mirrored formula in aot_emit_function.c in sync.
* keep the CSP offset padded within the frame.

That leaves every access site unchanged, adds nothing to the hot paths, and costs at most 4 bytes per frame on 64-bit. With both changes, a local Apple A12 (iPhone XS) -fsanitize=alignment build is clean for odd-cell functions, internal calls, and the 3-cell-result entry scenarios. Since the UBSan preset builds with -fno-sanitize=alignment, I'll make the test assert frame and CSP alignment directly (unless you want me to change the alignment sanitizer to on as a goal?).

what do you think? am I missing some key detail or have my math wrong somewhere?  separately, if you'd like the accessors themselves to be well-defined on x86-64/AArch64, a fixed-size memcpy compiles to the same single load there (that's how DrumBrake does it). I'm happy to do that as a follow-up PR, if it's not antithetical to the "micro" platform goals :)

PR #5062, opened 2026-08-22
classic-interp: align branch control stack

matthargett:
## Problem

The classic interpreter places `WASMBranchBlock` entries immediately after a stack measured in 32-bit cells. An odd cell count leaves the control stack only four-byte aligned on 64-bit targets, causing undefined behavior when branch blocks containing pointers are accessed.

## Fix

Round the control-stack offset up to pointer alignment and include the padding in the interpreter-frame allocation.

## Tests

`InterpreterTest.ExecutesControlFlowWithOddCellStackLayout` runs a function with one i32 local and a nested block. It passes normally and with ASan+UBSan (`detect_stack_use_after_return=1`).

## possible Follow-on

Architectures with stricter aggregate alignment can extend the offset calculation to use an explicit branch-block alignment helper.


lum1n0us (changes requested):
Good catch. Code should consistently access CSP variables via `PUT/GET_xxx_TO/FRAME_ADDR`. Deviating from this adds unnecessary complexity and hurts maintainability

Issue #5068, opened 2026-08-22
ESP-IDF: libc-wasi build uses os_timespec as an integer type


78:
## Problem

With libc-wasi enabled, the bundled ESP-IDF example fails because `core/shared/platform/esp-idf/platform_internal.h` defines `os_timespec` as `int`, while the sandboxed system primitives use it as a `struct timespec`.

Representative errors from `locking.h` include:

```
error: field name not in record or union initializer
error: request for member 'tv_sec' in something not a structure or union
error: passing argument 3 of 'pthread_cond_timedwait' from incompatible pointer type
```

Reproduction environment:

- ESP-IDF commit `6a9c44fe7e725af45cb99293ae38afd7d481f1e3`
- GCC 15.2.0
- ESP32-P4
- bundled `product-mini/platforms/esp-idf` example
- libc-wasi enabled by the example defaults

This appears after the independent warning-option and POSIX declaration blockers in #5065 and #5066 are resolved.

## Possible fix

Use `struct timespec` for the ESP-IDF `os_timespec` definition, then validate the libc-wasi timestamp and condition-variable paths on supported ESP-IDF versions.

PR #5069, opened 2026-08-22
esp-idf: include required POSIX declarations

### LINKED ISSUE #5066: ESP-IDF 6.1: platform sources rely on removed transitive POSIX declarations
## Problem

The bundled ESP-IDF example does not compile against ESP-IDF 6.1 / GCC 15 because the ESP-IDF platform adapter relies on POSIX declarations that used to arrive through transitive includes.

After resolving the independent `-Wformat-security` option conflict reported in #5065, the build fails in `espidf_file.c` and `espidf_platform.c` with errors including:

```
error: invalid use of undefined type 'const struct stat'
error: implicit declaration of function 'fstatat'
error: implicit declaration of function 'renameat'
error: 'struct stat' declared inside parameter list
```

Reproduction environment:

- ESP-IDF commit `6a9c44fe7e725af45cb99293ae38afd7d481f1e3` (6.1 development branch)
- GCC 15.2.0
- ESP32-P4
- `product-mini/platforms/esp-idf`

## Proposed fix

Make the ESP-IDF platform adapter self-contained by including `<sys/stat.h>` in `platform_internal.h` and declaring the local `renameat()` fallback that is implemented in `espidf_platform.c` but used from `espidf_file.c`.

## Validation plan

Build the bundled ESP-IDF example for ESP32-P4 with warnings treated as errors. This is a declaration/header compatibility fix and does not change runtime behavior.

78:
Fixes #5066.

## Summary

- include `<sys/stat.h>` explicitly in the ESP-IDF platform header
- declare the local `renameat()` fallback implemented by `espidf_platform.c`
- avoid relying on POSIX declarations that ESP-IDF 6.1 no longer provides transitively

## Validation

Tested with the bundled `product-mini/platforms/esp-idf` example using:

- ESP-IDF commit `6a9c44fe7e725af45cb99293ae38afd7d481f1e3`
- GCC 15.2.0
- ESP32-P4
- the independent warning-option fix from #5067 applied locally

Before this change, `espidf_file.c` and `espidf_platform.c` fail on an undefined `struct stat` and implicit declarations of `fstatat`, `renameat`, and related POSIX functions. After this change, those ESP-IDF platform sources compile successfully. The example later reaches a separate existing libc-wasi `os_timespec` incompatibility tracked in #5068.

Formatting and checks:

- formatted with Espressif clang-format 21.1.3 using the repository style
- `git diff --check` passes
- `clang-tidy-diff.py` reports no directly matched compilation unit for this header-only change

No runtime behavior is changed.

### DIFF
--- a/core/shared/platform/esp-idf/platform_internal.h
+++ b/core/shared/platform/esp-idf/platform_internal.h
@@ -18,6 +18,7 @@
 #include <unistd.h>
 #include <pthread.h>
 #include <arpa/inet.h>
+#include <sys/stat.h>
 #include <sys/socket.h>
 #include <sys/uio.h>
 #include <dirent.h>
@@ -30,6 +31,11 @@
 #ifdef __cplusplus
 extern "C" {
 #endif
+
+/* Implemented by espidf_platform.c when ESP-IDF doesn't provide it. */
+int
+renameat(int old_dirfd, const char *old_path, int new_dirfd,
+         const char *new_path);
 
 #ifndef BH_PLATFORM_ESP_IDF
 #define BH_PLATFORM_ESP_IDF
--- a/core/shared/platform/esp-idf/platform_internal.h
+++ b/core/shared/platform/esp-idf/platform_internal.h
@@ -32,10 +32,12 @@
 extern "C" {
 #endif
 
-/* Implemented by espidf_platform.c when ESP-IDF doesn't provide it. */
+#if defined(__PICOLIBC__)
+/* Picolibc omits this declaration; espidf_platform.c defines the stub. */
 int
 renameat(int old_dirfd, const char *old_path, int new_dirfd,
          const char *new_path);
+#endif
 
 #ifndef BH_PLATFORM_ESP_IDF
 #define BH_PLATFORM_ESP_IDF


lum1n0us (commented):
#include <sys/stat.h> is fine (fixes struct stat, consistent with how the platform header centralizes POSIX includes).

Two concerns on the renameat declaration:

- fstatat coverage unverified. Both are gated by the same __ATFILE_VISIBLE macro in newlib, and renameat was implicit even with <stdio.h> included — so <sys/stat.h> may not declare fstatat (or the other at-functions used by espidf_file.c) either. Please confirm it does on 6.1, or declare the full set of at-functions espidf_platform.c stubs.
- Unconditional declaration is environment-dependent. On ESP-IDF versions where newlib does declare renameat (4.4/5.x), it becomes a duplicate declaration — legal while signatures match, but a latent conflict. Suggest guarding like NuttX (#if !defined(AT_FDCWD)) and testing on 4.4 / 5.x / 6.1.

Minor: the comment "when ESP-IDF doesn't provide it" doesn't match reality — the stub is defined unconditionally.

lum1n0us (changes requested):

PR #5078, opened 2026-08-28
windows: support native invocation across MSVC and MinGW architectures

cpunion:
## Summary

- add GNU/COFF native-invocation bridges for MinGW ARM64 (scalar and SIMD) and x86, following the dedicated MinGW x64 bridge established by #1753;
- model the Windows ARM64 ABI with eight independent integer and floating-point argument registers while retaining the four shared positional slots used by Windows x64;
- keep MSVC targets on MASM and select GNU assembly only for MinGW targets;
- make the standalone native-invocation test build and run on Windows and return failure when arguments or return values are corrupted;
- cover MSVC x64/x86/ARM64 and MinGW GCC/Clang x64/x86/ARM64 in Windows CI, including both ARM64 SIMD modes;
- use native Win32 definitions where MinGW's headers already provide them and avoid the POSIX-only time entry point that blocked native MinGW builds.

## Validation

- [Full fork CI run](https://github.com/cpunion/wasm-micro-runtime/actions/runs/33185446392): 55 checks passed, 0 failed, 0 pending. The other 50 pull-request counterparts were intentionally skipped by the fork's per-SHA dedup gate; the push run executed the real jobs.
- Windows 11 ARM64 VM:
  - MSVC x64, x86, and ARM64 native-invocation tests pass;
  - MinGW GCC x64/x86 and MinGW Clang x64/ARM64 tests pass;
  - ARM64 scalar and SIMD bridges both pass argument and return-value validation.
- macOS ARM64 scalar and SIMD standalone native-invocation tests pass as non-Windows regressions.
- LLVM 19 cross-assembly produces the expected COFF-ARM64 `invokeNative` symbols and decorated COFF-i386 `_invokeNative` symbol.
- actionlint, clang-format 21, and the repository coding-guideline workflow pass.


core/iwasm/common/wasm_runtime_common.c
```diff
@@ -5984,7 +5984,7 @@ wasm_runtime_invoke_native(WASMExecEnv *exec_env, void *func_ptr,
 #undef v128
 #endif
 
-#if defined(_WIN32) || defined(_WIN32_)
+#ifdef _MSC_VER
```
lum1n0us: This narrows the scope from all Windows platforms down to MSVC only, right? Is it on purpose or a typo?

core/iwasm/common/wasm_runtime_common.c
```diff
@@ -6000,6 +6000,10 @@ typedef union __declspec(intrin_type) __declspec(align(8)) v128 {
     || defined(BUILD_TARGET_RISCV64_LP64)
 typedef long long v128
     __attribute__((__vector_size__(16), __may_alias__, __aligned__(1)));
+#if defined(__MINGW32__) \
+    && (defined(BUILD_TARGET_X86_64) || defined(BUILD_TARGET_AMD_64))
+#define MINGW_X64_PASS_V128_BY_REFERENCE
```
lum1n0us, line 6005: I'm no Windows ABI expert. Could you help clarify if passing v128 by reference is a MinGW-specific behavior (for both GCC/Clang), or does MSVC follow the same calling convention?

.github/workflows/compilation_on_windows.yml
```diff
@@ -128,6 +129,156 @@ jobs:
...
+        shell: pwsh
+        run: |
+          foreach ($simd in "${{ matrix.simd_modes }}".Split(" ")) {
+            $buildDir = "build-invoke-native-simd-$simd"
+            cmake -S tests/standalone/test-invoke-native -B $buildDir `
+              -G "NMake Makefiles" `
+              -DWAMR_BUILD_TARGET=${{ matrix.target }} `
+              -DWAMR_BUILD_AOT=0 `
+              -DWAMR_BUILD_FAST_INTERP=1 `
+              -DWAMR_BUILD_SIMD=$simd `
```
lum1n0us, line 180: Not sure if `$simd` will reliably evaluate to 0 or 1 as expected. Could you double-check/verify its value under different build configurations?

lum1n0us (changes requested):
Needs a rebase against latest `main`.

PR #5083, opened 2026-09-02
zephyr: make platforms/zephyr a standalone application repo

lum1n0us:
Locate the runtime through `ZEPHYR_WASM_MICRO_RUNTIME_MODULE_DIR` instead of relative paths, split the manifest into west.yml (lists WAMR, for users) and let twister pick samples and platforms from the sample.yaml files instead of a CI matrix.


.github/workflows/compilation_on_zephyr.yml
```diff
@@ -164,8 +148,7 @@ jobs:
         # The same command reproduces this job locally with the Docker image,
         # without --no-docker. twister reports the verdict.
         run: |
-          python3 product-mini/platforms/zephyr/build_and_run.py \
-            --no-docker --sim ${{ matrix.sim }} ${{ matrix.sample }}
+          python3 product-mini/platforms/zephyr/build_and_run.py --no-docker
```
TianlongLiang, line 151: Do you think it's worth adding a `--all` option so that it's explicit that the default will run all

TianlongLiang (approved):

.github/workflows/compilation_on_zephyr.yml
lum1n0us: Twister orchestrates all the samples under *platform/zephyr/*. It recursively discovers test configs via *samples.yaml* or *testcases.yaml* to dynamically construct the platform matrix and decide between build-only vs. run targets. Leaving the default empty isn't about setting `-all` — it's about letting Twister do its magic out of the box.

PR #5090, opened 2026-09-04
Add pre-commit local checks

lum1n0us:
Add a staged coding guidelines mode, a clang-tidy diff helper, and a fail-fast pre-commit dispatcher for local best-effort checks.

> [!NOTE]
> currently, clang-tidy is using default compilation flags. It should be improved by flag-checks in https://github.com/wasm-micro-runtime/wasm-micro-runtime/pull/5064. Let's apply it in a new PR.


TianlongLiang (approved):
Checking the staged changes is a good idea, as it's more convenient to make sure the changes are formatted before committing.

PR #5100, opened 2026-09-14
Re-org compilation flags and their related macros

lum1n0us:
Compilation flags and the macros they control were only loosely coupled: a
few flags defaulted to on inside build-scripts/, some macros had no
pre-definition at all, and nothing stopped a new one from drifting. This
establishes the discipline and the checks that keep it.


lum1n0us:
@matthargett Instead of searching for the perfect macro combination, I went with a complete 1-wise macro matrix to cover the diff. This compiled and passed locally on my setup. Could you give it a run on your end?

build-scripts/config_common.cmake
```diff
@@ -95,19 +95,116 @@ if (NOT WAMR_BUILD_AOT EQUAL 1)
...
+if (WAMR_BUILD_WASM_SPEC2 EQUAL 1)
+  message ("     Wasm 2.0 preset enabled via WAMR_BUILD_WASM_SPEC2")
+  # The only configurable switches among the Wasm 2.0 proposals; Multi-value,
+  # Non-trapping float-to-int Conversions, Sign-extension Operators and
+  # Import/Export of Mutable Globals are always on and have no switch.
+  set (WAMR_BUILD_BULK_MEMORY 1)
+  set (WAMR_BUILD_REF_TYPES 1)
+  set (WAMR_BUILD_SIMD 1)
+  #TODO: SPEC2 should use SPEC1 as a base and enable additional features
+elseif (WAMR_BUILD_WASM_SPEC1 EQUAL 1)
```
TianlongLiang, line 189: Could the Wasm 1.0 preset explicitly turn off the post-1.0 features that the root entry point enables? This reproduces with a fresh build directory, so it does not depend on stale CMake cache state: run `cmake -S . -B /tmp/wamr-spec1 -DWAMR_BUILD_WASM_SPEC1=1`. Root `CMakeLists.txt:112-120` first executes `set(WAMR_BUILD_SIMD 1)` and `set(WAMR_BUILD_REF_TYPES 1)`, then `CMakeLists.txt:124` includes `build-scripts/runtime_lib.cmake`, whose line 210 includes this file in the same CMake variable scope. This SPEC1 branch only executes `set(WAMR_BUILD_BULK_MEMORY 0)` at lines 189-194. Consequently, the later `if(NOT DEFINED ...)` defaults at `config_common.cmake:368-378` do not reset REF_TYPES or SIMD, and the derived rule at lines 433-435 sees REF_TYPES=1 and enables CALL_INDIRECT_OVERLONG as well. The effective “Wasm 1.0” build is therefore `BULK_MEMORY=0`, but `SIMD=1`, `REF_TYPES=1`, and `CALL_INDIRECT_OVERLONG=1`. For example, code guarded by `#if WASM_ENABLE_SIMD` can compile and pass in the SPEC1 job even though that code would be absent from an actual minimal Wasm 1.0 configuration, so the preset can no longer detect an accidental dependency on SIMD. Could this branch set SIMD, REF_TYPES, and CALL_INDIRECT_OVERLONG to 0 explicitly, or could the preset be applied before the root entry-point defaults?

build-scripts/config_common.cmake
```diff
@@ -268,9 +393,32 @@ if (NOT DEFINED WAMR_BUILD_LIME1)
...
+
+if (NOT DEFINED WAMR_BUILD_APP_FRAMEWORK)
+  set (WAMR_BUILD_APP_FRAMEWORK 0)
+endif ()
+
+if (NOT DEFINED WAMR_BUILD_FUZZ_TEST)
+  set (WAMR_BUILD_FUZZ_TEST 0)
+endif ()
 
 if (WAMR_BUILD_LIME1 EQUAL 1)
```
TianlongLiang, line 423: I think the same entry-point leakage changes the meaning of the Lime1 profile here. Starting clean with `cmake -S . -B /tmp/wamr-lime1 -DWAMR_BUILD_LIME1=1`, root `CMakeLists.txt:112-120` has already set SIMD=1 and REF_TYPES=1. In this file, the compatibility default at lines 336-338 additionally sets BULK_MEMORY=1, while this Lime1 block only enables BULK_MEMORY_OPT, CALL_INDIRECT_OVERLONG, and EXTENDED_CONST_EXPR. The resulting profile is therefore the three documented Lime1 switches plus BULK_MEMORY, SIMD, and REF_TYPES; REF_TYPES also independently drives CALL_INDIRECT_OVERLONG through lines 433-435. That matters when this profile is used as a minimal-configuration check: a source that accidentally requires bulk-memory or SIMD still builds, so the check cannot demonstrate that Lime1 works with only the feature set described at lines 134-137. Could this block explicitly clear BULK_MEMORY, SIMD, and REF_TYPES before enabling the three Lime1 switches?

core/config.h
```diff
@@ -674,17 +691,21 @@ unless used elsewhere */
...
 /* Support AOT intrinsic functions which can be called from the AOT code
    when `--disable-llvm-intrinsics` flag or
    `--enable-builtin-intrinsics=<intr1,intr2,...>` is used by wamrc to
-   generate the AOT file */
+   generate the AOT file.
+   Derived from the running mode by build-scripts/config_common.cmake: on for
+   AOT, off otherwise. */
 #ifndef WASM_ENABLE_AOT_INTRINSICS
-#define WASM_ENABLE_AOT_INTRINSICS 1
+#define WASM_ENABLE_AOT_INTRINSICS 0
```
TianlongLiang, line 708: Could we migrate the non-CMake AOT builds before changing this fallback to 0? AliOS is a concrete example: its aos.mk enables AOT and includes aot_intrinsic.c, but it never defines WASM_ENABLE_AOT_INTRINSICS. Imagine compiling an application with wamrc using an intrinsic-enabled option and then loading that AOT file in the AliOS runtime. wamrc can emit references to the intrinsic helpers, while the runtime now sees this macro as disabled and does not expose the corresponding support; the failure is deferred until the AOT module is loaded rather than appearing at build time. I think we should either add the explicit definition to every non-CMake AOT build in this PR or keep the old fallback until those build systems have been migrated.

tests/unit/unsupported-features/CMakeLists.txt
```diff
@@ -3,36 +3,28 @@
...
-if(NOT DEFINED WAMR_BUILD_FAST_JIT)
-  set(WAMR_BUILD_FAST_JIT 0)
+# A fake target to satisfy runtime_lib.cmake's requirement for a running mode.
+if(NOT DEFINED WAMR_BUILD_INTERP AND NOT DEFINED WAMR_BUILD_AOT)
+  set(WAMR_BUILD_INTERP 1)
 endif()
 
-include(../unit_common.cmake)
-add_library(unsupported_features_tests ${WAMR_RUNTIME_LIB_SOURCE})
+include(${WAMR_ROOT_DIR}/build-scripts/runtime_lib.cmake)
```
TianlongLiang, line 27: Would you mind making this test verify the specific unsupported-feature diagnostic rather than accepting any failure through WILL_FAIL? For example, on a machine without the expected LLVM installation, the LLVM-JIT case stops at “Cannot find LLVM dir.” CTest sees the nonzero result and marks the test as passed, even though runtime_lib never evaluated the invalid feature combination. I saw the same false positive when the fast-JIT case failed while downloading asmjit. In both situations CI is green for exactly the wrong reason, so a future regression in the validation would go unnoticed. A configure-only probe that matches the expected error text, or an equivalent PASS/FAIL regular expression, would distinguish the intended rejection from setup failures.

ci/clang_tidy_diff_check.py
```diff
@@ -49,46 +96,93 @@ def get_staged_paths(root: Path) -> list:
...
+def resolve_build_dirs(root: Path) -> list:
+    """The (preset, build directory) pairs to scan with, in report order.
+
+    Every preset is re-configured on every run.  A compile DB left over from
+    an earlier commit describes the source files and feature flags of that
+    commit, and a change to either is exactly the kind of thing this check is
+    supposed to notice."""
+    build_dirs = []
+    for preset in PRESETS:
+        if not generate_preset_compile_commands(root, preset):
```
TianlongLiang, line 176: Could a failed required preset make the check fail instead of being skipped? Suppose a change breaks configuration only for the AOT preset, or that preset loses a required dependency. This exception is caught, the script continues with the remaining presets, and the job can finish successfully. The files or conditional branches unique to AOT then receive no clang-tidy analysis, even though that is likely the configuration where the change needs scrutiny most. If some presets are intentionally optional, could we identify those explicitly and propagate failures from the rest?

ci/clang_tidy_diff_check.py
```diff
@@ -167,84 +354,166 @@ def run_clang_tidy_diff(
...
+        path = root.joinpath(source).resolve()
+        lines = get_changed_lines(get_diff(root, commits, [source]))
+        if not lines:
+            continue
 
-    build_dir = platform_dir.joinpath("build-clang-tidy")
-    if not generate_compile_commands(root, build_dir):
-        return True
+        changed_lines[source] = lines
+        if not any(path in db for _, _, db in compile_dbs):
```
TianlongLiang, line 427: Could we avoid treating “not present in any compilation database” as success for a changed source file? Take samples/file/src/main.c from this PR: none of the generated minimum-mode databases owns it, so the script returns the skip result and the workflow maps that to exit 0. If a later edit introduces a clang-tidy finding in that file, the check still reports green without ever invoking clang-tidy on it. That makes it difficult to distinguish “clean” from “not checked.” I think an unmatched changed C/C++ file should either fail with a message naming the missing build configuration, or its owning sample configuration should be added to the database set.

PR #5103, opened 2026-09-15
 add a gcovr toolchain for the spec and unit test suites

lum1n0us:
Replace the lcov/genhtml chain (collect_coverage.sh) with gcovr-based coverage
measurement for the spec and unit suites.

The regression build_run.py gains --llvm-dir, optional and defaulting to the
bundled LLVM build.


tests/wamr-test-suites/test_wamr.sh
```diff
@@ -967,38 +964,36 @@ function build_wamrc()
...
+        local COV_REPORT_DIR=${WORK_DIR}/coverage-report
+        local BUILD_DIRS=()
 
-        CODE_COV_FILE=""
-        if [[ -z "${CODE_COV_FILE}" ]]; then
-            CODE_COV_FILE="${WORK_DIR}/wamr.lcov"
-        else
-            CODE_COV_FILE="${CODE_COV_FILE}"
-        fi
+        rm -fr ${COV_REPORT_DIR}
```
TianlongLiang, line 971: Could we preserve the coverage produced by earlier running-mode passes here? With a normal `./test_wamr.sh -C` run, classic-interp writes `coverage-report`, then fast-interp deletes and replaces it, and each following mode does the same; by the end, the apparently complete report contains only the final mode. The smaller `-t jit` case has the same issue inside one mode: eager JIT is collected first, this removal happens again, and lazy-JIT replaces it. The previous lcov path appended each tracefile, so this is a behavioral regression that can make overall coverage look lower or hide paths exercised only by an earlier mode. Could we keep per-pass gcovr data and merge it, or collect once after all requested passes have finished?

tests/wamr-test-suites/coverage/run_coverage.py
```diff
@@ -0,0 +1,796 @@
...
+    # 'jit' (spec/test_wamr.sh naming) is an alias for 'llvm-jit' (unit-test
+    # naming); normalize early so fingerprints and build dirs are stable.
+    args.mode = ["llvm-jit" if m == "jit" else m for m in args.mode]
+
+    # Each --report starts a new group; --mode, --spec and --feature are paired
+    # by position (mode[i], spec[i], feature[i]); defaults fill the rest.
+    reports = []
+    for i, name in enumerate(args.report):
+        combo = {
+            "mode": args.mode[i] if i < len(args.mode) else "classic-interp",
```
TianlongLiang, line 766: Could we bind mode/spec/feature options to the `--report` they follow rather than pairing four independent lists by index? For example, a user may run `--report default --report gc --spec=-G`, intending only the second report to enable GC. argparse produces two report names but only one spec value, so this loop assigns `-G` to `default` and gives `gc` the empty default. Both runs can complete successfully, but their directory names and fingerprints describe the opposite configurations from what the user requested. Omitting a mode or feature in the middle causes the same left-shift for every later report. Parsing report groups in argument order, or accepting one structured report specification, would prevent silently mislabeled coverage.

tests/wamr-test-suites/coverage/run_coverage.py
```diff
@@ -0,0 +1,796 @@
...
+    collected from (the unit build dirs, and for the spec layer a copy of the
+    gcov files of test_wamr.sh's in-place product build), so the merge
+    re-collects the union of those dirs."""
+    merged_out = os.path.join(out_dir, "_merged")
+    os.makedirs(merged_out, exist_ok=True)
+    build_dirs = []
+    for report in reports:
+        work = os.path.join(out_dir, "_work", report)
+        if os.path.isdir(work):
+            for d in os.listdir(work):
```
TianlongLiang, line 616: Could the merge consume an explicit list of artifacts from the report being merged instead of every directory left under `_work/<name>`? One realistic sequence is to run a report named `nightly` in classic-interp mode, then reuse the same name for an AOT run. `run_report()` does not clear the old work directory, so both `spec-coverage-classic-interp` and `spec-coverage-aot` remain. A later `--merge nightly` walks this directory and includes both, even though the current `nightly` report represents only the AOT invocation. The merged percentage may increase and look healthy, but it now includes stale execution data and cannot be reproduced from the current command. Clearing the work directory at the start, keying it by fingerprint, or recording the exact current build directories in a manifest would keep the merge attributable to one run.

tests/wamr-test-suites/coverage/run_coverage.py
```diff
@@ -0,0 +1,796 @@
...
+                   f"{repo_relative(log_path)}")
+        print(f"      FAILED (rc={status}); collecting the data it did produce")
+
+    if not os.path.isdir(build_dir):
+        print("      WARNING: spec iwasm build dir not found: "
+              f"{repo_relative(build_dir)}")
+        return None, failure
+    snapshot = os.path.join(workdir, f"spec-coverage-{mode}")
+    copied = snapshot_coverage_data(build_dir, snapshot)
+    if not copied:
```
TianlongLiang, line 300: Could the absence of gcov artifacts be recorded as a failed step rather than only a warning? For example, CMake and the spec tests may return 0 while a custom compiler ignores the GCC coverage flags, or a configuration change may stop forwarding COLLECT_CODE_COVERAGE. The build directory then exists but contains no `.gcno`/`.gcda` files, so this returns `(None, None)`. In a spec-only run, `collect()` has nothing to do, the runner still writes fingerprint metadata, and the process exits 0 without `index.html` or a coverage summary. An automation job would report success even though it measured nothing. Please return a failure message here—and for a missing expected build directory—so the partial-report machinery produces `failures.txt` and a nonzero exit status.

tests/wamr-test-suites/coverage/README.md
```diff
@@ -0,0 +1,316 @@
...
+cd tests/regression/ba-issues
+
+# build + run every active case (runtimes auto-derived from running_config.json)
+python3 build_run.py
+
+# only classic-interp cases
+python3 build_run.py --mode classic-interp
+
+# build with gcov data (-DCOLLECT_CODE_COVERAGE=1)
+python3 build_run.py --coverage
```
TianlongLiang, line 299: Am I missing a companion change for the `--llvm-dir` behavior mentioned in the PR description? `tests/regression/ba-issues/build_run.py` is unchanged: its argument parser does not accept `--llvm-dir`, and `build_wamrc()` invokes CMake without forwarding `-DLLVM_DIR`. A developer whose LLVM package lives outside the bundled default will therefore try the advertised option and get an argparse error; without the option, CMake may fail to find LLVM or pick up an unintended installation. Could we add and thread the option through the wamrc configuration, or update the PR description if regression tests are intentionally outside this change?

PR #5109, opened 2026-09-20
test: add Zephyr coverage measurement pipeline

TianlongLiang:
<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


.github/workflows/compilation_on_zephyr.yml
lum1n0us: How about we pass `--coverage` here and upload the coverage data right after *Summarize twister results*? Also, I'm not entirely sure we need a standalone report for this—perhaps a unified dashboard entry is enough?

like https://github.com/wasm-micro-runtime/wasm-micro-runtime/actions/runs/35626127074

